PR #188, opened 2015-06-24
Allow size_t from multiple headers.

bungeman:
According to the C language specification, size_t is defined by all
of stddef.h, stdio.h, stdlib.h, string.h, time.h, and wchar.h. Allow
size_t to be provided by any of these.

The tests must be updated to reflect this change. Currently cstdlib
and time.h are being used to obtain NULL, but these also now provide
size_t, which interferes with tests for size_t. Fortunately, locale.h
provides NULL but not size_t.

### DIFF
--- a/gcc.symbols.imp
+++ b/gcc.symbols.imp
@@ -93,6 +93,11 @@
   { symbol: [ "useconds_t", private, "<unistd.h>", public ] },
   # glob.h seems to define size_t if necessary, but it should come from stddef.
   { symbol: [ "size_t", private, "<stddef.h>", public ] },
+  { symbol: [ "size_t", private, "<stdio.h>", public ] },
+  { symbol: [ "size_t", private, "<stdlib.h>", public ] },
+  { symbol: [ "size_t", private, "<string.h>", public ] },
+  { symbol: [ "size_t", private, "<time.h>", public ] },
+  { symbol: [ "size_t", private, "<wchar.h>", public ] },
   # Macros that can be defined in more than one file, don't have the
   # same __foo_defined guard that other types do, so the grep above
   # doesn't discover them.  Until I figure out a better way, I just
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -171,7 +171,14 @@ const IncludeMapEntry libc_symbol_map[] = {
   { "useconds_t", kPrivate, "<sys/types.h>", kPublic },
   { "useconds_t", kPrivate, "<unistd.h>", kPublic },
   // glob.h seems to define size_t if necessary, but it should come from stddef.
-  { "size_t", kPrivate, "<stddef.h>", kPublic },
+  // It is unspecified if the cname headers provide ::size_t.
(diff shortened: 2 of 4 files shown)


kimgr:
This looks good to me. The <locale> thing is a little weird, but so were the prior includes.

Which platforms did you test on? I can run this on FreeBSD and WIndows just to make sure nothing weird crops up.

kimgr:
Tests on FreeBSD and Windows were failing before this change, but now they're faiilng slightly differently.

I have diffs between the original failures and failures with your patch, I put them on Pastebin here: http://pastebin.com/PZaP4bBc (diffs of diffs, yay!)

I can't quite make out if these changes are significant, but the patch seems to affect the `wchar` header, and it looks like NULL is first seen in `stdio.h` on FreeBSD.

bungeman:
I tested this on Ubuntu Trusty. The main issue I was running into is that this existing test is trying to do so much... but on the other hand, that's somewhat realistic as well.

Before it was being reported that stddef.h should be included because that was the only stated source of size_t. Now stdio.h is being suggested because it is marked as providing size_t (which it does) but it is also indirectly included, overriding the suggestion of stddef.h. This is taken care of in the change by changing the test regex to match any of the files which may provide size_t. I think this is a more general issue as to how to report when a symbol may have multiple correct sources.

Windows suggesting wchar.h for 'stdout' is just a 'feature' in their library implementation, it isn't supposed to provide it directly. They should be undef'ing the 'stdin', 'stdout',' and 'stderr' macros they define there if they defined them in that header for the forward declarations (or just give them internal names). I'm sure this can't be fixed for compatibility reasons, but a '.imp' file for VC++ headers should mark these as private coming from this header.

So yes, aside from the non-conformance of the Window's headers themselves, the output referenced in the above comment is what I would expect.

kimgr:
Thank you! I'm assuming the test suite still runs green on Ubuntu/GCC/libstdc++ (IWYU's 'home' platform).

Agree on badinc, that test is hard to please. We try to actively migrate details out of it into dedicated feature tests instead. But I agree that the big bang integration test is also useful, we'll see if we can make it less attached to system headers.

I'm not too worried about spurious failures in MSVC and FreeBSD/clang/libc++ since we haven't managed to sort correct mappings yet. I'm not sure if there's a strict method to the current mappings system, but I think ideally we should have a 'preferred' header for a symbol, and then a set of 'allowed' providers, so we get the preferred unless one of the allowed is already provided.

From my viewpoint this patch is good as-is, but I'd like to leave it to vsapsai to test it on his OSX environment too.

vsapsai:
Everything looks fine on Mac OS X (things are failing in the same way as earlier).

Also checked if command `grep -R '__.*_defined' /usr/include | perl -nle 'm,/usr/include/([^:]*):#\s*\S+ __(.*)_defined, and print qq@    { "$2", kPublic, "<$1>", kPublic },@' | sort -u` will discover more size_t definitions, but it doesn't discover any size_t definitions at all.

kimgr:
Thanks for the patch, now merged!

Issue #127, opened 2015-06-15
IWYU and Static members


vsapsai:
Originally reported on Google Code with ID 127
```

There seems to be some issue on how to handle static members, when making the report.
I have attached two different cases that do the same but through different ways.

Both samples will declare a static member inside a class by forward declaring the type.
Then on the cpp of that class they will invoke some method on the static member.

1) Plain: In this sample we declare a static member directly inside ZMyTestClass by
forward declaring the type. If you run iwyu notice how it suggests to remove the forward
declaration and replace it by the include itself, while it says that the cpp is correct.
It's not needed, right?

2) ThroughMacro: This sample is closer to what I have here. We have a header that defines
some macros that are used in many classes. Some macros are used for the declaration
and some used for the definition. In those cpp where definition macros are used we
need to include the corresponding header (in this case, myutil.h). The report is suggesting
to remove myutil.h, which doesn't compile.

What do you think?
```

Reported by `dpunset` on 2014-04-14 19:54:05

<hr>
 * *Attachment: [Plain.rar](https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-127/comment-0/Plain.rar)*
 * *Attachment: [ThroughMacro.rar](https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-127/comment-0/ThroughMacro.rar)*

kimgr:
Something messed up with locations in `<scratch space>`, I'm hoping to get it fixed. Here's the failure from badinc.cc:

    <scratch space>:254: Unexpected diagnostic:
    kI1ConstInt is defined in "tests/cxx/badinc-i1.h", which isn't directly #included.

    <scratch space>:255: Unexpected diagnostic:
    I1_Class needs a declaration, but does not provide or directly #include one.

    tests/cxx/badinc.cc:1452: Unmatched regex:
    kI1ConstInt is...*badinc-i1.h

    tests/cxx/badinc.cc:1454: Unmatched regex:
    I1_Class needs a declaration

(see http://buildbot.include-what-you-use.com:8010/builders/ubuntu-builder/builds/403/steps/test%20IWYU/logs/stdio)

conchurnavid:
This change causes #334

PR #189, opened 2015-06-27
Fix issue #127: new rules for uses inside macros

kimgr:
This isn't 100% finished, but I wanted to show it off and get some feedback. This patch as-is *should* fix issue 127, but I'd like to expand tests a little bit to cover a few corner cases.

What do you think of the overall approach? I'm not all that happy with the details (the pair returned from GetUseLocationForMacroExpansion, for example), but I think the general direction is right.

See commit comment for more details.


vsapsai:
Can you please describe in general what went wrong in issue #127? I see you are changing how we attribute decls in macro, to caller or to macro author. But I am missing why are you doing this. A small test case that demonstrates the problem will be very useful.

kimgr:
The original problem as reported was a construct like this:

    // util.h
    class Util {
    public:
      void Foo() {
         // Interesting stuff here, I'm sure.
      }
    };

    // macros.h
    #define DECLARE_UTIL(CLASSNAME) \
        static class Util ms_util; \
        void Foo();

    #define DEFINE_UTIL(CLASSNAME) \
        Util CLASSNAME::ms_util; \
        void CLASSNAME::Foo() { \
          ms_util.Foo();  \
        }

(this is cut down from the attachment from issue #127 on Google Code.)

The original problem was that they wanted to push the responsibility for `util.h` onto clients. This makes sense because using `DECLARE_UTIL` does not require `util.h` to be included, but using `DEFINE_UTIL` does, and it seems intuitive to keep the macros together in the same file.

So I suggested the same heuristic for author intent that we use for templates and typedefs: if the macro file forward-declares a type, responsibility for that type should be pushed onto files instantiating the macro.

In trying to understand IWYU's current handling of macro locations I thought it committed to the location too early (in `CurrentLoc`), so I delayed that decision until we were ready to report the use. That way I could include any forward-declares in the analysis, but I also had to re-work the handling of macro arguments (which worked magically before.)

I'm now pretty convinced that the stuff I'm doing with macro arguments is wrong, and I'd like to explore that further, but hopefully this sheds some light on the background. Any ideas on how to document this better in code would be welcome.

vsapsai:
Delaying macro location from `CurrentLoc` to `GetUseLocationForMacroExpansion` seems to be fine. Not sure about returning `isForwardDeclaredInMacroFile` as the second part of a pair. It is set only in one place inside `GetUseLocationForMacroExpansion`, though it has multiple branches. And it is used only in one place. So I think it is cleaner not to return a pair. The only benefit I see is unknown performance improvement. `GetUseLocationForMacroExpansion` will calculate `isForwardDeclaredInMacroFile` anyway so if it is too heavy, recalculating it in other places can be too expensive.

Is all the magic with macro arguments finished and ready for review? Because all this `FindNextIdentifier` looks tricky.

kimgr:
Yes, the reason I returned the pair was because it seemed unnecessary to calculate the same thing more than once, but it does muddy up the logic, so I'll try to kill that darling to get something that's easier to understand.

No, I think I'm going about the macro arguments the wrong way. Here's an example where it will break, for example:

    // macros.h
    #define APPLY(a, op, b) \
        a op b

    // fiasco.h
    int operator+(int a, int b) {
      return a * b;
    }

    // main.cc
    #include "macros.h"
    #include "fiasco.h"

    void foo() {
       int c = APPLY(100, +, 200);
    }

Because the operator comes in as a macro argument, its use should be attributed to `main.cc`. But the current code is just looking for identifiers, so it can never work.

I saw test failures with the classical array-size macro:

    #define ARRAY_SIZE(x) sizeof(x) / sizeof(*x)

Something caused the use-location for `x` to point somewhere else, that's why I started searching through the macro body with `FindNextIdentifier`. Now I hope I can adjust the use-location so it points to the right thing, instead, and thereby solve the operator-as-argument case as well.

So, long story short, no, I don't think the macro argument magic is ready for review, I think it's broken :-)

kimgr:
Abandoning this, I have a new branch with a working approach.

PR #193, opened 2015-07-14
Assume override keyword support.

kimgr:
We're past the time where `override` is not a supported keyword across all compilers necessary to compile LLVM.

This patch retires `IWYU_OVERRIDE` and just uses the keyword outright.


vsapsai:
Looks good to me.

PR #195, opened 2015-07-19
Update to reflect changes in Clang.

kimgr:
Clang r242176 disables exceptions by default for MSVC. They can be
re-enabled by passing `Xclang -fcxx-exceptions -Xclang -fexceptions` on
the command-line.

This patch does that for the only test that uses try/catch.

### DIFF
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -81,7 +81,8 @@ class OneIwyuTest(unittest.TestCase):
     clang_flags_map = {
       'alias_template.cc': ['-std=c++11'],
       'auto_type_within_template.cc': ['-std=c++11'],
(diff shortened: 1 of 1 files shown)


vsapsai:
Do you need `-Xclang`? For me `-fcxx-exceptions -fexceptions` works and `-fno-cxx-exceptions -fno-exceptions` breaks the test.

kimgr:
Huh, yeah, that seems cleaner. I was working from the e-mail thread tied to the commit that changed this:
https://www.mail-archive.com/cfe-commits@cs.uiuc.edu/msg123866.html

kimgr:
Oh, they're talking in the context of `clang-cl`, that's why they need `-Xclang`. I'll update the PR, thanks for noticing!

vsapsai:
LGTM. Sorry, I haven't noticed the PR was updated.

Issue #166, opened 2015-06-15
ext/alloc_traits.h suggested when indexing std::vector<std::string>


vsapsai:
Originally reported on Google Code with ID 166

```
(extracted from issue #163)

--
../hc/hc.cpp:1154:24: warning: __gnu_cxx::__alloc_traits<std::allocator<std::basic_string<char>
> >::value_type is defined in <ext/alloc_traits.h>, which isn't directly #included.
../hc/hc.cpp:1215:31: warning: __gnu_cxx::__alloc_traits<std::allocator<std::basic_string<char>
> >::value_type is defined in <ext/alloc_traits.h>, which isn't directly #included.
../hc/hc.cpp:1219:38: warning: __gnu_cxx::__alloc_traits<std::allocator<std::basic_string<char>
> >::value_type is defined in <ext/alloc_traits.h>, which isn't directly #included.

../hc/hc.cpp should add these lines:
#include <ext/alloc_traits.h>           // for __alloc_traits<>::value_type
--

IWYU should not suggest including ext/alloc_traits, which is a private header.

> ../hc/hc.cpp:1154:24: ...
> ../hc/hc.cpp:1215:31: ...
> ../hc/hc.cpp:1219:38: ...

All these three point to indexing into a vector of strings;

  std::vector<std::string> title;
  ...

  title[count][len] = '\0'; // <- alloc_traits is required here.

(see https://github.com/LegalizeAdulthood/iterated-dynamics/blob/master/hc/hc.cpp#L1154,
line numbers are off by one, so I'm probably linking to the wrong revision, but this
is good enough)

IWYU log attached.
```

Reported by `kim.grasman` on 2015-01-03 12:17:39

<hr>
- _Attachment: [verbose-7.txt.gz](https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-166/comment-0/verbose-7.txt.gz)_

vsapsai:
It's not obvious in this issue but pull request #214 contains a few test cases that are still failing. Full fix for the issue requires addressing those failing tests.

kimgr:
I can no longer repro this on mainline as of 205ef58df95e95d92f1d59e897398263780b8883. Closing as fixed.

Issue #163, opened 2015-06-15
tells me I need ext/alloc_traits.h but not why


vsapsai:
Originally reported on Google Code with ID 163

```
The full include-list for ../hc/hc.cpp:
#include <ctype.h>                      // for isalpha, isdigit
#include <ext/alloc_traits.h>
#include <fcntl.h>                      // for open, O_RDONLY, O_RDWR
#include <stdio.h>                      // for printf, fwrite, fclose, etc
#include <stdlib.h>                     // for atoi, malloc, qsort, exit, etc
#include <string.h>                     // for strlen, strcpy, strcmp, etc
#include <unistd.h>                     // for close, read, lseek, unlink, etc
#include <algorithm>                    // for min
#include <cassert>                      // for assert
#include <cstdarg>                      // for va_list, va_end, va_start
#include <string>                       // for string
#include <vector>                       // for vector
#include "helpcom.h"                    // for PD_INFO, setint, etc
#include "port.h"                       // for BYTE, VOIDPTR, VOIDCONSTPTR
#include "unix.h"                       // for strlwr, O_BINARY, _MAX_EXT, etc

See the source code here:
<https://github.com/LegalizeAdulthood/iterated-dynamics>
changeset 97459e23a88337fd66eb9f0e3739d85102241c55

```

Reported by `legalize@xmission.com` on 2014-12-22 22:54:53

kimgr:
This has been fixed.

PR #196, opened 2015-07-21
Fix issue 163: briefer why comments.

kimgr:
Instead of reserving the last 5 chars for ', etc' in why comments, just
truncate the last symbol with '...'. This generally gives more relevant
output.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -72,53 +72,53 @@ class OutputLine {
   explicit OutputLine(const string& line)
       : line_(line) {}
   OutputLine(const string& line, const vector<string>& symbols)
-      : line_(line), symbols_(symbols) {}
+      : line_(line),
+        symbols_(symbols) {
+    symbols_.erase(std::remove_if(symbols_.begin(), symbols_.end(),
+                                  [](const string& v) { return v.empty(); }),
+                   symbols_.end());
+  }
 
   size_t line_length() const { return line_.size(); }
   bool needs_alignment() const { return !symbols_.empty(); }
   void add_prefix(const string& prefix) { line_ = prefix + line_; }
-  string printable_line(size_t desired_length, size_t max_length) const;
+  string printable_line(size_t min_length, size_t max_length) const;
 
  private:
   string line_;                     // '#include XXX' or 'class YYY;'
   vector<string> symbols_;          // symbols used from included header
 };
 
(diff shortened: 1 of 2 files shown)


kimgr:
ref issue #163

LegalizeAdulthood:
Try running it on the code referenced in that issue.  If you get some reasonable sense of why it wants to include `<ext/alloc_traits.h>`, or even better, it no longer nags me that I need to include this implementation detail header, then it's fixed.  I'm a user of iwyu, not a developer, so having me review the code is rather pointless.  If this changeset fixes the issue, great, otherwise, the issue is still open.

kimgr:
Sorry, yes, I meant if you could run a patched IWYU against the code in question, and get the behavior you expect. I don't have a gcc/libstdc++ environment where I can reproduce the original issue.

The problem with the private symbol leaking out is not fixed, only the disappearing "// for ..." comment. We pulled the functional bug out into a separate issue, #166.

vsapsai:
First of all, I don't like the name `Truncate`, it is too generic. Sorry, right now I don't have a better suggestion.

And now let's get to edge cases. Can we get `  // f...`? I.e. part of "for" and space are truncated. It looks pretty ugly.

Can we get something like `  // for T...`? It is ugly, but we can live with it and I'm not sure it is bad. When it is the only visible symbol, it is good to indicate there is a truncated string. But if we have something like `  // for vector, remove_if, T...`, then "etc." looks nicer.

I've noticed you increased `max_length` by 1. Have you checked that clang-format excludes new line characters when enforces line length limit? It shouldn't be an issue, just need to check it.

kimgr:
> I've noticed you increased max_length by 1. Have you checked that clang-format excludes
> new line characters when enforces line length limit? It shouldn't be an issue, just need to check it.

Oops, no the convention seems to be to put the newline at col 80, so the max width should be 79, as it was.

As for the rest, I think they are all possible, depending on how long your include lines are. I wish we had test coverage for this, so it was easier to experiment.

I'll come back with an updated patch, thanks for reviewing!

kimgr:
Oops, I accidentally killed this. Updated in new PR #204.

I addressed everything but this comment

> But if we have something like // for vector, remove_if, T..., then "etc." looks nicer.

It could probably be made to work, but I didn't want to introduce even more complexity for the sake of this corner case.

PR #197, opened 2015-07-23
Clang in cl-mode now supports -fsyntax-only.

kimgr:
Per http://lists.cs.uiuc.edu/pipermail/cfe-commits/Week-of-Mon-20150413/127452.html.

Remove special-casing with /Zs.
Adjust test not to pass invalid args (/C now requires /P)

### DIFF
--- a/iwyu_driver.cc
+++ b/iwyu_driver.cc
@@ -179,11 +179,7 @@ CompilerInstance* CreateCompilerInstance(int argc, const char **argv) {
   // FIXME: This is a hack to try to force the driver to do something we can
   // recognize. We need to extend the driver library to support this use model
   // (basically, exactly one input, and the operation mode is hard wired).
-  driver.ParseDriverMode(args);
-  if (driver.IsCLMode())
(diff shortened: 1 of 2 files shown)


vsapsai:
Looks good to me. The only issue is I don't understand the purpose of `/GF` flag for clmode.cc.

kimgr:
I'm just passing some random cl.exe args to make sure we don't choke. Clang-cl added validation for `/C`, requiring that it's used together with `/P`, which in turn writes preprocessed output to a file (I think this is `-E` in GCC terms). To avoid polluting the disk with preprocessed output, I just picked another non-controversial flag, `/GF` (Enable string pooling).

There's a comment about this in `clmode.cc`, maybe I should add something similar to `run_iwyu_test.py`?

vsapsai:
Thanks for explanation. I don't think extra comments in `run_iwyu_test.py` are required. I was looking for some secret meaning in selected flags but there is no secret meaning, stupid me.

Have no objections regarding the change.

PR #198, opened 2015-07-23
Make run_iwyu_tests.py more Pythonic.

kimgr:
Python 3-compatible subprocess output collection.
Use python on PATH.

### DIFF
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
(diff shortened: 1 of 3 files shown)


vsapsai:
Subprocess communication works fine on Python 2.7 (which is our min supported version).

If you are fixing Python path, we are still using `/usr/bin/python` in iwyu_test_util.py, make_readme.py. It doesn't matter much, but for consistency it would be nice to update these files too.

kimgr:
Done, thanks for noticing!

Issue #174, opened 2015-06-15
using declarations of overloaded functions greedily add unnecessary includes


vsapsai:
Originally reported on Google Code with ID 174
```
What steps will reproduce the problem?  Give the *exact* arguments passed
to include-what-you-use, and attach the input source file that gives the
problem (minimal test-cases are much appreciated!)

1. Add a using declaration for something like std::swap

#include <vector>
#include <list>

void use_overload() {  
  std::vector<int> a(10);
  std::vector<int> b(11);
  using std::swap;
  swap(a, b);
}

2. Run IWYU on this file using default options

What is the expected output? What do you see instead?

I expect IWYU to tell me I can remove #include <list>.  Instead I get the following
output;

using_overload.cc should add these lines:
#include <algorithm>                    // for swap

using_overload.cc should remove these lines:

The full include-list for using_overload.cc:
#include <algorithm>                    // for swap
#include <list>                         // for swap
#include <vector>                       // for swap, vector


What version of the product are you using? On what operating system?

Ubuntu 14.10 / Clang 3.5 / IWYU Svn Rev. 599


Please provide any additional information below.

I tracked down the problem to how IWYU internally handles using declarations.  What
happens is IWYU finds all of the declarations that a using statement could be referencing
(essentially anything with that name). Those names are then added as forward declarable.
However, in post processing, IWYU converts all function references to needing to be
fully included, not forward declarable.  This makes sense, but it does mean that all
overloads of functions accessed through a using declaration will always need to be
#included.  Attached is a test to catch this plus my attempt at a fix.  The fix is
outlined as follows;
- Change adding all references of a using declaration from immediate to deferred
- Detect when something is accessed though a using declaration and add it as forward
declarable.
- If a using declaration is never referenced, arbitrarily add one of the symbols it
is referencing so that we don't remove all headers it refers to and break compilation.

This fix is not small and contains a questionable check at line 2577 of iwyu.cc that
I think warrants some discussion.

Note:  This patch was made against my github fork, if you want one against svn I can
do that.
```

Reported by `c.d.glover` on 2015-02-07 15:55:54

<hr>
 * *Attachment: [using_decl_fix.patch](https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-174/comment-0/using_decl_fix.patch)*

cdglove:
As the original submitter of this issue, I wonder if it makes more sense to convert this (and possibly others) from a patch to a pull request now that the project is on git hub.

[...]

PR #199, opened 2015-07-24
Add support for using declarations

cdglove:
Add support for using declarations such that "using std::swap;" does not cause all std::swap headers to be included.

This was caused by the fact that IWYU marks all function references as requiring a full include (ie:  not forward declarable).  This makes sense for normal functions but not in the context of using declarations.

This change alters the logic as follows;

- Change adding all references of a using declaration from immediate to deferred
- Detect when something is accessed though a using declaration and add it as forward
declarable.
- If a using declaration is never referenced, arbitrarily add one of the symbols it
is referencing so that we don't remove all headers it refers to and break compilation.

The change also adds two tests to try to catch the problem.  The tests fail prior to the change and pass after the change.

Resolves issue #174

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1601,13 +1601,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // the namespace of the decl.  If we have 'using std::vector;' +
     // 'std::vector<int> foo;' we don't actually care about the
     // using-decl.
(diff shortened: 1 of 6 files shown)


cdglove:
I just realized that for some reason this change put the tests into the wrong directory.  I will fix that and reissue the pull request.

PR #200, opened 2015-07-25
Add support for using declarations

cdglove:
Add support for using declarations such that "using std::swap;" does not cause all std::swap headers to be included.

This was caused by the fact that IWYU marks all function references as requiring a full include (ie: not forward declarable). This makes sense for normal functions but not in the context of using declarations.

This change alters the logic as follows;

Change adding all references of a using declaration from immediate to deferred
Detect when something is accessed though a using declaration and add it as forward declarable.
If a using declaration is never referenced, arbitrarily add one of the symbols it is referencing so that we don't remove all headers it refers to and break compilation.
The change also adds two tests to try to catch the problem. The tests fail prior to the change and pass after the change.

Resolves issue #174


iwyu.cc
```diff
@@ -2594,8 +2583,17 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     vector<const UsingDecl*> using_decls
         = FindInMultiMap(visitor_state_->using_declarations, decl);
     for (Each<const UsingDecl*> it(&using_decls); !it.AtEnd(); ++it) {
-      if (!(*it)->getDeclContext()->Encloses(using_context))
-        continue;
+      const DeclContext* using_decl_context = (*it)->getDeclContext();
+      if (!using_decl_context->Encloses(use_context)) {
+        // DeclContext::Encloses fails if the decl context is a function. In
+        // practice this doesnt matter because we only care about Encloses
```
kimgr: s/doesnt/doesn't/

iwyu.cc
```diff
@@ -2594,8 +2583,17 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        // when we're exporting symbols from a namespace, class, etc which can
+        // not happen if the using decl is in a function.
+        if(using_decl_context->getDeclKind() != clang::Decl::Function) {
```
kimgr: Space after if

iwyu.cc
```diff
@@ -3483,17 +3492,17 @@ class IwyuAstConsumer
...
     // The shadow decls hold the declarations for the var/fn/etc we're
     // using.  (There may be more than one if, say, we're using an
-    // overloaded function.)  We check to make sure nothing we're
-    // using is an iwyu violation.
-    for (UsingDecl::shadow_iterator it = decl->shadow_begin();
-         it != decl->shadow_end(); ++it) {
-      ReportDeclForwardDeclareUse(CurrentLoc(), (*it)->getTargetDecl());
-    }
-
+    // overloaded function.)  We don't want too add all of them at once
```
kimgr: s/too/to/

iwyu.cc
```diff
@@ -3483,17 +3492,17 @@ class IwyuAstConsumer
...
+    // though because that will drag in every overload even if we're
```
kimgr: Comma after "though".

iwyu_output.cc
```diff
@@ -605,6 +636,28 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
   LogSymbolUse("Marked fwd-decl use of decl", symbol_uses_.back());
 }
 
+void IwyuFileInfo::ReportUsingDeclUse(SourceLocation use_loc,
+                                      const UsingDecl* using_decl,
+                                      const NamedDecl* target_decl,
+                                      bool in_cxx_method_body,
+                                      const char* comment) {  
+  for(auto&& saved_decl : using_decl_lines_) {
+    if(saved_decl.matches(using_decl)) {
```
kimgr: Space after `for` and `if`

iwyu_output.cc
```diff
@@ -605,6 +636,28 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
   LogSymbolUse("Marked fwd-decl use of decl", symbol_uses_.back());
 }
 
+void IwyuFileInfo::ReportUsingDeclUse(SourceLocation use_loc,
+                                      const UsingDecl* using_decl,
+                                      const NamedDecl* target_decl,
+                                      bool in_cxx_method_body,
+                                      const char* comment) {  
+  for(auto&& saved_decl : using_decl_lines_) {
```
kimgr: I'd prefer spelling out the type here, instead of using `auto` -- I tend to use the latter only when the type is named on the right-hand side of the expression.

iwyu_output.cc
```diff
@@ -1844,6 +1897,27 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
...
+  // the decls that the using decl is referencing, which would result in a
+  // compilation error at best. A possible solution is to remove the using
+  // decl if it's not used, but that doesn't work for header files because a
+  // using decl in a header is an exported symbol, so we don't want to do that
+  // either. As a compromise, we arbitrarily add the first shadow decl to make
+  // sure everything still compiles instead of removing the using decel. A
+  // more thorough approach would be to scan the current list of includes that
+  // alredy name this decl (like in the overloaded function case) and include
+  // one of those so we don't include a file we dont actually need.
+  for(auto&& using_line : using_decl_lines_) {
```
kimgr: Please spell out the type here instead of using `auto`.

iwyu_output.cc
```diff
@@ -1844,6 +1897,27 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
...
+    if(!using_line.is_referenced()) {
```
kimgr: Space after `for` and `if`.

iwyu_output.h
```diff
@@ -214,6 +232,9 @@ class IwyuFileInfo {
   // the fwd-decl be removed, even if we don't see any uses of it.
   void AddForwardDeclare(const clang::NamedDecl* forward_declare_decl,
                          bool definitely_keep_fwd_decl);
+    // definitely_keep_fwd_decl tells us that we should never suggest
+  // the fwd-decl be removed, even if we don't see any uses of it.
```
kimgr: Comment doesn't seem to apply here, copy-paste residue, right?

iwyu_output.h
```diff
@@ -321,6 +347,9 @@ class IwyuFileInfo {
   // Holds all the lines (#include and fwd-declare) that are reported.
   vector<OneIncludeOrForwardDeclareLine> lines_;
 
+  // Holds all of the using delcs that ar reported.
```
kimgr: delcs -> decls and ar -> are

iwyu_output.cc
```diff
@@ -467,6 +468,26 @@ string OneIncludeOrForwardDeclareLine::LineNumberString() const {
   return buf;
 }
 
+OneUsingDeclLine::OneUsingDeclLine(
+    const UsingDecl* using_decl)
```
kimgr: No need to break this line, it fits onto 80 chars.

iwyu_output.cc
```diff
@@ -1844,6 +1897,27 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
...
+  // sure everything still compiles instead of removing the using decel. A
+  // more thorough approach would be to scan the current list of includes that
+  // alredy name this decl (like in the overloaded function case) and include
+  // one of those so we don't include a file we dont actually need.
+  for(auto&& using_line : using_decl_lines_) {
+    if(!using_line.is_referenced()) {
+      const UsingDecl* using_decl = using_line.using_decl();
+      ReportForwardDeclareUse(using_decl->getUsingLoc(),
+                              using_decl->shadow_begin()->getTargetDecl(), 
+                              false, nullptr);
```
kimgr: Can you add `/* comments */` with the argument names here to simulate named args? It's easier to understand what `false` and `nullptr` refers to, that way.

iwyu_output.h
```diff
@@ -214,6 +232,9 @@ class IwyuFileInfo {
   // the fwd-decl be removed, even if we don't see any uses of it.
   void AddForwardDeclare(const clang::NamedDecl* forward_declare_decl,
                          bool definitely_keep_fwd_decl);
+    // definitely_keep_fwd_decl tells us that we should never suggest
+  // the fwd-decl be removed, even if we don't see any uses of it.
+  void AddUsingDecl(const clang::UsingDecl* using_decl_decl);
```
kimgr: One too many `_decl` suffixes on `using_decl_decl`

iwyu.cc
```diff
@@ -2604,6 +2602,17 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return retval;
   }
 
+ private:
```
kimgr: Did this code really have to move? I can't see that your changes rely on it...

iwyu.cc
```diff
@@ -3623,7 +3632,11 @@ class IwyuAstConsumer
   // Called whenever a variable, function, enum, etc is used.
   bool VisitDeclRefExpr(clang::DeclRefExpr* expr) {
     if (CanIgnoreCurrentASTNode())  return true;
-    ReportDeclUse(CurrentLoc(), expr->getDecl());
+    if(const UsingShadowDecl* found_decl = DynCastFrom(expr->getFoundDecl())) {
```
kimgr: Space after `if`

iwyu_output.cc
```diff
@@ -537,6 +558,16 @@ void IwyuFileInfo::AddForwardDeclare(const clang::NamedDecl* fwd_decl,
            << internal::GetQualifiedNameAsString(fwd_decl) << "\n";
 }
 
+void IwyuFileInfo::AddUsingDecl(const clang::UsingDecl* using_decl_decl) {
```
kimgr: `using_decl_decl`

iwyu_output.cc
```diff
@@ -605,6 +636,28 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
...
+                                      bool in_cxx_method_body,
+                                      const char* comment) {  
+  for(auto&& saved_decl : using_decl_lines_) {
+    if(saved_decl.matches(using_decl)) {
+      saved_decl.set_referenced();
+      break;
+    }
+  }
+
+  ReportFullSymbolUse(use_loc, using_decl, in_cxx_method_body, comment);
```
kimgr: It'd be nice if you could add comments to these two Report-calls describing what they do.

As I'm reading it, this marks the using-decl itself as used, say if it comes from an included header, is that it?

iwyu_output.cc
```diff
@@ -605,6 +636,28 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
...
+  ReportForwardDeclareUse(using_decl->getSourceRange().getBegin(),
```
kimgr: ... and this marks use of the target decl as a forward-declare use. How do we know that's sufficient?

iwyu_output.cc
```diff
@@ -1844,6 +1897,27 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
...
+  // using it. If that happens, we might try to remove all of the headers with
+  // the decls that the using decl is referencing, which would result in a
+  // compilation error at best. A possible solution is to remove the using
+  // decl if it's not used, but that doesn't work for header files because a
+  // using decl in a header is an exported symbol, so we don't want to do that
+  // either. As a compromise, we arbitrarily add the first shadow decl to make
+  // sure everything still compiles instead of removing the using decel. A
+  // more thorough approach would be to scan the current list of includes that
+  // alredy name this decl (like in the overloaded function case) and include
+  // one of those so we don't include a file we dont actually need.
```
kimgr: dont -> don't

iwyu_output.cc
```diff
@@ -1844,6 +1897,27 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
...
 
+  // Resolve using declarations before continuing.  This handles the case
+  // where there's a using dclaration in the file but no code is actually
+  // using it. If that happens, we might try to remove all of the headers with
+  // the decls that the using decl is referencing, which would result in a
+  // compilation error at best. A possible solution is to remove the using
+  // decl if it's not used, but that doesn't work for header files because a
+  // using decl in a header is an exported symbol, so we don't want to do that
+  // either. As a compromise, we arbitrarily add the first shadow decl to make
+  // sure everything still compiles instead of removing the using decel. A
```
kimgr: decel -> decl

iwyu_output.cc
```diff
@@ -1844,6 +1897,27 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
   // we *do* need to add it.
   set<string> associated_desired_includes = AssociatedDesiredIncludes();
 
+  // Resolve using declarations before continuing.  This handles the case
+  // where there's a using dclaration in the file but no code is actually
```
kimgr: dclaration -> declaration

iwyu.cc
```diff
@@ -2594,8 +2583,17 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     vector<const UsingDecl*> using_decls
         = FindInMultiMap(visitor_state_->using_declarations, decl);
     for (Each<const UsingDecl*> it(&using_decls); !it.AtEnd(); ++it) {
-      if (!(*it)->getDeclContext()->Encloses(using_context))
-        continue;
+      const DeclContext* using_decl_context = (*it)->getDeclContext();
+      if (!using_decl_context->Encloses(use_context)) {
```
kimgr: You said in the original issue:
> This fix is not small and contains a questionable check at line 2577 of iwyu.cc that I think warrants some discussion.

Is this the one? I can't quite make out what this means -- can you say something about what this code block does, in plain English?

kimgr:
I went through a first pass with this patch -- I still haven't tried it in my environments, so these were mostly just cursory style nits.

I'm still not really sure how it works, I'll have to read through and mull on this some more.

Thanks!

iwyu_output.cc
cdglove: Fair enough.  I've personally come around to the almost-always-auto point of view, but I understand you're point of view as well.  I will change it.

iwyu_output.h
cdglove: Looks like it yes.  I will examine.

iwyu_output.h
cdglove: "delcs -> decls and ar -> are"

Thanks.  In general it looks like my spelling is atrocious.

"'of' seems redundant here."

Not where I'm from. Potato, potato?

iwyu_output.cc
cdglove: Yes.  

As a side note; this is a continuation of the pattern found throughout IWYU.  Personally, bool and nullable parameters are a pet peeve of mine;  I much prefer named enums and some kind of optional object.  As a minimum I will add the parameter names as a comment.

iwyu_output.h
cdglove: Actually this was intentional because the type is UsingDecl I felt that using_decl_decl was more descriptive in saying 'this is the declaration of some UsingDecl'.  But I can understand how it looks wrong.  Can you think of another name?

iwyu.cc
cdglove: No, this is just a reorganization of the original check.  The new check is explained directly below, starting on line 2588 (lines must have changed after I merged in the latest after submitting the original patch).

iwyu.cc
cdglove: I don't think it did move.  The diff is just confused because I inserted new code above it in the protected: section.

cdglove:
Thanks for the review Kim.  I will apply your requested changes on the spelling error and wait for any follow up discussion on the others.  If you need an overview of how it works, I can provide that, but the gist of it is that instead of marking all functions referenced through a using decl as 'full use', we instead defer that decision until *after* we're done with everything else so that we can make a proper decision on if the using decl requires an include or not.  The basic idea is that using foo; by itself does not require a #include because the *actual* use of foo later on will trigger the #include for us.

iwyu_output.cc
cdglove: This is not for using decl's from headers.  It's for using decl's in source. What this is doing is saying 'I have just accessed something through a using decl', which means we don't need to include anything to make that using decl valid because the actual use of it will force the correct include.  Here's an example;

#include <vector>
#include <list>
using std::swap;
// EOF

in the above case, std::swap is never referenced, so we will need to include vector or list to keep the file compiling.  (the old behaviour was to include both vector and list)

If we change the code;

#include <vector>
#include <list>
using std::swap;

void foo()
{
   std::vector<int> a, b;
   swap(a, b); // This will call ReportUsingDeclUse
}

Now we know which file we need to include; vector.

Does this make sense?

iwyu_output.h
kimgr: It's probably some kind of potato, I was working from the comment on line 347 ("// Holds all the lines...") and figured it might make sense to keep them consistent :-)

iwyu_output.cc
kimgr: I'll put it on my todo list to find these and clean them up.

The more "solutions" I see to the lack of named arguments in C++, the more I like the lo-fi comment employed by LLVM & Clang. It's cheap and gets the job done (until the argument is renamed, of course :-))

kimgr:
Thanks for your additional comments, I'll try to apply the changes locally tomorrow (CET) and see if I can understand the logic!

kimgr:
Now that I look at this context, I think `OneUsingDeclLine` is superfluos and a little misleading. The `OneIncludeOrForwardDeclareLine` it's modelled on is used in the final cleanup phase where include- and fwd-decl-lines are added or removed. `OneUsingDeclLine` doesn't appear to be used for that at all, just for delayed processing. So it would make more sense to me to remove the new data structure, and make `using_decl_lines_` an `std::map<UsingDecl*, bool>` instead, to keep track of whether the using declaration has been used or not.

I'm still trying to get my head around `ReportUsingDeclUse`. It feels like we're reporting the full symbol use of the using-decl and the forward-declare use of the target decl (I'm not even sure what this target decl is anymore -- the original decl that the using-decl is bringing into the namespace?) too early -- won't everything already have been reported by the time we get to `IwyuFileInfo::CalculateAndReportIwyuViolations` where the unreferenced using-decls are forced as forward-declared?

iwyu_output.cc
cdglove: We don't know it's sufficient, but we also don't know it's not. There's logic later on to convert all functions to FullSymbolUse. This is the same as how it worked previously but now we're running the report through the using decl.

cdglove:
> So it would make more sense to me to remove the new data structure, and make using_decl_lines_ an std::map<UsingDecl*, bool> instead, to keep track of whether the using declaration has been used or not.

We could do this.  I was continuing the pattern I saw there, but possibly I didn't understand the original intent. This structure is also used for logging however, so that would need to be changed too.

cdglove:
> I'm still trying to get my head around ReportUsingDeclUse. It feels like we're reporting the full symbol use of the using-decl and the forward-declare use of the target decl too early -- won't everything already have been reported by the time we get to IwyuFileInfo::CalculateAndReportIwyuViolations where the unreferenced using-decls are forced as forward-declared?

You have a point here that I need think about.  This might actually simplify the implementation Here's the reasoning I used;

- Previously, using decls were not treated special, so a using decl would result in the target decl(s) being treated the same as if they had been referenced outside of a using decl and would be marked as a 'use' appropriately.
- New behaviour is to treat using decls special.  Instead of automatically bringing in the target decl(s), we simply record the fact that we have a using decl that refers to some target decl(s). 
- This is the same as saying 'using decls never require an include, because the target decl(s) that get referenced will eventually bring in the required header'
- This creates a potential bug though where if nothing references the target decl(s) that the using decl refers to, we will try to remove the headers that the using decl is importing symbols from and the source would not compile. This is fixed by the OneUsingDeclLine; we are simply tracking when a symbol is referenced thorough a using decl so we know it's safe to ignore the using decl entirely.

It's been a few months since I did this now so I have to go back and step through some code to figure it out again.

kimgr:
Yeah, there is a pattern there, but it's a pattern for something else :-)

As for the logging, I think you can just inline the relevant parts from `LogSymbolUse` instead of cooking up a fake `OneUse`.

Oh, your follow-up comment just came in while I was typing this. Thanks for the added detail, this looks like a nice high-level description, I'll think some more about this as time permits.

cdglove:
I pushed a few changes to my master fork, which seems to have automatically updated the pull request.  In the future, I suppose I should do this in a branch so I can void that.  Is that the expected workflow?  

The changes are as follows;

- Rename OneUsingDeclLine to UsingDeclStatus as that more accurately reflects what it's doing.  I could just inline the code where necessary and use a map to bool, as Kim mentioned, but this felt a little cleaner as inline code has a tendency to gain tentacles.

- Remove unnecessary ReportForwardDeclareUse of UsingDecl's target decl because it will actually be picked up separately in the AST traversal.  This was as theorized in previous discussion.
- Add a test to catch the case where we have a header importing symbols from another header via using decls. If we're not careful in this case, we can end up removing the header with the using decls which would be undesirable.

iwyu_test_util.py
```diff
@@ -69,7 +69,8 @@ def _Which(program, paths):
 
 def _GetIwyuPath(iwyu_paths):
   """Returns the path to IWYU or raises IOError if it cannot be found."""
-  iwyu_path = _Which('include-what-you-use', iwyu_paths)
+  #iwyu_path = _Which('include-what-you-use', iwyu_paths)
+  iwyu_path = "/home/chris/src/build/llvm/_target/gcc/bin/include-what-you-use"
```
kimgr: Oops, we probably don't want to merge this :-)

The test runner is designed to search a number of known paths + the system `PATH`. So you can add your IWYU build dir to the `PATH` in your shell and it will find it.

iwyu_output.cc
kimgr: Cool, this looks much clearer!

A few things --

  * Isn't it a hard error If the `UsingDecl` is not found in the vector? I'd reorganize it so that `ReportFullSymbolUse` is called inside the loop, when the decl is found, and replace the `break` with a `return`. Then `CHECK_UNREACHABLE_` after the loop.
  * Can you add a comment above `ReportFullSymbolUse` to describe cases that make it necessary? I think it's for the so-called 'imported' usings from header files, is that it?

iwyu_output.cc
kimgr: I still think this should just be `using_decl`. When I try to pronounce this as "using declaration declaration" I get even more confused :-)

iwyu.cc
kimgr: Oh, I see what happened now; you moved `GetUsingDeclarationOf` from private to protected, makes sense, thanks.

using_overload.cc
```diff
@@ -0,0 +1,30 @@
+//===--- using_overload.cc - test input file for iwyu ---------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include <vector>
```
kimgr: This file seem to have ended up in the wrong place.

using_specialization.cc
```diff
@@ -0,0 +1,31 @@
+//===--- using_specialization.cc - test input file for iwyu ---------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+namespace ns {
```
kimgr: This file seem to have ended up in the wrong place.

tests/cxx/declare_symbol.h
```diff
@@ -0,0 +1,13 @@
+//===--- declare_symbol.h - test input file for iwyu ----------------------===//
```
kimgr: There's a naming convention for test files to make it easier to correlate test `.cc` files with their headers.

There are some exceptions to this, like `direct.h` and `indirect.h`, which are 'reusable' among tests, but otherwise we try to keep headers prefixed to avoid accidental coupling between tests. There's a little blurb on this here, soon to be Markdownized:
https://code.google.com/p/include-what-you-use/wiki/InstructionsForDevelopers#Running_the_Tests

I think this should be `using_import-declare.h`.

tests/cxx/import_symbol.h
```diff
@@ -0,0 +1,13 @@
+//===--- import_symbol.cc - test input file for iwyu ----------------------===//
```
kimgr: I think this should be `using_import-import.h`.

iwyu_output.cc
```diff
@@ -605,6 +635,21 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
   LogSymbolUse("Marked fwd-decl use of decl", symbol_uses_.back());
 }
 
+void IwyuFileInfo::ReportUsingDeclUse(SourceLocation use_loc,
+                                      const UsingDecl* using_decl,
+                                      const NamedDecl* target_decl,
```
kimgr: This argument is no longer used here.

iwyu_output.h
```diff
@@ -175,6 +176,23 @@ class OneIncludeOrForwardDeclareLine {
   const clang::NamedDecl* fwd_decl_;
 };
 
+class UsingDeclStatus {
```
kimgr: I don't mind this class as much now that it's disassociated from the other `One...` classes.

But I still find it lacks cohesion, see below.

iwyu_output.h
```diff
@@ -241,6 +261,13 @@ class IwyuFileInfo {
                                const clang::NamedDecl* decl,
                                bool in_cxx_method_body, const char* comment);
 
+  // Called whenever a NamedDecl is accesed through a UsingDecl.
```
kimgr: accessed

kimgr:
> I pushed a few changes to my master fork, which seems to have automatically updated the pull
> request. In the future, I suppose I should do this in a branch so I can void that. Is that the
> expected workflow?

That's what I normally do, yes.

iwyu_output.cc
cdglove: OK.  I'll change it.

iwyu_output.cc
cdglove: Yes, it's unclear to me why I wrote it that way.  It should fail if we can't find the using decl.

iwyu_test_util.py
cdglove: Oops indeed.

iwyu_output.cc
cdglove: Noted, nice catch.

tests/cxx/declare_symbol.h
cdglove: OK got it. On a related note, what I really wanted to do was group my using decl tests into a directory.  It seems the test runner script supports this. That change would be for another day, but do you have any thoughts on that?

kimgr: I'm not sure that works, the files currently in subdirs (IIRC) are just headers that need to be included with a special name. But it might very well work, it hasn't occurred to me before.

kimgr: OK, so I tested. It does work (see `internal/internal_files.cc`), but I wouldn't group things with directories unless it makes a difference for the IWYU analysis.

iwyu_output.cc
cdglove: Interestingly, I just made this edit and a few tests are now failing (not just my tests, but some others too).  What this means is that somehow we don't know about all of the using decls in the source.  I will investigate why, but probably wont get to it until the weekend now.

tests/cxx/using_overload.cc
```diff
@@ -0,0 +1,30 @@
+//===--- using_overload.cc - test input file for iwyu ---------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
```
vsapsai: Please add comments in tests what exactly the test is supposed to catch. For example,

> Tests that if we encounter using overloaded function, iwyu recommends to keep only files with actually used function signatures.

And maybe add a test case where you use both `swap(vector, vector)` and `swap(list, list)`.

vsapsai:
In process of reviewing, just added a note about comments in tests to simplify further reviewing.

iwyu_output.cc
```diff
@@ -537,6 +538,19 @@ void IwyuFileInfo::AddForwardDeclare(const clang::NamedDecl* fwd_decl,
            << internal::GetQualifiedNameAsString(fwd_decl) << "\n";
 }
 
+void IwyuFileInfo::AddUsingDecl(const clang::UsingDecl* using_decl) {
+  CHECK_(using_decl && "using_decl unexpectedly NULL");
+  using_decl_referenced_.insert(std::make_pair(using_decl, false));
```
cdglove: I've gotten rid of the UsingDeclStatus and gone with a map<UsingDecl*, bool> instead as you suggested.  Feels cleaner overall.

tests/cxx/using_overload.cc
cdglove: I agree.  I'll do this for the tests I've added. At some point the existing tests will need to be updated.

cdglove:
This latest version feels much better than that first patch I submitted. I think the code is clearer and easier to understand.

iwyu_output.cc
```diff
@@ -605,6 +619,29 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
...
+
+  // If accessing a symbol through a using decl in the same file that contains
+  // the using decl, we must mark the using decl as referenced. At the end of
+  // traversing the AST, we check to see if a using decl is unreferenced and a
+  // full use of one of its shadow decls so that the source file continues to
+  // compile.
+  typename map<const UsingDecl*, bool>::iterator using_decl_status = 
+    using_decl_referenced_.find(using_decl);
+
+  if(using_decl_status != using_decl_referenced_.end()) {
```
cdglove: In the last review, we hypothesized that this should be a hard error if the using decl is not found.  This actually isn't the case.  Here, we are looking to see if a using decl is referenced in the *same* file it is declared in.  Only then do we care to mark the using decl as referenced.  This is strictly to cover the case where a header file exports a symbol though a using decl but doesn't actually use the using decl at all.

iwyu.cc
kimgr: I find this a little hard to understand. On the one hand, you say that `Encloses` returns false for functions, but then on the other hand, you seem to say that it's never called for functions. Could the inner `if` be replaced by a `CHECK_`?

It seems wrong not to `continue` if it actually is a function. And if the prior predicate failed for functions, I'm not sure why you changed this.

iwyu.cc
```diff
@@ -1601,12 +1601,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // the namespace of the decl.  If we have 'using std::vector;' +
     // 'std::vector<int> foo;' we don't actually care about the
     // using-decl.
-    // TODO(csilvers): maybe just insert our own using declaration
-    // instead?  We can call it "Use what you use". :-)
```
kimgr: Sentimental and silly, but this was one of my favorite comments, it always makes me smile :-)

At first I thought you removed it because you'd fixed the TODO, but it doesn't seem like it. so care to leave it to cheer me up?

iwyu.cc
```diff
@@ -3623,7 +3630,11 @@ class IwyuAstConsumer
   // Called whenever a variable, function, enum, etc is used.
   bool VisitDeclRefExpr(clang::DeclRefExpr* expr) {
     if (CanIgnoreCurrentASTNode())  return true;
-    ReportDeclUse(CurrentLoc(), expr->getDecl());
+    if (const UsingShadowDecl* found_decl = DynCastFrom(expr->getFoundDecl())) {
+      ReportDeclUse(CurrentLoc(), found_decl->getTargetDecl());
```
kimgr: What about the shadow decl itself? I'm thinking of something like:

    // a.h
    namespace A { void foo(); }

    // b.h
    #include "a.h"
    namespace B { using A::foo; }

    // main.cc
    #include "b.h"
    void use() {
        return B::foo();
    }

It seems like this would record a use of `A::foo` but not `B::foo`? Actually, with this example in mind, it seems more right to just use the decl, regardless of whether it's a shadow decl or not, but I'm probably missing something :-)

iwyu_output.cc
```diff
@@ -605,6 +619,29 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
   LogSymbolUse("Marked fwd-decl use of decl", symbol_uses_.back());
 }
 
+void IwyuFileInfo::ReportUsingDeclUse(SourceLocation use_loc,
+                                      const UsingDecl* using_decl,
+                                      bool in_cxx_method_body,
+                                      const char* comment) {  
+
```
kimgr: Kill blank line

iwyu_output.cc
kimgr: Space after if.

iwyu_output.cc
```diff
@@ -537,6 +538,19 @@ void IwyuFileInfo::AddForwardDeclare(const clang::NamedDecl* fwd_decl,
            << internal::GetQualifiedNameAsString(fwd_decl) << "\n";
 }
 
+void IwyuFileInfo::AddUsingDecl(const clang::UsingDecl* using_decl) {
```
kimgr: You shouldn't need to namespace-qualify `UsingDecl` here, there was a `using` of it above.

iwyu.cc
cdglove: Ah yes, possibly this warrants more discussion, but I removed the comment because we can't actually do what it says. We could *possibly* remove the using from a source file, but not a header because in a header the using becomes an exported symbol.  In source, it might be possible, then IWYU would have different behaviour between source and header files, and I thought that would be undesirable.

iwyu_output.cc
```diff
@@ -605,6 +619,29 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
...
 }
 
+void IwyuFileInfo::ReportUsingDeclUse(SourceLocation use_loc,
+                                      const UsingDecl* using_decl,
+                                      bool in_cxx_method_body,
+                                      const char* comment) {  
+
+  // If accessing a symbol through a using decl in the same file that contains
+  // the using decl, we must mark the using decl as referenced. At the end of
+  // traversing the AST, we check to see if a using decl is unreferenced and a
```
kimgr: "and a" -> "and add a"?

iwyu.cc
cdglove: The original code failed to find the using decl of a symbol if the using decl is in side a function.  For example;

```cpp
void foo() {
  using std::swap;
  std::vector a, b;
  swap(a, b);
}
```
In this example, GetUsingDeclarationOf 'swap' returns NULL, which is not what we want.  As far as I can tell, 
```cpp
(!(*it)->getDeclContext()->Encloses(using_context))
```
returns false whenever the using decl is in the same function it is used in and I don't know why.  This is what I meant by a questionable check.  But, as noted in the comment, in practise, it doesn't matter because we only care about the encloses case to make sure we're not finding a using decl across namespaces.

I think....

iwyu.cc
cdglove: Sounds like a good test.  I'll add one for this case.

iwyu_output.cc
cdglove: Correct.  Copy/paste surely.

iwyu_output.cc
cdglove: Noted.

iwyu_output.cc
cdglove: Noted.

iwyu_output.cc
cdglove: Yes, noted.

kimgr:
Thank you so much for working on this! GitHub's UI for commenting is a little wonky, so the review gets a little confused for me...

I still don't have all the concepts clear in my mind (I don't have an intuitive sense of the difference between a using decl and a using shadow decl, for example), so this is a little hard for me, but it looks like we're getting there.

It's a little disconcerting that you needed to change the existing tests, do you have a sense of whether those changes make the tests more correct, or whether you just had to update them to get them to pass?

cdglove:
> Thank you so much for working on this! GitHub's UI for commenting is a little wonky, so the review gets a little confused for me...

Yes, it doesn't accumulate comment properly with incremental changes.  Though, that gets confusing too if you've ever used Code Collaborator or Review Board plugged into Perforce...

> I still don't have all the concepts clear in my mind (I don't have an intuitive sense of the difference between a using decl and a using shadow decl, for example), so this is a little hard for me, but it looks like we're getting there.

Yes we're getting there.  And I'm happy to be doing it; it's not a simple problem to solve so I am not surprised it's taking some iteration time.

As for the using decl/shadow decl thing,aAs far as I can tell, the using decl is the using statement, and the shadow decl is the thing it might refer too;

```cpp
#include <vector>
#include <list>
using std::swap;  // Here UsingDecl is this line, and it will have two ShadowDecls -- one for each overload of swap
```

> It's a little disconcerting that you needed to change the existing tests, do you have a sense of whether those changes make the tests more correct, or whether you just had to update them to get them to pass?

The tests themselves didn't need to change, I only needed to remove some IWYU tags that silence some IWYU warnings (ie:  the warnings no longer occur). The reason the warnings no longer occur around using decls is because we are no longer considering a using decl a "use" because doing so drags in all of it's shadow decls.  I can't think of a straight forward way around that.

iwyu_output.cc
```diff
@@ -1844,6 +1881,29 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
...
+  // where there's a using declaration in the file but no code is actually
+  // using it. If that happens, we might try to remove all of the headers with
+  // the decls that the using decl is referencing, which would result in a
+  // compilation error at best. A possible solution is to remove the using
+  // decl if it's not used, but that doesn't work for header files because a
+  // using decl in a header is an exported symbol, so we don't want to do that
+  // either. As a compromise, we arbitrarily add the first shadow decl to make
+  // sure everything still compiles instead of removing the using decl. A
+  // more thorough approach would be to scan the current list of includes that
+  // alredy name this decl (like in the overloaded function case) and include
```
kimgr: Oops: "alredy"

iwyu_output.cc
cdglove: There are a few other places in this file that are fully qualifying the names.  I have made a not to remove them in a future change.

iwyu.cc
kimgr: Thanks for the explanation, this makes more sense to me now.

There's one branch where this seems broken; if Encloses returns false and the decl-kind is a function, we'll always register it, whether it's the *right* function (i.e. one enclosing the using-decl) or not.

It seems like this could cause problems, but I'm not sure.

I think if Encloses fails you could just check if the using-decl's parent is the same as the decl's parent, and if either of them is a FunctionDecl. That seems like a naive implementation of Encloses :-). I'll ask the Clang devs if Encloses is broken in this regard, it could be that we could get it fixed.

iwyu_output.cc
kimgr: Thanks for the update, I can now understand the logic based on the comments.

iwyu.cc
kimgr: Yes, it does seem a little pretentious.

On the other hand, I think IWYU already has this behavior wrt `#include`. When run on `a.cc`, it modifies the associated header `a.h` in ways that might break consumers of the header (e.g. removing unused includes that were exported and used downstream.)

I'm not necessarily arguing for keeping the comment, I'm just thinking out loud... :-)

iwyu.cc
kimgr: Also, I'd love it if you could add a comment here for each of the two reports with the kind of constructs they're intended to catch.

iwyu_output.cc
kimgr: OK, thanks!

kimgr:
> The tests themselves didn't need to change, I only needed to remove some IWYU tags that silence 
> some IWYU warnings (ie: the warnings no longer occur). The reason the warnings no longer occur 
> around using decls is because we are no longer considering a using decl a "use" because doing so 
> drags in all of it's shadow decls. I can't think of a straight forward way around that.

And this is essentially the gist of your fix, isn't it -- to delay reports for shadow decls until we've proven they are used? If the test changes support that change, then they're fine.

But this one:
> -// IWYU: i1_ns::I1_NamespaceStruct is...*badinc-i1.h
> +// IWYU: I1_NamespaceStruct ...*badinc-i1.h

seems to have changed behavior. Can you see why it's no longer reporting the namespace-qualified name?

kimgr:
Hey,

I had an idea. I think the whole GetUsingDeclarationOf construct is unnecessary.

I've been digging into the AST of this reduced case:

    namespace thing {
    void swap(int& a, int& b) {
      int tmp = a;
      a = b;
      b = tmp;
    }
    }

    void foo() {
      using thing::swap;
      int a = 10, b = 20;
      swap(a, b);
    }

It looks like the call to `swap` generates a branch like this in the AST:

    `-CallExpr 0xdbd378 <line:12:3, col:12> 'void'
      |-ImplicitCastExpr 0xdbd368 <col:3> 'void (*)(int &, int &)' <FunctionToPointerDecay>
      | `-DeclRefExpr 0xdbd330 <col:3> 'void (int &, int &)' lvalue Function 0xdbceb0 'swap' 'void (int &, int &)' (UsingShadow 0xdbd198 'swap')
      |-DeclRefExpr 0xdbd300 <col:8> 'int' lvalue Var 0xdbd1e0 'a' 'int'
      `-DeclRefExpr 0xdbd318 <col:11> 'int' lvalue Var 0xdbd248 'b' 'int'

Note how the `DeclRefExpr` denoting `swap` shows its associated using-shadow. I now understand a using-shadow to be a special decl forming a link between a using-decl and a found decl, so that we can go in both directions after lookup.

Since we have this information, how is `GetUsingDeclarationOf` useful?

We should be able to, in `ReportDecl...`, always do something like:

    if (const UsingShadowDecl* using_shadow = DynCastFrom(decl)) {
      const UsingDecl* using_decl = using_shadow->getUsingDecl();
      preprocessor_info().FileInfoFor(used_in)->ReportUsingDeclUse(
          used_loc,
          using_decl,
          IsNodeInsideCXXMethodBody(current_ast_node()),
          "(for using decl)");
    }

Unless I'm mistaken, `GetUsingDeclarationOf` seems like a bad approximation of the using shadow decl concept.

With this in place, `VisitDeclRefExpr` shouldn't need any special processing, and any other constructs producing a using-shadow decl would work transparently.

What do you think?

cdglove:
> seems to have changed behavior. Can you see why it's no longer reporting the namespace-qualified name?

I'm not sure actually. If I had to guess I would say it is due to the change in VisitDeclRefExpr where previously we were blindly adding the decl, but now we check to see if the decl is actually a shadow decl and add that instead.  This change was to make GetUsingDeclarationOf, which is called deeper in the callstack, work correctly.  Possibly that is not necessary anymore based on your hypothesis above, which I am going to look at today if I can find some time.

kimgr:
Thanks, I just did a quick experiment based on the hypothesis above and it doesn't pass your test suite, so I'm probably missing something. I think my interpretation of `UsingShadowDecl` is broken after all.

iwyu_output.cc
```diff
@@ -605,6 +619,29 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
...
+    using_decl_referenced_.find(using_decl);
+
+  if(using_decl_status != using_decl_referenced_.end()) {
+    using_decl_status->second = true;
+  }
+
+  // When a symbol is accessed through a using decl, we must report
+  // that as a full use of the using decl because whatever file that
+  // using decl is in is now required.
+  ReportFullSymbolUse(use_loc, using_decl, in_cxx_method_body, comment);
```
vsapsai: What if we also add that using_decl is making ForwardDeclareUse of shadow target? I.e. something like `ReportForwardDeclareUse(using_decl->getUsingLoc(), decl, ...)`? I haven't found a case where it is crucial, because `decl` is marked as used anyway. But seems like it is correct thing to do.

iwyu.cc
```diff
@@ -3483,17 +3490,17 @@ class IwyuAstConsumer
       AddUsingDeclaration((*it)->getTargetDecl(), decl);
```
vsapsai: Maybe rename `AddUsingDeclaration` to `AddUsingShadowDecl`? Because now we have `AddUsingDeclaration` and `AddUsingDecl` which is confusing.

tests/cxx/using_import.cc
```diff
@@ -0,0 +1,29 @@
...
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "using_import-import.h"
+
+void use_symbol() {  
+  // IWYU: ns::symbol is defined in ...*using_import-declare.h", which isn't directly #included.
+  symbol();
+}
```
vsapsai: Does this test catch something that is not caught by include_with_using.cc? It is kinda similar to include_with_using-d3.h, include_with_using-d3b.h. But it's not exactly the same and I don't know how important is the full use in this test.

And please don't use 'import' in test name because there are also `#import` and `@import`. Test name in this case is misleading.

iwyu_output.cc
```diff
@@ -1844,6 +1881,29 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
...
+  for (typename map<const UsingDecl*, bool>::value_type using_decl_status
+       : using_decl_referenced_) {
+    if (!using_decl_status.second) {
+      const UsingDecl* using_decl = using_decl_status.first;
+      ReportForwardDeclareUse(using_decl->getUsingLoc(),
+                              using_decl->shadow_begin()->getTargetDecl(), 
+                              /* in_cxx_method_body */ false,
+                              "(for un-referenced using)");
+    }
+  }
```
vsapsai: Does it matter that you are doing this after `AssociatedDesiredIncludes`, not before? And it feels a little bit dirty doing it in `CalculateAndReportIwyuViolations`. Maybe we should add something like `IwyuFileInfo::ResolvePendingAnalysis` or `IwyuFileInfo::DidFinishFileProcessing` and call it from `HandleTranslationUnit`. What do you think?

iwyu.cc
cdglove: > I think if Encloses fails you could just check if the using-decl's parent is the same as the decl's parent, and if either of them is a FunctionDecl.

I'm pretty sure I tried that as I tried a few permutations before I settled here.

> I'll ask the Clang devs if Encloses is broken in this regard, it could be that we could get it fixed.

Fantastic!

iwyu.cc
cdglove: This is a good point, I'll clean it up.

iwyu_output.cc
cdglove: This is an interesting thing to consider because I think the difference in functionality is subtle. In the end, the correct includes will be reported either way, but things like logs and error messages might be slightly different. One difference already seems to be that the decls are reported as they appear in source code.  Consider the following;

```cpp
using std::swap;
std::vector a, b;
swap(a, b);
```

If we report swap as being used through the using decl, as you suggest, it is reported as std::swap.  If we are lazy about it and let the AST traversal handle it, it's reported as simply swap.  I'm not certain which one we ultimately want.  Previous functionality (before my change) is to report as std::swap because the shadow decls were added greedily.

I will experiment with this to see what else is different.

iwyu_output.cc
cdglove: > Does it matter that you are doing this after AssociatedDesiredIncludes, not before?

Curious, I would think it would matter.  I'll write a test to see if I can break it.

> And it feels a little bit dirty doing it in CalculateAndReportIwyuViolations. Maybe we should add something like IwyuFileInfo::ResolvePendingAnalysis or IwyuFileInfo::DidFinishFileProcessing and call it from HandleTranslationUnit. What do you think?

I agree.  I like ResolvePendingAnalysis.  I'll make that change.

tests/cxx/using_import.cc
cdglove: > Does this test catch something that is not caught by include_with_using.cc? It is kinda similar to include_with_using-d3.h, include_with_using-d3b.h. But it's not exactly the same and I don't know how important is the full use in this test.

It might, but I'm not certain because include_with_using.cc is, in my personal opinion, trying to do too much. I actually think many of the iwyu tests are guilty of this. I prefer to start with very simple, specific tests, and progress to more complicated cases that look more like integration tests, but that might just be me. At any rate, I was trying to isolate the specific case I was fixing while making sure I didn't break basic functionality.

> And please don't use 'import' in test name because there are also #import and @import. Test name in this case is misleading.

Ah, right.  I hadn't considered that.  I will find a better name.

kimgr:
Ping! How's it going with this? It's a really ambitious patch, and I'd love to see this improved in IWYU. Sorry we drowned you in review comments :-)

cdglove:
Hi Kim, I was working on it this weekend actually. No worries on all of the comments, as one of them really exposed an issue.

One of your suggestions was to create a test as follows;

// a.h
namespace A { void foo(); }

// b.h
#include "a.h"
namespace B { using A::foo; }

// main.cc
#include "b.h"
void use() {
    return B::foo();
}

When running this, the detection that 'b.h' is used fails and IWYU suggests removing b.h, which obviously doesn't work. I experimented a little bit with taking a different approach where we could do away with the map from Decls to UsingDecls, but that also doesn't work because by the time something is detected as used, we've lost the context of if it was used through a UsingDecl or not.  This made me realize that despite my changes, the system is still conservative when it comes to using decls (though my changes do improve the situation a lot).

I discovered this a few weeks ago only this weekend found the time to look into it. Since the different approach failed, I need to revert my changes a take a look at the AST again to try to figure out where it's going wrong.  At any rate, I haven't forgotten about it. It just turns out to be a hard problem to solve.  :-)

kimgr:
Good to hear!

>  It just turns out to be a hard problem to solve. :-)

Aren't they all :-)

iwyu_output.cc
```diff
@@ -605,6 +619,29 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
...
+                                      const UsingDecl* using_decl,
+                                      bool in_cxx_method_body,
+                                      const char* comment) {  
+
+  // If accessing a symbol through a using decl in the same file that contains
+  // the using decl, we must mark the using decl as referenced. At the end of
+  // traversing the AST, we check to see if a using decl is unreferenced and a
+  // full use of one of its shadow decls so that the source file continues to
+  // compile.
+  typename map<const UsingDecl*, bool>::iterator using_decl_status =
```
kimgr: MSVC doesn't accept `typename` here, only inside a template.

iwyu_output.cc
```diff
@@ -1844,6 +1881,29 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
...
+  // the decls that the using decl is referencing, which would result in a
+  // compilation error at best. A possible solution is to remove the using
+  // decl if it's not used, but that doesn't work for header files because a
+  // using decl in a header is an exported symbol, so we don't want to do that
+  // either. As a compromise, we arbitrarily add the first shadow decl to make
+  // sure everything still compiles instead of removing the using decl. A
+  // more thorough approach would be to scan the current list of includes that
+  // alredy name this decl (like in the overloaded function case) and include
+  // one of those so we don't include a file we don't actually need.
+  for (typename map<const UsingDecl*, bool>::value_type using_decl_status
```
kimgr: MSVC doesn't accept `typename` here either.

kimgr:
I tried to get to the bottom of the weirdness with `Encloses` today, and it turned out to be fairly straightforward -- `Encloses` does not include the current decl context itself, so when we check the use's decl context against the using-declaration's decl context, `Encloses` will only return true if one is nested inside the other.

So the code I would expect to work is this:

    if (!using_decl_context->Encloses(use_context) &&
        !using_decl_context->Equals(use_context))
      continue;

... but unfortunately that causes the `using_overload` test to fail (both `<vector>` and `<list>` come out as needed.) Including `<vector>` after `<list>` causes it to pick the right `std::swap` and the test passes, so apparently the include order affects lookup and visibility somehow.

That in turn inspired me to try and discard `GetUsingDeclarationOf` entirely, and use `UsingShadowDecls` instead to find which using-declaration gave access to a used underlying decl. Unfortunately that caused `include_with_using` to fail entirely, and I never sorted it out.

Anyway, I thought I'd give you an update on `Encloses` -- it's not broken, it's just that we (or at least I) misunderstood how it works.

iwyu_output.cc
cdglove: Fixed.

iwyu_output.cc
cdglove: Also fixed.

cdglove:
Hi Kim,
Funny, I tried everything you just described.  In the end, I went with a combination and it seems to be working. Now, when I encounter a UsingShadowDecl, instead of throwing it away and passing in the result of getTargetDecl, I pass the UsingShadowDecl into ReportDeclUse.  Inside ReportDeclUse, I convert the decl to the result of getTargetDecl, but I pass the UsingShadowDecl into GetUsingDeclarationOf().  Inside that function, if we have a UsingShadowDecl, we just return the UsingDecl from that (because at that point we know we're going through a using decl so we don't need to check the namespaces), otherwise, it's business as usual.
I'm considering issuing a fresh pull request with the changes because this one has gotten a bit cluttered.  Thoughts?

kimgr:
Very cool!

Yeah, I think a new PR with cleaned up commit history would be good. I see some small things in the existing diff, but overall it would be good to start from scratch.

cdglove:
OK, new pull request incoming and we can clean it up from there.

cdglove:
Closing this old PR as it's superceded by PR #221

PR #202, opened 2015-07-30
Fix issue 176: Build out-of-source fails on Fedora

insysion:
Allow separate CMake arguments for LLVM include and lib paths.
Rename LLVM_PATH -> IWYU_LLVM_ROOT_PATH and add deprecation message.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,12 +5,27 @@ project(include-what-you-use)
 if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
   message(STATUS "IWYU out-of-tree configuration")
 
-  if( NOT DEFINED LLVM_PATH )
-    message(FATAL_ERROR "LLVM_PATH must be provided using -DLLVM_PATH=<path to llvm package root>")
+  if( DEFINED LLVM_PATH )
+    set(IWYU_LLVM_ROOT_PATH ${LLVM_PATH})
+    message(WARNING "LLVM_PATH is deprecated, use IWYU_LLVM_ROOT_PATH instead.")
   endif()
 
-  link_directories(${LLVM_PATH}/lib)
-  include_directories(${LLVM_PATH}/include)
+  if( DEFINED IWYU_LLVM_INCLUDE_PATH AND DEFINED IWYU_LLVM_LIB_PATH )
+    # User provided include/lib paths, fall through
+  elseif ( DEFINED IWYU_LLVM_ROOT_PATH )
+    # Synthesize include/lib relative to a root.
+    set(IWYU_LLVM_INCLUDE_PATH ${IWYU_LLVM_ROOT_PATH}/include)
+    set(IWYU_LLVM_LIB_PATH ${IWYU_LLVM_ROOT_PATH}/lib)
+  else()
+    # If none provided, fail.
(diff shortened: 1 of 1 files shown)


kimgr:
Instinctively I didn't want to mess with the `clang_3.4` branch, since it's already been released (and quite a long while ago). But if this is useful/necessary, I suppose it can't hurt to backport.

@vsapsai, what do you think?

insysion:
I think it's worth bearing in mind that clang 3.4 is still the version used
by RedHat et al and so is useful for enterprise users.

daveisfera:
I'm the packager for iwyu in Fedora and RHEL/EPEL and I've actually already worked around this issue with my own similar patch in RHEL/EPEL. I probably should have flowed the change upstream way back when and asked that it be rolled in, but I think that having some sort of "official" support/fix for this would be good. If you'd like, I can test this change to verify that it fixes the issue in the RHEL/EPEL build as well.

kimgr:
Hey @daveisfera, thanks for the additional detail. Yes, I'd appreciate it if you could test it from a packaging perspective.

daveisfera:
The patch works on x86 builds, but it currently uses a hard coded path for lib (on line 18). On x64 the path is lib64 and so the build fails on 64-bit builds.
I don't know anything about CMake, but is there a way to make that be lib on 32-bit and lib64 on 64-bit? Because this without a change like that, this change just isn't usable for building on RHEL/EPEL.

vsapsai:
I am OK with backporting. Just want to figure out what is the best way to do this. Do you need to use branch and tag `clang_3.4` or you can use a different branch/tag? Repository is enough for you, you don't need source archive, correct? What RHEL/EPEL versions are using clang 3.4 (asking to know what is the delay in updating clang version)?

daveisfera:
As far as current Fedora releases are concerned, this issue has been fixed and a similar fix is already in use in the packaging done for EPEL 6 and 7, so from my perspective this fix isn't explicitly necessary but it would be nice to have a working "official" fix to for consistency.
As far as the version of clang goes, EPEL 6 and 7 are both using clang 3.4 and according to the update policy ( https://fedoraproject.org/wiki/Updates_Policy#Stable_Releases ) should stay at that version for the entire life of the corresponding RHEL release, but I'm not sure what will happen in the long run.

kimgr:
Where are we with this? I'd like to move forward, but I'm not sure where to...

daveisfera:
From my perspective, it would be nice to have "official" support for this because then I could clean up the packaging to not have to manually patch in this support. But the patch is already in place and working, so life will go on if nothing else happens with this.

kimgr:
OK, thanks! So we want to backport that commit to clang_3.4? And we don't need to release again, it's just for packagers? Makes sense to me, I'll see if I can make time soon.

daveisfera:
It would be nice to also have a tagged release, but for the EPEL/Fedora
packages I point at a github commit (which usually corresponds to a
release) and so it's nice but not necessary.
http://pkgs.fedoraproject.org/cgit/rpms/iwyu.git/commit/?id=e9c0f036a9952f2e1936c30bb1c57c66e85daf26

kimgr:
OK, so I think I fixed this now. It turns out we have both a tag and a branch called `clang_3.4`. This was pushed into the branch, the tag is still at our last public release. I think that's as it should be.

PR #204, opened 2015-08-05
Fix issue #163: briefer why comments

kimgr:
Oops, I seem to have killed the old PR. Here's an updated one with review comments addressed.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -72,53 +72,53 @@ class OutputLine {
   explicit OutputLine(const string& line)
       : line_(line) {}
   OutputLine(const string& line, const vector<string>& symbols)
-      : line_(line), symbols_(symbols) {}
+      : line_(line),
+        symbols_(symbols) {
+    symbols_.erase(std::remove_if(symbols_.begin(), symbols_.end(),
+                                  [](const string& v) { return v.empty(); }),
+                   symbols_.end());
+  }
 
   size_t line_length() const { return line_.size(); }
   bool needs_alignment() const { return !symbols_.empty(); }
   void add_prefix(const string& prefix) { line_ = prefix + line_; }
-  string printable_line(size_t desired_length, size_t max_length) const;
+  string printable_line(size_t min_length, size_t max_length) const;
 
  private:
   string line_;                     // '#include XXX' or 'class YYY;'
   vector<string> symbols_;          // symbols used from included header
 };
 
-string OutputLine::printable_line(size_t desired_length, size_t max_length)
-    const {
-  // Reduce max_length to make room for the ", etc" suffix.
-  max_length -= strlen(", etc");
+// Append a helpful 'why' comment to the include line, containing the symbols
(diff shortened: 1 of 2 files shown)


iwyu_output.cc
```diff
@@ -72,53 +72,60 @@ class OutputLine {
...
-      ++symbols_printed;
-      prefix = ", ";       // before 2nd and subsequent symbols, print ', '
-    } else {
-      // Truncate at max_length cols.
-      if (symbols_printed > 0)
-        retval += ", etc";
-      break;
+// Append a helpful 'why' comment to the include line, containing the symbols
+// used from the header. Align nicely at lower verbosity levels.
+string OutputLine::printable_line(size_t min_length, size_t max_length) const {
```
vsapsai: Should we have some CHECK_ that max_length >= min_length and that max_length > 0 or some other reasonable value?

iwyu_output.cc
```diff
@@ -72,53 +72,60 @@ class OutputLine {
...
+  if (symbols_.empty())
+    return line_;
+
+  --max_length;  // Spare a char for newline character.
+
+  // Before first symbol, print '  // for ' and pad it so 'why' comments are
+  // nicely aligned.
+  string symbol_prefix = "  // for ";
+  if (min_length > line_.length())
+    symbol_prefix.insert(0, min_length - line_.length(), ' ');
```
vsapsai: I find `if (line_.length() < min_length)` more readable. Cannot explain why, maybe it is just a personal taste. It is up to you to decide in which order to write this comparison.

iwyu_output.cc
```diff
@@ -72,53 +72,60 @@ class OutputLine {
...
+  string result = line_;
+  for (string symbol : symbols_) {
+    // At verbose levels 0-2, truncate output to max_length columns.
+    if (!ShouldPrint(3)) {
+      // Calculate number of chars remaining.
+      size_t remaining = 0;
+      size_t result_length = result.length() + symbol_prefix.length();
+      if (max_length > result_length)
+        remaining = max_length - result_length;
+
```
vsapsai: And here too, I prefer `result_length < max_length`. Interesting, if there is statistic data regarding usage of `<` and `>`.

vsapsai:
Code looks correct, just a few suggestions.

iwyu_output.cc
kimgr: `max_length > 0` is required for correctness, so I added a `CHECK_` for that.

I think it's legal for `min_length` to be larger than `max_length`, however, e.g. if an `#include` line exceeds the configured max length, and that should be handled by the algorithm as-is.

kimgr:
Thanks, updated the diff with the conditions turned around and a `CHECK_` added.

iwyu_output.cc
vsapsai: Good point that `min_length` can be larger than `max_length`.

vsapsai:
No more comments, you are good to go.

PR #205, opened 2015-08-20
Update to reflect changes in LLVM.

vsapsai:
LLVM r245318 removed IPA library, folding the code into the main
Analysis library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,7 +133,6 @@ target_link_libraries(include-what-you-use
   LLVMScalarOpts
   LLVMInstCombine
   LLVMTransformUtils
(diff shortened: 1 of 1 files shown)


kimgr:
This builds on Windows, so I'm happy with it.

I just noticed I've already created a `clang_3.7` branch, so this can go right in, as far as I'm concerned. There are a few commits we may be able to cherry-pick from `master`, but that's a separate activity.

vsapsai:
Thanks for checking this, Kim.

Will need to check how cherry-picking goes. Using Git instead of SVN makes things different. Will write to the mailing list with my thoughts on this subject.

PR #203, opened 2015-07-30
Markdown readme

tpltnt:
Hi,

i converted the README into proper markdown for better display on github.
Cheers,
tpltnt

PR #206, opened 2015-08-22
Update docs

kimgr:
Inspired by @tpltnt's PR #203, this is an attempt to convert all the docs to Markdown and update them to reflect the current state of the world:

- No more referencing code.google.com
- Remove outdated information
- Simplify build instructions
- ...

Once this goes in, I think `make_readme.py` can be updated to basically just merge the files together in a reasonable order.

### DIFF
--- a/docs/IWYUMappings.txt
+++ b/docs/IWYUMappings.txt
@@ -1,4 +1,4 @@
-# IWYU Mappings #
+## IWYU Mappings ##
 
 One of the difficult problems for IWYU is distinguishing between which header contains a symbol definition and which header is the actual documented header to include for that symbol.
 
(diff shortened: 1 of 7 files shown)


vsapsai:
I would suggest to rename files to *.md so that GitHub renders them as Markdown. I've tried it at [my fork](https://github.com/vsapsai/include-what-you-use/blob/update-docs/docs/WhyIWYU.md) and it looks nice, except of broken links to *.txt files.

The rest looks fine, thanks for you and @tpltnt for working on this.

As for updating `make_readme.py` I was thinking about creating a new short introductory README file because after concatenation you receive a wall of text which is too intimidating.

kimgr:
Thanks, yes, I thought I'd make the content edits first, and then rename them in a separate changeset.

A shorter README sounds like a good idea, but the build instructions are pretty useful. Maybe we should have make_readme exclude e.g. the What* and Why* files?

I'll merge this as-is, anyway, and then push a change to rename all pages to .md.

PR #208, opened 2015-09-02
Rewrite relative links so they can be followed on GitHub.

kimgr:
I'll just merge this, it seems non-controversial. Feel free to revert it if you disagree.


vsapsai:
The change is fine. I believe there might be edge cases that `make_readme.py` won't be able to cope with. But I don't care about those cases.

kimgr:
Yes, I figure we'll know 'em when we see 'em.

Issue #70, opened 2015-06-15
Do no include "detail" headers


vsapsai:
Originally reported on Google Code with ID 70
```
Boost often separates out implementation details into a "detail" folder, e.g. <boost/thread/detail/thread.hpp>.
It would be nice if iwyu did not suggest these detail headers.

```

Reported by `leszek.swirski` on 2012-06-18 23:47:53

delaitre:
Just wondering if it would be a good idea to commit those mapping files so that anybody could easily propose improvements. That would hopefully allow to have a fine tuned mapping in few months.

vsapsai:
Thank you, @delaitre. That was a great idea to commit those mapping files.

Closing the issue as it seems to be resolved. For other Boost mapping fixes and enhancements please open separate issues/pull requests.

PR #209, opened 2015-09-07
Collect Boost and Qt mappings created by Scott Howard aka @maqifrnswa.

vsapsai:
As @delaitre suggested in issue #70, here are Boost and Qt mappings created by Scott Howard. I haven't tested the mappings, this pull request is mostly for visibility. So that other people can check out these mappings easily and play with them.


maqifrnswa:
Those were for an older version of boost, but you're right - visibility is needed both as an example for others to use to generate mappings for other libraries an to improve these. The commands to generate them programmatically is included and should be a good starting point.

kimgr:
Great! I started thinking about where the mappings should be installed, etc, and then I lost track of it entirely.

I wonder if it would be useful to include Boost version in the mapping filenames? That way we can add user contributions for other Boost versions side-by-side, if they turn out to be different.

Not sure about versioning for Qt4, I don't use Qt myself, so I don't know how stable it is.

One thing we might want to do moving forward is to build a library of Python scripts for generating mappings for various big libraries. Windows is sadly undertalented when it comes to shell one-liners.

cdglove:
For what it's worth, I tested this over the weekend on one personal project, and it worked well.  The scripts aren't perfect however because the boost headers and directories are also not perfect.  For example, a couple of headers that were suggested to be included were obviously base/impl headers that were not meant to be individually included.  This actually broke my compile in those cases. I'm not sure how to solve that.

kimgr:
Let's commit this as-is, I think it's better to iterate on it in-tree.

maqifrnswa:
> ...The scripts aren't perfect however because the boost headers and directories are also not perfect.

Boost is hard because there is no standard way of laying out boost libraries, so creating a general script is hard. The best I could do is try to find general trends, and then manually handling corner cases. However, now that it's in tree, hopefully it will start to work itself out as there are more eyes on it. Thanks all

vsapsai:
Merged change manually to fix typo in commit message. Didn't specify Boost version because it is not clear how much different versions differ.

PR #212, opened 2015-09-17
Fix issue #127: macro locations

kimgr:
This is a replacement for PR #189 and I believe it actually works (with test to prove it!)

Background: the challenge is that macro authors want to be able to push #include responsibility onto callers of the macro by introducing a heuristic forward-declare (only used for the purpose of signalling to IWYU that the macro file is not responsible.)

IWYU's prior approach to macro locations was too eager, it committed to the instantiation loc already in `CurrentLoc`, where we didn't really know what the location would be used for.

While working on delaying this decision, I found a bug with locations of unary operators, which confused the location for things like `#define ARRAYSIZE(x) sizeof(x) / sizeof(*(x))` where the second `sizeof` would register as a use, but its location could not be mapped to the macro argument.

Then I instituted the new rules:

- If a use location is known to be part of a macro argument, we attribute it to callers. Anything passed as argument to a macro is the caller's responsibility.
- If a use location is in scratch space, it's the result of concatenation of macro arguments. See above
- If there's a forward declaration of the used decl in the file defining the macro, the macro rejects responsiblity and the caller becomes responsible (this matches what we do for typedefs.)

Unfortunately the last one turned out to be problematic -- if the macro file is part of the IWYU analysis, a "nudging" forward-declare will typically not have any use inside the macro file, and IWYU would suggest that we remove it. Since it was basically just a marker, IWYU would short-circuit itself, and the next IWYU run would attribute the symbol to the macro, causing all sorts of weirdness.

The last commit fixes that problem by reporting the marker fwd-decl as used by the macro's spelling location.

These changes finally made it possible for me to change the macro_location test to also include the forward-declare construct.

@dpunset, please try this out in your codebase. @vsapsai, please review this critically, I'd like to refine it a bit more, but I've worked on it so long I'm basically blind.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -431,17 +431,7 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
 
   SourceLocation CurrentLoc() const {
     CHECK_(current_ast_node_ && "Call CurrentLoc within Visit* or Traverse*");
-    const SourceLocation loc = current_ast_node_->GetLocation();
-    if (!loc.isValid())
-      return loc;
-    // If the token is formed via macro concatenation, the spelling
-    // location will be in <scratch space>.  Use the instantiation
-    // location instead.
-    const SourceLocation spelling_loc = GetSpellingLoc(loc);
-    if (IsInMacro(loc) && StartsWith(PrintableLoc(spelling_loc), "<scratch "))
-      return GetInstantiationLoc(loc);
-    else
-      return spelling_loc;
+    return current_ast_node_->GetLocation();
   }
 
   string CurrentFilePath() const {
@@ -1380,30 +1370,71 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return type;
   }
 
-  // Re-assign uses from macro authors to macro callers when possible.
-  // With macros, it can be very difficult to tell what types are the
-  // responsibility of the macro-caller, and which the macro-author.  e.g.
-  //    #define SIZE_IN_FOOS(ptr)  ( sizeof(*ptr) / sizeof(Foo) )
-  // Here the type of *ptr is the responsibility of the caller, while the
-  // type of Foo is the responsibility of the author, even though the
-  // Exprs for *ptr and Foo are both located inside the macro (the Expr
-  // for ptr is located at the macro-calling site, but not for *ptr).
-  // To help us guess the owner, we use this rule: if the macro-file
-  // intends-to-provide the type, then we keep ownership of the type
-  // at the macro.  Otherwise, we assume it's with the caller.  This
-  // works well as long as the file defining the macro is well-behaved.
-  //   This function should be called with a use-loc that is within
-  // an expanded macro (so the use-loc will point to either inside a
-  // macro definition, or to an argument to a macro call).  If it
-  // points within a macro definition, and that macro-definition file
-  // does not mean to re-export the symbol being used, then we reassign
-  // use of the decl to the macro-caller.
-  SourceLocation GetUseLocationForMacroExpansion(SourceLocation use_loc,
-                                                 const Decl* used_decl) {
-    CHECK_(IsInMacro(use_loc) && "Unexpected non-macro-expansion call");
-    if (!preprocessor_info().PublicHeaderIntendsToProvide(
-            GetFileEntry(use_loc), GetFileEntry(used_decl)))
-      return GetInstantiationLoc(use_loc);
+  // Return any forward-declare of decl earlier in the same file as loc.
+  const NamedDecl* GetForwardDeclareInSameFile(const Decl* decl,
+                                               SourceLocation loc) {
+    set<const NamedDecl*> redecls = GetClassRedecls(DynCastFrom(decl));
+    for (const NamedDecl* redecl : redecls) {
+      if (IsBeforeInSameFile(redecl, loc)) {
+        return redecl;
+      }
+    }
+
+    return nullptr;
+  }
+
+  // Get the canonical use location for a (location, decl) pair.
+  // Decide whether the file expanding the macro or the file defining the macro
+  // should be held responsible for a use.
+  SourceLocation GetCanonicalUseLocation(SourceLocation use_loc,
+                                         const NamedDecl* decl) {
+    // If we're not in a macro, just echo the use location.
+    if (!IsInMacro(use_loc))
+      return use_loc;
+
+    // If we're in a macro, and the macro file forward-declares decl, make sure
+    // we keep that forward-declaration.
+    const NamedDecl* fwd_decl = GetForwardDeclareInSameFile(decl, use_loc);
+    if (fwd_decl) {
+      const FileEntry* used_in = GetFileEntry(use_loc);
+      preprocessor_info().FileInfoFor(used_in)->ReportForwardDeclareUse(
+          use_loc, decl, IsNodeInsideCXXMethodBody(current_ast_node()),
+          nullptr);
+    }
+
+    // Resolve the best use location based on our current knowledge.
+    //
+    // 1) If the macro definition file forward-declares the used decl, that's a
+    //    hint that it wants the expansion location to take responsibility.
+    // 2) If the symbol being used is passed as an argument to the macro, we
+    //    want the expansion location to take responsibility (that thing they
+    //    passed in is something they know about, and the macro doesn't.)
+    // 3) If the use_loc is in <scratch space>, we assume it's formed by macro
+    //    argument concatenation, and attribute responsibility to the expansion
+    //    location.
+    //
+    // Otherwise, the macro file is responsible for including the symbol.
+    SourceLocation expansion_loc = GetInstantiationLoc(use_loc);
+    const char* responsible = "expansion";
+
+    if (fwd_decl != nullptr) {
+      VERRS(5) << "Macro file has a forward-decl attribution hint.\n";
+      use_loc = expansion_loc;
+    } else if (IsInScratchSpace(use_loc)) {
+      VERRS(5) << "Decl was used in <scratch space>, presumably as a result of "
+                  "macro arg concatenation.\n";
+      use_loc = expansion_loc;
+    } else if (GlobalSourceManager()->isMacroArgExpansion(use_loc)) {
+      VERRS(5) << "Use location is a macro arg expansion.\n";
+      use_loc = expansion_loc;
+    } else {
+      responsible = "definition";
+    }
+
+    VERRS(5) << "Attributing " << decl->getNameAsString() << " use to macro "
+             << responsible << " location at " << PrintableLoc(use_loc)
+             << ".\n";
+
     return use_loc;
   }
 
@@ -1621,11 +1652,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreDecl(target_decl))
       return;
 
-    // Figure out the best location to attribute uses inside macros.
-    if (IsInMacro(used_loc))
-      used_loc = GetUseLocationForMacroExpansion(used_loc, target_decl);
+    // Canonicalize the use location and report the use.
+    used_loc = GetCanonicalUseLocation(used_loc, target_decl);
     const FileEntry* used_in = GetFileEntry(used_loc);
-
     preprocessor_info().FileInfoFor(used_in)->ReportFullSymbolUse(
         used_loc, target_decl, IsNodeInsideCXXMethodBody(current_ast_node()),
         comment);
@@ -1694,11 +1723,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreDecl(target_decl))
       return;
 
-    // Figure out the best location to attribute uses inside macros.
-    if (IsInMacro(used_loc))
-      used_loc = GetUseLocationForMacroExpansion(used_loc, target_decl);
+    // Canonicalize the use location and report the use.
+    used_loc = GetCanonicalUseLocation(used_loc, target_decl);
     const FileEntry* used_in = GetFileEntry(used_loc);
-
     preprocessor_info().FileInfoFor(used_in)->ReportForwardDeclareUse(
         used_loc, target_decl, IsNodeInsideCXXMethodBody(current_ast_node()),
         comment);
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2194,7 +2194,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       if (!CanIgnoreType(dereftype))
         // This reports even if the expr ends up not being a reference, but
         // that's ok (if potentially redundant).
-        ReportTypeUse(GetLocation(arg_expr), dereftype);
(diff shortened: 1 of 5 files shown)


dpunset:
should I just sync the latest trunk to test this?

kimgr:
Yes, it's developed against this repo's master (it's been a while since I synced the LLVM tree, though, so I hope it's up to date...)

iwyu.cc
```diff
@@ -1380,30 +1370,53 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+  const NamedDecl* GetForwardDeclareInSameFile(const Decl* decl,
+                                               SourceLocation loc) {
+    set<const NamedDecl*> redecls = GetClassRedecls(DynCastFrom(decl));
+    for (Each<const NamedDecl*> it(&redecls); !it.AtEnd(); ++it) {
+      if (IsBeforeInSameFile(*it, loc)) {
+        return *it;
+      }
+    }
+
+    return NULL;
```
vsapsai: Not critical, you might want to try range-based for loop. And maybe also `nullptr`.

iwyu.cc
```diff
@@ -1380,30 +1370,53 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+
+  // Assuming use_loc is part of a macro expansion, decide whether the file
+  // expanding the macro or the file defining the macro should be held
+  // responsible for the used_decl.
+  // The rules are:
+  // 1) If the macro definition file forward-declares the used decl, that's a
+  //    hint that it wants callers to take responsibility.
+  // 2) If the symbol being used is passed as an argument to the macro, we want
+  //    callers to take responsibility (that thing they passed in is something
+  //    they know about, and the macro doesn't)
```
vsapsai: Here the sentence doesn't end with a period.

iwyu.cc
```diff
@@ -1380,30 +1370,53 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+  // 3) If the use_loc is in <scratch space>, we assume it's formed by macro
+  //    argument concatenation, and attribute responsibility to callers.
+  // Otherwise, the macro file is responsible for including the symbol.
+  //
+  // Returns the responsible location.
   SourceLocation GetUseLocationForMacroExpansion(SourceLocation use_loc,
                                                  const Decl* used_decl) {
```
vsapsai: Looks like now `used_decl` isn't used in the function. Does it matter we aren't doing private-to-public mapping check here? If not, it is better to remove `used_decl` from the signature.

iwyu.cc
```diff
@@ -1380,30 +1370,53 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+                  "macro arg concatenation.\n";
+      use_loc = caller_loc;
+    } else if (GlobalSourceManager()->isMacroArgExpansion(use_loc)) {
+      VERRS(5) << "Use location is a macro arg expansion.\n";
+      use_loc = caller_loc;
+    } else {
+      responsible = "macro";
+    }
+
+    VERRS(5) << "Attributing use to " << responsible << ".\n";
```
vsapsai: Do comments have enough context? Especially in the case of attributing to macro it looks like the comment will be "Attributing use to macro". And it is unclear what use, to what macro.

iwyu.cc
```diff
@@ -1585,10 +1598,23 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+            comment);
+
+        // Attribute use to caller
+        used_loc = GetInstantiationLoc(used_loc);
+      } else {
+        used_loc = GetUseLocationForMacroExpansion(used_loc, decl);
+      }
+    }
 
+    used_in = GetFileEntry(used_loc);
```
vsapsai: You could set `used_in` in different ways, for example, inside previous if-block. But it doesn't matter, it's just a matter of taste and I have no arguments that some of the approaches is better, the only difference so far is probably nicer diff.

iwyu.cc
```diff
@@ -1650,13 +1676,34 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        // Make sure we keep the heuristic forward decl.
+        preprocessor_info().FileInfoFor(used_in)->ReportForwardDeclareUse(
+            used_loc, decl, IsNodeInsideCXXMethodBody(current_ast_node()),
+            comment);
+
+        // Attribute use to caller
+        used_loc = GetInstantiationLoc(used_loc);
+      } else {
+        used_loc = GetUseLocationForMacroExpansion(used_loc, decl);
+      }
```
vsapsai: It is the same code as in `ReportDeclUseWithComment` but I don't know how to avoid code duplication. This code block requires `decl`, `used_loc`, returns `used_loc`, `fwd_decl` - looks like extracting it in a function will have too many arguments for input and output. Kim, have you tried this? Maybe it won't be so bad.

iwyu.cc
```diff
@@ -1650,13 +1676,34 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    }
 
-    preprocessor_info().FileInfoFor(used_in)->ReportForwardDeclareUse(
-        used_loc, decl, IsNodeInsideCXXMethodBody(current_ast_node()),
-        comment);
+    used_in = GetFileEntry(used_loc);
+    if (isMacroDefinedAndUsedInSameFile || !fwd_decl) {
```
vsapsai: It requires more attention to understand that in case of no macro `isMacroDefinedAndUsedInSameFile` will be false and `fwd_decl` will be NULL, so you need to report forward declare use. Maybe have this condition inside `IsInMacro` block and maybe it will be easier to understand when condition is telling when you want to avoid `ReportForwardDeclareUse`.

Another issue, don't know how important it is. `fwd_decl` is a pointer but is used as a boolean variable. Will it be better to have it as a boolean variable?

iwyu_location_util.cc
```diff
@@ -142,6 +143,10 @@ SourceLocation GetLocation(const clang::Stmt* stmt) {
   } else if (const ConditionalOperator* conditional_op =
              DynCastFrom(stmt)) {
     return conditional_op->getQuestionLoc();
+  } else if (const UnaryOperator* unary_op = DynCastFrom(stmt)) {
+    // Drill through unary operators and parentheses, to get at the underlying
+    // DeclRefExpr or whatever, e.g. '*(x)' should give the location of 'x'
+    stmt = unary_op->getSubExpr()->IgnoreParenImpCasts();
```
vsapsai: Nice how you are handling unary operator location. It is better than lexing macros.

dpunset:
Kim forgive my noobness with github but I don't see how to sync all your changes from the pull request into my local repository. I synced the main branch but obviously your fixes are not there, and I can't find an option to fork your branch or apply your changes. Apologies, it's the first time I use GitHub. Can you tell me please?

vsapsai:
@dpunset, at the bottom of the thread there is a message "This branch is up-to-date with the base branch", it has a link "command line instructions". To get Kim's changes locally you need to do
```
git checkout -b kimgr-127-macro-locations master
git pull https://github.com/kimgr/include-what-you-use.git 127-macro-locations
```

And then you'll have Kim's changes in branch kimgr-127-macro-locations. If Kim updates the code, you'll be able to pull his changes in this branch again.

dpunset:
@vsapsai thank you. I don't see the link you mention, I only see

"This branch is up-to-date with the base branch
Only those with write access to this repository can merge pull requests."

However I managed to run your commands on git shell and it worked. I get the idea now. I merged his changelists and will do some tests today hopefully.

Thank you both!

dpunset:
Hi Kim,

I've been using your changelist today, but I still have issues where macros that I think are related to this problem.

There is a macro that implements a global function. This global function is just declaring a static variable whom in the constructor takes as a parameter the return value of a function declared in a namespace. This function is declared in a namespace, and IWYU suggests to remove this header in the places where this macro is used.

Something like this:
''''
function.h
namespace mynamespace
{
    cmyclass& getmyclassmydomain();
}

macro.h
#define macroA(myparam) \
void myfunc() { \
static mynamespace::cotherclass s_var(mynamesapce::getmyclass##myparam()); }

#define macroB(myparam) \
some_other_macro() \
macroA(myparam)

usage.cpp
#include "macro.h"
#include "function.h"
macroB(mydomain)

''''

IWYU suggests to remove #include "function.h"

Debugging a bit I see that macroB appears in the list of "use" but macroA does not.

kimgr:
@vsapsai: oops, I didn't see your comments until now, I didn't get any notification for your comments. I'll try and look through them tonight, thanks!

kimgr:
@dpunset: I haven't had a chance to debug it yet, but you shouldn't be worrying about the use of `macroA` (that's an implementation detail of `macroB`), but rather why `macroA´ doesn't attribute the use of `mynamespace::getmyclass##myparam` to the caller. It might be that it gets stuck in the nested macro expansion, could you try inlining `macroA` into `macroB` and see if it makes any difference?

dpunset:
@kimgr I tried, but it doesn't seem to appear either. macroB reads like this now:

#define macroB(myparam) \
some_other_macro() \
void myfunc() { \
static mynamespace::cotherclass s_var(mynamesapce::getmyclass##myparam()); }

no reference to myfunc, s_var, mynamespace::cotherclass, etc...

dpunset:
I am trying to replicate a few of the issues we have when we have macros involved. I haven't managed to fully replicate the most annoying one we have, but at least I managed to reproduce a strange situation. Consider this:
```
baseclass.h
#pragma once

#define BASE_MACRODEF(NAME)  \
    static const char* GetName() { return #NAME; }

class CBaseClass
{
public:

    BASE_MACRODEF(CBaseClass)

private:
};


myclass.h
#pragma once

#include "baseclass.h"

class CMyClass : public CBaseClass
{
public:

    BASE_MACRODEF(CMyClass)

private:
};


mytemplate.h
#pragma once

template <class T>
class CMyTemplate
{
public:
    CMyTemplate() : m_name(T::GetName()) {}

private:
    const char*             m_name;
};


test.cpp
#include "stdafx.h"
#include "myclass.h"
#include "mytemplate.h"

int _tmain(int argc, _TCHAR* argv[])
{
    CMyTemplate<CMyClass> myTemplate;

	return 0;
}
```
It suggest this:

#include "stdafx.h"
#include "ConsoleApplication3/baseclass.h"   // for CMyClass::GetName
#include "ConsoleApplication3/myclass.h"     // for CMyClass
#include "ConsoleApplication3/mytemplate.h"  // for CMyTemplate

baseclass.h is not needed in fact, it's wrong to say that it is needed for CMyClass::GetName

I debugged the issue, and I found that in IwyuFileInfo::ReportFullSymbolUse, calling GetFileEntry on the decl received from ReportDeclUseWithComment, it points to baseclass.h, not myclass.h

Isn't that strange? Is it because the macro is defined in baseclass.h?

kimgr:
That does seem backwards. I can't explain off-hand how this happens. But it looks like there's more macro expansion going on in iwyu_output, so maybe I haven't covered all of them yet :-/

iwyu.cc
kimgr: I'm working through this now, and I can't make any sense of it myself. I'll see if I can make it clearer.

kimgr:
@vsapsai  - addressed your comments, which led to an almost-rewrite, much clearer now, in my opinion!

kimgr:
@dpunset - thanks for testing this, and sorry about the long hiatus.

The first case (where a symbol is conjured in the macro based on a macro argument) *should* be attributed to the macro expansion location, so that looks like a bug.

The second case (where the symbol is misattributed to the macro spelling location instead of the expansion location) is, as you hinted at, due to additional macro location logic in `iwyu_output.cc`. I think this logic is wrong (it appears entirely backwards) and needs to be smarter overall, following similar rules as when we find the use location.

I'll look into these, but possibly separately. I'd like to get this larger change in place first.

kimgr:
> The first case (where a symbol is conjured in the macro based on a macro argument)
> should be attributed to the macro expansion location, so that looks like a bug.

OK, so I finally think I understand what's going on here.

The macro is defined like so:

    #define MACRO(param) \
        static ns::Class var = ns::getClass##param();

Our rules say that any use that depends on in-arguments to a macro should be attributed to the expansion location, so it would seem natural that ns::Class is attributed to the macro definition and ns::getClassFoo is attributed to the expansion (when called as `MACRO(Foo)`.)

However, the logic is entirely based on locations, and locations appear to be token-granular. So when we ask the source manager if the location of `ns::getClassFoo` is the result of a macro concatenation it will say 'no', because it only evaluates the `ns` token, not the actual `getClassFoo` token. So it seems like more complex uses in macros (with elaborated types, larger expressions, etc) won't be correctly attributed.

I haven't come up with a good way of solving this, either. There's no information about the semantic structure *of the use* available at this point in processing, all we have is a location and the declaration of what was used.

I'm tempted to call this a known limitation and move on.

iwyu.cc
```diff
@@ -1380,30 +1370,70 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      use_loc = expansion_loc;
+    } else if (GlobalSourceManager()->isMacroArgExpansion(use_loc)) {
+      VERRS(5) << "Use location is a macro arg expansion.\n";
+      use_loc = expansion_loc;
+    } else {
+      responsible = "definition";
+    }
+
+    VERRS(5) << "Attributing use to macro " << responsible << " location at "
+             << PrintableLoc(use_loc) << ".\n";
```
vsapsai: What about something like
```c++
    VERRS(5) << "Attributing " << decl->getNameAsString() << " use "
             << "to macro " << responsible << " location at "
             << PrintableLoc(use_loc) << ".\n";
```
? I also tried `PrintableDecl(decl)` but it outputs
```
Attributing class Foo {
} use to macro expansion location at tests/cxx/macro_location.h:19:15.
```

iwyu.cc
```diff
@@ -1622,10 +1652,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       return;
 
     // Figure out the best location to attribute uses inside macros.
-    if (IsInMacro(used_loc))
-      used_loc = GetUseLocationForMacroExpansion(used_loc, target_decl);
+    used_loc = GetCanonicalUseLocation(used_loc, target_decl);
```
vsapsai: Comment is still correct but not as good as before. I don't know if you have better ideas to change the comment.

iwyu.cc
```diff
@@ -1380,30 +1370,70 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    for (const NamedDecl* redecl : redecls) {
+      if (IsBeforeInSameFile(redecl, loc)) {
+        return redecl;
+      }
+    }
+
+    return nullptr;
+  }
+
+  // Get the canonical use location for a location,decl pair.
```
vsapsai: Shouldn't be there a space after comma?

vsapsai:
Everything looks fine, have a suggestion to improve logging.

kimgr:
Thanks, pushed a revised change set.

kimgr:
I'd like to push this to get more macro cases working, and then we can work on all the variations separately (e.g. by pushing `ASTNode` forward in the pipeline). @vsapsai - are you OK with that?

vsapsai:
I'm OK with pushing this change.

Issue #169, opened 2015-06-15
Fails to see through std::remove_pointer


vsapsai:
Originally reported on Google Code with ID 169

```
Put this code into a source file in tests/cxx:

traits_remove_pointer.cc:
#include <type_traits>                                                            

#include "tests/cxx/indirect.h"                                                   


std::remove_pointer<IndirectClass*>::type x;   

Run iwyu like so:

$ include-what-you-use -I. tests/cxx/traits_remove_pointer.cc

What is the expected output? What do you see instead?

This should produce no IWYU changes, but IWYU suggests that IndirectClass should be
forward-declared rather than included.

Issue found by Daniel Jump, reported on the mailing list.
```

Reported by `kim.grasman` on 2015-01-11 17:05:13

PR #214, opened 2015-09-21
Part of fix for the issue #166.  Follow typedef chains in templates.

vsapsai:
First of all, it doesn't fix the issue #166 completely. Now IWYU suggests to include `<memory>  // for allocator_traits<>::value_type` instead of `<ext/alloc_traits.h>  // for __alloc_traits<>::value_type`. But I believe it is the movement in the right direction.

I want to mention that in this change I didn't try to handle alias templates and nested templates. Another issue is that most likely `VisitMemberExpr` is not the only place that should use `GetFirstNonTemplateTypedef`. My plan is to investigate if `GetCallerResponsibleTypesForTypedef` should use it, todos look quite promising.

Also I wasn't able to come up with a test case that executes lines
```c++
if (underlying_typedef->getDecl()->getLexicalDeclContext() == parent) {
  return GetFirstNonTemplateTypedef(underlying_typedef, parent);
}
```
Looks like Clang doesn't allow such cases but I prefer to keep this code to not rely on current Clang implementation and because its intention is correct.

In the process of fixing the issue I've discovered the following cases:
```c++
// tests/cxx/issue166_extra_test-i1.h
#ifndef DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_ISSUE166_EXTRA_TEST_I1_H_
#define DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_ISSUE166_EXTRA_TEST_I1_H_

template<typename T>
class Container {
 public:
  typedef T value;
  typedef value& reference;

  value content_;
  reference getContent() { return content_; }
};

#endif  // DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_ISSUE166_EXTRA_TEST_I1_H_
```
```c++
// tests/cxx/issue166_extra_test-d1.h
#include "tests/cxx/issue166_extra_test-i1.h"

template<typename T>
class NonContainer {
 public:
  typedef Container<T> value;
  typedef value& reference;

  value container_;
  reference getContainer() { return container_; }
};
```
```c++
// tests/cxx/issue166_extra_test.cc
#include "tests/cxx/issue166_extra_test-d1.h"

void usage() {
  NonContainer<IndirectClass> nc;
  nc.getContainer().getContent().Method();
}
```
```c++
// tests/cxx/issue166_extra_test2.cc
#include "tests/cxx/issue166_extra_test-i1.h"

void usage() {
  Container<IndirectClass> c;
}
```

`tests/cxx/issue166_extra_test.cc` is not that serious, it merely suggest to add `#include "tests/cxx/issue166_extra_test-i1.h"  // for Container`. `tests/cxx/issue166_extra_test2.cc` on the other hand suggests to replace `#include "tests/cxx/indirect.h"` with a forward declaration which breaks the compilation. I think it is almost identical to the issue #169 and can be fixed by applying `GetFirstNonTemplateTypedef` in the right place.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1475,6 +1475,37 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  // Returns the first type that is not a typedef in a template.  For example,
+  // for template
+  //
+  //   template<typename T> class Foo {
+  //     typedef T value_type;
+  //     typedef value_type& reference;
+  //   };
+  //
+  // for type 'reference' it will return type T with which Foo was instantiated.
+  const Type* DesugarDependentTypedef(const TypedefType* typedef_type) {
+    const DeclContext* parent
+        = typedef_type->getDecl()->getLexicalDeclContext();
+    if (const ClassTemplateSpecializationDecl* template_parent
+        = DynCastFrom(parent)) {
+      return DesugarDependentTypedef(typedef_type, template_parent);
+    }
+    return typedef_type;
+  }
+
+  const Type* DesugarDependentTypedef(
+      const TypedefType* typedef_type, const RecordDecl* parent) {
+    const Type* underlying_type
+        = typedef_type->getDecl()->getUnderlyingType().getTypePtr();
+    if (const TypedefType* underlying_typedef = DynCastFrom(underlying_type)) {
+      if (underlying_typedef->getDecl()->getLexicalDeclContext() == parent) {
+        return DesugarDependentTypedef(underlying_typedef, parent);
+      }
+    }
+    return underlying_type;
+  }
+
   set<const Type*> GetCallerResponsibleTypesForTypedef(
       const TypedefDecl* decl) {
     set<const Type*> retval;
@@ -2086,6 +2117,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         = expr->isArrow() ? RemovePointerFromType(base_type) : base_type;
     if (CanIgnoreType(base_type) && CanIgnoreType(deref_base_type))
       return true;
+    if (const TypedefType* typedef_type = DynCastFrom(deref_base_type)) {
+      deref_base_type = DesugarDependentTypedef(typedef_type);
+    }
     // Technically, we should say the type is being used at the
     // location of base_expr.  That may be a different file than us in
     // cases like MACRO.b().  However, while one can imagine
--- a/tests/cxx/typedef_chain_in_template-d1.h
+++ b/tests/cxx/typedef_chain_in_template-d1.h
@@ -1,0 +1,22 @@
+//===--- typedef_chain_in_template-d1.h - test input file for iwyu --------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// This header mimics std::vector in libstdc++.
+
+#include "tests/cxx/typedef_chain_in_template-i1.h"
+
+template<typename T>
+class ContainerAsLibstdcpp {
+ public:
+  typedef T value_type;
+  typedef typename TypedefWrapper<T>::reference reference;
+  reference GetContent() { return content_; }
+ private:
+  value_type content_;
+};
--- a/tests/cxx/typedef_chain_in_template-d4.h
+++ b/tests/cxx/typedef_chain_in_template-d4.h
@@ -1,0 +1,25 @@
+//===--- typedef_chain_in_template-d4.h - test input file for iwyu --------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Container that tests typedef chains longer than in libc++.
+
+template<typename T>
+class ContainerLongTypedefChain {
+ public:
+  typedef T value_type;
+  typedef value_type& reference;
+  typedef reference& reference_reference;
+  reference_reference GetContent1() { return content_; }
+
+  typedef reference reference2;
+  typedef reference2 reference3;
+  reference3 GetContent2() { return content_; }
+ private:
+  value_type content_;
+};
--- a/tests/cxx/typedef_chain_class.h
+++ b/tests/cxx/typedef_chain_class.h
@@ -1,0 +1,21 @@
+//===--- typedef_chain_class.h - test input file for iwyu -----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Simple class to use in typedef chains.  Similar to IndirectClass but without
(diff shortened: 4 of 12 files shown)


iwyu.cc
```diff
@@ -1475,6 +1475,37 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+  // Returns the first type that is not a typedef in a template.  For example,
+  // for template
+  //
+  //   template<typename T> class Foo {
+  //     typedef T value_type;
+  //     typedef value_type& reference;
+  //   };
+  //
+  // for type 'reference' it will return type T with which Foo was instantiated.
+  const Type* GetFirstNonTemplateTypedef(const TypedefType* typedef_type) {
```
kimgr: The name is a little clunky. Isn't this a form of desugaring? `DesugarTemplateTypedef` sounds more decisive :-)

tests/cxx/typedef_chain_in_template-d1.h
```diff
@@ -0,0 +1,22 @@
...
+// This header mimics std::vector in libstdc++.
+
+#include "tests/cxx/typedef_chain_in_template-i1.h"
+
+template<typename T>
+class ContainerAsLibstdcpp {
+ public:
+  typedef T value_type;
+  typedef typename TypedefWrapper<T>::reference reference;
+  reference getContent() { return content_; }
```
kimgr: Here and throughout; you've started using camelCase rather than PascalCase for functions.

tests/cxx/typedef_chain_in_template-d4.h
```diff
@@ -0,0 +1,25 @@
...
+//===----------------------------------------------------------------------===//
+
+// Container that tests typedef chains longer than in libc++.
+
+template<typename T>
+class ContainerLongTypedefChain {
+ public:
+  typedef T value_type;
+  typedef value_type& reference;
+  typedef reference& reference_reference;
```
kimgr: This is only valid in C++11, right? Or is it valid to form a universal/rvalue reference using nested typedefs? Apparently Clang is happy with it anyway...

I guess `-std=c++11` (or later) is enabled everywhere by default now, so it's probably OK.

kimgr:
Looks neat! This is another example of a general theme -- uses of caller-provided symbols are pushed toward the caller, others are pulled into the containing decl (e.g. template arguments, macro arguments, etc)

iwyu.cc
vsapsai: I like your suggestion. I haven't thought about it as a desugaring but it makes perfect sense.

tests/cxx/typedef_chain_in_template-d1.h
vsapsai: Good point, will change.

tests/cxx/typedef_chain_in_template-d4.h
vsapsai: I think `reference_reference` is not a rvalue. Don't remember exactly but AST doesn't look like rvalue. And also when I run `clang -fsyntax-only -std=c++03 -I . tests/cxx/typedef_chain_in_template.cc` or `clang -fsyntax-only -std=c++98 -I . tests/cxx/typedef_chain_in_template.cc` - clang doesn't complain.

I wasn't trying to test rvalues, these are my frantic attempts to trigger execution of
```c++
if (underlying_typedef->getDecl()->getLexicalDeclContext() == parent) {
  return GetFirstNonTemplateTypedef(underlying_typedef, parent);
}
```

I decided to leave this test case for more complete test suite and to test that cases like this don't confuse IWYU.

vsapsai:
I renamed functions and left typedefs intact. For typedefs we are using PascalCase but tests are based on libc++ and libstdc++ so I think it is better to use the same typedefs `value_type`, `reference`.

tests/cxx/typedef_chain_in_template-d4.h
kimgr: I think this is reference collapsing -- http://stackoverflow.com/a/3772234/96963 -- if it affects IWYU analysis it's definitely a good thing to cover.

kimgr:
I'm trying to understand where `DesugarTemplateTypedef` stops -- it unfolds typedef until it reaches a non-typedef type, or a typedef declared inside another decl-context, right?

Oh, I think I see now, it tries to avoid crossing the streams over to another type, e.g. alloc_traits?

One case I'm missing is a non-template record containing a typedef of a traits-template, something like:

    struct A {
       typedef Traits<int>::reference_type reference_type;
    };

It looks like this is similar to one of your NonContainer tests, but they currently don't use TypedefWrapper, as far as I can see.

I'm not sure if this would work with the current desugaring routine, is it even called inside a non-template?

Looks good otherwise!

vsapsai:
`DesugarTemplateTypedef` unfolds typedefs declared in templates that depend on template arguments. And it stops when it hits non-typedef or when a typedef is declared outside of template. This check "declared outside of template" is done by checking parent decl-context and it might be wrong for nested templates, for example. I've decided not to go further in this direction until we have real code that experience problems with nested templates.

> Oh, I think I see now, it tries to avoid crossing the streams over to another type, e.g. alloc_traits?
Check for parent is more for the cases when you pass typedef as a template argument.

If typedef parent is non-template, attempt to cast it to `ClassTemplateSpecializationDecl` will fail and we won't desugar it.

```c++
    const DeclContext* parent
        = typedef_type->getDecl()->getLexicalDeclContext();
    if (const ClassTemplateSpecializationDecl* template_parent
        = DynCastFrom(parent)) {
      return DesugarTemplateTypedef(typedef_type, template_parent);
    }
```

All tests for typedefs inside non-templates are the same and are covered by `typedef_chain_no_follow-d2.h`.

tests/cxx/typedef_chain_in_template-d4.h
vsapsai: It definitely looks like reference collapsing. There was no indication in debugger it affects IWYU - Clang hides this from us. I'll keep the test as a specific case of elongating typedef chain.

kimgr:
On second thought 'template typedef' might be a bad choice of words, that's a language feature that's been suggested several times over the years for a construct like:

    template<class T>
    typedef std::vector<T> Container;

    Container<int> c; // this is a std::vector<int>

but it never seems to have come around to anything. The closest we have now is alias templates. I wonder if `DesugarDependentTypedefs` would be less semantically overloaded? Sorry about the bikeshedding.

> If typedef parent is non-template, attempt to cast it to `ClassTemplateSpecializationDecl` will fail
> and we won't desugar it.

OK, so for my example, the absence of desugaring would return `A::reference_type`, right? Yep, that seems like the right thing to do.

So I saw you added this to `VisitMemberExpr` -- that's somewhat arbitrary, right? I guess it catches a lot of cases we've seen in practice, such as method calls returning typedefs, etc. It sounds like it would be nice to get this analysis into `GetCallerResponsibleTypesForTypedef`, but I can't really tell what kind of consequences that would have. Time will tell, I guess :-)

I'm perfectly happy with this as-is, but if you think you can iron out more issues, feel free to keep hammering at it.

vsapsai:
Renamed to `DesugarDependentTypedef`. No problem with better naming. In this case I am failing to understand how clear is the name.

Yes, with absent desugaring you'll get `A::reference_type`.

Adding desugaring to `VisitMemberExpr` is done because that's where the problem is in issue #166. But that's not the right place or at least not the only place to do so.

PR #218, opened 2015-09-27
Fix #216: Allow user to specify IWYU path to test

### LINKED ISSUE #216: Test script doesn't look in the PWD for include-what-you-use binary.
Sorry if this is a duplicate but after having built `include-what-you-use` I attempted to run `./run_iwyu_tests.py` with the following error:

```
earnest % ./run_iwyu_tests.py
  File "./run_iwyu_tests.py", line 25, in <module>
    import iwyu_test_util
  File "/home/earnest/build/lab/include-what-you-use-git/src/include-what-you-use/iwyu_test_util.py", line 97, in <module>
    _IWYU_PATH = _GetIwyuPath(_IWYU_PATHS + _SYSTEM_PATHS)
  File "/home/earnest/build/lab/include-what-you-use-git/src/include-what-you-use/iwyu_test_util.py", line 76, in _GetIwyuPath
    raise IOError('Failed to locate IWYU.\nSearched\n %s' % '\n '.join(iwyu_paths))
IOError: Failed to locate IWYU.
Searched
 ../../../../Debug+Asserts/bin
 ../../../../Release+Asserts/bin
 ../../../../Release/bin
 ../../../../build/Debug+Asserts/bin
 ../../../../build/Release+Asserts/bin
 ../../../../build/Release/bin
 ../../../../../build/Debug+Asserts/bin
 ../../../../../build/Release+Asserts/bin
 ../../../../../build/Release/bin
 ../../../../../build/bin/Debug
 ../../../../../build/bin/Release
 ../../../../../build/bin/MinSizeRel
 ../../../../../build/bin/RelWithDebInfo
 /home/earnest/.local/bin
 /usr/local/sbin
 /usr/local/bin
 /usr/bin
 /usr/bin/site_perl
 /usr/bin/vendor_perl
 /usr/bin/core_perl
```

It seems the script is not looking for the `include-what-you-use` binary that was just created and currently resides in the same working directory as `run_iwyu_tests`.

What am I doing wrong?

kimgr:
Add an optional command-line argument to run_iwyu_tests.py, to specify
which IWYU binary to test with. Usage example:

  $ ./run_iwyu_tests.py -- ./include-what-you-use

The '--' delimiter is used to separate args going to Python's unittest
from the new arg used by the test runner.

### DIFF
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -67,15 +67,7 @@ def _Which(program, paths):
     return None
 
 
-def _GetIwyuPath(iwyu_paths):
-  """Returns the path to IWYU or raises IOError if it cannot be found."""
-  iwyu_path = _Which('include-what-you-use', iwyu_paths)
-  if iwyu_path:
-      return iwyu_path
-
-  raise IOError('Failed to locate IWYU.\nSearched\n %s' % '\n '.join(iwyu_paths))
-
-
+_IWYU_PATH = None
 _SYSTEM_PATHS = [p.strip('"') for p in os.environ["PATH"].split(os.pathsep)]
 _IWYU_PATHS = [
     '../../../../Debug+Asserts/bin',
@@ -94,7 +86,27 @@ _IWYU_PATHS = [
     '../../../../../build/bin/MinSizeRel',
     '../../../../../build/bin/RelWithDebInfo',
     ]
-_IWYU_PATH = _GetIwyuPath(_IWYU_PATHS + _SYSTEM_PATHS)
+
+
+def SetIwyuPath(iwyu_path):
+  """Set the path to the IWYU executable under test.
+  """
+  global _IWYU_PATH
+  _IWYU_PATH = iwyu_path
+
+
+def _GetIwyuPath():
+  """Returns the path to IWYU or raises IOError if it cannot be found."""
+  global _IWYU_PATH
+
+  if not _IWYU_PATH:
+    iwyu_paths = _IWYU_PATHS + _SYSTEM_PATHS
+    _IWYU_PATH = _Which('include-what-you-use', iwyu_paths)
+    if not _IWYU_PATH:
+      raise IOError('Failed to locate IWYU.\nSearched\n %s' %
+                    '\n '.join(iwyu_paths))
+
+  return _IWYU_PATH
 
 
 def _GetCommandOutput(command):
@@ -404,7 +416,7 @@ def TestIwyuOnRelativeFile(test_case, cc_file, cpp_files_to_check,
 
   # TODO(csilvers): verify that has exit-status 0.
   cmd = '%s %s %s %s' % (
-      _IWYU_PATH, ' '.join(iwyu_flags), ' '.join(clang_flags), cc_file)
+      _GetIwyuPath(), ' '.join(iwyu_flags), ' '.join(clang_flags), cc_file)
   if verbose:
     print('>>> Running %s' % cmd)
   output = _GetCommandOutput(cmd)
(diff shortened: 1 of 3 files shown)


Earnestly:
This does not seem to function with either: `./run_iwyu_tests.py -- include-what-you-use
` or `./run_iwyu_tests.py -- ./include-what-you-use`. Is the innvocation incorrect?

kimgr:
Uh, oops. It searches the path before applying the command-line-provided path. I'll look into fixing that, but it will need a bigger effort, so it won't be today.

kimgr:
There, this should be better.

Earnestly:
Yes, this works.  Thanks

iwyu_test_util.py
```diff
@@ -404,7 +416,7 @@ def TestIwyuOnRelativeFile(test_case, cc_file, cpp_files_to_check,
 
   # TODO(csilvers): verify that has exit-status 0.
   cmd = '%s %s %s %s' % (
-      _IWYU_PATH, ' '.join(iwyu_flags), ' '.join(clang_flags), cc_file)
+      _GetIwyuPath(), ' '.join(iwyu_flags), ' '.join(clang_flags), cc_file)
```
vsapsai: Need to measure how repeated calling _GetIwyuPath() affects testing time.

run_iwyu_tests.py
```diff
@@ -167,6 +176,11 @@ def RegisterFilesForTesting(rootdir, pattern):
 
 
 if __name__ == '__main__':
+  unittest_args, additional_args = Partition(sys.argv, '--')
+  if additional_args:
+    iwyu_test_util.SetIwyuPath(additional_args[0])
+  sys.argv = unittest_args
```
vsapsai: I think it is cleaner to do `unittest.main(argv=unittest_args)` if I understand [Python docs](https://docs.python.org/2/library/unittest.html#unittest.main) correctly.

kimgr: Oh, cool, that completely escaped me when I looked up the docs for `unittest`. Will fix.

iwyu_test_util.py
kimgr: The new `_GetIwyuPath` caches the result of the search for IWYU, so I can't imagine it makes much difference. But I'll try before/after tests on Windows and FreeBSD.

kimgr:
Here are some timings:

    Before
    ------
    Windows hog
        84 tests in 196.924s
        84 tests in 200.092s
        84 tests in 201.192s

    FreeBSD monster
        84 tests in 30.368s
        84 tests in 29.964s
        84 tests in 29.987s

    After
    -----
    Windows hog
        84 tests in 206.032s
        84 tests in 197.168s
        84 tests in 197.289s 

    FreeBSD monster
        84 tests in 31.948s
        84 tests in 30.404s
        84 tests in 30.014s

It looks like it might be slightly slower, but I'd say it's still within the noise.

I'll push the latest diff in a minute.

Earnestly:
> “Any measurement that you make without the knowledge of its uncertainty is completely meaningless.”
— Walter Lewin

iwyu_test_util.py
vsapsai: Sorry, I've missed that `_GetIwyuPath` caches its result. Performance concern was a false alarm.

vsapsai:
This change looks fine to me. It might be useful to tell about this option in Instructions for Developers

kimgr:
Good idea, done.

vsapsai:
Docs update is good, have nothing else to add.

kimgr:
OK, I'll create a new branch and clean up commit history a little bit, and then push it. Thanks!

Earnestly:
Thank you very much too

PR #219, opened 2015-09-27
Fix #211: Install scripts to bin/

### LINKED ISSUE #211: Install fix_includes.py and iwyu_tool.py
I think will be good idea to include helper scripts (fix_includes.py and iwyu_tool.py) into installation.

Such scripts are installed for Clang-format and Clang-tidy. See http://reviews.llvm.org/rL247890 for Clang-tidy implementation for CMake and configure.

kimgr:
Have CMake install both fix_includes.py and iwyu_tool.py to bin/.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,5 +167,5 @@ else()
   endif()
 endif()
 
-install(TARGETS include-what-you-use
-  RUNTIME DESTINATION bin)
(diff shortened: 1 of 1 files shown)


vsapsai:
The change looks good from the point of view of somebody who doesn't install include-what-you-use in order to use it. So I don't know if it follows the best practices.

kimgr:
Merged this manually. Closing, thanks!

PR #221, opened 2015-10-05
Add better support for using declarations to make them less greedy.

cdglove:
Cleaned up version of Pull Request #200 .  Resolves issue #174

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1608,26 +1608,33 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // The comment, if not NULL, is extra text that is included along
   // with the warning message that iwyu emits.
   virtual void ReportDeclUseWithComment(SourceLocation used_loc,
-                                        const NamedDecl* decl,
+                                        const NamedDecl* used_decl,
                                         const char* comment) {
+    const NamedDecl* target_decl = used_decl;
+
+    // Sometimes a shadow decl comes between us and the 'real' decl.
+    if (const UsingShadowDecl* shadow_decl = DynCastFrom(used_decl))
+      target_decl = shadow_decl->getTargetDecl();
+    
     // Map private decls like __normal_iterator to their public counterpart.
-    decl = MapPrivateDeclToPublicDecl(decl);
-    if (CanIgnoreDecl(decl))
+    target_decl = MapPrivateDeclToPublicDecl(target_decl);
+    if (CanIgnoreDecl(target_decl))
       return;
 
     // Figure out the best location to attribute uses inside macros.
     if (IsInMacro(used_loc))
-      used_loc = GetUseLocationForMacroExpansion(used_loc, decl);
+      used_loc = GetUseLocationForMacroExpansion(used_loc, target_decl);
     const FileEntry* used_in = GetFileEntry(used_loc);
 
     preprocessor_info().FileInfoFor(used_in)->ReportFullSymbolUse(
-        used_loc, decl, IsNodeInsideCXXMethodBody(current_ast_node()),
+        used_loc, target_decl, IsNodeInsideCXXMethodBody(current_ast_node()),
         comment);
 
     // Sometimes using a decl drags in a few other uses as well:
 
     // If we're a use that depends on a using declaration, make sure
-    // we #include the file with the using declaration.
+    // we #include the file with the using declaration. Need to check
+    // the original reported decl so we don't lose the shadow information.
     // TODO(csilvers): check that our getQualifier() does not match
     // the namespace of the decl.  If we have 'using std::vector;' +
     // 'std::vector<int> foo;' we don't actually care about the
@@ -1636,8 +1643,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // instead?  We can call it "Use what you use". :-)
     // TODO(csilvers): check for using statements and namespace aliases too.
     if (const UsingDecl* using_decl
-        = GetUsingDeclarationOf(decl, GetDeclContext(current_ast_node()))) {
-      preprocessor_info().FileInfoFor(used_in)->ReportFullSymbolUse(
+        = GetUsingDeclarationOf(used_decl, 
+              GetDeclContext(current_ast_node()))) {
+      preprocessor_info().FileInfoFor(used_in)->ReportUsingDeclUse(
           used_loc, using_decl, IsNodeInsideCXXMethodBody(current_ast_node()),
           "(for using decl)");
     }
@@ -1648,7 +1656,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // anywhere.  ('autocast' is similar, but is handled in
     // VisitCastExpr; 'fn-return-type' is also similar and is
     // handled in HandleFunctionCall.)
-    if (const TypedefDecl* typedef_decl = DynCastFrom(decl)) {
+    if (const TypedefDecl* typedef_decl = DynCastFrom(target_decl)) {
       // One exception: if this TypedefType is being used in another
       // typedef (that is, 'typedef MyTypedef OtherTypdef'), then the
       // user -- the other typedef -- is never responsible for the
@@ -1674,26 +1682,33 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // The comment, if not NULL, is extra text that is included along
   // with the warning message that iwyu emits.
   virtual void ReportDeclForwardDeclareUseWithComment(SourceLocation used_loc,
-                                                      const NamedDecl* decl,
+                                                      const NamedDecl* used_decl,
                                                       const char* comment) {
-    decl = MapPrivateDeclToPublicDecl(decl);
-    if (CanIgnoreDecl(decl))
+    const NamedDecl* target_decl = used_decl;
+
+    // Sometimes a shadow decl comes between us and the 'real' decl.
+    if (const UsingShadowDecl* shadow_decl = DynCastFrom(used_decl))
+      target_decl = shadow_decl->getTargetDecl();
+
+    target_decl = MapPrivateDeclToPublicDecl(target_decl);
+    if (CanIgnoreDecl(target_decl))
       return;
 
     // Figure out the best location to attribute uses inside macros.
     if (IsInMacro(used_loc))
-      used_loc = GetUseLocationForMacroExpansion(used_loc, decl);
+      used_loc = GetUseLocationForMacroExpansion(used_loc, target_decl);
     const FileEntry* used_in = GetFileEntry(used_loc);
 
     preprocessor_info().FileInfoFor(used_in)->ReportForwardDeclareUse(
-        used_loc, decl, IsNodeInsideCXXMethodBody(current_ast_node()),
+        used_loc, target_decl, IsNodeInsideCXXMethodBody(current_ast_node()),
         comment);
 
     // If we're a use that depends on a using declaration, make sure
     // we #include the file with the using declaration.
     if (const UsingDecl* using_decl
-        = GetUsingDeclarationOf(decl, GetDeclContext(current_ast_node()))) {
-      preprocessor_info().FileInfoFor(used_in)->ReportFullSymbolUse(
+        = GetUsingDeclarationOf(used_decl, 
+              GetDeclContext(current_ast_node()))) {
+      preprocessor_info().FileInfoFor(used_in)->ReportUsingDeclUse(
           used_loc, using_decl, IsNodeInsideCXXMethodBody(current_ast_node()),
           "(for using decl)");
     }
@@ -2600,10 +2615,11 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return visitor_state_->preprocessor_info;
   }
 
-  void AddUsingDeclaration(const NamedDecl* target_decl,  // what's being used
-                           const UsingDecl* using_decl) {
-    visitor_state_->using_declarations.insert(make_pair(target_decl,
-                                                        using_decl));
+  void AddShadowDeclarations(const UsingDecl* using_decl) {
+    for (const UsingShadowDecl* shadow : using_decl->shadows()) {
+      visitor_state_->using_declarations.insert(
+          make_pair(shadow->getTargetDecl(), shadow->getUsingDecl()));
+    }
   }
 
  private:
@@ -2618,17 +2634,23 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   }
 
   const UsingDecl* GetUsingDeclarationOf(const NamedDecl* decl,
-                                         const DeclContext* using_context) {
-    // We look through all the using-decls of the given decl.  We
-    // limit them to ones that are visible from the decl-context we're
-    // currently in (that is, what namespaces we're in).  Of those, we
-    // pick the one that's in the same file as decl, if possible,
-    // otherwise we pick one arbitrarily.
+                                         const DeclContext* use_context) {
+    // First, if we have a UsingShadowDecl, then we don't need to do anything
+    // because we can just directly return the using decl from that.
+    if (const UsingShadowDecl* shadow = DynCastFrom(decl))
+      return shadow->getUsingDecl();
+
+    // But, if we don't have a UsingShadowDecl, then we need to look through
+    // all the using-decls of the given decl.  We limit them to ones that are
+    // visible from the decl-context we're currently in (that is, what
+    // namespaces we're in), via the check through 'Encloses'. Of those, we
+    // pick the one that's in the same file as decl, if possible, otherwise we
+    // pick one arbitrarily.
     const UsingDecl* retval = NULL;
     vector<const UsingDecl*> using_decls
         = FindInMultiMap(visitor_state_->using_declarations, decl);
     for (Each<const UsingDecl*> it(&using_decls); !it.AtEnd(); ++it) {
-      if (!(*it)->getDeclContext()->Encloses(using_context))
+      if (!(*it)->getDeclContext()->Encloses(use_context))
         continue;
       if (GetFileEntry(decl) == GetFileEntry(*it) ||    // in same file, prefer
           retval == NULL) {        // not in same file, but better than nothing
@@ -3427,14 +3449,24 @@ class IwyuAstConsumer
     ParseFunctionTemplates(context.getTranslationUnitDecl());
 
     TraverseDecl(context.getTranslationUnitDecl());
+   
+    const set<const FileEntry*>* const files_to_report_iwyu_violations_for
+        = preprocessor_info().files_to_report_iwyu_violations_for();
+
+    // Some analysis, such as UsingDecl resolution, is deferred until the
+    // entire AST is visited because it's only at that point that we know if
+    // the symbol was actually used or not.
+    // We perform that analysis here before CalculateAndReportIwyuViolations.
+    for (const FileEntry* file : *files_to_report_iwyu_violations_for) {
+      CHECK_(preprocessor_info().FileInfoFor(file));
+      preprocessor_info().FileInfoFor(file)->ResolvePendingAnalysis();
+    }
 
     // We have to calculate the .h files before the .cc file, since
     // the .cc file inherits #includes from the .h files, and we
     // need to figure out what those #includes are going to be.
     size_t num_edits = 0;
     const FileEntry* const main_file = preprocessor_info().main_file();
-    const set<const FileEntry*>* const files_to_report_iwyu_violations_for
-        = preprocessor_info().files_to_report_iwyu_violations_for();
     for (Each<const FileEntry*> file(files_to_report_iwyu_violations_for);
          !file.AtEnd(); ++file) {
       if (*file == main_file)
@@ -3512,21 +3544,17 @@ class IwyuAstConsumer
     // information here.  Note: we have to store this even if this is
     // an ast node we would otherwise ignore, since other AST nodes
     // (which we might not ignore) can depend on it.
-    for (UsingDecl::shadow_iterator it = decl->shadow_begin();
-         it != decl->shadow_end(); ++it) {
-      AddUsingDeclaration((*it)->getTargetDecl(), decl);
-    }
-
-    if (CanIgnoreCurrentASTNode())  return true;
+    AddShadowDeclarations(decl);
 
     // The shadow decls hold the declarations for the var/fn/etc we're
     // using.  (There may be more than one if, say, we're using an
-    // overloaded function.)  We check to make sure nothing we're
-    // using is an iwyu violation.
-    for (UsingDecl::shadow_iterator it = decl->shadow_begin();
-         it != decl->shadow_end(); ++it) {
-      ReportDeclForwardDeclareUse(CurrentLoc(), (*it)->getTargetDecl());
-    }
+    // overloaded function.)  We don't want to add all of them at once
+    // though, because that will drag in every overload even if we're
+    // only using one.  Instead, we keep track of the using decl and
+    // mark it as touched when something actually uses it.
+    preprocessor_info().FileInfoFor(CurrentFileEntry())->AddUsingDecl(decl);
+
+    if (CanIgnoreCurrentASTNode())  return true;
 
     return Base::VisitUsingDecl(decl);
   }
@@ -3657,7 +3685,14 @@ class IwyuAstConsumer
   // Called whenever a variable, function, enum, etc is used.
   bool VisitDeclRefExpr(clang::DeclRefExpr* expr) {
     if (CanIgnoreCurrentASTNode())  return true;
-    ReportDeclUse(CurrentLoc(), expr->getDecl());
+    // Special case for UsingShadowDecl to track UsingDecls correctly. The
+    // actual decl will be reported by obtaining it from the UsingShadowDecl
+    // once we've tracked the UsingDecl use.
+    if (const UsingShadowDecl* found_decl = DynCastFrom(expr->getFoundDecl())) {
+      ReportDeclUse(CurrentLoc(), found_decl);
+    } else {
+      ReportDeclUse(CurrentLoc(), expr->getDecl());
+    }
     return Base::VisitDeclRefExpr(expr);
   }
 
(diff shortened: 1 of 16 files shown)


tests/cxx/using_aliased_symbol-alias.h
```diff
@@ -0,0 +1,15 @@
+//===--- using_aliased_symbol-declare.h - test input file for iwyu --------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
```
cdglove: I will add a description to this.

tests/cxx/using_aliased_symbol-alias.h
```diff
@@ -0,0 +1,15 @@
+//===--- using_aliased_symbol-declare.h - test input file for iwyu --------===//
```
cdglove: Name is wrong, I'll correct it.

tests/cxx/using_aliased_symbol-declare.h
```diff
@@ -0,0 +1,13 @@
+//===--- using_aliased_symbol-declare.h - test input file for iwyu --------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
```
cdglove: Also description missing here.  I'll add one.

cdglove: Just realized this is a header, which is why there's no test description; there's no test.

tests/cxx/using_aliased_symbol-alias.h
cdglove: Just realized this is a header, which is why there's no test description; there's no test.

iwyu.cc
```diff
@@ -1577,26 +1577,34 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // The comment, if not NULL, is extra text that is included along
   // with the warning message that iwyu emits.
   virtual void ReportDeclUseWithComment(SourceLocation used_loc,
-                                        const NamedDecl* decl,
+                                        const NamedDecl* used_decl,
                                         const char* comment) {
+
```
kimgr: You can kill this blank line.

iwyu.cc
```diff
@@ -2566,10 +2582,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return visitor_state_->preprocessor_info;
   }
 
-  void AddUsingDeclaration(const NamedDecl* target_decl,  // what's being used
-                           const UsingDecl* using_decl) {
-    visitor_state_->using_declarations.insert(make_pair(target_decl,
-                                                        using_decl));
+  void AddShadowDeclarations(const UsingDecl* using_decl) {
+    for (UsingDecl::shadow_iterator it = using_decl->shadow_begin();
```
kimgr: You could use a range for-loop here and make this much nicer:

    for (const UsingShadowDecl* shadow : using_decl->shadows()) {
       ...
    }

iwyu.cc
```diff
@@ -2585,11 +2603,17 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
   const UsingDecl* GetUsingDeclarationOf(const NamedDecl* decl,
                                          const DeclContext* using_context) {
```
kimgr: I liked when you renamed this to `use_context`, I thought that was much better, can you make that change here too?

iwyu.cc
```diff
@@ -2585,11 +2603,17 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-    // otherwise we pick one arbitrarily.
+    // First, if we have a UsingShadowDecl, then we don't need to do anything
+    // because we can just directly return the using decl from that.
+    if (const UsingShadowDecl* shadow = DynCastFrom(decl))
+      return shadow->getUsingDecl();
+
+    // But, if we don't have a UsingShadowDecl, then we need to look through
+    // all the using-decls of the given decl.  We limit them to ones that are
+    // visible from the decl-context we're currently in (that is, what
+    // namespaces we're in), via the check through 'Encloses'. Of those, we
```
kimgr: I wonder if this should also include `Equals` as I mentioned on the other PR? See, `Encloses` only checks if the `using_context` is an "ancestor" of `using_decl_context`, not if they're the same. Then again, maybe the shadow decl checks cover this case?

I've also been thinking about this more generally, I wonder if what we really want to do is a proper lookup? Not sure how that works, though, but it might remove some boilerplate here. For future reference :-)

iwyu.cc
```diff
@@ -3393,14 +3417,25 @@ class IwyuAstConsumer
...
     TraverseDecl(context.getTranslationUnitDecl());
+   
+    const set<const FileEntry*>* const files_to_report_iwyu_violations_for
+          = preprocessor_info().files_to_report_iwyu_violations_for();
+
+    // Some analysis, such as UsingDecl resolution, is deferred until the
+    // entire AST is visited because it's only at that point that we know if
+    // the symbol was actually used or not.
+    // We perform that analysis here before CalculateAndReportIwyuViolations.
+    for (Each<const FileEntry*> file(files_to_report_iwyu_violations_for);
```
kimgr: Range-for again. We're slowly trying to transition to range-for as we change code, but we don't want to do a global conversion just yet.

iwyu_output.cc
```diff
@@ -613,6 +627,28 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
...
+  // traversing the AST, we check to see if a using decl is unreferenced and
+  // add a full use of one of its shadow decls so that the source file
+  // continues to compile.
+  map<const UsingDecl*, bool>::iterator using_decl_status = 
+    using_decl_referenced_.find(using_decl);
+
+  if (using_decl_status != using_decl_referenced_.end()) {
+    using_decl_status->second = true;
+  }
+
```
kimgr: ... but what happens if it's not found? I seem to recall there was some valid case where that could happen, but I don't remember now...

iwyu_output.cc
```diff
@@ -1845,6 +1881,31 @@ size_t PrintableDiffs(const string& filename,
 
 }  // namespace internal
 
+void IwyuFileInfo::ResolvePendingAnalysis() {
+  // Resolve using declarations before continuing.  This handles the case
```
kimgr: 'before continuing' seems redundant...

iwyu_output.cc
```diff
@@ -1845,6 +1881,31 @@ size_t PrintableDiffs(const string& filename,
...
+  // sure everything still compiles instead of removing the using decl. A
+  // more thorough approach would be to scan the current list of includes that
+  // already name this decl (like in the overloaded function case) and include
+  // one of those so we don't include a file we don't actually need.
+  for (map<const UsingDecl*, bool>::value_type using_decl_status
+    : using_decl_referenced_) {
+    if (!using_decl_status.second) {
+      const UsingDecl* using_decl = using_decl_status.first;
+      ReportForwardDeclareUse(using_decl->getUsingLoc(),
+        using_decl->shadow_begin()->getTargetDecl(),
```
kimgr: Is it guaranteed that `using_decl->shadows()` is non-empty?

tests/cxx/badinc.cc
```diff
@@ -1237,11 +1235,13 @@ int main() {
   i1_template_method_only_class.s<I2_Class*>(local_i2_class_ptr);
   // IWYU: I1_Class is...*badinc-i1.h
   // IWYU: I1_Class needs a declaration
+  // IWYU: I1_TemplateMethodOnlyClass<I1_Class, I1_Union>::s is...*badinc-i1.h
```
kimgr: All the changes in this file make me nervous... Given how completely impenetrable `badinc.cc` is, I get the feeling you've just patched it after-the-fact to shut the test suite up... Which in turn could indicates that some critical functionality changed behavior. But I can't pretend to know what the basis for those changes are, so maybe you could explain?

tests/cxx/include_with_using.cc
```diff
@@ -41,7 +41,6 @@ void TestDeclFromD4() {
   var_in_d4 = 0;
 }
 
-// IWYU: i1::UsingInCc needs a declaration
```
kimgr: This is also really surprising to me, I don't understand which of your changes made IWYU stop diagnosing on this.

tests/cxx/using_aliased_symbol_unused.cc
```diff
@@ -0,0 +1,28 @@
...
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Tests that when we include a file that contains a using decl, that we are not
+// forced to include that file even if the decl is not used.
```
kimgr: scratch "even"?

tests/cxx/using_multiple_overload.cc
```diff
@@ -0,0 +1,31 @@
...
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Tests that when we use multiple function overloads through a using decl, 
+// that we correctly include all of the necessary files for the overload and
+// don't accidently remove files greedily.
```
kimgr: Typo: accidentally.

Really nice test-case, this one!

kimgr:
Thanks, this looks really good to me! I'm mostly worried about the `badinc` test, but I realize there's not much chance making any sense of it...

iwyu.cc
cdglove: Done.

iwyu.cc
cdglove: OK

iwyu.cc
cdglove: Sure can.

iwyu.cc
cdglove: Yes, I have a desire to change this whole thing to not use this map at all. But, I haven't figured out a way to do that (I failed at several attempts).  I believe the information is in there though, so it should be possible.

iwyu.cc
cdglove: OK.

iwyu_output.cc
cdglove: Not found is OK because here, we are looking to see if a using decl is referenced in the same file it is declared in. Only then do we care to mark the using decl as referenced. This is strictly to cover the case where a file locally declares a using decl but doesn't actually use it, so we need to force the include of a symbol that satisfies the the using decl;  

ie: if we have;

```cpp
#include <vector>
using std::vector;
```
But then never actually use a vector, we will try to remove the vector #include and break compilation.

iwyu_output.cc
cdglove: Yes.  This is my train of thought flowing out into comments. I'll clean it up.

iwyu_output.cc
cdglove: I believe so otherwise the file would not compile and we'd never get here.  That being said, I think at least an assert is in order to document that.

cdglove:
Yes, badinc.cc looks like it evolved from something much smaller. It's very difficult to reason about it because it's not contained.  That being said, I'm going to try to come up with a satisfactory explanation for the differences in the comments.  The first set, where I added extra silences seems to be OK because it seems like those messages were probably missing before. But why?  I'm not entirely sure.

The second ones, where the using decl warning is removed is probably because we don't add the shadow_decls early anymore so it's changed the violation logic around those decls.  I'll look into it.

tests/cxx/include_with_using.cc
cdglove: The reason is because we don't actually 'use' anything here at this point anymore, it's completely deferred to the use point.  Previously, we would add all of the shadow decls when coming across a using decl, but now we don't so there's nothing to issue a warning against.  I think this should be fine because the warning will simple occur at the point of use instead of here.

iwyu.cc
cdglove: When I change it from !Encloses to !Encloses  && !Equals, everything still works. I will try to come up with a test that breaks without the Equals to show that the Equals fixes something.

cdglove: I'm unable to break it. Everything I do that seems should require equials, passes with or without it, so I think we should leave it as is for now.

Interestingly, along the way, the only case that I could find that engages this code path is Variable declarations.  Everything else ends up passing though VisitDeclRefExpr, which catches the ShadowDecls.  It would be interesting to know *why* VarDecls aren't also caught by DeclRefExpr because if it's a bug in Clang, fixing it might remove the need for this map entirely.

iwyu_output.cc
kimgr: Oh, thanks for the explanation. So `using_decl_referenced_` is really `using_decl_referenced_in_same_file_`? I'm not suggesting a name change (too painfully long), but it might be worth making that clearer somehow.

iwyu.cc
kimgr: Thanks! No need to sweat it, I've tried to come up with relevant cases, too, but I can't seem to find anything that affects the flow the way I expect :-)

tests/cxx/include_with_using.cc
kimgr: Cool, I understand now that it's the deferred analysis that does it, and I ran it again locally and saw that the warning actually happens on line 48 instead. Now I'm happy! :-)

iwyu_output.cc
cdglove: It is explained in the comment;

```cpp
// If accessing a symbol through a using decl in the same file that contains
  // the using decl, we must mark the using decl as referenced. At the end of
  // traversing the AST, we check to see if a using decl is unreferenced and
  // add a full use of one of its shadow decls so that the source file
  // continues to compile.
```

Maybe using_decl_referenced_locally_ is a better name?

iwyu.cc
```diff
@@ -3623,7 +3651,7 @@ class IwyuAstConsumer
   // Called whenever a variable, function, enum, etc is used.
   bool VisitDeclRefExpr(clang::DeclRefExpr* expr) {
     if (CanIgnoreCurrentASTNode())  return true;
-    ReportDeclUse(CurrentLoc(), expr->getDecl());
+    ReportDeclUse(CurrentLoc(), expr->getFoundDecl());
```
cdglove: This is the change that triggered the other IWYU warnings from badinc.cc to change.  getFoundDecl may or may not return the same decl as getDecl.  The details are a little vague, but it seems the getFoundDecl is sort of more specific in that it includes nested types, which is causing things like I1_TemplateMethodOnlyClass<I1_Struct, I1_Union>::t to show up whereas previously only I1_TemplateMethodOnlyClass would.  I'm not sure what the exact right thing is.  It seems to me that we probably in general want the more specific thing, but in the case of nested types, this seems superfluous because the a missing parent decl implies a missing child decl.  In the mean time, I can restore the previous functionality by changing the code as follows;

'''cpp
  bool VisitDeclRefExpr(clang::DeclRefExpr* expr) {
    if (const UsingShadowDecl* found = DynCastFrom(expr->getFoundDecl())) {
      ReportDeclUse(CurrentLoc(), found);
    } else {
      ReportDeclUse(CurrentLoc(), expr->getDecl());
    }
    return Base::VisitDeclRefExpr(expr);
  }
'''

iwyu_output.cc
kimgr: Heh, I read the comment three times looking for clues before posting, but I was apparently not awake enough to see it :-)

I think the name is fine, and the comment makes it clear what's going on, let's drop this.

iwyu.cc
kimgr: Ah, makes sense! Yeah, let's change `VisitDeclRefExpr` to be backward-compatible, and we can think about simplifying later (that `UsingShadowDecl` cast seems like it's happening everywhere now :-))

cdglove: Done.

cdglove:
Ping. Just checking in. Is there any other feedback on this?

vsapsai:
I haven't reviewed the pull request in detail. On Mac OS X there is a regression. Or to be more correct, with libc++, I believe. `using_single_overload` fails with

```
AssertionError:
Unexpected summary diffs for tests/cxx/using_single_overload.cc:
+++
@@ -4,4 +4,4 @@
 - #include <list>  // lines XX-XX

 The full include-list for tests/cxx/using_single_overload.cc:
-#include <vector>  // for swap, vector
+#include <vector>  // for allocator, swap, vector
---
```

I don't see yet how your change subverts our attempts to ignore default `<vector>` stuff like `allocator`.

cdglove:
On Sun, 18 Oct 2015 at 20:37 Volodymyr Sapsai <notifications@github.com>
wrote:

> I haven't reviewed the pull request in detail. On Mac OS X there is a
> regression. Or to be more correct, with libc++, I believe.
> using_single_overload fails with
>
> AssertionError:
> Unexpected summary diffs for tests/cxx/using_single_overload.cc:
> +++
> @@ -4,4 +4,4 @@
>  - #include <list>  // lines XX-XX
>
>  The full include-list for tests/cxx/using_single_overload.cc:
> -#include <vector>  // for swap, vector
> +#include <vector>  // for allocator, swap, vector
> ---
>
> I don't see yet how your change subverts our attempts to ignore default
> <vector> stuff like allocator.
>
I can confirm that the issue occurs on Linux with libc++ as well, so it's
not Mac OS specific. However, that particular unit test, which I added as
part of this change, fails even when run against head with libc++.

…
Unexpected summary diffs for tests/cxx/using_single_overload.cc:
+++
@@ -1,7 +1,17 @@
 tests/cxx/using_single_overload.cc should add these lines:
+#include <__bit_reference>
+#include <__split_buffer>
+#include <memory>
+#include <type_traits>
+#include <utility>

 tests/cxx/using_single_overload.cc should remove these lines:
-- #include <list>  // lines XX-XX
…

 The full include-list for tests/cxx/using_single_overload.cc:
-#include <vector>  // for swap, vector
+#include <__bit_reference>  // for swap
+#include <__split_buffer>  // for swap
+#include <list>  // for swap
+#include <memory>  // for swap
+#include <type_traits>  // for swap
+#include <utility>  // for swap
…
+#include <vector>  // for allocator, swap, vector
---

Of course, I expect the additional includes for swap, because that's what I
was intending to fix, but unexpected allocator is present here as well.

vsapsai:
You are right, Chris. It's just that we have `stl_container_provides_allocator.cc` and `precomputed_tpl_args.cc` which also test we don't mention `allocator`. But your test discovered we don't handle the case when you specify vector size. I.e. `std::vector<int> v;` - no allocator, `std::vector<int> v(10);` - mention allocator. For this I'll test more on Linux and submit a separate issue.

iwyu.cc
```diff
@@ -3478,21 +3510,17 @@ class IwyuAstConsumer
...
-         it != decl->shadow_end(); ++it) {
-      ReportDeclForwardDeclareUse(CurrentLoc(), (*it)->getTargetDecl());
-    }
+    // overloaded function.)  We don't want to add all of them at once
+    // though, because that will drag in every overload even if we're
+    // only using one.  Instead, we keep track of the using decl and
+    // mark it as touched when something actually uses it.
+    preprocessor_info().FileInfoFor(CurrentFileEntry())->AddUsingDecl(decl);
+
+    if (CanIgnoreCurrentASTNode())  return true;
```
vsapsai: What was the reason to move `if (CanIgnoreCurrentASTNode())  return true;` from before `ReportDeclForwardDeclareUse` after `AddUsingDecl`? Because when I move it back, nothing fails.

iwyu_output.cc
```diff
@@ -475,7 +477,6 @@ string OneIncludeOrForwardDeclareLine::LineNumberString() const {
   return buf;
 }
 
-
```
vsapsai: I tend to overthink empty lines. Why was it there before? Why was it removed? In this case extra empty line might have been used to separate `OneIncludeOrForwardDeclareLine` methods from `IwyuFileInfo` methods.

tests/cxx/using_aliased_symbol_unused.cc
```diff
@@ -0,0 +1,28 @@
...
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Tests that when we include a file that contains a using decl, that we are not
+// forced to include that file if the decl is not used.
+
+#include "using_aliased_symbol-alias.h"
+
+void use_symbol() {
```
vsapsai: `use_symbol` is technically not correct, better is something like `not_use_symbol`.

tests/cxx/using_unused.cc
```diff
@@ -0,0 +1,21 @@
...
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Tests that when we import a name via a using statement, IWYU tells us we 
+// need to include the original thing we're referencing, despite the fact that
+// it's not actually used.
+
+#include <vector>
+using std::vector;
```
vsapsai: Here I think it will be better to use our own class, not STL class. Because for `using` you don't need full use, forward-declare use is sufficient. But we also have a rule not to forward-declare anything from the std namespace [tests/cxx/no_fwd_decl_std.cc](https://github.com/include-what-you-use/include-what-you-use/blob/master/tests/cxx/no_fwd_decl_std.cc). It is better to test these rules separately and not to use `std::vector` in this test case.

And I have discovered a bug that for
```c++
namespace std {
  class vector;
}
using std::vector;
```

IWYU recommends to remove forward declaration that breaks the build. But it's not a regression introduced by your change, master has the same behavior. Need to submit a separate issue for this.

vsapsai:
Amazing job, Chris. It looks really good.

In addition to inline comments I'd like to add that combination of forward declarations and using seems to be undertested. And mentioned bug with removing forward declaration proves it. Also in `ReportDeclForwardDeclareUseWithComment` removing
```c++
if (const UsingShadowDecl* shadow_decl = DynCastFrom(used_decl))
   target_decl = shadow_decl->getTargetDecl();
```
doesn't break anything. And `GetUsingDeclarationOf(target_decl,..` is incorrect but seems to be working fine.

Anyway, I don't think it is worth addressing the combination of forward declarations and using in this pull request. This PR is self-contained and new discovered bugs we should fix separately.

iwyu.cc
cdglove: Well, ReportDeclForwardDeclareUse isn't used at all any more, so it's not really about moving something as much as my choice to place the new code where I did.  My reasoning, and it may be incorrect, was that we're not actually doing any processing as this point, we're only setting up some tables, so we want disregard CanIgnoreCurrentASTNode to ensure the tables are setup correct for files further down the chain. My understanding of CanIgnoreCurrentASTNode is limited though.

iwyu_output.cc
cdglove: No idea why it was there -- also no idea why it was removed.  I probably thought I accidentally added it and so removed it, and then missed it in the diff.  I'll make sure my diff tool isn't set to ignore whitespace. which I have to set it to from time to time. I can add it back.

tests/cxx/using_aliased_symbol_unused.cc
cdglove: Good point.  I'll update it.

tests/cxx/using_unused.cc
cdglove: Yes, I also considered that using things from std is a little heavy, but decided it was good exercise.  For example, the issue with allocator on libc++ might not have been caught otherwise.  That being said, I don't have a strong opinion one way or the other so I am happy to change it.

cdglove:
Thanks Volodymyr.  I'll give the forward declaration stuff some though and perhaps I will look at that next.

kimgr:
Thanks for all the good work on this!

I'll try to pull your latest PR and test on Windows and FreeBSD tomorrow. If that works out, are we ready to merge?

cdglove:
Volodymyr had a couple of minor points to address. They'll take all of 5 minutes to change but I won't be able to get to that until tomorrow night. They will not affect your testing as they are style only (add a blank line back, change a function name in a unit test, possibly change one test to using a user defined struct instead of std::vector).

iwyu.cc
vsapsai: Thanks for explanation. The change is fine with me.

kimgr:
Windows: all tests pass none of the existing ones regress (hard to tell with badinc, but I don't think anything turned out worse)
FreeBSD: using_single_overload fails with the `allocator` diff, also libc++-related

It'd be nice if the new test could succeed on all platforms, so ditching the STL types in favor of simple user-defined types would be a definite improvement.

cdglove:
> It'd be nice if the new test could succeed on all platforms, so ditching
the STL types in favor of simple user-defined types would be a definite
improvement.

Agreed.  I will do that tonight.

vsapsai:
@cdglove, with your latest changes on Mac OS X everything is OK, only the usual badinc.cc failures are still there.

kimgr:
Yay! Same for Windows and FreeBSD.

I instinctively wanted to squash everything into one commit, but I'm not sure it's worth the bother. There's a nice writeup here [1] if you want to take a stab, Chris. I've been able to do this on open pull requests on other projects, so you can probably just re-push after rebasing.

Or I can do a manual pass and do it before pushing, if everyone agrees that it's reasonable to squash.

[1] http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

vsapsai:
Personally I prefer merge --squash. But it is only a matter of personal taste to prefer cleaned up commit history. @kimgr, downside of making merge --squash yourself is that GitHub doesn't understand it and won't show PR as merged. But it matters only in UI, it doesn't affect code itself.

cdglove:
I don't mind either way. I am personally against squashing the history because doing so erases the things that were tried, which I sometimes find to be valuable information.  But, this depends on providing good commit messages for every commit, which can be a pain. Sometimes I think being fast a loose and squashing might be better in the long run, but my primary experience is with Perforce, so I'm used to just expecting every commit to be eternal.  That being said, I am not attached and can easily completely squash that branch.

kimgr:
I thought the history as-is looked a little messy and distracted from the intended change.

But I have to go to bed now, and it's not a priority to me either, so anyone, please merge this however you see fit! :-)

cdglove:
I will prepare a cleaned up version in case we want to go with that.

cdglove:
Pull request now showing one commit.

vsapsai:
Merged the pull request. Thanks, @cdglove for elaborate commit message.

PR #234, opened 2015-11-10
Fix #232: Pick up revision from Git instead of SVN

### LINKED ISSUE #232: IWYU_SVN_REVISION is no longer accurate
Now that we're in Git, the use of SVN in CMake and the revision we attempt to put in `--help` is no longer useful.

We should update this to use the Git hash somehow.

kimgr:
Let me know what you think. Example output:

    $ include-what-you-use.exe --version
    include-what-you-use 0.5 (20c4f5f) based on clang version 3.8.0 (trunk 248605)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,15 +63,15 @@ else()
   message(STATUS "IWYU in-tree configuration")
 endif()
 
-# Pick up SVN revision so we can report it
-# in version information.
-include(FindSubversion)
-if( Subversion_FOUND AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.svn" )
-  Subversion_WC_INFO(${CMAKE_CURRENT_SOURCE_DIR} IWYU)
+# Pick up Git revision so we can report it in version information.
+include(FindGit)
+if( GIT_FOUND )
+  execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse --short HEAD
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    OUTPUT_VARIABLE IWYU_GIT_REV
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
 endif()
-add_definitions(
-  -DIWYU_SVN_REVISION="${IWYU_WC_REVISION}"
-)
+add_definitions(-DIWYU_GIT_REV="${IWYU_GIT_REV}")
 
 add_executable(include-what-you-use
   iwyu.cc
--- a/Makefile
+++ b/Makefile
@@ -35,11 +35,11 @@ ifneq (,$(filter $(HOST_OS), Cygwin MingW))
   LIBS += -lshlwapi
 endif
 
-# Provide SVN revision for version string like in clang/lib/Basic/Makefile.
-IWYU_SVN_REVISION := $(strip \
+# Provide Git revision for version string like in clang/lib/Basic/Makefile.
+IWYU_GIT_REV := $(strip \
         $(shell $(LLVM_SRC_ROOT)/utils/GetSourceVersion $(PROJ_SRC_DIR)))
 
-CPP.Defines += -DIWYU_SVN_REVISION='"$(IWYU_SVN_REVISION)"'
+CPP.Defines += -DIWYU_GIT_REV='"$(IWYU_GIT_REV)"'
 
 check-iwyu:: all
 	./run_iwyu_tests.py
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -103,10 +103,10 @@ static void PrintHelp(const char* extra_msg) {
 
 static void PrintVersion() {
   llvm::outs() << "include-what-you-use " << IWYU_VERSION_STRING;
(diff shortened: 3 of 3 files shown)


Bu11etmagnet:
Maybe a hint would be nice that IWYU is a Git hash fragment (there won't always be letters in the hash), whereas trunk is a svn revision number

vsapsai:
Can you also please update Makefile? Currently it results in
```include-what-you-use 0.5 (24f49a264f9520320968fd107f8192fe6a101ece) based on clang version 3.8.0 (trunk 250659)```

As for using short version you can submit a patch for GetSourceVersion to LLVM or add a file like
```shell
#!/bin/sh

set -eu

usage() {
  echo "usage: $0 <source root>"
  echo "  Prints the short Git commit hash of the given source directory."
  exit 1
}

if [ $# != 1 ] || [ ! -d $1 ]; then
  usage;
fi

cd $1
if [ -d .git ]; then
  git log -1 --pretty=format:%h
else
  exit 1;
fi

exit 0
```

and call it like ```IWYU_SVN_REVISION := $(strip $(shell $(PROJ_SRC_DIR)/GetSourceVersion.sh $(PROJ_SRC_DIR)))```

As for mentioning Git and SVN it is nice idea but we don't have control over what Clang outputs and I don't want to override Clang's version. @Bu11etmagnet, do you have an example how other projects explain it is Git commit hash? Because I don't know what is the common practice. And introducing something new might make things more confusing.

kimgr:
Autoconf/Makefile support in LLVM is about to die, as I understand it, do we really want to keep maintaining it?

I have a feeling I've seen git-marked hashes in LLVM/Clang's version numbering, something like:

    git20c4f5f

I wanted to check out an LLVM Git repo and build something to verify what their format is, but my infrastructure broke down yesterday. I'll check it out and see what they do.

vsapsai:
I agree that supporting Autoconf/Makefile isn't worthy endeavour. But I think we still need to rename IWYU_SVN_REVISION to IWYU_VC_REVISION to avoid breaking the build.

kimgr:
Oh, sorry. I missed the fact that the Makefile already did this. I'll try to build some basic support there as well.

kimgr:
OK, so Clang says this in its --version:

    $ bin/clang++ --version
    clang version 3.8.0 (http://llvm.org/git/clang.git d7177d2007477e52d9a6b1d038da735f408dbe37) (http://llvm.org/git/llvm.git 06941986ad10fe3f3ed5dc19561584e2649d8aa2)
    ...
    $

I might try to mimic that in our version, though I worry that a github.com link might get too long.

Still no luck with the Makefile, I haven't been able to generate a working Make system for IWYU.

kimgr:
I can't get autoconf to pick up either Clang (properly) or IWYU. The big warning sign here:
http://llvm.org/docs/BuildingLLVMWithAutotools.html

leads me to believe this may already be broken. I've tried on FreeBSD and Ubuntu without any luck.

Should I just remove the Makefile and the build instructions instead?

Attached is a preliminary patch to pick up the git revision, if you can find a way to test it.

[gitrev-makefile.txt](https://github.com/include-what-you-use/include-what-you-use/files/34585/gitrev-makefile.txt)

vsapsai:
`$(shell (cd $(PROJ_SRC_DIR) git rev-parse --short HEAD))` doesn't work (IWYU_VC_REVISION is empty).

I suggest just to rename IWYU_SVN_REVISION to IWYU_VC_REVISION, SVN to Git and keep using `$(shell $(LLVM_SRC_ROOT)/utils/GetSourceVersion $(PROJ_SRC_DIR))`.

kimgr:
Thanks! Something like this? I still can't test it, not sure what's going on...

kimgr:
@Bu11etmagnet: prefixed hash with 'git:' -- it's not great, but it's a little clearer.

vsapsai:
Last change doesn't handle correctly the case when there is no version control information, it will output just `[...] (git:) [...]`. The following code worked for me (no space before git is intended)
```Makefile
# Provide Git revision for version string like in clang/lib/Basic/Makefile.
IWYU_VC_REVISION := $(strip \
        $(shell $(LLVM_SRC_ROOT)/utils/GetSourceVersion $(PROJ_SRC_DIR)))
IWYU_VC_REVISION := $(if $(IWYU_VC_REVISION),git:$(IWYU_VC_REVISION))

CPP.Defines += -DIWYU_VC_REVISION='"${IWYU_VC_REVISION}"'
```

kimgr:
Oh, good point. How about moving the `git:` prefixing into `iwyu_globals.cc` instead? That would solve it in one place, even if it feels a little like a layering violation.

vsapsai:
You can move `git:` prefixing to `iwyu_globals.cc`, I have no objections. I've thought about it at first but still wanted to check if it is possible to do purely in Makefile.

kimgr:
OK, here's an updated patch.

I switched the build-system-generated symbol back to being more specific: `IWYU_GIT_REV`, which makes it feel less dirty to prefix the hash with `git:`.

vsapsai:
This one works fine.

PR #235, opened 2015-11-10
Fix #231: Allow spaces in paths in test runner

### LINKED ISSUE #231: Allow spaces in paths when running tests
In issue #215, @dpunset said:

> Oh and by the way I had to modify iwyu_test_util.py to support paths with spaces towards the iwyu
> executable.
>
> At line 406:
> -cmd = '%s %s %s %s' % (
> +cmd = '"%s" %s %s %s' % (

We should probably be passing a list to subprocess.call instead of a space-separated string. If we do, then Python will take care of correct escaping, etc.

kimgr:
This allows the test runner to use IWYU built into a path with spaces.

### DIFF
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -415,7 +415,7 @@ def TestIwyuOnRelativeFile(test_case, cc_file, cpp_files_to_check,
   iwyu_flags = ['-Xiwyu ' + flag for flag in iwyu_flags]
 
   # TODO(csilvers): verify that has exit-status 0.
-  cmd = '%s %s %s %s' % (
+  cmd = '"%s" %s %s %s' % (
       _GetIwyuPath(), ' '.join(iwyu_flags), ' '.join(clang_flags), cc_file)
   if verbose:
     print('>>> Running %s' % cmd)


vsapsai:
Why not passing a list to subprocess call instead of a space-separated string? I've [tried it](https://github.com/vsapsai/include-what-you-use/commit/949cdd37e841133eccae5d815e7bcb75fbe14a5f) and I think it is messier. But I want to hear your reasons.

kimgr:
It broke for two reasons:

1) `iwyu_flags` and `clang_flags` contain joint args, e.g. one list item contains `-Xiwyu --verbose=7`. Python will quote that into a single argument, and IWYU fails to parse it. I worked around it by doing `' '.join(iwyu_flags).split()`, but that seemed to add even more confusion.

2) Even with that fixed, some tests broke. Turned out to be the handling of `--check_also`. The test runner passes the argument as quoted, e.g. `-Xiwyu --check_also="tests/*-d1.h"`. This confused Python's argument handling again, not sure how, but removing the quotes worked.

At that point, there were too many workarounds going on that I thought it would make more sense to just quote the executable if it had spaces.

vsapsai:
Thanks for explanation. Why not just always quote executable? Whitespace check is simple but no check is simpler.

kimgr:
Ahem. Let's just do that instead :-)

vsapsai:
LGTM.

PR #236, opened 2015-11-11
Clean up third-party license listing

kimgr:
The only third-party package we use is an adapted copy of kimgr's
getopt_port, so mention that.

This is a fix for https://github.com/include-what-you-use/include-what-you-use.org/issues/1, filed in the wrong project.


vsapsai:
I am not a lawyer but I assume this change is fine.

PR #237, opened 2015-11-22
Add build target for unit tests

kimgr:
The tests under more_tests/ have been disabled forever.

This mega-hacky patch adds a build target for them --
include-what-you-use-tests -- which creates a working test runner.

There are lots of failing tests, including a bunch I just commented out to
get on with the task.

I think, once I clean this up, that we can remove a bunch of tests already
better covered by the end-to-end suite, and focus on using unit tests for
the various more stateless util-libraries. That way we can get better
coverage and more efficient testing of the small stuff.

Just to be clear, I'm not suggesting we should merge this as-is, I have
some ideas for cleaning up CMakeLists.txt, but it demonstrates that this
can be done. I thought I'd push followup commits as I find some spare time
to work on it.

Let me know what you think!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,40 +73,10 @@ if( GIT_FOUND )
 endif()
 add_definitions(-DIWYU_GIT_REV="${IWYU_GIT_REV}")
 
-add_executable(include-what-you-use
-  iwyu.cc
-  iwyu_ast_util.cc
-  iwyu_cache.cc
-  iwyu_driver.cc
-  iwyu_getopt.cc
-  iwyu_globals.cc
-  iwyu_include_picker.cc
-  iwyu_lexer_utils.cc
-  iwyu_location_util.cc
-  iwyu_output.cc
-  iwyu_path_util.cc
-  iwyu_preprocessor.cc
-  iwyu_verrs.cc
-)
+include_directories(${LLVM_MAIN_SRC_DIR}/utils/unittest/googletest/include)
 
-if( MSVC )
-  # Disable warnings for IWYU, and disable exceptions in MSVC's STL.
-  add_definitions(
-    -wd4722 # Suppress ''destructor'' : destructor never returns, potential memory leak
-    -D_HAS_EXCEPTIONS=0
-  )
-
-  # Put project in solution folder
-  set_target_properties(include-what-you-use
-    PROPERTIES FOLDER "Clang executables"
-  )
-else()
-  # Disable RTTI, use C++11 to be compatible with LLVM/Clang libraries.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti -std=c++11")
-endif()
-
-# Clang dependencies.
-target_link_libraries(include-what-you-use
+list(APPEND IWYU_LINK_LIBS
+  # Clang dependencies
   clangFrontend
   clangSerialization
   clangDriver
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -161,11 +128,66 @@ else()
...
+    -D_HAS_EXCEPTIONS=0
+  )
+
+  # Put project in solution folder
+  set_target_properties(include-what-you-use
+    PROPERTIES FOLDER "Clang executables"
+  )
+else()
+  # Disable RTTI, use C++11 to be compatible with LLVM/Clang libraries.
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti -std=c++11")
```
vsapsai: How can we set `-fno-rtti` only for include-what-you-use? Because it is used by GoogleTest and fails with
```
llvm/utils/unittest/googletest/include/gtest/internal/gtest-port.h:1041:16: error: cannot use typeid with -fno-rtti
  GTEST_CHECK_(typeid(*base) == typeid(Derived))
```

vsapsai:
Thanks for doing this, Kim. I am not CMake expert, your changes look fine to me. The only issue is `-fno-rtti`.

Also I have to add running unit tests at the buildbot and it is out-of-source build. I think defining `LLVM_MAIN_SRC_DIR` should be enough. But if you have any ideas regarding out-of-source build, I'll be glad to hear them.

kimgr:
> How can we set -fno-rtti only for include-what-you-use?
> Because it is used by GoogleTest and fails [...]

LLVM has this:

    list(APPEND LLVM_COMPILE_DEFINITIONS GTEST_HAS_RTTI=0)

so I'm guessing we can turn off RTTI in GTest rather than trying to enable it in the build. I'll try that.

vsapsai:
After changing in CMakeLists.txt
```
  # Disable RTTI, use C++11 to be compatible with LLVM/Clang libraries.
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti -DGTEST_HAS_RTTI=0 -std=c++11")
```
I was able to build `include-what-you-use-tests`. It has 2 warnings
```
llvm/utils/unittest/googletest/include/gtest/gtest.h:1320:16: warning: comparison of integers of different signs: 'const int' and 'const unsigned long' [-Wsign-compare]
```

When I run `include-what-you-use-tests`, there are a few failed tests and the test itself crashes. So we'll need to investigate and fix this. Also it would be nice to have a target to build-and-run tests instead of a target to build and manually run tests.

As for commented out tests, it might be better [to disable them](http://stackoverflow.com/questions/7208070/googletest-how-to-skip-a-test). I haven't checked what is the problem, if it is compilation error, disabling won't help.

kimgr:
Thanks for testing!

> As for commented out tests, it might be better to disable them. I haven't checked what is the problem, if it is compilation error, disabling won't help.

Right, the major challenge was getting it to compile.

I've now successfully ported `iwyu_lexer_utils_test.cc` and `iwyu_string_util_test.cc`, they compile and the tests pass.

The other two, are more challenging, however:

- `iwyu_output_test.cc` doesn't compile. I don't see a lot of value in these tests as they are so heavily dependent on Clang types (`Decl` and `SourceLocation`, notably) which are hard to fake. There are lots of hacks in `iwyu_output.cc` and the test file itself, but they seem to mostly add complexity for very little benefit. I vote to discard these tests and then start from scratch trying to get `OutputLine` under test.
- `iwyu_include_picker_test.cc` compiles, but there are several test failures. Many of them are due to Google-internal mappings. I'm inclined to remove the failing tests and just go with what happens to work.

If we can agree on that strategy, I can update the PR pretty soon with a more staged conversion,  better support for using system GTest and some other goodies.

vsapsai:
Need more time with failed tests in `iwyu_output_test.cc` and `iwyu_include_picker_test.cc`. Cannot tell now why PrintableDiffsTest.ProperlyOrdersIncludeLines, PrintableDiffsTest.ShowLineNumbersForDeletedIncludesEvenWithUses crash test executable. At first glance these tests seem to be reasonable and useful.

kimgr:
Sorry, that's my fault. We've changed `PrintableDiffs` to also take an `IwyuPreprocessorInfo*`, and I just passed `NULL` in the tests to get it to build.

I'll update the diff with what I have so far, but I'm at a stage where I can't get the output and include_picker tests to build, so I've just excluded them from the cmake configuration.

PR #238, opened 2015-11-22
Fix #210: Avoid mis-stripping of same path prefix

### LINKED ISSUE #210: Problem with stripped path
Hello, long time no see :)

I just encountered a problem with the function ConvertToQuotedInclude. I'll explain.

I noticed that some of the includes suggested by IWYU had a badly formed path. The name of the folder where the header exists was incomplete. I debugged it and here is what I found.

I have a filepath that reads like this as a parameter for this function:

R:/folderA/folderB/folderC/EngineServices/folderD/handle.h

The search_paths vector contains many entries to windows sdk, external middleware, and some of our own projects. One of the projects is this:

R:/folderA/folderB/folderC/Engine

When we iterate through them doing this:

for (Each<HeaderSearchPath> it(&search_paths); !it.AtEnd(); ++it) {
    if (StripLeft(&path, it->path)) {

my path is converted to this:

Services/folderD/handle.h

which is a bad path, this file doesn't exist.

A possible fix would be to apply the strip with the / at the end. If there is no /, we don't strip.

What do you think?

kimgr:
If two paths in the header search set shared the same prefix, IWYU would
sometimes pick the wrong one and produce an invalid quoted include. For
example:

Header search path: /a/foo, /a/foobar
File path: /a/foobar/header.h

This would yield "bar/header.h", as "/a/foo" was considered a basename of
the file path.

We fix this by ensuring all search path entries end with "/".

I didn't add a test for this, because an end-to-end test would be almost
nonsensical. Once we get the unit test suite up and running, I'd like to
add a test for this there instead.

### DIFF
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -297,7 +297,7 @@ static vector<HeaderSearchPath> ComputeHeaderSearchPaths(
            it = header_search->system_dir_begin();
        it != header_search->system_dir_end(); ++it) {
     if (const DirectoryEntry* entry = it->getDir()) {
-      const string path = CanonicalizeFilePath(entry->getName());
+      const string path = CanonicalizeHeaderSearchPath(entry->getName());
       search_path_map[path] = HeaderSearchPath::kSystemPath;
     }
   }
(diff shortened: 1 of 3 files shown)


vsapsai:
I confirm there are no regressions and my repro case for issue #210 is fixed.

What about adding trailing slash in `CanonicalizeFilePath`? Other options are `ComputeHeaderSearchPaths` and `NormalizeHeaderSearchPaths` but I prefer `CanonicalizeFilePath` because name corresponds to your goal.

kimgr:
That would be nice, but `CanonicalizeFilePath` doesn't know if it's working on a file path or a dir path. I could add a `CanonicalizeHeaderSearchPath` to do this if you prefer?

kimgr:
Here, changed to use new `CanonicalizeHeaderSearchPath` up-front. It simplifies `ConvertToQuotedInclude`, and I think I like this better, even if the patch is slightly larger.

vsapsai:
This works too and looks good to me.

kimgr:
Here we go, thanks! I'd like to nominate this for IWYU 0.5 as well.

PR #248, opened 2015-12-24
Update to reflect getOptionalExplicitTemplateArgs changes in Clang.

vsapsai:
Clang r256359 removed `getOptionalExplicitTemplateArgs` on various Expr
subclasses. Use
  `void copyTemplateArgumentsInto(TemplateArgumentListInfo &List) const`
instead.

`const ASTTemplateKWAndArgsInfo *getTemplateKWAndArgsInfo() const` doesn't look
like appropriate replacement because it checks only `HasTemplateKWAndArgsInfo`
and we'll need to do `hasExplicitTemplateArgs()` checks ourselves.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -45,7 +45,6 @@ namespace clang {
 class FileEntry;
 }  // namespace clang
 
-using clang::ASTTemplateArgumentListInfo;
 using clang::BlockPointerType;
 using clang::CXXConstructExpr;
 using clang::CXXConstructorDecl;
@@ -105,6 +104,7 @@ using clang::TagDecl;
 using clang::TagType;
 using clang::TemplateArgument;
 using clang::TemplateArgumentList;
+using clang::TemplateArgumentListInfo;
 using clang::TemplateArgumentLoc;
 using clang::TemplateDecl;
 using clang::TemplateName;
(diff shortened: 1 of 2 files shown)


vsapsai:
I am going to merge this change to fix the build. Any comments post-commit are welcome.

iwyu_ast_util.cc
```diff
@@ -711,17 +711,14 @@ static map<const Type*, const Type*> GetTplTypeResugarMapForFunctionNoCallExpr(
...
-  if (explicit_tpl_list) {
-    for (unsigned i = 0; i < explicit_tpl_list->NumTemplateArgs; ++i) {
-      const TemplateArgument& arg
-          = explicit_tpl_list->getTemplateArgs()[i].getArgument();
-      if (const Type* arg_type = GetTemplateArgAsType(arg)) {
-        retval[GetCanonicalType(arg_type)] = arg_type;
-        VERRS(6) << "Adding an explicit template-function type of interest: "
-                 << PrintableType(arg_type) << "\n";
-      }
+  for (unsigned i = 0; i < explicit_tpl_list.size(); ++i) {
```
kimgr: You should be able to convert this to a range loop like this:

    for (const TemplateArgumentLoc& loc : explicit_tpl_list.arguments()) {
        if (const Type* arg = GetTemplateArgAsType(loc.getArgument()) {
            ...
        }
    }

kimgr:
Looks good, thanks!

iwyu_ast_util.cc
vsapsai: Good suggestion, it simplifies the code a little bit. Pushed corresponding change.

PR #249, opened 2015-12-26
Delegate to CMake the usage of -ldl or not

sylvestre:


vsapsai:
This change doesn't break anything for me. And without testing it on Debian GNU/kFreeBSD, it looks fine. @kimgr supports IWYU on FreeBSD so I will rely on his opinion here.

kimgr:
I'm out traveling for a few more days, but I remember the Linux condition was added to get the FreeBSD build working. Stock freebsd didn't have/need dl at that point, maybe this changed or kFreeBSD is somehow different?

kimgr:
'm out traveling for a few more days, but I remember the Linux condition was added to get the FreeBSD build working. Stock freebsd didn't have/need dl at that point, maybe this changed or kFreeBSD is somehow different?

sylvestre:
Yes, kFreeBSD needs ld. See the log before we introduced the patch:
https://buildd.debian.org/status/fetch.php?pkg=iwyu&arch=kfreebsd-amd64&ver=3.6-1&stamp=1440159799

kimgr:
The build fails on stock FreeBSD 10.2 with your patch:

    FAILED: : && /usr/bin/CC   -B/home/kimgr/bin/gold -fPIC -fvisibility-inlines-hidden -Wall -W
    -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic
    -Wno-long-long -Wcovered-switch-default -Wdelete-non-virtual-dtor -std=c++11
    -fcolor-diagnostics -fno-common -Woverloaded-virtual -Wno-nested-anon-types -fno-rtti
    -std=c++11 -g  -Wl,-z,origin -Wl,-allow-shlib-undefined 
    ...
    /home/kimgr/bin/gold/ld: error: cannot find -ldl

It would be nice if we could test for the presence/necessity of `dl` instead of poking around with system names, but I don't know enough CMake to know if that's possible.

kimgr:
It appears the trick is to use `CMAKE_DL_LIBS` unconditionally, and it expands as necessary to `-ldl` or nothing depending on necessity:
https://cmake.org/cmake/help/v3.0/variable/CMAKE_DL_LIBS.html

Could you try that and update the PR? I'm a little pressed for time today.

sylvestre:
Much better indeed, thanks!

vsapsai:
Current diff has one extra `endif()`. And I was thinking about adding dl to the rest of dependencies, like
```
  # POSIX
  include(FindCurses)
  target_link_libraries(include-what-you-use
    pthread
    z
    ${CURSES_LIBRARIES}
    # Add dl for dynamic linking when needed.
    ${CMAKE_DL_LIBS}
  )
```

`${CMAKE_DL_LIBS}` works on Mac OS X, haven't tested Ubuntu yet.

sylvestre:
${CMAKE_DL_LIBS} tested on Debian without any issue. So, should work on Ubuntu without any issue

kimgr:
FreeBSD works too. Thanks for the patch, much appreciated.

I'll do a follow-up commit to get rid of all the ``CMAKE_SYSTEM_NAME MATCHES`` constructs -- we only added those to be consistent with the check for Linux needed for ``dl`` (apparently CMake does not expose a simple flag for Linux specifically.)

PR #250, opened 2015-12-27
Mappings for qt5.4

ThosRTanner:
This is the mappings for QT5 based on scanning the header directory and writing a whole list of mappings.

It's not ideal, but it mostly works. See issues raised about qt deciding it needs forward references though.


vsapsai:
At the first glance looks reasonable, thanks for working on this @ThosRTanner. It would be great if somebody else using Qt 5.4 tested this, just to get the second pair of eyes.

Is this mapping applicable only to Qt 5.4 or for all Qt 5.x versions?

Current Qt 4 mapping maps multiple files to `<QtCore>`, `<QtGui>`, `<QtNetwork>`. But this mapping , for example, maps to `<QtNetwork>` only symbol `QtNetwork`. Is this intended?

ThosRTanner:
Yes, this is intentional. I went from the the API documentation. For instance, although `<QDate>`, `<QDateTime>` and `<QTime>` are all just `#include "qqdatetime.h"`, the documentation from the web page says for the QDate class, you should include <QDate> (etc).

Which is why it is so big.

I suspect this is applicable to all Qt5 versions but I've only got QT5.4 installed. I haven't worked up the courage to install QT5.5 yet.

kimgr:
Thanks for working on this! I think you're the local Qt expert, so if this works for you then it's probably good.

I ran into this when reading through the file (line 1591):

> ## other things not picked up by the above
>    #{ symbol: [ "qobject_cast", "private", "<QObject>", "public" ] },
>    #{ symbol: [ "qApp", "private", "<QApplication>", "public" ] },
>    #{ symbol: [ "qHash", "private", "<QHash>", "public" ] },

They're all commented out -- if they're not necessary, can you remove them instead?

ThosRTanner:
Ooops. Removed.

kimgr:
This looks good to me. I think it's fine to start out with it named qt5.4, and we can revise the name if we get it proven to work with other 5.x versions. OK with you, @vsapsai?

vsapsai:
I am OK with this change. Any problems we can fix when we learn about them and can rename the file when we know it works for other 5.x versions or if it is not discoverable for other people.

kimgr:
Just pushed this as a collapsed commit, thanks!

PR #252, opened 2015-12-30
Remove use of CMAKE_SYSTEM_NAME

kimgr:
We only used it to detect Linux earlier. Using the Boolean flags makes
things much clearer.

Also, don't assume POSIX, but rather warn if it's a platform we don't
know anything about.

(side note: I wonder if ``APPLE``/Darwin is really the right thing to check for wrt ``-fvisibility-hidden``. Maybe that's what Clang does, though, and we need to follow suit?)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
     )
   endif()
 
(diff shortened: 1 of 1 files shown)


vsapsai:
With your change everything works fine.

I haven't found where in Mac build `-fvisibility-hidden` is set but I've tried with Clang 3.7.0 binaries and without it you have a bunch of linker warnings. Haven't seen those warnings on Ubuntu.

kimgr:
Thank you, pushed this just now.

PR #253, opened 2015-12-30
Add better build target for unit tests

kimgr:
Supersedes PR #237.
- Put IWYU sources in an OBJECT library
- Build an intermediate list of target link deps
- Add build target for unit tests
- Port iwyu_string_util_test.cc (tests succeed)
- Port iwyu_lexer_utils_test.cc (tests succeed)
- Port iwyu_include_picker_test.cc (tests fail)

I haven't managed to port iwyu_output.cc here yet, it's quite a lot of work due to code changes over the years.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,7 +110,7 @@ else()
 endif()
 
 # Clang dependencies.
-target_link_libraries(include-what-you-use
+list(APPEND IWYU_LINK_LIBS
   clangFrontend
   clangSerialization
   clangDriver
@@ -124,7 +124,7 @@ target_link_libraries(include-what-you-use
 )
 
 # LLVM dependencies.
-target_link_libraries(include-what-you-use
+list(APPEND IWYU_LINK_LIBS
   LLVMX86AsmParser # MC, MCParser, Support, X86CodeGen, X86Desc, X86Info
   LLVMX86CodeGen # Analysis, AsmPrinter, CodeGen, Core, MC, Support, Target, 
                  # X86AsmPrinter, X86Desc, X86Info, X86Utils
@@ -151,12 +151,12 @@ target_link_libraries(include-what-you-use
 
 # Platform dependencies.
 if( WIN32 )
-  target_link_libraries(include-what-you-use
+  list(APPEND IWYU_LINK_LIBS
     shlwapi
   )
 elseif( UNIX )
   include(FindCurses)
-  target_link_libraries(include-what-you-use
+  list(APPEND IWYU_LINK_LIBS
     pthread
     z
     ${CURSES_LIBRARIES}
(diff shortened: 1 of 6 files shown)


vsapsai:
Thanks, this works better. The only "issue" is inconsistent spaces around parenthesis. For example, `if ( DEFINED IWYU_GTEST_LIB_PATH)`.

kimgr:
Fixed the `if` spacing.

vsapsai:
Everything looks fine. Will update build bot to handle unit tests. I expect the build to fail due to missing gtest library and headers.

And I've noticed [LLVM has Windows-specific code in unit test driver](http://llvm.org/viewvc/llvm-project/llvm/trunk/utils/unittest/UnitTestMain/TestMain.cpp?view=markup). You might want to check that out.

kimgr:
Thanks, good tip. Windows is wired by default to show message boxes for critical errors, so this kind of stuff goes into every test runner. I would've expected gtest to take care of it, but apparently not. I'll update the driver.

Just so we're clear: the include_picker tests are still failing, so maybe we should disable them before setting up the buildbot?

vsapsai:
It is OK that include_picker test is failing. I am not adding the step to run unit tests. It's just the build is out-of-tree and IWYU_GTEST_INCLUDE_PATH, IWYU_GTEST_LIB_PATH are not defined. I plan to fix the build if running `ninja` fails.

kimgr:
Ah, I see what you're saying now, sorry about the confusion.

I pushed critical error suppression for _WIN32.

vsapsai:
Still works fine on Mac OS X.

vsapsai:
Checked the buildbot and adding GTest doesn't seem to be hard. For IWYU_GTEST_INCLUDE_PATH I can use `/mnt/buildbot_iwyu_trunk/workspace/sources/llvm/utils/unittest/googletest/include/` and for IWYU_GTEST_LIB_PATH `/mnt/buildbot_iwyu_trunk/workspace/build_llvm/lib/`. `workspace/build_llvm/lib` contains LLVM and Clang libraries like `workspace/installed` but I don't think it will be a problem, especially that IWYU itself isn't using libraries from IWYU_GTEST_LIB_PATH.

For the record, trying to build GTest separately with

`cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=YES /mnt/buildbot_iwyu_trunk/workspace/sources/llvm/utils/unittest/`

fails with

`CMake Error: The source directory "/mnt/buildbot_iwyu_trunk/workspace/sources/llvm/utils/unittest" does not appear to contain CMakeLists.txt.`

and
```
ls /mnt/buildbot_iwyu_trunk/workspace/sources/llvm/utils/unittest
CMakeLists.txt  LLVMBuild.txt  Makefile  UnitTestMain  googletest
```
I decided not to spend time trying to figure this out.

kimgr:
Abandoning this for now, it's been dormant for two years. We'll start over if we need this at some later date.

PR #261, opened 2016-02-01
Remove Makefile as it's not supported by LLVM+Clang.

vsapsai:
Autoconf support was removed in LLVM and Clang in r258861 and r258862
respectively.

Technically, I should branch `clang_3.8` before committing this change because 3.8.0 still supports autoconf. But I think we don't need to wait and can drop Makefile right now.

I think I've updated all places where we reference autoconf build but I am surprised how little changes were required. So I suspect I've missed something.


kimgr:
Oops, this slipped off my radar. LGTM!

Our autoconf build has always been treated a little step-motherly so I think it's fine to ship the next IWYU without it even if Clang technically supports it for 3.8.

vsapsai:
Thanks for review. Merged manually.

PR #263, opened 2016-02-07
Fix 262: Don't fail the build if checked out via SVN

kimgr:
Based on @JVApen's patch, this should allow, but discourage, SVN checkouts.


vsapsai:
Works for me.

kimgr:
Merged manually, thanks!

PR #264, opened 2016-02-09
Silence CMake CMP0048 warning

kimgr:
Whenever I run `cmake` I see the following warning:

    CMake Warning (dev) at tools/clang/tools/include-what-you-use/CMakeLists.txt:3 (project):
      Policy CMP0048 is not set: project() command manages VERSION variables.
      Run "cmake --help-policy CMP0048" for policy details.  Use the cmake_policy
      command to set the policy and suppress this warning.

      The following variable(s) would be set to empty:

        PROJECT_VERSION
        PROJECT_VERSION_MAJOR
        PROJECT_VERSION_MINOR
        PROJECT_VERSION_PATCH
    This warning is for project developers.  Use -Wno-dev to suppress it.

It doesn't look particularly useful to us, so I just picked up whatever LLVM's base `CMakeLists.txt` was doing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,10 @@
 cmake_minimum_required(VERSION 2.8)
 
+if( POLICY CMP0048 )
+  # Silence CMP0048 warning about missing project VERSION.
+  cmake_policy(SET CMP0048 NEW)
+endif()
+
 project(include-what-you-use)
 
 if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )


vsapsai:
Shouldn't it be `cmake_policy(SET CMP0048 OLD)`? Because LLVM specifies version in `project()` command and we don't. According to [CMP0048 docs](https://cmake.org/cmake/help/v3.0/policy/CMP0048.html) seems like no version in `project()` corresponds to OLD behavior.

kimgr:
I've now read both CMP0048 and the docs for `project` more thoroughly, and I have no idea what they mean anymore :-)

I was afraid IWYU might clobber LLVM's version variables, but I think they're per-project, not global.

We don't want IWYU to either clobber or silently pick up LLVM's version information, but it doesn't look like that can ever happen, can it?

Unless there's a functional difference, one reason to pick NEW is that it should eventually become the default supported option, and we can remove it.

vsapsai:
With this reasoning I agree to pick NEW. We don't use CMake VERSION variables anyway.

PR #265, opened 2016-02-09
Fix #230: Friendlier mapping error handling

### LINKED ISSUE #230: Could IWYU be more user friendly if you've made a mess of your mapping file
For instance if I have a line that says

    { include: [ "header", "private", "another header", "public"] }

I get an assertion error and I have no idea which part of the file it is objecting to.

kimgr:
Malformed mapping files were previously diagnosed with assertions. Now we
emit nice diagnostics using the YAML infrastructure instead.

These checks duplicate the assertions, which are more geared towards
programmatically added mappings.

### DIFF
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -1441,6 +1441,14 @@ void IncludePicker::AddMappingsFromFile(const string& filename,
           return;
         }
 
+        if (!IsQuotedInclude(mapping[2])) {
+          json_stream.printError(
+              current_node,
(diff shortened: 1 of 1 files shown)


kimgr:
Ping @ThosRTanner, try this and see if it makes more sense.

vsapsai:
Looks good to me. For reference, added diagnostic corresponds to assertions [here](https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_include_picker.cc#L1044-L1050).

kimgr:
Thanks, I added a diagnostic for `symbol` mappings too in the same style.

Issue #75, opened 2015-06-15
fix_includes.py --comments should add comments even if the actual includes didn't change


vsapsai:
Originally reported on Google Code with ID 75

```
What steps will reproduce the problem?  Give the *exact* arguments passed
to include-what-you-use, and attach the input source file that gives the
problem (minimal test-cases are much appreciated!)
1. Run iwyu on something, and update the file to match its recommendations.
2. Later on, run iwyu again, and feed the output to fix_includes.py --comments to add
comments.
3. Nothing happens!  "(skipping /path/to/foo.cpp: iwyu reports no contentful changes)".
 I don't want only "contentful" changes -- that's why I ran it with --comments!

Looking at the output of iwyu itself, it currently doesn't output anything for files
it's already happy with.  It should change to output the contents again with comments
if they don't already have comments.


What version of the product are you using? On what operating system?
LLVM r160007, iwyu r357, Ubuntu 12.04 LTS x86.
```

Reported by `ayg@aryeh.name` on 2012-07-11 12:42:58

PR #266, opened 2016-02-11
Add --no_comments arg to suppress why-comments

kimgr:
We've run IWYU continuously on a large code base for about a year, and applied its changes without `fix_includes.py` (developers apply changes manually and mindfully.)

We see the `// for` comments rotting pretty fast and the relatively usesless `// lines` comments being pasted into source code. This now creates new maintenance overhead for us, so I'd like to be able to disable the why-comments.

### DIFF
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -90,7 +90,8 @@ static void PrintHelp(const char* extra_msg) {
          "   --max_line_length: maximum line length for includes.\n"
          "        Note that this only affects comments and alignment thereof,\n"
          "        the maximum line length can still be exceeded with long\n"
-         "        file names (default: 80)."
+         "        file names (default: 80).\n"
+         "   --no_comments: do not add 'why' comments.\n"
          "   --verbose=<level>: the higher the level, the more output.\n"
          "\n"
          "In addition to IWYU-specific options you can specify the following\n"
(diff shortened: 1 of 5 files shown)


vsapsai:
Can you add a test for this, please? And I don't understand how `// lines` comments end up in code because these lines are supposed to be removed. And I though that without `// for` comments it is still useful to know which line to remove.

For reference, related issue is [#75 fix_includes.py --comments should add comments even if the actual includes didn't change](https://github.com/include-what-you-use/include-what-you-use/issues/75).

kimgr:
Thanks, I'll add a test -- it turns out to be straightforward, I figured it would be difficult.

We see `// lines` on forward decls, for some reason. That might be a bug in IWYU.

I think issue #75 is asking for the exact opposite of this patch -- always add comments, even if IWYU thinks the file is fine. My patch disables comments entirely.

kimgr:
Here's an update with test. Let me know what you think.

Flamefire:
I don't understand the need for this PR. fix_includes.py already has this option...

kimgr:
We don't use fix_includes.py, we run analysis off on a server, and then developers apply changes manually, fixing up obvious bugs manually.

Flamefire:
Ok, I misread your initial PR and confused "without" and "with"

vsapsai:
I didn't expect `// lines` on forward declares but this is a separate issue.

Everything is fine except of CRLF line endings in `tests/cxx/no_comments.cc`. I pulled your change with
```
git checkout -b kimgr-nocomments master
git pull https://github.com/kimgr/include-what-you-use.git nocomments
```

and on Mac OS X test failed. I expect it to fail on the buildbot too. After fixing line endings the test is passing.

kimgr:
Thanks, pushed a rebased atomic commit just now.

PR #274, opened 2016-02-23
Remove test dependency on <stdlib.h>

kimgr:
cc: @daveisfera, @vsapsai

Here's a stab at removing `rand` and `<stdlib.h>` from badinc. I get basically the same result on FreeBSD (a little less failure noise). I don't have a Linux box to try this on, but it should pretty much just  simplify matters.

### DIFF
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -1066,7 +1066,7 @@ class CC_TemplateClass {
   typedef I1_TemplateClass<A> i1_typedef;
 
   // Let's throw in per-class operator new/delete.
-  // IWYU: size_t is...*((<stddef.h>)|(stdio.h)|(stdlib.h)|(string.h)|(time.h)|(wchar.h))
+  // IWYU: size_t is...*((<stddef.h>)|(stdio.h)|(string.h)|(time.h)|(wchar.h))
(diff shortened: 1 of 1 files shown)


daveisfera:
Fedora Rawhide (24) updated to LLVM/Clang 3.8 a few days ago, so I can't test this patch with the 0.5 build that I have right now. Does trunk build against 3.8?

kimgr:
I think so, I think we got all of the compat changes out of the way before 3.8 branched.

daveisfera:
I built with a [recent commit](https://github.com/include-what-you-use/include-what-you-use/commit/e1fc7722673fd665c105d0aaed1dab1aeefce8c1) and all the tests passed, so it looks good to me.

kimgr:
Wait, what? You tested with this PR applied, right?

daveisfera:
Sorry for the confusion. I should have said "I built with this fix applied on top of a recent commit with LLVM/clang  3.8 and gcc 6.0 and all tests passed".

kimgr:
OK, I figured, thanks! :-)

vsapsai:
Looks good to me and tests are passing.

kimgr:
... and done, I just merged and pushed this manually. Thanks for confirming!

Issue #272, opened 2016-02-20
Don't include files included by corresponding header


Flamefire:
If you have a `foo.cpp` with `foo.h` where `foo.h` already includes `bar.h` and of course `foo.cpp` includes `foo.h` first then it should not be required to also include `bar.h`

kimgr:
This should already be happening. We need a small repro to be able to act on this.

Flamefire:
Before I start searching for one: Is it possible that IWYU gets confused, if the "corresponding" header is not the first one included? I got a common "defines.h" that (by convention) must be included first in every cpp file (similar probably to the stdafx.h on MSVC) so the "corresponding" header will be the 2nd one.

cdglove:
> This should already be happening. 

Not always I don't think. This really depends on what the foos and bars look like and how one feels about transitive includes. It's often not a good practise to depend on header including other headers for you, so we really need to see what these foos and bars look like.

kimgr:
>  Is it possible that IWYU gets confused, if the "corresponding" header is not the first one included?

Yes, the asssociated header must be the first included, and match the name of the .cpp file.

>  I got a common "defines.h" that (by convention) must be included first in every cpp file

If it's really included in *every* cpp file, you could probably use the `--pch_in_code` switch to ignore it (we added that to cope with `stdafx.h`).

> Not always I don't think. This really depends on what the foos and bars look like [...]

It only happens for so-called associated headers, i.e. the A.h for A.cpp. The theory is that A.cpp will always include A.h, and they belong together, so there's no point in duplicating includes between them.

Flamefire:
Well it is a rather large project with a couple of subprojects. One of them is even an external C project (library) so no it is no in ***ALL*** cpp files. In one of the projects it is `defines.h` in another it is `main.h` and in the C project there is no such thing.

Wouldn't it make sense, to extent the `pch_in_code` option so that it checks, if there is an additional header before the "corresponding" header and so detect the "corresponding" header also on 2nd position?

kimgr:
OK, then you pass `--pch_in_code` to the projects that have it.

That's basically what the pch-in-code logic does, but there's no conditional evaluation -- if you've specified pch-in-code, the first header is always marked as a PCH (and evaluated according to `--prefix_header_includes`). If not, the first header is assumed to be the associated header.

Flamefire:
[...]

Issue #271, opened 2016-02-20
Same directory not detected as include paths


Flamefire:
I have a file which misses an include to a header in the same directory. iwyu suggests including that header by its full path but the relative path (just the header name) is enough.
I'd consider this a major defect, as absolute paths to project headers will most likely break building on other systems.

Note: I'm using the iwyu_tool.py with a CMake build database on an in-source build (subfolder however)

Flamefire:
This also causes an interesting side-effect as described in #272: If you have a header in the same directory as a cpp file, that is included by the associated header, this header is suggested to add, also it should not. This goes away, when explicitly specifying the cpps directory via `-I`

Testcase: https://github.com/include-what-you-use/include-what-you-use/files/142703/incTest.zip

- running iwyu from inside the directory of the cpp results in correct includes :+1:
- using the absolute path of the cpp suggests adding the header :-1:
- running iwyu from the parent folder and specifing a relative path to the cpp suggests adding the header :-1:
- adding the path to the cpp via `-I` (either absolute or relative) results in correct includes :+1:

kimgr:
Thank you for this rundown! I wonder if this is the same issue as #5 (our oldest bug), they definitely look similar.

I don't have much time to look at this until next week, but I'd love to dig deeper here, and try to find a solution.

Flamefire:
Yes this seems to be a part of #5.
My idea would be, that includes are stored as absolute paths in the iwyu database so the same include even when spelled differently is always recognized. This would fix the issue in #272 
On picking an `#include` directive there should be a choice made, how to abbreviate the path. First impulse would be to go through search paths (converted to absolute paths) and then choose the shortest name. Otherwise one can choose the next-best spelling already found. So if the include appears in a header in the same directory, the same spelling is chosen. If not, pick just any spelling found if it is valid according to the search paths.
I'd even make this an option `--suggest-shortest-paths`

kimgr:
> First impulse would be to go through search paths (converted to absolute paths) and
> then choose the shortest name

See https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_path_util.cc#L181 -- there is logic like this. But it sounds like it's the reverse of what you're suggesting. Can you experiment with variations on this algorithm and see if you can figure out something that would work for you (and the existing test cases)?

Flamefire:
[...]

Issue #162, opened 2015-06-15
Pragma export not working for inline files


vsapsai:
Originally reported on Google Code with ID 162

```
What steps will reproduce the problem?  Give the *exact* arguments passed
to include-what-you-use, and attach the input source file that gives the
problem (minimal test-cases are much appreciated!)

I'm working on a smaller test case, but for now the error comes up here: 

0. Built IWYU in-tree
1. clone https://github.com/aherlihy/mongo-cxx-driver legacy branch.
2. add "// IWYU pragma: export" to src/mongo/base/status.h:150 (at https://github.com/aherlihy/mongo-cxx-driver/blob/iwyu-PNT/src/mongo/base/status.h#L150)
2. Run IWYU:
          a. Need to write little wrappers around iwyu because scons requires successful
exit status. I've attached my two wrapper scripts (they'll need to be changed to match
the path to iwyu).
          b. Run "scons --cc=iwyu.cc --cxx=iwyu.cxx build-base/parse_number_test"
4. Apply IWYU changes to code base: python fix_includes.py < iwyu_cxx.out

What is the expected output? What do you see instead?

We expect status.h to be included instead of status-inl.h. Instead, in places where
status.h is used, status-inl.h is added to the include list. One example of many is
in mongo/src/base/parse_number_test.cpp.


What version of the product are you using? On what operating system?

The most recent pull from svn: I ran svn co http://include-what-you-use.googlecode.com/svn/trunk/
include-what-you-use on Mac OSX 10.9.5 on 2.6GHz on Intel Core i5.


Please provide any additional information below.

I will try my best to recreate the issue with smaller test files, but so far I've been
unsuccessful. I will let you know as soon as I do. Thank you!


```

Reported by `anna.herlihy@10gen.com` on 2014-11-07 22:07:10

<hr>
- _Attachment: [iwyu.cxx](https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-162/comment-0/iwyu.cxx)_
- _Attachment: [iwyu.cc](https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-162/comment-0/iwyu.cc)_

kimgr:
This is still an issue after PR #276. When handling `pragma: export`, we don't have access to the includer of the includer (which would be the right thing to pass to `ConvertToQuotedInclude`). I think we could actually produce that information, and now that `ConvertToQuotedInclude` accepts it, this could be made to work.

PR #276, opened 2016-02-24
Also add include in same directory

Flamefire:
This fixes a bug where absolute paths are unnecessarily used for a header resulting even in inclusion of unwanted headers as described in #272 

The problem arises, as the directory of the file, which is implicitely added to the search path, is not honored. See #271 for this.

This supersedes the patch(es) in #5 and should also close those issues. This solution is better, than adding the main-files directory to the search path, as this does not add it globally which can cause problems if headers are checked, which might not be in the same directory as the main-file.

- [x] All tests run successfully
- [x] Fix verified by testing the test-case from #271 manually
- [x] Remove `IndirectSubfolderClass` from direct_subfolder.h
- [x] Rebase and remove `DEVTOOLS_MAINTENANCE_` from tests


Flamefire:
Some tests with the tests provided in #5 (https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-5/comment-32/issue_5_new_includes_r2.patch) show a kind of failure however. But it seems there a decision is required.

`new_header_path_local.cc` includes `direct_near.h` which includes `indirect.h` but does not use it but the `.cc` does. This unfortunately results in the removal if `indirect.h` (correct) and adds `tests/cxx/indirect.h` to the `.cc`. What is the behaviour wanted? Adding `indirect.h` is possible (and seems reasonable as its in the same folder) by using `insert` into the vector so it is before the longer include. This causes quite some other tests to fail, that expect `indirect.h` is added with the fuller path.

kimgr:
Many thanks for the patch! I'll try to dig into this at some point, but I'm busy with life for the rest of the week.

Until then, could you also clang-format your changes, so they follow style?

Flamefire:
I found a very related bug that makes the discussion kind of superflous: Running iwyu on a test without the `-I .` results in suggesting `#include tests/cxx/someinclude.h`which is plainly wrong when `-I .` is not specified.

The reason is, that the header can drop through the search path check in `ConvertToQuotedInclude, which should not really happen. If it happens, we have either an absolute include path, or need to use the base name, assuming the file is in the same folder as the includer. Also the `./` is never matched as a prefix in the current check which makes a fix a bit more complicated, but I'll come up with something.

Flamefire:
Ok, I added some tests and fixed the described bug. I tested the code with a few hand-made examples using absolute and relative paths and also with the whole test-suite.

vsapsai:
Main complexity comes from seemingly arbitrary usage of absolute paths in certain places. And it is unclear why in certain cases we need `includerPath`/`including_filepath` and in other cases default value "" is sufficient. Can you please explain these moments?

You adding `includerPath` to `ConvertToQuotedInclude` gave me an idea. Can we fix case 2? Comment

> // Case 2: Uses the implicit "-I." entry on the search path.  Always local.

is not entirely correct. In header search there is implicit "-I <includer_filepath>", not "-I .". Can we use `includerPath` to figure out that included file was found relatively to the includer?

Flamefire:
I agree. In my opinion ALL header files should be referenced by their absolute paths (as they would be found during include resolution) This avoids ambiguity among headers of the same name. Reduction of the path should then be done when suggesting includes based on a) usages in the same file b) usage scheme in general (might be hard...) c) shortest path

I don't quite get what moments you describe there, but I try to explain:
You can see one of the problems in the commit with the modified test cases: If you don't pass `-I.` to the compiler, and call iwyu from a different directory than the source, it will still assume, that `-I.`  is passed giving an include path relative to the invocation path (which is completely wrong)
In this case I want to path relative to the includer. But if I pass `-I<path>` (with "." beeing a possible `<path>`) to the compiler, a better guess might be the pass relative to this path (as the user explicitly stated that he wants this path)
This also helps as e.g. in my company we have a main directory (added via `-I`) and all includes must be relative to that one.

About the comment: Yes... I meant, that the includer file path is implicitly given. So what I do there after going through the search paths (which take precedence) is checking the common prefix of includer and includee. So the resulting include path is relative to the includer, if the includee is in the same or a subfolder. So yes, what you suggest is already happening.

iwyu_path_util.cc
```diff
@@ -109,8 +109,7 @@ string CanonicalizeHeaderSearchPath(const string& path) {
 
 string GetCanonicalName(string file_path) {
   // Get rid of any <> and "" in case file_path is really an #include line.
-  StripLeft(&file_path, "\"") || StripLeft(&file_path, "<");
-  StripRight(&file_path, "\"") || StripRight(&file_path, ">");
+  file_path = RemoveQuotes(file_path);
```
kimgr: Can you call this `StripQuotes` instead, to follow the Python terminology already used?

iwyu_path_util.cc
```diff
@@ -178,9 +177,13 @@ string GetParentPath(const string& path) {
 
 // Converts a file-path, such as /usr/include/stdio.h, to a
 // quoted include, such as <stdio.h>.
-string ConvertToQuotedInclude(const string& filepath) {
+string ConvertToQuotedInclude(const string& filepath, const string& includerPath) {
```
kimgr: Name `includePath` consistently with above, i.e. `including_filepath`.

iwyu_path_util.cc
```diff
@@ -178,9 +177,13 @@ string GetParentPath(const string& path) {
...
+  if (filepath == "<built-in>")
+    return filepath;
+
   // First, get rid of leading ./'s and the like.
   string path = NormalizeFilePath(filepath);
+  const bool isAbsolute = IsAbsolutePath(path);
```
kimgr: We usually don't use `const` on local values, please cut it here too.

iwyu_path_util.cc
```diff
@@ -190,7 +193,16 @@ string ConvertToQuotedInclude(const string& filepath) {
...
     // All header search paths have a trailing "/", so we'll get a perfect
     // quoted include by just stripping the prefix.
-    if (StripLeft(&path, it->path)) {
+    bool matched = StripLeft(&path, it->path);
+    if (!matched && !IsAbsolutePath(it->path)) {
+        if (isAbsolute) {
+            string absSearchPath = MakeAbsolutePath(it->path);
+            StripRight(&absSearchPath, "./");
+            matched = StripLeft(&path, absSearchPath);
+        } else
```
kimgr: else wants braces if the preceding if has braces.

iwyu_path_util.cc
kimgr: You probably want to name this so it's clearer what it refers to, e.g. `isIncludeeAbsolute`

iwyu_path_util.cc
```diff
@@ -190,7 +193,16 @@ string ConvertToQuotedInclude(const string& filepath) {
   for (Each<HeaderSearchPath> it(&search_paths); !it.AtEnd(); ++it) {
     // All header search paths have a trailing "/", so we'll get a perfect
     // quoted include by just stripping the prefix.
-    if (StripLeft(&path, it->path)) {
+    bool matched = StripLeft(&path, it->path);
+    if (!matched && !IsAbsolutePath(it->path)) {
```
kimgr: The nested conditionals here are killing me... Do you think it's possible to simplify this if you switch to guard-clause-style ifs with early returns?

iwyu_path_util.cc
```diff
@@ -190,7 +193,16 @@ string ConvertToQuotedInclude(const string& filepath) {
...
+        if (isAbsolute) {
+            string absSearchPath = MakeAbsolutePath(it->path);
+            StripRight(&absSearchPath, "./");
```
kimgr: Why would an absolute path ever start with "./"? I'm guessing you put this in because you needed it, but it indicates there's something wrong with `MakeAbsolutePath`

run_iwyu_tests.py
```diff
@@ -83,26 +83,29 @@ def setUp(self):
...
-      'lateparsed_template.cc': ['-fdelayed-template-parsing'],
+      'lateparsed_template.cc': ['-I .', '-fdelayed-template-parsing'],
       'ms_inline_asm.cc': ['-fms-extensions'],
       'prefix_header_attribution.cc': [
-          self.Include('prefix_header_attribution-d1.h')],
+          '-I .', self.Include('prefix_header_attribution-d1.h')],
       'prefix_header_includes_add.cc': prefix_headers,
       'prefix_header_includes_keep.cc': prefix_headers,
       'prefix_header_includes_remove.cc': prefix_headers,
+      'new_header_path_local.cc': [''], # no include paths
```
kimgr: This and and the manual adding of `-I .` everywhere makes me want to cry just a little bit :-)

I'll try and think up a cleaner way to do this, though I admit I can't see anything obvious right now.

tests/cxx/headerInSubfolder.cc
```diff
@@ -0,0 +1,30 @@
...
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Test that subfolders are correctly recognized
+
+#include "subfolder/directSub.h"
```
kimgr: No camelCase in header names, please.

tests/cxx/subfolder/indirectSub.h
```diff
@@ -0,0 +1,17 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+
+#ifndef DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_INDIRECT_SUB_H_
+#define DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_INDIRECT_SUB_H_
+
+class IndirectSubClass {};
```
kimgr: This name is a little unclear, I read it as sub-class in the "derived-class" sense at first.

tests/cxx/subfolder/indirectSub.h
```diff
@@ -0,0 +1,17 @@
+//===--- direct_near.h - test input file for iwyu -------------------------===//
```
kimgr: comment vs filename

tests/cxx/subfolder/directSub.h
```diff
@@ -0,0 +1,25 @@
+//===--- direct_near.h - test input file for iwyu -------------------------===//
```
kimgr: comment vs filename

tests/cxx/headerInSubfolder.cc
```diff
@@ -0,0 +1,30 @@
+//===--- new_header_path_local.cc - test input file for iwyu --------------===//
```
kimgr: comment vs filename, here and elsewhere

kimgr:
I've read through this a couple of times, and I'm still finding it hard to understand what's going on. But it looks like this patch generally solves the problem, so I'd like to see it refined a little bit. It would be great to get this solved.

This also reminds me of issue #162, it might be a good first step toward solving that.

run_iwyu_tests.py
vsapsai: I was thinking about introducing extra argument `search_paths` to `TestIwyuOnRelativeFile`. If it is `None`, we default to `['.']`. And add these search paths to other flags with `-I`. Not sure it is actually better.

vsapsai:
Sorry, @Flamefire, I wasn't clear enough. I thought about fixing *just* case 2 without changes in `GetCandidateHeadersForFilepathIncludedFrom`. Looks like it can reduce the complexity. Another argument for trying it is that based on commit log you started with changes in `GetCandidateHeadersForFilepathIncludedFrom` and later modified `ConvertToQuotedInclude`. Maybe invasive changes only in `ConvertToQuotedInclude` can be sufficient.

iwyu_path_util.cc
Flamefire: It is a Strip**Right**. The path can **end** with a ` ./`

iwyu_path_util.cc
Flamefire: You mean `includerPath` -> `includer_filepath`? So no CamelCasing for variables?

run_iwyu_tests.py
Flamefire: Sorry but nothing wrong there. Manually adding `-I .` is just the same as adding `-std=c++11` or any other compiler flag. There should be no default. And `['']` just says: "no compiler flags"
Please do not default to `.`. This is a mistake. Why should that be a default? Why would you do something like this: `cd build && clang++ ../src/foo.cpp -I.` ???

Flamefire:
@kimgr I know. The problem is IMO a flaw in the general logic. Using absolute paths until the very end consistently and reduce it only when suggesting an include would most likely make it better. But that would require major changes so far as I got the code...
@vsapsai Actually I wanted a change in `GetCandidateHeadersForFilepathIncludedFrom` (mostly) but that requires the change in `ConvertToQuotedInclude`. Actually I consider it a design flaw, to compare 'quoted includes' at this point (see above) If there were only absolute paths, it would be much easier as there is no need to convert the paths from and to quoted and from and to absolute paths.

kimgr:
@Flamefire We've wanted to move toward comparing `FileEntry*` instead of paths, they are identity pointers to known include files. But it's been hard to move from the current structure. I think absolute paths as an intermediate step might help.

iwyu_path_util.cc
kimgr: That does seem to be the prevailing style, I hadn't really noticed until I felt `includePath` looked weird :-)

I'm sure there are exceptions, but the vast majority seems to be all_lower_with_underscore.

iwyu_path_util.cc
kimgr: Ah, gotcha. Then I'd suggest you move that into MakeAbsolutePath!

run_iwyu_tests.py
kimgr: It's not wrong per se, it's just a total break with convention.

All test files are written with the implicit assumption that `-I .` is passed on the command-line, and only files with any other needs are listed in the clang_flags map above.

So I don't think it's indicative of any actual use case, it's just how the tests were written. I don't know what drove that default, originally, but if we need to break it, we should probably do that as a separate commit.

Flamefire:
@kimgr But I think this is beyond the scope of this PR. This PR just tries to fix some important issues/failures. However it smoothes the paths for going to absolute paths.

kimgr:
Yes, I think we're in agreement here.

iwyu_path_util.cc
Flamefire: Good idea!

run_iwyu_tests.py
Flamefire: I don't agree. The default `-I .` drove the current code in the wrong direction leading to (part of) this bug. The change of the "default" makes this bug verifiable. While still leaving the other tests in a valid setting (`-I .` manually added) so the tests relying on this are not changed.

Flamefire:
@kimgr Fixed the things you mentioned. Please check if I missed something. Tests are passing.

kimgr:
I've been working to understand your changes, and it's not going so well :-)

I feel like you're attacking too much at once; this makes the diff (and subsequent commit) difficult to understand and reason about the impact of.

I also have trouble understanding the details around the new optional `including_filepath` arguments -- when is it necessary to pass them? Why are they sometimes absolute and sometimes not? If I add a new call to either of these methods, how do I know whether to pass something? I think this is also what @vsapsai was asking for.

I think a lot of this could be simplified by coming to a more general convention that file paths are always absolute and normalized when they are first constructed. Now it looks like you're converting to absolutes at arbitrary places, but I could be misunderstanding.

Also, I have a vague feeling that the rule for the implicit header search path (where the current file's dirname is checked) is not 100% consistent among compilers. I seem to recall GCC and Clang only add the main compilation unit's path, and MSVC adds the current header's path. I've never properly verified this, but if there are differences, it might cause us to generate non-portable code. Then again, perhaps we already are. Just thought I'd mention it.

Do you think you could explain in brief which different problems you're attempting to solve, and which parts of the patch address which problem?

I'd love to move forward with this, but at this point I don't understand it well enough to feel comfortable maintaining it.

Flamefire:
Will answer this more complete, but about the include search path: http://stackoverflow.com/questions/179213/c-include-semantics --> The search path is always implementation defined but the includers path is searched first by all mentioned compilers (gcc, clang, msvc, ...)

Flamefire:
Ok, back from holiday: Skip to the text after the bold line for step-by-step guide through the change.
I already answered your question about the implicit search path. The current code added `-I .` as the implicit search path, which is wrong. If any, the implicit search path is the path of the current (including) file. The `including_filepath` argument is required to get this implicit path. The reason why it is optional is for places, where the path is not known. In my investigation so far I found, that it is not always required.

The place where it is required is in the function `GetCandidateHeadersForFilepathIncludedFrom` which is the main function, that I modified. All other changes are (mostly) to implement the required functionality in the callees (for example the ability to pass the `including_filepath` from `GetCandidateHeadersForFilepathIncludedFrom` to `ConverToQuotedInclude` and `GetCandidateHeadersForFilepath`)
`ConvertToQuotedInclude` needs an absolute path for its 2nd argument (`including_filepath`) because in https://github.com/include-what-you-use/include-what-you-use/pull/276/files#diff-e41a3f0a0d6c4056f2fb245b901b154dR228 the implicit search path is used to truncate as much as possible from the includees absolute path. The relative path would most likely be relative to the compilers invocation path, which would the wrong path to use (relative to includer would be ok) One could probably use `StripPathPrefix` here but need to check, that the `including_filepath` is non-empty and one passes a path relative to includer or an absolute path.

**So far for point-wise explanation. I'll try to explain the change itself now:**

`GetCandidateHeadersForFilepathIncludedFrom` is the function to change to fix the mentioned bugs as we need a) the includer and b) the includee. An include of the form `#include "foo.h"` is not generally wrong or right, but depends, where the includer lies (same path as `foo.h`)
So first we get the quoted includee by calling `ConvertToQuotedInclude` with extra includers filepath (absolute, see above). In this: If we can match a path from the explicit `-I` option(s) we take this. `StripPathPrefix` is like `StripLeft` but additionally tries absolute paths. This is required for something like `-I.` where `.` would not match correctly but the absolute version does. If no option matches and an includer_path was given, we remove that (this is the implicit one) unless the includee is absolute (If one writes an absolute include, it could be a system-dependent one, which probably is not a good idea to convert to a relative path) Back to `GetCandidateHeadersForFilepathIncludedFrom` we now have a good include (either relative to include option, including file or absolute path if everything else failed) and can check the maps for any mappings to other (private-public) headers. For no mappings we call into `GetCandidateHeadersForFilepath` which is pretty much the same (use `ConvertToQuotedInclude` with extra info, check and return)
If that found more than 1 header we try to use a writing, that was already used. If that fails, we calculate the path relative to the current file (strip prefix) and add it to the list, if it exists (if the includee is not in the same dir or a subdir the stripping will fail). The position depends on whether we (probably) used a search path option or not.

Final Note: Yes absolute paths everywhere would actually help, but one needs also the paths as written in the file(s) to choose a reasonable short version at the end. (If I have `myLib/foo.h` I don't want it replaced by `foo.h` even if it would be possible)

iwyu_include_picker.cc
```diff
@@ -1303,15 +1303,15 @@ vector<string> IncludePicker::GetCandidateHeadersForFilepath(
 vector<string> IncludePicker::GetCandidateHeadersForFilepathIncludedFrom(
     const string& included_filepath, const string& including_filepath) const {
   vector<string> retval;
-  const string quoted_includer = ConvertToQuotedInclude(including_filepath);
-  const string quoted_includee = ConvertToQuotedInclude(included_filepath);
+  const string quoted_includer = ConvertToQuotedInclude(including_filepath, MakeAbsolutePath(GetParentPath(including_filepath)));
```
vsapsai: This seems logically incorrect. We don't know who includes `including_filepath` and assuming it includes itself is far-fetched. Also keeping old 

`const string quoted_includer = ConvertToQuotedInclude(including_filepath);`

doesn't break anything (probably it indicates some tests are missing).

iwyu_include_picker.cc
```diff
@@ -1329,13 +1329,32 @@ vector<string> IncludePicker::GetCandidateHeadersForFilepathIncludedFrom(
...
+                       << "\n";
+              // If the included path is absolute, we prefer the short one
+              // else we prefer the current one, which is probably relative to
+              // a search directory
+              if(IsAbsolutePath(StripQuotes(*it)))
+                  retval.insert(it, "\"" + strippedIncludee + "\"");
+              else
+                  retval.push_back("\"" + strippedIncludee + "\"");
+          }
+      }
```
vsapsai: This `else` branch remains the most confusing part for me. It is unclear why we do these actions only when `MaybeGetIncludeNameAsWritten` doesn't return anything. In what cases `MaybeGetIncludeNameAsWritten` returns empty string? It seems it's empty because of deficiencies in other parts of the program, not that it is empty on purpose in specific cases.

Also I tried to comment out this `else` block and see how tests behave. Tests broke in a way that IWYU recommends to use
```c++
#include "/subfolder/indirect_subfolder.h"
#include "/indirect.h"
```
To me this method doesn't look like appropriate place to fix leading slash. And it indicates that in some place we didn't strip path prefix correctly and inadvertently created false absolute path. If you know why that happened, it would be very helpful. Otherwise I'll need to figure out in debugger where those paths come from.

run_iwyu_tests.py
vsapsai: The problem here is adding `-I .` manually (it is not really a default value if you have to spell it out explicitly). Another problem is perception. For me at cursory glance it looks broken and diverts attention. It's not an issue of correctness but of readability and understandability. Dealing with subjective issues can be frustrating but we cannot brush them away.

vsapsai:
Thanks for explanation. Currently I understand intention behind `ConvertToQuotedInclude`, `StripPathPrefix`. I still don't understand `else` block in `GetCandidateHeadersForFilepathIncludedFrom`, please see inline comments for more details.

run_iwyu_tests.py
Flamefire: Ok I stated it a bit wrong. `-I .` was not the default but a forced setting. So I changed it to be the default, to make as few changes as possible. A more reasonable default value would be an empty list.
But sure, we could also add another param for include paths to make it cleaner (only `.` instead of `-I .`). I still think the default should be empty...

iwyu_include_picker.cc
Flamefire: `MaybeGetIncludeNameAsWritten` is meant to be return something, if possible (hence the "maybe") I assume it should return the include line unmodified if it was already used. So IWYU does not change an include line the user has explicitly written and also to "adapt" to user style. If we did not find a user supplied include line (->New include line) we are in the else branch. So empty is on purpose.

This is NOT a fix for the leading slash, although I'm curious where it originated. The else is really for handling new headers. BTW: It is not introduced by this PR. So refactoring of this (like making the function return a bool and taking a pointer param) should be done outside of this PR.

iwyu_include_picker.cc
Flamefire: `quoted_includer` is only used to search the friends map. This map looks a bit strange to me. IMO the keys should be absolute paths but I think they aren't. I just included this change to have it consistent. It is basically the include as-if it would be included. (includer including itself, sounds kinda confusing, but I think you get the point)

kimgr:
OK, so here's a concrete suggestion to make this easier to digest: split the PR by test. If you can partition the change so you submit each breaking test with the change to fix it, it will be much easier to understand.

If we're convinced that the `-I .` needs to go, can you make that change to the test runner first, with all the existing tests in place?

Thanks!

Flamefire:
It will be hard as this is already small. But ok, I try. First I split off the part which allows tests without `-I .`. After this it is possible to add tests without the `-I .` that will then fail which this PR fixes.

I can probably split off another PR where the `-I .` is omitted from the clang flags, iwyu assumes it anyways and produces wrong results as it assumes the flag was given. However this split is not a good idea, as the reasonable include suggestion would be one relative to the current file, which means that this whole PR is required...
So I would only rebase this PR on #289 once it is merged as I don't think one could make it smaller.

iwyu_include_picker.cc
vsapsai: OK, accidental erroneous absolute path comes from `ConvertToQuotedInclude(including_filepath, MakeAbsolutePath(GetParentPath(...` `GetParentPath` uses [`llvm::sys::path::parent_path`](http://llvm.org/docs/doxygen/html/namespacellvm_1_1sys_1_1path.html#a734605aeef896f419c79c53d8af8f2e2) and it doesn't add trailing slash. In `GetParentPath` I tried to ensure non-empty path always has trailing slash and it didn't break anything. Moreover, I reverted this block I'm commenting on in PR back to

```c++
  const string& quoted_include_as_typed
      = MaybeGetIncludeNameAsWritten(including_filepath, included_filepath);
  if (!quoted_include_as_typed.empty()) {
    vector<string>::iterator it = std::find(retval.begin(), retval.end(),
                                            quoted_includee);
    if (it != retval.end())
      *it = quoted_include_as_typed;
  }
```
and new tests are still passing. Can you please verify trailing slash helps? Maybe there are other cases where your change is required but these cases aren't tested. Then we need better test coverage.

Flamefire: Yes, adding a slash indeed helps and tests pass. Running IWYU with the reverted code above on my real codebase shows however that I still get absolute paths. I've seen different behaviour when the cc file is passed as absolute vs relative. So I assume this is hat the test don't cover yet

Flamefire: I added 3 more commits so the test suite passes with absolute paths. If you like it smaller I suggest the following:
- Merge #289 to split off the test suite changes
- Remove the 3 commits I just added from this PR and revert the above changes (-> failures when using absolute paths)
- Add another PR with the above change and the 3 commits together with a change to the test suite, to test all tests with absolute input filepaths (I got that locally by using a modified version of `TestIwyuOnRelativeFile` which simply removes the current path from the output and compares this. Works on all current tests but might need a later refactoring. One can probably just use 1 function and check for an absolute path

Flamefire:
I rebased and restructured this PR. Absolute paths are now done in another PR: #294
I aimed to simplify this PR: Add tests, Fix general issues, Fix actual issue

iwyu_path_util.cc
```diff
@@ -173,7 +176,11 @@ string MakeAbsolutePath(const string& base_path, const string& relative_path) {
 
 string GetParentPath(const string& path) {
   llvm::StringRef parent = llvm::sys::path::parent_path(path);
-  return parent.str();
+  string result = parent.str();
+  // Ensure trailing slash
+  if(!result.empty() && result[result.size() - 1] != '/')
```
kimgr: This looks like it would be better spelled:

    if (!EndsWith(result, "/"))

iwyu_path_util.cc
```diff
@@ -149,6 +149,9 @@ string NormalizeFilePath(const string& path) {
   string result = CanonicalizeFilePath(path);
   while (StripLeft(&result, "./")) {
   }
+  // Don't use StripRight as we might have 'foo/../' here
```
kimgr: This comment is either wrong or needs more elaboration :-)

It looks like you're checking for `"/./"` but stripping only `"./"` -- is that why `StripRight` is not possible?

iwyu_path_util.cc
```diff
@@ -149,6 +149,9 @@ string NormalizeFilePath(const string& path) {
...
+  while (EndsWith(result, "/./"))
```
kimgr: Do you need to consider Windows `\` path separators here? I'm guessing yes.

iwyu_path_util.cc
kimgr: Oh, OK. You can't use `StripRight("./")` because it might cut half of a trailing `"../"` You can probably make the comment clearer here...

iwyu_path_util.h
```diff
@@ -69,14 +69,22 @@ string MakeAbsolutePath(const string& base_path, const string& relative_path);
...
 // "".  In the line '#include <time.h>', "<time.h>" is the quoted
 // include.
 
 // Converts a file-path, such as /usr/include/stdio.h, to a
 // quoted include, such as <stdio.h>.
-string ConvertToQuotedInclude(const string& filepath);
+string ConvertToQuotedInclude(const string& filepath, const string& includer_path = "");
+
+// Removes quotes ("" or <>) from a string
+string StripQuotes(string s);
```
kimgr: This would be more consistent with the other Strip functions if it took a `string*` and returned void.

iwyu_path_util.cc
Flamefire: Ok might be clearer that way

iwyu_path_util.cc
Flamefire: Nope. We only have forward slashes at this point (`CanonicalizeFilePath` does this)

iwyu_include_picker.cc
kimgr: Does it work if you leave out the `including_filepath` argument? If so, I'd prefer that.

iwyu_path_util.cc
Flamefire: Nope. I only want to add the slash, when the path is non-empty. So at least the empty check needs to stay. The 2nd part might be over-optimized... `result.back()` would be nicer, but that's C++11

iwyu_include_picker.cc
Flamefire: It might... But this may break user-defined mappings for that file. I can remove it and we wait for a bug report or open an issue to check that. Not sure if the mappings work with different relative paths.

iwyu_path_util.h
Flamefire: Agreed...

iwyu_path_util.cc
kimgr: And you might need to consider Windows path separators here, or use `CanonicalizeFilePath` to turn the slashes the right way.

iwyu_path_util.cc
```diff
@@ -167,18 +169,46 @@ string MakeAbsolutePath(const string& path) {
...
   llvm::StringRef parent = llvm::sys::path::parent_path(path);
-  return parent.str();
+  string result = parent.str();
+  // Ensure trailing slash
+  if(!result.empty() && result[result.size() - 1] != '/')
+      result += '/';
+  return result;
+}
+
+bool StripPathPrefix(string* path, const string& prefix_path) {
```
kimgr: I think I'd prefer if this just asserted that `path` and `prefix_path` are the same with respect to absoluteness, i.e.

    CHECK_(IsAbsolutePath(path) == IsAbsolutePath(prefix_path));

That way you can collapse the logic into `StripLeft`, I think.

kimgr:
I still think this patch is too big...  Is there a way to drive it test-case by test-case? There seem to be quite a few changes that are not covered by tests (e.g. stripping of `./`, `ConvertToQuotedInclude` handling `<built-in>`), so they seem pretty arbitrary. I understand they are necessary, but it would be nice to get test coverage for them.

That's partly why I'd like to start smaller, because it's easier to reason about design vs tests.

And I think I've mentioned this before, but it would be nice to do all path comparison with absolute paths. Your patch looks like it does the checks in a few places, but it's very optional due to the default values for `including_filepath`.

I realize I'm nagging at this point, so I'll see if I can put my money where my mouth is and post a series of PRs in the style I'm expecting, just to give you an idea what it might look like, and to better understand your suggested changes. Hopefully I'll have some time for it this week.

iwyu_path_util.cc
Flamefire: I think this is possible, if we make sure, this is only called with absolute, normalized paths (which I think should be done anyway)

Flamefire:
No, I don't see a reasonable way to make it smaller. This PR has 69 lines added (if you exclude the tests) and each of them is entirely related to fixing the testcases (proberly). The testcases itself are 2 examples: Adding and changing an include each with `-I .` and without for duality comparison. But they basically test the same thing where the counterpart simply makes sure, existing behaviour is not broken.

Stripping `./` is required, because `-I .` adds this. `<build-in>` must be handled there to a) make things faster and b) not run into problems comparing stuff as the code assumes actual paths. And it is a 2-liner! (and should have been there always, if you like, open a PR adding this 2 lines. But I think this is not required)

Sure it would be better to have absolute paths everywhere. This would also help the follow-up PR. However currently we are running around in circles: Make the PR smaller but also introduce more stuff which makes it bigger. I made the changes as small and less intrusive as possible and say merge these 2 PRs which use absolute paths where absolutely **necessary** and then iteratively change the whole codebase to work with absolute paths where ever **possible**

Again: There are very few codechanges. Most of the changes are the added testcases which are made like they are to highlight the problem at hand (the difference of `-I .`and no `-I .`) Introducing absolute paths throughout the codebase might actually worsen things, if there are no tests and absolute paths are not working. So do them **after** these 2 PRs.

Flamefire:
@kimgr I worked through your comments and now also store the search paths absolute and normalized. This was required to cleanly implement the suggestion for `StripPathPrefix`
Using more absolute paths may break something as that would touch to much of the codebase.

iwyu_path_util.cc
```diff
@@ -190,18 +214,26 @@ string ConvertToQuotedInclude(const string& filepath) {
...
+    else if (it->path_type == HeaderSearchPath::kSystemPath)
+      return "<" + path + ">";
+    else
+      return "\"" + path + "\"";
   }
 
-  // Case 2: Uses the implicit "-I." entry on the search path.  Always local.
+  // Case 2: Uses the implicit "-I <basename current file>" entry on the search path
+  if (!includer_path.empty())
+      StripLeft(&path, includer_path);
```
kimgr: It looks like this will only work if `includer_path` is absolute -- can you add a:

    CHECK_(includer_path.empty() || IsAbsolutePath(includer_path));

to the top of this method?

iwyu_path_util.cc
```diff
@@ -167,20 +178,33 @@ string MakeAbsolutePath(const string& path) {
 string MakeAbsolutePath(const string& base_path, const string& relative_path) {
   llvm::SmallString<128> absolute_path(base_path.c_str());
   llvm::sys::path::append(absolute_path, relative_path);
-
+
```
kimgr: Please remove trailing whitespace

iwyu_path_util.cc
```diff
@@ -97,20 +97,28 @@ string CanonicalizeFilePath(const string& path) {
 }
 
 string CanonicalizeHeaderSearchPath(const string& path) {
-  string result = CanonicalizeFilePath(path);
+  string result = path;
+  // Header search paths must be absolute
+  if (!IsAbsolutePath(result))
```
kimgr: Please collapse this to:

    result = MakeAbsolutePath(path);

I think it should amount to the same thing.

iwyu_path_util.cc
```diff
@@ -167,20 +178,33 @@ string MakeAbsolutePath(const string& path) {
...
-  return parent.str();
+  string result = parent.str();
+  // Ensure trailing slash on non-empty paths
+  if(!result.empty() && result[result.size() - 1] != '/' && result[result.size() - 1] != '\\')
+      result += '/';
+  return result;
+}
+
+bool StripPathPrefix(string* path, const string& prefix_path) {
+  // Makes only sense if both are absolute or both are relative (to same dir)
```
kimgr: // Only makes sense [...]

iwyu_path_util.cc
```diff
@@ -149,6 +157,9 @@ string NormalizeFilePath(const string& path) {
   string result = CanonicalizeFilePath(path);
   while (StripLeft(&result, "./")) {
   }
+  // Remove trailing './' but don't split '../'
```
kimgr: I wonder if we should now remove `NormalizeFilePath` entirely, and move this stuff to `CanonicalizeFilePath` instead?

`llvm::sys::path` has a `remove_dots` method that collapses all dot components, as far as I can tell:
http://llvm.org/docs/doxygen/html/namespacellvm_1_1sys_1_1path.html#a07079865968cc0fac090076bd8681750

If we move that into `CanonicalizeFilePath`, life should be better overall, I think?

iwyu_path_util.cc
```diff
@@ -167,20 +178,33 @@ string MakeAbsolutePath(const string& path) {
...
+  
   return absolute_path.str();
 }
 
 string GetParentPath(const string& path) {
   llvm::StringRef parent = llvm::sys::path::parent_path(path);
-  return parent.str();
+  string result = parent.str();
+  // Ensure trailing slash on non-empty paths
+  if(!result.empty() && result[result.size() - 1] != '/' && result[result.size() - 1] != '\\')
```
kimgr: This looks messy... What if you canonicalized the returned path before checking for trailing slash?

Something like:

```
  string parent = CanonicalizeFilePath(
      llvm::sys::path::parent_path(path).str());
  if (parent.empty())
    return parent;

  if (parent.back() != '/')
    parent += "/";

  return parent;
```

tests/cxx/subfolder/direct_subfolder.h
```diff
@@ -0,0 +1,25 @@
...
+#ifndef DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_DIRECT_SUBFOLDER_H_
+#define DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_DIRECT_SUBFOLDER_H_
+
+#include "indirect_subfolder.h"
+
+namespace { IndirectSubfolderClass foo; }
+
+#endif  // DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_DIRECT_SUBFOLDER_H_
+
+/**** IWYU_SUMMARY
```
kimgr: I don't think this is evaluated unless you add a `--check_also` argument to IWYU. You can just remove the IWYU summary matching from this header.

tests/cxx/direct_near.h
```diff
@@ -0,0 +1,29 @@
...
+// provided header search path -I .
+
+#ifndef DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_DIRECT_NEAR_H_
+#define DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_DIRECT_NEAR_H_
+
+#include "indirect.h"
+
+#endif  // DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_DIRECT_NEAR_H_
+
+/**** IWYU_SUMMARY
```
kimgr: No need for an IWYU summary pattern here, it's only evaluated for associated headers and headers named by `--check_also`

iwyu_path_util.h
```diff
@@ -69,14 +69,22 @@ string MakeAbsolutePath(const string& base_path, const string& relative_path);
 // Get the parent of path.
 string GetParentPath(const string& path);
 
+// Tries to strip the prefix_path from the front of path
+// path assumed to be normalized but either absolute or relative
+// Returns true, if path was stripped
```
kimgr: Remove comma after `true`

tests/cxx/subfolder/indirect_subfolder.h
```diff
@@ -0,0 +1,17 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+
+#ifndef DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_INDIRECT_SUBFOLDER_H_
+#define DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_INDIRECT_SUBFOLDER_H_
+
+class IndirectSubfolderClass {};
+
+
```
kimgr: Kill excess blank line

kimgr:
Thanks for the updates! Here's a new round of comments.

I agree with @vsapsai that the `GetCandidateHeaders` changes look arbitrary. I think it's worth taking bugs here if they occur, so we can work on clarifying how mappings are supposed to work.

As a spontaneous thought, maybe `GetCandidateHeadersForFilePathX` should disappear, and callers be responsible for constructing a quoted include and query on that directly (assuming callers have more context, they usually do.)

iwyu_path_util.cc
Flamefire: Sure. Question is: In this PR? You wanted to keep it small, didn't you?

iwyu_path_util.cc
Flamefire: I did not want to canonicalize the file path, as that would be beyond the scope of this function. Put in a backslash-path and get a slash-path although you only called `GetParent`?
And `parent.back()` is C++11

Flamefire:
Alright, fixed the mentioned things.
You mean rename `GetCandidateHeadersForFilePath` to `GetCandidateHeadersForQuotedInclude` and have it assume quoted includes as the param?  I could agree with that. Although the calls need to be modified. However one of the calls (the only one using the 2nd argument) already has the quoted include.

iwyu_path_util.cc
kimgr: C++11 is not a problem, we're already bound by the same language standards as LLVM/Clang, so we're at C++~12.5 (GCC 4.8 and VS2013, IIRC).

> I did not want to canonicalize the file path, as that would be beyond the scope of
> this function. Put in a backslash-path and get a slash-path although you 
> only called GetParent?

You're right, that makes no sense.

> I'd rather remove this and have an extra function NormalizeFolderPath that calls
> NormalizeFilePath and adds the slash

Sounds promising!

iwyu_path_util.cc
kimgr: Above all I want to keep it cohesive -- solve the smallest problem possible, with the right amount of redesign to make it fit well.

This one is borderline, I think, it would be nice if we could get a grip on paths, but maybe that's a separate change set.

I've always been confused by `Normalize` vs `Canonicalize`, I think it might be a case where someone just invented words to signify difference instead of profound meaning. And if one of them can be made redundant, I'm all for removing...

kimgr:
Can you also get clang-format and use it to format the changed lines? You have a few 80-column overruns and unconventional indentation. We have a .clang-format configuration file, so the tool should pick up our style directly. Also, we follow Google style to the letter, so comments should be full sentences and end with a full stop.

I went through all uses of `GetCandidateHeadersForFilepath.*`, but it doesn't look like there's a quoted include readily available in very many places there either. Perhaps best to leave this alone. I think longer term we'd want some kind of database for source files where we can resolve a `FileEntry*` from a path, include-names used to include them, a `SourceLocation`, etc.

iwyu_path_util.cc
Flamefire: Doing this in a new PR

Flamefire:
@kimgr Rebased and updated again. clang-format did not turn out to be helpful as it changes quite a lot more than just my changes. I fixed the 80-column overruns (hope I got them all) manually and tried to make sure, the comments are ok. Please check what I missed.

iwyu_path_util.cc
```diff
@@ -177,18 +189,27 @@ string ConvertToQuotedInclude(const string& filepath) {
...
+      return "<" + path + ">";
+    else
+      return "\"" + path + "\"";
   }
 
-  // Case 2: Uses the implicit "-I." entry on the search path.  Always local.
+  // Case 2:
+  // Uses the implicit "-I <basename current file>" entry on the search path.
+  if (!includer_path.empty())
+      StripPathPrefix(&path, NormalizeDirPath(includer_path));
```
kimgr: Over-indented.

iwyu_path_util.cc
```diff
@@ -177,18 +189,27 @@ string ConvertToQuotedInclude(const string& filepath) {
...
     // All header search paths have a trailing "/", so we'll get a perfect
     // quoted include by just stripping the prefix.
-    if (StripLeft(&path, it->path)) {
-      if (it->path_type == HeaderSearchPath::kSystemPath)
-        return "<" + path + ">";
-      else
-        return "\"" + path + "\"";
-    }
+
+    if (!StripPathPrefix(&path, it->path))
```
kimgr: I think this block was clearer before, i.e. something like:

    // All header search paths have a trailing "/", so we'll get a perfect
    // quoted include by just stripping the prefix.
    if (StripPathPrefix(&path, it->path)) {
      if (it->path_type == HeaderSearchPath::kSystemPath)
        return "<" + path + ">";
      else
        return "\"" + path + "\"";
    }

The `continue` and if/else made it harder to follow.

iwyu_path_util.h
```diff
@@ -66,14 +66,22 @@ string MakeAbsolutePath(const string& base_path, const string& relative_path);
...
+
 // Below, we talk 'quoted' includes.  A quoted include is something
 // that would be written on an #include line, complete with the <> or
 // "".  In the line '#include <time.h>', "<time.h>" is the quoted
 // include.
 
 // Converts a file-path, such as /usr/include/stdio.h, to a
 // quoted include, such as <stdio.h>.
-string ConvertToQuotedInclude(const string& filepath);
+string ConvertToQuotedInclude(const string& filepath, const string& includer_path = "");
```
kimgr: Line too long, wrap between the arguments

iwyu_path_util.cc
```diff
@@ -83,8 +83,7 @@ string Basename(const string& path) {
 
 string GetCanonicalName(string file_path) {
   // Get rid of any <> and "" in case file_path is really an #include line.
-  StripLeft(&file_path, "\"") || StripLeft(&file_path, "<");
-  StripRight(&file_path, "\"") || StripRight(&file_path, ">");
+  StripQuotes(&file_path);
```
kimgr: Actually, now that I look at it, I can't find a single place where this is called with a quoted path. So let's replace `StripQuotes` with:

    CHECK_(!IsQuotedInclude(file_path));

and remove `StripQuotes` entirely.

tests/cxx/subfolder/direct_subfolder.h
```diff
@@ -0,0 +1,19 @@
...
+//===----------------------------------------------------------------------===//
+
+// This file includes and uses only another file in the subfolder
+
+#ifndef DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_DIRECT_SUBFOLDER_H_
+#define DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_TESTS_DIRECT_SUBFOLDER_H_
+
+#include "indirect_subfolder.h"
+
+namespace { IndirectSubfolderClass foo; }
```
kimgr: Does this have any effect on the tests? I can't see that it's referenced anywhere...? If not, please remove it.

iwyu_include_picker.cc
```diff
@@ -1304,14 +1305,16 @@ vector<string> IncludePicker::GetCandidateHeadersForFilepathIncludedFrom(
     const string& included_filepath, const string& including_filepath) const {
   vector<string> retval;
   const string quoted_includer = ConvertToQuotedInclude(including_filepath);
```
kimgr: Can you add a comment above this line explaining why it's OK not to pass in any `including_filepath`?

iwyu_path_util.h
```diff
@@ -66,14 +66,22 @@ string MakeAbsolutePath(const string& base_path, const string& relative_path);
 // Get the parent of path.
 string GetParentPath(const string& path);
 
+// Try to strip the prefix_path from the front of path.
+// path assumed to be normalized but either absolute or relative
```
kimgr: Maybe `// The path argument is assumed [...] relative.`

Full sentence, trailing period.

kimgr:
Please find the latest (last?) round of comments inline.

Re: clang-format -- there's a git-clang-format script which can presumably clang-format a git commit. I haven't used it myself, but it sounds like the right granularity :-)

I've scrutinized the test cases more closely, and I think I understand what they're proving now. They look good!

iwyu_include_picker.cc
Flamefire: Because you said so? ;-) Don't know what to write there. I would have used its own path with the reasoning that this results in the filename only unless there is an `-I` that matches. W/O any path this will default to the absolute path.

iwyu_path_util.cc
Flamefire: Sure, but please don't change your mind a 3rd time ;)

tests/cxx/subfolder/direct_subfolder.h
Flamefire: You are right. Not anymore.

iwyu_path_util.cc
kimgr: :-) Everything changed since your original diff, so I'm not feeling too bad...

tests/cxx/subfolder/direct_subfolder.h
kimgr: OK, I think I can follow your reasoning, but I wonder if it's relevant to the indirect subfolder case. IWYU doesn't care about uses in files it doesn't analyze (i.e. that are not associated to the main source file or listed in `--check_also`), so this is really just adding noise.

If you disagree, can you clarify the test to reflect your reasoning?

An anonymous namespace in a header file is suspect, so maybe you can use something like this instead:

    void f() {
      // Use the class in this header to see that IWYU
      // sees through the include chain and recommends including
      // indirect_subfolder.h even if this header satisfies the dependency
      // on IndirectSubfolderClass.
      IndirectSubfolderClass x;
    }

Comment is just an example, please write something there that reflects your intent.

iwyu_include_picker.cc
kimgr: > I would have used its own path with the reasoning that this results in the
> filename only unless there is an -I that matches. 

This seems like a really good reason to use the file's own path. I didn't pick up on that last time I asked why it was there. But it introduces sufficient implicit coupling to `ConvertToQuotedInclude` that it needs a comment to explain *why* it's necessary to pass that file path.

tests/cxx/subfolder/direct_subfolder.h
Flamefire: The comment is already there: `// This file includes and **uses** only another file in the subfolder`
Note that this is the common way in the IWYU tests for indirect includes: Include a header that uses the used type but do not include the types header itself. So yes this single line is not required for the checks. But it IS required for the standard 'indirect' model used in the tests.

kimgr: I checked the original `direct.h` and `indirect.h`, and they don't have the use you're suggesting. I don't understand your rationale for this, but I can't muster more energy to debate it.

Flamefire: You are right. Strange... I was convinced I just copied the header when I created those tests. I'll remove this.

kimgr:
In the meantime, I've updated all header guards on master. Can you rebase, and remove `DEVTOOLS_MAINTENANCE_` from your test headers? Or, if you prefer, I can do that once we get to merging this.

All in all, I'm happy with the patch now, I think it's ready to go in! @vsapsai : Do you have any more concerns?

Flamefire:
Added those 2 to the task list. If this can be merged, I'll do them.

vsapsai:
I don't have comments regarding logic, just need to test.

Flamefire:
Ok, drop me a note, when you are done or found something. Together with #294 it works for my project.

vsapsai:
Currently when your branch `sameDirInc` is applied on top of `master`, all 97 tests fail with
> iwyu_path_util.cc:85: Assertion failed: !IsQuotedInclude(file_path)

Flamefire:
@vsapsai Seems I forgot a `make install` on the last test. There is a a "quoted include" `<built-in>` that was not caught before. Fixed this now.

iwyu_path_util.cc
```diff
@@ -82,9 +82,11 @@ string Basename(const string& path) {
 }
 
 string GetCanonicalName(string file_path) {
-  // Get rid of any <> and "" in case file_path is really an #include line.
-  StripLeft(&file_path, "\"") || StripLeft(&file_path, "<");
-  StripRight(&file_path, "\"") || StripRight(&file_path, ">");
+  // For this special 'path' we just remove the quotes.
+  if (file_path == "<built-in>")
+    return "built-in";
```
kimgr: Good catch, I just noticed this too. Can you just return `file_path` instead, so we maintain `<built-in>`?

kimgr:
Thanks, minor nit. I'll update locally and test my various systems.

iwyu_path_util.cc
Flamefire: Sure? This keeps the same behavior as before (removing the `<>`) and I don't know what happens for the `BaseName("<built-in>")` stuff, although I think it will just return it unmodified and we are better of using `<built-in>` to distinguish it from an include named "build-in".

kimgr:
- FreeBSD: no test regressions
- Windows/MSVC: no test regressions

With the above three things fixed (include guards, pruned IndirectSubfolderClass and `<built-in>` return) this looks great to me!

kimgr:
Nicely done! Tests still work on both my platforms.

vsapsai:
Now there are no regressions neither on Ubuntu nor on Mac OS X.

kimgr:
Closed as of 23119ba66a3186aef334a80f57e4d8754a956e41.

Wohoo! Thanks for hanging in there! I'm going to take a look at a couple of reported bugs and see if they can be closed as well.

Issue #280, opened 2016-03-09
Diff output in fix_includes


pfaffe:
If run with `-n`/`--dry_run`, `fix_includes.py` will output a unified diff without the --- +++ headers. This output is broken and can't be copy-pasted into a patch, because there is an extra newline between the `@@ ... @@` hunk header and the first line of the hunk. The cause is in `PrintFileDiff`, where difflib's `unified_diff` adds a trailing newline to headers, and the `join` adds another one.

Here's a possible fix:
```
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -613,7 +613,7 @@ def PrintFileDiff(old_file_contents, new_file_contents):
   try:
     next(diff)
     next(diff)
-    print('\n'.join(diff))
+    print('\n'.join(l.rstrip() for l in diff))
   except StopIteration:
     pass
```

Personally I'd prefer also printing the --- +++ headers so that I can even more quickly cherry-pick changes into patch files, but I think at least the hunks should be correct.

kimgr:
Can you think of a way to test that in `fix_includes_test.py`?

pfaffe:
Testing it isn't very nice because the output goes to stdout. Here's a way to do it by temporarily intercepting it:

```
--- a/fix_includes_test.py
+++ b/fix_includes_test.py
@@ -264,6 +264,49 @@ The full include-list for simple:
     self.RegisterFileContents({'simple': infile})
     self.ProcessAndTest(iwyu_output, expected_num_modified_files=1)
 
+  def testUnifiedDiffOutput(self):
+    """Test the unified diff output generated by dry runs."""
+    infile = """
+#include <notused.h>
+
+int main() { return 0; }
+"""
+    iwyu_output = """\
+unified_diff should add these lines:
+
+unified_diff should remove these lines:
+- #include <notused.h>  // lines 2-2
+
+The full include-list for unified_diff:
+---
+"""
+    diff_expect = """\
+>>> Fixing #includes in 'unified_diff'
+@@ -1,4 +1,2 @@
+-
+-#include <notused.h>
+
+ int main() { return 0; }
+IWYU edited 0 files on your behalf.
+
+"""
+
+    # the diff is printed to stdout, so intercept that
+    cout = StringIO()
+    dry_run = True
+
+    sys.stdout, cout = cout, sys.stdout
+    self.flags.dry_run, dry_run = dry_run, self.flags.dry_run
+    self.RegisterFileContents({'unified_diff': infile})
[...]

PR #281, opened 2016-03-10
Fix for #280

pfaffe:
Strip trailing whitespaces and newlines from the dry_run diff output in `fix_includes.py`


kimgr:
See inline comments. The test case can be simplified. Other than that, the fix looks good.

pfaffe:
Thanks for the review. Applied your comments.

kimgr:
This looks good to me. I'll pull it down, rebase it into a single commit and push it. Thanks for your help!

kimgr:
Pushed manually, now live.

PR #284, opened 2016-03-13
Normalize paths with realpath

sinelaw:
Gets rid of ../ 
realpath(3) is part of posix.
Tested only on linux (so far).

### DIFF
--- a/iwyu_path_util.cc
+++ b/iwyu_path_util.cc
@@ -12,6 +12,8 @@
 #include <stddef.h>
 #include <string.h>                     // for strlen
 #include <system_error>
+#include <limits.h>
+#include <cstdlib>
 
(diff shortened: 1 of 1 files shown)


sinelaw:
Sorry, just discovered about how to run UTs. Will correct it to pass and send a new one.

vsapsai:
Didn't test, for non-Posix systems you can also try to use [llvm::sys::path::remove_dots](http://llvm.org/docs/doxygen/html/namespacellvm_1_1sys_1_1path.html#a26d86d7e55aeaa9193b05d90141f89f3). Example of usage: [tools/clang/lib/Basic/FileManager.cpp](http://llvm-cs.pcc.me.uk/tools/clang/lib/Basic/FileManager.cpp#545).

PR #286, opened 2016-03-27
Naming

kimgr:
I realized just recently that the original IWYU convention for variable names is `lowercase_with_underscores`, not `camelCase`. A few atypical names had slipped through, so I thought I'd harmonize.

Ran clang-format on all changes, so indentation sometimes changed slightly.


kimgr:
Oops, I accidentally pushed history where the author info had gone awol. Updated and re-pushed.

vsapsai:
Looks good to me.

Issue #109, opened 2015-06-15
iwyu should test the include s


vsapsai:
Originally reported on Google Code with ID 109
```
Let's take the following file:
---
#include <glib.h>
#include <stdio.h>

int main() {
  char* f = g_strdup_printf("%s", "test");
  printf("%s\n", f);
  g_free(f);
  return 0;
}
---

iwyu recommends:
$ iwyu `pkg-config --cflags glib-2.0` foo.c 
[...]

The full include-list for foo.c:
#include <stdio.h>                      // for printf
#include "glib/gmem.h"                  // for g_free
#include "glib/gstrfuncs.h"             // for g_strdup_printf

If I rewrite the file as suggested, I get:

---
#include <stdio.h>                      // for printf
#include "glib/gmem.h"                  // for g_free
#include "glib/gstrfuncs.h"             // for g_strdup_printf

  int main() {
    char* f = g_strdup_printf("%s", "test");
    printf("%s\n", f);
    g_free(f);
    return 0;
  }
---

which will fail to build with:
---
In file included from foo.c:2:
/usr/include/glib-2.0/glib/gmem.h:31:2: error: "Only <glib.h> can be included
      directly."
#error "Only <glib.h> can be included directly."
 ^
---

A way for iwyu to check if it works would be to include the header in a temporary file
and build it.

```

Reported by `sylvestre.ledru` on 2013-09-09 07:04:25

vsapsai:
Fixed by PR #288.

vsapsai:
Turned out I was wrong about `#include_next`. It worked on Mac OS X because macros are reported in `IwyuPreprocessorInfo::HandlePreprocessingDone` after we exited from the file. So we don't see a bunch of macros and don't try to map limits.h to limits.h. Looks like handling macros defined by includers should be moved later.

I have an idea to solve limits.h by ignoring `#include_next` includes. I can distinguish `#include_next` includes but something is still not working on Ubuntu.

Another issue is `stddef.h`. `/usr/include/_G_config.h` does something like

```c++
#define __need_NULL
#include <stddef.h>
```

and then `../lib/clang/3.9.0/include/stddef.h` uses this macro to decide if it needs to define NULL. It completely follows the pattern of used macro being defined by includer. But I don't want to map `stddef.h` to `_G_config.h`. Still thinking how to fix this issue. So far, reusing the existing file mapping approach fails in many cases.

vsapsai:
[...]

PR #288, opened 2016-04-11
Handle internal headers guarded by macro and x-macros (fix issue #109).

vsapsai:
I've made a few trade-offs in implementation which I'd like to explain more.

First, code doesn't distinguish between guarded internal headers and
headers with x-macros. They are handled the same way. But in tests both
patterns are tested. It is done not to cover all code paths but to test
include-what-you-use from user's perspective.

Also I check if file defining macro is immediate includer. I decided not
to check if it includes file using the macro transitively until we have
such real-life use cases. Current implementation is strict in order to
avoid unexpected results.

For some cases I am reusing mechanism that keeps files included with the
"IWYU pragma: keep" comment. The downside is that it keeps all lines
including this file which might be not entirely correct for x-macros.
Though x-macros are close to pure textual includes and we cannot reason
about textual includes.

I also didn't include a few test cases because I think they don't
represent real-life use cases. These test cases are:
* when associated header is included by non-associated header;
* when file defining macro and file using macro are both included by the
  third file.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -602,6 +602,10 @@ void IwyuFileInfo::ReportMacroUse(clang::SourceLocation use_loc,
   LogSymbolUse("Marked full-info use of macro", symbol_uses_.back());
 }
 
+void IwyuFileInfo::ReportDefinedMacroUse(const clang::FileEntry* used_in) {
+  macro_users_.insert(used_in);
+}
+
 void IwyuFileInfo::ReportIncludeFileUse(const clang::FileEntry* included_file,
                                         const string& quoted_include) {
   symbol_uses_.push_back(OneUse("", included_file, quoted_include,
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -609,6 +609,10 @@ void IwyuFileInfo::ReportIncludeFileUse(const clang::FileEntry* included_file,
   LogSymbolUse("Marked use of include-file", symbol_uses_.back());
 }
 
+void IwyuFileInfo::ReportIncludedFileMacroUse(const FileEntry* included_file) {
+  kept_includes_.insert(included_file);
+}
+
 void IwyuFileInfo::ReportPragmaKeep(const clang::FileEntry* included_file) {
   kept_includes_.insert(included_file);
 }
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -609,11 +609,7 @@ void IwyuFileInfo::ReportIncludeFileUse(const clang::FileEntry* included_file,
   LogSymbolUse("Marked use of include-file", symbol_uses_.back());
 }
 
-void IwyuFileInfo::ReportIncludedFileMacroUse(const FileEntry* included_file) {
-  kept_includes_.insert(included_file);
-}
-
-void IwyuFileInfo::ReportPragmaKeep(const clang::FileEntry* included_file) {
+void IwyuFileInfo::ReportKnownDesiredFile(const FileEntry* included_file) {
   kept_includes_.insert(included_file);
 }
 
--- a/iwyu_output.h
+++ b/iwyu_output.h
@@ -235,9 +235,13 @@ class IwyuFileInfo {
                            const string& symbol);
   // TODO(dsturtevant): Can we determine in_cxx_method_body? Do we care?
 
+  // Called when using a macro in this file.
   void ReportMacroUse(clang::SourceLocation use_loc,
                       clang::SourceLocation dfn_loc,
                       const string& symbol);
+
+  // Called when somebody uses a macro defined in this file.
+  void ReportDefinedMacroUse(const clang::FileEntry* used_in);
 
   // We only allow forward-declaring of decls, not arbitrary symbols.
   void ReportForwardDeclareUse(clang::SourceLocation use_loc,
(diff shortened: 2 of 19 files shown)


iwyu_preprocessor.cc
```diff
@@ -772,25 +772,50 @@ void IwyuPreprocessorInfo::ReportMacroUse(const string& name,
                                           SourceLocation dfn_location) {
   const FileEntry* used_in = GetFileEntry(usage_location);
```
kimgr: You can move this down below the `dfn_location` check now.

iwyu_preprocessor.cc
```diff
@@ -772,25 +772,50 @@ void IwyuPreprocessorInfo::ReportMacroUse(const string& name,
...
+    // I think the solution is to have a 'soft' use -- don't remove it
+    // if it's there, but don't add it if it's not.  Or something.
+    GetFromFileInfoMap(used_in)->ReportMacroUse(usage_location, dfn_location,
+                                                name);
+  }
+  const FileEntry* defined_in = GetFileEntry(dfn_location);
+  const SourceLocation include_loc = GlobalSourceManager()->getIncludeLoc(
+      GlobalSourceManager()->getFileID(usage_location));
+  const FileEntry* use_includer = GetFileEntry(include_loc);
+  bool is_macro_defined_by_includer = (defined_in == use_includer);
```
kimgr: I'd inline the expression in the if instead of declaring a new bool. This is self-documentation in some sense, but it's already pretty obvious what the expression means.

iwyu_preprocessor.cc
```diff
@@ -772,25 +772,50 @@ void IwyuPreprocessorInfo::ReportMacroUse(const string& name,
...
+    if (ShouldReportIWYUViolationsFor(defined_in)) {
+      GetFromFileInfoMap(use_includer)->ReportIncludedFileMacroUse(used_in);
+      ERRSYM(defined_in) << "Keep #include " << used_in->getName()
+                         << " in " << defined_in->getName()
+                         << " because macro " << name
+                         << " is defined by includer.\n";
+    } else {
+      string private_include = ConvertToQuotedInclude(
+          GetFilePath(usage_location));
+      string public_include = ConvertToQuotedInclude(GetFilePath(dfn_location));
```
kimgr: It should be slightly more efficient to `GetFilePath` on the `FileEntry` pointers you've already resolved above.

iwyu_output.cc
```diff
@@ -609,6 +609,10 @@ void IwyuFileInfo::ReportIncludeFileUse(const clang::FileEntry* included_file,
   LogSymbolUse("Marked use of include-file", symbol_uses_.back());
 }
 
+void IwyuFileInfo::ReportIncludedFileMacroUse(const FileEntry* included_file) {
```
kimgr: I think it would be nicer to collapse this and `ReportPragmaKeep` into one more general concept, something like `ReportKnownDesiredFile`.

kimgr:
Nicely done! I'm glad we didn't go with my heuristic solution :-)

Tests pass fine on Windows and FreeBSD.

I ran this on the original test case from @sylvestre in #109, and it still suggests adding `glib/gmem.h`. The most pressing issue is fixed though -- the file still compiles after applying IWYU fixes. I think this is because it keeps `glib.h`, and so `glib/gmem.h` is already included and is eventually skipped.

But it would be nice if IWYU could avoid suggesting it...

vsapsai:
Agree with all comments, will do corresponding changes.

Need to investigate further the problem with actual glib. Looks like the problem is that `glib/gmem.h` is first encountered not through include in `glib.h` but in some other file. For example, `glib/glist.h` itself includes `glib/gmem.h` and `glib.h` includes `glib/glist.h` before `glib/gmem.h`.

vsapsai:
Update: I have also discovered that with my change IWYU triggers assertion while analysing `iwyu_preprocessor.cc`:

> ...llvm/tools/clang/tools/include-what-you-use/iwyu_include_picker.cc:986: Assertion failed: filepath_visibility_map_[quoted_filepath_pattern] == vis Same file seen with two different visibilities: <limits.h> Old vis: 1 New vis: 2

vsapsai:
Fixed discovered issues, didn't test with actual glib yet.

Not sure if test for nested guarded includes (`macro_defined_by_includer-g3.h `) is actually required. It's good to check it works but there is no indication such pattern exists in real code.

iwyu_output.h
```diff
@@ -235,10 +235,14 @@ class IwyuFileInfo {
                            const string& symbol);
   // TODO(dsturtevant): Can we determine in_cxx_method_body? Do we care?
 
+  // Called when using a macro in this file.
   void ReportMacroUse(clang::SourceLocation use_loc,
                       clang::SourceLocation dfn_loc,
                       const string& symbol);
 
+  // Called when somebody uses a macro defined in this file.
+  void ReportDefinedMacroUse(const clang::FileEntry* used_in);
```
kimgr: It bugs me a little that you can derive `used_in` already from the arguments to `ReportMacroUse`, but I see that you only call `ReportMacroUse` if `ShouldReportIWYUViolationsFor` is true, so it's hard to unify this.

kimgr:
Great! I ran this on my platforms with good results.

I also tested the original `glib` problem from #109, and it now works as expected! Unfortunately, it still suggests a change `<glib.h>` -> `"glib.h"`, but that's a separate problem, so let's skip it for now and hope we can get to a point where we identify headers based on `FileEntry*` sometime soon.

Please squash this into a single commit and push!

vsapsai:
Thanks for reviewing, Kim.

PR #289, opened 2016-04-11
Allow ommitting -I.

Flamefire:
This is a split-off of #276 

It allows tests without the `-I .` which is (wrongly) assumed to be implicit.
It does not change the code or tests, so no functional changes. It just allows tests without the `-I .`

Note: I would even go so far and remove the defaulted `-I .` completely so it has to be given explicitly.

### DIFF
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -99,17 +99,77 @@ class OneIwyuTest(unittest.TestCase):
       'lambda_fwd_decl.cc': ['-std=c++11'],
       'lateparsed_template.cc': ['-fdelayed-template-parsing'],
       'ms_inline_asm.cc': ['-fms-extensions'],
-      'prefix_header_attribution.cc': [
-          self.Include('prefix_header_attribution-d1.h')],
+      'prefix_header_attribution.cc': [self.Include('prefix_header_attribution-d1.h')],
       'prefix_header_includes_add.cc': prefix_headers,
       'prefix_header_includes_keep.cc': prefix_headers,
       'prefix_header_includes_remove.cc': prefix_headers,
+    }
+    include_map = {
+      'alias_template.cc': ['.'],
+      'array.cc': ['.'],
+      'associated_h_file_heuristic.cc': ['.'],
+      'associated_include.cc': ['.'],
+      'backwards_includes.cc': ['.'],
+      'badinc.cc': ['.'],
+      'casts.cc': ['.'],
+      'catch.cc': ['.'],
+      'clmode.cc': ['.'],
+      'comment_pragmas.cc': ['.'],
+      'computed_include.cc': ['.'],
+      'conversion_ctor.cc': ['.'],
+      'default_template_arg_other_file.cc': ['.'],
+      'depopulated_h_file.cc': ['.'],
+      'derived_function_tpl_args.cc': ['.'],
+      'double_include.cc': ['.'],
+      'elaborated_struct.c': ['.'],
+      'elaborated_type.cc': ['.'],
+      'forward_declare_in_macro.cc': ['.'],
+      'fullinfo_for_templates.cc': ['.'],
+      'fwd_decl_class_template.cc': ['.'],
+      'fwd_decl_static_member.cc': ['.'],
+      'fwd_decl_with_instantiation.cc': ['.'],
+      'implicit_ctor.cc': ['.'],
+      'include_with_using.cc': ['.'],
+      'iwyu_stricter_than_cpp.cc': ['.'],
+      'keep_mapping.cc': ['.'],
+      'lateparsed_template.cc': ['.'],
+      'macro_location.cc': ['.'],
+      'member_expr.cc': ['.'],
+      'multiple_include_paths.cc': ['.'],
+      'no_comments.cc': ['.'],
+      'no_h_includes_cc.cc': ['.'],
+      'non_transitive_include.cc': ['.'],
+      'overloaded_class.cc': ['.'],
+      'pch_in_code.cc': ['.'],
+      'precomputed_tpl_args.cc': ['.'],
+      'prefix_header_attribution.cc': ['.'],
+      'prefix_header_includes_add.cc': ['.'],
+      'prefix_header_includes_keep.cc': ['.'],
+      'prefix_header_includes_remove.cc': ['.'],
+      're_fwd_decl.cc': ['.'],
+      'redecls.cc': ['.'],
+      'remove_fwd_decl_when_including.cc': ['.'],
+      'sizeof_reference.cc': ['.'],
+      'specialization_needs_decl.cc': ['.'],
+      'system_namespaces.cc': ['.'],
+      'template_args.cc': ['.'],
+      'templated_constructor.cc': ['.'],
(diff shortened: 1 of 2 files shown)


kimgr:
> Note: I would even go so far and remove the defaulted -I . completely so it has to be given explicitly.

Yep, I think you should do that as part of this change, otherwise the whole change seems pointless :-)

Now that we're going down this path, should the long-term goal be to remove all `tests/cxx` prefixes from test includes?

I did a quick grep, and I see 469 includes with `tests/` and 35 without, so it's often used, but not universally.

Flamefire:
Ok will do.

I think this is a good idea. It would make the test suite cleaner after the `-I .` is no longer the default. However this must be after #276 as otherwise all tests will fail.

kimgr:
Oh, OK, I didn't expect that, but I'm beginning to get a feel for why this has been a problem for so long... Thanks!

Flamefire:
Jep, IWYU internally acts if this flag was passed. So it will not care that it was removed and produce the same output, as if it was there...
BTW: I would suggest to include a compile test with the fixed tests. So in addition to running iwyu we should invoke the compiler on the fixed test case with the flags specified in the test suite. Or at least add another tester (or flag), that does this. This way one can detect, that omitting the `-I. ` and following the (current) iwyu suggestions will result in source code that no longer compiles.

vsapsai:
Regarding removing `tests/cxx` prefix. Does majority of [multi-directory] C++ projects you've seen use relative includes or includes discovered through `-I`? Because I prefer tests to mimic common practice.

Flamefire:
I'm not sure about that. I think it is common practice to add the projects root include directory to the search paths. But it may also be mixed with relative includes (especially for the corresponding header)

However a `-I .` is mostly not what you want as you normally build in a separate build directory so the `-I .` would refer to the build rather than the src directory. But for these tests it should be ok.

Flamefire:
I removed the default and use a new map of includes. It is pretty long...

run_iwyu_tests.py
```diff
@@ -99,17 +99,77 @@ def setUp(self):
...
       'lateparsed_template.cc': ['-fdelayed-template-parsing'],
       'ms_inline_asm.cc': ['-fms-extensions'],
-      'prefix_header_attribution.cc': [
-          self.Include('prefix_header_attribution-d1.h')],
+      'prefix_header_attribution.cc': [self.Include('prefix_header_attribution-d1.h')],
       'prefix_header_includes_add.cc': prefix_headers,
       'prefix_header_includes_keep.cc': prefix_headers,
       'prefix_header_includes_remove.cc': prefix_headers,
     }
+    include_map = {
```
kimgr: So will we be able to shrink this as your actual IWYU patches come in? (hoping!)

run_iwyu_tests.py
```diff
@@ -139,6 +199,13 @@ def RunOneTest(self, filename):
     clang_flags = self._clang_flags_map.get(filename, None)
     if clang_flags:
       logging.info('%s: Using clang flags %s', filename, str(clang_flags))
+    include_flags = self._include_map.get(filename, None)
+    if include_flags:
+      logging.info('%s: Using include flags %s', filename, str(include_flags))
```
kimgr: I think you can skip this, and move the `if clang_flags:` logging down after you've decorated the clang flags with include flags.

run_iwyu_tests.py
```diff
@@ -139,6 +199,13 @@ def RunOneTest(self, filename):
     clang_flags = self._clang_flags_map.get(filename, None)
```
kimgr: Default to `[]` here instead of `None`...

run_iwyu_tests.py
```diff
@@ -139,6 +199,13 @@ def RunOneTest(self, filename):
...
     if clang_flags:
       logging.info('%s: Using clang flags %s', filename, str(clang_flags))
+    include_flags = self._include_map.get(filename, None)
+    if include_flags:
+      logging.info('%s: Using include flags %s', filename, str(include_flags))
+      if clang_flags:
```
kimgr: ... and you can use `.extend` unconditionally here.

kimgr:
Looks good to me, barring a few nits!

I think it's a clear improvement to be explicit about the `.`-dependency; hopefully we can remove it in short order, and reduce the list...

I did an experiment and tried to remove all `tests/cxx` prefixes from the tests (a quick sed exercise), but there were plenty of failures because IWYU's diff result output still used `tests/cxx`-qualified paths. I'm guessing this is the bug you're ultimately trying to fix?

kimgr:
Also @vsapsai: I agree that we should have tests in place for prefixed includes, but I think we should have focused tests for that specific case, not have it reflected in all tests.

To my mind, using a project-specific prefix is the right way to go about things, independent of what people generally do :-), but I don't see a reason for IWYU to have a strong opinion here.

run_iwyu_tests.py
Flamefire: Yes :)

run_iwyu_tests.py
Flamefire: I wanted to be explicit about the includes so one can easier see, where include paths were set.

Flamefire:
Yes your sed fix won't work as IWYU internally assumes the `-I.` no matter if it is there or not.

We may remove some, but we also need tests that have this option as we need to test both paths. So I would not rush removing this list. I rather though about a plain list of "basePathIncludes" where all files get listed, that should get only the `-I.` include so the list is much less verbose... But whatever... This is the test setup, so we can also be explicit.

Flamefire:
I changed the default for clang_flags as per your suggestion. I'd still keep the extra output for the includes as (especially when they are reduced) it is good to know, when and what includes are used at runtime.

kimgr:
I disagree about the `include_flags` output, it's not interesting in itself, and it's trivial to read it from the `clang_flags` output (anything -I-prefixed will be from there.)

Unless @vsapsai has more questions or comments, I can squash this and commit it when I get a chance.

Flamefire:
Ok alright. Makes the code more compact, I added that and squashed the changes

vsapsai:
As far as I understand, there is non-trivial difference between relative headers and includes discovered via `-I`. That's why I want tests by default to execute case that is prevalent in real code. And I want to have actual usage data, not anecdotal evidence and personal opinions. I value your opinion but I cannot tell it's representative.

I had in mind something like taking 50 most popular C++ projects on GitHub and checking how many of those projects use relative includes. Feel free to prepare this data, it will take me some time to get to it.

Flamefire:
Why is this important for this PR? In the end both ways need to work and this PR enables that. I think the discussion and decision what should be the default (overwritable) for the tests can be delayed to a later PR. Or even make the tests so generic, that (at least some) can be run with and without `-I .` (need some adjustments in the output checks which might not be completely trivial)

kimgr:
@vsapsai: I think this is a side discussion irrelevant to this PR, so I'll just go ahead and commit this.

Given that it allows starting work on #5 and friends, I think it's a strict improvement.

kimgr:
I rephrased the commit message, so this probably came in with a different SHA1, but this is now in master. Thanks for working on this!

PR #304, opened 2016-05-17
Remove CanonicalizeFilePath and add NormalizeFolderPath

Flamefire:
As discussed in #276 (https://github.com/include-what-you-use/include-what-you-use/pull/276#discussion_r63248447, https://github.com/include-what-you-use/include-what-you-use/pull/276/files#r63247740) the `CanonicalizeFilePath` is redundant and `NormalizeFilePath` should be used. Additionally we also need a function that adds the trailing slash to paths so compares for path suffixes are done correctly. This is now `NormalizeFolderPath`

### DIFF
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -300,7 +300,7 @@ static vector<HeaderSearchPath> ComputeHeaderSearchPaths(
            it = header_search->system_dir_begin();
        it != header_search->system_dir_end(); ++it) {
     if (const DirectoryEntry* entry = it->getDir()) {
-      const string path = CanonicalizeHeaderSearchPath(entry->getName());
+      const string path = NormalizeFolderPath(entry->getName());
       search_path_map[path] = HeaderSearchPath::kSystemPath;
     }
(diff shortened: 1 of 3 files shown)


kimgr:
This looks good to me! One nit: I tend to use the POSIX terminology for all file- and path-related things (see terminology here: http://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap03.html), so I'd prefer if you renamed `NormalizeFolderPath` -> `NormalizeDirPath` ('directory' tends to become too long).

I'll pull down and test your patch on my various systems to check for portability issues.

Thanks!

Flamefire:
Alright. Feel free to make that change before merging if you got the time. Otherwise I'll add that once I got time.

kimgr:
Thanks! Pushed a variant of this in 6cf6b76a4390a3c49bd50a55d640dc971f8c63d3.

I realized that `NormalizeFilePath` could be flattened since `std::replace` works on `SmallString` too.

PR #305, opened 2016-05-19
Install mapping files in share/include-what-you-use

EugeneZelenko:
Make mapping files  available for users who doesn't build include-what-you-use themselves.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,3 +180,7 @@ endif()
 
 install(TARGETS include-what-you-use RUNTIME DESTINATION bin)
 install(PROGRAMS fix_includes.py iwyu_tool.py DESTINATION bin)
+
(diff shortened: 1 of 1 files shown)


kimgr:
Thanks, looks good. I'm a bit of a CMake dummy, can you share the command for using this install step? Is CMake good enough to generate it for the respective build systems (I'm using ninja) or does it involve a `cmake` call?

EugeneZelenko:
I could not tell that I CMake expert too :-) I just borrowed bits of install from clang-tools-extra and Clazy (as in this case).

vsapsai:
Looks good to me. I'm not going to use this feature so I cannot tell how usable it is. I'll defer to others to decide if it solves their problems.

kimgr:
Pushed, thanks!

Issue #141, opened 2015-06-15
mappings for libc++ are missing


vsapsai:
Originally reported on Google Code with ID 141

```
IWYU doesn't come with default mappings for libc++ (the llvm stl implementation)

Example:
==> stringtest.cc <==
#include <string>
int main() {
    std::string* p = new std::string();
}

Compling as:
include-what-you-use stringtest.cc -stdlib=libc++ -I$LIBCXX/include

Gives:
stringtest.cc should add these lines:
#include "iosfwd"                       // for string
#include "new"                          // for operator new

To test this libc++ doesn't need to be compiled (nor are any dependencies required).
LIBCXX needs to be set to the path were the source folder is (or the install folder
if it is compiled&installed).
```

Reported by `roland@rschulz.eu` on 2014-05-31 17:36:18

EugeneZelenko:
I tried to run 3.8 release on trunk LLVM/Clang/LLDB built with libc++, so I got next false positives:

#include <__functional_base>     // for less
#include <iosfwd>                        // for string
#include <type_traits>                  // for move
#include <__mutex_base>           // for lock_guard

EugeneZelenko:
Problem is a little bit wider. At least less is used as default value for template parameter in map and set. Both includes functional explicitly. So functional should be suggested for inclusion only if less (or other template) is used in analyzed source file explicitly. So for often appeared type_traits/remove_reference.

chri7325:
I'm also hitting an issue now running IWYU with clang 9 using libc++ exclusively.  On some std types, it tries forward declaring them like so
```
namespace std
{
namespace __1
{
template <class _Tp>
struct __attribute__((type_visibility("default"))) hash;
} // namespace __1
} // namespace std
```
This produces a compiler error `error: inline namespace reopened as a non-inline namespace [-Werror]
namespace __1`

I've tried adding a mapping for this like so
```
{ symbol: [ "std::__1::hash", private, "<functional>", public ] },
{ symbol: [ "std::hash", private, "<functional>", public ] },
{ symbol: [ "std::hash<_Tp>", private, "<functional>", public ] },
```

Unfortunately none of these work. I'm thinking of possibly putting an iwyu pragma: keep on the functional header but I'm wondering if there's a better way to do this?

kimgr:
[...]

PR #307, opened 2016-05-20
Fix some LLVM STL false positives

EugeneZelenko:
Fix issues I reported in #141.

### DIFF
--- a/llvm.stl.headers.imp
+++ b/llvm.stl.headers.imp
@@ -1,0 +1,7 @@
+# LLVM STL headers
+[
+  { include: ["<__functional_base>", private, "<functional>", public ] },
+  { include: ["<__mutex_base>", private, "<mutex>", public ] },
+  { symbol: [ "std::move", private, "<algorithm>", public ] },
+  { symbol: [ "std::string", private, "<string>", public ] },
+]


kimgr:
Thanks! I see you've followed suit and named it after the larger toolchain (LLVM) instead of the standard library (libc++). I think that was a mistake on my part when I added the `gcc.*` mappings, I didn't fully understand the connection between GCC, libstdc++ and GNU libc at that point (not sure I do still...)

Anyway, I think this would be better named `libcxx.imp`

I know @vsapsai has done some work earlier to have IWYU principally detect patterns in libc++ and report them correctly without mappings. That's of course a much better way forward, but I think this is a good stepping stone to let people use IWYU even if it doesn't really work :-)

Also, I wonder if we should add this to IWYU's builtin mappings in `iwyu_include_picker.cc`. Apple's libc++ seems like a major platform, so it would be nice to support it natively without mappings.

EugeneZelenko:
I think will be good idea to solve libc++ mapping problem ASAP (ether internally or as mapping file). Number of bug reports indicate that it is long overdue feature.

vsapsai:
I see people are eager to help fix handling libc++. I'll try to summarise various issues, clean up my investigation notes and post them.

kimgr:
I'm going to clean this up and commit it. It opens up for people to IWYU with libc++ at their own discretion. I think we can and should do real improvements separately.

kimgr:
Adapted and pushed as 2205f1ac6a2b11352ba1917a9307889a600a0b8f.

PR #308, opened 2016-05-20
Fix Clang-tidy modernize warnings

EugeneZelenko:
Simplify header guard name, since Include What You Use is standalone tool and name is unique enough.

### DIFF
--- a/iwyu_preprocessor.h
+++ b/iwyu_preprocessor.h
@@ -55,8 +55,8 @@
 //      of the information about public vs private headers.
 //    * Testing and reporting membership in the main compilation unit.
 
-#ifndef DEVTOOLS_MAINTENANCE_INCLUDE_WHAT_YOU_USE_IWYU_PREPROCESSOR_H_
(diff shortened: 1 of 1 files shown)


kimgr:
Looks good to me. I'd prefer if we changed the header guards across all files in one fell swoop instead of changing just the one here. Can you remove that part and save it for a later sed exercise?

EugeneZelenko:
I'm planning to run Clang-tidy over all code base, so I'll fix inclusion guard gradually. It doesn't make sense to change only inclusion guards only.

kimgr:
Oh, the header guard change comes from Clang-tidy too? In that case I agree.

kimgr:
... I just noticed that all the tests use the same guard-style. I don't think we want to tidy the test suite, so let's update all include guards separately anyway. I'll prepare a PR for it.

kimgr:
Closed as of d9f88454ca3877bf89549949ac6bfde93c4853cb.

Rebased, fixed header guard and edited commit message.

PR #311, opened 2016-05-22
Remove DEVTOOLS_MAINTENANCE_ from header guards

kimgr:
@EugeneZelenko did this as part of a general Clang-tidy cleanup in PR #308 , but I'd prefer to get the tests and production code fixed in a single run.

Clang-tidy shouldn't need to complain once this is in place.


vsapsai:
Everything is still working after this change.

EugeneZelenko:
Please also remove trailing underscores. See LLVM/Clang/Clang-tidy headers as example.

vsapsai:
We are following [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#The__define_Guard).
> All header files should have `#define` guards to prevent multiple inclusion. The format of the symbol name should be `<PROJECT>_<PATH>_<FILE>_H_`.

EugeneZelenko:
Since this project is based on LLVM/Clang, it's reasonable to follow that standard.

kimgr:
Yes, but nothing else in IWYU follows that standard. It seems better to be self-consistent.

kimgr:
Added two commits to make header guards completely consistent (one to fix a couple of mismatched guards/filenames and one to rename all tests header guards to match the tests/cxx path.)

kimgr:
This seemed non-controversial, so I pushed the changes.

PR #322, opened 2016-05-30
Fix in tests "fatal error: file not found", make test suite detect such issue.

vsapsai:


kimgr:
Why was this necessary? It's surprising to me that the regex didn't already match, i.e. why doesn't "error" match "fatal error" too?

vsapsai:
There is no comma on the previous line
```python
_ACTUAL_DIAGNOSTICS_RE = re.compile(r'^(.*?):(\d+):\d+:\s*'
                                    r'(?:warning|error):\s*(.*)$')
```

so detected part is basically `:\s*(warning|error):`, i.e. "warning" or "error" between colons. That's why "fatal error" didn't match.

vsapsai:
Ping.

kimgr:
Oh, OK, makes sense! I'll just test this on my environments once I've built it all again...

kimgr:
Tests run fine on FreeBSD and Windows.

I reverted half your patch (the changes to run_iwyu_tests.py), and now I understand what bug you're fixing. The 62bc551a commit apparently broke lots of tests since the runner didn't fail when include files were not found. Might be worth mentioning that in the commit message.

Other than that, looks good, let's do it!

vsapsai:
Thanks for review. Updated commit message, pushed as d5cfaf0db3746622738fc36ec2be7a7eff5dadab.

PR #328, opened 2016-06-06
Handle internal headers guarded by macro and x-macros (fix issue #109).

vsapsai:
I've made a few trade-offs in implementation which I'd like to explain more.

First, code doesn't distinguish between guarded internal headers and
headers with x-macros. They are handled the same way. But in tests both
patterns are tested. It is done not to cover all code paths but to test
include-what-you-use from user's perspective.

Also I check if file defining macro is immediate includer. I decided not
to check if it includes file using the macro transitively until we have
such real-life use cases. Current implementation is strict in order to avoid
unexpected results.

For some cases I am reusing mechanism that keeps files included with the
"IWYU pragma: keep" comment. The downside is that it keeps all lines
including this file which might be not entirely correct for x-macros.
Though x-macros are close to pure textual includes and we cannot reason
about textual includes.

I also didn't include a few test cases because I think they don't
represent real-life use cases. These test cases are:
* when associated header is included by non-associated header;
* when file defining macro and file using macro are both included by the
  third file.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -602,6 +602,10 @@ void IwyuFileInfo::ReportMacroUse(clang::SourceLocation use_loc,
   LogSymbolUse("Marked full-info use of macro", symbol_uses_.back());
 }
 
+void IwyuFileInfo::ReportDefinedMacroUse(const clang::FileEntry* used_in) {
+  macro_users_.insert(used_in);
+}
+
 void IwyuFileInfo::ReportIncludeFileUse(const clang::FileEntry* included_file,
                                         const string& quoted_include) {
   symbol_uses_.push_back(OneUse("", included_file, quoted_include,
@@ -609,7 +613,7 @@ void IwyuFileInfo::ReportIncludeFileUse(const clang::FileEntry* included_file,
   LogSymbolUse("Marked use of include-file", symbol_uses_.back());
 }
 
-void IwyuFileInfo::ReportPragmaKeep(const clang::FileEntry* included_file) {
+void IwyuFileInfo::ReportKnownDesiredFile(const FileEntry* included_file) {
   kept_includes_.insert(included_file);
 }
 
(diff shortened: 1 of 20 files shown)


vsapsai:
This time tested on Mac OS X and Ubuntu. And tested on example from issue #109 - IWYU tells file has correct #includes/fwd-decls.

iwyu_output.cc
```diff
@@ -1889,6 +1893,41 @@ size_t PrintableDiffs(const string& filename,
...
+  bool should_report_violations = ShouldReportIWYUViolationsFor(file_);
+  std::list<const FileEntry*> direct_macro_use_includees;
+  std::set_intersection(macro_users_.begin(), macro_users_.end(),
+                        direct_includes_as_fileentries_.begin(),
+                        direct_includes_as_fileentries_.end(),
+                        std::inserter(direct_macro_use_includees,
+                                      direct_macro_use_includees.end()));
+  for (const FileEntry* macro_use_includee : direct_macro_use_includees) {
+    if (should_report_violations) {
+      ERRSYM(file_) << "Keep #include " << macro_use_includee->getName()
```
kimgr: OK, so the mere presence of a dependency on a includer-defined macro now forces us to keep the include?

I think that may be too strong, e.g. for 'control' macros, something like:

     #define XYZ_ENABLE_FOOBAR
     #include "unused.h"

where `unused.h` has something like:

    #ifdef XYZ_ENABLE_FOOBAR
    inline void foobar() {
    }
    #endif

(practical example: `__STDC_LIMIT_MACROS` and `__STDC_CONSTANT_MACROS` vs `stdint.h`).

That said, in practice I think this is fine. It will produce slightly less optimal IWYU output (for a narrow case), but should always produce code that can compile.

iwyu_output.h
```diff
@@ -256,15 +260,19 @@ class IwyuFileInfo {
   void ReportIncludeFileUse(const clang::FileEntry* included_file,
                             const string& quoted_include);
 
-  // This is used when we see an "IWYU pragma: keep" comment
-  // on an include line.
-  void ReportPragmaKeep(const clang::FileEntry* included_file);
+  // This is used when we see a file we want to keep not due to symbol-use
+  // reasons.  For example, it can be #included with an "IWYU pragma: keep"
+  // comment or it can be a x-macro.
```
kimgr: a*n* x-macro

kimgr:
One small nit inline (and a question), but overall I think this looks great!

Tests run clean on both FreeBSD and Windows/MSVC. Let's get this committed!

iwyu_output.cc
vsapsai: You are correct, 'control' macros can cause keeping unnecessary includes. Some cases are somewhat alleviated by putting these macros and includes in non-associated header like `config.h`.

One way to improve the situation is to distinguish 'control' macros, macros that occur only in #if, #ifdef, etc. This way we'll be able to tell if content of included file is required to be used to keep it. Because for x-macros we still want to keep #include even if it looks like it is unused.

vsapsai:
Pushed the updated commit. It was a forced push, so if you checked out PR locally, just pulling won't work.

kimgr:
Are we waiting on anything here? I'd love to get this merged before our next (imminent) release.

vsapsai:
I responded to your comment in June and didn't finalize this PR. Pushed 04f1c925370d0f1a70642147f07c70a4c2bfb530 minutes ago.

kimgr:
Yay!

PR #333, opened 2016-06-21
Allow iwyu_tool.py to accept source paths with symbolic links

EugeneZelenko:


kimgr:
I think `os.realpath` already does the equivalent of `os.abspath`, can you try removing `abspath`?

EugeneZelenko:
I think will be good idea to ask somebody who knows Python better then me.

kimgr:
I did a few tests, and `os.path.realpath` indeed appears to return absolute paths. I'll fix it up and pushd, thanks!

kimgr:
Fixed in 0341c6c0954ec6c36f2c15b8c3eec4b937039deb.

PR #339, opened 2016-07-13
Fix some Clang-tidy warnings

EugeneZelenko:
I run Clang-tidy with misc checks without  misc-unused-parameters and readability checks without readability-identifier-naming, readability-else-after-return, readability-braces-around-statements, readability-implicit-bool-cast, readability-named-parameter. readability-avoid-const-params-in-decls.

I also fixed modernize-loop-convert and modernize-use-emplace.

Most serious problem was found with misc-suspicious-missing-comma.

Regressions were OK on RHEL6.


EugeneZelenko:
Please ignore change in  iwyu_output.cc.

iwyu_path_util.cc
```diff
@@ -36,7 +37,7 @@ const char* source_extensions[] = {
   ".cxx",
   ".CXX",
   ".cpp",
-  ".CPP"
+  ".CPP",
```
kimgr: Ouch! I remember reading about this check and thinking it looked cool, but I never thought I'd put in bugs for it to find :-)

kimgr:
Preliminarily looks good. I'll take a closer look tonight and try to merge.

iwyu_include_picker.cc
```diff
@@ -1414,14 +1416,13 @@ void IncludePicker::AddMappingsFromFile(const string& filename,
       return;
     }
 
-    for (MappingNode::iterator it = mapping->begin();
-        it != mapping->end(); ++it) {
+    for (KeyValueNode &node : *mapping) {
```
vsapsai: I don't like shadowing `node` variable in outer loop. What if we name variable in outer loop `array_item_node` and this one `mapping_item_node`?

iwyu_include_picker.h
```diff
@@ -89,14 +89,13 @@ class IncludePicker {
...
-  void MarkIncludeAsPrivate(const string& quoted_include);
+  void MarkIncludeAsPrivate(const string& quoted_filepath_pattern);
 
   // Add this to say that "any file whose name matches the
   // friend_regex is allowed to include includee_filepath".  The regex
   // uses the POSIX Entended Regular Expression syntax and should
   // match a quoted-include (starting and ending with "" or <>).
-  void AddFriendRegex(const string& includee_filepath,
-                      const string& quoted_friend_regex);
+  void AddFriendRegex(const string& includee, const string& friend_regex);
```
vsapsai: Here I would change argument names in .cc file to be the same as in .h as names in .h are more specific.

iwyu_output.h
```diff
@@ -213,7 +215,7 @@ class IwyuFileInfo {
                   const string& quoted_includee, int linenumber);
   // definitely_keep_fwd_decl tells us that we should never suggest
   // the fwd-decl be removed, even if we don't see any uses of it.
-  void AddForwardDeclare(const clang::NamedDecl* forward_declare_decl,
+  void AddForwardDeclare(const clang::NamedDecl* fwd_decl,
```
vsapsai: Here I think it is better to sync implementation with declaration, not vice versa. Mostly because I think that this argument is not forward declaration per se but a NamedDecl that can be forward declared.

iwyu_preprocessor.h
```diff
@@ -202,13 +203,13 @@ class IwyuPreprocessorInfo : public clang::PPCallbacks,
...
                    clang::SrcMgr::CharacteristicKind file_type,
-                   clang::FileID PrevFID) override;
+                   clang::FileID exiting_from) override;
   void FileSkipped(const clang::FileEntry& file, const clang::Token &filename,
                    clang::SrcMgr::CharacteristicKind file_type) override;
   // FileChanged is actually a multi-plexer for 4 different callbacks.
   void FileChanged_EnterFile(clang::SourceLocation file_beginning);
   void FileChanged_ExitToFile(clang::SourceLocation include_loc,
-                              clang::FileID exiting_from);
+                              clang::FileID exiting_from_id);
```
vsapsai: Can we please make names consistent? So it is either `exiting_from` or `exiting_from_id` in both places.

vsapsai:
Added my comments. Tested on Mac OS X, no regressions.

iwyu_include_picker.h
```diff
@@ -80,7 +80,7 @@ class IncludePicker {
   // perhaps.
   void AddDirectInclude(const string& includer_filepath,
                         const string& includee_filepath,
-                        const string& quoted_include_as_written);
+                        const string& quoted_include_as_typed);
```
kimgr: I think `quoted_include_as_written` was more 'right' here.

iwyu_include_picker.h
```diff
@@ -186,11 +186,12 @@ class IncludePicker {
   void AddImplicitThirdPartyMappings();
 
   // Adds an entry to filepath_visibility_map_, with error checking.
-  void MarkVisibility(const string& quoted_include, IncludeVisibility vis);
+  void MarkVisibility(const string& quoted_filepath_pattern,
+                      IncludeVisibility vis);
```
kimgr: Now that the line is broken, I think we can afford to spell out `visibility`.

kimgr:
A few more comments from me, I can run tests tomorrow.

iwyu_include_picker.h
EugeneZelenko: In this case includer_and_includee_to_include_as_typed_ member and quoted_include_as_typed local variable should be renamed too.

kimgr:
Builds and tests pass on Windows and FreeBSD too.

kimgr:
I'll take ownership of this and commit it in pieces -- I'd like to commit the bugfix separately.

kimgr:
There, I think I got all of them. I split the various fixes up into individual commits.

Issue #332, opened 2016-06-17
IWYU removes an include which is required for pointer arithmetic


Bothari:
If a type is used in a file for pointer arithmetic, the header file defining that type still needs to be included. See attached code for an example. 

` 14:15 > ./include-what-you-use.exe c:/Users/mdyckhoff/iwyu-test/test.cpp`

`c:/Users/mdyckhoff/iwyu-test/test.cpp should add these lines:`
`class Foo;`

`c:/Users/mdyckhoff/iwyu-test/test.cpp should remove these lines:`
`- #include "foo.h"  // lines 2-2`

`The full include-list for c:/Users/mdyckhoff/iwyu-test/test.cpp:`
`class Foo;`
`---`

However if I perform this change, the file will not compile, with the error:

`c:/Users/mdyckhoff/iwyu-test/test.cpp:6:15: error: arithmetic on a pointer to an incomplete type 'Foo'
        return (pPtr + count);`

[EDIT: Adding version information. Running IWYU on Windows.

`include-what-you-use 0.4 (645) based on clang version 3.6.1 (branches/release_36 238234)`]

[test-remove-header-pointer-arith.zip](https://github.com/include-what-you-use/include-what-you-use/files/321281/test-remove-header-pointer-arith.zip)

Bothari:
I realised I am running a rather out of date version of IWYU (necessary because I'm integrating it with our toolchain which currently runs on clang 3.6). I have confirmed that the same error occurs with the latest Windows binary available:

`include-what-you-use 0.6 (git:f09b88a) based on clang version 3.8.0 (tags/RELEASE_380/final 269221)`

kimgr:
Thanks for the concise test case. It looks like IWYU doesn't handle binary operators at all -- this should be easily solved by adding a `VisitBinaryOperator` implementation that reports type use of both the left- and right-hand side of the binary op expression.

kimgr:
I did some hacking on this when it came up, then work came in the way.

It turned out to be slightly more complicated, because pointer arithmetic is only a narrow subset of all binary operator uses, I found a clang-tidy check that identifies it like in this patch:
http://reviews.llvm.org/D14582

I'll try to pick this up at some point, but I'm still in crunch.

PR #340, opened 2016-07-14
Classify pointer arithmetic as a full use

kimgr:
For pointer arithmetic, the compiler needs to know the size of the
pointed-to types.

Detect all valid forms of pointer arithmetic and report both the
right-hand and left-hand expression types.

This fixes issue #332

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -124,6 +124,7 @@
 #include "clang/AST/DeclBase.h"
 #include "clang/AST/DeclTemplate.h"
 #include "clang/AST/NestedNameSpecifier.h"
+#include "clang/AST/OperationKinds.h"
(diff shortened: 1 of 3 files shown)


kimgr:
I just realized that I can simplify the test case by putting all of it in a big function, and not saving the results of pointer arithmetic expressions. I'll follow up with that later.

vsapsai:
Overall looks good to me. In test I would add pointer arithmetic with built-in types like `char`, `int`.

kimgr:
Pushed a new commit with negative tests for builtins.

vsapsai:
Looks good to me.

vsapsai:
Merged squashed commits as 2141bde3be54724a5d31eca18ca5bd6ccc0df2d5.

Issue #325, opened 2016-06-02
It'd be nice if include what you use told you *why* it felt you needed a #include


ThosRTanner:
For instance, include what you use is telling me in a QT project I need QStringList.

However, I've examined the source code and the headers it references and I cannot see any way there's a need for QStringList (admittedly this is not helped by every QT header accidentally including every other QT header). 

If it could tell me what piece of code had lead it to the conclusion a file was necessary then either I could say 'ah yes', or I could file a report with some confidence.

Bu11etmagnet:
Don't examine the source code; that's not what the compiler (and iwyu) sees. Examine the preprocessed source.

kimgr:
I think that's why we have so much logging in IWYU -- most of the decision paths are spewed to the log. Run with `-Xiwyu -v7` and see if you can glean something from there.

ThosRTanner:
Ah. I've managed to interpret the output a bit better after skipping the ages of lexical analysis.

It's because there's a function prototype in the header that returns a `QStringList` which is being called by the source code in question, but then using `QStringList`'s `join` method to turn the result into a string.

So it's correct, but  the actual use is well hidden

This is at line 53:

    .... call_function_that_takes_QString(
                            info.arguments().join(" "),
    ....

Where info.arguments returns `QStringList`. That produces this with -v7

```
(For type class QStringList):
Marked full-info use of decl 0631DA00 QStringList (from C:/Apps/Qt/5.5/msvc2013/include/QTCore/qstringlist.h:93:7) at source/organizer/executableslist.cpp:53:46
source/organizer/executableslist.cpp:53:29: (14) [ ImplicitCastExpr ] 08BD64D0 ImplicitCastExpr 0x8bd64d0 <source\organizer\executableslist.cpp:53:29, col:44> 'const struct QListSpecialMethods<class QString>' <UncheckedDerivedToBase (QList -> QListSpecialMethods)>
`-CXXBindTemporaryExpr 0x8bd6418 <col:29, col:44> 'class QStringList' (CXXTemporary 0x8bd6410)
  `-CXXMemberCallExpr 0x8bd63f0 <col:29, col:44> 'class QStringList'
    `-MemberExpr 0x8bd63d0 <col:29, col:34> '<bound member function type>' .arguments 0x6639680
[...]

PR #341, opened 2016-07-16
Remove logging from FindArgumentsToDefined

kimgr:
This came up in issue #325, and I've always found this logging to be in the way.

If we think it might be useful, I could increase the log level instead.

### DIFF
--- a/iwyu_lexer_utils.cc
+++ b/iwyu_lexer_utils.cc
@@ -107,7 +107,7 @@ vector<Token> FindArgumentsToDefined(
   const string range_str(text, range_length);
   const char* range_cstr = range_str.c_str();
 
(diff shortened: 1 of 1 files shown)


kimgr:
OK, makes sense. Let me recast the patch.

kimgr:
Done, PTAL.

vsapsai:
Looks good to me. And it doesn't contradict to [verbosity description in `iwyu_globals.h`](https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_globals.h#L73).

vsapsai:
Merged in 0cfdff27dce040787499c77f01237b99482ea70e.

PR #342, opened 2016-07-16
Don't classify anonymous types as fwd-decls

### LINKED ISSUE #334: 3.8 recommends includes for things not used in file
Reported by Franz Schrober on http://bugs.debian.org/828734

It was noticed that the upgrade from 3.7 to 3.8 introduced a regression. iwyu
is now recommending includes for things which are nowhere used in a file.

First problem was that this version is not depending on libclang-common-3.8-dev
and thus didn't find some of the headers (see #722132). But it turns out that
this is not the only problems 

E.g it wants me to now include

```
#include <linux/compiler.h> // for __u, __read_once_size, __read_o...
#include "linux/swab.h" // for __fswab16
```

But these things are nowhere used in the file. The only appearances are inside
included headers.

I've tried to create a preprocessed file but it seems that iwyu doesn't report
any problems anymore when the file is already preprocessed. But I've created a
small example which shows the problem and how it related to macros. The test.c
file just uses the functionality from memops.h (this is the abstraction layer)
- stdio.h. memops.h itself is using compiler.h. No functionality from
  compiler.h is used directly in test.c

Compile with macros in memops.h:

```
    $ iwyu test.c -o test.o
    #include "compiler.h"

    #include <stdio.h> // for printf
    #include "compiler.h"
    #include "memops.h" // for get_memory_pointer_once
```

And now with the same functionality but implemented as static inline function:

```
    $ iwyu -DIWYU_INLINE_FUNCTION_TEST test.c -o test.o
    (test.c has correct #includes/fwd-decls)
```

So my guess is that the handling of macros is currently broken in iwyu/clang.

kimgr:
When looking for forward-decl hints in macro definition files, avoid
anonymous types.

This fixes #334.

Additional note: there should be other cases where this can break. I think the real fix is to find any declarations outside of the macro definition range, but I'm not entirely sure how to do that, so this is a good start.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1374,6 +1374,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
                                                SourceLocation loc) {
     set<const NamedDecl*> redecls = GetClassRedecls(DynCastFrom(decl));
     for (const NamedDecl* redecl : redecls) {
+      // Anonymous types can't be forward-decls.
+      if (redecl->getName().empty())
+        continue;
       if (IsBeforeInSameFile(redecl, loc)) {
         return redecl;
       }
--- a/tests/cxx/macro_location-i3.h
+++ b/tests/cxx/macro_location-i3.h
@@ -10,3 +10,22 @@
 #define DECLARE_FRIEND(cls)  friend class cls
 
 class Foo {};
+
+
+// Reduced example from the Linux kernel:
+// http://lxr.free-electrons.com/source/tools/include/linux/compiler.h#L112
+//
+// (The below test code is absolute nonsense, but it triggers the same IWYU
+// behavior as the actual READ_ONCE macro.)
+//
+// This makes sure we don't treat anonymous type declarations as forward-decls
+// in the context of macro use locations.
+#define ANONYMOUS_TYPE_IN_MACRO(value) \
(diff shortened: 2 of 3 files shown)


kimgr:
Oh. I was so excited I got this to work that I forgot about the test case. I'll get back with that some time soon.

kimgr:
There.

vsapsai:
The change looks good to me and it works on Mac OS X and Ubuntu.

Just to help me to understand the issue better. Do you need 2 `ANONYMOUS_TYPE_IN_MACRO` calls so you have 2 anonymous types which are considered to be redeclarations of each other? Or just 1 `ANONYMOUS_TYPE_IN_MACRO` is enough?

I also had and idea to add an assertion that `NamedDecl` name is not empty in `ReportForwardDeclareUse` and `ReportFullSymbolUse`. Maybe in `AddForwardDeclare` too, not entirely sure about it. What do you think?

kimgr:
I think one is enough, as the macro declares a function with a local type definition. Two macro expansions wouldn't work with the current macro form as it would introduce two functions w/ the same name. 

The assert sounds like a good idea, we occasionally get fwd decl suggestions for anonymous types.

kimgr:
Better/bigger solution in #350, closing this.

Issue #244, opened 2015-12-18
IWYU requires you to include header files for enums but there's no way you can't have included it 


ThosRTanner:
For instance, with QT, you have

    QFile::open(QIODevice::some flag)

Because this parameter is an enum and enums can't be forward declared, then the include has to be there for QFile::open to be declared.

So when in your code you then do

    QFile file("xxx");
    file.open(QIODevice::Read)

include what you use says you are missing `#include <QIODevice>`.

But it has to be implicitly available because of the way open is declared.

vsapsai:
Good point. IWYU discourages transitive includes but it looks reasonable to treat enum as part of exported API. Still need to see how it integrates with existing re-export cases.

Note: this is applicable to traditional enums, strongly typed enums is a different beast.

ThosRTanner:
Oh, quite agree about strongly typed ones (on the basis you can forward declare them anyway).

firewave:
Suggesting enum forward declaration was added a few versions ago.

Issue #105, opened 2015-06-15
types used in functions marked "override" need not have the corresponding header included


vsapsai:
Originally reported on Google Code with ID 105

```
1. Run IWYU on a subclass that overrides a virtual function that uses a type, and has
the c++11 override specifier.
2. IWYU suggests that the subclass should provide an include for the type in the overridden
method.

> What is the expected output? What do you see instead?

The base class header file must somehow include the type, and so the subclass header
is guaranteed to have a definition of the type since it must include the base header
(otherwise "override" will cause an error).  Therefore it is redundant to include the
type's header in the subclass.

> What version of the product are you using? On what operating system?

SVN linux build from 2013-08-27.

```

Reported by `mostynb@opera.com` on 2013-08-29 20:05:21

<hr>
- _Attachment: [override-include.tar.gz](https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-105/comment-0/override-include.tar.gz)_

PR #345, opened 2016-07-20
Automatically reexport enums in call-expressions

kimgr:
When an enum is used in a call-expression, the target of that
call-expression must have included the enum definition to be able to
declare its prototype.

Since we already depend on the callee itself, the enum must be
indirectly available and we can ignore its use in this context.

Commentary: This is a tentative fix for #244. I've moved code around most of the code base until I found this to be the cleanest cut to ignore uses of enumerators in call-expressions. Let me know what you think. 

@ThosRTanner, it'd be great if you could test this on your Qt example.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -901,9 +901,25 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
   // This is to catch assigning template functions to function pointers.
   // For instance, 'MyFunctionPtr p = &TplFn<MyClass*>;': we need to
   // expand TplFn to see if it needs full type info for MyClass.
+  //
+  // We also use it to ignore unscoped enumerators in call-expressions.
+  // They are never forward-declarable, which means the author of a called
+  // function will always have had to include their definition. Avoid forcing
+  // that tax on callers as well.
   bool TraverseDeclRefExpr(clang::DeclRefExpr* expr) {
-    if (!Base::TraverseDeclRefExpr(expr))  return false;
-    if (CanIgnoreCurrentASTNode())  return true;
+    // The base TraverseDeclRefExpr traverses 'upwards' in the
+    // hierarchy so for Class::Enum, we'll see Enum -> NestedNameSpecifier ->
+    // Class, which is kind-of weird.
+    // Short-circuit unscoped enums entirely before further traversal.
+    if (current_ast_node()->template ParentIsA<CallExpr>() &&
+        IsUnscopedEnum(expr)) {
+      VERRS(5) << "Ignoring enumerator in call-expression: "
+               << PrintableStmt(expr) << " (automatic re-export)\n";
+      return true;
+    }
+
+    if (!Base::TraverseDeclRefExpr(expr)) return false;
+    if (CanIgnoreCurrentASTNode()) return true;
 
     // If it's a normal function call, that was already handled by a
     // CallExpr somewhere.  We want only assignments.
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -70,6 +70,7 @@ using clang::DependentTemplateName;
 using clang::DependentTemplateSpecializationType;
 using clang::ElaboratedType;
 using clang::EnumDecl;
+using clang::EnumType;
 using clang::ExplicitCastExpr;
 using clang::Expr;
 using clang::ExprWithCleanups;
@@ -377,6 +378,13 @@ bool IsMemberOfATypedef(const ASTNode* ast_node) {
     if (nns->getAsType() && isa<TypedefType>(nns->getAsType()))
       return true;
   }
+  return false;
+}
+
+bool IsUnscopedEnum(const DeclRefExpr* expr) {
+  if (const EnumType* enum_type = DynCastFrom(GetTypeOf(expr->getDecl())))
+    return !enum_type->getDecl()->isScoped();
+
   return false;
 }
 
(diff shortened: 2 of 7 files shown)


ThosRTanner:
@kimgr as soon as i work out how to build iwyu on windows...

ThosRTanner:
Well, assuming I built this right (which is admittedly questionable)

```
#include <QFile>

void func(QString const &s)
{
  QFile x(s);
  x.open(QIODevice::ReadOnly);
}
```

results in

> source/organizer/testdevice.cpp should add these lines:
> #include <QIODevice>  // for QIODevice, QIODevice::OpenModeFlag::ReadOnly
> #include <QString>    // for QString

note that i did this as a patch onto the 3.8 label

kimgr:
I found a Qt installation on my machine and tried your example with:

    $ include-what-you-use -Wno-microsoft-enum-value -I C:\Qt\Qt5.4.1\5.4\msvc2012_opengl\include \
        -I C:\Qt\Qt5.4.1\5.4\msvc2012_opengl\include\QtCore -Xiwyu --mapping_file=qt5_4.imp \
        -Xiwyu -v7 tests\244\useqt.cpp

It turns out QIODevice declares the OpenMode enum as;

    enum OpenModeFlag {
        NotOpen = 0x0000,
        ReadOnly = 0x0001,
        WriteOnly = 0x0002,
        ReadWrite = ReadOnly | WriteOnly,
        Append = 0x0004,
        Truncate = 0x0008,
        Text = 0x0010,
        Unbuffered = 0x0020
    };
    Q_DECLARE_FLAGS(OpenMode, OpenModeFlag)

and the `Q_DECLARE_FLAGS` macro expands to a template `QFlags<OpenModeFlag>`. This seems to be a Qt pattern for type-safe enums.

So `QFlags<OpenModeFlag>` is declared in `QIODevice` and then implicitly constructed as a temporary in the call, so IWYU is right to attribute it to `QIODevice`.

The AST fragment for the call looks like this:

    tests/244/useqt.cpp:6:3: (4) [ CXXMemberCallExpr ] 0634AE78 CXXMemberCallExpr 0x634ae78 <tests\244\useqt.cpp:6:3, col:29> '_Bool'
    |-MemberExpr 0x634ae50 <col:3, col:5> '<bound member function type>' .open 0x6345e10
    | `-DeclRefExpr 0x634ada8 <col:3> 'class QFile' lvalue Var 0x634ad18 'x' 'class QFile'
    `-CXXConstructExpr 0x634aee0 <col:10, col:21> 'OpenMode':'class QFlags<enum QIODevice::OpenModeFlag>' 'void (class QFlags<enum QIODevice::OpenModeFlag> &&) __attribute__((thiscall)) noexcept' elidable
      `-MaterializeTemporaryExpr 0x634aed0 <col:10, col:21> 'class QFlags<enum QIODevice::OpenModeFlag>' xvalue
        `-ImplicitCastExpr 0x634aec0 <col:10, col:21> 'OpenMode':'class QFlags<enum QIODevice::OpenModeFlag>' <ConstructorConversion>
          `-CXXConstructExpr 0x634ae98 <col:10, col:21> 'OpenMode':'class QFlags<enum QIODevice::OpenModeFlag>' 'void (enum QIODevice::OpenModeFlag) __attribute__((thiscall))'
            `-DeclRefExpr 0x634ae30 <col:10, col:21> 'enum QIODevice::OpenModeFlag' EnumConstant 0x632a118 'ReadOnly' 'enum QIODevice::OpenModeFlag'

I'm not sure how to pattern-match this -- attempting to find a call argument being a class template over an enum type quickly becomes ambiguous/nonsensical.

But maybe this is another instance of #105 -- could it be that all public symbols in a base class should be automatically exported from a derived class? I can't tell what kind of fallout that would cause, but it sounds like a viable strategy. Unfortunately there's also no architecture in IWYU for that kind of relationship yet, and I haven't been able to think one up so far.

Anyway, the Qt specifics appear to be a separate problem; the enum re-export needs to happen either way.

ThosRTanner:
I could imagine working round this (for QT) if we could put export statements in the .imp file which would probably be an OK workround until #105 got solved. Of course, that depends on another issue and I'm not sure what state that's in.

iwyu.cc
```diff
@@ -901,9 +901,25 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
...
+    // The base TraverseDeclRefExpr traverses 'upwards' in the
+    // hierarchy so for Class::Enum, we'll see Enum -> NestedNameSpecifier ->
+    // Class, which is kind-of weird.
+    // Short-circuit unscoped enums entirely before further traversal.
+    if (current_ast_node()->template ParentIsA<CallExpr>() &&
+        IsUnscopedEnum(expr)) {
+      VERRS(5) << "Ignoring enumerator in call-expression: "
+               << PrintableStmt(expr) << " (automatic re-export)\n";
+      return true;
+    }
```
vsapsai: Having this check at this place seems a little bit strange. Need more time to think about it and to check other options.

kimgr: Yes, after thinking more about #105, it seems to me all these decisions should be made in iwyu_output -- iwyu.cc should just collect enough information to let iwyu_output handle automatic reexport and other peculiarities.

I'll start sketching on some architecture to make that happen.

kimgr:
Looks like we're in agreement that this should be solved some other way. Closing this.

PR #302, opened 2016-05-11
Add associated pragma to set associated header

ivankoster:
**Rationale:**
IWYU assumes that associated headers have the same file base name as the main translation unit file.
So foo.h is the associated file for foo.c

Many codebases in the wild, and also at the company i work for, do not follow this rule.
foo.h may be the associated header for foo_bar.c. Also, a *.c file may possibly have more than one associated header, or a header could be the associated header for more than one *.c files.

Since it is hard to define the rules for associated header detection on all the codebases in the wild, I propose the following pragma: // IWYU pragma: associated
So that the user can set associated headers for main translation unit files.

**About the code in this pull request:**
I took a bit of a long way to detect the 2 files that are supposed to be associated according to the pragma comment. In the handle pragma function I store the location of the pragma. Later, when exitting a header file to the stored pragma location, I know which two files must be linked.

I suppose there is a shorter way, if we are handling the pragma comment, maybe it's possible to search the #include statement right before the pragma comment and more directly link the files together. I have not explored this further due to my lack of knowledge of clang codebase and AST.

Comments / feedback on this pull request is welcome (this is my github first pull request)

PR #347, opened 2016-08-01
Don't require grep to run tests.

vsapsai:
While checking PR #302 I was able to reproduce failing comment_pragmas test. It was because of `grep` usage and I don't have `grep` installed on Windows.

I verified there are no performance regressions in running test suite. Tested on Python 2.7.5, 3.2.3.

### DIFF
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -11,7 +11,7 @@
 
 """Utilities for writing tests for IWYU.
 
-This script has been tested with python 2.4, 2.7, 3.1.3 and 3.2.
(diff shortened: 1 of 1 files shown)


kimgr:
Good catch!

Tests still pass as well as they do on Windows and FreeBSD.

Code looks good style-wise. I think it could be simplified now, but I can worry about that in a follow-up, please merge.

vsapsai:
Thanks for review, committed 2f826a3323b4ef8dcab13ad31e8ac59113ec84ae.

Issue #338, opened 2016-07-03
mappings behaviour changed from 0.5 to 0.6


ThosRTanner:
This

```
#include <QDebug>

int main()
{
    qDebug() << "fish";
    return 0;
}

```

with this compile line (on windows)

```

bin\include-what-you-use.exe -Xiwyu --transitive_includes_only -D_MT -D_DLL -m32 -DBOOST_USE_WINDOWS_H -Wno-inconsistent-missing-override --system-header-prefix=Q --system-header-prefix=boost/ -fexceptions -fmsc-version=1800 -DQ_COMPILER_INITIALIZER_LISTS -DQ_COMPILER_DECLTYPE -DQ_COMPILER_VARIADIC_TEMPLATES -D_DEBUG -Xiwyu --mapping_file=share\qt5_4.imp -DUNICODE -DWIN32 -DNOMINMAX -DQT_CORE_LIB -DQT_GUI_LIB -DQT_WIDGETS_LIB -DGAMEFALLOUT4_LIBRARY -IC:\Apps\Qt\5.5\msvc2013\mkspecs\win32-msvc2013 -IC:\Apps\Qt\5.5\msvc2013\include -IC:\Apps\Qt\5.5\msvc2013\include -IC:\Apps\Qt\5.5\msvc2013\include\QTCore -IC:\Apps\Qt\5.5\msvc2013\include\QTGui -IC:\Apps\Qt\5.5\msvc2013\include\QTWidgets  -I"C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\INCLUDE" test.cc

```

with IWYU 0.5 produces this:

```
test.cc should add these lines:
#include <QtDebug>  // for qDebug

test.cc should remove these lines:

The full include-list for test.cc:
#include <QDebug>   // for QDebug
#include <QtDebug>  // for qDebug

---

```

but in 0.6 produces this

```
test.cc should add these lines:
#include <QtDebug>  // for qDebug

test.cc should remove these lines:

The full include-list for test.cc:
#include <QDebug>   // for QDebug
#include <QMessageLogger>  // for QMessageLogger
#include <QtDebug>  // for qDebug

---

```

This addition of `QMessageLogger` is an unexpected regression.

In passing, if I add `QtDebug`, then IWYU 0.5 tells me to **remove** `QDebug`, after which it is happy. However, 0.6 behaves differently and requires `QDebug` as well.

[...]

Issue #335, opened 2016-06-30
Mappings in qt5_4.imp not working


estan:
With

```cpp
#include <QDebug>

int main() {
    qWarning() << "sweet";
}
```

I get

```
main.cpp should add these lines:
#include <qlogging.h>      // for qWarning
#include <QMessageLogger>  // for QMessageLogger

main.cpp should remove these lines:

The full include-list for main.cpp:
#include <qlogging.h>      // for qWarning
#include <QDebug>          // for QDebug
#include <QMessageLogger>  // for QMessageLogger
```

With more complex Qt code, I also get many other wrong suggestions (`qobjectdefs.h`, `qnamespace.h`, ...) that, by the looks of it, the mapping file should take care of.

This is with Qt 5.7.0. I know the mapping file is made for Qt 5.4, but it should still catch these things, right?

estan:
[iwyu.out.txt](https://github.com/include-what-you-use/include-what-you-use/files/342221/iwyu.out.txt) is the output of the invocation

```
include-what-you-use -Xiwyu --verbose=5 -Xiwyu --mapping_file=qt5_4.imp -std=c++11 -c -pipe -O2 -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -Wall -W -D_REENTRANT -fPIC -DQT_NO_DEBUG -DQT_GUI_LIB -DQT_CORE_LIB -I. -I. -isystem /usr/include/qt -isystem /usr/include/qt/QtGui -isystem /usr/include/qt/QtCore -I. -I/usr/lib/qt/mkspecs/linux-g++ -o main.o main.cpp
```

which is hopefully of some use to debug the issue.

estan:
This is with `include-what-you-use 0.6 based on clang version 3.8.0 (tags/RELEASE_380/final)` BTW.

ThosRTanner:
That's odd. As far as I understand it, this line:

     { include: [ "@\"(QtCore/)?qlogging\\.h\"", "private", "<QtDebug>", "public" ] }, #qDebug, qWarning, etc

should remove qlogging.h and ask for <QtDebug> (not QDebug. I plumped for the former, because there's two headers, one of which gets a more flexible version of qDebug()). Having said which, I have no idea where QMessageLogger is coming from. I suppose the way the headers are included might have changed so things are no longer recognised. I should get sufficient courage up to upgrade my QT install soon and will have a look see.

estan:
[...]

PR #350, opened 2016-08-09
Improve macro use-attribution heuristics

kimgr:
Bugs #334, #335 and #338 were all caused by the new macro location
rule that considers a forward-decl in the macro definition file a
hint that uses of the declared type inside a macro are attributed
to the macro's expansion location.

Several different kinds of declarations were mis-characterized as
forward decls.

This patch improves IsForwardDecl to better detect this, and
simplifies the heuristic rule so that any forward-declaration
outside of a macro in the same file as the macro definition will
count as a use-attribution hint.

This deprecates #342.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1369,19 +1369,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return type;
   }
 
-  // Return any forward-declare of decl earlier in the same file as loc.
-  const NamedDecl* GetForwardDeclareInSameFile(const Decl* decl,
-                                               SourceLocation loc) {
-    set<const NamedDecl*> redecls = GetClassRedecls(DynCastFrom(decl));
-    for (const NamedDecl* redecl : redecls) {
-      if (IsBeforeInSameFile(redecl, loc)) {
-        return redecl;
-      }
-    }
-
-    return nullptr;
-  }
-
   // Get the canonical use location for a (location, decl) pair.
   // Decide whether the file expanding the macro or the file defining the macro
   // should be held responsible for a use.
@@ -1391,12 +1378,22 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (!IsInMacro(use_loc))
       return use_loc;
 
-    // If we're in a macro, and the macro file forward-declares decl, make sure
-    // we keep that forward-declaration.
-    const NamedDecl* fwd_decl = GetForwardDeclareInSameFile(decl, use_loc);
+    // If the macro definition file contains a forward-declaration for decl,
+    // we treat that as a use-attribution hint. See below for details.
+    const FileEntry* macro_def_file = GetFileEntry(GetSpellingLoc(use_loc));
+    const NamedDecl* fwd_decl = nullptr;
+    for (const NamedDecl* redecl : GetClassRedecls(decl)) {
+      if (GetFileEntry(redecl) == macro_def_file &&
+          IsForwardDecl(redecl)) {
+        fwd_decl = redecl;
+        break;
+      }
+    }
+
     if (fwd_decl) {
-      const FileEntry* used_in = GetFileEntry(use_loc);
-      preprocessor_info().FileInfoFor(used_in)->ReportForwardDeclareUse(
+      // Make sure we keep that forward-declaration, even if it's probably
+      // unused in this file.
+      preprocessor_info().FileInfoFor(macro_def_file)->ReportForwardDeclareUse(
           use_loc, decl, IsNodeInsideCXXMethodBody(current_ast_node()),
           nullptr);
     }
--- a/iwyu_ast_util.h
+++ b/iwyu_ast_util.h
@@ -589,11 +589,11 @@ bool IsDefaultNewOrDelete(const clang::FunctionDecl* decl,
 // Returns true if this decl is part of a friend decl.
 bool IsFriendDecl(const clang::Decl* decl);
 
-// Returns true if a Record-decl looks like a forward-declaration of a
+// Returns true if a named decl looks like a forward-declaration of a
 // class (rather than a definition, a friend declaration, or an 'in
 // place' declaration like 'struct Foo' in 'void MyFunc(struct Foo*);'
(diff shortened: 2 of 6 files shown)


vsapsai:
Overall looks good to me, just need to check a few details like relationship between `TagDecl`, `NamedDecl`, `RecordDecl`; do we need `IsBefore` part from `IsBeforeInSameFile`.

kimgr:
Thanks!

I intentionally skipped the `IsBefore` part to allow patterns like:

    // Public API
    #define MYMACRO(x) \
        MyClass().Method(x); \
        MyOtherClass(x).SomeOtherMethod();

    #define MYOTHERMACRO(y) \
        TheThirdClass y;

    // Forward-declare implementation details to force them on users.
    class MyClass;
    class MyOtherClass;

The precise location of the forward declaration doesn't seem important.

kimgr:
Also, just to be clear on the `TagDecl`/`RecordDecl`/`NamedDecl` change -- I did that to push casts from callers into `IsForwardDecl`. Since the function was already doing dynamic type checking, we might as well leave it less constrained at compile-time.

This opens up for detecting forward-decls of functions, for example.

iwyu.cc
```diff
@@ -1392,12 +1379,23 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    const FileEntry* macro_def_file = GetFileEntry(use_loc);
+    const NamedDecl* fwd_decl = nullptr;
+    for (const NamedDecl* redecl : GetClassRedecls(decl)) {
+      if (GetLocation(redecl).isFileID() &&
+          GetFileEntry(redecl) == macro_def_file &&
+          IsForwardDecl(redecl)) {
+        fwd_decl = redecl;
+        break;
+      }
+    }
```
vsapsai: What case `isFileID` is covering? Just curious, nothing wrong with it.

It might be useful to mention that for macros it's not required for forward declaration to be before usage. Not sure about it, I struggled with it at first but your response in PR made it clear.

And the biggest problem for me was how did we get from `use_loc` to `macro_def_file` because use should be not in macro definition file. And the answer to that is that [`GetFileEntry` looks past spelling location](https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_location_util.h#L145-L168) and arrives at macro definition file. I don't know, having something like
`const FileEntry* macro_def_file = GetFileEntry(GetSpellingLoc(use_loc));`
might express the intention better.

Anyway, how forward declare hint is supposed to work with macro token concatenation? Because I came up with test case
```c++
// tests/cxx/issue350-i1.h
class debugLogger {
public:
  void Log(const char* msg) {}
};

// ================
// tests/cxx/issue350-d1.h
#include "tests/cxx/issue350-i1.h"

#define LOG(level, x) \
  level ## Logger().Log(x);

// ================
// tests/cxx/issue350.cpp
#include "tests/cxx/issue350-d1.h"

void foo() {
    LOG(debug, "test");
}
```

And IWYU recommends to `#include "tests/cxx/issue350-i1.h"` for `debugLogger`. Though my understanding is that it shouldn't because there is no forward declaration and `issue350-d1.h` is responsible for including `issue350-i1.h`.

vsapsai:
Thanks for explaining issues I had concerns about. Everything seems fine, I've found a test case with macro token concatenation, not sure how your patch is supposed to handle it.

kimgr:
> What case isFileID is covering? Just curious, nothing wrong with it.

I wanted to make sure a forward-declaration inside a macro definition would not get flagged as a hint, e.g.

    DECLARE_X(name) \
        X x;

    SOME_OTHER_MACRO(a, b) \
        X().print(a, b);

I don't think users of `SOME_OTHER_MACRO` should be forced to include `X.h` because of `DECLARE_X`.

kimgr:
That said, I wonder if I'm doing this right at all. Considering #358, nested macros don't seem to work with or without this change.

I'm currently looking at the Clang diagnostic macro backtrace, that seems like it should need to solve many of the same problems I'm addressing.

iwyu.cc
kimgr: That one's difficult :-)

The forward-declare rule never comes into play here -- all uses involving macro arguments will be attributed to the expansion location, because, in this case, issue350.cpp is passing in part of the used symbol, so it's assumed to know what it wants to "bind to".

For example, think of how your reasoning would change if it passed in an unknown Logger, e.g.:

    LOG(critical, "test");

Assuming the log framework does not have a `criticalLogger` it's intuitive that the caller is responsible for providing the type.

But as I said, this one is difficult, because your example and conclusion is equally sensible.

kimgr:
Pushed a new version.

Remove `isFileID` check, I couldn't think of a case where that was relevant.
Use `GetSpellingLoc` explicitly to get the macro-definition file.

iwyu_ast_util.h
```diff
@@ -589,11 +589,11 @@ bool IsDefaultNewOrDelete(const clang::FunctionDecl* decl,
 // Returns true if this decl is part of a friend decl.
 bool IsFriendDecl(const clang::Decl* decl);
 
-// Returns true if a Record-decl looks like a forward-declaration of a
+// Returns true if a named decl looks like a forward-declaration of a
 // class (rather than a definition, a friend declaration, or an 'in
 // place' declaration like 'struct Foo' in 'void MyFunc(struct Foo*);'
 // Always returns false for enums.
```
vsapsai: Don't know if we should mention in comment that anonymous `RecordDecl` aren't forward declarations. Leaning more towards not mentioning it because comments don't have to repeat the code and I don't think someone can expect anonymous structure to be a forward declaration.

vsapsai:
Tested with LLVM+Clang at r283206 and everything looks fine, no regressions. It's up to you to push the change now or to wait till we fix broken build.

iwyu_ast_util.h
kimgr: I agree, Until the example that raised the bug came up, I never would've considered an anonymous type to be a forward decl.

kimgr:
I'll push it now, to be able to merge it to the release branch (which should currently be building green).

kimgr:
Pushed in 3e98eed.

Issue #331, opened 2016-06-17
Incorrectly removing necessary forward declarations


Bothari:
I've run into a pretty common and annoying bug with IWYU, where it removes a necessary forward declaration. See the attached test.cpp/.h files for an example. The output from IWYU is:

`11:14 > ./include-what-you-use.exe c:/Users/mdyckhoff/iwyu-test/test.cpp`

`c:/Users/mdyckhoff/iwyu-test/test.h should add these lines:`

`c:/Users/mdyckhoff/iwyu-test/test.h should remove these lines:`
`- class Container::Iterator;  // lines 9-9`

`The full include-list for c:/Users/mdyckhoff/iwyu-test/test.h:`
`---`

However if that line is removed, the file fails to compile, as Element is defined before Iterator:

`c:/Users/mdyckhoff/iwyu-test/test.h:13:31: error: no class named 'Iterator' in 'Container<T>'
                        friend class Container<T>::Iterator;
                                     ~~~~~~~~~~~~~~^`

I am using IWYU on Windows, in Cygwin. Version info:

`include-what-you-use 0.4 (645) based on clang version 3.6.1 (branches/release_36 238234)`

Any advice or fixes?

Bothari:
Attached files test.cpp and test.h.

[test.zip](https://github.com/include-what-you-use/include-what-you-use/files/321065/test.zip)

kimgr:
Thanks for the report! I can reproduce on Windows/MSVC and I agree it looks like a bug.

kimgr:
This appears to have been fixed in IWYU 0.7!

PR #351, opened 2016-08-11
Recognize nested classes in friend declarations

kimgr:
Fragments such as this:

    friend class Container<T>::Iterator;

would not be recognized as needing a forward-declaration, so IWYU
would suggest removing forward-decls of Iterator.

This patch changes this and expands the test suite for nested classes
to cover containing class templates and friend declarations.

Fixes issue #331.


vsapsai:
Looks good to me. No regressions and fix is working on Mac OS X, Ubuntu.

kimgr:
Thanks, pushed in 9c432a97639f61a842bbc3f058e249fde6df39d6.

PR #352, opened 2016-08-12
Clean up output from test runner

kimgr:
Here's a small one with some aesthetics that have been bugging me for a while:

* Only quote the IWYU executable name when necessary
* Remove logging of IWYU and Clang flags, they are already visible on
  the command-line

### DIFF
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -206,14 +206,8 @@ class OneIwyuTest(unittest.TestCase):
     files_to_check = [PosixPath(f) for f in files_to_check]
 
     iwyu_flags = self._iwyu_flags_map.get(filename, None)
-    if iwyu_flags:
-      logging.info('%s: Using iwyu flags %s', filename, str(iwyu_flags))
-
     clang_flags = self._clang_flags_map.get(filename, [])
     clang_flags.extend(self._include_map.get(filename, []))
-    if clang_flags:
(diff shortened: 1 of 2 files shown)


run_iwyu_tests.py
```diff
@@ -206,14 +206,8 @@ def RunOneTest(self, filename):
...
 
     iwyu_flags = self._iwyu_flags_map.get(filename, None)
-    if iwyu_flags:
-      logging.info('%s: Using iwyu flags %s', filename, str(iwyu_flags))
-
     clang_flags = self._clang_flags_map.get(filename, [])
     clang_flags.extend(self._include_map.get(filename, []))
-    if clang_flags:
-      logging.info('%s: Using clang flags %s', filename, str(clang_flags))
-
```
vsapsai: I see it turned out not to be particularly useful. `-Xiwyu` is sufficient to tell if flag is IWYU flag or Clang flag.

vsapsai:
`_ShellQuote` is not bulletproof but will work in most cases. No regressions, OK to commit.

kimgr:
Yeah, I figured we could patch `_ShellQuote` as necessary in the future (spaces in paths are likely the most common issue now). I've tried `pipes.quote` which is the undocumented but official way to do this, but it considers Windows path separators escape-worthy, so it didn't really work.

kimgr:
Thanks, pushed as f09dba693412463fcca0898a202763761fe60ccc.

PR #355, opened 2016-08-16
Terminology: as-typed -> as-written

kimgr:
Clang uses mostly "as-written" in its APIs. Follow that convention
more consistently.

Sparked by Eugene's comment on the recent clang-tidy patches.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3088,7 +3088,7 @@ class InstantiatedTemplateVisitor
     const Type* class_type = GetTypeOf(expr);
     if (CanIgnoreType(class_type))  return true;
 
-    // If the ctor type is a SubstTemplateTypeParmType, get the type-as-typed.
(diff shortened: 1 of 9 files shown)


vsapsai:
No regressions. Do you want to fix `GetIncludeNameAsTyped` in [more_tests/iwyu_lexer_utils_test.cc](https://github.com/include-what-you-use/include-what-you-use/blob/master/more_tests/iwyu_lexer_utils_test.cc#L243) too?

And I checked that Clang in interfaces isn't using "asTyped" (only as a part of "hasTyped"). While at the same time you have NameAsWritten, FeaturesAsWritten, UmbrellaAsWritten, isVirtualAsWritten, getTypeInfoAsWritten, and many more asWritten.

kimgr:
Oops, good catch. Fixed.

kimgr:
This seemed non-controversial, so I merged it after fixing up the unit tests too.

Issue #247, opened 2015-12-23
Include C++ header directory in binary release


jasonmp85:
The `0.5.0` (binary) release now organizes files into `bin`, `docs`, and `lib` folders where previously the binaries were just included at the top level of the release archive.

I was pretty excited that the `lib` directory appeared to contain headers needed by `include-what-you-use`, as the homebrew package I've created previously had to create symlinks into similar headers provided by Xcode (see jasonmp85/homebrew-iwyu#8).

I develop in C, so the headers now provided are sufficient for me, but I've been told (see jasonmp85/homebrew-iwyu#5) that some extra work is needed so `include-what-you-use` can find C++ standard headers.

I'd like to simplify my build (doing so would increase the likelihood of the homebrew project itself accepting my formula) to use the packaged C headers, but with no C++ headers this is still an insufficient replacement for the Xcode symlinking I presently have.

Am I misunderstanding the purpose of this new `lib` directory? If not, why is there not also an `includes/c++/v1` directory for C++ projects? If it's just an oversight, can we get them added?

I looked for existing issues discussing the reorganization of the release archive, but couldn't find anything useful.

vsapsai:
Including internal headers in released binary was discussed in ["Clang 3.6-compatible release"  thread](https://groups.google.com/forum/#!topic/include-what-you-use/npGnf4c_fk0). It was my mistake I haven't included these headers in IWYU 0.4 Mac build.

`includes/c++/v1` headers aren't included on purpose. These are libc++ headers included explicitly in code. And IWYU expects them to be already present on the system. Internal headers aren't included in analysed code directly, but Clang libraries require them.

[...]

PR #360, opened 2016-09-12
On macOS search for libc++ headers in command-line tools and Xcode (fix issue #247).

vsapsai:


vsapsai:
Tested on macOS with `/usr/include/c++/v1` absent and present - no regressions in tests. And the tests that were failing without this directory are passing now. Also tested that there are no regressions on Ubuntu.

kimgr:
No regressions on Windows or FreeBSD. LGTM!

vsapsai:
Thanks for review, Kim. Manually merged changes in 9ab7286d56d05193d1c0e1118e5dcdfa705bedf3.

Also I thought about it a little bit more and think it can complicate life for people building their own libc++ because in this case most likely system-wide libc++ will be discovered first. I decided not to fix it right now as I don't know if anybody uses include-what-you-use with the latest libc++ and if it is crucial that libc++ is the latest. And I don't know where exactly people are installing libc++. Potential fix can be something like
```c++
invocation->getHeaderSearchOpts().AddPath(
    remove_filename(path) + "../include/c++/v1",
    clang::frontend::CXXSystem,
    /*IsFramework=*/false, /*IgnoreSysRoot=*/true);
```
before `/Library/Developer/CommandLineTools/usr/include/c++/v1/` and `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1`.

But I don't know if this fix will work, so decided against adding it now.

PR #367, opened 2016-10-25
Fix Clang-tidy readability-redundant-member-init warnings. Modernize > >

EugeneZelenko:


iwyu_output.cc
```diff
@@ -240,11 +240,7 @@ OneUse::OneUse(const NamedDecl* decl, SourceLocation use_loc,
       use_loc_(use_loc),
       use_kind_(use_kind),             // full use or fwd-declare use
       in_cxx_method_body_(in_cxx_method_body),
-      comment_(comment ? comment : ""),
-      public_headers_(),
-      suggested_header_(),             // figure that out later
-      ignore_use_(false),
-      is_iwyu_violation_(false) {
```
kimgr: We shouldn't be removing `ignore_use_` and `is_iwyu_violation_` here, should we?

Won't they be garbage? I think I've seen recent claims that members are default-initialized, but I can't find a reference now. I'd prefer to keep the explicitly initialized to false anyway.

iwyu_output.cc
```diff
@@ -257,12 +253,7 @@ OneUse::OneUse(const string& symbol_name, const FileEntry* dfn_file,
       decl_filepath_(dfn_filepath),
       use_loc_(use_loc),
       use_kind_(kFullUse),
-      in_cxx_method_body_(false),
-      comment_(),
-      public_headers_(),
-      suggested_header_(),
-      ignore_use_(false),
-      is_iwyu_violation_(false) {
```
kimgr: Keep Boolean inits

iwyu_output.cc
```diff
@@ -434,9 +425,10 @@ string MungedForwardDeclareLine(const NamedDecl* decl) {
 OneIncludeOrForwardDeclareLine::OneIncludeOrForwardDeclareLine(
     const NamedDecl* fwd_decl)
     : line_(internal::MungedForwardDeclareLine(fwd_decl)),
-      start_linenum_(-1), end_linenum_(-1),     // set 'for real' below
-      is_desired_(false), is_present_(false), symbol_counts_(),
```
kimgr: Keep Boolean inits

iwyu_output.cc
```diff
@@ -448,9 +440,10 @@ OneIncludeOrForwardDeclareLine::OneIncludeOrForwardDeclareLine(
 OneIncludeOrForwardDeclareLine::OneIncludeOrForwardDeclareLine(
     const FileEntry* included_file, const string& quoted_include, int linenum)
     : line_("#include " + quoted_include),
-      start_linenum_(linenum), end_linenum_(linenum),
-      is_desired_(false), is_present_(false), symbol_counts_(),
```
kimgr: Keep Boolean inits

iwyu_output.h
```diff
@@ -108,8 +108,8 @@ class OneUse {
   string comment_;                 // If not empty, append to clang warning msg
   vector<string> public_headers_;  // header to #include if dfn hdr is private
   string suggested_header_;        // header that allows us to satisfy use
-  bool ignore_use_;                // set to true if use is discarded
-  bool is_iwyu_violation_;         // set to false when we figure out it's not
+  bool ignore_use_ = false;        // set to true if use is discarded
```
kimgr: Ah, I see what happened now. Can we please keep initialization in the init-list? Better to have initialization in one place.

iwyu_output.h
```diff
@@ -108,8 +108,8 @@ class OneUse {
...
+  bool is_iwyu_violation_ = false; // set to false when we figure out it's not
```
kimgr: Move back to init list

iwyu_output.h
```diff
@@ -169,8 +169,8 @@ class OneIncludeOrForwardDeclareLine {
   string line_;                     // '#include XXX' or 'class YYY;'
   int start_linenum_;
   int end_linenum_;
-  bool is_desired_;                 // IWYU will recommend this line
-  bool is_present_;                 // line was present before the IWYU run
+  bool is_desired_ = false;         // IWYU will recommend this line
```
kimgr: Move back to init list

EugeneZelenko:
I reverted C++11-style member initialization. I also fixed Clang-tidy modernize-use-default warnings.

kimgr:
Looks good to me. @vsapsai ?

vsapsai:
Looks good to me and there are no regressions. My knowledge is outdated but as long as MSVC version we are using supports defaulted constructors and destructors, I have no objections.

EugeneZelenko:
Minimum MSVC version required by LLVM should support = default. Problem was some default constructors/destructors with GCC 4.7, but requirement was changed to GCC 4.8.

kimgr:
Let me double-check MSVC (Windows) and clang-3.6 builds (FreeBSD). I'll merge if it goes well.

kimgr:
Done, eb7f5d1ef02836. Thanks!

PR #370, opened 2016-11-08
Handle symbolic links in source paths correctly

svenpanne:
When use realpath on source files given on the command line, so we should
better use realpath for the paths in the compilation DB, too.  Otherwise we
don't find source paths which contain a symbolic link and incorrectly emit a
warning instead.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -62,6 +62,10 @@ def main(compilation_db_path, source_files, verbose, iwyu_args):
     with open(compilation_db_path, 'r') as fileobj:
         compilation_db = json.load(fileobj)
 
+    # expand symlinks
+    for entry in compilation_db:
+        entry['file'] = os.path.realpath(entry['file'])
+
     # Cross-reference source files with compilation database
     source_files = [os.path.realpath(s) for s in source_files]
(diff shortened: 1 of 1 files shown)


kimgr:
I've run into this problem as well. I blindly accepted the suggestion that we use realpath for source files, but I'm starting to wonder whether that's necessary. What would suffer if we didn't use realpath for source files?

svenpanne:
I think using `realpath` is conceptually the right thing to do: We get paths from two sources, namely the compilation DB and the command line. We have no clue about the format we get, e.g. [Build EAR](https://github.com/rizsotto/Bear) emits absolute paths into the compilation DB, but without expanding symlinks. No idea what CMake's `CMAKE_EXPORT_COMPILE_COMMANDS` option does, and absolutely no clue what the user passes via the command line.

So if we want to compare files for identity at all, we need to use some normalized form. Under Unix, a device/inode pair would be that canonical form, but that't not portable, so `realpath` is the best thing we got.

kimgr:
Makes perfect sense to me, thanks! I'm guessing this is non-controversial, so I'll just merge it.

PR #371, opened 2016-11-08
Use a variable name which makes pylint happy

svenpanne:


kimgr:
Heh, I actually think the original name was `double_dash` and I shortened it to make the list slicing easier to read. But I'm all for pylint happiness. I'll merge this when I have some time.

PR #372, opened 2016-11-08
Add a new option for a more tool-friendly output.

svenpanne:
The standard output format of iwyu is not very friendly for tools like
Jenkins' warnings-plugin (https://github.com/jenkinsci/warnings-plugin) or
editors.  This change adds a new -c/--clang-style command line option for
mangling the output into something clang-like, enabling nice integration
into emacs/Jenkins/etc. out of the box.

Somehow this should probably be the default and be handled on the native
side (not as a post-processing step), but the current change is probably the
least controversial and easiest one for now.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -25,9 +25,77 @@ import sys
 import json
 import argparse
 import subprocess
+import re
 
 
-def run_iwyu(cwd, compile_command, iwyu_args, verbose):
+def iwyu_formatter(output):
+    """ Process iwyu's output, basically a no-op. """
+    print('\n'.join(output))
+
+
+CORRECT_RE = re.compile(r'^\((.*?) has correct #includes/fwd-decls\)$')
+SHOULD_ADD_RE = re.compile(r'^(.*?) should add these lines:$')
+SHOULD_REMOVE_RE = re.compile(r'^(.*?) should remove these lines:$')
+FULL_LIST_RE = re.compile(r'The full include-list for (.*?):$')
+END_RE = re.compile(r'^---$')
+LINES_RE = re.compile(r'^- (.*?)  // lines ([0-9]+)-[0-9]+$')
+
+
+GENERAL, ADD, REMOVE, LIST = range(4)
+
+
+def clang_formatter(output):
+    """ Process iwyu's output into something clang-like. """
+    state = (GENERAL, None)
+    for line in output:
+        match = CORRECT_RE.match(line)
+        if match:
+            print('%s:1:1: note: #includes/fwd-decls are correct' % match.groups(1))
+            continue
+        match = SHOULD_ADD_RE.match(line)
+        if match:
+            state = (ADD, match.group(1))
+            continue
+        match = SHOULD_REMOVE_RE.match(line)
+        if match:
+            state = (REMOVE, match.group(1))
+            continue
+        match = FULL_LIST_RE.match(line)
+        if match:
+            state = (LIST, match.group(1))
+        elif END_RE.match(line):
+            state = (GENERAL, None)
+        elif not line.strip():
+            continue
+        elif state[0] == GENERAL:
+            print(line)
+        elif state[0] == ADD:
+            print('%s:1:1: error: add the following line' % state[1])
+            print(line)
+        elif state[0] == REMOVE:
+            match = LINES_RE.match(line)
+            line_no = match.group(2) if match else '1'
+            print('%s:%s:1: error: remove the following line' % (state[1], line_no))
+            print(match.group(1))
+
+
+DEFAULT_FORMAT = 'iwyu'
+FORMATTERS = {
+    'iwyu': iwyu_formatter,
+    'clang': clang_formatter
+}
+
+def get_output(cwd, command):
+    """ Run the given command and return its output as a string. """
+    process = subprocess.Popen(command,
+                               cwd=cwd,
+                               shell=True,
+                               stdout=subprocess.PIPE,
+                               stderr=subprocess.STDOUT)
+    return process.communicate()[0].decode("utf-8").splitlines()
+
+
+def run_iwyu(cwd, compile_command, iwyu_args, verbose, formatter):
     """ Rewrite compile_command to an IWYU command, and run it. """
     compiler, _, args = compile_command.partition(' ')
     if compiler.endswith('cl.exe'):
@@ -43,10 +111,10 @@ def run_iwyu(cwd, compile_command, iwyu_args, verbose):
     if verbose:
         print('%s:' % command)
 
-    return subprocess.call(command, cwd=cwd, shell=True)
+    formatter(get_output(cwd, command))
 
 
-def main(compilation_db_path, source_files, verbose, iwyu_args):
+def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
     """ Entry point. """
     # Canonicalize compilation database path
     if os.path.isdir(compilation_db_path):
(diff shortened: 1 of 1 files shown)


iwyu_tool.py
```diff
@@ -25,9 +25,67 @@
...
+
+CORRECT_RE = re.compile(r'^\((.*?) has correct #includes/fwd-decls\)$')
+SHOULD_ADD_RE = re.compile(r'^(.*?) should add these lines:$')
+SHOULD_REMOVE_RE = re.compile(r'^(.*?) should remove these lines:$')
+FULL_LIST_RE = re.compile(r'The full include-list for (.*?):$')
+END_RE = re.compile(r'^---$')
+LINES_RE = re.compile(r'.*// lines ([0-9]+)-[0-9]+$')
+
+
+GENERAL, ADD, REMOVE, LIST = range(4)
```
kimgr: I think this needs to be `list(range(4))` to be Python 3-compatible (quirky but nice idiom for enums, btw, I've never seen that before)

svenpanne: Just out of curiosity: What's the minimum Python version we have to support? Looking at the code, it seems to be Python 2 and 3, but which versions exactly? `namedtuple` could be used for the state tuple to make my diff a bit nicer, but that's only available for Python >= 2.6, and real enums for >= 3.4. Both are not really crucial for such short code, but anyway...

kimgr: iwyu_test_util.py says:

    This script has been tested with python 2.7, 3.1.3 and 3.2.

So I think 2.7 and later is fine. I didn't even know real enums were available in Python, but 3.4 is probably a little too harsh a requirement. `namedtuple` sounds good to me!

iwyu_tool.py
```diff
@@ -123,6 +185,8 @@ def custom_help():
 
     parser.add_argument('-v', '--verbose', action='store_true',
                         help='Print IWYU commands')
+    parser.add_argument('-c', '--clang-style', action='store_true',
```
kimgr: I wonder if you could make this a `--output` argument instead, with a set of fixed options, e.g. `iwyu`, `clang`, etc. That way it's more extensible. I'm already dreaming of plain diff output...

Something like:

    parser.add_argument('-o', '--output', type=str,
                        choices=['iwyu', 'clang'], default='iwyu',
                        help='Output format')

svenpanne:
OK, I've uploaded a new version with an -o/--output-format option and restructured the code slightly to add more formats easily.

Regarding `range(4)`: This is OK as it is, because unpacking is based on _sequences_, not lists.

Regarding `namedtuple` and friends: I'll leave it as it is, let's not over-engineer things.

vsapsai:
How can we test `iwyu_tool.py`? It was supposed to be fast and loose solution. But as we are adding more stuff there I feel uneasy without tests.

kimgr:
The same thing occurred to me. But I think we can survive without tests until we fold this into include-what-you-use proper.

kimgr:
Finally got around to testing this. On FreeBSD:

    [kimgr@beast ~/code/llvm-trunk/llvm/tools/clang/tools/include-what-you-use]$ ./iwyu_tool.py -o clang -p $IWYU_BUILD_ROOT iwyu_verrs.cc

    /home/kimgr/code/llvm-trunk/llvm/tools/clang/tools/include-what-you-use/iwyu_verrs.h:15:1: error: remove the following line
    #include "llvm/Support/raw_ostream.h"  // lines 15-15
    /home/kimgr/code/llvm-trunk/llvm/tools/clang/tools/include-what-you-use/iwyu_verrs.h:1:1: error: full include list follows
    namespace clang { class FileEntry; }  // lines 18-18
    
    /home/kimgr/code/llvm-trunk/llvm/tools/clang/tools/include-what-you-use/iwyu_verrs.cc has correct #includes/fwd-decls

It looks weird with the `error: full include list follows`. I wonder if that output should be omitted, or replaced with e.g. a note, to match Clang parlance?

Same thing with `...iwyu_verrs.cc has correct #includes/fwd-decls` -- would it be better to suppress that? It's not like the compiler says: `generated correct object file abc.o`.

kimgr:
On Windows:

    D:\code\llvm-trunk\llvm\tools\clang\tools\include-what-you-use>python iwyu_tool.py -o clang -p %build_root% iwyu_verrs.cc

    D:/code/llvm-trunk/llvm/tools/clang/tools/include-what-you-use/iwyu_verrs.h:15:1: error: remove the following line

    #include "llvm/Support/raw_ostream.h"  // lines 15-15
    D:/code/llvm-trunk/llvm/tools/clang/tools/include-what-you-use/iwyu_verrs.h:1:1: error: full include list follows

    namespace clang { class FileEntry; }  // lines 18-18

    D:/code/llvm-trunk/llvm/tools/clang/tools/include-what-you-use/iwyu_verrs.cc has correct #includes/fwd-decls

No platform-specific problems, as far as I can see, just the same issues as last comment.

svenpanne:
Regarding `error: full include list follows`: I think the output itself is very helpful, regardless of the chosen output format. I don't care about the exact wording, perhaps you can suggest something concrete which is better? (I don't know exactly what the clang-way of saying something informative is in detail.) We could tie the output to the `--verbose`-flag, but this would be different from the traditional behavior of the tool. It somehow feels wrong that choosing the output format dictates _what_ is printed, IMHO it should only control _how_ it is printed.

The same reasoning holds for printing `... has correct ...`, I would really like to see that, especially because running iwyu takes ages for non-trivial projects, so printing something from time to time is a sign that things are still alive. But I agree that it probably should not be flagged with `error`, but with the clang-way of saying something informative. Suggestions?

In a nutshell: We could omit the full include list and the `... has correct ...` messages per default and tie both to `--verbose`, but then we should do the same for the standard output format for consistency. But I fear that this would break a few tools/scripts out there. 

Anyway, upstream has the last word... :wink:

kimgr:
I don't like the `--verbose` strategy either, let's not do that :-)

Clang has two more neutral diagnostic types: notes and remarks, see
http://clang.llvm.org/docs/UsersManual.html#diagnostic-mappings

and the following for examples:
http://clang.llvm.org/docs/UsersManual.html#rpass
http://clang.llvm.org/docs/UsersManual.html#cmdoption-fno-elide-type

Remarks seem to be geared toward optimizations, so I think a `note:` prefix would be suitable here (even if notes normally come with additional information for an error or warning).

Off the top of my head, the `full include list` message is unnecessary because it is what remains after all the adds/removes are acted on (and those are flagged as errors).

But the summary at the end makes sense as a closing message if nothing else has been emitted, as you say, so progress can be monitored. Can I ask for a colon to make it more diagnostic-like?

    /path/to/file.cc: has correct #includes/fwd-decls

svenpanne:
OK, I've uploaded a new version of this change:

- It prints the `... has correct ...` as a note, which makes e.g. emacs/flycheck/Code::Blocks/... etc. more happy.
- The full include list is really unnecessary, so let's remove it from the output when using clang-style format (yes, it is a bit inconsistent with the "classic" output, but on second thoughts, it is probably nicer).
- As an added bonus, strip away the decoration iwyu adds for lines which should be removed, so we present the user the line he actually wrote.

kimgr:
This looks great to me. @vsapsai: any further concerns?

kimgr:
(Oh, also, could you squash this into a single commit? Use `get rebase -i` and `git push -f` to overwrite the diff of this PR)

svenpanne:
OK, done. There is always the tension between easy reviews of patches and a sensible, compact history of  the upstream project. And finally there is `git bisect`, where small incremental patches are easier to handle when hunting down a bug. So there are various ways how projects handle this, and my question (for possible future patches) is: What is the preferred way for this project? Small changes and no rebases during review plus a final squash after the OK, right before merging? Just curious...

kimgr:
I like it that way, i.e. separate commits during review and then a final squash, especially if the proposed changes are relatively small and cohesive. Overall I hate merge commits, I find they just disrupt the history. But I'm sure there's a git expert out there who can explain to me why I'm wrong :-)

There are rare cases where I think several commits in a single PR makes sense, e.g. cleanup in one commit, functional change in another.

Your point about `bisect` is well taken, but we tend to try and keep PRs small enough that the need for micro-commits largely goes away. IMHO.

vsapsai:
Code looks fine to me. I don't use `iwyu_tool.py` regularly so don't know what edge cases to test. Hope that people using it regularly will catch the problems.

PR #376, opened 2016-11-23
Add tweaks for declval, forward and nullptr_t in libc++

EugeneZelenko:


kimgr:
I'd be happy to merge this, but I can't check that it's a strict improvement. Should we be testing mappings? I've been considering letting the test suite add libc++ mappings implicitly for FreeBSD and Mac targets, but I don't know if that's achievable.

kimgr:
@vsapsai, I think we should test this manually for now -- 

* Locally add libcxx.imp onto all test cases in run_iwyu_test.py/iwyu_test_util.py
* Run tests on the Mac (you) and FreeBSD (me)

If this is better or equivalent to the current behavior, I think it passes our best-effort testing for mappings and can be merged. Sounds OK to you?

kimgr:
On FreeBSD, applying the following patch:

diff --git a/iwyu_test_util.py b/iwyu_test_util.py
index dfe37b6..8ad8460 100755
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -425,7 +425,7 @@ def TestIwyuOnRelativeFile(test_case, cc_file, cpp_files_to_check,
   # variable, or by iwyu_flags, for easy debugging.  (We put the
   # envvar-based flag first, so user flags can override it later.)
   iwyu_flags = ['--verbose=%s' % os.getenv('IWYU_VERBOSE', '3')] + iwyu_flags
-
+  iwyu_flags += ['--mapping_file=libcxx.imp']
   # clang reads iwyu flags after the -Xiwyu clang flag: '-Xiwyu --verbose=6'
   iwyu_flags = ['-Xiwyu ' + flag for flag in iwyu_flags]

makes badinc.cc a little better (lots of the confusion around `<iosfwd>` and `<fstream>` vs `<string>` is resolved by the mapping.

The symbols added in this patch don't make any difference to the test suite.

I vote to merge.

vsapsai:
Tested and there are no regressions on macOS.

PR #378, opened 2016-11-28
Fix handling template creation as an array.

vsapsai:
It fixes case in `tests/cxx/badinc.cc` when for

    new I1_TemplateClass<I2_Class, I1_Struct>[kI1ConstInt]

we erroneously don't need full declaration for `I1_Struct`.

Clang r283406 improved handling array creation which caused us to see
not template type but array type. The fix is to take element type in
this case.

Another attempted approach is

    GetTypeOf(constructor_expr->getConstructor()->getParent());

But it doesn't work because in this case we receive not template type
but instantiated RecordType that has no template arguments. For example,
with mentioned code we get type like

    RecordType 0x7fb7ccc251f0 'class I1_TemplateClass<class EmptyClass, struct Fail_Struct>'
    `-ClassTemplateSpecialization 0x7fb7ccc250f0 'I1_TemplateClass'

instead of desired

    TemplateSpecializationType 0x7fb7ccc253f0 'I1_TemplateClass<class EmptyClass, struct Fail_Struct>' sugar I1_TemplateClass
    |-TemplateArgument type 'class EmptyClass'
    |-TemplateArgument type 'struct Fail_Struct'
    `-RecordType 0x7fb7ccc251f0 'class I1_TemplateClass<class EmptyClass, struct Fail_Struct>'
      `-ClassTemplateSpecialization 0x7fb7ccc250f0 'I1_TemplateClass'

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1027,6 +1027,14 @@ const Type* GetTypeOf(const Expr* expr) {
   return expr->getType().getTypePtr();
 }
 
+const Type* GetTypeOf(const CXXConstructExpr* expr) {
+  const Type* type = expr->getType().getTypePtr();
+  if (const clang::ArrayType* array_type = type->getAsArrayTypeUnsafe()) {
+    type = array_type->getElementType().getTypePtr();
+  }
+  return type;
+}
+
 const Type* GetTypeOf(const ValueDecl* decl) {
   return decl->getType().getTypePtr();
 }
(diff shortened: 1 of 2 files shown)


kimgr:
Looks good!

kimgr:
Tests run as expected on FreeBSD, and the I1_Struct warning is gone.

On Windows, LLVM now requires VS2015, so I had to rebuild with a whole new compiler and thereby a new standard lib, which affects the test output (standard headers have changed names). I found one reference to I1_Struct:

    tests/cxx/badinc.cc:1468: Unexpected diagnostic:
    I1_Struct needs a declaration, but does not provide or directly #include one.

Not sure if that's a result of your patch or not. I can try and rebuild master and check.

kimgr:
Oh, that went quicker than expected. Without your patch, there's also:

    tests/cxx/badinc.cc:1626: Unmatched regex:
    I1_Struct is...*badinc-i1.h

So the expectation failure on 1468 is probably some other MSVC problem.

Ship it!

vsapsai:
Thanks for the quick review.

PR #379, opened 2016-11-30
fix a bug

xuzhen1994:
Filepath may contains the symbol ":" in Windows, and the line break is different.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -286,10 +286,10 @@ class IWYUOutputParser(object):
   # #includes and forward-declares should be added to the output file,
   # what should be removed, and what the end result is.  The first line
   # of each section also has the filename.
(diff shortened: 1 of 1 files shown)


kimgr:
Thanks!

I'm surprised your changes work, though...

Won't `(.*)` eagerly consume the trailing `:`? I would expect you'd need non-greedy matching here.

Also, doesn't `$` match `\r\n` on Windows and `\n` on Unix?

Bu11etmagnet:
Maybe he's running iwyu compiled on Windows from a Python script running on Unix?

kimgr:
I completely misread this patch -- I thought it was for iwyu_tool.py.

fix_includes.py, on the other hand, has tests -- can you add test cases for the stuff that doesn't work prior to your patch?

kimgr:
No input for more than six months. I've added a new PR with tests based on this one. Closing.

PR #380, opened 2016-12-03
Ignore friend decls when looking for definition-as-written

kimgr:
Here's a preliminary patch to fix the test failures in master.

I'm not sure if this solves a problem on our end, or works around a bug in Clang. I've talked to the author of the Clang patch, and we haven't yet been able to figure it out together,

I think we're at the point where it's more useful for us to have the test suite green than correct :-/

Let me know if this is OK to merge, otherwise we'll have to keep digging to get to the root of the issue, and I haven't been able to spare more than an hour at a time since this broke back in October.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -846,9 +846,14 @@ const NamedDecl* GetDefinitionAsWritten(const NamedDecl* decl) {
     if (const ClassTemplateDecl* tpl_decl = DynCastFrom(decl))
       decl = tpl_decl->getTemplatedDecl();  // convert back to CXXRecordDecl
   } else if (const FunctionDecl* func_decl = DynCastFrom(decl)) {
-    if (const FunctionDecl* tpl_pattern =
-        func_decl->getTemplateInstantiationPattern())
-      decl = tpl_pattern;
+    // If we're instantiated from a template, use the template pattern as the
+    // decl-as-written.
+    // But avoid friend declarations in templates, something happened in Clang
+    // r283207 that caused them to form a dedicated redecl chain, separate
+    // from all other redecls.
+    const FunctionDecl* tp_decl = func_decl->getTemplateInstantiationPattern();
+    if (tp_decl && tp_decl->getFriendObjectKind() == Decl::FOK_None)
+      decl = tp_decl;
   }
   // Then, get to definition.
   if (const NamedDecl* class_dfn = GetDefinitionForClass(decl)) {


kimgr:
Pushed a new version with simpler implementation and better commit message.

iwyu_ast_util.cc
```diff
@@ -846,9 +846,14 @@ const NamedDecl* GetDefinitionAsWritten(const NamedDecl* decl) {
...
   } else if (const FunctionDecl* func_decl = DynCastFrom(decl)) {
-    if (const FunctionDecl* tpl_pattern =
-        func_decl->getTemplateInstantiationPattern())
-      decl = tpl_pattern;
+    // If we're instantiated from a template, use the template pattern as the
+    // decl-as-written.
+    // But avoid friend declarations in templates, something happened in Clang
+    // r283207 that caused them to form a dedicated redecl chain, separate
+    // from all other redecls.
+    const FunctionDecl* tp_decl = func_decl->getTemplateInstantiationPattern();
```
vsapsai: The only concern is `tp_decl` naming. Should it be `tpl` instead? But we already have `const ClassTemplateDecl* tpl_decl` above so maybe keeping `tpl_pattern` is better but it can be causing more line wrapping.

vsapsai:
The patch looks good to me, it fixes the problem. And I agree with having a bandaid without waiting for Clang to revert this change.

Do you know reasons for this change in Clang? It would be nice to include in commit message another point of view because to me this
> foo();  // attributed to Template&lt;T&gt;:foo

part looks completely broken and it's unclear why is this not reverted.

And small correction, there is 1 missing `:` before `foo` in the comment.

kimgr:
No, I don't have a very good understanding of this.

The Clang commit added support for friend definitions in templates (something that looks really weird to begin with):

    void f();  // declared here

    template<class T>
    class Template {
      // defined here!
      friend void f() {
      }
    }:

The fact that `foo` in the commit message example is attributed to `Template<T>::foo` is an artefact of IWYU's behavior -- `GetDefinitionAsWritten` looks for the 'best' declaration to attribute a use to.

Something changed in Clang's building of redecl chains, but I haven't been able to put my finger on exactly what, and what the correct behavior should be.

I thought I'd bring a reduced example up on the Clang list and get some feedback on what we're doing and how Clang works.

Thanks for the review, I'll fix the nit.

iwyu_ast_util.cc
kimgr: Right, I shortened the name to avoid line wrapping, it made it much easier to read.

I thought of `tp_decl` as an effective abbreviation of `template pattern decl`. That name is so short-lived that I don't worry too much about naming.

vsapsai:
Thanks for explanation. I have no comments then. Ship it.

Issue #365, opened 2016-10-11
crash when forgetting "  like here:  // IWYU pragma: private, include header.h"


gunrot:
crash when forgetting " like here: // IWYU pragma: private, include header.h"

kimgr:
Should be fixed on master now.

PR #381, opened 2016-12-05
Better validate pragma syntax

kimgr:
IWYU pragma: private, include "foo" and
IWYU pragma: no_include "foo" both require quoted includes for their
arguments, but only had assertions to validate.

Add more explicit error messages for invalid syntax in these pragmas.

Fixes issue #365.


vsapsai:
Everything is great. Code looks fine, tests are passing, cannot find other places where we need to add a warning. Thanks for improving pragma validation.

kimgr:
Thanks, it's live!

PR #385, opened 2017-01-01
Handle CK_ZeroToOCLQueue cast kind

kimgr:
Added in Clang in r290431, and handled here to silence warning.


kimgr:
I'm guessing this is non-controversial, merging now.

Issue #384, opened 2016-12-26
Build fails after Clang r289250


vsapsai:
Build is failing during tests with

```
======================================================================
FAIL: runTest (__main__.badinc)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/mnt/buildbot_iwyu_trunk/workspace/sources/iwyu/run_iwyu_tests.py", line 244, in <lambda>
    {'runTest': lambda self, f=filename: self.RunOneTest(f),
  File "/mnt/buildbot_iwyu_trunk/workspace/sources/iwyu/run_iwyu_tests.py", line 215, in RunOneTest
    iwyu_flags, clang_flags, verbose=True)
  File "/mnt/buildbot_iwyu_trunk/workspace/sources/iwyu/iwyu_test_util.py", line 455, in TestIwyuOnRelativeFile
    test_case.assertTrue(not failures, ''.join(failures))
AssertionError: 
tests/cxx/badinc.cc:1500: Unmatched regex:
I1_Class is...*badinc-i1.h

tests/cxx/badinc.cc:1506: Unmatched regex:
I1_const_ptr is...*badinc-i1.h

tests/cxx/badinc.cc:1767: Unexpected diagnostic:
std::vector<float, std::allocator<float> >::const_iterator is defined in <vector>, which isn't directly #included.

tests/cxx/badinc.cc:1788: Unexpected diagnostic:
std::vector<float, std::allocator<float> >::reverse_iterator is defined in <vector>, which isn't directly #included.

tests/cxx/badinc.cc:1801: Unexpected diagnostic:
std::vector<float, std::allocator<float> >::const_reverse_iterator is defined in <vector>, which isn't directly #included.

Unexpected summary diffs for tests/cxx/badinc.cc:
+++ 
@@ -37,7 +37,7 @@
 #include <fstream>  // for fstream
 #include <list>  // for list
 #include <new>  // for operator new
-#include <string>  // for allocator, basic_string, basic_string<>::iterator, char_traits, operator+, string
+#include <string>  // for allocator, basic_string, basic_string<>::iterator, operator+, string
 #include <typeinfo>  // for type_info
[...]

PR #386, opened 2017-01-01
Remove incorrect test case assumption on desugared types

kimgr:
Clang r289250 caused more type sugar to remain in the AST. That, in turn, caused
derived_function_tpl_args to fail, because we had a compensating assertion
(marked with a TODO indicating that it was undesirable.)

Now that type sugar survives, we won't see a need for IndirectClass
here. Rather, the use will be attributed to the LocalClass typedef, which exists
in the same file. No diagnostic expected.

(Fixes part of the tracking bug #384)


vsapsai:
The change is fine and testing proves it.

PR #387, opened 2017-01-01
Pick up more sugar for function arguments

kimgr:
Type sugar (typedefs, more specifically) were previously lost for
arguments passed by value.

Introduce a new helper function to glean sugared type info from
ImplicitCastExpr nodes in the AST, as present in later revs of Clang.

Fixes a TODO in derived_function_tpl_args.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -719,6 +719,23 @@ GetTplTypeResugarMapForFunctionExplicitTplArgs(
   return retval;
 }
 
+// Get the type of an expression while preserving as much type sugar as
+// possible. This was originally designed for use with function argument
+// expressions, and so might not work in a more general context.
+static const Type* GetSugaredTypeOf(const Expr* expr) {
+  // First, try to find an ImplicitCastExpr under the expr, and let that provide
+  // the type. This has a higher probability of yielding a sugared type.
+  for (const Stmt* child_expr : expr->children()) {
+    if (const auto* cast_expr = dyn_cast<ImplicitCastExpr>(child_expr)) {
+      return cast_expr->getType().getTypePtr();
+    }
+  }
+
+  // If we didn't find a type via ImplicitCastExpr, just return the type of the
+  // expr itself.
+  return GetTypeOf(expr);
+}
+
 map<const Type*, const Type*> GetTplTypeResugarMapForFunction(
     const FunctionDecl* decl, const Expr* calling_expr) {
   map<const Type*, const Type*> retval;
@@ -788,7 +805,7 @@ map<const Type*, const Type*> GetTplTypeResugarMapForFunction(
   //                 under it, take the pre-cast type instead?
   set<const Type*> fn_arg_types;
   for (unsigned i = 0; i < num_args; ++i) {
-    const Type* argtype = GetTypeOf(fn_args[i]);
+    const Type* argtype = GetSugaredTypeOf(fn_args[i]);
     // TODO(csilvers): handle RecordTypes that are a TemplateSpecializationDecl
     InsertAllInto(GetComponentsOfType(argtype), &fn_arg_types);
   }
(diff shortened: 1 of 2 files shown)


iwyu_ast_util.cc
```diff
@@ -719,6 +719,37 @@ GetTplTypeResugarMapForFunctionExplicitTplArgs(
...
+// possible. This was originally designed for use with function argument
+// expressions, and so might not work in a more general context.
+static const Type* GetSugaredTypeOf(const Expr* expr) {
+  // First, try to find an ImplicitCastExpr under the expr, and let that provide
+  // the type. This has a higher probability of yielding a sugared type.
+  struct Visitor : public RecursiveASTVisitor<Visitor> {
+    Visitor() : cast_type(nullptr) {}
+
+    bool VisitImplicitCastExpr(ImplicitCastExpr* cast_expr) {
+      VERRS(6) << "Seeing ImplicitCastExpr: " << PrintableStmt(cast_expr)
```
kimgr: Remove this log.

vsapsai:
Isn't RecursiveASTVisitor too blunt tool in this case? I am thinking if it will be thrown off by other ImplicitCastExpr present in expression. For example,

```c++
class Number {
public:
  Number(int x) {}
};
class Foo {
public:
  Foo(Number number) {}
};
void bar(Foo foo) {}

void bazz() {
  bar(Foo(3));  // <- multiple ImplicitCastExpr
}
```

There are no templates here but I think the idea with multiple ImplicitCastExpr is clear.

kimgr:
I think you're right, the recursive visitation is unnecessary and wrong. See new revision -- much simpler!

I couldn't get your example to produce more than one ImplicitCastExpr, but it was littered with CXXConstructExprs. Anyway, this new routine is much nicer without the RAV.

I've only run tests on Windows, and derived_function_tpl_args succeeds. I have a lot of other platform-dependent failures, so it's hard to see if I've introduced anything new. Can you give it a shot on Ubuntu/GCC?

Thanks!

vsapsai:
There are no regressions on Ubuntu and macOS. But I think I can have an example to break your code. Currently checking all expression children seems suspicious. Maybe it would work if we are kind of collapsing a part of the tree where you have only single child. For example, if you have

```
x
`-ImplicitCastExpr
   `-y
     |-foo
     `-bar
```

you can use "y" instead of "x" but not if

```
x
|-ImplicitCastExpr
| `-y
`-z
```

Did you check how `GetComponentsOfType` behaves for this case? At first it looked promising but after some thought I'm not so sure. Would be glad to hear what you think about it.

kimgr:
Thanks for working through this!

I'm having a hard time coming up with code examples to produce the AST fragments you're proposing. I see what you're saying in the abstract, but I can't tell if there's a way to get such ASTs from valid code.

As for `GetComponentsOfType`, it seems to do something different -- `GetTypeOf` and `GetSugaredTypeOf` return the type of an expression, where `GetComponentsOfType` starts from a `Type` -- I can't quite see what you had in mind?

I think you're right that there are a large number of cases for which this doesn't work. But it might be interesting as a small improvement to the status quo, unless it introduces false positives elsewhere.

vsapsai:
I thought `GetComponentsOfType` could be useful because it drills down into a type and returns not only immediate type but also inner components. But I didn't come up with any usage for this problem, so it's just useless suggestion.

I also wasn't able to come up with AST I wanted. I wasn't able to trick your code into using incorrect `ImplicitCastExpr` because in my attempts there was `MaterializeTemporaryExpr` node preventing it from happening. I cannot find any real problems with this change, please merge it.

kimgr:
Thanks! Yes, I also kept getting `MaterializeTemporaryExpr` nodes in the way... I think we'll need to revisit this once we learn more about actual ASTs in the wild, but I think it's a strict improvement for now.

PR #388, opened 2017-01-03
Adjust badinc test for improved Clang behavior

kimgr:
badinc.cc would have several intentional misattributions for iterators to
<vector>. Clang changed in r289250 to better preserve type sugar, which caused
these tests to produce more correct diagnostics. Test assertions were now
protecting broken behavior, and failing.

Adjust assertions to reflect goodness.

### DIFF
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -1765,6 +1765,8 @@ int main() {
   const std::vector<float>::const_iterator float_constit = float_vector.begin();
   // IWYU: std::vector<.*>::const_iterator is...*<vector>
   (void)(float_it == float_constit);
+  // IWYU: std::vector<.*>::const_iterator is...*<vector>
+  (void)(float_constit == float_it);
   // IWYU: std::vector is...*<vector>
   // IWYU: std::vector<.*>::const_iterator is...*<vector>
   std::vector<float>::const_iterator float_forit;
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -1763,7 +1763,7 @@ int main() {
   // IWYU: std::vector is...*<vector>
   // IWYU: std::vector<.*>::const_iterator is...*<vector>
   const std::vector<float>::const_iterator float_constit = float_vector.begin();
-  // IWYU: std::vector is...*<vector>
+  // IWYU: std::vector<.*>::const_iterator is...*<vector>
   (void)(float_it == float_constit);
   // IWYU: std::vector is...*<vector>
   // IWYU: std::vector<.*>::const_iterator is...*<vector>
@@ -1785,8 +1785,8 @@ int main() {
            // IWYU: std::vector is...*<vector>
            float_reverse_it = float_vector.rbegin();
        // IWYU: std::vector is...*<vector>
+       // IWYU: std::vector<.*>::reverse_iterator is...*<vector>
        float_reverse_it != float_vector.rbegin();
-       // IWYU: std::vector is...*<vector>
        // IWYU: std::vector<.*>::reverse_iterator is...*<vector>
        ++float_reverse_it) ;
   // IWYU: std::vector is...*<vector>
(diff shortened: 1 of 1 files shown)


kimgr:
Based on @vsapsai's excellent analysis in #384, it looks right to just update the test assertions.

vsapsai:
I was also thinking about moving bunch of iterator-related test code out of badinc.cc. Something like [this](https://github.com/vsapsai/include-what-you-use/commit/d8151b45199fe913c7b7592ee9033cfab36e68c0). Another idea is to harden test with comparing iterators `float_constit`, `float_it` in different order like [here](https://github.com/vsapsai/include-what-you-use/commit/6f738664e4691f1bece2b53cfa0755728f6f6cc1). And `// IWYU: std::vector is...*<vector>` on line 1766 has some complications. I plan to discuss that in #384.

kimgr:
It makes it easier to troubleshoot tests, so from that perspective I'm all for.

But it would be nice to get the test suite green first, and then start shuffling around, maybe?

vsapsai:
I agree to extract iterator tests later, especially that doing it in naive way causes broken test. Still think that having

`(void)(float_constit == float_it);`

will be useful.

kimgr:
OK, done. Assertion holds. Good to go?

vsapsai:
`// IWYU: std::vector is...*<vector>` is still required for calling operators on iterators for libstdc++. Tested on Ubuntu 16.04 and we need `// IWYU: std::vector is...*<vector>` for lines 1767, 1769, 1793, 1806. All line numbers are after applying change.

Investigated it a little bit and it happens because [in `ReportDeclUseWithComment`](https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L1647-L1648) `MapPrivateDeclToPublicDecl` maps `operator==` to `std::vector` like

`template<> inline bool operator==<const float *, std::vector<float, std::allocator<float> >>(const __normal_iterator<const float *, std::vector<float, std::allocator<float> > > &__lhs, const __normal_iterator<const float *, std::vector<float, std::allocator<float> > > &__rhs)`

to

`template<> class vector<float, std::allocator<float>> : protected _Vector_base<float, std::allocator<float> >`

Didn't really investigate how to fix it, I think we can do this after fixing the build. Preliminary idea is to map `__normal_iterator` not to `std::vector` but to `std::vector::iterator`.

kimgr:
OK, I'll try to get rid of the std::vector removals then... I have some network problems at home, so I'll see if I can get in touch with my dev machine again..

kimgr:
There, could you test this on your Ubuntu rig? badinc.cc now fails for all the vector assertions on FreeBSD.

vsapsai:
This way it works. Output on Ubuntu is
```
AssertionError:
tests/cxx/badinc.cc:1500: Unmatched regex:
I1_Class is...*badinc-i1.h

tests/cxx/badinc.cc:1506: Unmatched regex:
I1_const_ptr is...*badinc-i1.h

+++
@@ -37,7 +37,7 @@
 #include <fstream>  // for fstream
 #include <list>  // for list
 #include <new>  // for operator new
-#include <string>  // for allocator, basic_string, basic_string<>::iterator, char_traits, operator+, string
+#include <string>  // for allocator, basic_string, basic_string<>::iterator, operator+, string
 #include <typeinfo>  // for type_info
 #include "tests/cxx/badinc-d1.h"  // for D1CopyClassFn, D1Function, D1_Class, D1_CopyClass, D1_Enum, D1_Enum::D11, D1_I1_Typedef, D1_StructPtr, D1_Subclass, D1_TemplateClass, D1_TemplateStructWithDefaultParam, MACRO_CALLING_I4_FUNCTION
 #include "tests/cxx/badinc-d4.h"  // for D4_ClassForOperator, operator<<
---
```

kimgr:
OK, great! I'll merge this then? The two unmatched regexes are related to the SubstTemplateTypeParmType disappearing.

I haven't started looking into why char_traits shows up on the `<string>` include line.

vsapsai:
Yes, merge please.

PR #391, opened 2017-01-08
Add dev script for log scrubbing

kimgr:
I've been comparing so many logs lately that I've begun to settle on a repeatable procedure.

Once component in that procedure is this script, which scrubs irrelevant details from logs, to make it possible to diff the essence. Should be generally useful, so I'll post it here.

Let me know what you think.

### DIFF
--- a/scrub-logs.py
+++ b/scrub-logs.py
@@ -26,7 +26,7 @@ def strip_path_prefix(line):
 
 
 def strip_addrs(line):
-    line = re.sub(r'\b(0x)?[0-9A-Fa-f]{6,16}\b', '', line)
+    line = re.sub(r'\b(0x)?[0-9A-Fa-f]{6,16}', '', line)
     return line
 
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Include What You Use #
 
-This README was generated on 2016-02-01 00:55:37 UTC.
+This README was generated on 2017-01-16 18:38:58 UTC.
 
 For more in-depth documentation, see http://github.com/include-what-you-use/include-what-you-use/tree/master/docs.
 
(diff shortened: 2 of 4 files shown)


scrub-logs.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python
+
+##===---------- scrub-logs.py - generate README from Wiki sources ---------===##
```
vsapsai: Usually having dashes in Python file names is undesirable because you won't be able to use the file as a module. But I believe here it should be fine as it's not supposed to be used as a library.

scrub-logs.py
```diff
@@ -0,0 +1,45 @@
...
+import sys
+
+
+def strip_path_prefix(line):
+    line = re.sub(r'<.*(llvm[\\/]tools[\\/].*):', r'<\1:', line)
+    return line
+
+
+def strip_addrs(line):
+    line = re.sub('\s(0x)?[0-9A-Fa-f]{6,9}', '', line)
```
vsapsai: On macOS I have addresses like 0x7fe1d81b3e10 which aren't caught by your regexp. The tricky part is it will apply to the part of the address leaving it in inconsistent state. And I don't know if it is better to remove addresses entirely or to replace with something generic. I was replacing it with 0xxxxx, for instance.

scrub-logs.py
kimgr: Oh, yeah. I usually use that naming when I want to discourage imports, because it's really just a tool.

scrub-logs.py
kimgr: Oops, that's a big one :-) I was sloppy and kept adding hex digits until it worked with my logs. But 64-bit numbers should max out at 16 hex digits, right?

I think it's better to remove them entirely, since that leaves more horizontal space for useful information.

kimgr:
Sorry, I missed your comments until now. I'll try to clean this up and post a new version.

kimgr:
There. I think this can be improved infinitely, but I also think it's a good, useful start.

kimgr:
I actually stumbled on a case where `\b` did not work -- Clang's AST dumper writes out `0x12345678nullptr` in a single word at some point, which is arguably a Clang bug. So I eventually opted for just a leading `\b`, I'll post a new rev.

PR #394, opened 2017-01-16
How to Run: Redirect stderr to /tmp/iwyu.out

mineo:
The messages from include-what-you-use are printed on stderr, so redirect that
to the log file.


kimgr:
This looks good to me, thanks! I'll merge this as-is.

PR #399, opened 2017-01-22
Fix detecting type responsible for operator function for function templates

vsapsai:
Relevant discussion can be found in issue [#384: Build fails after Clang r289250](https://github.com/include-what-you-use/include-what-you-use/issues/384).

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1272,12 +1272,14 @@ const Expr* GetFirstClassArgument(CallExpr* expr) {
       return expr->getArg(0);
     }
     // Handle free functions.
+    CHECK_(callee_decl->getNumParams() == expr->getNumArgs() &&
+        "Require one-to-one match between call arguments and decl parameters");
     int params_count = callee_decl->getNumParams();
     for (int i = 0; i < params_count; i++) {
-      // View argument types from the perspective of function body, not from
-      // the caller's perspective.  For example, function parameter can have
-      // template type but function argument is not necesserily a template
-      // when the function is called.
+      // View argument types from the perspective of function declaration,
+      // not from the caller's perspective.  For example, function parameter
+      // can have template type but function argument is not necessarily
+      // a template when the function is called.
       const Type* param_type = GetTypeOf(callee_decl->getParamDecl(i));
       param_type = RemovePointersAndReferencesAsWritten(param_type);
       // Make sure we do the right thing given a function like
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1266,20 +1266,33 @@ const Type* TypeOfParentIfMethod(const CallExpr* expr) {
 }
 
 const Expr* GetFirstClassArgument(CallExpr* expr) {
-  for (CallExpr::arg_iterator it = expr->arg_begin();
-       it != expr->arg_end(); ++it) {
-    const Type* argtype = GetTypeOf(*it);
-    // Make sure we do the right thing given a function like
-    //    template <typename T> void operator>>(const T& x, ostream& os);
-    // In this case ('myclass >> os'), we want to be returning the
-    // type of os, not of myclass, and we do, because myclass will be
-    // a SubstTemplateTypeParmType, not a RecordType.
-    if (isa<SubstTemplateTypeParmType>(argtype))
-      continue;
-    argtype = argtype->getUnqualifiedDesugaredType();  // see through typedefs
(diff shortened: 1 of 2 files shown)


iwyu_ast_util.cc
```diff
@@ -1266,20 +1266,33 @@ const Type* TypeOfParentIfMethod(const CallExpr* expr) {
...
-      return *it;
+  if (const FunctionDecl* callee_decl = expr->getDirectCallee()) {
+    if (isa<CXXMethodDecl>(callee_decl)) {
+      // If a method is called, return 'this'.
+      return expr->getArg(0);
+    }
+    // Handle free functions.
+    int params_count = callee_decl->getNumParams();
+    for (int i = 0; i < params_count; i++) {
+      // View argument types from the perspective of function body, not from
```
kimgr: s/body/declaration/

iwyu_ast_util.cc
```diff
@@ -1266,20 +1266,33 @@ const Type* TypeOfParentIfMethod(const CallExpr* expr) {
...
+      // the caller's perspective.  For example, function parameter can have
+      // template type but function argument is not necesserily a template
```
kimgr: typo: necesserily

iwyu_ast_util.cc
```diff
@@ -1266,20 +1266,33 @@ const Type* TypeOfParentIfMethod(const CallExpr* expr) {
...
+      // when the function is called.
+      const Type* param_type = GetTypeOf(callee_decl->getParamDecl(i));
+      param_type = RemovePointersAndReferencesAsWritten(param_type);
+      // Make sure we do the right thing given a function like
+      //    template <typename T> void operator>>(const T& x, ostream& os);
+      // In this case ('myclass >> os'), we want to be returning the
+      // type of os, not of myclass, and we do, because myclass will be
+      // a SubstTemplateTypeParmType, not a RecordType.
+      if (isa<SubstTemplateTypeParmType>(param_type))
```
kimgr: Is this path ever hit now? It's a bit surprising that a parameter (as opposed to an argument) would be a `SubstTemplateTypeParmType`. I would have expected STTPT to only show up in the AST when instantiating a template/invoking a function template, but it's hard to tell from the docs what the respective nodes mean.

iwyu_ast_util.cc
```diff
@@ -1266,20 +1266,33 @@ const Type* TypeOfParentIfMethod(const CallExpr* expr) {
...
-    if (isa<RecordType>(argtype) ||
-        isa<TemplateSpecializationType>(argtype)) {
-      return *it;
+  if (const FunctionDecl* callee_decl = expr->getDirectCallee()) {
+    if (isa<CXXMethodDecl>(callee_decl)) {
+      // If a method is called, return 'this'.
+      return expr->getArg(0);
+    }
+    // Handle free functions.
+    int params_count = callee_decl->getNumParams();
```
kimgr: Can we put in a CHECK_ here that params_count == expr's arg count?

kimgr: ... and actually, is that a strong invariant? I would expect things like:

    void f(int a, int b=0);

    f(10);

to break that.

Can't think of a case with operators and default arguments, so not sure if it's relevant.

vsapsai: Added CHECK_. And regarding default arguments, I've tried it and hit a compiler error

```
tests/cxx/repo.cpp:28:39: error: parameter of overloaded 'operator==' cannot have a default argument
bool operator==(Cc_Class& object, int x = 0) {
                                      ^   ~
```

iwyu_ast_util.cc
vsapsai: Changed.

iwyu_ast_util.cc
vsapsai: Fixed. Thanks for catching this.

iwyu_ast_util.cc
vsapsai: This is still required. Tried to remove it and GetFirstClassArgument wasn't doing its job.

vsapsai:
Addressed the comments, please check again.

iwyu_ast_util.cc
kimgr: OK, let's leave it at that :-)

kimgr:
Looks good to me, please merge.

vsapsai:
Merged manually 4e957bf33a337f5fe5e275a354cfc46650e11ac2 and 1007562363756627cdab664124e37669dfbf0c6b.

PR #390, opened 2017-01-07
WIN32: Link against version.dll to fix link error

nocnokneo:
Fixes three missing symbols linker errors:

- `GetFileVersionInfoSizeW`
- `GetFileVersionInfoW`
- `VerQueryValueW`

Fixes the following build error:

```
[14/14] Linking CXX executable include-what-you-use.exe
FAILED: include-what-you-use.exe
[...]

PR #400, opened 2017-01-23
Add /bigobj for MSVC builds

kimgr:
Inspired by PR #390, apparently /bigobj can be necessary for MSVC too.

We enabled extended object format to work around 'too many sections' for
MinGW, but not for MSVC.

Do so now.


vsapsai:
Looks good to me.

PR #403, opened 2017-01-26
Adjust CMakeLists for out-of-tree builds

kimgr:
PR #390 showed that IWYU doesn't actually build out-of-tree with Clang
3.9, as advertised.

The PR only added a link dependency on version.lib, but there were several
other problems:

- Warnings disabled by Clang/LLVM that we hadn't picked up
- Extended object format needed to be enabled (/bigobj)
- Sane exception handling (/EHsc) needed to explicitly enabled

I noticed these when building with MSVC 2015 for 64-bit, so some may be
sparked by that configuration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,8 +45,10 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
       -D_SCL_SECURE_NO_WARNINGS
 
       # Disabled warnings.
+      -wd4141 # Suppress ''modifier' : used more than once' (because of __forceinline combined with inline)
       -wd4146 # Suppress 'unary minus operator applied to unsigned type, result still unsigned'
       -wd4244 # Suppress ''argument' : conversion from 'type1' to 'type2', possible loss of data'
+      -wd4267 # Suppress ''var' : conversion from 'size_t' to 'type', possible loss of data'
       -wd4291 # Suppress ''declaration' : no matching operator delete found; memory will not be freed if initialization throws an exception'
       -wd4345 # Suppress 'behavior change: an object of POD type constructed with an initializer of the form () will be default-initialized'
       -wd4355 # Suppress ''this' : used in base member initializer list'
@@ -109,6 +111,9 @@ if( MSVC )
     -D_HAS_EXCEPTIONS=0
   )
 
+  # Enable bigobj-support and sane C++ exception semantics.
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj /EHsc")
+
   # Put project in solution folder
   set_target_properties(include-what-you-use
     PROPERTIES FOLDER "Clang executables"
@@ -163,6 +168,7 @@ target_link_libraries(include-what-you-use
 if( WIN32 )
   target_link_libraries(include-what-you-use
     shlwapi
+    version # For clangDriver's MSVCToolchain
   )
 elseif( UNIX )
   include(FindCurses)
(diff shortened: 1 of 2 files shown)


nocnokneo:
Works for me! Built with:

```bash
cmake -GNinja -DCMAKE_BUILD_TYPE:STRING=Release -DIWYU_LLVM_ROOT_PATH:PATH=~/opt/LLVM-master ../include-what-you-use
ninja
```

Had to make sure that the CMAKE_BUILD_TYPE matched that used for LLVM or else I get linking errors like:

```
LLVMSupport.lib(circular_raw_ostream.cpp.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in iwyu.cc.obj
LLVMSupport.lib(circular_raw_ostream.cpp.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MDd_DynamicDebug' in iwyu.cc.obj
LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library
```

But I don't think there's any good way around that.

kimgr:
Ah, yes. Installed LLVM libs are probably release builds, so that needs to be matched. I'll add something to the docs about that and include in this PR.

kimgr:
Done. @vsapsai, can you take a look?

kimgr:
Also, how do we want to handle "historical" releases?

At the stage IWYU is in it feels better to focus on the future. But given how interconnected IWYU and Clang are (via the builtin headers), having to force people to upgrade Clang to be able to use the latest IWYU also seems a bit harsh.

That said, maintaining more than trunk tends to consume so much resources that nothing really gets done, so I'm reluctant to do it.

Thoughts?

kimgr:
Here's the corresponding diff for master: https://github.com/kimgr/include-what-you-use/commit/2fb8a2816384ae1cae24e03f301973b641ce1ae7

I'll test that separately.

CMakeLists.txt
```diff
@@ -109,6 +111,9 @@ if( MSVC )
     -D_HAS_EXCEPTIONS=0
   )
 
+  # Enable bigobj-support and sane C++ exception semantics.
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj /EHsc")
```
vsapsai: Reminder to clean up duplication with b5d3e7250a96df6627530b6d8e79c7945a95dcbb.

CMakeLists.txt
```diff
@@ -163,6 +168,7 @@ target_link_libraries(include-what-you-use
 if( WIN32 )
   target_link_libraries(include-what-you-use
     shlwapi
+    version # For clangDriver's MSVCToolchain
```
vsapsai: I assume WIN32 is just a bad name and actually covers 64-bit as well, right?

docs/InstructionsForUsers.md
```diff
@@ -35,7 +35,7 @@ In an out-of-tree configuration, we assume you already have compiled LLVM and Cl
         iwyu-trunk/include-what-you-use$ git checkout clang_3.2
         iwyu-trunk/include-what-you-use$ cd ..
 
-  * Create a build root and use CMake to generate a build system linked with LLVM/Clang prebuilts, e.g.
+  * Create a build root and use CMake to generate a build system linked with LLVM/Clang prebuilts. Note that CMake settings need to match exactly, so you may need to add `-DCMAKE_BUILD_TYPE=Release` or more to the command-line below:
```
vsapsai: Reminder to regenerate README.md.

vsapsai:
I would avoid backporting fixes, we don't really have capacity for that. We can do that on case-by-case basis, for example when new IWYU+LLVM+Clang require new MSVC version and old IWYU version is horribly broken. But I think that should be an exception, not a rule.

kimgr:
I agree. Posted PR #404 in place of this one. Closing.

CMakeLists.txt
nocnokneo: Right.

```
$ cmake --help-variable WIN32
WIN32
-----

``True`` on Windows systems, including Win64.

Set to ``true`` when the target system is Windows.
```

PR #404, opened 2017-01-29
Fix out-of-tree build for MSVC

kimgr:
We don't build with MSVC out-of-tree regularly, so a number of
compatibility issues with LLVM had snuck in:

- Suppressed warnings in header LLVM files
- Extended object format support (/bigobj)
- Use of noexcept requires sane C++ exception semantics (/EHSc)
- Clang's MSVCToolChain.cpp now uses functions from version.dll, so add it
  to linker deps

While dabbling with this, we also found it was necessary to CMake with
-DCMAKE_BUILD_TYPE=Release to match LLVM's release-built libraries.
Mention this in the user docs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,8 +45,10 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
       -D_SCL_SECURE_NO_WARNINGS
 
       # Disabled warnings.
+      -wd4141 # Suppress ''modifier' : used more than once' (because of __forceinline combined with inline)
       -wd4146 # Suppress 'unary minus operator applied to unsigned type, result still unsigned'
       -wd4244 # Suppress ''argument' : conversion from 'type1' to 'type2', possible loss of data'
+      -wd4267 # Suppress ''var' : conversion from 'size_t' to 'type', possible loss of data'
       -wd4291 # Suppress ''declaration' : no matching operator delete found; memory will not be freed if initialization throws an exception'
       -wd4345 # Suppress 'behavior change: an object of POD type constructed with an initializer of the form () will be default-initialized'
       -wd4355 # Suppress ''this' : used in base member initializer list'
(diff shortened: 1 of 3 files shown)


kimgr:
Supersedes #390 and #403.

This is for master. Maintainers prefer not to backport this to IWYU 0.7 (for Clang 3.9). Users who need it can apply the patch from PR #403.

CMakeLists.txt
```diff
@@ -112,6 +114,9 @@ if( MSVC )
     -D_HAS_EXCEPTIONS=0
   )
 
+  # Enable bigobj support and sane C++ exception semantics.
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj /EHsc")
```
vsapsai: Reminder to clean up duplication with b5d3e7250a96df6627530b6d8e79c7945a95dcbb.

docs/InstructionsForUsers.md
```diff
@@ -35,7 +35,7 @@ In an out-of-tree configuration, we assume you already have compiled LLVM and Cl
         iwyu-trunk/include-what-you-use$ git checkout clang_3.2
         iwyu-trunk/include-what-you-use$ cd ..
 
-  * Create a build root and use CMake to generate a build system linked with LLVM/Clang prebuilts, e.g.
+  * Create a build root and use CMake to generate a build system linked with LLVM/Clang prebuilts. Note that CMake settings that affect link-compatibility need to match exactly, so you may need to add `-DCMAKE_BUILD_TYPE=Release` or more (depending on how your LLVM/Clang libraries were built) to the command-line below:
```
vsapsai: Reminder to regenerate README.md.

kimgr: Reminder very welcome, I had forgotten.

CMakeLists.txt
kimgr: Oops! Will fix.

kimgr:
Thanks for reviewing!

> By the way, are Windows prebuilt binaries always include libraries?
> Because I think I tried it at some point and it contained only final
> executables, no libraries.

You mean LLVM's prebuilts from http://releases.llvm.org/download.html? You're right, they're executables only -- I started with that package for testing, but eventually had to rebuild it all from scratch.

kimgr:
There, this should be better. @nocnokneo, can you take a look?

PR #405, opened 2017-01-29
Remove test expectation for char_traits

kimgr:
Clang no longer desugars as aggressively. As a consequence, char_traits no
longer 'leaks' through the definition of std::string and is not noted in
the why-comment for `<string>`.


kimgr:
This should bring us back to a green build!

PR #408, opened 2017-02-11
Add optional terse flag to PrintableDecl

kimgr:
When working on the nested macros bug, I found it useful to be able log decls with less formatting.

I hacked up my own ugly post-formatting to get what I wanted, but then I realized that Clang has a printing policy specifically for this.

Hoping to use this in my upcoming nested-macros patch.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -402,10 +402,16 @@ string PrintableSourceRange(SourceRange range) {
   return PrintableLoc(range.getBegin()) + " - " + PrintableLoc(range.getEnd());
 }
 
-string PrintableDecl(const Decl* decl) {
-  std::string buffer;    // llvm wants regular string, not our versa-string
+string PrintableDecl(const Decl* decl, bool terse/*=true*/) {
+  // Use the terse flag to limit the level of output to one line.
+  clang::PrintingPolicy policy = decl->getASTContext().getPrintingPolicy();
+  policy.TerseOutput = terse;
+  policy.SuppressInitializers = terse;
+  policy.PolishForDeclaration = terse;
+
+  std::string buffer;
   raw_string_ostream ostream(buffer);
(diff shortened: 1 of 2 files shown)


vsapsai:
Looks good to me. Tried to switch to be terse by default and didn't notice any problems. Do you know any pitfalls of terse output? For example, if in some cases it is ambiguous.

kimgr:
I don't know the details of terse output, I just noticed that it turned decls into one-liners, which is better for logging.

But I like the idea of turning it on by default. I'm feeling brave today, let's try it!

kimgr:
PrintableDecl doesn't appear to be used very much, so I think it's safe to make it terse by default. Pushed an updated commit.

vsapsai:
Looks good to me.

No strong opinion, just for discussion. What about not changing `PolishForDeclaration` for terse output? [Documentation mentions](https://clang.llvm.org/doxygen/structclang_1_1PrintingPolicy.html#a22f302528b047a689fac6c4900d55d91) in this case it won't print attributes. As for me, attributes have high chance of being overlooked and shouldn't be too noisy. So it might be safer to keep them with terse logging.

kimgr:
I actually thought the opposite -- at least GCC attributes are pretty verbose, e.g. `__attribute__(no_caller_saved_registers)`, and I wanted to produce something short that is still easy to map back into source code. Including attributes shouldn't make it easier to distinguish between different decls, I think.

Note also that this does not affect the AST dump, `PrintableDecl` is only used in a few select scenarios for logging.

vsapsai:
I trust your judgement that it is better to omit attributes.

kimgr:
> Note also that this does not affect the AST dump, PrintableDecl is only used in a few select
> scenarios for logging.

I just noticed I was wrong about this -- somehow the `PrintableDecl` calls from AST traversal escaped my attention.

Let's see how this works out in practice, we can always flip the default back if we find we're missing information.

Issue #368, opened 2016-11-01
iwyu 0.7 - iwyu_include_picker.cc:1047: Assertion failed


WildCard-25:
output from 0.7 verbose 9
>[ Use macro   ] D:/Repositories/shareaza/shareaza/Buffer.cpp:986:9: TRUE (from C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/atlmfc/include/afx.h:273:9)
Marked full-info use of macro TRUE (from C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/atlmfc/include/afx.h) at D:/Repositories/shareaza/shareaza/Buffer.cpp:986:9
Marked full-info use of macro DEBUG_NEW (from C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/atlmfc/include/afx.h) at D:/Repositories/shareaza/shareaza/Buffer.cpp:33:13
Adding mapping from \<sdkddkver.h> to 
D:\dev\private\llvm39\llvm\tools\clang\tools\include-what-you-use\iwyu_include_picker.cc:1047: Assertion failed: IsQuotedInclude(map_to) && "All map values must be quoted includes"

output from 0.6 verbose 9
>[ Use macro   ] D:/Repositories/shareaza/shareaza/Buffer.cpp:986:9: TRUE (from C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/atlmfc/include/afx.h:273:9)
Marked full-info use of macro TRUE (from C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/atlmfc/include/afx.h) at D:/Repositories/shareaza/shareaza/Buffer.cpp:986:9
Marked full-info use of macro DEBUG_NEW (from C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/atlmfc/include/afx.h) at D:/Repositories/shareaza/shareaza/Buffer.cpp:33:13
Marked use of include-file  (from \<math.h>) at Invalid location
Marked use of include-file  (from \<float.h>) at Invalid location

Looking back through the output, the only real difference is a missing slash at the end of the path for Buffer.cpp, which probably isn't good if this is the actual full path used and not just a pretty printing issue.

0.7
[...]

PR #410, opened 2017-02-13
Exclude prefix headers from macro-defined-by-includer heuristics (issue #368).

vsapsai:
Also considered options of not populating `macro_users_` or
`direct_includes_as_fileentries_` for prefix headers but decided not to
do so. I think it is better to have correct information for prefix
headers but not to give recommendations based on that instead of having
incomplete information.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1889,6 +1889,13 @@ size_t PrintableDiffs(const string& filename,
 void IwyuFileInfo::HandlePreprocessingDone() {
   // Check macros defined by includer.  Requires file preprocessing to be
   // finished to know all direct includes and all macro usages.
+  //
+  // Exclude prefix headers from mapping heuristics.  Includes in prefix
+  // headers are kept regardless of their usage in includer.  And entire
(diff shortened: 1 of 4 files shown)


iwyu_output.cc
```diff
@@ -1889,6 +1889,13 @@ size_t PrintableDiffs(const string& filename,
 void IwyuFileInfo::HandlePreprocessingDone() {
   // Check macros defined by includer.  Requires file preprocessing to be
   // finished to know all direct includes and all macro usages.
+  //
+  // Exclude prefix headers from mapping heuristics.  Includes in prefix
+  // headers are kept regardless of their usage in includer.  And entire
```
kimgr: And the entire

kimgr:
Looks like our mailbot doesn't pick up the new GitHub reviews. See above.

iwyu_output.cc
vsapsai: Will fix.

vsapsai:
Merged 7cb24d257b4576a1bfc82b4dcdce81e90ea8218d with comment fix.

Issue #358, opened 2016-08-22
False positive when using htons and friends


svenpanne:
Consider the following short file `foo.c`:

``` C
#include <netinet/in.h>
int my_htons(int x) { return htons(x); }
```

(Yes, `int` is not totally correct here, but this is just for demonstration purposes.) On my Ubuntu 16.04 box `include-what-you-use` (`clang_3.9` branch) issues a false positive:

```
$ include-what-you-use -O -c foo.c

foo.c should add these lines:
#include <byteswap.h>    // for __x

foo.c should remove these lines:

The full include-list for foo.c:
#include <byteswap.h>    // for __x
#include <netinet/in.h>  // for htons

---
```

Note that the `-O` is essential here, otherwise the macro definitions will be different. I was able to boil this example down to 3 files (see [boiled_down.tar.gz](https://github.com/include-what-you-use/include-what-you-use/files/429795/boiled_down.tar.gz), too):
- `foo.c`:

``` C
#include "my_netinet_in.h"
int my_htons(int x) { return htons(x); }
```
- `my_netinet_in.h`:

``` C
#include "my_byteswap.h"
#define htons(x) bswap(x)
```
- `my_byteswap.h`:

``` C
#define bswap(x) ({ int __x = (x); bswap2(__x); })
#define bswap2(x) (x)
```

I've got no clue if this iwyu bug is related to the use of the non-standard statement expression in `bswap` and/or the macro structure itself.

kimgr:
I think it has to do with the macros. It sounds like you're building from source, could you try and apply this patch: https://github.com/include-what-you-use/include-what-you-use/pull/350

Thanks!

svenpanne:
Yes, I'm building from source, but the patch didn't help.

kimgr:
Sorry to hear. Thanks for the reduced example, gives me something to work from.

I won't have time to look at this for a while yet, but I'll try to pick it up soon.

kimgr:
It looks to be a problem with nested macros. If I remove ``bswap2``, everything works as expected.

It's because `__x`` is passed as a macro argument, that fact attributes the use to the passing location, which is in itself a macro. I'll see if I can figure out a way to short-circuit this.

kimgr:
After many moons, I've now circled back to this. And solved it, I think!

You can try this patch on top of master:
https://github.com/kimgr/include-what-you-use/commit/5606e046589cd4e976cff82f2edfa8f2a4258c7a.patch

[...]

PR #413, opened 2017-02-22
Handle locations of nested macros better

kimgr:
Issue #358 describes that htons from inet.h completely misses the mark
when finding the use location.

This patch should handle nested macros better. Test case from #358 added
to macro_location.cc

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1375,62 +1375,71 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   SourceLocation GetCanonicalUseLocation(SourceLocation use_loc,
                                          const NamedDecl* decl) {
     // If we're not in a macro, just echo the use location.
-    if (!IsInMacro(use_loc))
+    if (!use_loc.isMacroID())
       return use_loc;
 
-    // If the macro definition file contains a forward-declaration for decl,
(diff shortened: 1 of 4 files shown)


iwyu.cc
```diff
@@ -1375,62 +1375,71 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    clang::SourceManager* sm = GlobalSourceManager();
+
+    VERRS(5) << "Trying to determine use location for '"
+             << PrintableDecl(decl) << "'\n";
+
+    // Spin over the expansions from the inside out, looking for the first
+    // expansion location (i.e. not concatenated or part of a macro argument.)
+    // Inspired by clang's DiagnosticRenderer::emitMacroExpansions.
+    SourceLocation loc = use_loc;
+    while (loc.isMacroID() && sm->isMacroArgExpansion(loc) &&
```
kimgr: Looking at this now, this must be wrong -- should be `sm->isMacroArgExpansion(loc) || IsInScratchSpace(loc)`.

This passes all the tests, hopefully the fixed-up version will, too.

kimgr: Ugh. Tests fail with the or-expression. I'll retract the PR.

kimgr:
See review comments.

PR #415, opened 2017-03-02
Handle locations of nested macros better

kimgr:
Issue #358 describes that htons from inet.h completely misses the mark
when finding the use location.

This patch should handle nested macros better. Test case from #358 added
to macro_location.cc

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1375,62 +1375,61 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   SourceLocation GetCanonicalUseLocation(SourceLocation use_loc,
                                          const NamedDecl* decl) {
     // If we're not in a macro, just echo the use location.
-    if (!IsInMacro(use_loc))
+    if (!use_loc.isMacroID())
       return use_loc;
 
-    // If the macro definition file contains a forward-declaration for decl,
-    // we treat that as a use-attribution hint. See below for details.
-    const FileEntry* macro_def_file = GetFileEntry(GetSpellingLoc(use_loc));
+    VERRS(5) << "Trying to determine use location for '"
+             << PrintableDecl(decl) << "'\n";
+
+    clang::SourceManager* sm = GlobalSourceManager();
+    SourceLocation spelling_loc = sm->getSpellingLoc(use_loc);
+    SourceLocation expansion_loc = sm->getExpansionLoc(use_loc);
+
+    // If the file defining the macro contains a forward decl, keep it around
+    // and treat it as a hint that the expansion loc is responsible for the
+    // symbol.
+    const FileEntry* macro_def_file = GetLocFileEntry(spelling_loc);
+    VERRS(5) << "Macro is defined in file '" << GetFilePath(macro_def_file)
+             << "'. Looking for fwd-decl hint...\n";
+
     const NamedDecl* fwd_decl = nullptr;
     for (const NamedDecl* redecl : GetClassRedecls(decl)) {
-      if (GetFileEntry(redecl) == macro_def_file &&
-          IsForwardDecl(redecl)) {
+      if (GetFileEntry(redecl) == macro_def_file && IsForwardDecl(redecl)) {
         fwd_decl = redecl;
+
+        // Make sure we keep that forward-declaration, even if it's probably
+        // unused in this file.
+        IwyuFileInfo* file_info =
+            preprocessor_info().FileInfoFor(macro_def_file);
+        file_info->ReportForwardDeclareUse(
+            spelling_loc, fwd_decl,
+            IsNodeInsideCXXMethodBody(current_ast_node()), nullptr);
         break;
       }
     }
 
-    if (fwd_decl) {
-      // Make sure we keep that forward-declaration, even if it's probably
-      // unused in this file.
-      preprocessor_info().FileInfoFor(macro_def_file)->ReportForwardDeclareUse(
-          use_loc, decl, IsNodeInsideCXXMethodBody(current_ast_node()),
-          nullptr);
-    }
-
     // Resolve the best use location based on our current knowledge.
     //
-    // 1) If the macro definition file forward-declares the used decl, that's a
-    //    hint that it wants the expansion location to take responsibility.
-    // 2) If the symbol being used is passed as an argument to the macro, we
-    //    want the expansion location to take responsibility (that thing they
-    //    passed in is something they know about, and the macro doesn't.)
-    // 3) If the use_loc is in <scratch space>, we assume it's formed by macro
+    // 1) If the use_loc is in <scratch space>, we assume it's formed by macro
     //    argument concatenation, and attribute responsibility to the expansion
     //    location.
+    // 2) If the macro definition file forward-declares the used decl, that's a
+    //    hint that it wants the expansion location to take responsibility.
     //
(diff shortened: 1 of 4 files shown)


vsapsai:
Tested there are no regressions, need more time to understand the fix.

iwyu.cc
```diff
@@ -1375,62 +1375,70 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-      VERRS(5) << "Macro file has a forward-decl attribution hint.\n";
+      // If the macro file has a forward-decl, the expansion loc is responsible.
+      VERRS(5) << "Found a forward-decl in macro definition file.\n";
       use_loc = expansion_loc;
-    } else if (IsInScratchSpace(use_loc)) {
-      VERRS(5) << "Decl was used in <scratch space>, presumably as a result of "
-                  "macro arg concatenation.\n";
+    } else if (sm->isMacroArgExpansion(spelling_loc)) {
+      // For macro arguments, the expansion loc is responsible.
+      VERRS(5) << "Spelling location is in a macro-arg expansion.\n";
```
vsapsai: I got a little bit suspicious about checking `spelling_loc` for macro arg expansion and for scratch space. But I don't have concrete thoughts about that yet.

What is puzzling me is that this branch isn't invoked during `tests/cxx/macro_location.cc` test. Is it intended? Because I expected it to be used for arguments in nested macros.

kimgr: Thanks, good feedback!

It sounds suspicious that this branch isn't hit by the test suite; I would have expected the new nested htons test to trigger this.

As for spelling vs expansion, my theory was that macro expansions are 'incremental'. So the 'unwind' at the top of the method would find the right level of expansion to make responsible for the use by repeatedly following immediate macro expansions.

Once we find the right point in the 'macro backtrace', the spelling loc should be that point and the expansion loc should be one level up. I've used debug traces of the macro backtrace to come to this conclusion. So `spelling_loc` and `expansion_loc` are relative, not the spelling/expansion locs you'd get if you asked the original use location.

I'm fully prepared to learn that this is all a huge misunderstanding :-)

I'll try to find some time to design a test case that hits the `isMacroArgExpansion` branch, or come up with an explanation for why it's not currently covered.

kimgr:
Thanks! See response inline.

iwyu.cc
kimgr: Actually, the more precise test case to provoke this should be:

    // macro_location-d2.h
    #define CREATE_VAR(typ)   typ create_var

    // macro_location.h
    CREATE_VAR(HClass);

But I can see that the expansion is not working as I expected:

    tests/cxx/macro_location.h:26:1 <Spelling=./tests/cxx/macro_location-d2.h:37:31>: (1) [ VarDecl ] HClass create_var
    tests/cxx/macro_location.h:26:1 <Spelling=./tests/cxx/macro_location.h:26:12>: (2) [ RecordType ] class HClass     
    Trying to determine use location for 'class HClass {}'                                                             
    Original use loc: tests/cxx/macro_location.h:26:1 <Spelling=./tests/cxx/macro_location.h:26:12>                    
    Spelling loc: tests/cxx/macro_location.h:26:12                                                                     
    Expansion loc: tests/cxx/macro_location.h:26:12                                                                    
    Macro is defined in file 'tests/cxx/macro_location.h'. Looking for fwd-decl hint...                                
    Attributing use of 'class HClass {}' to location at tests/cxx/macro_location.h:26:12.                              

The original use location looks wrong here -- I would have expected the spelling loc to point into `macro_location-d2.h`. So maybe this is an issue with Clang's location handling, that implicitly locates macro arg expansions at some offset from the expansion loc?

I'll try and example with bad syntax inside a macro, and see if I can reproduce the error in Clang's error message.

vsapsai: After checking the code a little bit more, my suspicions about `spelling_loc` turned out to be incorrect. Looks like you are right and it is correct to use `spelling_loc` in those 2 checks. Still would be great to cover `isMacroArgExpansion` branch with a test.

kimgr:
I'm really glad you poked at this -- I have found absolutely no way to cover `isMacroArgExpansion`. That in turn made me question whether the 'macro unwind' was really doing something useful, and it appears not.

I was able to remove that and the `isMacroArgExpansion` branch without failing any tests.

I'll walk through a couple of old bug reports and compare notes, and see if I can break it, but otherwise I have a much smaller patch to put up for review soon.

kimgr:
PR updated with simplified implementation. Please take another look.

vsapsai:
Tested on macOS and Ubuntu - everything works fine. Didn't look into the code yet. Though it does look simpler than earlier.

iwyu.cc
```diff
@@ -1375,62 +1375,55 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      if (GetFileEntry(redecl) == macro_def_file && IsForwardDecl(redecl)) {
         fwd_decl = redecl;
+
+        // Make sure we keep that forward-declaration, even if it's probably
+        // unused in this file.
+        IwyuFileInfo* file_info =
+            preprocessor_info().FileInfoFor(macro_def_file);
+        file_info->ReportForwardDeclareUse(
+            spelling_loc, fwd_decl,
+            IsNodeInsideCXXMethodBody(current_ast_node()), nullptr);
```
vsapsai: I suspect this `ReportForwardDeclareUse` isn't covered by tests because I tried with `use_loc` and `decl` and nothing was broken. It's not a blocker, just mention it for general awareness.

iwyu.cc
```diff
@@ -1375,62 +1375,55 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-    // Resolve the best use location based on our current knowledge.
-    //
-    // 1) If the macro definition file forward-declares the used decl, that's a
-    //    hint that it wants the expansion location to take responsibility.
-    // 2) If the symbol being used is passed as an argument to the macro, we
-    //    want the expansion location to take responsibility (that thing they
-    //    passed in is something they know about, and the macro doesn't.)
-    // 3) If the use_loc is in <scratch space>, we assume it's formed by macro
-    //    argument concatenation, and attribute responsibility to the expansion
-    //    location.
```
vsapsai: I like these removed comments more than new comments. Old comments explain intention and reasoning behind our actions. New comments are more like implementation summary. That's why they have less value, as for me. What do you think?

vsapsai:
Reviewed the code, it looks fine. There is one inline question about comments.

iwyu.cc
kimgr: There is a quirk in the test cases for this:
https://github.com/include-what-you-use/include-what-you-use/blob/master/tests/cxx/macro_location-d2.h#L12-L16

I think here's what happens:

- We're inside a macro, and there's a forward-declaration of the real decl being used
- The use-location is therefore in the same file as the nearest decl-location (the fwd decl)

So the `use_loc` and `spelling_loc` resolve to the same file, and possibly `fwd_decl` and `decl` to the same decl.

I think this may be a special case, but I'm not 100% sure. The logic as-is covers the general case, where both location and decl can differ.

iwyu.cc
kimgr: Good point, I'll put them back!

kimgr:
Thanks, I'll clear up the last few things and post a new patch.

PR #416, opened 2017-03-03
Add pragma to set associated header

kimgr:
Sometimes IWYU's auto-detection of associated headers is insufficient.
Add an IWYU pragma to explicitly set associated header.

This is based on an original patch by Ivan Koster.

I added a test and some documentation.

This supersedes PR #302.

### DIFF
--- a/tests/cxx/pragma_associated.cc
+++ b/tests/cxx/pragma_associated.cc
@@ -1,0 +1,27 @@
+//===--- pragma_associated.cc - test input file for iwyu ------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/pragma_associated-d1.h"  // IWYU pragma: associated
+#include "tests/cxx/pragma_associated-d2.h"  // IWYU pragma: associated
+#include "tests/cxx/pragma_associated.h"  // This still counts as associated.
+#include "tests/cxx/direct.h"  // This is unused.
+
+/**** IWYU_SUMMARY
+
+tests/cxx/pragma_associated.cc should add these lines:
+
+tests/cxx/pragma_associated.cc should remove these lines:
+- #include "tests/cxx/direct.h"  // lines XX-XX
+
+The full include-list for tests/cxx/pragma_associated.cc:
+#include "tests/cxx/pragma_associated-d1.h"
+#include "tests/cxx/pragma_associated-d2.h"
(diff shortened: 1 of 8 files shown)


kimgr:
@ivankoster, please take a look to make sure I'm doing your work justice.

ivankoster:
@kimgr It looks great to me. You added the test I couldn't get to work and documentation. "Doing my work justice" is awefully humble of you!

tests/cxx/pragma_associated.cc
```diff
@@ -0,0 +1,25 @@
+//===--- pragma_associated.cc - test input file for iwyu ------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/pragma_associated-d1.h"  // IWYU pragma: associated
```
vsapsai: I'm not entirely sure if we need to have also `pragma_associated-d2.h` to test that pragma works for multiple files.

kimgr:
I based the decision on the assumption that a `.cc` file can already have multiple associated headers -- it looks like it considers all of `foo.h`, `foo.inl` and `foo-inl.h` associated headers of `foo.cc`. We might as well add to that set with `IWYU pragma: associated`. 

Moving forward, if necessary, we could consider a new `IWYU pragma: no_associated` pragma to un-mark an automatically detected header, but I think it we can hold that off until we know it's necessary.

tests/cxx/pragma_associated.cc
kimgr: I figured that would be covered by the auto-associated header, but I guess double manual association would be nice to test as well. Will add.

kimgr: Done.

kimgr:
Added another associated header to test. Please take another look.

tests/cxx/pragma_associated.cc
vsapsai: My motivation for testing multiple files is that we change `associated_pragma_location_` state (valid/invalid) in multiple places. And that is somewhat error-prone.

vsapsai:
I was thinking if we should have `no_associated` pragma and my desire is to cover most common use case. If people usually don't want auto-association when provide custom pragma - better not to do auto-association and make users mark all headers manually. If folks usually want to keep auto-associated headers, requiring changes for those headers creates unnecessary work and it is better to have `no_associated` pragma.

Checked the change, looks good to me.

kimgr:
Thanks!

I think I agree with your analysis of `no_aggregate`. Time will tell what's a good way forward.

PR #417, opened 2017-03-17
Added command line argument to iwyu_tool.py to allow for directories …

johnbaker32:
…to ignore during processing. 

Useful for cases where third party software is dynamically pulled into build directory and the iwyu tool output gets very cluttered analyzing large amounts of tps.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -152,9 +152,10 @@ def main(compilation_db_path, source_files, verbose, ignore_dirs, formatter, iwy
                       source)
 
     # Remove entries with directory belonging to any ignore_dirs
-    if ignore_dirs:
-        entries[:] = [entry for entry in entries if all(not entry['directory']
-                     .startswith(os.path.realpath(ignore)) for ignore in ignore_dirs)]
+    def is_ignored(dirpath):
+        return any(dirpath.startswith(os.path.realpath(i)) for i in ignore_dirs)
+
+    entries[:] = [e for e in entries if not is_ignored(e['directory'])]
 
     # Run analysis
     try:
@@ -208,7 +209,7 @@ def _bootstrap():
     parser.add_argument('source', nargs='*',
                         help='Zero or more source files to run IWYU on. '
                         'Defaults to all in compilation database.')
-    parser.add_argument('-i', '--ignore', nargs='+',
+    parser.add_argument('-i', '--ignore', nargs='+', default=[],
                         help='Directories in build path to ignore')
 
     def partition_args(argv):
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -151,11 +151,13 @@ def main(compilation_db_path, source_files, verbose, ignore_dirs, formatter, iwy
                 print('WARNING: \'%s\' not found in compilation database.' %
                       source)
 
-    # Remove entries with directory belonging to any ignore_dirs
-    def is_ignored(dirpath):
-        return any(dirpath.startswith(os.path.realpath(i)) for i in ignore_dirs)
 
-    entries[:] = [e for e in entries if not is_ignored(e['directory'])]
+    # Remove entries with directory belonging to any ignore_dirs
+    ignore_dirs = [os.path.realpath(i) for i in ignore_dirs]
+    def is_ignored(dirpath):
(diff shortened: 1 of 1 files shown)


iwyu_tool.py
```diff
@@ -151,6 +151,11 @@ def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
                 print('WARNING: \'%s\' not found in compilation database.' %
                       source)
 
+    # Remove entries with directory belonging to any ignore_dirs
+    if ignore_dirs:
+        entries[:] = [entry for entry in entries if all(not entry['directory']
```
kimgr: Ugh, nested comprehensions always makes my head spin. And the `all(not` feels like a double negation.

If you can default `ignore_dirs` to an empty list, you can introduce a nested helper function to make this more obvious:

    def is_ignored(dirpath):
         return any(dirpath.startswith(i) for i in ignore_dirs)

    entries[:] = [e for e in entries if not is_ignored(os.path.realpath(e['directory']))]

I think that would be easier to penetrate.

kimgr:
Thanks for your contribution! Posted some review feedback.

iwyu_tool.py
```diff
@@ -203,6 +208,8 @@ def custom_help():
     parser.add_argument('source', nargs='*',
                         help='Zero or more source files to run IWYU on. '
                         'Defaults to all in compilation database.')
+    parser.add_argument('-i', '--ignore', nargs='+',
```
johnbaker32: I think it just makes it require at least 1 argument following if `-i` or `--ignore` is specified.

iwyu_tool.py
johnbaker32: Works for me, i'll get that changed up.

iwyu_tool.py
```diff
@@ -151,6 +151,12 @@ def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
                 print('WARNING: \'%s\' not found in compilation database.' %
                       source)
 
+    # Remove entries with directory belonging to any ignore_dirs
+    def is_ignored(dirpath):
+        return any(dirpath.startswith(os.path.realpath(i)) for i in ignore_dirs)
```
kimgr: Let's also hoist the `realpath` call;

      ignore_dirs = [os.path.realpath(i) for i in ignore_dirs

      def is_ignored(dirpath):
          return any(dirpath.startswith(i) for i in ignore_dirs)

iwyu_tool.py
```diff
@@ -151,6 +151,12 @@ def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
...
+
+    entries[:] = [e for e in entries if not is_ignored(e['directory'])]
```
kimgr: I think you can just rebind `entries` to the comprehension, rather than overwriting the array:

    entries = [e for e...]

Not sure which is better/faster here, but this seems clearer.

iwyu_tool.py
```diff
@@ -203,6 +209,8 @@ def custom_help():
     parser.add_argument('source', nargs='*',
                         help='Zero or more source files to run IWYU on. '
                         'Defaults to all in compilation database.')
+    parser.add_argument('-i', '--ignore', nargs='+', default=[],
```
kimgr: Shouldn't  it be nargs='*'? I expected `+` to require at least one ignore-dir, which we don't want, right?

kimgr:
Thanks, I had a few more ideas when I saw this.

One more thought: which Python versions have you tested this with? We try to support Python 2.7 and onwards, and all Python 3 versions.

iwyu_tool.py
johnbaker32: `+` only requires an argument IF you specify `-i` or `--ignore`. It doesn't make you specify `-i`. While `*` wouldn't be functionally incorrect, it would allow for specifying the ignore flag without specifying any ignore directories, which doesn't make any sense.

johnbaker32:
I've been running 2.7.13, but I'll run on 3.x before submitting next changes

iwyu_tool.py
kimgr: Aha, got it, thanks!

kimgr:
@johnbaker32 Ping. Can you take another look?

kimgr:
This has gone stale, closing.

PR #418, opened 2017-03-19
Update README --no_comments flag

nilsdeppe:
The --no_comments flag to iwyu was --nocomments in the README, which is incorrect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ We also include, in this directory, a tool that automatically fixes up your sour
 If you don't like the way `fix_includes.py` munges your `#include` lines, you can control its behavior via flags. `fix_includes.py --help` will give a full list, but these are some common ones:
 
   * `-b`: Put blank lines between system and Google includes
-  * `--nocomments`: Don't add the 'why' comments next to includes
+  * `--no_comments`: Don't add the 'why' comments next to includes
 
 
 


kimgr:
Thanks for the patch!

It looks like `fix_includes.py` takes `--nocomments` and `include-what-you-use` takes `--no_comments`.

I think maybe the right fix is to change `include-what-you-use`'s switch to also be `--nocomments`. It's a breaking change, but it seems silly to have two different spellings.

nilsdeppe:
You're welcome!

I hadn't used the `fix_includes.py` so I didn't notices this. It seems like there's a greater inconsistency with the command line flags. In `include-what-you-use` most have underscores between words but then there's `--howtodebug`. I personally find the underscores make it easier to parse the flag and since most flags in `include-what-you-use` have underscores it might be better to go with that convention since it would break fewer things. Thoughts?

kimgr:
I don't have a strong preference on underscore vs not (my favorite is dashes, e.g. `--no-comments`), but the reason I mildly prefer `--nocomments` is because `fix_includes.py` had it first.

I added the `--no_comments` switch recently (about a year ago), and I missed that `fix_includes.py` already had a switch for this (we're not using `fix_includes.py` internally).

But since `include-what-you-use` already has a `--no_default_mappings` switch, maybe it would be nice to use the underscore spelling.

OK, so here's an idea:

- Add `--no_comments` to `fix_includes.py`
- Keep accepting `--nocomments`, but hide it from help text (see e.g. http://stackoverflow.com/a/11114654/96963)
- Update the docs as you've done to only mention `--no_comments`.

That way we get both consistent spelling and backward compatibility. Would you be willing to take a stab at it?

nilsdeppe:
If we're making this change and leaving the current for backwards compatibility, making it makes sense to move to dashes then and update the documentation for the dashes. That's my personal favorite too and would be in line with what most tools use. I'd be happy to take a stab at either change. Let me know if the dashes are okay or not.

kimgr:
I'm not sure, I'll defer to @vsapsai, my wise and seasoned co-maintainer. I'd be happy to revise all command-line switches, but I'm not sure there's enough bang for the buck, especially with the resulting backward-compat mess.

As a first step, I suggest we just go with underscore spelling, and then treat the dashes as a separate problem.

nilsdeppe:
Alright, that sounds like a good plan to me. Thanks!

vsapsai:
I would keep underscores and wouldn't switch to dashes as it is more work and the risk of breaking something is higher. As a user I don't remember what style of options different tools are using but I will notice if something breaks. So for users migration to dashes has no real value.

kimgr:
This does not appear to be going anywhere. I'll close this for now.

PR #427, opened 2017-04-30
Remove TODOs fixed by upstream Clang commit

kimgr:
This should fix our broken bot.

Clang r300938 caused these TODO:ed assertions to become unnecessary, which is nice. I haven't quite understood why. It looks like r300398 changes what is considered the definition in some circumstances, so it probably just shuffles responsibility around somehow.

It could be that this exposes another weakness in our `GetDefinitionAsWritten`, but it could also be something else entirely. I'd rather not spend more time on it.

### DIFF
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -1417,13 +1417,10 @@ int main() {
   local_i2_template_class.InlFileTemplateClassFn();
   // IWYU: I2_TemplateClass is...*badinc-i2.h
   local_i2_template_class.a();
-  // TODO(csilvers): these first three errors are wrong, due to a bug
-  // in IntendsToProvide.  The file defining this template method
-  // *should* be providing badinc-i2-inl.h, but isn't (iwyu will
-  // suggest we add it).  So we don't detect that that file is
-  // responsible for these symbols, and not us.
-  // IWYU: I2_TemplateClass::I2_TemplateClass<.*> is...*badinc-i2-inl.h
-  // IWYU: I2_TemplateClass::~I2_TemplateClass<.*> is...*badinc-i2-inl.h
+  // TODO(csilvers): this first error is wrong, due to a bug in
+  // IntendsToProvide.  The file defining this template method *should* be
+  // providing badinc-i2-inl.h, but isn't (iwyu will suggest we add it).
+  // So we don't detect that file as responsible for these symbols, and not us.
   // IWYU: I2_TemplateClass::InlFileTemplateClassFn is...*badinc-i2-inl.h
   // IWYU: I2_TemplateClass is...*badinc-i2.h
   local_i2_template_class.CcFileFn();


vsapsai:
How far did you get with your investigation? Because I figured out that those 3 uses are triggered by `template<typename FOO> int I2_TemplateClass<FOO>::CcFileFn()` implementation that uses all 3 of them. I find that quite bizarre and want to try to understand it. What about limiting my investigation for 5 days? If I don't find anything by end of Friday, May 5th, then I agree to fix the symptoms and postpone root causing till the better times.

kimgr:
Not very far, I'm afraid.

`I2_TemplateClass` has declarations spread out over three different files, and I guess when Clang changes what it considers the definition, our declaration location shuffles between these three, But this is all very wishy-washy :)

I'd be happy to wait until Friday, you're usually much better at grasping the big picture. Understanding the root cause is always better than not.

vsapsai:
TL;DR your change is OK to commit, I don't have another fix.

Before r300938 we were reporting use for `I2_TemplateClass` constructor and destructor [in `HandleFunctionCall`](https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L2479-L2490)

```c++
  bool HandleFunctionCall(FunctionDecl* callee, const Type* parent_type,
                          const clang::Expr* calling_expr) {
      return false;
    if (!callee || CanIgnoreCurrentASTNode() || CanIgnoreDecl(callee))
      return true;
    // We may have already been checked in a previous
    // VisitOverloadExpr() call.  Don't check again in that case.
    if (IsProcessedOverloadLoc(CurrentLoc()))
      return true;

    ReportDeclUse(CurrentLoc(), callee);
```

After r300938 we don't get to `ReportDeclUse` because `CanIgnoreDecl(callee)` returns true. And [`InstantiatedTemplateVisitor::CanIgnoreDecl`](https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L2882-L2889) tells us

```c++
  // We ignore function calls in nodes_to_ignore_, which were already
  // handled by the template-as-written, and function names that we
  // are not responsible for because the template code is (for
  // instance, we're not responsible for a vector's call to
  // allocator::allocator(), because <vector> provides it for us).
  bool CanIgnoreDecl(const Decl* decl) const override {
    return nodes_to_ignore_.Contains(decl);
  }
```

When you check how `nodes_to_ignore_` are populated in [`TraverseExpandedTemplateFunctionHelper`](https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L3184-L3203)

```c++
    // We want to ignore all nodes that are the same in this
    // instantiated function as they are in the uninstantiated version
    // of the function.  The latter will be reported when we traverse
    // the uninstantiated function, so we don't need to re-traverse
    // them here.
    AstFlattenerVisitor nodeset_getter(compiler());
    // This gets to the decl for the (uninstantiated) template-as-written:
    const FunctionDecl* decl_as_written
        = fn_decl->getTemplateInstantiationPattern();
    if (!decl_as_written) {
      if (fn_decl->isImplicit()) {     // TIP not set up for implicit methods
        // TODO(csilvers): handle implicit template methods
      } else {                         // not a templated decl
        decl_as_written = fn_decl;
      }
    }
    if (decl_as_written) {
      FunctionDecl* const daw = const_cast<FunctionDecl*>(decl_as_written);
      nodes_to_ignore_.AddAll(nodeset_getter.GetNodesBelow(daw));
    }
```

you see that we intentionally ignore function calls in instantiated template that are the same as in uninstantiated version. When in my repro case I replaced

```c++
I2_TemplateClass<int>* ptr = new I2_TemplateClass<int>(42, "hi");
```
with
```c++
I2_TemplateClass<FOO>* ptr = new I2_TemplateClass<FOO>(a_, "hi");
```

it started telling me again
```
tests/local/april_test_failure.cpp:20:3: warning: I2_TemplateClass::I2_TemplateClass<FOO> is defined in "tests/cxx/badinc-i2-inl.h", which isn't directly #included.
tests/local/april_test_failure.cpp:20:3: warning: I2_TemplateClass::InlFileTemplateClassFn is defined in "tests/cxx/badinc-i2-inl.h", which isn't directly #included.
tests/local/april_test_failure.cpp:20:3: warning: I2_TemplateClass::~I2_TemplateClass<FOO> is defined in "tests/cxx/badinc-i2-inl.h", which isn't directly #included.
```

Doing the same trick in `tests/cxx/badinc.cc` was unsuccessful and I didn't investigate it further.

Wrapping up the last loose end. `TraverseExpandedTemplateFunctionHelper` calls `decl_as_written = fn_decl->getTemplateInstantiationPattern();` and [r300938](http://llvm.org/viewvc/llvm-project?view=revision&revision=300938) changed `CXXRecordDecl::getTemplateInstantiationPattern`, that's how we work now with different template instantiation pattern.

kimgr:
Thanks! That helped me craft a better commit message. Merged.

Issue #420, opened 2017-04-04
iwyu asserts when on using declaration.


zjturner:
```
template<typename T>
class Bar {
public:
  void foo() { }
};

class Baz : public Bar<Baz> {
  typedef Bar<Baz> BaseT;
    
  using BaseT::foo;
  void foo() { }
};

int main(int argc, char **argv) {
  return 0;
}
```

When you run iwyu against this, it asserts with: `iwyu_output.cc:1951: Assertion failed: using_decl->shadow_size() > 0`

It seems to be due to the fact that you brought the base declaration in, but then overrode all of the base class declarations anyway.  If you declare a second overload of foo() in the base class but keep the derived class the same, the assertion doesn't happen.  But if you add the second overload to the list in the derived as well, the assertion happens again.  So it is only caused by the fact that every overload has been re-declared.

kimgr:
Thanks for the report!

I reduced this to:

    class Bar {
    public:
      void foo() {}
    };

    class Baz : public Bar {
      using Bar::foo;
      void foo() {}
    };

I think the relevant language rule is described here:
http://en.cppreference.com/w/cpp/language/using_declaration

> If the derived class already has a member with the same name,
> parameter list, and qualifications, the derived class member hides
> or overrides (doesn't conflict with) the member that is introduced
> from the base class.

This probably makes Clang ignore the using-decl in favor of the more derived declaration.

I'll think about whether it's relevant to keep the assertion in a modified form, or just cut it.

kimgr:
@cdglove, we discussed this in:
https://github.com/include-what-you-use/include-what-you-use/pull/221#discussion-diff-41559781

The phrasing in the code around using decls and shadow decls makes my head spin... Do you think we could omit the synthesized forward-declare use if there are no shadows?

Thanks!

kimgr:
Here's a PR that does this. I also sent out a question to the Clang list on using vs. using shadow, because I can't make sense of it.

kimgr:
Now fixed.

PR #428, opened 2017-04-30
Accept methods hiding using declarations

kimgr:
When a method in a derived class hides a name introduced by a
using-declaration, Clang discards the using shadow declarations for the
hidden name.

We used to have an assertion that every using-declaration must have a
shadow declaration. This patch removes it, to placate the above.

Fixes issue #420.


vsapsai:
The code looks correct to me. Still need to run tests.

vsapsai:
Tested on macOS and Ubuntu - everything is OK.

kimgr:
Thanks, merged!

Issue #425, opened 2017-04-06
IWYU incorrect forward declare


jcageman:
Repro:

DataElement.h

```
class CDataElement
{
};

```
main.cpp

```
#include <tchar.h>
#include "DataElement.h"


template < class T >
class CCreator
{
public:
	static void* Create()
	{
		return new T;
	}
};

typedef void* (*TConstructor)(void);

void RegisterType( TConstructor /*i_pfnConstructor*/ )
{
}

int _tmain(int argc, _TCHAR* argv[])
{
	RegisterType(CCreator<CDataElement>::Create);
	return 0;
}
```

IWYU suggestions:

```
D:/src/IWYUReproProject/IWYUReproProject/IWYUReproProject.cpp should add these lines:
class CDataElement;

D:/src/IWYUReproProject/IWYUReproProject/IWYUReproProject.cpp should remove these lines:
- #include "DataElement.h"  //lines 5-5

The full include-list for D:/src/IWYUReproProject/IWYUReproProject/IWYUReproProject.cpp:
#include <tchar.h>
class CDataElement;
---

```

applying the suggestions results in the following compile error in visual studio 2012
` 'CDataElement' : no appropriate default constructor available`

If you need more logging let me know.

If i change:
`RegisterType(CCreator<CDataElement>::Create);`
to
`RegisterType(&CCreator<CDataElement>::Create);`

 iwyu will make the correct decision by the way. E.g.  the & is crucial here. I have to admit with & would make more sense, but its legal C++ i think.

I tried debugging it myself, but i have to admit i do not understand the AST parser well enough to fix it.

jcageman:
An update on this issue:

I managed to actually understand a bit. Like i mentioned previously the difference between & and without is crucial here so i just made an ast-dump and diffed the dumps.

Without &
```
 | | | |-ExprWithCleanups <line:116:4, col:87> 'void'
  | | | | `-CXXMemberCallExpr <col:4, col:87> 'void'
  | | | |   |-MemberExpr <col:4, col:7> '<bound member function type>' .RegisterCompositeConstructor   | | | |   | `-DeclRefExpr <col:4> 'class CParameterHolder' lvalue Var 'PH' 'class CParameterHolder'
  | | | |   |-CXXConstructExpr <col:37, col:46> 'class CParameter::TToken':'class CStringToken' 'void (class CStringToken &&) __attribute__((thiscall)) noexcept' elidable
[...]

PR #429, opened 2017-05-01
Remove second-guessing for DeclRefExprs in CallExprs

kimgr:
In TraverseDeclRefExpr we would exit early if we were under a
CallExpr, with the motivation that CallExpr had already handled the
relevant cases.

However, CallExpr only looks specifically for reference arguments
passed to vararg functions, and ignores anything else.

Let TraverseDeclRefExpr work on all expressions, so that binding of
function pointers to templates works in all contexts, including
CallExprs.

Also took the time to extract function pointer binding from badinc.cc to a new test suite, and add full coverage for (free functions, static members, instance members) x templates.

Fixes issue #425.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -898,20 +898,12 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
         const_cast<CXXDestructorDecl*>(dtor), parent_type, expr);
   }
 
-  // This is to catch assigning template functions to function pointers.
+  // This is to catch function pointers to templates.
   // For instance, 'MyFunctionPtr p = &TplFn<MyClass*>;': we need to
   // expand TplFn to see if it needs full type info for MyClass.
   bool TraverseDeclRefExpr(clang::DeclRefExpr* expr) {
     if (!Base::TraverseDeclRefExpr(expr))  return false;
     if (CanIgnoreCurrentASTNode())  return true;
-
-    // If it's a normal function call, that was already handled by a
-    // CallExpr somewhere.  We want only assignments.
-    if (current_ast_node()->template ParentIsA<CallExpr>() ||
-        (current_ast_node()->template ParentIsA<ImplicitCastExpr>() &&
-         current_ast_node()->template AncestorIsA<CallExpr>(2))) {
-      return true;
-    }
 
     if (FunctionDecl* fn_decl = DynCastFrom(expr->getDecl())) {
       // If fn_decl has a class-name before it -- 'MyClass::method' --
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -1891,42 +1891,6 @@ int main() {
   local_d1_copy_class = D1CopyClassFn(I12);
   local_d1_copy_class.a();
 
-  // Test (templated) function pointers and method pointers.
-  // IWYU: I1_Class needs a declaration
-  // IWYU: I1_Enum is...*badinc-i1.h
-  // IWYU: I1_Function is...*badinc-i1.h
-  I1_Enum (*local_fn_ptr)(I1_Class*) = &I1_Function;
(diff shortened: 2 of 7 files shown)


tests/cxx/badinc.cc
```diff
@@ -1894,42 +1894,6 @@ int main() {
...
-  // IWYU: I1_Class needs a declaration
-  // IWYU: I1_Enum is...*badinc-i1.h
-  // IWYU: I1_Function is...*badinc-i1.h
-  I1_Enum (*local_fn_ptr)(I1_Class*) = &I1_Function;
-  // IWYU: I1_Class is...*badinc-i1.h
-  int (*static_method_ptr)() = &I1_Class::s;
-  // IWYU: I1_Struct is...*badinc-i1.h
-  // IWYU: I1_Struct needs a declaration
-  // IWYU: I1_TemplateMethodOnlyClass is...*badinc-i1.h
-  int (*tpl_fn_ptr)() = &I1_TemplateMethodOnlyClass<I1_Struct>::stat;
```
vsapsai: `I1_TemplateMethodOnlyClass::stat` seems to be not used anymore. Do you think it's worth removing?

iwyu.cc
```diff
@@ -898,21 +898,13 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
         const_cast<CXXDestructorDecl*>(dtor), parent_type, expr);
   }
 
-  // This is to catch assigning template functions to function pointers.
+  // This is to catch binding of function pointers to templates.
```
vsapsai: I find this sentence confusing. After some thought I understood what you mean but it took time. One of the biggest problems is how "bind" is used. It is used both "bind a variable to a value" and "bind data to parameter", i.e. direction of binding is ambiguous. And do we care if actual binding takes place? For example, in `&TplFn<MyClass*>;`? I am thinking about a comment like "This is to handle function pointers to template functions."

iwyu.cc
```diff
@@ -898,21 +898,13 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
...
   // For instance, 'MyFunctionPtr p = &TplFn<MyClass*>;': we need to
   // expand TplFn to see if it needs full type info for MyClass.
   bool TraverseDeclRefExpr(clang::DeclRefExpr* expr) {
     if (!Base::TraverseDeclRefExpr(expr))  return false;
     if (CanIgnoreCurrentASTNode())  return true;
 
-    // If it's a normal function call, that was already handled by a
-    // CallExpr somewhere.  We want only assignments.
```
vsapsai: In commit message you mention
> [...] CallExpr only looks specifically for reference arguments passed to vararg functions [...]

It is embarrassing but I cannot find `HandleFunctionCall` handling vararg functions. Can you please point me to the corresponding place?

tests/cxx/badinc.cc
kimgr: Yes, I can get rid of that too. I'm hoping to decimate badinc.* slowly.

iwyu.cc
kimgr: I was worried "binding" was too semantically overloaded, and I see what you're saying. I regret naming the whole test case "bind_funcptr" now :-)

I'll evalute your test case in context with the others and see if it makes sense there too.

iwyu.cc
kimgr: I took that from `VisitCallExpr` on line 2334. Should update the commit message to say *Visit*CallExpr.

tests/cxx/badinc.cc
kimgr: Done

iwyu.cc
kimgr: Done, toned down binding and added test cases for simple, "naked" function pointer expressions.

kimgr:
Address review comments. Please take another look.

PR #430, opened 2017-05-01
Let scrub-logs accept input from stdin

kimgr:
Use the fileinput module to take input from stdin or a named file.

This enables:

    $ include-what-you-use -Xiwyu -v7 file.cc 2>&1 | ./scrub-logs.py

which is very useful for comparing execution flows.


kimgr:
I'm guessing this is non-controversial. Merging.

Issue #412, opened 2017-02-22
CE after fix_includes.py


arseny30:
a.cpp:
```
template <class T>
struct Owner {
  using AliasT = T;
  ~Owner();
};

template <class T>
Owner<T>::~Owner() {
}

struct Query {
  class Callback {};
  Owner<Callback> callback_;
};

int main(void) {
  return 0;
}
```
Run with:
`iwyu --std=c++11 a.cpp `

Output:
```
The full include-list for a.cpp:
class Query::Callback;
---
```

This all leads to a CE after suggested fixes are applied.
Also this code looks like a minimal test for me. All further test simplifications I've tried are fixing the problem.

I used `include-what-you-use 0.7 (git:136736f) based on clang version 3.9.0 (tags/RELEASE_390/final)`

kimgr:
Thanks for the report!

Interesting case, since there are no `#include` directives anywhere :-)

I think the problem might be in `iwyu_output.cc`, the function called `DeclIsVisibleToUseInSameFile`. It looks like it doesn't take template uses into account.

kimgr:
Fixed in PR #433.

Issue #431, opened 2017-05-01
Typedefs of template arguments aren't required anywhere


kimgr:
While investigating #412, I realized that IWYU's current handling of typedefs of template arguments must be broken.

Consider this test case:

```c++
// class.h
class Class {};

// typedef_tpl.cc
#include "class.h"

template<class T>
class Container {
 public:
  // Should not be an iwyu violation for T inside the template...
  typedef T value_type;
};

void Instantiations() {
  // ... but when this typedef is used, surely the underlying type must be
  // required?
  Container<Class>::value_type vt;
}
```

Running IWYU on typedef_tpl.cc gives the following:
```
$ include-what-you-use typedef_tpl.cc

typedef_tpl.cc should add these lines:
class Class;

typedef_tpl.cc should remove these lines:
- #include "class.h"  // lines 1-1

The full include-list for typedef_tpl.cc:
class Class;
---
```

So it suggests that Class should be forward-declared instead of included. When I do that, on the other hand:
```
$ include-what-you-use typedef_tpl.cc
typedef_tpl.cc:14:32: error: variable has incomplete type 'Container<Class>::value_type' (aka 'Class')
  Container<Class>::value_type vt;
                               ^
typedef_tpl.cc:2:7: note: forward declaration of 'Class'
class Class;
      ^

(typedef_tpl.cc has correct #includes/fwd-decls)
```

IWYU produces broken code in this case. I haven't isolated where this happens yet, but I thought I'd write it down while I have it repro:ed.

jcageman:
Just as an addition to this, i noticed in our own source code that we had something similar, although a bit more difficult. Just to ensure this is also taken into account in a future fix.

```
template<T>
void Instantiations() {
  typedef SValueTraits<T>::TValueType vt;
}

valuetype.h

template<>
struct SValueTraits<CType>
{
  typedef CValueType TValueType;
}

```

As you can see, to know the outcome of this typedef we have a header which defines the SValueTraits<T>::TValueType specialization. Hence, when you use Instantations for a specific type T you have to include the header which defines the SValueTraits<T>::TValueType. Currently we add this include at the calling side and not in the the file that implements void Instantiations(). I currently manually fix these errors, since they don't occur a lot.

kimgr:
Thanks for the report! Does this happen only for nested typedefs, or is it a general problem with template specializations? If the latter, we may want to split it out into a new bug.

jcageman:
[...]

PR #433, opened 2017-05-02
Equate alias declarations and typedef in templates

kimgr:
Typedefs are exempted from IWYU checks in templates, so treat C++11
alias declarations the same.

(side note: this behavior is broken, but at least now it's consistent. See bug #431.)

Fixes issue #412.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -205,6 +205,7 @@ using clang::TranslationUnitDecl;
 using clang::Type;
 using clang::TypeLoc;
 using clang::TypedefDecl;
+using clang::TypedefNameDecl;
 using clang::TypedefType;
 using clang::UnaryExprOrTypeTraitExpr;
 using clang::UsingDecl;
@@ -3042,7 +3043,7 @@ class InstantiatedTemplateVisitor
     // you do 'Foo<MyClass>::value_type m;'?
     for (const ASTNode* ast_node = current_ast_node();
          ast_node != caller_ast_node_; ast_node = ast_node->parent()) {
-      if (ast_node->IsA<TypedefDecl>())
+      if (ast_node->IsA<TypedefNameDecl>())
         return Base::VisitSubstTemplateTypeParmType(type);
     }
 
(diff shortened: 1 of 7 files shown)


vsapsai:
Haven't tested extensively and observing some strange behaviour for non-alias with old code and new test case. Need more time, it is possible I've done some mistake during testing.

Have you considered using `ast_node->IsA<clang::TypedefNameDecl>()`? It *seems* like the right direction, though I'm not entirely sure, want your opinion.

kimgr:
Ah, I hadn't seen that they share a common base decl type. Yes, that should be generally better. I can try it and report back at some point.

I'm not surprised the "old" code is causing surprises, I found a number of inconsistencies when I logged issue #431.

kimgr:
Done. Please take another look.

vsapsai:
Looks good to me. Strange behavior turned out not to be strange. I expected alias to trigger full `Class` usage only in `Container<Class>::alias_type at;` But full usage was triggered just by `Container<Class>` in `Container<Class>::value_type` which is correct (though not so easy to reason about).

kimgr:
That sounds wrong, though. The rule in IWYU seems to be that template arguments should only require forward-decls. See e.g. https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L2565

So I would expect `Container<Class>::value_type` to break down into:

- full use of `Container`
- fwd-decl use of `<Class>`
- full use of `value_type` for the typedef *and*
- full use of `Class` for the typedef type

I think IWYU is broken as per #431, the full use of `value_type` doesn't seem to be recorded anywhere.

Also, I think IWYU's analysis is too simplistic regarding templates and arguments. Given e.g.

```c++
template<class T>
struct Foo {
  T full;
};

template<class T>
struct Bar {
  const T* fwd;
};
```

I have a sense that instantiating `Foo<Class>` would need the size of `Class`, because it's full-used inside the template. But `Bar<Class>` would be fine with a forward-decl.

I haven't had a check to investigate this deeper, but I think that fits well into #431 as well.

As far as I can see this change does not push things either way.

PR #435, opened 2017-05-21
Add option to remove arguments from compile_commands

pkryger:
There are use cases when project's compile_commands.json has some
extra arguments that should be removed before passing to clang.
This can happen when project is compiled with gcc and it  flags that
are not supported to LLVM/Clang, i.e., `-fno-default-inline`. This
patch allows to achieve that behavior.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -95,7 +95,7 @@ def get_output(cwd, command):
     return process.communicate()[0].decode("utf-8").splitlines()
 
 
-def run_iwyu(cwd, compile_command, iwyu_args, verbose, formatter):
+def run_iwyu(cwd, compile_command, iwyu_args, remove_args, verbose, formatter):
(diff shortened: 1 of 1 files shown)


vsapsai:
Is it possible to edit compile_commands.json directly? Currently I'm reluctant to add this feature as having arguments to remove other arguments seems unnecessarily complicated. Another reason to solve your problem at compile_commands.json level is that not only include-what-you-use will benefit from it. All other tools using compilation DB (e.g. clang-tidy) will be able to work on your codebase.

kimgr:
While I can see how this would be convenient, I agree with @vsapsai that it would be nicer not to mess with this in `iwyu_tool.py`. There's an interesting project called `jq` that you could use to rewrite the compilation database before using: https://stedolan.github.io/jq/.

Vote to close.

kimgr:
No activity here for a while, and we'd prefer not to make iwyu_tool.py smarter. Closing.

PR #436, opened 2017-05-22
adding x86intrin.h to builtin mapping rules

pseyfert:
I'm trying to run iwyu against https://gitlab.cern.ch/lhcb/LHCb and suspect importing x86intrin.h rather than e.g. emmintrin.h makes sense.


vsapsai:
How should it work when code using intrinsics is targeting non-x86 architecture?

Is this mapping consistent with recommendations in header files? For example, lwpintrin.h mentions `#error "Never use <lwpintrin.h> directly; include <x86intrin.h> instead."` and clflushoptintrin.h mentions `#error "Never use <clflushoptintrin.h> directly; include <immintrin.h> instead."`

Do you think it is better to have this mapping built-in or specifying mapping in .imp file will work too? Currently I don't know which approach is better. Just it is OK to have .imp file for a single architecture but built-in rules for x86 shouldn't break other architectures.

pseyfert:
For non-x86 architectures … good point, no idea. Do they use the same private headers?

I thought it is consistent. My clflushoptintrin.h says `# error "Never use <clflushoptintrin.h> directly; include <x86intrin.h> instead."` (means you spotted that the mapping is not portable.)

I had no strong preference for .imp or built-in either. I went for the built-in since x86intrin.h comes with libgcc and i saw libc "stuff" is already built-in. Given the mismatch above, I suspect an .imp file is the more flexible solution?

vsapsai:
Sorry for the delay, I need more time to investigate the difference in intrinsics headers between Clang and GCC and to check which headers are architecture-specific and which are not.

vsapsai:
I've checked Clang intrinsics that Apple bundles with Xcode ([.dot file](https://gist.github.com/vsapsai/3dcb89ff3775aea84a34705cb6b949a8)). Everything maps to `immintrin.h` or `x86intrin.h`. There are a few exceptions like `f16cintrin.h` and `prfchwintrin.h` but I think it's OK to prefer `immintrin.h` (Intel intrinsics) over `emmintrin.h` (only SSE2 intrinsics) and `x86intrin.h` over `mm3dnow.h`.

Now I am thinking how to make mapping not to recommend to replace `immintrin.h` with `x86intrin.h`. Both headers are valid and [recommended to be used](https://stackoverflow.com/questions/11228855/header-files-for-x86-simd-intrinsics). If you are already using `immintrin.h`, replacing it with `x86intrin.h` can be considered as unnecessary churn.

vsapsai:
Added to the aforementioned gist mappings for GCC 5.4.0. There are a few differences but looks like these differences usually have "x86intrin.h" as a common denominator. So I think in these cases it's OK to map to "x86intrin.h" and it will also have the side effect of having code more portable.

vsapsai:
To clarify, do you have problems with files that have
```
#error "Never use {header} directly; include {header} instead."
```
or with files that don't have such directive? In your change you have both of them but currently I'm thinking they should be handled in different ways.

kimgr:
There was a patch in circulation years ago (by yours truly) that provided auto-mappings for `#error` hints like this.

I still can't help but think this is a pretty nice heuristic, but it's risky enough that we may want to hide it behind a flag (actually, maybe we should put *all* heuristics behind flags, something analogous to `-W` like `-H`, e.g. `-Hheadername` vs `-Hnoheadername`, `-Hforward-typedef` vs `-Hnoforward-typedef`, etc.)

Prior discussion in #109, and patch here:
https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-109/comment-3/109r2.patch

vsapsai:
Well, after my testing I came to the conclusion that headers with `#error` are handled correctly and we want mostly to allow both `<immintrin.h>` and `<x86intrin.h>` depending on which one is already present in the code. And there are also a few exceptional headers that we might want to handle the same way both for GCC and Clang. But I'm not sure if my testing results are the same as @pseyfert's so I want to confirm what is exactly the problem.

kimgr:
Ah, sorry, I misunderstood.

pseyfert:
Sorry for the delay. My original motivation was to get rid of the `#error` files in the suggestions, but then concluded that `x86intrin.h` was a reasonable main include for working with intrinsics. Your suggestions sound good to me.

kimgr:
@pseyfert What's the status of this PR? I can't make out a clear way forward from the discussion.

pseyfert:
Rethinking, I think we want a mapping file (should make it easier for a user to find where the suggestions come from, and modify them. Modifyability seems desirable as we see the suggestions can change from compiler to compiler).
I now base the mappings on the `#error` messages among the intrin.h files from the gcc and clang versions I have here. When the two suggest different inclusions or say `<x86intrin.h> or <mm3dnow.h>`, both are in the mapping file.

My understanding of mapping files is (tested on a minimal example, please confirm) that two lines
```
   { include: ["<adxintrin.h>", private, "<x86intrin.h>", public ] },
   { include: ["<adxintrin.h>", private, "<immintrin.h>", public ] },
```
will keep either `<x86intrin.h>` or `<immintrin.h>` untouched if present, but will suggest the former `<x86intrin.h>` if none of them is present. For double public headers, I preferred the x86 version over imm.

kimgr:
I think that's almost right. I'm not 100% clear on the logic here, but there appears to be an element of frequency-of-use as well;

```
  // Step (3): Now we have a set-cover problem: we need to end up with
  // a set of headers, called cover, so that for every i:
  //    intersection(cover, public_headers[i]) != empty_set
  // We do this greedily: we find the header that's listed the most
  // often.  Among those, we prefer the one that's listed first in
  // public_headers[i] the most often (each list is in approximate
  // best-fit order).  Among those, we choose arbitrarily.  We repeat
  // until we cover all sets.
```

from https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_output.cc#L800.

So if both headers are included, it appears to prefer the first one in the mapping unless the second is included more often, but I'm honestly not sure.

kimgr:
Is there anything to do here?

kimgr:
I'll close this as there was no input for a long while. Feel free to reopen if there's something we should do here.

PR #442, opened 2017-06-09
Allow IWYU pragma: keep on forward declarations as an escape hatch fo…

KristofferHenriksson:
Allow IWYU pragma: keep on forward declarations as an escape hatch for cases where IWYU incorrectly concludes that a forward declaration is not necessary and recommends its removal. This allows for fast temporary fixes while a proper fix is being worked on.

Include x64 output paths in test util to allow it to locate Windows x64 builds of include-what-you-use.exe automatically.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -105,6 +105,7 @@
 #include "iwyu_ast_util.h"
 #include "iwyu_cache.h"
 #include "iwyu_globals.h"
+#include "iwyu_lexer_utils.h"
 #include "iwyu_location_util.h"
 #include "iwyu_output.h"
 #include "iwyu_path_util.h"
@@ -3660,6 +3661,12 @@ class IwyuAstConsumer
             if (decl == first_decl)
               definitely_keep_fwd_decl = true;
           }
+        }
+      } else {
+        SourceLocation decl_end_location = decl->getSourceRange().getEnd();
+        if (LineHasText(decl_end_location, "// IWYU pragma: keep") ||
+            LineHasText(decl_end_location, "/* IWYU pragma: keep")) {
+          definitely_keep_fwd_decl = true;
         }
       }
 
--- a/iwyu_lexer_utils.cc
+++ b/iwyu_lexer_utils.cc
@@ -8,6 +8,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "iwyu_lexer_utils.h"
+#include "iwyu_globals.h"
 
 #include <cstring>
 #include <string>
@@ -28,10 +29,17 @@ using clang::SourceLocation;
 using clang::SourceManager;
 using clang::SourceRange;
 using clang::Token;
+using llvm::StringRef;
 using std::string;
 using std::vector;
 
 namespace include_what_you_use {
+
+bool LineHasText(SourceLocation source_location, StringRef text) {
+  const StringRef data =
+      GetSourceTextUntilEndOfLine(source_location, DefaultDataGetter());
+  return data.find(text) != StringRef::npos;
+}
 
 // SourceManagerCharacterDataGetter method implementations.
 SourceManagerCharacterDataGetter::SourceManagerCharacterDataGetter(
@@ -48,14 +56,13 @@ const char* SourceManagerCharacterDataGetter::GetCharacterData(
   return data;
 }
 
-string GetSourceTextUntilEndOfLine(
-    SourceLocation start_loc,
-    const CharacterDataGetterInterface& data_getter) {
+StringRef GetSourceTextUntilEndOfLine(
+    SourceLocation start_loc, const CharacterDataGetterInterface& data_getter) {
   const char* data = data_getter.GetCharacterData(start_loc);
   const char* line_end = strchr(data, '\n');
   if (!line_end)
     return data;
-  return string(data, line_end - data);
+  return StringRef(data, line_end - data);
 }
 
 SourceLocation GetLocationAfter(
--- a/iwyu_lexer_utils.h
+++ b/iwyu_lexer_utils.h
@@ -25,6 +25,11 @@ namespace include_what_you_use {
 using std::string;
 using std::vector;
 
+// For a particular source line that source_location points to,
+// returns true if the given text occurs on the line.
+// (Case sensitive.)
+bool LineHasText(clang::SourceLocation source_location, llvm::StringRef text);
+
 // Interface to get character data from a SourceLocation. This allows
 // tests to avoid constructing a SourceManager yet still allow iwyu to
 // get the character data from SourceLocations.
@@ -46,7 +51,7 @@ class SourceManagerCharacterDataGetter : public CharacterDataGetterInterface {
 };
 
 // Returns the source-code line from the current location until \n.
-string GetSourceTextUntilEndOfLine(
+llvm::StringRef GetSourceTextUntilEndOfLine(
     clang::SourceLocation start_loc,
     const CharacterDataGetterInterface& data_getter);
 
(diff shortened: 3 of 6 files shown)


kimgr:
Thanks for the patch, this looks useful!

I'm not so happy about the sudden Doxygen syntax. I'm guessing it's the easiest way to get at the comments using Clang's API, but it's an odd bird compared to our other pragmas.

Our other symbol-related pragmas spell out symbol names in the pragma text, so they're easier. Not sure what infrastructure would be necessary to handle comment-lexing on forward-declarations.

KristofferHenriksson:
I updated it to parse all comments, not just doxygen-style comments. It's off by default, presumably for performance reasons.

kimgr:
Cool, thanks!

Sorry to be picky, but can we expect a trailing comment instead, like how it's done for the `#include` directives? The forms I'd expect are:

    class ForwardDecl; // IWYU pragma: keep
    class ForwardDecl; /* IWYU pragma: keep */

Also, you mention that enabling comment parsing might affect performance negatively -- do you have any numbers? IWYU is already dog-slow,, and it would be a shame if we made it noticeably slower for a fringe feature like this...

I wonder if you could fall back on raw string processing like we do for the other `pragma: keep` handling... You should be able to just `GetSourceTextUntilEndOfLine` from the decl's location. It's way uglier, but if it's measurable faster, I think it's definitely worth it.

As for the test; the `comment_pragmas.cc` test is already huge, but I think this belongs there, rather than in its private test file (every new test file requires a new IWYU run, which increases test time.)

KristofferHenriksson:
Pragmas are now required to be trailing and re-uses the same code for finding pragmas as the #include parsing.
Moved tests to comment_pragmas.cc

kimgr:
Nice, thanks! Much neater diff. Please find nits inline.

kimgr:
Hm, forgot to publish the review comments. Should be out there now.

iwyu_preprocessor.cc
```diff
@@ -93,14 +93,11 @@ static string GetIncludeNameAsWritten(SourceLocation include_loc) {
...
-// returns true if the given text occurs on the line.
-// (Case sensitive.)
-static bool IncludeLineHasText(SourceLocation include_loc,
-                               const string& text) {
-  const string data = GetSourceTextUntilEndOfLine(include_loc,
-                                                  DefaultDataGetter());
-  return data.find(text) != string::npos;
+bool LineHasText(SourceLocation source_location, StringRef text)
+{
+    const StringRef data = GetSourceTextUntilEndOfLine(source_location,
```
KristofferHenriksson: clang-format makes hundreds of changes when run on these files, the vast bulk of which are not related to my changes. I'm using clang-format from my local build branched off of the Clang 4.00 tagged release with -style=file which should pick up the .clang-format in the include-what-you-use directory. I even tried moving it to a new directory so there was no parent .clang-format that could be picked up, only the one in include-what-you-use.
Am I using it wrong or has clang-format not been run on these for a long time?

KristofferHenriksson:
To make sure this doesn't get hidden: I ran into problems running clang-format with it making tons of changes unrelated to mine. Either I'm running it wrong or clang-format hasn't been run on these files recently. If I'm not doing it wrong it would be helpful to have a checkin that just runs clang-format on these so future users can just run clang-format and not have to worry about the formatting.

kimgr:
Re clang-format: I don't know exactly how you run it, but I usually just run it on code I've changed (in a region in emacs). Some people swear by clang-format-diff, which operates on a diff. But maybe a global clang-format would be useful.

iwyu_lexer_utils.h
```diff
@@ -25,6 +25,12 @@ namespace include_what_you_use {
 using std::string;
 using std::vector;
 
+
+// For a particular source line that source_location points to,
+// returns true if the given text occurs on the line.
+// (Case sensitive.)
+bool LineHasText(clang::SourceLocation source_location, clang::StringRef text);
```
kimgr: llvm::StringRef here too

iwyu_lexer_utils.cc
```diff
@@ -8,6 +8,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "iwyu_lexer_utils.h"
+#include "iwyu_globals.h"
```
kimgr: Is this necessary?

iwyu.cc
```diff
@@ -3661,6 +3662,13 @@ class IwyuAstConsumer
               definitely_keep_fwd_decl = true;
           }
         }
+      } else {
+        SourceLocation decl_end_location = decl->getSourceRange().getEnd();
+        if (LineHasText(decl_end_location, "// IWYU pragma: keep") ||
+            LineHasText(decl_end_location, "/* IWYU pragma: keep"))
+        {
```
kimgr: Brace at end of line

iwyu_lexer_utils.cc
```diff
@@ -28,11 +29,19 @@ using clang::SourceLocation;
...
 using clang::SourceRange;
 using clang::Token;
+using llvm::StringRef;
 using std::string;
 using std::vector;
 
 namespace include_what_you_use {
 
+bool LineHasText(SourceLocation source_location, StringRef text)
+{
```
kimgr: Brace at end of line

kimgr:
Some more comments, but looks good for the most part

iwyu_lexer_utils.cc
KristofferHenriksson: It's needed for DefaultDataGetter()

iwyu.cc
KristofferHenriksson: Fixed brace and indent here.

iwyu_lexer_utils.h
KristofferHenriksson: D'oh. My search for clang::StringRef missed this because the generated Visual Studio projects don't include the header files, only cc files.

iwyu_lexer_utils.cc
```diff
@@ -48,14 +56,13 @@ const char* SourceManagerCharacterDataGetter::GetCharacterData(
   return data;
 }
 
-string GetSourceTextUntilEndOfLine(
-    SourceLocation start_loc,
-    const CharacterDataGetterInterface& data_getter) {
+StringRef GetSourceTextUntilEndOfLine(
+    SourceLocation start_loc, const CharacterDataGetterInterface& data_getter) {
```
KristofferHenriksson: clang-format-diff put these two parameters on the same line.

iwyu_lexer_utils.cc
kimgr: Ah, I see.

iwyu_lexer_utils.cc
kimgr: That's fine by me, we might as well follow clang-format's lead when we're touching functions.

Personally, I think the stacked parameter list is nicer, but I bow to the tooling.

kimgr:
Thanks! Could you rebase and squash into a single commit?

    $ git rebase -i master
    # squash all commits
    # collapse into a nice commit message

Cheers.

vsapsai:
Looks good to me and no regressions on macOS and Ubuntu. Initially I had concerns about Doxygen syntax as it creates inconsistent user interface. But now you are consistent with other IWYU pragmas so everything is good with UI.

KristofferHenriksson:
Manually squashed and pushed. Does github not allow you to choose to squash changes when merging? That option should really be under the control of the maintainers.

vsapsai:
Merged the changes. Thanks, Kristoffer, for your contribution.

PR #444, opened 2017-06-12
run multithread iwyu_tool.py

pseyfert:
wrap the call to run_iwyu (and include-what-you-use) into python's
multiprocessing.Pool to run multiple compilation concurrently

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -26,7 +26,7 @@ import json
 import argparse
 import subprocess
 import re
-from multiprocessing import Pool
+import multiprocessing
 
 
 def iwyu_formatter(output):
@@ -155,17 +155,15 @@ def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
     # Run analysis
     try:
 
-        pool = Pool()
-        # results doesn't contain actual results, it serves exception handeling
-        # https://stackoverflow.com/a/28660669
+        pool = multiprocessing.Pool()
+        # No actual results in `results`, it's only used for exception handling.
+        # Details here: https://stackoverflow.com/a/28660669.
         results = []
         for entry in entries:
             cwd, compile_command = entry['directory'], entry['command']
             results.append(pool.apply_async(run_iwyu,
                                             (cwd, compile_command, iwyu_args, verbose),
-                                            callback = formatter
-                                           )
-                          )
+                                            callback=formatter))
         pool.close()
         pool.join()
         for r in results:
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -154,12 +154,22 @@ def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
 
     # Run analysis
     try:
+
         pool = Pool()
+        # results doesn't contain actual results, it serves exception handeling
+        # https://stackoverflow.com/a/28660669
+        results = []
         for entry in entries:
             cwd, compile_command = entry['directory'], entry['command']
-            pool.apply_async(run_iwyu, (cwd, compile_command, iwyu_args, verbose), callback = formatter)
+            results.append(pool.apply_async(run_iwyu,
+                                            (cwd, compile_command, iwyu_args, verbose),
+                                            callback = formatter
+                                           )
+                          )
         pool.close()
         pool.join()
+        for r in results:
+            r.get()
     except OSError as why:
         print('ERROR: Failed to launch include-what-you-use: %s' % why)
         return 1
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -115,7 +115,7 @@ def run_iwyu(cwd, compile_command, iwyu_args, verbose):
     return get_output(cwd, command)
 
 
-def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
+def main(compilation_db_path, source_files, verbose, formatter, jobs, iwyu_args):
     """ Entry point. """
     # Canonicalize compilation database path
     if os.path.isdir(compilation_db_path):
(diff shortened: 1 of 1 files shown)


kimgr:
Nice idea!

I'll have to test run this on Windows -- multiprocessing doesn't always work well there.

Do you need to synchronize the terminal somehow to avoid having output from multiple runs interleaved? Or does that just magically work?

pseyfert:
i was also wondering about output interleaving, but my understanding is that that's exactly what the `callback` argument is good for and why i moved the formatter there. i saw it on SO in examples how to let the subthreads append to a list. I didn't see issues when testing (though haven't yet run on a large stack and checked only 4 concurrent threads).

vsapsai:
`run_iwyu` prints some output in verbose mode so that might be interleaved.

What happens when one of the workers throws an exception? Checked quickly [`multiprocessing` docs](https://docs.python.org/2/library/multiprocessing.html) but haven't found the answer.

pseyfert:
I think output shouldn't be an issue because stdout and stderr get collected in the return

```
                               stdout=subprocess.PIPE,
                               stderr=subprocess.STDOUT)
    return process.communicate()[0].decode("utf-8").splitlines()
```

For exceptions, trying around I think the thread just stops and produces no output and the Pool looks like all is fine. This must be fixed. On a side note, things like `include_what_you_use file or directory not found` (in case `$PATH` is not set up correctly) would not through an exception but just end up as error message in the output.

vsapsai:
OK, so the output shouldn't be interleaved. And thanks for handling exceptions.

Let me play a devil's advocate more.
* Why wouldn't `xargs -P` work? Or is it undesirable for portability reasons?
* How would you run other Clang tools like clang-tidy in multiple processes?

@pseyfert, I want to apologize for my negativistic tone, let it not discourage you. Leveraging multiple CPU cores is a great suggestion and I appreciate your contribution. I am trying to find if we can come up with a better implementation.

iwyu_tool.py
```diff
@@ -26,6 +26,7 @@
 import argparse
 import subprocess
 import re
+from multiprocessing import Pool
```
kimgr: I prefer importing modules and then referring to `module.Type` -- can you change this to `import multiprocessing`?

iwyu_tool.py
```diff
@@ -153,9 +154,22 @@ def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
 
     # Run analysis
     try:
+
+        pool = Pool()
```
kimgr: `pool = multiprocessing.Pool()`

iwyu_tool.py
```diff
@@ -153,9 +154,22 @@ def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
...
+        # results doesn't contain actual results, it serves exception handeling
+        # https://stackoverflow.com/a/28660669
+        results = []
         for entry in entries:
             cwd, compile_command = entry['directory'], entry['command']
-            run_iwyu(cwd, compile_command, iwyu_args, verbose, formatter)
+            results.append(pool.apply_async(run_iwyu,
+                                            (cwd, compile_command, iwyu_args, verbose),
+                                            callback = formatter
+                                           )
```
kimgr: Don't put closing parentheses on their own line

iwyu_tool.py
```diff
@@ -153,9 +154,22 @@ def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
...
+        pool = Pool()
+        # results doesn't contain actual results, it serves exception handeling
+        # https://stackoverflow.com/a/28660669
+        results = []
         for entry in entries:
             cwd, compile_command = entry['directory'], entry['command']
-            run_iwyu(cwd, compile_command, iwyu_args, verbose, formatter)
+            results.append(pool.apply_async(run_iwyu,
+                                            (cwd, compile_command, iwyu_args, verbose),
+                                            callback = formatter
```
kimgr: No spaces around keyword-argument `=`

iwyu_tool.py
```diff
@@ -153,9 +154,22 @@ def main(compilation_db_path, source_files, verbose, formatter, iwyu_args):
 
     # Run analysis
     try:
+
+        pool = Pool()
+        # results doesn't contain actual results, it serves exception handeling
```
kimgr: This comment could be clearer. Perhaps:

    # No actual results in `results`, it's only used for exception handling.
    # Details here: https://stackoverflow.com/a/28660669.

kimgr:
It looks like it's working well on both Windows and FreeBSD, I see as many Python and `include-what-you-use` processes as I have CPUs in both cases.

I wonder if parallellization should be behind a switch? Something like `-j` in Make, so you can choose the level of parallellism? We could always use multiprocessing but set pool size to 1 by default.

iwyu_tool.py
pseyfert: fine with me

iwyu_tool.py
pseyfert: fine with me

iwyu_tool.py
pseyfert: okay

iwyu_tool.py
pseyfert: okay

iwyu_tool.py
pseyfert: okay

pseyfert:
Quickly added a draft for `-j`. the default right now is the multiprocessing default (one subprocess per virtual core). Could as well make 1 the default.

kimgr:
Looks good to me, style-wise! I'd prefer for -j to default to 1, to mirror what Make does.

Can you also rebase on top of the latest master? I couldn't fast-forward when I last pulled your changes.

pseyfert:
Okay, just did. Probably I'm biased towards the ninja default and not the Make default. Nothing we can make consistent here.

kimgr:
Looks good to me, thanks!

I'll pull down and squash this into a single commit; worst case I'll lose the ancestry and then I'll attribute you in the commit message.

kimgr:
Just pushed a squashed version with a small bugfix. Thank you!

PR #448, opened 2017-06-25
Fix requiring allocator for getting container iterator with libc++.

vsapsai:
libc++ code is

    template <class _Tp, class _Allocator>
    inline _LIBCPP_INLINE_VISIBILITY
    typename vector<_Tp, _Allocator>::iterator
    vector<_Tp, _Allocator>::begin() _NOEXCEPT
    {
        return __make_iter(this->__begin_);
    }

and we reported use of allocator when visiting `vector<_Tp, _Allocator>`
because here `_Allocator` is not a default template argument. There was
no such problem in libstdc++ because there `begin` is defined inside a class.

Also I wasn't able to reproduce the issue with non-STL types because for
not precomputed types we replayed template arguments usage. And when we
populate the cache we see which template arguments are default so there
are no problems during replay.

Update test to remove method calls using object template arguments. As result,
on macOS it fixed errors

    tests/cxx/badinc.cc:1534: Unexpected diagnostic:
    I2_Enum is defined in "tests/cxx/badinc-i2.h", which isn't directly #included.

    tests/cxx/badinc.cc:1580: Unexpected diagnostic:
    I2_Enum is defined in "tests/cxx/badinc-i2.h", which isn't directly #included.

but introduced another error

    tests/cxx/badinc.cc:1527: Unmatched regex:
    I2_Enum is...*badinc-i2.h

I think it's a reasonable trade-off.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2552,17 +2552,22 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // If we're in an nns (e.g. the Foo in Foo::bar), we're never
     // forward-declarable, even if we're part of a pointer type, or in
     // a template argument, or whatever.
-    current_ast_node()->set_in_forward_declare_context(false);
+    ASTNode* ast_node = current_ast_node();
+    ast_node->set_in_forward_declare_context(false);
 
-    // When a method is defined outside of a class it has nns as qualifier.
-    // But for method call it doesn't matter if method is defined inside
-    // a class or outside of it.  Skip traversing NestedNameSpecifier
-    // in this case.
-    const ASTNode* ast_node = current_ast_node();
+    // For method calls it doesn't matter if a method is defined inside
+    // a class or outside of it.  We detect out-of-class method calls with
+    // the pattern
+    //
+    //   CallExpr
+    //   `-CXXMethodDecl
+    //     `-NestedNameSpecifier
+    //
+    // and skip traversing method qualifier as in-class methods don't have it.
     if (const CXXMethodDecl* parent = ast_node->GetParentAs<CXXMethodDecl>()) {
       if ((nns == parent->getQualifier()) &&
           ast_node->AncestorIsA<CallExpr>(2)) {
-        VERRS(7) << "Skip traversing FunctionDecl qualifier "
+        VERRS(7) << "Skipping traversal of CXXMethodDecl qualifier "
                  << PrintableNestedNameSpecifier(nns) << "\n";
         return false;
       }
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2553,6 +2553,20 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // forward-declarable, even if we're part of a pointer type, or in
     // a template argument, or whatever.
     current_ast_node()->set_in_forward_declare_context(false);
+
+    // When a method is defined outside of a class it has nns as qualifier.
+    // But for method call it doesn't matter if method is defined inside
+    // a class or outside of it.  Skip traversing NestedNameSpecifier
+    // in this case.
+    const ASTNode* ast_node = current_ast_node();
+    if (const CXXMethodDecl* parent = ast_node->GetParentAs<CXXMethodDecl>()) {
+      if ((nns == parent->getQualifier()) &&
+          ast_node->AncestorIsA<CallExpr>(2)) {
+        VERRS(7) << "Skip traversing FunctionDecl qualifier "
+                 << PrintableNestedNameSpecifier(nns) << "\n";
+        return false;
+      }
+    }
     return true;
   }
 
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -1841,7 +1841,6 @@ int main() {
   // IWYU: I1_Class needs a declaration
   H_TemplateFunction<I1_Class*>(&i1_class);
   H_TemplateFunction(&i1_class);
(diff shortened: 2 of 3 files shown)


iwyu.cc
```diff
@@ -2553,6 +2553,20 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // forward-declarable, even if we're part of a pointer type, or in
     // a template argument, or whatever.
     current_ast_node()->set_in_forward_declare_context(false);
+
+    // When a method is defined outside of a class it has nns as qualifier.
+    // But for method call it doesn't matter if method is defined inside
+    // a class or outside of it.  Skip traversing NestedNameSpecifier
+    // in this case.
+    const ASTNode* ast_node = current_ast_node();
```
kimgr: Can you move this up and use it for the call to `set_in_forward_declare_context` too?

iwyu.cc
```diff
@@ -2553,6 +2553,20 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    if (const CXXMethodDecl* parent = ast_node->GetParentAs<CXXMethodDecl>()) {
+      if ((nns == parent->getQualifier()) &&
+          ast_node->AncestorIsA<CallExpr>(2)) {
```
kimgr: `CallExpr`? Why is there a `CallExpr` above the method declaration/definition?

iwyu.cc
```diff
@@ -2553,6 +2553,20 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        VERRS(7) << "Skip traversing FunctionDecl qualifier "
```
kimgr: Maybe 'Skipping traversal of ...', the current message looks imperative.

tests/cxx/badinc.cc
```diff
@@ -1841,7 +1841,6 @@ int main() {
   // IWYU: I1_Class needs a declaration
   H_TemplateFunction<I1_Class*>(&i1_class);
   H_TemplateFunction(&i1_class);
-  // IWYU: I1_Enum is...*badinc-i1.h
```
kimgr: Cool that this disappears! Is it because the method is out-of-line-defined and the method decl thereby mentions the template argument?

kimgr:
Review comments posted.

iwyu.cc
vsapsai: I was playing with that but `set_in_forward_declare_context` is mutating and doesn't work with `const`. I still want to use const for querying ancestors and current approach seems to be short and avoiding strange naming like mutable_ast_node/ast_node or ast_node/const_ast_node.

iwyu.cc
vsapsai: The idea was that when you call a method, it shouldn't matter if it was defined out of line or inside the class. But in other cases method qualifier can be useful and we shouldn't skip visiting it. I tried to remove this check and 9 tests were failing. One of the simplest failing tests is [tests/cxx/no_deque.cc](https://github.com/include-what-you-use/include-what-you-use/blob/master/tests/cxx/no_deque.cc)

```c++
#include <queue>

int main() {
  std::queue<int> q;
  (void)q;
}
```

and without `ast_node->AncestorIsA<CallExpr>(2)` check we recommend to remove `#include <queue>`. Though I didn't check were exactly we go wrong.

But I have some doubts about these lines. I'm not sure about checking parent first and then in  nested block checking node itself and its grandparent. Maybe checking grandparent, then parent, then node itself would be less surprising and easier to understand? There will be more nesting but only 1 extra level and complexity is pretty much the same, just formatted in a different way.

iwyu.cc
vsapsai: Will change. In other places we have adding, ignoring.

tests/cxx/badinc.cc
vsapsai: Yes. At first it was a little bit surprising but I think it is correct to remove it. And we still need `I1_Enum` where we create `H_TemplateClass<I1_Enum> h_templateclass2(I11);`

vsapsai:
Thanks for the review. Responded inline. Unfortunately the assertion on line 1527 is still required on Ubuntu and removing it will break the test. I didn't investigate it further so don't know what exactly triggers its usage.

tests/cxx/badinc.cc
vsapsai: And it's nice we don't need `allocator` for `string` because we don't use allocator explicitly.

iwyu.cc
kimgr: Given the scope is so small for `ast_node`, I think the value of `const` is limited here.

iwyu.cc
kimgr: I was confused -- I thought the `Visit` was seeing the definition of a function, and couldn't see why there would be a `CallExpr` in the tree at all. But I see now that the declaration is present in the AST below the `CallExpr`.

That is, I was thinking of:

    TranslationUnitDecl
      FunctionDecl
        NestedNameSpecifier
          ...

But the actual pattern you're looking for is:

    TranslationUnitDecl
      ...
        CallExpr
          ...
          FunctionDecl
            NestedNameSpecifier
          ...

So this makes perfect sense. Maybe a comment would help describe the pattern you're looking for and why.

vsapsai:
Tried to improve the comment. Please check if it explains things better.

kimgr:
Looks good to me!

- A net reduction in test assertions for FreeBSD
- No major change on Windows, but allocator is no longer required for std::string!

Issue #447, opened 2017-06-24
Test failure on Fedora Rawhide (27)


daveisfera:
I'm trying to update the packaging on Fedora Rawhide (27) to the clang_4.0 branch ( https://github.com/include-what-you-use/include-what-you-use/commit/953970c487b004050d44925bfcc5133be9fa0d7f ) and I'm getting 1 test failing.

Here's the output:
```
FAIL: runTest (__main__.associated_h_file_heuristic)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "run_iwyu_tests.py", line 246, in <lambda>
    {'runTest': lambda self, f=filename: self.RunOneTest(f),
  File "run_iwyu_tests.py", line 217, in RunOneTest
    iwyu_flags, clang_flags, verbose=True)
  File "/builddir/build/BUILD/include-what-you-use-953970c487b004050d44925bfcc5133be9fa0d7f/iwyu_test_util.py", line 455, in TestIwyuOnRelativeFile
    test_case.assertTrue(not failures, ''.join(failures))
AssertionError: 
tests/cxx/associated_h_file_heuristic.cc:17: Unexpected diagnostic:
FILE is defined in <bits/types/FILE.h>, which isn't directly #included.

Unexpected summary diffs for tests/cxx/associated_h_file_heuristic.cc:
+++ 
@@ -1,8 +1,10 @@
 tests/cxx/associated_h_file_heuristic.cc should add these lines:
+#include <bits/types/FILE.h>
 
 tests/cxx/associated_h_file_heuristic.cc should remove these lines:
+- #include <stdio.h>  // lines XX-XX
 - #include <time.h>  // lines XX-XX
 
 The full include-list for tests/cxx/associated_h_file_heuristic.cc:
 #include "tests/cxx/internal/associated_h_file_heuristic.h"
-#include <stdio.h>  // for FILE
+#include <bits/types/FILE.h>  // for FILE
---
```

kimgr:
Aargh. I'm tempted to just add a built-in mapping from ``<bits/types/FILE.h>`` to ``<stdio.h>``. @vspasai, do you have the requisite energy for something more principled...?

vsapsai:
[...]

PR #449, opened 2017-06-26
Map FILE to <stdio.h>

kimgr:
Newer glibc has FILE in a separate header.

Fix for issue #447.


kimgr:
Oh, @daveisfera, could you try your Fedora package build with this patch applied?

vsapsai:
Hmm, requested change in GitHub UI looks more aggressive than I intended. We'll see though if the mechanism offers advantages over our less formal approach.

kimgr:
Oops, good idea. Done.

daveisfera:
For the record, all the tests now pass on Fedora Rawhide (27). Thanks for fixing this.

PR #450, opened 2017-06-28
Copy InstructionsForDevelopers to CONTRIBUTING

kimgr:
According to:
https://github.com/include-what-you-use/include-what-you-use/community

we're lacking a CONTRIBUTING.md file in the root of the project.

This is pretty much covered by docs/InstructionsForDevelopers.md, so I'd like to move into something more GitHub-friendly.

Unfortunately include-what-you-use.org links to InstructionsForDevelopers.md, so we'll do this in two steps -- first create CONTRIBUTING.md, then update the site, and then remove InstructionsForDevelopers.md.

Makes sense?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,62 @@
+## Instructions for Developers ##
+
+
+### Submitting Patches ###
+
+We welcome patches and rely on your contributions to make IWYU smarter.
+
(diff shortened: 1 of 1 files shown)


vsapsai:
I agree with the plan to migrate from InstructionsForDevelopers.md to CONTRIBUTING.md and the change looks good to me. On related subject, should we also migrate InstructionsForUsers.md to README.md and don't bother keeping them in sync?

kimgr:
Oh, good idea! I can do that as a separate PR.

PR #453, opened 2017-06-30
Add .H as a known header extension

kimgr:
No test coverage, I couldn't find any mention of the header extensions in our current test suite, and I don't think the cost outweighs the benefit here.


kimgr:
Thanks. I'll wait for some input from @tavaresdong before merging.

kimgr:
Merged this with a fix for another extension detection in fix_includes.py.

PR #466, opened 2017-08-11
pick include-what-you-use from same path as iwyu_tool.py

pseyfert:
Thereby iwyu_tool.py doesn't rely on finding include-what-you-use from
the PATH, i.e. even if iwyu_tool.py is not in the path and gets executed
by absolute path, it should still succeed in finding the executable
binary include-what-you-use.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -86,10 +86,21 @@ FORMATTERS = {
     'clang': clang_formatter
 }
 
+
 def get_output(cwd, command):
     """ Run the given command and return its output as a string. """
+    # Environment dictionary handling from https://stackoverflow.com/a/4453495
+    # ensures that: The current environment is not altered, instead it is
+    # copied for the child process, then the directory of iwyu_tool.py is
+    # prepended to the PATH variable.
+    env = os.environ.copy()
+    path_variable = os.path.dirname(__file__)
+    if 'PATH' in env:
+        path_variable += os.pathsep + env['PATH']
+    env['PATH'] = path_variable
     process = subprocess.Popen(command,
                                cwd=cwd,
+                               env=env,
                                shell=True,
                                stdout=subprocess.PIPE,
                                stderr=subprocess.STDOUT)


kimgr:
Can we instead make this into a search path, with `$PWD` at the head? It could be that my dev workflow is broken, but I set up an environment with the CMake build dir in the `PATH` and then work from the include-what-you-use source dir.

How do you work with this patch applied? Copy include-what-you-use into the source dir?

pseyfert:
I came across this drafting [this function](https://gitlab.cern.ch/lhcb/Gaudi/merge_requests/103/diffs#13be37d7401bd5ef7ac978bddb0b97b5e6f2fac3_0_43). Which leads works in the cmake run when include-what-you-use is in the path. however the ninja.build files are then written with the call to iwyu_tool.py hard coded as absolute path:
```
pseyfert@pc82 /tmp/foobarpath/LHCb/build.x86_64-centos7-gcc62-opt > ninja -n -v HltDAQ_run_iwyu_1
[1/3] cd /tmp/foobarpath/LHCb/build.x86_64-centos7-gcc62-opt && python /tmp/foobarpath/LHCb/build.x86_64-centos7-gcc62-opt/strip-json.py WORKING_DIR=/tmp/foobarpath/LHCb/build.x86_64-centos7-gcc62-opt
[2/3] cd /tmp/foobarpath/LHCb/build.x86_64-centos7-gcc62-opt/Hlt/HltDAQ && /tmp/pseyfert/tmp.iCzgEqwMfJ/install/bin/iwyu_tool.py -p /tmp/foobarpath/LHCb/build.x86_64-centos7-gcc62-opt/iwyu /tmp/foobarpath/LHCb/Hlt/HltDAQ/src/lib/HltEvaluator.cpp -- --mapping_file=/tmp/foobarpath/Gaudi/InstallArea/x86_64-centos7-gcc62-opt/cmake/IWYU_maps/include_rules.imp > /tmp/foobarpath/LHCb/build.x86_64-centos7-gcc62-opt/Hlt/HltDAQ/HltEvaluator.iwyu.log 2>&1
[3/3] cd /tmp/foobarpath/LHCb/build.x86_64-centos7-gcc62-opt/Hlt/HltDAQ && /tmp/pseyfert/tmp.iCzgEqwMfJ/install/bin/iwyu_tool.py -p /tmp/foobarpath/LHCb/build.x86_64-centos7-gcc62-opt/iwyu /tmp/foobarpath/LHCb/Hlt/HltDAQ/src/lib/HltTrackingCoder.cpp -- --mapping_file=/tmp/foobarpath/Gaudi/InstallArea/x86_64-centos7-gcc62-opt/cmake/IWYU_maps/include_rules.imp > /tmp/foobarpath/LHCb/build.x86_64-centos7-gcc62-opt/Hlt/HltDAQ/HltTrackingCoder.iwyu.log 2>&1
```
When trying to recompile in a new shell, the call to `iwyu_tool.py` would still work (gets called by its absolute path), but the `include-what-you-use` binary is not in the path anymore. I.e. at the moment I only need include-what-you-use in the `PATH` at the time `cmake` is run but not for the actual build. (and that's even up to review how we will deal with that, at the moment we don't rely on the right `g++` in the `PATH`, so it would be consistent if we don't get `include-what-you-use` from the `PATH`)

Now this is not production code yet (as you see I have include-what-you-use installed in `/tmp/`), but `include-what-you-use` will for sure neither end up in a standard installation directory nor in the llvm installation directory. It will end up in a network filesystem where only `include-what-you-use` is built. So we were aiming for a situation that relies as little as possible on the environment. With the patch I have put here, we're about down to that when we have the absolute path to `iwyu_tool.py` we "only" need libcxx in the LD_LIBRARY_PATH.

Your suggestion, making this a search path sounds good to me.

pseyfert:
Okay, @kimgr I think this is now what you suggested and covers it covers my use case.

iwyu_tool.py
```diff
@@ -86,10 +86,21 @@ def clang_formatter(output):
     'clang': clang_formatter
 }
 
+
 def get_output(cwd, command):
     """ Run the given command and return its output as a string. """
+    # https://stackoverflow.com/a/4453495
+    # do not alter the current environment,
```
kimgr: Please reflow this comment to 80 chars, use sentence case (leading uppercase char) and a trailing period. We want comments as full sentences.

iwyu_tool.py
```diff
@@ -86,10 +86,21 @@ def clang_formatter(output):
...
+    # but copy all of it for the child process,
+    # then prepend the directory of iwyu_tool.py to PATH
+    env = os.environ.copy()
```
kimgr: I'm not a huge fan of the Python "ask for forgiveness" adage, but I can't come up with a cleaner path append in short order.

iwyu_tool.py
```diff
@@ -86,10 +86,21 @@ def clang_formatter(output):
...
+    try:
+        env["PATH"] = os.path.dirname(__file__) + os.pathsep + env["PATH"]
```
kimgr: Prevalent style in this file is single-quoted strings. I found one example of a double-quoted string, but that was probably just overlooked.

iwyu_tool.py
pseyfert: @cherti just suggested an alternative to me (see updated commit)

iwyu_tool.py
```diff
@@ -86,10 +86,22 @@ def clang_formatter(output):
     'clang': clang_formatter
 }
 
+
 def get_output(cwd, command):
     """ Run the given command and return its output as a string. """
+    # The environment dictionary handling taken from
+    # https://stackoverflow.com/a/4453495 .
```
kimgr: No space before '.'

pseyfert: okay, i rephrased to avoid having the link at the end of the sentence and have the punctuation get merged with the url

pseyfert:
updated, and also rebased to the current master

kimgr:
Thanks, looks good to me!

PR #469, opened 2017-08-20
Add library dependency for backtrace

kimgr:
The out-of-tree build on my FreeBSD 11 machine failed with an unresolved
symbol for `backtrace`.

Use CMake's find provider for backtrace to satisfy the dependency.


kimgr:
Built on FreeBSD and Ubuntu 17.04, seems to be working.

Issue #456, opened 2017-07-04
iwyu_tool cannot handle compilation database with "arguments" instead of "command"


bartlibert:
The JSON [Database format specification](https://clang.llvm.org/docs/JSONCompilationDatabase.html) allows for two ways to specify the compilation commands, using either "command" or "arguments". The iwyu_tool can only handle the format with the "command" parameter, on databases that use "arguments", the tool crashes:

`
Traceback (most recent call last):
  File "/usr/bin/iwyu_tool", line 146, in <module>
    _bootstrap()
  File "/usr/bin/iwyu_tool", line 142, in _bootstrap
    sys.exit(main(args.dbpath, args.source, args.verbose, iwyu_args))
  File "/usr/bin/iwyu_tool", line 85, in main
    cwd, compile_command = entry['directory'], entry['command']
KeyError: 'command'
`

Since the [Build EAR](https://github.com/rizsotto/Bear) tool generates compilation databases in that format, it would be nice to support this as well.

kimgr:
I agree that would be useful, the `command` key always requires splitting off `argv[0]`, which is kind-of annoying.

There's one minimal complication -- we use the name of `argv[0]` to decide whether the arguments are for Microsoft's `cl.exe` or for a GCC/clang-compatible compiler. Do you think there's a way to tell from `arguments` too? Otherwise it would be useful to have `command` specified also, but separate from the arguments.

bartlibert:
In the output of bear, arguments is a JSON array where the first element is argv[0], in my case this is g++ or gcc. The spec is not fully clear, but I guess command and arguments are exactly the same, just in a different format.

kimgr:
Ah, so it's just a matter of space-separated string (command) vs actual list (arguments)? Well, that makes it even easier to support :-) I'll try to take a look at this at some point.

kimgr:
Can you test if #470 fixes this for you?

bartlibert:
Yep, everything works great with that version!

PR #470, opened 2017-08-20
Support 'arguments' key in compilation databases

kimgr:
Move things around a little to make this easier to implement -- run_iwyu
now translates from a compilation database entry to a command-line.

This should fix issue #456.


kimgr:
Manually rebased on top of latest master.

PR #473, opened 2017-08-24
Fixed CMakeLists.txt for build with LLVM 4.0.1

adhokshajmishra:


kimgr:
Can you make the change in a topic branch off your repo's `clang_4.0` branch, and create a PR off of that instead?

We've seen some LLVM problems when building 4.0.1, so I'm not sure how to take the patch anyway, but it would be nice to see where we have specific compat issues.

adhokshajmishra:
Sure, will do that :)

kimgr:
This doesn't look like it's going anywhere. Please open a new PR if you want to pursue this.

PR #477, opened 2017-08-30
Improve path normalization in windows

jru:
When using --clang-mode=cl and the flag /I., paths will be reported relative to the current directory. To make sure associate includes are always found, convert to full path before making canonical and
comparing. Additionally, make all lowercase in GetCanonicalName() as the last step, only for windows systems.

Also fixed iwyu_tool.py

### DIFF
--- a/iwyu_location_util.h
+++ b/iwyu_location_util.h
@@ -92,6 +92,11 @@ inline string GetFilePath(const clang::FileEntry* file) {
           NormalizeFilePath(file->getName()));
 }
 
+inline string GetFileFullPath(const clang::FileEntry* file) {
+  return (IsBuiltinFile(file) || file->tryGetRealPathName().empty())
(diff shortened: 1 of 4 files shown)


jru:
Just realized this one is very similar to #294, is the idea to merge it? It seems to have been idling for a while now..

kimgr:
Both this and #294 are probably on to something. But this PR has no tests, and #294 has a rewrite of the entire testing infrastructure, in a way that requires running all tests twice.

I've made some attempts at smaller surgery to give individual tests more control over path format, but I haven't gotten anything working yet. This bug (or possibly bugs?) is bad enough that I want to get it solved, but we need solid testing for it. I'll take a look at options for the testing infrastructure again. Thanks!

kimgr:
@jru Is this still misbehaving after PR #681 was merged?

kimgr:
No activity here for a while, closing for now.

PR #478, opened 2017-08-31
Add directory support to iwyu_tool.py

jru:
For each directory that is given, iwtyu_tool will search for all source files contained in it (or in its subdirectories) and process them all.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -145,9 +145,8 @@ def main(compilation_db_path, source_files, verbose, formatter, jobs, iwyu_args)
         # warn for the rest.
         entries = []
         for source in source_files:
-            matches = [e for e in compilation_db if e['file'] == source]
-            if matches:
-                entries.extend(matches)
+            if os.path.isdir(source) or os.path.isfile(source):
+                entries.extend([e for e in compilation_db if e['file'].startswith(source)])
             else:
                 print('WARNING: \'%s\' not found in compilation database.' %
                       source)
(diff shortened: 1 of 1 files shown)


kimgr:
The Windows programmer in me likes this. The UNIX programmer in me doesn't understand why it's necessary; this is how you'd typically do this in a UNIX environment:

    $ find some/sub/dir -name "*.cpp" -exec iwyu_tool.py -p build/ {}  \;

I'd prefer to avoid piling functionality onto `iwyu_tool.py` -- ideally it should disappear in the future. Is there a way for you to do this in Windows? E.g. something like:

    FOR f in (some\sub\dir\*.cpp) DO ( python iwyu_tool.py -p build %f )

?

jru:
Your examples are not completely equivalent to accepting directories since they would be executed in a serial fashion, defeating the purpose of this little wrapper tool (which I think is great to have, btw!). But, anyway, sure we could find some command line trick that does the same thing, assuming the command-line limit on windows doesn't hit.

My point was that it felt a very natural extension to the tool (in fact, the first time I used I was surprised it didn't work) given that it already accepts one, few and all files, why not all the files within a directory?

At least for me, it is a extremely common use case to want process files one sub-project at a time, especially for large code bases. But I am not going to lose my head over it, I just though it would be useful :)

Note: Just made the change even tinier.

kimgr:
Meh, what the heck. This looks good to me, let's merge it!

Issue #479, opened 2017-09-10
fix_includes.py --dry_run exit code does not match documentation


suertreus:
The documentation for fix_includes.py states:

> The exit code is the number of files that were modified (or that would
> be modified if --dry_run was specified) unless that number exceeds 100,
> in which case 100 is returned.

However in --dry_run mode it is zero regardless of the number of files that would be modified.

Here's a simple repro case where fix_includes.py ought to exit with status 1 but instead exits with status 0.

indirect.h:
`void Symbol() {}`

direct.h:
`#include "indirect.h"`

source.cc:
```
#include "direct.h"  // should be "indirect.h"
void Function() { Symbol(); }
```

```
$ include-what-you-use source.cc 2> output
$ fix_includes.py -n source.cc < output 
>>> Fixing #includes in 'source.cc'
@@ -1,2 +1,3 @@
-#include "direct.h"  // should be "indirect.h"
+#include "indirect.h"
+
 void Function() { Symbol(); }
IWYU edited 0 files on your behalf.

$ echo $?
0
```

FWIW I would prefer to see this resolved by changing the code to match the documentation, not vice-versa.

kimgr:
Thanks, makes sense. This should be easy to patch in `FixManyFiles`; I don't have enough of an environment up to test right now.

kimgr:
Can you apply #482 and see if it matches your expectations better?

suertreus:
Works great, thanks!

kimgr:
Thanks, merged!

PR #482, opened 2017-09-16
Issue #479: Make dry-run return proper exit code

kimgr:
This makes the script match the documentation better.


kimgr:
Built on Ubuntu 17.04 and FreeBSD, seems to be working.

PR #485, opened 2017-10-01
Skip non-ctors during instantiation of implicit methods

### LINKED ISSUE #402: C++11 inheriting constructors cause IWYU assertion failures
The following code snippet causes crashes under IWYU 0.5 - 0.7 for me:

```C++
struct Base { Base(int); };
struct Derived : Base { using Base::Base; };
```

Under 0.5 and 0.6 this hits a [possibly bogus assert](https://github.com/include-what-you-use/include-what-you-use/blob/eb7f5d1ef02836575aebfe5a95977f067064cee2/iwyu_output.cc#L1944):

```
$ docker run -it --rm ubuntu:yakkety-20170104
# apt update && apt install iwyu clang-3.8
...
# cat > test.cpp
...
# clang++-3.8 -c -std=c++11 test.cpp; echo $?
0
# iwyu -c -std=c++11 test.cpp; echo $?
/build/iwyu-1de5QU/iwyu-3.8/iwyu_output.cc:1912: Assertion failed: using_decl->shadow_size() > 0
Aborted (core dumped)
134
# clang++-3.8 --version; iwyu --version
clang version 3.8.1-12ubuntu1 (tags/RELEASE_381/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
include-what-you-use 0.6 based on clang version 3.8.0-2ubuntu3 (tags/RELEASE_380/final)
```

And under 0.7 something scarier happens:

```
$ docker run -it --rm ubuntu:zesty-20170118
# apt update && apt install iwyu clang-3.9
...
# cat > test.cpp
...
# clang++-3.9 -c -std=c++11 test.cpp; echo $?
0
# iwyu -c -std=c++11 test.cpp; echo $?
iwyu: /usr/lib/llvm-3.9/include/llvm/Support/Casting.h:237: typename llvm::cast_retty<X, Y*>::ret_type llvm::cast(Y*) [with X = clang::CXXConstructorDecl; Y = clang::NamedDecl; typename llvm::cast_retty<X, Y*>::ret_type = clang::CXXConstructorDecl*]: Assertion `isa<X>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
Aborted (core dumped)
134
# clang++-3.9 --version; iwyu --version
clang version 3.9.1-2ubuntu1 (tags/RELEASE_391/rc2)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
include-what-you-use 0.7 based on clang version 3.9.1-1ubuntu1 (tags/RELEASE_391/rc2)
```

jru:
InstantiateImplicitMethods() wrongly assumed a ctor decl must be present just by checking that it is not a template. This is no longer true in C++11 where a using decl is also a valid ctor type.

Fixes #402

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -163,6 +163,7 @@ using clang::CallExpr;
 using clang::ClassTemplateDecl;
 using clang::ClassTemplateSpecializationDecl;
 using clang::CompilerInstance;
+using clang::ConstructorUsingShadowDecl;
 using clang::Decl;
 using clang::DeclContext;
 using clang::DeclRefExpr;
@@ -618,8 +619,10 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
     clang::Sema& sema = compiler_->getSema();
     DeclContext::lookup_result ctors = sema.LookupConstructors(decl);
     for (NamedDecl* ctor_lookup : ctors) {
-      // Ignore templated constructors.
-      if (isa<FunctionTemplateDecl>(ctor_lookup))
+      // Ignore templated or inheriting constructors.
+      if (isa<FunctionTemplateDecl>(ctor_lookup) ||
+          isa<UsingDecl>(ctor_lookup) ||
+          isa<ConstructorUsingShadowDecl>(ctor_lookup))
         continue;
       CXXConstructorDecl* ctor = cast<CXXConstructorDecl>(ctor_lookup);
       if (!ctor->hasBody() && !ctor->isDeleted() && ctor->isImplicit()) {
--- a/tests/cxx/inheriting_ctor-i1.h
+++ b/tests/cxx/inheriting_ctor-i1.h
@@ -1,0 +1,20 @@
+//===--- inheriting_ctor-i1.h - test input file for iwyu ------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_INHERITING_CTOR_I1_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_INHERITING_CTOR_I1_H_
+
+struct Base {
+  Base(int);
+};
+struct Derived : Base {
+  using Base::Base;
+};
+
+#endif  // INCLUDE_WHAT_YOU_USE_TESTS_CXX_INHERITING_CTOR_I1_H_
--- a/tests/cxx/inheriting_ctor.cc
+++ b/tests/cxx/inheriting_ctor.cc
@@ -1,0 +1,26 @@
+//===--- inheriting_ctor.cc - test input file for iwyu --------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "inheriting_ctor-d1.h"
+
+// IWYU: Derived is defined in .*-i1.h
+void func() { Derived d(1); }
+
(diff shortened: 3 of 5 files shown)


iwyu.cc
```diff
@@ -618,8 +618,8 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
     clang::Sema& sema = compiler_->getSema();
     DeclContext::lookup_result ctors = sema.LookupConstructors(decl);
     for (NamedDecl* ctor_lookup : ctors) {
-      // Ignore templated constructors.
-      if (isa<FunctionTemplateDecl>(ctor_lookup))
+      // Ignore templated or inheriting constructors.
+      if (!isa<CXXConstructorDecl>(ctor_lookup))
```
kimgr: I think it would be nicer to do this as a `dyn_cast` now:

     CXXConstructorDecl* ctor = dyn_cast<CXXConstructorDecl>(ctor_lookup);

     // Ignore templated or inheriting constructors
     if (!ctor)
       continue;

    ...

tests/cxx/inheriting_ctor-i1.h
```diff
@@ -0,0 +1 @@
+#include "inheriting_ctor-i2.h"
```
kimgr: This needs a license header and include guard

tests/cxx/inheriting_ctor-i2.h
```diff
@@ -0,0 +1,4 @@
+
```
kimgr: License header and include guard

tests/cxx/inheriting_ctor-i1.h
kimgr: Also, I think this should be called `...-d1.h` -- I believe 'd' is for 'direct' and 'i' is is for 'indirect'.

tests/cxx/inheriting_ctor.cc
```diff
@@ -0,0 +1,18 @@
+
```
kimgr: License header

tests/cxx/inheriting_ctor.cc
```diff
@@ -0,0 +1,18 @@
...
+tests/cxx/inheriting_ctor.cc should add these lines:
+#include "inheriting_ctor-i2.h"
+
+tests/cxx/inheriting_ctor.cc should remove these lines:
+- #include "inheriting_ctor-i1.h"  // lines XX-XX
+
+The full include-list for tests/cxx/inheriting_ctor.cc:
+#include "inheriting_ctor-i2.h"  // for Derived
+
+***** IWYU_SUMMARY */
```
kimgr: Please add trailing newline

tests/cxx/inheriting_ctor.cc
```diff
@@ -0,0 +1,18 @@
+
+#include "inheriting_ctor-i1.h"
+
+//IWYU: Derived is defined in .*-i2.h
+void func() { Derived(1); }
```
kimgr: This would be easier to read if you broke it over multiple lines and named the `Derived` instance, something like:

    void func() {
      // IWYU ...
      Derived d(1);
    }

iwyu.cc
kimgr: Hmm, but it also makes this code a little more bug-prone. The current code explicitly spells out FunctionTemplateDecl as an exception to the norm (CXXConstructorDecl). It would be more conservative and informative to just add UsingDecl (which I guess this is?), something like:

    // Ignore templated and inheriting constructors.
    if (isa<FunctionTemplateDecl>(ctor_lookup) || isa<UsingDecl>(ctor_lookup))
      continue;

and then leave the `cast` as-is, to make sure we break if something unexpected creeps in here.

kimgr:
Nice, thanks! Please find some review comments inline.

jru:
All changes addressed.

kimgr:
Great, thanks! Can you also re-word the commit message to mention that it fixes #402?

jru:
Done

kimgr:
Thank you!

PR #489, opened 2017-10-05
Fix Python3 TypeError in fix_includes.py

dakotahawkins:
The error is "TypeError: a bytes-like object is required, not 'str'".

Because the file is opened in binary mode, we need to call .encode() on
stings being written.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -543,8 +543,8 @@ def _ReadWriteableFile(filename, ignore_writeable):
 
 def _WriteFileContentsToFileObject(f, file_lines, line_ending):
   """Write the given file-lines to the file."""
-  f.write(line_ending.join(file_lines))
-  f.write(line_ending)
+  f.write(line_ending.join(file_lines).encode())
+  f.write(line_ending.encode())
 
 def _DetectLineEndings(filename):
   """Detect line ending of given file."""


dakotahawkins:
@kimgr I looked at how the tests work, and because none of them (I think) use actual files it might be difficult to test. I think they'd either have to use real files or some kind of in-memory stream that supports the same binary open/read/write interface. The takeaway is that as-is I don't think anything is testing the binary read/write that lets you keep line-endings in the real code.

This also exposed another issue with some files I had that weren't `ascii` (what this will use on my system, anyway) originally. Some cursory googling left me with this:

```python
def _WriteFileContentsToFileObject(f, file_lines, line_ending):
  """Write the given file-lines to the file."""
  try:
    f.write(line_ending.join(file_lines).encode('utf-8'))
    f.write(line_ending.encode('utf-8'))
  except UnicodeDecodeError:
    f.write(line_ending.join(file_lines).decode('cp1252').encode('utf-8'))
    f.write(line_ending.decode('cp1252').encode('utf-8'))
```

That worked for my files, but obviously it's not ideal. Do you have any ideas about handling different encodings? In this case, there was no reason for that file not to be in `ascii` or `utf8`.

I think at the very least an exception there shouldn't bring the whole process down -- as-is it can try to fix hundreds of files and wind up fixing nothing because it had a problem with one of them.

kimgr:
Hmm. It would be nice if fix_includes.py could work exclusively with `bytes`, since it's only modifying existing files, and only adding ASCII content. But that is probably a pretty massive change, especially in a way that works with both py2 and 3.

Just for the record, I found that `_WriteFileContents`, the only caller of `_WriteFileContentsToFileObject` was stubbed out in the test, so that's why this wasn't covered. I didn't see any obvious way to fix that in a meaningful way.

I'll try to look into this again soon.

dakotahawkins:
I think it can't work with `bytes` as is because the lines are implicitly converted to `str`s, and they probably need to be that way to process them in any meaningful way.

What about pulling in (or just requiring, or maybe optionally using if installed) [chardet](https://github.com/chardet/chardet), which will work with Python 2/3, and using that to try to write the file back out in it's original encoding?

kimgr:
I'm reluctant to add third-party dependencies, so let's just iron this out bit by bit. I'll try to do some work on this today.

dakotahawkins:
It's a big dependency, but a non-trivial problem. Maybe spitting out an error with the filename would be a good option if you can make it do that, I need to fix all those files in our codebase anyway, but to do it I'd have to run chardet or something on all of them and then something else to convert them :(

kimgr:
Hey @dakotahawkins -- can you attach some source files with exotic encodings for me to test with? Anything synthetic would be welcome, as long as they expose the same kind of problems you're seeing in the real world. I'm interested in both the original case, and the followup that drove you to use cp1252.

Thanks!

dakotahawkins:
The `.decode('cp1252').encode('utf-8')` came from a windows/VS-specific StackOverflow answer, and I'm pretty sure I just got lucky. Or maybe cp1252 maps to enough things that it happened to work.

I could make some files, but it might take me some time. I know we've had issues in the past with engineers putting unicode symbols in comments (like math symbols) and visual studio auto-changing the encoding, but I'm guessing that to cause a failure in a synthetic file I will have to figure out what character we have that doesn't map to utf-8, which might be tricky (hex-editor diff against a converted file, I guess).

If you want to take a faster stab at it, uncoding some file in utf-16 (not sure what you'd use, I'd use Notepad++ probably) might bring it down. Maybe shove an emoji in a comment somewhere to spice things up. Otherwise I'll try to find a specific encoding/character that breaks it and make a file for you.

Basically, I don't know what `str` is doing behind the scenes and if python is trying to map characters and chokes on a few that it can't or if it bails as soon as it sees a different encoding anywhere :(

Back to the original issue, python3 wants (requires) `encode()` to be called there, and even then anything it can't handle currently raises an exception that brings down the whole run. I think most people should be able to live with catching those, building a list of problem files, and asking the user to fix them and try again at the end. Does that make sense?

dakotahawkins:
Here's a couple of files that may or may not cause problems:

[wat.zip](https://github.com/include-what-you-use/include-what-you-use/files/1413115/wat.zip)

kimgr:
Thank you, and sorry it took me so long to circle back to this.

I have a hacky worktree locally with home-cooked heuristic encoding detection that seems to work for both python2 and python3. I'm working on the assumption that source files come in a small, limited number of encodings (currently ascii, windows-1250, windows-1252 and utf-8). This does not cover all possible inputs, but I think it's likely to cover a large percentage. It's systematic enough that we should be able to support more encodings by adding their name to the list as problems are discovered.

A couple of observations/insights that came up when I worked through this:

- IWYU only ever adds ASCII content to the file (any non-ASCII comes from the input itself)
- Even if we guess the encoding wrong (e.g. something that's nominally windows-1252 but can be decoded as windows-1250), I _think_ it should be fine, as the non-ASCII content is simply roundtripped.

It would be nice to have test coverage for this. I built a little one-off harness based on your test files, but it's hard to fold this into `fix_includes_test.py` as it must be cross-tested with both python2 and 3. We might consider using tox for Python testing, but I'd rather not spend time on that right now.

I'll try and clean this up for easier review and post, but I'm not sure how soon I'll have time to put it together.

kimgr:
Superseded by #528

Issue #293, opened 2016-04-19
Add CI suite and code coverage


Flamefire:
I'm surprised to see no CI set up for IWYU. It is pretty easy (travis), widely used already and allows to check, if IWYU compiles with different compilers and OSs and also run the runtime test and produce coverage results.
The last point gives a great tool to check if the current tests fail to tests some code portions of IWYU

vsapsai:
There is a buildbot at http://buildbot.include-what-you-use.com:8010/grid and there is a link to it from http://include-what-you-use.org/ Corresponding repo is [include-what-you-use/iwyu_buildbot](https://github.com/include-what-you-use/iwyu_buildbot).

As for [travis-ci.org](https://travis-ci.org/), I'm not sure it can build IWYU. Main source of build breakage is API changes in Clang. So we want to build IWYU with the latest Clang. The problem is that LLVM+Clang build takes time and can exceed Travis CI limits ([documented here](https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts)). I didn't try Travis CI, anybody can try to set up full build with it and see if it works.

As for code coverage, do you expect from web version to give you something that cannot be achieved with checking code coverage locally? I agree that code coverage is nice to have but I think there are more important issues to fix. Just want to clarify that I'm not missing some use cases.

Flamefire:
As for travis: I'd rather suggest to stick to a specified version of clang and update manually if required. This has 2 advantages: 1. It is clear, with which version IWYU is/can be build 2. One can use the travis cache, which allows skipping the rebuild of clang when it is not changed.
The container based (sudo-less) environment has got 2 cores for compiling and is often faster anyways. So parallel compiling may be enough to stay in the 50 minutes limit. Although I remember that it did took a lot of time to compile clang. If it is not enough, one can probably use one of the prebuild-binaries (http://llvm.org/releases/download.html)

As for coverage: Sure. But I found the code coverage setup to be "easy enough" to have (although I'm not sure about the python wrapper and the multiple execution here) And it can provide useful without having individuals run this. They can just look at the visualisation, see that some part is not covered and add a test. Or see, when an addition reduces coverage e.g. by a misplaced `if`

kimgr:
If you can get an environment up and running, I would be interested in seeing it (I know GitHub integrates well with Travis, too, so it would be generally helpful.)

[...]

PR #490, opened 2017-10-08
Create a travis script to build with latest LLVM

jru:
Hi,

After some fighting I got the Travis integration working. Phew!

You can check the result of the latest run (as configured on my fork) here: 
https://travis-ci.org/jru/include-what-you-use/builds/285073561

By default it will build PRs and master immediately after pushing. You just need to configure you account and enable the iwyu repo by going to https://travis-ci.org/

I figured it would be most useful if we keep building against the latest LLVM snapshot (as published in apt.llvm.org) so that we can react proactively should something break. What do you think?

Cheers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,38 @@
+
+language: cpp
+
+addons:
+  apt:
+    sources:
+     - ubuntu-toolchain-r-test
+     - llvm-toolchain-trusty
+    packages:
+     - ninja-build
+     - llvm-dev
+     - libclang-dev
+     - clang
+
+before_install:
+ # Install a supported cmake version (>= 3.4.3)
+ - wget -O cmake.sh https://cmake.org/files/v3.10/cmake-3.10.0-rc1-Linux-x86_64.sh 
+ - sudo sh cmake.sh --skip-license --exclude-subdir --prefix=/usr/local
+
+ # Extract the version number from the most-recently installed LLVM
+ - VERSION=`ls -t /usr/lib/ | grep '^llvm-' | head -n 1 | sed -E 's/llvm-(.+)/\1/'`
+
+ # Absolute paths to LLVM's root and bin directory
+ - ROOT_PATH=`llvm-config-$VERSION --prefix`
+ - BIN_PATH=`llvm-config-$VERSION --bindir`
+
+script:
+# Build IWYU
+ - mkdir build
+ - cd build
+ - cmake -GNinja -DIWYU_LLVM_ROOT_PATH=$ROOT_PATH -DCMAKE_C_COMPILER=$BIN_PATH/clang -DCMAKE_CXX_COMPILER=$BIN_PATH/clang++ -DCMAKE_INSTALL_PREFIX=./ ../
+ - ninja install
+
+# Test IWYU
+ - cd ..
+ - python run_iwyu_tests.py -- build/bin/include-what-you-use
+ - python fix_includes_test.py
+
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Include What You Use #
 
+[![Build Status](https://travis-ci.org/include-what-you-use/include-what-you-use.svg?branch=master)](https://travis-ci.org/include-what-you-use/include-what-you-use)
+
 For more in-depth documentation, see http://github.com/include-what-you-use/include-what-you-use/tree/master/docs.
 
 


kimgr:
Thank you! This looks cool.

A few questions:

- I see this builds on some Ubuntu -- do we know which version?
- LLVM/Clang are installed with a version spec. Does that mean they'll stick to v6.0 even after trunk has moved to 7.0? Can we omit the version, or does that cause other problems?
- Rather than pulling CMake off the web, is there a repo with recent releases we could point to under `sources` instead?
- How hard is it to add support for another platform (Windows)? Would that go in the same travis.yml? I'm not suggesting we do it now, but I'd like to know that it's possible and how to do it

I know preciously little about Travis, so these may be stupid questions :)

janisozaur:
> * I see this builds on some Ubuntu -- do we know which version?

14.04.5 "Trusty". It is the default on Travis (https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments). You can also see this under `Build system information`: https://travis-ci.org/jru/include-what-you-use/builds/285073561#L18 and guess from the installed repo name: `llvm-toolchain-trusty`

> * LLVM/Clang are installed with a version spec. Does that mean they'll stick to v6.0 even after trunk has moved to 7.0? Can we omit the version, or does that cause other problems?

The change explicitly requests llvm-6.0. It is required to specify the version, if you want anything else than the default. The way packages are done imposes the version information in them, usually `llvm` would be the transitive package pulling the default, e.g. `llvm-3.9`. Do note that you can have different ways to install packages based on which kind of virtualisation you request from Travis. In this case it is container, which means you can only install [whitelisted packages](https://github.com/travis-ci/apt-package-whitelist) from [whitelisted repos](https://github.com/travis-ci/apt-source-whitelist).

> * Rather than pulling CMake off the web, is there a repo with recent releases we could point to under sources instead?

See previous point and note no `cmake` other than default in here: https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-trusty

> * How hard is it to add support for another platform (Windows)? Would that go in the same travis.yml? I'm not suggesting we do it now, but I'd like to know that it's possible and how to do it

Travis only [offers Linux and macOS workers](https://docs.travis-ci.com/user/multi-os/), with latter being run on much smaller farm due to licensing and cost reasons. There is, however, another service similar to Travis (also free for open-source): https://www.appveyor.com/ which offers Windows VMs for CI purposes. I have no idea how hard it is to get clang/llvm on Windows, but they offer some pre-installed versions too, currently 4.0.0: https://www.appveyor.com/docs/build-environment/#llvm

jru:
Thanks for the info @janisozaur. 

I was pretty much settled on having to provide the version but after experimenting a bit I managed to get the latest LLVM picked by the script using some bash magic. The most recent LLVM seems to be installed by default if the version is left out but it is not set as the default. It feels hacky but it works :)

Regarding CMake, I could not find any apt source that supplies 3.4.3 or greater, so I decided to go the easy way and download it directly.

In the meantime I will try AppVeyor. I don't think it brings so much now given that the test suite does not run clean, but at least we could build it.

.travis.yml
```diff
@@ -0,0 +1,33 @@
...
+
+ # Extract the version number from the most-recently installed LLVM
+ - VERSION=`ls -t /usr/lib/ | grep '^llvm\-' | head -n 1 | sed 's/llvm\-\(.\+\)/\1/'`
+
+ - ROOT_PATH=`llvm-config-$VERSION --prefix`
+ - BIN_PATH=`llvm-config-$VERSION --bindir`
+
+script:
+# Build IWYU
+ - mkdir build && cd build && cmake -GNinja -DIWYU_LLVM_ROOT_PATH=$ROOT_PATH -DCMAKE_C_COMPILER=$BIN_PATH/clang -DCMAKE_CXX_COMPILER=$BIN_PATH/clang++ -DCMAKE_INSTALL_PREFIX=./ ../
```
kimgr: Could you split this line into the three steps (mkdir, cd, cmake) instead? Or does the one-liner add something?

.travis.yml
```diff
@@ -0,0 +1,33 @@
...
+     - libclang-dev
+     - clang
+
+before_install:
+ # Install a supported cmake version (>= 3.4.3)
+ - wget -O cmake.sh https://cmake.org/files/v3.10/cmake-3.10.0-rc1-Linux-x86_64.sh 
+ - sudo sh cmake.sh --skip-license --exclude-subdir --prefix=/usr/local
+
+ # Extract the version number from the most-recently installed LLVM
+ - VERSION=`ls -t /usr/lib/ | grep '^llvm\-' | head -n 1 | sed 's/llvm\-\(.\+\)/\1/'`
```
kimgr: The escaping of `\+` looks wrong here, but apparently it works...? I would guess `\(.+\)` would be a functional capturing group. Likewise, the escaping of `-` is probably not necessary in the sed expression (grep requires it, IIRC).

.travis.yml
```diff
@@ -0,0 +1,33 @@
...
+ - ROOT_PATH=`llvm-config-$VERSION --prefix`
+ - BIN_PATH=`llvm-config-$VERSION --bindir`
+
+script:
+# Build IWYU
+ - mkdir build && cd build && cmake -GNinja -DIWYU_LLVM_ROOT_PATH=$ROOT_PATH -DCMAKE_C_COMPILER=$BIN_PATH/clang -DCMAKE_CXX_COMPILER=$BIN_PATH/clang++ -DCMAKE_INSTALL_PREFIX=./ ../
+ - ninja install
+
+# Test IWYU
+ - cd .. && python run_iwyu_tests.py -- $TRAVIS_BUILD_DIR/build/bin/include-what-you-use
```
kimgr: Since you must be in `$TRAVIS_BUILD_DIR` at this point, I wonder if it's necessary to spell it out. Maybe just `build/bin/include-what-you-use`?

kimgr: Also, could this be split onto two lines?

README.md
```diff
@@ -1,3 +1,5 @@
+[![Build Status](https://travis-ci.org/jru/include-what-you-use.svg?branch=master)](https://travis-ci.org/jru/include-what-you-use)
```
kimgr: Reminder for myself: this needs to be updated before we merge, to point to an "official" IWYU build.

jru:
Hi,

We could definitely find a way to skip the failing tests directly in the build script, but I think it would be much easier to extend the runner to ignore certain tests passed by argument.

Regarding PRs, yes, that script will be run for all of them and there will be some notifications sent informing of failure or success. That should make reviewing small changes much easier as you won't necessarily need to build & test locally.

I will address your comments soon! Thanks.

.travis.yml
jru: Yeah, I think can simplify it by using the extended regex mode (-E). I never remember which mode escapes what.

.travis.yml
jru: Each line in the script block will be executed even if the previous failed. So, it adds nothing really for *mkdir* and *cd* since they can't fail, but for *cmake* and *ninja* I think it makes sense. I'll change it as you suggested

README.md
jru: I can do that too!

jru:
Done. 

Btw, the Travis build seems to be failing now due to the llvm package being unavailable (404). I hope is a transient issue..

janisozaur:
> How does this work, admin-wise -- does travis-ci.org pick up automatically or do we need to register an account, or something?

You need to _register_ with Travis. Note that it will use your GitHub login information, so there's no need to create another login. I think it can be anyone who is a member of https://github.com/include-what-you-use organization. Then, go hover over your login, go to "Accounts" and switch to "include-what-you-use" and enable this repository.

It should automatically pick up any new commits to the repository and new PRs (do check in options that they are enabled; as this one is already created, it won't get picked up unless re-PR'd).

Do note the above statement means that PRs coming from branches inside this repository will get two distinct runs – one "push" and one "PR".

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+# Build IWYU
+ - mkdir build
+ - cd build
+ - cmake -GNinja -DIWYU_LLVM_ROOT_PATH=$ROOT_PATH -DCMAKE_C_COMPILER=$BIN_PATH/clang -DCMAKE_CXX_COMPILER=$BIN_PATH/clang++ -DCMAKE_INSTALL_PREFIX=./ ../
+ - ninja install
+
+# Test IWYU
+ - cd ..
+ - python run_iwyu_tests.py -- build/bin/include-what-you-use
+
```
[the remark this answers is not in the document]
jru: Sure thing!

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+ # Absolute paths to LLVM's root and bin directory
+ - ROOT_PATH=`llvm-config-$VERSION --prefix`
+ - BIN_PATH=`llvm-config-$VERSION --bindir`
+
+script:
+# Build IWYU
+ - mkdir build
+ - cd build
+ - cmake -GNinja -DIWYU_LLVM_ROOT_PATH=$ROOT_PATH -DCMAKE_C_COMPILER=$BIN_PATH/clang -DCMAKE_CXX_COMPILER=$BIN_PATH/clang++ -DCMAKE_INSTALL_PREFIX=./ ../
+ - ninja install
```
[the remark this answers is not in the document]
jru: It searches both, in the system dirs and in the installation directory. Not sure which ones, but I do rememeber seeing it fail if I didn't install it.

kimgr:
Thanks, I'll register now.

> Do note the above statement means that PRs coming from branches
> inside this repository will get two distinct runs – one "push" and one "PR".

We typically don't create PRs from this repo, iwyu maintainers have forks and PR from there, so this should not be an issue. But thanks for bringing it up, I could see myself getting confused!

kimgr:
There, I have registered `include-what-you-use/include-what-you-use` now, to only build if `.travis.yml` is available. So it *should* pick up when this is merged.

.travis.yml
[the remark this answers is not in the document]
kimgr: OK, that explains it, thanks!

jru:
Just updated the PR with the corrections and Jenkins picked it up already. First green build :)

I added fix_includes_test and squashed the corrections. Should I also mention in the commit that it fixes #293?

kimgr:
Very nice! Thank you for working on this.

PR #504, opened 2017-11-30
Use LLVM 5.0 for this branch.

kimgr:
Experiment to see if this builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,24 +5,21 @@ addons:
   apt:
     sources:
      - ubuntu-toolchain-r-test
-     - llvm-toolchain-trusty
+     - llvm-toolchain-trusty-5.0
     packages:
      - ninja-build
-     - llvm-dev
-     - libclang-dev
-     - clang
+     - llvm-5.0-dev
+     - libclang-5.0-dev
+     - clang-5.0
 
 before_install:
  # Install a supported cmake version (>= 3.4.3)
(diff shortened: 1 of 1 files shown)


kimgr:
Umm, no, apparently not.

kimgr:
@jru Could you please take a look and see if this makes sense?

I want to make a similar change every time we take a release branch, to make sure the branch stays true to upstream and vice versa. This should *not* go onto master, which should continue to build against LLVM/Clang snapshots.

I tried putting in a variable for the LLVM version, but I never got that to work across the apt sources/packages, so I gave up. It's simple enough to change if you know which packages are affected.

Let me know what you think!

jru:
Hi! I think is a good idea. So, are you planning to rebase this branch onto master until LLVM 6.0 is out? 

Regarding the patch, I would also get rid of the `VERSION` var and hardcode it to 5.0 for the two llvm-config lines, just in case there is a newer one installed in the VM.

jru:
I just realized that there are clang_X.Y branches for old LLVM releases already.. :)
Also, correcting myself above: rebasing would only work if the commits with breaking changes on master are reverted on the clang_5.0 branch.

kimgr:
Good point on VERSION, thanks. And yeah, we typically branch only as soon as it is necessary, because llvm trunk breaks iwyu master. Any compatible changes are cherry-picked onto the branch until release.

PR #506, opened 2017-12-10
Use PRIVATE in target_link_libraries

kimgr:
LLVM introduced PRIVATE scope for target_link_libraries for executables
in r319840. Since executables don't export link dependencies, this makes
sense.

CMake does not allow mixing target_link_libraries invocations
with/without scope specifiers, so add PRIVATE in IWYU's CMakeLists.txt


kimgr:
The Travis failure is orthogonal to this change -- a recent change in Clang caused the test failure above. Still haven't isolated what.

Issue #354, opened 2016-08-16
Is it possible to make IWYU not add forward declarations?


EFanZh:
I don’t care if my code’s compiling time is a little longer, but I want my code to look clean and short. Is it possible to make IWYU only organize includes, without introducing any forward declarations?

kimgr:
No, it's not currently possible, but I'm starting to think it might be a nice feature.

kimgr:
Also requested on the mailing list here:
https://groups.google.com/d/topic/include-what-you-use/ITG9mhL-Mt0/discussion

cdglove:
On Tue, 16 Aug 2016 at 15:00 Kim Gräsman <notifications@github.com> wrote:

> No, it's not currently possible, but I'm starting to think it might be a
> nice feature.
>
In my opinion, this is an odd thing to want -- I try to fwd-declare
everything and tell my teams to do the same. I can't think of a good reason
not to do that.

If this is really a feature people want then perhaps it should be added,
but it might also be a point that requires education over a technical fix.
Thoughts?

-- chris

>

kimgr:
Personally I agree, but I think the recent trend comes from Google's updated style guide:
https://google.github.io/styleguide/cppguide.html#Forward_Declarations

I think their motivation holds water for the most part, forward declarations are really brittle in C++.

So for small code bases (not Google), or when you can massively parallellize and cache your builds (Google), this might actually be good advice. For the rest of us; not so much :-)

cdglove:
On Tue, 16 Aug 2016 at 16:01 Kim Gräsman <notifications@github.com> wrote:

> Personally I agree, but I think the recent trend comes from Google's
> updated style guide:
> https://google.github.io/styleguide/cppguide.html#Forward_Declarations
>
> I think their motivation holds water for the most part, forward
> declarations are really brittle in C++.
>
Well, the points are valid, I just don't agree with the actual severity of
the problem -- I have never been bitten by any of the issues listed there
(other than the verbosity of fwd-declarations).

Of course, the size of the projects makes a difference, and I might
misunderstand the scale, but the projects I work on are not exactly small
-- they are somewhere between 10 and 15 million lines.

All of this being said, if it's in the google style guide, it's not an
unreasonable feature to add.

-- chris

ThosRTanner:
As long as it's not forced.

[...]

PR #508, opened 2017-12-14
Add --no_fwd_decls option

alex-xnor:
As requested in #354. Marks all uses as full uses. Useful for compliance with Google C++ style guide.


kimgr:
Cool, thanks! Can you also add a test case to prove this behavior exists? There's a list of per-testcase IWYU switches in `run_iwyu_tests.py` where you can add `--no_fwd_decls` for your new test to check that it short-circuits the include analysis.

alex-xnor:
Sure. I added a test.

kimgr:
Looks great to me. Could you squash the commits with a message mentioning the fixed issue as well? And then just force-push over the PR branch, and I can merge to master. Thank you!

cdglove:
How does this handle the case where a fwd decl is required? Sometimes it's not possible to #include due to circular dependencies so I assume that in those cases the build will break and someone will need to mark the decl as "IWYU keep"?

kimgr:
Oops, this should also target the master branch, not clang_4.0

alex-xnor:
@kimgr Done.

@cdglove I didn't consider that case, but experimentally it looks like it will suggest removing the required forward declaration (which would break compilation); adding `// IWYU pragma: keep` as you mention does work to prevent it from suggesting removal of the required forward declaration. I haven't thought about that situation very much so I'm not sure if this behavior is reasonable or not.

kimgr:
@cdglove I think that's a case of "Doctor, it hurts when I pass this flag" -- if you're modding IWYU's default behavior to never forward-declare, then it's hard for it to do the right thing.

But maybe it would be nice if it only forward-declared when it was absolutely  necessary. I don't really have a good idea for how to do that.

PR #511, opened 2017-12-23
iwyu_tool.py: add support for assumed paths

vmiklos:
When the implementation of foo.h is in bar.cpp and baz.cpp in a legacy
codebase, IWYU currently never analyzes foo.h. Add a new flag that
allows analyzing such headers explicitly by adding a switch that tells
what implementation file has a relevant compiler command to parse that
header.

Example usage:

iwyu_tool.py -v -p . -a source/bar.cpp include/foo.h

The "assume path" concept is inspired by clang-format.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -27,6 +27,7 @@ import argparse
 import subprocess
 import re
 import multiprocessing
+import copy
 
 
 def iwyu_formatter(output):
@@ -127,7 +128,8 @@ def run_iwyu(dbentry, iwyu_args, verbose):
     return get_output(cwd, command)
 
 
-def main(compilation_db_path, source_files, verbose, formatter, jobs, iwyu_args):
+def main(compilation_db_path, source_files, verbose, formatter, jobs,
+         assume_path, iwyu_args):
     """ Entry point. """
     # Canonicalize compilation database path
     if os.path.isdir(compilation_db_path):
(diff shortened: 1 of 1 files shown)


kimgr:
I'm very reluctant to add functionality to `iwyu_tool.py` -- it's really only intended as a shim for the compilation database format (something we would eventually want to build into `include-what-you-use`)

Also, couldn't you achieve this with `-Xiwyu --check_also=...`?

vmiklos:
Just using --check_also would be nice, but I didn't manage get that work. Perhaps I miss something.

E.g. I have 'sw/source/core/doc/doc.cxx' that is an entry in the compilation database, and I would like to get analysis for 'sw/inc/doc.hxx'. So I tried:

```
iwyu_tool.py -v -p . sw/source/core/doc/doc.cxx -- --check_also=sw/inc/doc.hxx
```

(Paths are from the LibreOffice source code, but should not matter, I guess.)

But that still seem to only give analysis for doc.cxx. (I built against system clang-3.8 from the clang_3.8 branch, perhaps that old branch has problematic --check-also support?)

Thanks.

kimgr:
As far as I can tell, that should work. `check_also` has been untouched since basically forever (as far as I know), so I think it should be OK in `clang_3.8`.

Can you post the actual include-what-you-use command echoed by `-v`? And maybe try tinkering with that directly and see if you can get better results? Thanks!

vmiklos:
Hrm, I've tried IWYU master against LLVM trunk, but the result is the same. Thanks for the hints, I'll try to debug this and see where it leads, or failing that with a self-contained reproducer. Closing this pull request in any case, clearly it's not the direction wanted.

Issue #499, opened 2017-10-23
In-tree build is not working (inconsistency in registered CommandLine or core dumped)


TheErk:
When building IWYU in tree with either clang 4.0.1 or clang 5.0, the compilation succeed but I do end up with an unusable executable:

```
$ include-what-you-use
: CommandLine Error: Option 'asan-instrument-assembly' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```

I seek the issue a little bit and found some LLVM bug reference (https://reviews.llvm.org/rL240104) speaking of double link to some LLVM static lib.
I did examine the link line for IWYU and discover a lot of duplicates.
In fact all LLVM libs where duplicated on the command line.

I commented out the 
```
 target_link_libraries(include-what-you-use
    LLVMX86AsmParser # MC, MCParser, Support, X86CodeGen, X86Desc, X86Info
...
    LLVMDemangle
    )
```

in IWYU CMakeLists.txt and I end up with an apparently working iwyu:
```
$ include-what-you-use
error: no input files
error: unable to handle compilation, expected exactly one compiler job in ''
$ include-what-you-use --version
include-what-you-use 0.9 (git:576e80f) based on clang version 5.0.0 (tags/RELEASE_500/final)

```
However on a "real" call using compilation database generated by CMake on my project
(which correctly builds with either clang 4.0.1 or clang 5.0.0) I get core dumped.

My LLVM+Clang+IWYU build uses CMake + Ninja with the following parameters:

MY_LIBS=clang
MY_VERSION=5.0
INSTALL_DIR=<pathto>/local/clang
cmake .. -G Ninja \
         -DCLANG_ENABLE_BOOTSTRAP=On \
         -DCMAKE_BUILD_TYPE=MinSizeRel \
         -DCMAKE_CXX_FLAGS="-Wl,-rpath=${INSTALL_DIR}/lib/" \
         -DLLVM_TARGETS_TO_BUILD="X86" \
         -DLLVM_ENABLE_CXX1Y=On \
         -DLLVM_LINK_LLVM_DYLIB=On \
         -DLIBOMP_TSAN_SUPPORT=On \
         -DCMAKE_INSTALL_PREFIX:PATH=${INSTALL_DIR}/${MY_LIBS}/${MY_VERSION}/ 

What should I do to get a working IWYU in-tree build?

kimgr:
[...]

PR #514, opened 2018-01-04
Experiment with a better CMakeLists.txt

kimgr:
Let's see if this builds.

Relies entirely on `find_package` to get at LLVM/Clang build configurations. This gets rid of all duplication of build details from LLVM and Clang.

There are some known issues with Clang packaging, e.g. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=862328, so this may prove insufficient for ALL environments, but it should be a good baseline.


kimgr:
Not looking so good, so far :-)

TheErk:
Did you see you have a version mismatch between llvm/clang in the previous travis log:
```
-- IWYU: found LLVM (version 6.0.0)
-- IWYU: found Clang (at /usr/local/clang-5.0.0)
```

TheErk:
As of today. This branch is fine for my usage as explained in issue #499

kimgr:
This will not be workable until the Debian packaging is fixed, closing.

PR #517, opened 2018-02-11
Add CMake flag IWYU_USE_LLVM_DYLIB

kimgr:
Adding this flag on the CMake line should force the IWYU build to try
and link with LLVM dynamically.

Also add a heuristic to auto-detect dynamic linking for in-tree
builds. IWYU_USE_LLVM_DYLIB will be set to ON if there's a CMake target
called 'LLVM' or if the LLVM_BUILD_LLVM_DYLIB flag is set in the LLVM
build system.

Fixes issue #499.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
 
   if( DEFINED IWYU_LLVM_INCLUDE_PATH AND DEFINED IWYU_LLVM_LIB_PATH )
     # User provided include/lib paths, fall through
-  elseif ( DEFINED IWYU_LLVM_ROOT_PATH )
+  elseif( DEFINED IWYU_LLVM_ROOT_PATH )
     # Synthesize include/lib relative to a root.
     set(IWYU_LLVM_INCLUDE_PATH ${IWYU_LLVM_ROOT_PATH}/include)
     set(IWYU_LLVM_LIB_PATH ${IWYU_LLVM_ROOT_PATH}/lib)
@@ -68,6 +68,13 @@ if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
   endif()
 else()
   message(STATUS "IWYU in-tree configuration")
+
+  if( NOT DEFINED IWYU_USE_LLVM_DYLIB )
+    if( TARGET LLVM OR LLVM_BUILD_LLVM_DYLIB )
+      # Link with LLVM dynamically if LLVM produces that target.
+      set(IWYU_USE_LLVM_DYLIB ON)
+    endif()
+  endif()
 endif()
 
 # Pick up Git revision so we can report it in version information.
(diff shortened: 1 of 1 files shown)


TheErk:
This branch does not work as-is in our context.
First IWYU_USE_LLVM_DYLIB should work both in in-tree or out-of-tree but
when using it in out-of-tree I have some remaining undefined Clang symbol.
I'll investigate that and send a patch/pull request based in this branch.

In out-of-tree config one has to provide IWYU_LLVM_ROOT_PATH so that 
  include_directories(${IWYU_LLVM_INCLUDE_PATH})
  link_directories(${IWYU_LLVM_LIB_PATH})
do their job.
It should work for both LLVM **and** Clang libs that are target_linked afterwards because they are given with their "short name" i.e.

"clangFrontend", "clangLex" or "LLVM".

This works very well when in-tree because they correspond to target known to CMake and it
works equally well when those target are "imported targets".

When they are bare lib (.a or .so) files (in ${IWYU_LLVM_LIB_PATH}) it should work but
you don't get transitive deps which may explain my undefined symbol issue.

I'll work on this next week and try to come up with a PR.

kimgr:
Thanks for trying it out! I actually did try this out-of-tree, and it worked for me :tm:

```
build$ cmake -G Ninja -DIWYU_LLVM_ROOT_PATH=/usr/lib/llvm-7 ../include-what-you-use/
build$ ninja
build$ rm -rf *
build$ cmake -G Ninja -DIWYU_LLVM_ROOT_PATH=/usr/lib/llvm-7 -DIWYU_USE_LLVM_DYLIB=ON ../include-what-you-use/
build$ ninja
build$ ldd include-what-you-use
	linux-vdso.so.1 =>  (0x00007ffc227fd000)
	libLLVM-7.so.1 => /usr/lib/llvm-7/lib/libLLVM-7.so.1 (0x00007f9a81301000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f9a810e2000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f9a80d5c000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f9a80b45000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9a80765000)
	libffi.so.6 => /usr/lib/x86_64-linux-gnu/libffi.so.6 (0x00007f9a8055d000)
	libedit.so.2 => /usr/lib/x86_64-linux-gnu/libedit.so.2 (0x00007f9a80326000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f9a80109000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f9a7ff05000)
	libtinfo.so.5 => /lib/x86_64-linux-gnu/libtinfo.so.5 (0x00007f9a7fcdc000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f9a7f986000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f9a866a2000)
```
(successful build output elided)

Can you tell me more about your platform? In my (Ubuntu) packaging, it looks like only LLVM is available as a dylib, all the clang libraries are still statically linked.

kimgr:
It could be that your packaging depends on some third-party library that we don't link (`terminfo` shows up at times)

TheErk:
Sorry for the long delay.
My platform is a linux platform (either Debian or CentOS, various version) however since we cannot rely on distribution specific version of software we do compile from source almost all (beside libc and a couple of others) our dependencies *including compilers* (gcc, clang) and many libraries. We compile clang with dylib enable and install it in custom prefix.
Concerning the reason of the failure I'll give another try next week, since I don't remember the reason of the failure. I'm sorry I didn't report it properly last time. I'll do it next time.

nilsdeppe:
This is necessary to have IWYU work on Arch Linux with Clang 6. Could this be merged soon?

kimgr:
@nilsdeppe IWYU 0.10 for Clang 6 is already released, so I would prefer not to touch that at this point.

I just learned today that the Debian packaging problems have been fixed, so I'll post a new PR soon (hopefully today) with a stripped down CMakeLists.txt that relies on the LLVM/Clang CMake modules installed by the packages.

kimgr:
@TheErk @nilsdeppe By the way, can someone tell me what the dylib target of Clang is called? Assuming it's a single one, like with LLVM?

nilsdeppe:
@kimgr sorry, I meant it in a forward looking sense. "Since this is required for Clang 6 on Arch it'll likely be true for Clang 7, so it would be great to have it for then." was what I meant :) I look forward to the PR you mention :)

The Clang shared object is called `libclang.so`, is that enough info?

kimgr:
@nilsdeppe Thanks for the clarification, hopefully I can have something for you to test later tonight :)

`libclang.so` is Clang's stable C wrapper, as opposed to `libLLVM.so`, which is LLVM's entire C++ API. It sounds from earlier in the discussion that there is a Clang build target to produce a similar dylib for Clang, but I'm not sure what it's called. I can probably dig it out, though.

kimgr:
FWIW, I can't find any way to build a Clang dylib target, so I'll just keep pretending there's no way to do it :). Happy to be convinced otherwise.

kimgr:
I got this working for Ubuntu and FreeBSD yesterday, should probably connect these two PRs: https://github.com/include-what-you-use/include-what-you-use/pull/542

kimgr:
This is superseded by #542, closing

Issue #441, opened 2017-05-28
include-what-you-use suggests removal of header which declares a namespace


pseyfert:
Hi,
running IWYU on [lhcb/LHCb](https://gitlab.cern.ch/lhcb/LHCb) I encountered a few wrong suggestions which I don't understand yet and I'd like to figure out what's wrong without just putting pragmas all over the code. I tried to put a stripped down version on github:

[one failing file example](https://github.com/pseyfert/iwyu-failure-on-lhcbmath)

(commands I ran for compilation are in the build directory, I used clang 3.9 and iwyu from the `clang_3.9` branch). It basically is a header file `Package/file.h` and an implementation file `src/file.cpp`. The header declares
```
namespace Math {
  unsigned long long choose(const unsigned short n, const unsigned short k);
}
```
and the implementation file has
```
#include "Package/file.h"
unsigned long long Math::choose(const unsigned short n, const unsigned short k) {…}
```
IWYU suggests to remove the inclusion of `Package/file.h` (which then leads to a compiler error since the namespace `Math` hasn't been declared yet). w/o using the namespace at all, the suggestion can be implemented.

That said, there has to be something else relevant I'm missing, because I haven't reproduced the wrong suggestion in a more minimal example yet.

Thanks in advance for suggestions and pointers.

vsapsai:
Thanks for the report, I was able to reproduce it with

```c++
// issue441-d1.h
namespace ns {
  void foo();
}

// issue441.cc
#include "issue441-d1.h"

void ns::foo() {}
```

It is important that .h and .cc files have different names. Probably that's why you had trouble creating a smaller repro case.

pseyfert:
Thanks for checking, however the stems of the filenames are the same (Choose.h, Choose.cpp), they are just in different directories and the include statement contains that:
```
../Kernel/LHCbMath/src/Choose.cpp should remove these lines:
- #include "LHCbMath/Choose.h"  // lines 14-14
```

and I'm not able to reproduce with the same-stem-different-directory setup:
```
// issue/issue441.h
namespace ns {
  void foo();
}

// issue441.cc
#include "issue/issue441.h"

void ns::foo() {}

//outputs:
//(issue/issue441.h has correct #includes/fwd-decls)
//
//(issue441.cc has correct #includes/fwd-decls)
//
```

kimgr:
The associated-header game has a few more rules than that, see:
[...]

Issue #179, opened 2015-06-15
iwyu removes differently named headers with declarations even despite  -Wmissing-declarations -Werror


vsapsai:
Originally reported on Google Code with ID 179

```
Issue is present with C code. In project I participate gcc options -Wmissing-declarations
-Werror are required, also declarations in header files can be split into public and
those available only inside the repository (private), with whose iwyu has a problem.

Problem can be reproduced on attached files. iwyu gives output:
$ iwyu StringTest.c -Wmissing-declarations -Werror
StringTest.c should add these lines:

StringTest.c should remove these lines:
- #include "StringTestPrivate.h"  // lines 1-1

The full include-list for StringTest.c:
#include <stdlib.h>                     // for calloc
#include <string.h>                     // for memcpy

---

Clearly applying such fix would lead to:
$ gcc StringTest.c -Wmissing-declarations -Werror
StringTest.c:6:46: error: no previous declaration for ‘memcpytest’ [-Werror=missing-declarations]
 __attribute__ ((visibility ("hidden"))) int *memcpytest(int **var, int **val) {
                                              ^
cc1: all warnings being treated as errors

With these flags I expect iwyu to include headers providing required declarations regardless
of where they are.

Tested on Ubuntu 14.04
$ iwyu --version
include-what-you-use 0.3 (590) based on clang version 3.5.0 (tags/RELEASE_350/final)
$ gcc --version
gcc (Ubuntu 4.8.2-19ubuntu1) 4.8.2

```

Reported by `mrzyglocki.marcin` on 2015-03-03 13:16:30

<hr>
- _Attachment: [StringTest.c](https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-179/comment-0/StringTest.c)_
- _Attachment: [StringTestPrivate.h](https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-179/comment-0/StringTestPrivate.h)_

bungeman:
A relevant llvm issue: https://llvm.org/bugs/show_bug.cgi?id=16286
The gcc documentation for 'missing-declarations': https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wmissing-declarations-558
See also 'missing-prototypes' in the gcc documentation.

[...]

Issue #491, opened 2017-10-11
iwyu wants to reomve header with prototype when not the same name as the c file


BradleySmall:
If we have a project such that the header file and the c file don't share the same name, and the prototype is located in the header file, it will suggest removing the header file. This leads to warnings about function defined without prototype in scope. Here is a quick example:
---interface.h---
#ifndef INTERFACE_H
#define INTERFACE_H 1
#ifdef __cplusplus
extern "C" {
#endif
int junk(void); 
#ifdef __cplusplus
}
#endif
#endif
--- end ----
---func.c---
#include "interface.h"
#include <stdio.h>

int junk (void) {
   printf("Hello, world!\n");
   return 0;
}
--- end func.c ----
--- output----
$ iwyu func.c
$
$ func.c should add these lines:
$
$ func.c should remove these lines:
$ - #include "interface.h"  // lines 1-1
$
$ The full include-list for func.c:
$ #include <stdio.h>                      // for printf
$ ---

jcageman:
Have seen the same issue, should be easy to reproduce.

http404error:
Experiencing this as well.

kimgr:
This is the same as #179, closing as duplicate (sorry I didn't notice earlier).

PR #519, opened 2018-03-01
Use flags to solve the definition-is-use problem

kimgr:
Here's a stab at fixing the problem where IWYU doesn't recognize the definition of free functions as uses of previous declarations of the functions.

This has been reported in #179, and with a slight wrinkle around namespaces in #441.

@bungeman, @EdSchouten, @pseyfert and @Tim020 -- could you take a look and see if this works in your respective contexts? Thanks!

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1400,7 +1400,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
             preprocessor_info().FileInfoFor(macro_def_file);
         file_info->ReportForwardDeclareUse(
             spelling_loc, fwd_decl,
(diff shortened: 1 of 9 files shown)


bungeman:
I just built everything at tip of tree with your patch, and it seems to do the same thing my original patch did but better. I haven't had a lot of time to evaluate, but this seems to fix my issue.

pseyfert:
This fixes the false exclusions I reported in #441.

bungeman:
I think this would also solve #491 .

kimgr:
Thanks for the feedback! I have a foggy idea for an even better design, but I think I'll stop here for now, in the interest of improving user experience.

Thank you @bungeman for the additional bug, I'll double-check that too and make sure it's closed.

PR #520, opened 2018-03-07
Reworked fix done in c182d6a to allow compiling with stable clang/llv…

Hinterwaeldlers:
The original commit disallow compiling iwyu using actual stable version 5.0.1. So better check, if the version fits. It might be necessary, to check further parts of the version, if the related SVN commit is applied to upcoming stable releases before 6.x.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -452,16 +452,26 @@ string PrintableTemplateName(const TemplateName& tpl_name) {
 string PrintableTemplateArgument(const TemplateArgument& arg) {
   std::string buffer;
   raw_string_ostream ostream(buffer);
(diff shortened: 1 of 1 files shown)


Hinterwaeldlers:
The compatibility branch is working well. As this behaviour is not obvious, a note related to this would be nice. Maybe:

### How to Build ###
... and fixed promptly.
Therefore the master branch should always work with the latest development version of Clang. For the usage of stable compiler version, ensure to make use of the related compatibility branches. Each of them use the same name as the compiler version (patch version number is omitted).

So think, the pull request can be closed accordingly

kimgr:
Thanks, good idea. I'll try to make this more prominent in the docs.

Issue #486, opened 2017-10-05
Suggests <algorithm> for std::move (correct is <utility>)


sesse:
Hi,

As the title says:

```c++
#include <utility>
#include <algorithm>

int foo(int x)
{
  int k = 3;
  return std::move(k);
}
```

Running this with IWYU removes the wrong header

```
atum17:~> /usr/bin/iwyu -std=c++14 -c test.cc       

test.cc should add these lines:

test.cc should remove these lines:
- #include <utility>  // lines 1-1

The full include-list for test.cc:
#include <algorithm>  // for move
---
```

There _is_ an std::move in <algorithm>, but it's a different overload (`OutputIt move( InputIt first, InputIt last, OutputIt d_first )`). This is probably what's tripping it up.

kimgr:
Can you run IWYU with a higher verbosity level? That might give some clues as to its reasoning. I can't find any direct mapping from `std::move` to `<algorithm>`, so maybe there's a deeper mistake somewhere.

sesse:
The reason why it wants \<algorithm\> seems pretty clear to me; there are two entirely different functions called std::move, and one of them (although the wrong one) is in \<algorithm\>.

Attaching a log with --verbose=10.

[iwyu-log.txt](https://github.com/include-what-you-use/include-what-you-use/files/1362181/iwyu-log.txt)

kimgr:
No, I think there's a mapping bug:

```
Marked full-info use of decl 0x5555575a7468 k (from test.cc:6:7) at test.cc:7:20
test.cc:7:10: (5) [ FunctionCall ] 0x5555575a7d00 template<> constexpr typename std::remove_reference<int &>::type &&move<int &>(int &&__t) noexcept
Marked full-info use of decl 0x55555720d3c0 std::move (from /usr/bin/../lib/gcc/x86_64-linux-gnu/7.2.0/../../../../include/c++/7.2.0/bits/move.h:98:5) at test.cc:7:10
...
Mapped /usr/bin/../lib/gcc/x86_64-linux-gnu/7.2.0/../../../../include/c++/7.2.0/bits/move.h to <algorithm> for std::move (#include already present)
```
iwyu_include_picker.cc has this:

    // This didn't come from the grep, but seems to be where swap()
    // is defined?
    { "<bits/move.h>", kPrivate, "<algorithm>", kPublic },   // for swap<>()

Maybe we should revisit this mapping; it's from the GCC 4.4-4.6 timeframe.

[...]

PR #522, opened 2018-03-08
<bits/move.h> is a part of <utility>.

bungeman:
`<bits/move.h>` provides std::forward, std::move, and std::swap. These are defined to come from `<utility>` and not `<algorithm>`.

In c++11 and later std::swap comes from `<utility>`, previously it came from `<algorithm>`. Fortunately, `<bits/move.h>` is a c++11 only header, so if any symbols are being seen in it, that's where they should come from.

This was reported in #486.

### DIFF
--- a/gcc.stl.headers.imp
+++ b/gcc.stl.headers.imp
@@ -114,9 +114,7 @@
   { include: ["<tr1_impl/unordered_set>", private, "<unordered_set>", public ] },
   { include: ["<tr1_impl/utility>", private, "<tr1/utility>", public ] },
   { include: ["<tr1_impl/utility>", private, "<utility>", public ] },
-  # This didn't come from the grep, but seems to be where swap()
-  # is defined?
-  { include: ["<bits/move.h>", private, "<algorithm>", public ] },   # for swap<>()
+  { include: ["<bits/move.h>", private, "<utility>", public ] },
   # Hash and hashtable-based containers.
   { include: ["<tr1_impl/functional_hash.h>", private, "<tr1/functional>", public ] },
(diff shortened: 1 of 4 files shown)


kimgr:
Thanks! Any chance for a testcase?

Also, we have freestanding mappings in gcc.*.imp, could you make sure they're updated similarly?

bungeman:
Updated the *.imp, add the test from #486. I'm doing this between other things, so I just did all this in the github webui, resulting in all these one file commits. I think you can just "Squash and merge" and clean up the commit message. If not, let me know and I'll see if there's a good way to squash all this.

kimgr:
Neat, thanks! I'll take another look, test on my freebsd system and clean it up if I find anything.

bungeman:
I made it back to a real machine so I squashed this merge request myself. Makes it easier to look at.

kimgr:
Thank you! I just gave this a whirl on FreeBSD. Of course the test fails :-/. The system C++ library on BSD is libc++ and all the builtin mappings are for libstdc++, so we don't have mappings and `<type_traits>` is suggested for `std::move`.

I'm not sure how I feel about this. On the one hand test coverage is nice, but on the other hand non-portable tests are a pain. We don't have good test infrastructure to detect these things either.

We actively try to keep new test cases independent of the environment by creating stub headers for constructs in the standard library, when we run into issues. But the mappings are naturally environment-specific.

Sigh. I'll rework your commit and remove the test (sorry about the wasted effort).

kimgr:
Did that and merged in 314fc2d182a99386d073ebc44f7924a7d99cba43. Thanks for the patch!

PR #524, opened 2018-03-14
use a ; separated list like CMake requires it in documentation

fiesh:
Fix documentation of CMake integration


kimgr:
Looks good to me!

I'm a bit torn about the args -- only arguments intended for include-what-you-use need the `-Xiwyu` prefix, but sometime we also need to pass special Clang arguments to IWYU, undecorated.

I think this is a strict improvement, so let's do this for now, and update again if someone complains.

Thanks for your contribution!

PR #525, opened 2018-03-15
fix dangling reference and unused argument

fiesh:
Minor fix


kimgr:
Thank you! Note we follow LLVM's lead and avoid `auto` unless the type is visible elsewhere on the line. So please spell out `std::set<string>` instead of `auto`.

fiesh:
I updated and rebase!

kimgr:
Looks good to me, thanks for the patch!

PR #527, opened 2018-03-18
Remove Google VCS support from fix_includes.py

kimgr:
The script had rudimentary support for checking out files from Google's
version control system, which complicated testing and implementation
somewhat.

Now that this is not primarily a Google tool and most version control
systems do not use the read-only flag for preventing concurrent edits,
remove this code to simplify.

No functional change intended, unless you were using this support.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -19,11 +19,6 @@ mentioned in the output, removing their old #include lines and
 replacing them with the lines given by the include_what_you_use
 script.
 
-We only edit files that are writeable (presumably open for p4 edit),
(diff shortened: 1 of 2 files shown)


kimgr:
cc @bungeman -- You may know someone who is affected by this. Let me know if it's a deal-breaker for Googlers.

bungeman:
According to my sources, this should not affect any use at Google. In general, if something seems like it would only make sense with a Google specific project setup, feel free to change it.

kimgr:
Thanks, just what I needed to know!

PR #528, opened 2018-03-18
Better handling of encoding in fix_includes.py file I/O

kimgr:
Replaces PR #489.

Still a little hacky, but this appears to get the job done.

The first commit is already covered by PR #527, so only the last four apply to this problem.

@dakotahawkins, can you try this and see if it works in your context? Also, a thorough code review would be welcome from anyone.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -501,6 +501,35 @@ class LineInfo(object):
             % (line, type_id, self.key, self.move_span, self.reorder_span))
 
 
+class FileInfo(object):
+  """ Details about a file's storage encoding  """
+  DEFAULT_LINESEP = os.linesep
+
+  def __init__(self, linesep):
+    self.linesep = linesep
+
+  @staticmethod
+  def parse(filename):
+    """ Return a FileInfo object describing file encoding details. """
+    with open(filename, 'rb') as f:
+      content = f.read()
+
+    linesep = FileInfo.guess_linesep(content)
+    return FileInfo(linesep)
+
+  @staticmethod
+  def guess_linesep(bytebuf):
+    """ Return most frequent line separator of buffer. """
+    win = bytebuf.count(b'\r\n')
+    unix = bytebuf.count(b'\n') - win
+    if win > unix:
+      return '\r\n'
+    elif unix > win:
+      return '\n'
+
+    return FileInfo.DEFAULT_LINESEP
+
+
 def _ReadFile(filename):
   """Read from filename and return a list of file lines."""
   try:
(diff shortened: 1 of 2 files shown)


dakotahawkins:
@kimgr I'll try to take a look this week! Thanks for pinging me about it :)

kimgr:
Rebased on top of new master with VCS checkout removed and some other simplifying refactors applied.

fix_includes.py
```diff
@@ -501,40 +501,75 @@ def __str__(self):
...
+  def parse(filename):
+    with open(filename, 'rb') as f:
+      content = f.read()
+
+    linesep = FileInfo.guess_linesep(content)
+    encoding = FileInfo.guess_encoding(content)
+    return FileInfo(linesep, encoding)
+
+  @staticmethod
+  def guess_linesep(bytebuf):
```
dakotahawkins: I wonder if something like this would be faster than `count`-ing through the whole file twice:

```python
win = 0
unix = 0
i = 0
while True:
  i = value.find(b'\n', i)
  if i == -1:
    break
  if i > 0 and value[i - 1] == b'\r':
    win += 1
  else:
    unix += 1
  i += 1
```

Think it's worth trying on a very large file?

dakotahawkins: That way you could also break early: If one is 0 and the other is some threshold X (or if one is 2x the other or something), you're hitting consistent line endings so you might as well assume the trend will continue.

It would also be awesome to be able to specify this as a preference and skip the check entirely. We work on windows but use unix line-endings (windows line-endings would be regarded as an error that snuck in to git somehow).

kimgr: I figured I might as well measure, so I hacked up this:
https://gist.github.com/kimgr/53fba53d71c1aa83a7f9ea0006089bc3

I haven't used `timeit` before, so hopefully I got it right. I ran this on GCC7's basic_string.h, 240282 bytes:
/tmp$ python linesep_speed.py basic_string.h 
kim: 3.14161586761 us
dakota: 14.5403268337 us

So unless I've completely botched something, the naive implementation is faster (yay, less loops!)

kimgr: I'd rather not add any more options. The whole point of this detection is that IWYU will follow the majority in the file -- if you have a file that's accidentally Windows line endings, IWYU will build on that mistake, but I don't see how that's worse than mixing in something else. If you want to force a particular linesep, that sounds like a job for another tool (`unix2dos` and `dos2unix` from gnuwin32 are great for this).

kimgr:
Any more input on this?

kimgr:
Added unit test coverage and some doc comments. I think we're done here. Will merge assuming the build passes.

Issue #526, opened 2018-03-15
add_only_with_remove mode


fiesh:
I've been trying to implement the following mode, but to no real avail.  So I was hoping to discuss it here and get some feedback on if and how to implement this:

I'd like to have a mode `add_only_with_remove`, triggered by `--add_only_with_remove`.  What it should do is only add includes whose adding allows for another removal.  This is partly motivated by the fact that removal of includes is, for many people, more important than adding them (hey it compiles!), and partly by #121 which, for us, seems to make it impossible to use exports correctly.  (Even with #121 fixed I'd like this kind of mode.)

Consider the following include tree of translation unit `A`.
```
   B* -- C*
  /
A -- F
  \
   D --- E*
```
The `*` symbolizes that the include satisfies a Use.

In its current mode, IWYU would like to change that to `A`'s includes to `B`, `C`, `E`.
However, in `add_only_with_remove` mode, I'd like the includes to be changed to `B`, `E`.
`C`, already hereditarily present through `B`, should not be added explicitly.

I understand that IWYU aims to perfect includes, and this takes away one step.  However, the world being imperfect, I think this would make IWYU useable for us.

kimgr:
Nice ASCII graph, that might actually win me over :)

The switch name isn't great. How about `--allow_transitive`? I don't have a clear idea of how to do this, but if you look in `iwyu_output.cc`, `ProcessFullUse`, I wonder if you could detect transitive "double-uses", and cross-check them against each other?

If you can come up with a neat patch that doesn't add lots of complexity, I'd be open to taking it, as this will probably produce better results in a lot of cases.

fiesh:
If there is a working patch, I'll make sure to include the ASCII artwork in the source code's documentation :)

I agree the name is not ideal.  I'm not too happy about `--allow-transitive` though either as just like the original name, it doesn't really convey the idea.  How about something like `--purge_subincludes` or `--purge_transitive_includes`, since we purge things that are hereditarily included?

My idea was the following.  After every `Use` was assigned a header, we go through all `Use`s and and remove those whose header is transitively included by another `Use`.

Does that sound like a reasonable approach?

fiesh:
After playing around with a test implementation for a bit, I think the feature would have to be changed like this:

Let's call it `--tolerate-transitive`.  If, like above, `B` is needed, then both having `C` present as an explicit include or not having it present are acceptable.  IWYU should neither recommend adding it nor removing it.

[...]

Issue #424, opened 2017-04-05
iwyu asserts on include cycle


zjturner:
Shouldn't this just output a diagnostic like everything else?  I mean it's a legitimate coding error that someone can make which would not cause a compile error (which is indeed how I found this assertion, I had a header file which inadvertently included itself).

```
// Foo.h
#ifndef FOO_H
#define FOO_H
#include "Foo.h"  // iwyu asserts, but any compiler will happily accept this.
#endif
```

kimgr:
Boost has this a lot too, apparently, see #406.

I haven't dug into the details yet, but I can imagine the cycle will throw IWYU's analysis for a fit, since it sometimes needs to follow the include graph.

zjturner:
Even if it follows the include graph, the preprocessor should magically
handle it the second time around, just like in the real program (ie when
you do the second include, the include guard is defined, so the whole file
evaluates to nothing)

vsapsai:
@zjturner, can you reproduce the problem with exact code you've provided? Because we try to handle at least simple cycles (see [tests/cxx/include_cycle.cc](https://github.com/include-what-you-use/include-what-you-use/blob/master/tests/cxx/include_cycle.cc)). I am wondering if there is something else that triggers this problem.

kimgr:
I can't reproduce this. Tried with a `Foo.cc` file including your `Foo.h` above, a `Foo_main.cc` including `Foo.h` to test when they're not associated, and running IWYU on `Foo.h` directly. No assert.

I did see some annoying absolute path suggestions, though, but it feels like you're seeing something we're not. I can see if I can get my environment up and running on Windows, and try there, in case this is something Windows-specific.

zjturner:
I encountered this when building llvm / lldb, so you might try to just run
against llvm on whatever platform and see if it repros. If you use that
magic cmake define it makes it really easy

banjo74:
Our organization uses -isystem for third party software (including boost).  We do this to also ignore compiler warnings that arise from such headers.  Maybe iwyu could ignore such issues that are included via -isystem.

dpelle:
I encountered the assert failure too. In case it helps, here is what I see:

{noformat}
Cycle in include-mapping:
  <boost/phoenix/support/detail/iterate.hpp> ->
  <boost/phoenix/operator/detail/cpp03/mem_fun_ptr_eval_result_of.hpp> ->
[...]

Issue #406, opened 2017-01-30
iwyu chokes on boost/move/algorithm.hpp


Bu11etmagnet:
`#include <boost/chrono.hpp>
`

results in

```
Cycle in include-mapping:
  <boost/move/algorithm.hpp> ->
  <boost/move/algorithm.hpp>
llvm/tools/clang/tools/include-what-you-use/iwyu_include_picker.cc:844: Assertion failed: Cycle in include-mapping
```

Ubuntu's Boost 1.54 boost/move/algorithm.h does indeed include itself:

```
#ifndef BOOST_MOVE_ALGORITHM_HPP
#define BOOST_MOVE_ALGORITHM_HPP

#include <boost/move/detail/config_begin.hpp>

#include <boost/move/utility.hpp>
#include <boost/move/iterator.hpp>
#include <boost/move/algorithm.hpp>
#include <boost/detail/no_exceptions_support.hpp>

#include <algorithm> //copy, copy_backward
#include <memory>    //uninitialized_copy
```

I don't understand the logic of exempting "third_party" and "internal".

Maybe a one-length cycle could also receive a free pass.

kimgr:
Sigh. Sometimes Boost is wonderful, and sometimes I wonder what's going on.

> I don't understand the logic of exempting "third_party" and "internal"

IWYU started as a Google-internal project, and there are still some Googleisms hard-coded in it. It looks like Google puts all their third-party code in a root called `third_party`, so that's a convenient way of excluding code they have less control over. Outside of Google, this is less useful :-)

Thanks for the bug report, we should probably handle this more gracefully. Not sure how.

Bu11etmagnet:
Tried it with a file including `boost/optional.hpp` and got

```
Cycle in include-mapping:
  <boost/utility/in_place_factory.hpp> ->
  <boost/preprocessor/iteration/detail/iter/forward1.hpp> ->
  <boost/preprocessor/iteration/detail/iter/forward1.hpp>
llvm/tools/clang/tools/include-what-you-use/iwyu_include_picker.cc:844: Assertion failed: Cycle in include-mapping
```

This is even worse than `algorithm.h` because `algorithm.h` including itself is likely a mistake, but the cycle in `boost/preprocessor` might be intentional :(

apollo13:
What can be done to work around this? Boost seems to be pretty annoying :D

Bu11etmagnet:
I've done this privately:

```
[...]

PR #529, opened 2018-03-20
add --tolerate_transitive mode

fiesh:
This adds `--tolerate_transitive` mode as discussed in #526.

It seems that due to other issues, we still can't really use IWYU though :( (#406, #424 at least).

I think the mode is quite reasonable though and should most often improve users' experience.

### DIFF
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -93,6 +93,9 @@ static void PrintHelp(const char* extra_msg) {
          "        file names (default: 80).\n"
          "   --no_comments: do not add 'why' comments.\n"
          "   --no_fwd_decls: do not use forward declarations.\n"
+         "   --tolerate_transitive: If a file uses symbols from both a.h and b.h\n"
+         "        and includes a.h, tolerate b.h both when it is included and \n"
+	 "        when\n it is not.\n"
          "   --verbose=<level>: the higher the level, the more output.\n"
          "\n"
          "In addition to IWYU-specific options you can specify the following\n"
@@ -162,7 +165,8 @@ CommandlineFlags::CommandlineFlags()
       prefix_header_include_policy(CommandlineFlags::kAdd),
       pch_in_code(false),
       no_comments(false),
-      no_fwd_decls(false) {
+      no_fwd_decls(false),
+      tolerate_transitive(false) {
 }
 
 int CommandlineFlags::ParseArgv(int argc, char** argv) {
@@ -179,8 +183,10 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
     {"max_line_length", optional_argument, nullptr, 'l'},
     {"no_comments", optional_argument, nullptr, 'o'},
     {"no_fwd_decls", optional_argument, nullptr, 'f'},
+    {"tolerate_transitive", optional_argument, nullptr, 'r'},
     {nullptr, 0, nullptr, 0}
   };
+  // TODO: This doesn't cover all options!
   static const char shortopts[] = "d::p:v:c:m:n";
   while (true) {
     switch (getopt_long(argc, argv, shortopts, longopts, nullptr)) {
@@ -193,6 +199,7 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
       case 'n': no_default_mappings = true; break;
       case 'o': no_comments = true; break;
       case 'f': no_fwd_decls = true; break;
+      case 'r': tolerate_transitive = true; break;
       case 'x':
         if (strcmp(optarg, "add") == 0) {
           prefix_header_include_policy = CommandlineFlags::kAdd;
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -95,7 +95,7 @@ static void PrintHelp(const char* extra_msg) {
          "   --no_fwd_decls: do not use forward declarations.\n"
          "   --tolerate_transitive: If a file uses symbols from both a.h and b.h\n"
          "        and includes a.h, tolerate b.h both when it is included and \n"
-	 "        when\n it is not.\n"
+         "        when it is not.\n"
          "   --verbose=<level>: the higher the level, the more output.\n"
          "\n"
          "In addition to IWYU-specific options you can specify the following\n"
@@ -186,7 +186,6 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
     {"tolerate_transitive", optional_argument, nullptr, 'r'},
     {nullptr, 0, nullptr, 0}
   };
-  // TODO: This doesn't cover all options!
   static const char shortopts[] = "d::p:v:c:m:n";
   while (true) {
     switch (getopt_long(argc, argv, shortopts, longopts, nullptr)) {
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -468,8 +468,8 @@ bool OneIncludeOrForwardDeclareLine::HasSymbolUse(const string& symbol_name)
   return ContainsKey(symbol_counts_, symbol_name);
 }
 
-void OneIncludeOrForwardDeclareLine::AddSymbolUse(const string& symbol_name, int multiplicity) {
-  symbol_counts_[symbol_name] += multiplicity;
+void OneIncludeOrForwardDeclareLine::AddSymbolUses(const string& symbol_name, int count) {
+  symbol_counts_[symbol_name] += count;
 }
 
 bool OneIncludeOrForwardDeclareLine::IsIncludeLine() const {
@@ -1430,47 +1430,59 @@ void CalculateIwyuForFullUse(OneUse* use,
 // but not present but transitively included by some
 // OneIncludedOrForwardDeclareLine b that is desired and present
 // and adds a's symbol counts to b.
-vector<OneIncludeOrForwardDeclareLine> TolerateTransitively(vector<OneIncludeOrForwardDeclareLine> const & oiofdls,
+vector<OneIncludeOrForwardDeclareLine> PruneTransitivelyPresent(
+    vector<OneIncludeOrForwardDeclareLine> const& lines,
     const IwyuPreprocessorInfo* preprocessor_info) {
-  // In the first step, we copy all oiofdl's that
+  // In the first step, we copy all lines that
   // * are either no include lines,
   // * or that are an include file and present,
   // * or that are an include file, not present and not desired,
   // * or that are an include file, not present, desired, and whose include_file
-  //   isn't transitively included by another oiofdl which is an include file
+  //   isn't transitively included by another line which is an include file
   // into one range and the remaining ones into another range.
-  const auto pred = [&oiofdls, preprocessor_info] (OneIncludeOrForwardDeclareLine const & target) {
-    if(!target.IsIncludeLine() || target.is_present() || !target.is_desired()) {
+  const auto pred = [&lines, preprocessor_info](
+                        OneIncludeOrForwardDeclareLine const& target) {
+    if (!target.IsIncludeLine() || target.is_present() ||
+        !target.is_desired()) {
       return true;
     }
     // Check if target is transitively included by some source.
     // This must be proper, equality does not suffice.
-    for(const OneIncludeOrForwardDeclareLine & source : oiofdls) {
-      if(!source.IsIncludeLine()) {
+    for (const OneIncludeOrForwardDeclareLine& source : lines) {
+      if (!source.IsIncludeLine()) {
         continue;
       }
-      if(source.included_file() == target.included_file()) {
+      if (source.included_file() == target.included_file()) {
         continue;
       }
 
-      if(preprocessor_info->FileTransitivelyIncludes(source.included_file(), target.included_file())) {
+      if (preprocessor_info->FileTransitivelyIncludes(source.included_file(),
+                                                      target.included_file())) {
         return false;
       }
     }
     return true;
   };
   vector<OneIncludeOrForwardDeclareLine> first, second;
-  first.reserve(oiofdls.size());
-  second.reserve(oiofdls.size());
-  std::partition_copy(oiofdls.cbegin(), oiofdls.cend(), std::back_inserter(first), std::back_inserter(second), pred);
-  // Now add the symbols from range second to the corresponding elements of range first.
-  for(const OneIncludeOrForwardDeclareLine & target : second) {
-    const auto iter = std::find_if(first.begin(), first.end(), [&target, preprocessor_info] (const OneIncludeOrForwardDeclareLine & source) {
-        return preprocessor_info->FileTransitivelyIncludes(source.included_file(), target.included_file());
-        });
-    CHECK_(iter != first.end() && "Internal error, cannot find the source for a transitive tolerate target");
-    for(const auto & p : target.symbol_counts()) {
-      iter->AddSymbolUse(p.first, p.second);
+  first.reserve(lines.size());
+  second.reserve(lines.size());
+  std::partition_copy(lines.cbegin(), lines.cend(), std::back_inserter(first),
+                      std::back_inserter(second), pred);
+  // Now add the symbols from range second to the corresponding elements of
+  // range first.
+  for (const OneIncludeOrForwardDeclareLine& target : second) {
+    const auto iter =
+        std::find_if(first.begin(), first.end(),
+                     [&target, preprocessor_info](
+                         const OneIncludeOrForwardDeclareLine& source) {
+                       return preprocessor_info->FileTransitivelyIncludes(
+                           source.included_file(), target.included_file());
+                     });
+    CHECK_(iter != first.end() &&
+           "Internal error, cannot find the source for a transitive tolerate "
+           "target");
+    for (const auto& p : target.symbol_counts()) {
+      iter->AddSymbolUses(p.first, p.second);
     }
     // iter is desired now because target was.
     CHECK_(target.is_desired() && "Internal error, target must be desired");
@@ -2061,7 +2073,7 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
   }
 
   if (GlobalFlags().tolerate_transitive) {
-	  lines_ = internal::TolerateTransitively(lines_, preprocessor_info_);
+	  lines_ = internal::PruneTransitivelyPresent(lines_, preprocessor_info_);
   }
 
   internal::CleanupPrefixHeaderIncludes(preprocessor_info_, &lines_);
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1426,33 +1426,31 @@ void CalculateIwyuForFullUse(OneUse* use,
   }
 }
 
-// This function removes all OneIncludeOrForwardDeclareLine's a that are desired
+// This function removes all OneIncludeOrForwardDeclareLines a that are desired
 // but not present but transitively included by some
-// OneIncludedOrForwardDeclareLine b that is desired and present
-// and adds a's symbol counts to b.
+// OneIncludedOrForwardDeclareLine b that is desired and present and adds a's
+// symbol counts to b.
 vector<OneIncludeOrForwardDeclareLine> PruneTransitivelyPresent(
-    vector<OneIncludeOrForwardDeclareLine> const& lines,
+    const vector<OneIncludeOrForwardDeclareLine> & lines,
     const IwyuPreprocessorInfo* preprocessor_info) {
-  // In the first step, we copy all lines that
-  // * are either no include lines,
-  // * or that are an include file and present,
-  // * or that are an include file, not present and not desired,
-  // * or that are an include file, not present, desired, and whose include_file
-  //   isn't transitively included by another line which is an include file
-  // into one range and the remaining ones into another range.
-  const auto pred = [&lines, preprocessor_info](
-                        OneIncludeOrForwardDeclareLine const& target) {
+  // In the first step, we copy all lines that are an include line, not present,
+  // desired, and whose include_file is transitively included by another line
+  // which is a desired include line into a range second and the remaining ones
+  // into another range first.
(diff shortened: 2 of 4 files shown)


kimgr:
Thanks! There seems to be some build error -- see Travis logs. I'll try and look into this tonight.

fiesh:
Ah sorry, in all rebasing madness I made some stupid mistake.  I fixed and... rebased ;)

iwyu_globals.cc
```diff
@@ -179,8 +183,10 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
     {"max_line_length", optional_argument, nullptr, 'l'},
     {"no_comments", optional_argument, nullptr, 'o'},
     {"no_fwd_decls", optional_argument, nullptr, 'f'},
+    {"tolerate_transitive", optional_argument, nullptr, 'r'},
     {nullptr, 0, nullptr, 0}
   };
+  // TODO: This doesn't cover all options!
```
kimgr: I think that's intentional, not all options are exposed as short-opts.

iwyu_globals.cc
```diff
@@ -93,6 +93,9 @@ static void PrintHelp(const char* extra_msg) {
          "        file names (default: 80).\n"
          "   --no_comments: do not add 'why' comments.\n"
          "   --no_fwd_decls: do not use forward declarations.\n"
+         "   --tolerate_transitive: If a file uses symbols from both a.h and b.h\n"
+         "        and includes a.h, tolerate b.h both when it is included and \n"
+	 "        when\n it is not.\n"
```
kimgr: Misaligned, needs another leading space. And spurious \n after "when"

iwyu_output.cc
```diff
@@ -468,8 +468,8 @@ bool OneIncludeOrForwardDeclareLine::HasSymbolUse(const string& symbol_name)
   return ContainsKey(symbol_counts_, symbol_name);
 }
 
-void OneIncludeOrForwardDeclareLine::AddSymbolUse(const string& symbol_name) {
-  ++symbol_counts_[symbol_name];
+void OneIncludeOrForwardDeclareLine::AddSymbolUse(const string& symbol_name, int multiplicity) {
```
kimgr: Can you remove this method to `AddSymbolsUses` and call `multiplicity` `count` instead?

iwyu_output.cc
```diff
@@ -1426,6 +1426,59 @@ void CalculateIwyuForFullUse(OneUse* use,
   }
 }
 
+// This function removes all OneIncludeOrForwardDeclareLine's a that are desired
+// but not present but transitively included by some
+// OneIncludedOrForwardDeclareLine b that is desired and present
+// and adds a's symbol counts to b.
+vector<OneIncludeOrForwardDeclareLine> TolerateTransitively(vector<OneIncludeOrForwardDeclareLine> const & oiofdls,
```
kimgr: Can you call variables of type `OneIncludeOrForwardDeclareLine` `line[s]` instead? I like the `oiofdls` acronym, but it's hard to read and spell correctly :)

iwyu_output.cc
```diff
@@ -1426,6 +1426,59 @@ void CalculateIwyuForFullUse(OneUse* use,
   }
 }
 
+// This function removes all OneIncludeOrForwardDeclareLine's a that are desired
```
kimgr: Please reflow comment to fill 80 chars

iwyu_output.cc
kimgr: `TolerateTransitively` does not describe what this method does, more a general attitude :)

I wonder what a good name would be... `RemoveTransitivelyProvided` or something? Shorter, preferably.

iwyu_output.h
```diff
@@ -161,7 +161,7 @@ class OneIncludeOrForwardDeclareLine {
   void clear_desired() { is_desired_ = false; }
   void clear_line_numbers() { start_linenum_ = end_linenum_ = -1; }
   // Another symbol we're using that's defined in this file.
-  void AddSymbolUse(const string& symbol_name);
+  void AddSymbolUse(const string& symbol_name, int multiplicity = 1);
```
kimgr: I think I'd prefer to remove the default arg and update callers to pass 1 where applicable.

iwyu_globals.cc
fiesh: I'll remove it then

iwyu_globals.cc
fiesh: Oops, nvim put a tab in there, and I didn't see that because it aligns with tab width 8.

fiesh:
I incorporated your suggestions hopefully to your liking.  The only thing missing is the clang-format, which I'd gladly do because I'm actually not used to formatting code by hand any more.  However, I don't see and clang-format style file in the project.  Is there something?  The defaults don't seem to match.

kimgr:
There's a .clang-format file in the project root. Maybe it's hidden because of the leading dot?

WSoptics:
Ah ok heh, I had actually used that and got confused by `clang-format` still changing tons of lines.  So what should I do, only reformat my lines?  Or alter the whole file?

It would seem that everything should be uniformly formatted?  Then again, that shouldn't be in my commit I guess.

fiesh:
Ah ok heh, I had actually used that and got confused by clang-format still changing tons of lines. So what should I do, only reformat my lines? Or alter the whole file?

It would seem that everything should be uniformly formatted? Then again, that shouldn't be in my commit I guess.

kimgr:
Yes, it would be nice if you could constrain it to your changed lines. We'll have to take a closer look at global reformatting or not later.

fiesh:
OK, please look at the updated pull request.  Would rebase once everything is fine!

iwyu_output.cc
```diff
@@ -1426,6 +1426,71 @@ void CalculateIwyuForFullUse(OneUse* use,
...
+        std::find_if(first.begin(), first.end(),
+                     [&target, preprocessor_info](
+                         const OneIncludeOrForwardDeclareLine& source) {
+                       return preprocessor_info->FileTransitivelyIncludes(
+                           source.included_file(), target.included_file());
+                     });
+    CHECK_(iter != first.end() &&
+           "Internal error, cannot find the source for a transitive tolerate "
+           "target");
+    for (const auto& p : target.symbol_counts()) {
```
kimgr: Better name than `p` -- `symbol_count`, maybe?

iwyu_output.cc
```diff
@@ -1426,6 +1426,71 @@ void CalculateIwyuForFullUse(OneUse* use,
...
 }
 
+// This function removes all OneIncludeOrForwardDeclareLine's a that are desired
+// but not present but transitively included by some
+// OneIncludedOrForwardDeclareLine b that is desired and present
+// and adds a's symbol counts to b.
+vector<OneIncludeOrForwardDeclareLine> PruneTransitivelyPresent(
+    vector<OneIncludeOrForwardDeclareLine> const& lines,
+    const IwyuPreprocessorInfo* preprocessor_info) {
+  // In the first step, we copy all lines that
```
kimgr: Everything you describe here is pretty obvious from the code itself. Could you say something more about the intent and motivation of the algorithm instead? There's a lot of different states influencing the outcome here, and I'm having trouble keeping it all in my head without a higher purpose.

iwyu_output.cc
```diff
@@ -1426,6 +1426,71 @@ void CalculateIwyuForFullUse(OneUse* use,
...
+  const auto pred = [&lines, preprocessor_info](
+                        OneIncludeOrForwardDeclareLine const& target) {
+    if (!target.IsIncludeLine() || target.is_present() ||
+        !target.is_desired()) {
+      return true;
+    }
+    // Check if target is transitively included by some source.
+    // This must be proper, equality does not suffice.
+    for (const OneIncludeOrForwardDeclareLine& source : lines) {
+      if (!source.IsIncludeLine()) {
```
kimgr: I think you can combine these two if-statements into a single `||`-separated one.

iwyu_output.cc
```diff
@@ -1426,6 +1426,71 @@ void CalculateIwyuForFullUse(OneUse* use,
...
+      }
+
+      if (preprocessor_info->FileTransitivelyIncludes(source.included_file(),
+                                                      target.included_file())) {
+        return false;
+      }
+    }
+    return true;
+  };
+  vector<OneIncludeOrForwardDeclareLine> first, second;
```
kimgr: Blank line after the lambda, please

iwyu_output.cc
```diff
@@ -1426,6 +1426,71 @@ void CalculateIwyuForFullUse(OneUse* use,
...
+    vector<OneIncludeOrForwardDeclareLine> const& lines,
+    const IwyuPreprocessorInfo* preprocessor_info) {
+  // In the first step, we copy all lines that
+  // * are either no include lines,
+  // * or that are an include file and present,
+  // * or that are an include file, not present and not desired,
+  // * or that are an include file, not present, desired, and whose include_file
+  //   isn't transitively included by another line which is an include file
+  // into one range and the remaining ones into another range.
+  const auto pred = [&lines, preprocessor_info](
```
kimgr: Can you name `pred` after its purpose. It looks something like `is_transitively_present`, maybe...?

iwyu_output.cc
```diff
@@ -1426,6 +1426,71 @@ void CalculateIwyuForFullUse(OneUse* use,
...
+      }
+    }
+    return true;
+  };
+  vector<OneIncludeOrForwardDeclareLine> first, second;
+  first.reserve(lines.size());
+  second.reserve(lines.size());
+  std::partition_copy(lines.cbegin(), lines.cend(), std::back_inserter(first),
+                      std::back_inserter(second), pred);
+  // Now add the symbols from range second to the corresponding elements of
```
kimgr: This could also do with some more detail. E.g. why are we doing this?

iwyu_output.cc
kimgr: Typo: a that

No need for the apostrophe on OneIncludeOrForwardDeclareLines

iwyu_output.cc
fiesh: I went for `p` since it's a pair because we're iterating over a map.  Other names always have the problem that they somewhat obfuscate we're dealing with a key-value-pair.  But I can change it if you think it makes it more readable.

fiesh:
Thank you for your review!  I'll try to incorporate your suggestions and commit a new release.

I don't know why, but I cannot comment on your last comment, maybe a github bug.  So for the apostrophe between `OneIncludeOrForwardDeclareLine` and `s`, I like putting it there because it makes clear what the symbol name is and that's it's merely a plural form. I can change it if you like that, but I think it makes it worse.  I can also use markdown-like syntax and put the `OneIncludeOrForwardDeclareLine` into \`\`s.

kimgr:
Huh, I don't see a Reply field for the last comment either. Weird.

I've only seen apostrophes in English to indicate either the possessive form ("my cat's fur is silky") or to indicate word contraction ("don't ask"). We typically don't use backticks in comments in IWYU, single quotes are more common, but that won't look very nice here. I think you should just drop the apostrophe.

iwyu_output.cc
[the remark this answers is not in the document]
fiesh: Not quite sure about this.  `first` and `second` have the advantage of making clear to the user what arguments they are in the call to `std::partition_copy`.  What the two ranges contain is rather difficult and explained in length further up.

iwyu_output.cc
fiesh: Good suggestion!

fiesh:
I included almost all of your suggestions, I think they have improved the code a lot.  If you can have a look at it again, I'd be most thankful!

iwyu_output.cc
```diff
@@ -1426,6 +1426,70 @@ void CalculateIwyuForFullUse(OneUse* use,
...
+
+      if (preprocessor_info->FileTransitivelyIncludes(source.included_file(),
+                                                      target.included_file())) {
+        return false;
+      }
+    }
+    return true;
+  };
+
+  vector<OneIncludeOrForwardDeclareLine> first, second;
```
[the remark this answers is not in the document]
kimgr: "direct"/"indirect"?

iwyu_output.cc
```diff
@@ -1426,6 +1426,70 @@ void CalculateIwyuForFullUse(OneUse* use,
...
+    const vector<OneIncludeOrForwardDeclareLine> & lines,
+    const IwyuPreprocessorInfo* preprocessor_info) {
+  // In the first step, we copy all lines that are an include line, not present,
+  // desired, and whose include_file is transitively included by another line
+  // which is a desired include line into a range second and the remaining ones
+  // into another range first.
+  // This is because elements of range second are then precisely those whose
+  // symbols are already exposed through a different header that will is to be
+  // included anyway, so they can be omitted.
+  const auto is_transitively_present = [&lines, preprocessor_info](
```
kimgr: I wonder if this name is right. I realize I suggested it, but I meant `is_transitively_present` as an example to make you describe the purpose. It looks to me like it now returns `false` for lines that have transitive representation but `true` for everything else, so is it backwards?

I'm also having trouble with the `source` and `target` terminology. I think `target` could be renamed `line` and `source` could be renamed `other`. That way it's clearer that you cross-reference every include-line against every other.

fiesh:
@kimgr Oh yes, I am.  Sorry for the delay.  It'll be a couple more days before I get to it though, I'm occupied with real life's work.

kimgr:
@fiesh Oh, great! I was considering picking up where you left off, but then I can focus on other things, thanks!

iwyu_output.cc
```diff
@@ -1426,6 +1426,70 @@ void CalculateIwyuForFullUse(OneUse* use,
   }
 }
 
+// This function removes all OneIncludeOrForwardDeclareLines a that are desired
```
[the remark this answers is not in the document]
fiesh: It's not a typo, the variable name is `a`.  I'll rename it to make the explanation clearer.

iwyu_output.cc
fiesh: You're right, I flipped its logic, so that it does match the name `is_trasitively_present` now and accordingly switched `first` and `second`.  (And then renamed `first` and `second`.)

fiesh:
I have incorporated nearly all your suggestions in the new commit.  The formatting was also fixed by applying `clang-format`, don't know why I hadn't done that before, sorry.

The only thing I did not change is the renaming of `source` and `target`.  I do feel they describe the situation in the dependency graph quite well, the `source`s being the sources of paths with the `target`s being their targets.

I would very much appreciate some first test from someone else though since I think me writing the test might lead to a false feeling of it working correctly.  I developed it using some include files whose behavior I wanted a certain way, and I think I might integrate certain assumptions.  Plus I alas am rather constrained time wise recently.  I do want to see this feature integrated though :)

Thanks for all your feedback, @kimgr

iwyu_output.cc
```diff
@@ -2007,6 +2072,10 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
     }
   }
 
+  if (GlobalFlags().tolerate_transitive) {
+	  lines_ = internal::PruneTransitivelyPresent(lines_, preprocessor_info_);
```
kimgr: Too much indentation here

iwyu_output.cc
```diff
@@ -1426,6 +1426,71 @@ void CalculateIwyuForFullUse(OneUse* use,
...
+  // desired, and whose include_file is transitively included by another line
+  // which is a desired include line into a range `indirect` and the remaining
+  // ones into another range `direct`. This is because elements of range
+  // `indirect` are then precisely those whose symbols are already exposed
+  // through a different header that is to be included anyway, so they can be
+  // omitted.
+  const auto is_transitively_present =
+      [&lines,
+       preprocessor_info](const OneIncludeOrForwardDeclareLine& target) {
+        if (!target.IsIncludeLine() || target.is_present() ||
```
kimgr: `is_present` and `is_desired` are both cheaper than `IsIncludeLine`, so please put them first.

iwyu_output.cc
```diff
@@ -1426,6 +1426,71 @@ void CalculateIwyuForFullUse(OneUse* use,
...
+  indirect.reserve(lines.size());
+  direct.reserve(lines.size());
+  std::partition_copy(lines.cbegin(), lines.cend(), std::back_inserter(indirect),
+                      std::back_inserter(direct), is_transitively_present);
+  // Now add the symbol uses from elements of range indirect to the
+  // corresponding source elements of range direct so that these sources are now
+  // listed as including the corresponding symbols because they do transitively.
+  for (const OneIncludeOrForwardDeclareLine& target : indirect) {
+    const auto source =
+        std::find_if(direct.begin(), direct.end(),
```
kimgr: Hmm. It looks a little off that you need to search for this again -- you've already worked every include line against every other once.

It looks like you already have an O(n^2) algorithm: `partition_copy` traverses all lines, and the lambda you're using to partition traverses all lines. And now this traverses all lines invoking `FileTransitivelyIncludes` again. I wonder if you could flatten this so it only needs to do the N*N loop once?

jru: Also think so. If the lambda returned a pointer to the file that transitively includes the target or null, instead of true or false, then a single pass through the lines should be enough. No extra vectors or partition/find_if would be needed either.

The lambda could then be called `transitively_included_by` or smth along those lines.

iwyu_output.cc
```diff
@@ -1426,6 +1426,71 @@ void CalculateIwyuForFullUse(OneUse* use,
...
+                           s.included_file(), target.included_file());
+                     });
+    CHECK_(source != direct.end())
+        << "Cannot find the source for an element"
+           "that was sorted out because it had a source";
+    for (const auto& p : target.symbol_counts()) {
+      source->AddSymbolUses(p.first, p.second);
+    }
+    // source is desired now because target was.
+    CHECK_(target.is_desired()) << "Target must be desired";
```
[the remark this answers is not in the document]
fiesh: I'm not entirely certain what else to say?  `source` is the direct provider, `target` was indirect, and `target` was desired, so that now source is.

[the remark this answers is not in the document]
fiesh: Or are you talking about the `CHECK_`?  It's just an assertion for what already happened in `is_transitively_present`.

iwyu_output.cc
fiesh: Yes good catch, that's really unnecessary!

fiesh: OK, now that I've looked at it a bit, I'm not sure what the best venue would be.  Returning a pointer isn't really that good of an option because it would still require two iterations over the range: one to add the `target`s' symbols to the `source`s, and one with a `std::remove_if` or so to eliminate these `target`s then.  I don't see how to do this in one go.

One option would be to copy everything into a `vector` of `pair`s of `OneIncludeOrForwardDeclarLine`s and pointers to such, then fill the pointers, then add the symbols, then call `remove_if` on all whose pointer is a `nullptr`, and then `transform` the range back into one where the pointers are removed.  Not sure if I like that option better, also not sure if it's actually faster or slower since it involves more allocations and copying, which might be worse than doing some work twice.

Thoughts?

fiesh:
The new commit addresses all comments except the two that I in turn commented.  Thanks for the feedback!

iwyu_output.cc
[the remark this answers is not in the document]
kimgr: Yes, the check message. If it ever fires, it's nice to know up-front *why* target must be desired (presumably because you only selected desired targets before?)

iwyu_output.cc
kimgr: I need to sit down and think about this for a while. It must be possible to make this better than O(N*N) + O(N) (where `FileTransitivelyIncludes` is also non-trivial)

jru: I thought briefly about this and I believe it would actually need two passes through the lines to prevent aliasing issues in its current form: one to save the <source,target>pairs and one more to process the vector, and add the symbols from each target to the corresponding source. For the first pass, only the sources that found a valid target would need to be stored (skipping the ones that return null)

That said, I can't actually try anything right now, so please take everything with a grain of salt :)

kimgr:
I haven't had much time to experiment with this, and my personal schedule is pretty packed in the coming month, so it might be a while until I can get to it.

One other idle thought that popped up: you're discarding a lot of includes in the new `Prune` method. Should you instead just mark them as non-desired? That way we could add a later pass and still include them in the analysis (if that becoms necessary for some reason).

fiesh:
Oh wow!  I will look into this in about two weeks, alas I cannot do it before then.  I'm very interested!

jru:
I think it looks great! I added a couple comments.

kimgr:
Here's a stable link with @jru's `is_present` comment addressed: https://github.com/kimgr/include-what-you-use/commit/tolerate_transitive.

I would really need to write up some test cases to see if this works according to expectations...

kimgr:
I wrote a simple test case and debugged this. I think the big takeaway from that was to only fiddle with desred/present in the second pass, after the include map is built. The transitive graph is the same whether lines are present and desired or not.

I think the test boiled down to three distinct cases:
1) direct.h and indirect.h are both included, only indirect.h symbols are used: remove direct.h because IWYU
2) direct.h and indirect.h are both included, symbols from both are used: leave indirect.h because IWYU
3) Only direct.h is included, symbols from both are used: do nothing because tolerate-transitive

Does that jive with everybody else's assumptions? Can you see a case I've missed?

I'll turn these into real test cases and push a branch soon.

fiesh:
@kimgr Sorry for my late reply, I had a lot of things going on in real life.

Your three cases look correct and cover all relevant cases.

I looked at your code and find it very clean and much better readable, great job!  There's one minor recommendation from my side: Ditch the `multimap` in favor of a plain `vector`, which is then `sort`ed and `erase`d + `unique`d.  This tends to be much faster.

kimgr:
Posted #594 to replace this. @fiesh -- please review that version and check if it meets your needs. Closing.

PR #531, opened 2018-03-24
Clang 6.0

adhokshajmishra:


kimgr:
Thanks for your contribution!

Note that all version numbers in the documentation are *examples* (and many of the mentions say so explicitly). I don't think updating them to a single yet unsupported version is making things clearer.

If you have other ideas for how to make this more obvious, I'd be happy to hear them, but unless I'm missing something, this patch will not be accepted.

PR #535, opened 2018-04-08
Ignore class qualifier in out of line methods

jru:
Instead of aborting the AST traversal completely whenever a class
qualifier is found, skip only the travesal of the class template.

The reasoning here is that a class template will be inspected during the
regular instantiation anyway, so there is no need to duplicate efforts.

Apart from performance implications, scanning the template here makes
IWYU report all template parameters as if specified explicitly by the
user (effectively ignoring default template params). For example,
std::allocator would get reported as a full-use during a call to
push_back() because it is defined outsidef of the std::vector class.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2583,23 +2583,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     ASTNode* ast_node = current_ast_node();
     ast_node->set_in_forward_declare_context(false);
 
-    // For method calls it doesn't matter if a method is defined inside
-    // a class or outside of it.  We detect out-of-class method calls with
-    // the pattern
-    //
-    //   CallExpr
-    //   `-CXXMethodDecl
-    //     `-NestedNameSpecifier
-    //
-    // and skip traversing method qualifier as in-class methods don't have it.
-    if (const CXXMethodDecl* parent = ast_node->GetParentAs<CXXMethodDecl>()) {
-      if ((nns == parent->getQualifier()) &&
-          ast_node->AncestorIsA<CallExpr>(2)) {
-        VERRS(7) << "Skipping traversal of CXXMethodDecl qualifier "
-                 << PrintableNestedNameSpecifier(nns) << "\n";
-        return false;
-      }
-    }
     return true;
   }
 
@@ -3000,8 +2983,27 @@ class InstantiatedTemplateVisitor
   bool TraverseTemplateSpecializationTypeHelper(
       const clang::TemplateSpecializationType* type) {
     if (CanIgnoreCurrentASTNode())  return true;
-    if (CanForwardDeclareType(current_ast_node()))
-      current_ast_node()->set_in_forward_declare_context(true);
+
+    // Skip the template traversal if this occurrence of the template name is
+    // just a class qualifier for an out of line method, as opposed to an object
+    // instantiation, where the templated code would need to be inspected.
+    //
+    // Class<T,U>::method() {
+    // |-Type---^
+    // |-NNS------^
+    // |-CXXMethodDecl--^
+    ASTNode* ast_node = current_ast_node();
+    if (const auto* nns = ast_node->GetParentAs<NestedNameSpecifier>()) {
+      if (nns->getAsType() == type) {
+        if (const auto* method = ast_node->GetAncestorAs<CXXMethodDecl>(2)) {
+          CHECK_(nns == method->getQualifier());
+          return true;
+        }
+      }
+    }
+
+    if (CanForwardDeclareType(ast_node))
+      ast_node->set_in_forward_declare_context(true);
     return TraverseDataAndTypeMembersOfClassHelper(type);
   }
 
--- a/tests/cxx/out_of_line-dep-char.h
+++ b/tests/cxx/out_of_line-dep-char.h
@@ -1,0 +1,14 @@
+//===--- out_of_line-dep-char.h - test input file for iwyu ----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_OUT_OF_LINE_DEP_CHAR_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_OUT_OF_LINE_DEP_CHAR_H_
+
+void Dependent(char);
+
+#endif  // INCLUDE_WHAT_YOU_USE_TESTS_CXX_OUT_OF_LINE_DEP_CHAR_H_
--- a/tests/cxx/out_of_line.cc
+++ b/tests/cxx/out_of_line.cc
@@ -1,0 +1,78 @@
+//===--- out_of_line.cc - test input file for iwyu ------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "out_of_line-dep.h"
+
+template <class T, class U = char>
+struct Class {
+  using type = T;
+
+  void MethodIn() { Dependent(T()); }
+  void MethodOut();
+
+  static void MethodStaticIn() { Dependent(T()); }
+  static void MethodStaticOut();
+};
+
+template <class T, class U>
+void Class<T, U>::MethodOut() {
+  Dependent(T());
+}
+
+template <class T, class U>
+void Class<T, U>::MethodStaticOut() {
+  Dependent(T());
+}
+
+template <>
+void Class<int, int>::MethodOut() {
+  // IWYU: Dependent is.*dep-int.h
+  Dependent(type{});
+}
+
+
+int main()
+{
+    //Inline template method
+    Class<int> c;
+    // IWYU: Dependent is.*dep-int.h
+    c.MethodIn();
+
+    //Out-of-line template method
+    Class<int> c2;
+    // IWYU: Dependent is.*dep-int.h
+    c2.MethodOut();
+
+    //A explicit specialization of MethodOut(). It does not
+    //require Dependent(int) here, because it must be already
+    //available at the definition.
+    Class<int, int> c3;
+    c3.MethodOut();
+
+    //Inline static method
+    // IWYU: Dependent is.*dep-int.h
+    Class<int>::MethodStaticIn();
+
+    //Out-of-line static method
+    // IWYU: Dependent is.*dep-int.h
+    Class<int>::MethodStaticOut();
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/out_of_line.cc should add these lines:
+#include "out_of_line-dep-int.h"
+
+tests/cxx/out_of_line.cc should remove these lines:
+- #include "out_of_line-dep.h"  // lines XX-XX
+
+The full include-list for tests/cxx/out_of_line.cc:
+#include "out_of_line-dep-int.h"  // for Dependent
+
+***** IWYU_SUMMARY */
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -1520,9 +1520,7 @@ int main() {
   // IWYU: std::vector is...*<vector>
   // IWYU: I2_Enum is...*badinc-i2.h
   std::vector<I2_Enum> local_enum_vector;
-  // TODO: In C++03 and earlier I2_Enum is required below. The fact that this
-  // depends on standard version indicates a bug in our template handling, as
-  // the instantiation chain is different in old vs. new standard libraries.
+  // IWYU: I2_Enum is...*badinc-i2.h
   // IWYU: std::vector is...*<vector>
   // IWYU: I21 is...*badinc-i2.h
   local_enum_vector.push_back(I21);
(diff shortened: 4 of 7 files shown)


jru:
>  I can see that that declaration requires I1_Enum, but I don't understand why the method call does... Is it because the return type of static_inline is I1_Enum in this instantiation? 

Yes, that's correct. IWYU reports all types from the template substitution.

To verify this is consistent with instance inline methods I added this line in the test:
```cpp
h_templateclass2.a(); //Returns a FOO -> I1_Enum
```
And IWYU correctly reports *I1_Enum* as well

> Do we really need a complete type for a return value we don't use?

Good question. I can check what we do in other cases.

> Huh, so why is I1_Enum not required here? Some other quirk that we're missing?

Actually, it is not required because *tplnested()* does not return or otherwise use any FOO, but I found that the line below does return a FOO and it is not reported.. I will investigate! 
https://github.com/include-what-you-use/include-what-you-use/blob/5082fddccb3d5aabaace2208f1162029a27c0334/tests/cxx/badinc.cc#L1850

kimgr:
> > Do we really need a complete type for a return value we don't use?
>
> Good question. I can check what we do in other cases.

I just did a micro-test here:
```c++
// api.h
#ifndef INCLUDED_API_H
#define INCLUDED_API_H

// Forward-declare return type
struct Type;

Type function();

#endif // INCLUDED_API_H

// type.h -- unused
#ifndef INCLUDED_TYPE_H
#define INCLUDED_TYPE_H

struct Type {
	int value;
};
#endif // INCLUDED_TYPE_H

// main.cc
#include "api.h"

int main() {
  function();
  return 0;
}
```

This fails with:
```
$ clang main.cc 
main.cc:4:3: error: calling 'function' with incomplete return type 'Type'
  function();
  ^~~~~~~~~~
./api.h:7:6: note: 'function' declared here
Type function();
     ^
./api.h:5:8: note: forward declaration of 'Type'
struct Type;
       ^
1 error generated.
```

And on reflection this makes sense, as `Type` might have a non-trivial destructor that still needs to be called even if the return value is not bound to a variable.

Oh, but there's also IWYU's idea of Author Intent:
https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhatIsAUse.md#author-intent

That is, `api.h` can't compile on its own without the `Type` declared. If we `#include "type.h"` we say that `api.h` is responsible for exporting `Type`, but if we forward-declare `Type`, we push that responsibility to callers. This is codified in `GetCallerResponsibleTypeForFnReturn`: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L1616.

I'm not sure if there's something about templates that changes this analysis, but I suspect that a template in general can never be responsible for its template arguments. An explicit specialization, on the other hand, supposedly could.

Anyway, that answered my own question in general, but I'm not sure it helped much :)

tests/cxx/out_of_line.cc
```diff
@@ -0,0 +1,70 @@
...
+  dependent(T{});
+}
+
+template <class T, class U>
+void Class<T, U>::method_static_out() {
+  dependent(T{});
+}
+
+template <>
+void Class<int, int>::method_out() {
```
kimgr: As per my earlier comment, it would be interesting to see an explicit specialization with one of the specialized arguments as return type, something like:

```
// somewhere-else.h
struct Type {};

// template.h
Type Class<Type, int>::method_static_out() {
    return Type();
}
```
This won't work exactly as-is, because the inline implementation will require the complete type of `Type`, but it might be possible to do something tricky to provoke the case in question.

iwyu.cc
```diff
@@ -3000,8 +2983,27 @@ class InstantiatedTemplateVisitor
   bool TraverseTemplateSpecializationTypeHelper(
       const clang::TemplateSpecializationType* type) {
     if (CanIgnoreCurrentASTNode())  return true;
-    if (CanForwardDeclareType(current_ast_node()))
-      current_ast_node()->set_in_forward_declare_context(true);
+
+    ASTNode* const ast_node = current_ast_node();
```
kimgr: I'd move this declaration down below the comment.

iwyu.cc
```diff
@@ -3000,8 +2983,27 @@ class InstantiatedTemplateVisitor
...
+    // If, in this context, the template type is only a qualifier for a method
+    // defined out of line, don't treat it as an actual specialization. Just
+    // ignore it.
+    //
+    // Class<T,U>::method() {
+    // |-Type---^
+    // |-NNS------^
+    // |-CXXMethodDecl--^
+    //
+    if (auto nns = ast_node->GetParentAs<NestedNameSpecifier>())
```
kimgr: Please use braces around all nested ifs (when an inner if-block is braced, the freehanging outer ones look weird)

kimgr: `const auto*`

iwyu.cc
```diff
@@ -3000,8 +2983,27 @@ class InstantiatedTemplateVisitor
...
+      if (nns->getAsType() == type)
+        if (auto method = ast_node->GetAncestorAs<CXXMethodDecl>(2)) {
```
kimgr: `const auto*`

jru:
> That is, api.h can't compile on its own without the Type declared. If we #include "type.h" we say that api.h is responsible for exporting Type, but if we forward-declare Type, we push that responsibility to callers. This is codified in GetCallerResponsibleTypeForFnReturn

Thanks for looking into this. I took a look at the docs (hey! there is actually great info there :)) and then tried a few more examples. After giving it some thought I think I might now what is going on:

- I believe *CodeAuthorWantsJustAForwardDeclare* is not implemented correctly: As it is now, if no forward declare was found in the same file before the return decl, it already assumes that the code author provides the definition and that the user doesn't have to. This does not seem right. What if the definition is not provided because the return type is a template parameter defined somewhere else (like in the example above)? 

- Also, the implementation seems to be in contradiction with the [docs](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhatIsAUse.md#automatic-re-export-typedefs): As the first point reads, in order to decide that the author of the declaration intends to re-export the definition of the return type, it needs to provide the definition in its includes. The next two points unconditionally assume that the author does not want to re-export. So, the only possible way of concluding that a re-export is intended is through (1), making (2) redundant. That seems wrong too.

- The return type is required for *h_templateclass2.static_inline(I22)* and *h_templateclass2.static_out_of_line(I22)* because they are templates and both make full use of *FOO*. If they were regular functions *I1_Enum* wouldn't be reported because of the previous points (although it should)

- In the case of *h_templateclass2.h_nested_struct.static_tplnested(I22)*, the return type is not reported becuase IWYU fails to analyze the template for dependent nested structs, given that the parent isn't itself a template. This seems like an unrelated defect. I can create a new report for it.


I'll go through your review comments in the meantime.
Thanks!

jru:
I went through the corrections.

> When you say "the template type is only a qualifier for a method defined out of line", I got the sense that it was not a specialization, but it still is, right?

I reworded the comment to make it clearer. What I meant was not that this is not a specialization per-se but rather than in this specific context we are safe to not explore the type like a regular instantiation would. 

>On the one hand, it looks a little bad to short-circuit the specialization case when we got here and it wasn't a specialization. 

Note that the only thing that it's skipped is the IWYU logic that traverses the template type and its methods, which will happen anyway whenever the type is mentioned. The recursive traversal from Clang is completely unaffected (as opposed to the change I undid, where it was really aborting it)

> It would seem better to not call this code if this case is true. 

This helper is called by *TraverseTemplateSpecializationType* and *TraverseTemplateSpecializationTypeLoc*, both of which are entry points in the AST traversal, so there seems to be no other way.

tests/cxx/out_of_line.cc
jru: I am not sure what I would be testing with that. I mean, the specialization is really like a regular method so the template logic wouldn't really trigger here.

Can you be more specific what we try to verify here?

jru:
By the way, I also prepared a [minimal test](https://github.com/include-what-you-use/include-what-you-use/compare/master...jru:test/author_intent_broken) to better illustrate the point about the Author Intent logic being kind of dubious. IWYU suggests to remove the full type declaration where it is actually needed.

I also looked for similar issues and mentioned a few in the commit.

iwyu.cc
```diff
@@ -3000,8 +2983,29 @@ class InstantiatedTemplateVisitor
...
+    // a class qualifier for an out of line method and not an actual type
+    // instantiation. Note also that in this case the method decl will not
+    // have any dependent types and so will be treated as non-templated code.
+    //
+    // Class<int>::method() {
+    // |-Type---^
+    // |-NNS------^
+    // |-CXXMethodDecl--^
+    //
+    ASTNode* const ast_node = current_ast_node();
```
kimgr: Can you spell this `const ASTNode* ast_node = ...` instead? That seems to be prevalent style.

iwyu.cc
```diff
@@ -3000,8 +2983,29 @@ class InstantiatedTemplateVisitor
...
+    // Skip the template traversal if this occurrence of the template name is
+    // a class qualifier for an out of line method and not an actual type
+    // instantiation. Note also that in this case the method decl will not
+    // have any dependent types and so will be treated as non-templated code.
+    //
+    // Class<int>::method() {
+    // |-Type---^
+    // |-NNS------^
+    // |-CXXMethodDecl--^
+    //
```
kimgr: Can you also remove this empty comment line?

tests/cxx/out_of_line-dep-char.h
```diff
@@ -0,0 +1,14 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_OUT_OF_LINE_DEP_CHAR_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_OUT_OF_LINE_DEP_CHAR_H_
+
+void Dependent(char);
+
+#endif  //INCLUDE_WHAT_YOU_USE_TESTS_CXX_OUT_OF_LINE_DEP_CHAR_H_
```
kimgr: Space after `//`, please, here and throughout

tests/cxx/out_of_line.cc
```diff
@@ -0,0 +1,78 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "out_of_line-dep.h"
+
+template <class T, class U = char>
+struct Class {
+  using type = T;
```
kimgr: I'd like for this type name to be PascalCase too, can you rename it `Type`?

tests/cxx/out_of_line.cc
```diff
@@ -0,0 +1,78 @@
...
+
+  void MethodIn() { Dependent(T{}); }
```
kimgr: Use `T()` instead of C++11 initialization syntax

kimgr: ... here and throughout

iwyu.cc
```diff
@@ -3000,8 +2983,29 @@ class InstantiatedTemplateVisitor
...
+    ASTNode* const ast_node = current_ast_node();
+    if (const auto* nns = ast_node->GetParentAs<NestedNameSpecifier>()) {
```
kimgr: One some level I feel that maybe this code should move into iwyu_ast_util.cc, but I have a really hard time coming up with a name for it, so maybe it should just stay :)

iwyu.cc
jru: Not the same :) Anyway, I got rid of it. For such a small function it is noisy.

iwyu.cc
jru: Yeah, I think so too. Not clear this functionality can be so useful to deserve its own function anyway. And it can always be changed if it turns out otherwise.

kimgr:
Re your Author Intent example, I believe that's actually working as intended. I suspect if you replace the fwd-header with an inline forward-decl, it will behave as expected. So the problem is we don't understand that fwd-headers legally provide forward-decls. But I really don't want to commingle these two issues in the same GitHub issue, so could you please open a new one if you have more concerns? Thanks!

jru:
I just reported both open issues discussed here: #536 , #537 

Other than that, I finished the code suggestions and should be ready to merge. Let me know if there is anything else I should do!

kimgr:
I have a few small nits I would like to get straightened out, but I'll adopt your branch and make the requisite changes while I test on my various platforms. Thank you for this contribution!

kimgr:
On my FreeBSD system, I see an additional IWYU diagnostic:

    tests/cxx/badinc.cc:1579: Unexpected diagnostic:
    I2_Enum is defined in "tests/cxx/badinc-i2.h", which isn't directly #included.

This comes from this line in badinc.cc:

    // TODO(csilvers): IWYU: I2_Enum is...*badinc-i2.h
    // IWYU: std::vector is...*<vector>
    delete newed_vector;

Since there's a TODO for this construct, I think IWYU is doing something right :) Probably libc++ (which is the system C++ library on BSD) is constructed in a way that triggers this, and your standard library implementation is not.

kimgr:
Merged after small adjustments!

jru:
> Probably libc++ (which is the system C++ library on BSD) is constructed in a way that triggers this, and your standard library implementation is not.

My dist is based on Ubuntu Xenial, which uses libstdc++. I installed the available libc++ in the package manager (3.7) and there are some unexpected warnings, but not the one you mentioned. Maybe you have a newer version? 

But anyway, getting a clean run everywhere (especially Windows) is going to take a while. At least as much as getting rid of the library-specific code (MapPrivate.*) in IWYU :)

PR #540, opened 2018-05-11
Update for Clang r332021

kimgr:
PPCallbacks::InclusionDirective grew a new parameter.


kimgr:
Thanks, we should be back to green!

PR #543, opened 2018-05-13
Add alternative mapping for std::function

### LINKED ISSUE #487: std::function gets private headers suggested
Hi,

As per IWYU 0.8 (based on clang 4.0.1-1, from Debian), anything std::function gets the wrong headers included:

```c++
#include <functional>

void bar();

void foo()
{
  std::function<void()>(foo) = bar;
}
```

```
atum17:~> /usr/bin/iwyu -std=c++14 -c test.cc

test.cc should add these lines:
#include <bits/std_function.h>  // for function
#include <algorithm>            // for move

test.cc should remove these lines:
- #include <functional>  // lines 1-1

The full include-list for test.cc:
#include <bits/std_function.h>  // for function
#include <algorithm>            // for move
---
```

This is a double error; not only does it suggest <algorithm> when no move was made, but it also suggests a private header from bits/.

kimgr:
libstdc++ 7.1 and later defines std::function in bits/std_function.h
instead of bits/stl_function.h. Support both.

Fixes #487.

### DIFF
--- a/gcc.stl.headers.imp
+++ b/gcc.stl.headers.imp
@@ -39,6 +39,7 @@
   { include: ["<bits/ostream_insert.h>", private, "<ostream>", public ] },
   { include: ["<bits/postypes.h>", private, "<iostream>", public ] },
   { include: ["<bits/slice_array.h>", private, "<valarray>", public ] },
+  { include: ["<bits/std_function.h>", private, "<functional>", public ] },
   { include: ["<bits/stl_algo.h>", private, "<algorithm>", public ] },
   { include: ["<bits/stl_algobase.h>", private, "<algorithm>", public ] },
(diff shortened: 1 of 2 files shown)


jru:
I am wondering if it makes sense to have a test case for this. Even though it would be library-specific.

kimgr:
We don't have a way yet to limit tests to only selected environments. This mapping is only valid for libstdc++, not for libc++, MSVC's STL or any of the other standard library impls in circulation. So far we've punted on tests for mappings, for that reason.

kimgr:
This seems like a simple improvement, and the Travis build is happy, so I'll just merge it. It would be nice if we could figure something out for testing mappings later.

PR #545, opened 2018-05-22
imp files for Intel intrinsics

pseyfert:
(reopening #436, found time to look into this again)

Rules are based on the files' error messages when included directly.
These differ for different compilers, hence separate files for gcc and
clang. When two possible public headers are suggested, both are allowed
and it's left up to the iwyu standard behaviour which to suggest.

Iiuc we left #436 with how to deal with cases where multiple public headers are possible. Thinking about this now, I think "both are fine and suggest whatever the iwyu logic prefers" is a sane approach and doesn't need any work for the imp file.


kimgr:
I've never used the intrinsics headers in a project, so I don't fully understand what kinds of issues one might run into. I wonder if these mapping files should be version-tagged as well? So `clang-6.intrinsics.imp` and `gcc-8.intrinsics.imp`. That way we don't have to worry about backwards/forwards compat as much, we can just generate new mappings for whatever version needs something else.

Otherwise, looks good to me.

pseyfert:
Indeed good point. I checked the gcc versions I could easily access (4.8.4, 4.9.3, 6.2.0, 7.3.0, 8.1.0). All of them add new headers in new revisions (support for more instructions/instruction sets). Other than additions there was only one change I grepped:

 - `pcommitintrin.h` disappears from gcc6 to gcc7

I add `pcommitintrin.h` to `gcc.intrinsics.imp` so it should work for these 5 gcc versions (i.e. we have some backwards compatibility). But it also shows that forwards compatibility is not guaranteed.

gcc-8.intrinsics.imp
```diff
@@ -0,0 +1,61 @@
+# based on the #error pragmas in the header files below
+# based on their version in gcc 8.1.0
+# backwards compatibility tested ongcc 4.8.4, 4.9.3, 6.2.0, 7.3.0
```
kimgr: Typo: ongcc

gcc-8.intrinsics.imp
```diff
@@ -0,0 +1,61 @@
+# based on the #error pragmas in the header files below
```
kimgr: Can probably collapse these two lines into one sentence, e.g "These mappings are based on #error pragmas in the header files below in gcc 8.1.0."

clang-6.intrinsics.imp
```diff
@@ -0,0 +1,60 @@
+# based on the #error pragmas in the header files below
```
kimgr: Can probably collapse these two lines into a single sentence, see below

pseyfert:
updated, and also rebased to the current master

kimgr:
Thanks, merged!

PR #546, opened 2018-05-23
Add boost import reference files for version 1.64

scramsby:


kimgr:
Were these changes generated by the command-lines embedded in the old .imp files? It's always nice if there's a recipe for generating them that doesn't require manual intervention.

I think we should keep the old .imp files and create new, version-tagged ones, so users can choose which mappings they want depending on which Boost version they're using. Can you generate entirely new files called `boost-1.64-all.imp` and `boost-1.64-all-private.imp` instead? Thanks!

scramsby:
These were generated using the script commented at the top of the reference file. I'll make these new versions newly named as you suggest and leave the old.

kimgr:
Commit message now needs some love -- `Re-add` is a little brief :)

scramsby:
Ack! Looks like my rebase went a bit awry there. Will update the commit message.

scramsby:
Done!

kimgr:
Thanks, looks great!

PR #547, opened 2018-05-23
Add additional mapping for size_t

scramsby:


kimgr:
Thanks for the patch!

Please don't make editorial and functional changes in the same commit. Can you split out the typo fix into its own commit?

When updating mappings, please make the corresponding change in both iwyu_include_picker.cc and the corresponding .imp file. In this case, the same mapping should be added to `gcc.symbols.imp`.

kimgr:
Thanks, looks good to me, style-wise. Can you please squash into a single commit?

Hmm. `<uchar.h>` is not shown in http://www.cplusplus.com/reference/cstring/size_t/. I'm guessing your libstdc++ implementation happens to export it from there?

I wonder if we should allow this in the interest of minimizing churn, or if it's better to be strict and only expose `size_t` from headers that are documented to provide it?

scramsby:
You can see on cppreference.com (which I've found to be more accurate and easier to use) that it's defined in uchar.h: http://en.cppreference.com/w/c/types/size_t

I also double checked and it's documented as such in the C11 standard documentation: http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf. See page section 7.28 on page 398.

kimgr:
Thanks! I keep getting the two confused :)

Looks good to me!

PR #548, opened 2018-05-23
Small change to return a string by const&

scramsby:


kimgr:
I think I prefer to keep the lifetime of the string disconnected from the lifetime of the `OneIncludeOrForwardDeclare`. Or did this stand out as a big cost in a profile?

scramsby:
No, just an opportunistic micro-optimization. Not a big deal so I'll close if you disagree.

PR #550, opened 2018-05-25
Refactor header sorting logic to be more clear

scramsby:


iwyu_output.cc
```diff
@@ -1774,29 +1774,45 @@ OutputLine PrintableIncludeOrForwardDeclareLine(
...
+
+LineSortOrdinal GetLineSortOrdinal(const OneIncludeOrForwardDeclareLine& line,
+                                   const set<string>& associated_quoted_includes,
+                                   const IwyuFileInfo* file_info) {
+  if (!line.IsIncludeLine())
+    return LineSortOrdinal::kForwardDeclaration;
+  if (file_info && file_info->is_pch_in_code())
+    return LineSortOrdinal::kPrecompiledHeader;
+
+  const auto& quotedInclude = line.quoted_include();
```
kimgr: No `auto` if the type is not visible on the rhs of the assignment, please. Except in emergency cases like map iteration :)

kimgr: And I'd prefer the concrete type to be `std::string` (sans reference) -- I find lifetime extension to be a wart on the language (among many)

iwyu_output.cc
```diff
@@ -1774,29 +1774,45 @@ OutputLine PrintableIncludeOrForwardDeclareLine(
     GetSymbolsSortedByFrequency(line.symbol_counts()));
 }
 
-typedef pair<int, string> LineSortKey;
+enum class LineSortOrdinal {
+  kPrecompiledHeader,
```
kimgr: Hmm, I'm also unsure about the naming of enum class enumerators. I think the `k` prefix has outlived its usefulness here, as the names are scoped in `LineSortOrdinal` anyway. I vote to skip the k-prefix.

iwyu_output.cc
scramsby: Sure thing.

iwyu_output.cc
scramsby: I totally agree. Was just trying to follow existing convention. I'll update and squash.

kimgr:
Thanks, looks good to me!

PR #554, opened 2018-05-26
Fix erroneous explanation of MSVC difference

scramsby:


docs/WhyIWYUIsDifficult.md
```diff
@@ -154,7 +154,7 @@ Conditional `#includes` are a problem for IWYU when the condition is false:
     #include <foo>
     #endif
 
-If we're not running under windows (and IWYU does not currently run under windows), we have no way of telling if foo is a necessary `#include` or not.
+Since we're not running using the Microsoft Visual C++ compiler (IWYU is built on top of Clang), we have no way of telling if foo is a necessary `#include` or not.
```
kimgr: Come to think of it, this is actually a terrible example nowadays, because IWYU/Clang can be made to define `_MSC_VER` in MSVC compatibility modes. Maybe the section should be generalized to talk about conditional compilation at large instead?

scramsby: Updated with a different example. Please let me know what you think.

scramsby:
Updated with a different example. Please let me know what you think.

kimgr:
Wonderful, thanks!

PR #556, opened 2018-05-30
Add option to change header ordering to place quoted headers first

scramsby:


scramsby:
All tests are now passing so this is ready for review.

tests/cxx/quoted_headers_first.cc
```diff
@@ -0,0 +1,36 @@
+//===--- prefix_header_operator_new.cc - test input file for iwyu ---------===//
```
kimgr: `quoted_headers_first.cc`

tests/cxx/quoted_headers_first.cc
```diff
@@ -0,0 +1,36 @@
...
+// Tests that IWYU will respect the --quoted_headers_first option.
+
+#include <iostream>
+#include <list>
+#include <map>
+#include <memory>
+#include "subfolder/indirect_subfolder.h"
+
+std::unique_ptr<IndirectSubfolderClass> CreateIndirectSubfolderClass()
+{
```
kimgr: Opening brace goes on the preceding line

tests/cxx/quoted_headers_first.cc
```diff
@@ -0,0 +1,36 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+// Tests that IWYU will respect the --quoted_headers_first option.
+
+#include <iostream>
+#include <list>
+#include <map>
+#include <memory>
+#include "subfolder/indirect_subfolder.h"
```
kimgr: Hmm, I wonder if it would be beneficial to test this with an associated header, too? To see that associated headers (usually included with quotes) are sorted before other quoted headers. Also, that would provide an opportunity to make sure `--quoted_headers_first` works in header files too (but then you'd probably need to add another quoted include in this `.cc` file).

iwyu_globals.cc
```diff
@@ -94,6 +94,8 @@ static void PrintHelp(const char* extra_msg) {
          "   --no_comments: do not add 'why' comments.\n"
          "   --no_fwd_decls: do not use forward declarations.\n"
          "   --verbose=<level>: the higher the level, the more output.\n"
+         "   --quoted_headers_first: when sorting includes, place quoted\n"
```
kimgr: "Quoted" is a quality of the `#include` line, not of the header.

All the other `LineSortOrdinal`s talk about "headers", but I think it makes sense to name this separately. The others use all sorts of heuristics to identify the kind of header, but this is really about the include line.

So I'd suggest `--quoted_includes_first`, and a thorough rename :-/

iwyu_output.cc
```diff
@@ -1778,6 +1778,7 @@ enum class LineSortOrdinal {
   PrecompiledHeader,
   AssociatedHeader,
   AssociatedInlineDefinitions,
+  QuotedHeader,
```
kimgr: And this would then become `QuotedInclude`

iwyu_output.cc
```diff
@@ -1800,6 +1801,9 @@ LineSortOrdinal GetLineSortOrdinal(const OneIncludeOrForwardDeclareLine& line,
       return LineSortOrdinal::AssociatedInlineDefinitions;
     return LineSortOrdinal::AssociatedHeader;
   }
+
+  if (GlobalFlags().quoted_headers_first && EndsWith(quotedInclude, "\""))
```
kimgr: Oops, I just noticed now that `quotedInclude` doesn't follow the `hacker_case` naming convention. I'll push a commit to fix that, you may need to resolve conflicts when rebasing.

tests/cxx/quoted_headers_first.cc
kimgr: Also, might be a good idea to try this with a PCH.

scramsby:
Will address your suggestions hopefully today. Agreed that a configuration mechanism would be a good idea at some point. Perhaps using the same serialization format as the .imp files would be a good way to go.

iwyu_output.cc
scramsby: Done.

iwyu_output.cc
scramsby: I can include in my fix if you prefer.

iwyu_globals.cc
scramsby: Done.

tests/cxx/quoted_headers_first.cc
scramsby: Done.

tests/cxx/quoted_headers_first.cc
scramsby: Done.

tests/cxx/quoted_headers_first.cc
scramsby: Done, including both associated header and pch.

scramsby:
Changes made to address all previous iteration suggestions.

iwyu.cc
```diff
@@ -332,7 +332,7 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
 
   bool TraverseStmt(Stmt* stmt) {
     if (current_ast_node_->StackContainsContent(stmt))
-      return true;               // avoid recursion
+      return true;               // avoid recursion
```
scramsby: Looks like a bad merge. Fixing.

tests/cxx/quoted_includes_first.h
```diff
@@ -0,0 +1,38 @@
+//===--- quoted_includes_first.cc - test input file for iwyu --------------===//
```
kimgr: `quoted_includes_first.h`

tests/cxx/quoted_includes_first.h
```diff
@@ -0,0 +1,38 @@
...
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
```
kimgr: I think we prefer old-school include guards, but for test headers we typically don't bother. Please remove the `#pragma` for now -- there are portability subtleties around `#pragma once` that I would prefer to keep out of the test suite.

tests/cxx/quoted_includes_first.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <iostream>
+#include <list>
+#include <map>
+#include <memory>
+#include "subfolder/indirect_subfolder.h"
+
+class quoted_includes_class {
+public:
+    static std::unique_ptr<IndirectSubfolderClass> CreateIndirectSubfolderClass();
```
kimgr: Two-space indentation.

tests/cxx/quoted_includes_first.h
```diff
@@ -0,0 +1,38 @@
...
+
+#pragma once
+
+#include <iostream>
+#include <list>
+#include <map>
+#include <memory>
+#include "subfolder/indirect_subfolder.h"
+
+class quoted_includes_class {
```
kimgr: Use PascalCase for class names. That said, I'm not sure the class adds anything here, the method might as well be a free function (?)

tests/cxx/quoted_includes_first.cc
```diff
@@ -0,0 +1,39 @@
...
+#include "tests/cxx/pch.h"  // this is the precompiled header
+#include <iostream>
+#include <list>
+#include <map>
+#include <memory>
+#include "subfolder/indirect_subfolder.h"
+#include "quoted_includes_first.h"
+
+std::unique_ptr<IndirectSubfolderClass> quoted_includes_class::CreateIndirectSubfolderClass() {
+    return std::make_unique<IndirectSubfolderClass>();
```
kimgr: Only two spaces indentation

kimgr: Oh, hmm... `make_unique` is C++14. I think this is now Clang's default, but it could be that there are platforms where this is not true (the default has been different for different targets before). Could you dumb this down to `return std::unique_ptr<IndirectSubfolderClass>(new IndirectSubfolderClass);`? Thanks!

scramsby:
Will address today. I think I'll keep the quoted include rename as part of the single commit because it is new to this PR so not much value to show addressing PR feedback separately.

tests/cxx/quoted_includes_first.h
scramsby: Done.

tests/cxx/quoted_includes_first.h
scramsby: Done.

tests/cxx/quoted_includes_first.h
scramsby: Done.

tests/cxx/quoted_includes_first.h
scramsby: Removed class.

tests/cxx/quoted_includes_first.cc
scramsby: Done and done.

scramsby:
Done!

kimgr:
Beautiful, thank you!

Issue #557, opened 2018-05-30
fix_includes.py removes forward declares in different namespaces


chri7325:
fix_includes.py will remove duplicate forward declare lines that exist in seperate namespaces.  This appears to happen at _DeleteDuplicateLines(file_lines, toplevel_reorder_spans).  This function only checks the string of each class for duplicates to remove.  The toplevel_reorder_spans don't seem to take into account different namespaces, causing the python script to remove the duplicated lines.

For example,
```
// test.cpp
namespace Outer_namespace
{
namespace Inner_namespace_1
{
class Test;
class Fake_class;
} // namespace Inner_namespace_1

namespace Inner_namespace_2
{
class Test;
} // namespace Inner_namespace_2
} // namespace Outer_namespace

Outer_namespace::Inner_namespace_1::Test *test_inner_namespace_1;
Outer_namespace::Inner_namespace_2::Test *test_inner_namespace_2;
```
Will produce
```

test.cpp should add these lines:

test.cpp should remove these lines:
- namespace Outer_namespace { namespace Inner_namespace_1 { class Fake_class; } }  // lines 6-6

The full include-list for test.cpp:
namespace Outer_namespace { namespace Inner_namespace_1 { class Test; } }  // lines 5-5
namespace Outer_namespace { namespace Inner_namespace_2 { class Test; } }  // lines 11-11
```
After running fix_includes.py on the file, the end result is
```
namespace Outer_namespace
{
namespace Inner_namespace_1
{
class Test;
} // namespace Inner_namespace_1

namespace Inner_namespace_2
{
} // namespace Inner_namespace_2
} // namespace Outer_namespace

Outer_namespace::Inner_namespace_1::Test *test_inner_namespace_1;
Outer_namespace::Inner_namespace_2::Test *test_inner_namespace_2;
```

chri7325:
It looks as though this is happening because of the namespace style.  For our codebase, we use the AfterNamespace style for clang format so { is always on the next line.  The script is making assumptions that the style is done with { on the same line as the namespace so the regex for namespace detection doesn't quite work for our coding style.

kimgr:
Thanks, that would be great! Put up a pull request, please!

chri7325:
[...]

PR #562, opened 2018-06-11
Enchance fix_includes.py to properly scan allman namespaces

chri7325:
The fix_includes.py script has been tailored to work for
namespaces that have { on the same line as the namespace.
For styles that do not follow this format, namespace
detection does not work properly as the line numbers and
regex statement are not able to be evaluated.  Logic has
been added to detect when an allman namespace has been
found and shifts the line numbers appropriately.

Fixes issue https://github.com/include-what-you-use/include-what-you-use/issues/557

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -104,6 +104,10 @@ _ENDIF_RE = re.compile(r'\s*#\s*endif\b')
 _NAMESPACE_START_RE = re.compile(r'\s*(namespace\b[^{]*{\s*)+(//.*)?$|'
                                  r'\s*(U_NAMESPACE_BEGIN)|'
                                  r'\s*(HASH_NAMESPACE_DECLARATION_START)')
+# Also detect Allman style namespaces.  Use a continue regex for validation
+# but don't add it to the normal regex list.
+_NAMESPACE_START_ALLMAN_RE = re.compile(r'\s*(namespace\b[^{]*)+(//.*)?$')
+_NAMESPACE_CONTINUE_ALLMAN_RE = re.compile(r'\s*{\s*(//.*)?$')
 _NAMESPACE_END_RE = re.compile(r'\s*(})|'
                                r'\s*(U_NAMESPACE_END)|'
                                r'\s*(HASH_NAMESPACE_DECLARATION_END)')
@@ -125,8 +129,8 @@ _HEADER_GUARD_DEFINE_RE = re.compile(r'\s*#\s*define\s+')
 # Note that not all of the above RE's are represented here; for instance,
 # we fold _C_COMMENT_START_RE and _C_COMMENT_END_RE into _COMMENT_LINE_RE.
 _LINE_TYPES = [_COMMENT_LINE_RE, _BLANK_LINE_RE,
-               _NAMESPACE_START_RE, _NAMESPACE_END_RE,
-               _IF_RE, _ELSE_RE, _ENDIF_RE,
+               _NAMESPACE_START_RE, _NAMESPACE_START_ALLMAN_RE,
+               _NAMESPACE_END_RE, _IF_RE, _ELSE_RE, _ENDIF_RE,
                _INCLUDE_RE, _FORWARD_DECLARE_RE,
                _HEADER_GUARD_RE, _HEADER_GUARD_DEFINE_RE,
                _PRAGMA_ONCE_LINE_RE,
@@ -683,6 +687,7 @@ def _CalculateLineTypesAndKeys(file_lines, iwyu_record):
   """
   seen_types = set()
   in_c_style_comment = False
+  in_allman_namespace = False
   for line_info in file_lines:
     if line_info.line is None:
       line_info.type = None
@@ -706,6 +711,9 @@ def _CalculateLineTypesAndKeys(file_lines, iwyu_record):
       in_c_style_comment = False
     elif in_c_style_comment:
       line_info.type = _COMMENT_LINE_RE
+    elif in_allman_namespace and _NAMESPACE_CONTINUE_ALLMAN_RE.match(line_info.line):
+      in_allman_namespace = False
+      line_info.type = _NAMESPACE_CONTINUE_ALLMAN_RE
     else:
       for type_re in _LINE_TYPES:
         # header-guard-define-re has a two-part decision criterion: it
@@ -719,6 +727,9 @@ def _CalculateLineTypesAndKeys(file_lines, iwyu_record):
           line_info.type = type_re
           if type_re == _INCLUDE_RE:
             line_info.key = m.group(1)   # get the 'key' for the #include.
+          elif type_re == _NAMESPACE_START_ALLMAN_RE:
+            # set in_allman_namespace to true in order to find {}
+            in_allman_namespace = True
           break
       else:    # for/else
         line_info.type = None   # means we didn't match any re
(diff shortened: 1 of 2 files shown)


fix_includes.py
```diff
@@ -968,24 +979,37 @@ class Myclass;
...
+    if line_info.type == _NAMESPACE_START_RE:
+      # Because multiple namespaces can be on one line
+      # ("namespace foo { namespace bar { ..."), we need to count.
+      # We use the max because line may have 0 '{'s if it's a macro.
+      # TODO(csilvers): ignore { in comments.
+      namespace_depth = max(line_info.line.count('{'), 1)
+    elif line_info.type == _NAMESPACE_START_ALLMAN_RE:
+      # For Allman namespaces, keep the start line and increment
+      # the namespace depths when the actual brace is encountered.
+      namespace_depth = 0
```
kimgr: It's a little disconcerting that `namespace_depth` is unbound if type is neither `_NAMESPACE_START_RE` nor `_NAMESPACE_START_ALLMAN_RE`. After some scanning, I found that that's guaranteed by the conditional above, but maybe it would be nice to add an `else` leg with an `assert False` to make sure that invariant doesn't break.

fix_includes_test.py
```diff
@@ -335,6 +335,80 @@ class Alert;                         ///-
 The full include-list for empty_internal_namespace:
 namespace maps_transit_realtime { namespace service_alerts { class StaticServiceAlertStore; } }   // lines 5-5
 ---
+"""
+    self.RegisterFileContents({'empty_internal_namespace': infile})
+    self.ProcessAndTest(iwyu_output)
+
+  def testRemoveEmptyAllmanNamespace(self):
```
kimgr: I haven't quite understood what the "continued Allman" regex matches -- should there be a test with nested namespaces to cover that code? E.g.
```
namespace A { namespace B { namespace C
{
}
}
}
```

or however it looks.

chri7325: @kimgr the continued allman regex simply matches the { on the next line.
namespace A // allman start re
{ // allman continued re
} // namespace end re

It is only triggered when an allman namespace is found (namespace without any { on the same line.  I'll add that test case and see if I can handle it somehow but it is a mix of both styles so it might be a bit tricky.

chri7325: Sorry, just to clarify a bit more, the "allman" style is something I'm getting from clang-format.  We use allman style for our clang-format option which states to always break before braces so for nested namespaces, it would be like the test cases supplied here
namespace A
{
namespace B
{
namespace C
{
} // namespace C
} // namespace B
} // namespace A

In order to support this formatting style, I added the regex statements for allman style as well as a regex for finding the appropriate continuation { line.

fix_includes.py
chri7325: @kimgr great point.  I've added an assert to guard against future additions.

fix_includes_test.py
chri7325: Logic and test cases were added for mixed namespaces.  Now all namespace styles should be supported (hopefully haha).

fix_includes.py
```diff
@@ -1344,6 +1389,19 @@ def _GetFirstNamespaceLevelReorderSpan(file_lines):
         break
       namespace_prefix += ('namespace %s { ' % m.group(1).strip())
 
+    elif line_info.type in (_NAMESPACE_START_ALLMAN_RE,
+                            _NAMESPACE_START_MIXED_RE):
+      # Trim any possibly comments on namespace to properly build
```
kimgr: Typo: possibly -> possible

kimgr:
Cool, we're nearly there!

chri7325:
@kimgr I've fixed the typo and rebased onto my last commit to keep things clean.  As for the line indicators, the way IWYU produces the output data is only to show which forward declares need to be removed.  They only function on the forward declare itself.  They only use multi-line indicators when you have forward declares that span multiple lines such as templated forward declares.  The actual removal of empty namespaces is done by the python script itself, not by the IWYU program which don't operate on them.  I have a bit more work that I was hoping to submit soon that does better empty namespace detection as well as multi-level namespace detection.  I've been running my enhancements over our entire codebase (~2 million loc) and haven't found any issues as of yet with the next PR request I was hoping to submit.  Please let me know if this would be okay and if I should open an issue first before submitting another PR.

fix_includes.py
chri7325: Woops, nice catch.

fix_includes.py
kimgr: :+1:

kimgr:
No need to open a bug issue for the subsequent changes, just post a PR, please!

chri7325:
@kimgr please be as picky as possible.  IWYU was an amazing tool that allowed us to reduce our build times in an nearly automated fashion so anything to keep the projects code quality high is a much appreciated.  I'll amend the messages right away.

fix_includes.py
```diff
@@ -706,6 +717,10 @@ def _CalculateLineTypesAndKeys(file_lines, iwyu_record):
       in_c_style_comment = False
     elif in_c_style_comment:
       line_info.type = _COMMENT_LINE_RE
+    elif (in_allman_or_mixed_namespace and
+         _NAMESPACE_CONTINUE_ALLMAN_MIXED_RE.match(line_info.line)):
```
kimgr: Ouch, line is under-indented by one space!

fix_includes.py
```diff
@@ -968,24 +987,43 @@ class Myclass;
   start_line = 0
   while start_line < len(file_lines):
     line_info = file_lines[start_line]
-    if line_info.deleted or line_info.type != _NAMESPACE_START_RE:
+    if (line_info.deleted or 
+       (line_info.type != _NAMESPACE_START_RE and
```
kimgr: Under-indented

chri7325:
Updated

fix_includes.py
```diff
@@ -1189,7 +1227,9 @@ def _ShouldInsertBlankLine(decorated_move_span, next_decorated_move_span,
            file_lines[next_line].type == _COMMENT_LINE_RE):
       next_line += 1
     return (next_line and next_line < len(file_lines) and
-            file_lines[next_line].type in (_NAMESPACE_START_RE, None))
+            file_lines[next_line].type in
```
kimgr: I'd align this as:
```
file_lines[next_line].type in (_NAMESPACE_START_RE,
                               _NAMESPACE_START_ALLMAN_RE,
                               _NAMESPACE_START_MIXED_RE,
                               None))
```

Not sure Python agrees that's a good thing.

fix_includes_test.py
```diff
@@ -299,6 +299,78 @@ class Baz;       ///-
...
 
+The full include-list for empty_namespace:
+#include <stdio.h>
+---
+"""
+    self.RegisterFileContents({'empty_namespace': infile})
+    self.ProcessAndTest(iwyu_output)
+
+  def testRemoveEmptyAllmanNamespace(self):
+    """Tests we remove a namespace with allman braces if we remove all fwd-decls inside it."""
```
kimgr: Allman

fix_includes_test.py
```diff
@@ -335,6 +407,78 @@ class Alert;                         ///-
 The full include-list for empty_internal_namespace:
 namespace maps_transit_realtime { namespace service_alerts { class StaticServiceAlertStore; } }   // lines 5-5
 ---
+"""
+    self.RegisterFileContents({'empty_internal_namespace': infile})
+    self.ProcessAndTest(iwyu_output)
+
+  def testRemovePartOfEmptyAllmanNamespace(self):
+    """Tests we remove a namespace with allman braces if empty, but not enclosing namespaces."""
```
kimgr: Allman

fix_includes_test.py
```diff
@@ -2557,6 +2701,59 @@ class Bar;       ///-
...
 - class Bar;  // lines 5-5
 
+The full include-list for iterative_namespace:
+---
+"""
+    self.RegisterFileContents({'iterative_namespace': infile})
+    self.ProcessAndTest(iwyu_output)
+
+  def testIterativeAllmanNamespaceDelete(self):
+    """Tests deleting an allman namespace with an emptied #ifdef inside it."""
```
kimgr: Allman

chri7325:
Updated.  I'm thinking if everything is okay now, I can squash everything into a single commit but I'm leaving them as is now so you can see what I just changed instead of rebasing now.  If you feel this is good, please also feel free to squash when merging too.

fix_includes.py
```diff
@@ -968,24 +987,43 @@ class Myclass;
   start_line = 0
   while start_line < len(file_lines):
     line_info = file_lines[start_line]
-    if line_info.deleted or line_info.type != _NAMESPACE_START_RE:
+    if (line_info.deleted or 
+        (line_info.type != _NAMESPACE_START_RE and
+        line_info.type != _NAMESPACE_START_ALLMAN_RE and
```
kimgr: These two lines should be indented one more space to align with the `line_info` above

kimgr:
Squashed and merged. Thank you for your contribution!

PR #563, opened 2018-06-14
Add --only_re option to fix_includes.py

vpoughon:
Add an option to `fix_includes.py`:

```
  --only_re=ONLY_RE     fix_includes.py will skip editing any file whose name
                        does not match this regular exression.
```

This is complementary to `--ignore_re` and can be useful for large codebases. Both `--ignore_re` and `--only_re` can be provided.


kimgr:
Looks good! Reviewing this on my phone, so I could be missing something, but I'll try and take another look later.

vpoughon:
typo fixed

PR #567, opened 2018-07-04
[fix-includes] Enable specifiying a 'base path' relative to which the IWYU output should be interpreted

pfaffe:
iwyu likes to output relative filenames, which can make running fix-includes tedious sometimes. This PR adds an option to specify a base-path which gets prepended to relative filenames.

### DIFF
--- a/fix_includes_test.py
+++ b/fix_includes_test.py
@@ -3978,7 +3978,7 @@ namespace ns { namespace ns4 { class Baz; } }
     self.RegisterFileContents({'add_fwd_declare_keep_iwyu_namespace': infile})
     self.ProcessAndTest(iwyu_output, expected_num_modified_files=1)
 
-  def testbasedir(self):
+  def testBasedir(self):
     self.flags.basedir = "/project/build/"
     iwyu_output = """\
 ../src/source.cc should add these lines:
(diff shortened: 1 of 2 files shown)


kimgr:
What's the use case? Are you saving IWYU output and applying it at a later date or something?

pfaffe:
My usecase is piping the output of iwyu_tool into fix_includes. Like a good student, I build my project out-of-tree. When I run iwyu_tool _in tree_ though, it prints output such as ```../include/adaptive/Indicators.h should add these lines:```, which is relative to the build directory. With the base-path option I can fix that, without having to run the pipeline in the build-directory.

pfaffe:
Added the chdir option you suggested. How should this be tested though?

kimgr:
Excellent question. The only reasonable way I can think of is to get rid of the `chdir` trick and go back to you original approach :-/

Do you still have the original branch in your reflog so I can see them side-by-side?

Sorry/thanks!

pfaffe:
Pushed it to [pfaffe:FixIncludesBasePath.Old](https://github.com/pfaffe/include-what-you-use/tree/FixIncludesBasePath.Old)

pfaffe:
True. Fixed.

kimgr:
Thanks! That led me to stumble on a couple of tests invoking `fix_includes.main`: https://github.com/pfaffe/include-what-you-use/commit/4514a636146487102eba35d1386f35662a95f6a6#diff-cb7792622fdf9e6204f9dadb046539dbR3452 (expand for more).

So one possible test case would be something like:
```
  def testChdir(self):
    """ Ensure -C switches the current working directory """
    # Give an empty stdin so we don't actually try to parse anything.
    old_stdin = sys.stdin
    old_cwd = os.getcwd()
    try:
      sys.stdin = StringIO()
      # This assumes we're not executing in /
      fix_includes.main(['fix_includes.py', '-C', '..'])
      self.assertEqual(os.path.dirname(old_cwd), os.getcwd())
    finally:
      sys.stdin = old_stdin
      os.chdir(old_cwd)
```

But overall, that's a pretty bad test case (doesn't say anything about the end-to-end behavior). Let's go with your original plan, could you update this PR back to that? I have some suggestions to offer there.

fix_includes_test.py
```diff
@@ -3713,6 +3717,27 @@ def testAddForwardDeclareInNamespaceWithKeepIwyuNamespaceFormat(self):
     self.RegisterFileContents({'add_fwd_declare_keep_iwyu_namespace': infile})
     self.ProcessAndTest(iwyu_output, expected_num_modified_files=1)
 
+  def testBasePath(self):
+    self.flags.base_path = "basepath/"
```
kimgr: I think this test would be clearer if you mimicked the build-dir/src-dir tension and with absolute paths, something like:

```
self.flags.base_path = "/project/build"
iwyu_output = """\
../src/source.cc should add these lines:
...
"""
...
self.RegisterFileContents({'/project/src/source.cc': infile})
```

And a doc comment for the test would also help.

pfaffe:
Rebased.

fix_includes_test.py
```diff
@@ -23,6 +23,7 @@
 except ImportError:
     from io import StringIO
 
+import os
```
kimgr: You no longer need `os`

kimgr:
I squashed and merged this as 6b236db.

Your test case still didn't cover the case I was worried about -- the combination of `--basedir` and relative paths on the command line -- so I added a testcase and fixed that bug.

Also noticed after merge that the new function didn't have any doc comment, and formatting was inconsistent with the rest of the file, so I just pushed a followup.

Thanks for your contribution!

PR #568, opened 2018-07-04
[CMake] Fix LLVM libs linked against.

pfaffe:
When built against trunk LLVM with SHARED_LIBS enabled, LLMVOption is missing from the linked libraries. Instead of hardcoding LLVM libnames, use the appropriate macro to resolve libraries from components.


kimgr:
Thank you!

I'd like to see this as two commits -- one refactor to use the macro, and one functional change to add the Option dependency.

You say in the commit message this is for the SHARED_LIBS build, but we're only applying it in what we interpret as a static build (i.e. if there's no LLVM target).

I wonder if the macro solves all this, so that we can skip the `if (TARGET LLVM)` check entirely?

TheErk:
I didn't know that macro (which is defined in cmake/modules/LLVM-Config.cmake by the way)
From reading it I don't think the macro spit out LLVM when LLVM_LINK_LLVM_DYLIB is set to ON.
This is two different setup to build SHARED_LIBS (i.e. build shared library instead static)
and LLVM_LINK_LLVM_DYLIB which build the LLVM shared lib and use it for linking LLVM tools.

TheErk:
The doc:
https://github.com/llvm-mirror/llvm/blob/master/docs/CMake.rst

says:
LLVM_BUILD_LLVM_DYLIB:BOOL
If enabled, the target for building the libLLVM shared library is added. This library contains all of LLVM's components in a single shared library. Defaults to OFF. This cannot be used in conjunction with BUILD_SHARED_LIBS. Tools will only be linked to the libLLVM shared library if LLVM_LINK_LLVM_DYLIB is also ON. The components in the library can be customised by setting LLVM_DYLIB_COMPONENTS to a list of the desired components.
LLVM_LINK_LLVM_DYLIB:BOOL
If enabled, tools will be linked with the libLLVM shared library. Defaults to OFF. Setting LLVM_LINK_LLVM_DYLIB to ON also sets LLVM_BUILD_LLVM_DYLIB to ON.

pfaffe:
Yes, these are orthogonal things. with the DYLIB build, it produces a single libLLVM.so (and the matching cmake target). In a BUILD_SHARED_LIBS build, it will on generate shared libraries for the _individual components_ instead of static ones. These are independent of each other.

pfaffe:
Regarding the macro: It's purpose is to give you library targets for components you want to link against. The LLVM codebase is grouped into logical components that serve a specifc purpose, such as `codegen`, `option`, or `transformutils` (You can check `llvm-config --components` for a complete list). The macro gives you all the libraries you need if you want to use a specific component.

kimgr:
I'm not sure the macro is useful here -- now we actually have all CMake targets available, and adding a target dependency will automatically include transitive deps. Note how your macro expansion lists more components than we had lib targets.

Does IWYU itself depend on LLVMOption, or is it a transitive dependency from e.g. Support? If the latter, there could be a missing dependency in LLVM's CMake system.

pfaffe:
LLVM's CMake does not encode inter-target dependencies. The macro's purpose is to encode them at least per component. Fundamentally the macro isn't mandatory of course, linking against individual libraries is entirely possible.

In this specific case, libclangFrontend and libclangDriver pull in libLLVMOption via shared object dependencies, but linking fails nonetheless:
```
/usr/bin/x86_64-linux-gnu-ld: CMakeFiles/include-what-you-use.dir/iwyu_driver.cc.o: undefined reference to symbol '_ZTVN4llvm3opt12InputArgListE'
/opt/llvm/lib/../lib/libLLVMOption.so.7: error adding symbols: DSO missing from command line
```

kimgr:
> LLVM's CMake does not encode inter-target dependencies.

I would argue that it does, but I can imagine there are quirks with dynamic linking.

I started looking at the LLVM/Clang CMake, and most CMake targets seem to use something like:
```
set( LLVM_LINK_COMPONENTS
  ${LLVM_TARGETS_TO_BUILD}
  Option
  Support
  )
```

at the top. It looks like the CMake system automatically picks this up and adds the relevant target deps (and I found at least one place where it defaulted to the `LLVM` target for when `LLVM_LINK_LLVM_DYLIB` is on).

I'll try that in my statically-linked environment and see what happens. I can post a little diff and maybe you could try it in your env?

pfaffe:
> I would argue that it does, but I can imagine there are quirks with dynamic linking.

In modern CMake, you can export inter-target dependencies publicly, which is what happens when you use e.g. `target_link_libraries(<Target> PUBLIC ...)`. Then, whenever you link against the CMake target <Target>, transitive dependencies get pulled in automatically. But this is only CMake internal. The actual artifacts encode dependencies differently, i.e. for shared objects, they're encoded for the linker as they'd normally be, in the .so. For static objects, they're not encoded.

Setting `LLVM_LINK_COMPONENTS` is equivalent to calling the macro and then saying `target_link_libraries()`. I don't have a preference here. In your example I'm missing  X86Desc and X86Info:
```
/usr/bin/x86_64-linux-gnu-ld: CMakeFiles/include-what-you-use.dir/iwyu.cc.o: undefined reference to symbol 'LLVMInitializeX86TargetMC'
/opt/llvm/lib/../lib/libLLVMX86Desc.so.7: error adding symbols: DSO missing from command line
/usr/bin/x86_64-linux-gnu-ld: CMakeFiles/include-what-you-use.dir/iwyu.cc.o: undefined reference to symbol 'LLVMInitializeX86TargetInfo'
/opt/llvm/lib/../lib/libLLVMX86Info.so.7: error adding symbols: DSO missing from command line
```

Here's my CMake settings for LLVM: 
```
-GNinja -DCLANG_DEFAULT_OPENMP_RUNTIME=libgomp -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DLLVM_TARGETS_TO_BUILD="X86;NVPTX" -DLLVM_BUILD_DOCS=OFF -DCMAKE_INSTALL_PREFIX="${RELEASE_TARGET}" -DLLVM_ENABLE_PROJECTS="llvm clang compiler-rt lld lldb"
```

kimgr:
I just tested my change with `LLVM_LINK_COMPONENTS`.

Current status:
- IWYU configured with fully static LLVM build - works!
- IWYU configured with `BUILD_SHARED` LLVM build - works!
- IWYU configured with `LLVM_BUILD/USE_DYLIB` LLVM build - fails. Both `LLVM*.a` and `LLVM.so` are linked and we end up with symbol conflicts.

I'll give this a shot with your original patch and see if it behaves differently.

kimgr:
This worked better, all configurations ran successfully! Unsurprisingly, since it still has special handling of the LLVM shared-object target.

I'll do some more digging the LLVM/Clang CMake and see if there's a third strategy.

pfaffe:
In LLVM and clang, when you add a library or executable and LLVM_LINK_LLVM_DYLIB is enabled, this library is linked against the DYLIB per default, otherwise against components defined by LLVM_LINK_COMPONENTS.

Can we just use the the add_clang_executable or add_clang_tool macro and have it do this?

kimgr:
I'm using the Ubuntu llvm-7 package from http://apt.llvm.org -- it ships with both static libraries for LLVM and Clang *and* the all-encompassing LLVM.so.

For some reason these Clang libraries have a dependency toward the .so instead of the corresponding static LLVM libs.

pfaffe:
I've just landed [r337366](http://llvm.org/r337366) to help with this. Should take a bit to bubble through packaging. That should take care of the duplicate options and symbols error.

kimgr:
This works for all three configurations I mentioned above! Would you like me to post a new PR with the branch I have (which only sets `LLVM_LINK_COMPONENTS`), or do you want to keep working on this one?

pfaffe:
I've updated this PR to use LLVM_LINK_COMPONENTS, since your commit was missing two components for me (I'm building shared libs but not libLLVM).

kimgr:
Great! This is identical to the patch I tested with locally, except I had the components alpha-sorted. Not sure if the order here affects link order? If not, I'd prefer to keep them sorted.

pfaffe:
Sorted them for you :)

kimgr:
Great, please squash the commits and write something coherent :) I'll merge it tomorrow.

I'll send out an announcement to the list once this is merged, to let people know they need to update their llvm/clang packages past r337336.

pfaffe:
I've pushed the squash to pfaffe/include-what-you-use@1d2e6e5, since I don't think I should force-push here. Do you want me to open a new PR for this?

kimgr:
Force-pushing to PR branches is fine, we do it routinely.

It only becomes a problem if someone has checked out your PR branch and added additional work on top of it, when they pull out the rewritten branch, their commits are no longer connected to the same ancestry. Git will complain and force them to resolve the conflicts, and there's always the reflog if something gets lost. People typically don't do that anyway, so I'm not worried.

pfaffe:
Done.

PR #569, opened 2018-07-07
Add option to disable line re-ordering

Phantal:
Current default behavior is to re-order modified lines relative to lines of a similar type (all forward decls get sorted, all project headers get sorted, etc).

This commit makes two changes:

* Adds options to enable/disable line re-ordering
* Defaults to no re-ordering; to get the original behavior the user must specify --reorder

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -2172,10 +2172,16 @@ def main(argv):
   parser.add_option('--nosafe_headers', action='store_false',
                     dest='safe_headers')
 
+  # --reorder and --no-reorder share the same destination variable.  Omitting
+  # the default here would still produce the intended behavior, however if
+  # the code is re-organized such that these two specifications are swapped
+  # relative to each-other there will be a change in behavior.
   parser.add_option('--reorder', default=False, action='store_true', dest='reorder',
                     help=('Re-order lines relative to other similar lines '
                           '(eg, headers relative to other headers)'))
 
+  # --reorder and --no-reorder share the same destination variable.  Omitting
+  # the default here would make the default True.
   parser.add_option('--no-reorder', default=False, action='store_false', dest='reorder',
                     help=('Do not re-order lines relative to other similar lines.'))
 
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1929,7 +1929,7 @@ def FixFileLines(iwyu_record, file_lines, flags):
   # For every move-span in our file -- that's every #include and
   # forward-declare we saw -- 'decorate' the move-range to allow us
   # to sort them.
-  move_spans = OrderedSet([fl.move_span for fl in file_lines if fl.move_span])  decorated_move_spans = []
+  move_spans = OrderedSet([fl.move_span for fl in file_lines if fl.move_span])
   for (start_line, end_line) in move_spans:
     decorated_span = _DecoratedMoveSpanLines(iwyu_record, file_lines,
                                              file_lines[start_line:end_line],
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1929,7 +1929,8 @@ def FixFileLines(iwyu_record, file_lines, flags):
   # For every move-span in our file -- that's every #include and
   # forward-declare we saw -- 'decorate' the move-range to allow us
   # to sort them.
-  move_spans = set([fl.move_span for fl in file_lines if fl.move_span])
+  seen = set()
+  move_spans = [fl.move_span for fl in file_lines if fl.move_span and not (fl.move_span in seen or seen.add(fl.move_span))]
   decorated_move_spans = []
   for (start_line, end_line) in move_spans:
     decorated_span = _DecoratedMoveSpanLines(iwyu_record, file_lines,
@@ -1963,7 +1964,10 @@ def FixFileLines(iwyu_record, file_lines, flags):
   # Add a sentinel decorated move-span, to make life easy, and sort.
   decorated_move_spans.append(((len(file_lines), len(file_lines)),
                                _EOF_KIND, '', []))
-  decorated_move_spans.sort()
+  if flags.reorder:
+    decorated_move_spans.sort()
+  else:
+     decorated_move_spans.sort(key=lambda x: x[0:-2])
 
   # Now go through all the lines of the input file and construct the
   # output file.  Before we get to the next reorder-span, we just
@@ -2168,6 +2172,13 @@ def main(argv):
                           ' header files; just add new ones [default]'))
   parser.add_option('--nosafe_headers', action='store_false',
                     dest='safe_headers')
+
+  parser.add_option('--reorder', default=False, action='store_true', dest='reorder',
+                    help=('Re-order lines relative to other similar lines '
+                          '(eg, headers relative to other headers)'))
+
+  parser.add_option('--no-reorder', default=False, action='store_false', dest='reorder'
+                    help=('Do not re-order lines relative to other similar lines.'))
 
   parser.add_option('-s', '--sort_only', action='store_true',
                     help=('Just sort #includes of files listed on cmdline;'
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -2177,7 +2177,7 @@ def main(argv):
                     help=('Re-order lines relative to other similar lines '
                           '(eg, headers relative to other headers)'))
 
-  parser.add_option('--no-reorder', default=False, action='store_false', dest='reorder'
+  parser.add_option('--no-reorder', default=False, action='store_false', dest='reorder',
                     help=('Do not re-order lines relative to other similar lines.'))
 
   parser.add_option('-s', '--sort_only', action='store_true',
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1930,6 +1930,7 @@ def FixFileLines(iwyu_record, file_lines, flags):
   # forward-declare we saw -- 'decorate' the move-range to allow us
   # to sort them.
   move_spans = OrderedSet([fl.move_span for fl in file_lines if fl.move_span])
+  decorated_move_spans = []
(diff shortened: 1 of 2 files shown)


Phantal:
I don't have a build environment setup that has internet access so I'm having to do this through the web interface.  Getting that test fixed is going to have to wait 'til I have a bit more free time.

fix_includes.py
```diff
@@ -2169,6 +2173,13 @@ def main(argv):
   parser.add_option('--nosafe_headers', action='store_false',
                     dest='safe_headers')
 
+  parser.add_option('--reorder', default=False, action='store_true', dest='reorder',
```
Flamefire: If you use `action='store_true'` you don't need the default as the parser already knows it

fix_includes.py
```diff
@@ -2169,6 +2173,13 @@ def main(argv):
...
+                    help=('Re-order lines relative to other similar lines '
+                          '(eg, headers relative to other headers)'))
+
+  parser.add_option('--no-reorder', default=False, action='store_false', dest='reorder',
```
Flamefire: And this is why you don't need a default: Using `--no-reorder` will store `false` into `reorder`, not using the flag will store the default `false` into `reorder` making this whole option having no effect.

I'd suggest you go for exactly 1 of those 2

fix_includes.py
```diff
@@ -1963,7 +1964,10 @@ def FixFileLines(iwyu_record, file_lines, flags):
   # Add a sentinel decorated move-span, to make life easy, and sort.
   decorated_move_spans.append(((len(file_lines), len(file_lines)),
                                _EOF_KIND, '', []))
-  decorated_move_spans.sort()
+  if flags.reorder:
+    decorated_move_spans.sort()
+  else:
+     decorated_move_spans.sort(key=lambda x: x[0:-2])
```
kimgr: What's the intent here? To sort by one-letter file extension (i.e. `.h`)? I don't think this holds water -- headers can be arbitrarily named, and stuff like `.hpp` and `.inl` is pretty common at least.

fix_includes.py
```diff
@@ -1929,7 +1929,8 @@ def FixFileLines(iwyu_record, file_lines, flags):
   # For every move-span in our file -- that's every #include and
   # forward-declare we saw -- 'decorate' the move-range to allow us
   # to sort them.
-  move_spans = set([fl.move_span for fl in file_lines if fl.move_span])
+  seen = set()
+  move_spans = [fl.move_span for fl in file_lines if fl.move_span and not (fl.move_span in seen or seen.add(fl.move_span))]
```
kimgr: You're changing `move_spans` from a set to a list to preserve order, right? I wonder if you could use `OrderedSet`, defined earlier in fix_includes.py, instead?

    move_spans = OrderedSet([fl.move_span for fl in file_lines if fl.move_span])

fix_includes.py
kimgr: Over-indented line

kimgr: Oh wait, the slice is the entire string except the trailing two chars... I still don't understand.

kimgr:
Thanks, added some comments.

fix_includes.py
Phantal: I don't have a strong preference for "on by default" or "requires a flag" for this feature.  Some people were of the opinion this should be the default.  I'll agree with whatever the maintainer(s) want.

fix_includes.py
Phantal: Each element of `decorated_move_spans` is a tuple; the intent is to remove the key and text from each tuple so it can only sort by the span and kind.  This will preserve the order within each kind, though it only preserves the order it's already in.

fix_includes.py
Phantal: Yes, the intention is to retain the order from `file_lines`.  This is the first time I've written anything in python, I'll yield to your judgement if you think it's appropriate.

fix_includes.py
Phantal: In the comment below you indicated the option will have no effect.  From my testing this is the behavior I see:

```
$ fix_includes.py -n -s /tmp/blah.cc | md5sum
513f864111d7efd9f66540eb373bebfc  -
$ fix_includes.py -n -s /tmp/blah.cc --no-reorder | md5sum
513f864111d7efd9f66540eb373bebfc  -
$ fix_includes.py -n -s /tmp/blah.cc --reorder --no-reorder | md5sum
513f864111d7efd9f66540eb373bebfc  -
$ fix_includes.py -n -s /tmp/blah.cc --reorder | md5sum
48316666f509f3d37221fb5b118a3574  -
```

Things to note from that test:

* `--no-reorder` doesn't change default behavior
* `--no-reorder` inverts the behavior of `--reorder`
* `--reorder` produces different output from the default

In my initial implementation I wasn't specifying `default`; in that scenario the default was based on the order arguments were declared.  I'm sure you'd agree that isn't ideal.

Unless I'm missing something I see only 2 worthwhile choices (unless someone devises a better answer):

* Maintainer(s) decide what the default should be, then eliminate the flag that corresponds with the default
* Keep it the way it is

Flamefire: Thanks for the explanation, I see what you mean now. Suggestion from here: Remove the default from the one not matching the default. From:
>In my initial implementation I wasn't specifying default; in that scenario the default was based on the order arguments were declared. I'm sure you'd agree that isn't ideal.

This is not fully true. You used `default=False, action='store_false'` here (on purpose as I see now). Omitting the default clause would implicitly change that to `default=True, action='store_false'`. Hence your observation about order. But removing the default from `default=False, action='store_true'` will not change anything.

As what you are doing here is "strange", a comment in the code would be good to highlight, that you have 2 switches for the same option and `default` is used on purpose.

Flamefire: Err, no! As I explained above removing the `default=False` from the first line does not change anything as that is the default for `store_true`. Removing it from the 2nd does change the meaning as it is **not** the default for `store_false` which makes that required.

Phantal:
Unless I'm mistaken this change is ready to go in.  For obvious reasons it would be best to squash merge.

If any further changes are needed, let me know.

fix_includes.py
```diff
@@ -2169,6 +2172,19 @@ def main(argv):
   parser.add_option('--nosafe_headers', action='store_false',
                     dest='safe_headers')
 
+  # --reorder and --no-reorder share the same destination variable.  Omitting
+  # the default here would still produce the intended behavior, however if
+  # the code is re-organized such that these two specifications are swapped
+  # relative to each-other there will be a change in behavior.
+  parser.add_option('--reorder', default=False, action='store_true', dest='reorder',
```
Flamefire: Sorry to complain again but the 2nd part is wrong: Omitting the default here (and only here) would not change anything even when reordering stuff. Both set the default to `False`: This one implicitly (if default removed) and the other one explicitly. Just remove the part starting with `Omitting...` would be fine with me, keep the default if you like, but again: It's completely optional here.

Flamefire:
LGTM 👍

fix_includes.py
Phantal: No worries, I'm not easily offended.  I am, however, a little slow sometimes; I understand now what you were getting at.

fix_includes.py
kimgr: Aah, it's a tuple, I completely missed that! Please add a little comment to describe the intent, something like `# Sort only by (span, kind), not the rest of the decoration`.

Now that I understand this better -- couldn't you just take `x[0:2]` to make sure you get the first two, rather than everything but the last two?

kimgr:
Thanks! This looks good to me too, I just want a little more clarification/cleanup on the sorting in the new mode.

fix_includes.py
```diff
@@ -2169,6 +2172,16 @@ def main(argv):
   parser.add_option('--nosafe_headers', action='store_false',
                     dest='safe_headers')
 
+  # --reorder and --no-reorder share the same destination variable.
+  parser.add_option('--reorder', action='store_true', dest='reorder',
+                    help=('Re-order lines relative to other similar lines '
+                          '(eg, headers relative to other headers)'))
```
kimgr: I'd prefer to spell `eg,` as `e.g.` here

fix_includes_test.py
```diff
@@ -41,6 +41,7 @@ def __init__(self):
     self.invoking_command_line = 'iwyu.py my_targets'
     self.find_affected_targets = True
     self.keep_iwyu_namespace_format = False
+    self.reorder = True
```
kimgr: It's a little annoying that the default is True in tests, but False in production... But we don't want to update all tests as part of this change set anyway, so let's keep this as-is for now

fix_includes.py
```diff
@@ -2169,6 +2172,16 @@ def main(argv):
...
+
+  # --reorder and --no-reorder share the same destination variable.  Omitting
+  # the default here would make the default True.
+  parser.add_option('--no-reorder', default=False, action='store_false', dest='reorder',
+                    help=('Do not re-order lines relative to other similar lines.'))
```
kimgr: No need for parentheses around the `help` string, parenthesized strings are only useful for multiline strings.

kimgr:
Would you like for me to pick up and finalize this change? I think it looks useful and would like to get it merged.

Phantal:
Sure.  I thought it was already finished / waiting to be merged in.

kimgr:
Merged as 5266ab8f5aa0029b8896c4c79ae178e96ffe241d with the following changes:

* "eg," -> "e.g." in help text
* `--no-reorder` -> `--noreorder` for consistency with other switches
* Mirror `--safe_headers`/`--nosafe_headers` in setup of switch
* Fix doc comments on tests to reflect actual behavior
* Fix long lines in tests

Thanks for the patch!

Phantal:
Thank you, much appreciated.

PR #571, opened 2018-07-14
Enhance fix_includes.py to understand multi-level namespaces

chri7325:
The fix_includes.sh script has logic to append new forward declares at
the namespace level if possible.  This logic only tries on the first
namespace encountered, however, so projects with a lot of namespaces
require a lot of manual tidying up.  Add multi-level detection so nested
namespaces have includes automatically inserted without creating new
namespace blocks when it is possible to do so cleanly.  If any issues
are encountered when evaluating namespaces, new forward declares will
be added at the top level.  Ifdeffed code will also be ignored when
trying to evaluate namespaces as well.

@kimgr I've run IWYU on our codebase with these changes and the results
were very successful.  We have our entire codebase formatted with Allman
namespaces so it is very normalized.  I changed from Allman namespaces to
nonAllman and had the same success.  I also tried nonAllman with compact
namespaces.  This covers the cases of
```
// Allman
namespace ns
{
namspace ns1
{
} // namespace ns1
} // namespace ns

// NonAllman
namespace ns {
namspace ns1 {
} // namespace ns1
} // namespace ns
```

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1407,12 +1407,14 @@ def _GetNamespaceLevelReorderSpans(file_lines):
       elif ifdef_depth != 0:
         continue # skip lines until we're outside of an ifdef block
 
-      # Build the simplified namespace lists and namespace depths.  When
-      # any new namespace is encountered, add the namespace to the list
-      # using the next line to cover namespaces without forward declares.
-      # When a forward declare is found, add the concatenated namespaces
-      # to the list.  Once a contentful line (None) has been found, add
-      # the final namespace if we're still inside one and stop searching.
+      # Build the simplified namespace dictionary.  When any new namespace is
+      # encountered, add the namespace to the list using the next line to cover
+      # namespaces without forward declarations.  When a forward declare is
+      # found, update the dictionary using the existing namespace span that the
+      # forward declare contains.  Once a contentful line (None) has been found
+      # or any exception occurs, return the results that have been found.  Any
+      # forward declare that wasn't able to have a proper namespace name found
+      # will still propigate to the top of the file.
       elif line_info.type == _NAMESPACE_START_RE:
         for namespace in _GetNamespaceNames(line_info.line):
           if not namespace:
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1424,7 +1424,12 @@ def _GetNamespaceLevelReorderSpans(file_lines):
           line_number+1, line_number+1)
 
       elif line_info.type == _NAMESPACE_START_ALLMAN_RE:
-        for namespace in _GetNamespaceNames(line_info.line):
+        pending_namespace_prefix = ''
+        namespaces = _GetNamespaceNames(line_info.line)
+        if len(namespaces) != 1:
+          raise FixIncludesError('Allman namespace found containing multiple '
+                                 'names: %s', line_info.line)
+        for namespace in namespaces:
           if not namespace:
             pending_namespace_prefix += 'namespace'
           else:
@@ -1437,6 +1442,7 @@ def _GetNamespaceLevelReorderSpans(file_lines):
         # nature of mixed namespaces, there will always be more than
         # one namespace so it is okay to assume that _GetNamespaceNames
         # will always return multiple records.
+        pending_namespace_prefix = ''
         namespaces = _GetNamespaceNames(line_info.line)
         for namespace in namespaces[:-1]:
           if not namespace:
@@ -1451,11 +1457,13 @@ def _GetNamespaceLevelReorderSpans(file_lines):
 
       elif line_info.type == _NAMESPACE_CONTINUE_ALLMAN_MIXED_RE:
         # Append to the simplified allman namespace.
+        if pending_namespace_prefix == '':
+          raise FixIncludesError('Namespace bracket found without an associated '
+                                 'namespace name at line: %s', line_number)
         pending_namespace_prefix += ' {'
         namespace_prefixes.append(pending_namespace_prefix)
         namespace_reorder_spans[' '.join(namespace_prefixes)] = (
           line_number+1, line_number+1)
-        pending_namespace_prefix = '' # reset this in case we find more
 
       elif line_info.type == _NAMESPACE_END_RE:
         # Remove C++ comments and count the ending brackets.
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1880,21 +1880,22 @@ def _DecoratedMoveSpanLines(iwyu_record, file_lines, move_span_lines, flags):
   if reorder_span is None:     # must be a new #include we're adding
     # If we're a forward-declare inside a namespace, see if there's a
     # reorder span inside the same namespace we can fit into.
-    namespace_reorder_spans = _GetNamespaceLevelReorderSpans(file_lines)
-    for namespace_prefix, possible_reorder_span in namespace_reorder_spans:
-      if (namespace_prefix and possible_reorder_span and
-          firstline.line.startswith(namespace_prefix)):
-        # Great, we can go into this reorder_span.  We also need to
-        # modify all-lines because this line doesn't need the
-        # namespace prefix anymore.  Make sure we can do that before
-        # succeeding.
-        new_firstline = _RemoveNamespacePrefix(firstline.line, namespace_prefix)
-        if new_firstline:
-          assert all_lines[first_contentful_line] == firstline.line
-          all_lines[first_contentful_line] = new_firstline
-          sort_key = re.sub(r'\s+', '', new_firstline)
-          reorder_span = possible_reorder_span
-          break
+    if kind == _FORWARD_DECLARE_KIND:
+      namespace_reorder_spans = _GetNamespaceLevelReorderSpans(file_lines)
+      for namespace_prefix, possible_reorder_span in namespace_reorder_spans:
+        if (namespace_prefix and possible_reorder_span and
+            firstline.line.startswith(namespace_prefix)):
+          # Great, we can go into this reorder_span.  We also need to
+          # modify all-lines because this line doesn't need the
+          # namespace prefix anymore.  Make sure we can do that before
+          # succeeding.
+          new_firstline = _RemoveNamespacePrefix(firstline.line, namespace_prefix)
+          if new_firstline:
+            assert all_lines[first_contentful_line] == firstline.line
+            all_lines[first_contentful_line] = new_firstline
+            sort_key = re.sub(r'\s+', '', new_firstline)
+            reorder_span = possible_reorder_span
+            break
 
     # If that didn't work out, find a top-level reorder span to go into.
     if reorder_span is None:
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1324,12 +1324,12 @@ def _GetToplevelReorderSpans(file_lines):
   return good_reorder_spans
 
 
-def _GetFirstNamespaceLevelReorderSpan(file_lines):
-  """Returns the first reorder-span inside a namespace, if it's easy to do.
+def _GetNamespaceLevelReorderSpans(file_lines):
+  """Returns a list of reorder-spans inside namespaces, if it's easy to do.
 
   This routine is meant to handle the simple case where code consists
   of includes and forward-declares, and then a 'namespace
-  my_namespace'.  We return the reorder span of the inside-namespace
+  my_namespace'.  We return the reorder spans of the inside-namespace
   forward-declares, which is a good place to insert new
   inside-namespace forward-declares (rather than putting these new
   forward-declares at the top level).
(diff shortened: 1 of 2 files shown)


fix_includes.py
```diff
@@ -1339,87 +1339,159 @@ def _GetFirstNamespaceLevelReorderSpan(file_lines):
...
+        # one namespace so it is okay to assume that _GetNamespaceNames
+        # will always return multiple records.
+        namespaces = _GetNamespaceNames(line_info.line)
+        for namespace in namespaces[:-1]:
+          if not namespace:
+            namespace_prefixes.append('namespace { ')
+          else:
+            namespace_prefixes.append('namespace %s { ' % namespace)
+          namespace_depth += 1
+        else:
```
kimgr: for/else always piques my interest...

You say in the comment above that there will "always be more than one namespace" -- you strip off the last one, which means there should be at least one left. You have no `break` in the loop body, so this `else` leg will never execute, right?

kimgr: Or is the comment wrong?

fix_includes.py
```diff
@@ -1339,87 +1339,159 @@ def _GetFirstNamespaceLevelReorderSpan(file_lines):
...
+      else:
+        # We should have handled all the cases above!
+        assert False, ('unknown line-info type',
+                       _LINE_TYPES.index(line_info.type))
+  except:
+    # namespace detection could be tricky so take what we have
+    pass
+
+  # return a reverse sorted list so longest matches are checked first
+  return sorted(list(namespace_reorder_spans.items()), reverse=True)
```
kimgr: I think Python 3 will return a generator from `sorted`, not a list. So depending on how this is used outside, you may need to materialize a list around `sorted`. In that vein, do you really need to materialize the `items()`? I would expect `sorted` to work on a generator/sequence-like thing.

fix_includes.py
```diff
@@ -1944,7 +2017,9 @@ def _GetSymbolNameFromForwardDeclareLine(line):
   """
   iwyu_namespace_re = re.compile(r'namespace ([^{]*) { ')
   symbolname_re = re.compile(r'([A-Za-z0-9_]+)')
-  namespaces_in_line = iwyu_namespace_re.findall(line)
+  # turn anonymous namespaces into their proper symbol representation
```
kimgr: Google style is proper sentence casing and punctuation for comments, so "Turn ... symbol representation." with a closing period.

fix_includes.py
```diff
@@ -1944,7 +2017,9 @@ def _GetSymbolNameFromForwardDeclareLine(line):
...
+  namespaces_in_line = iwyu_namespace_re.findall(line.replace(
```
kimgr: I'm curious how this is used, I hope there's a test to demonstrate :)

fix_includes_test.py
```diff
@@ -1099,20 +1099,14 @@ class Bar;  // lines 8-8
...
-    """Tests that for 'confusing' code, we keep fwd-decl at the top."""
+  def testAddForwardDeclareInsideNamespaceWithIfDef(self):
+    """Tests that ifdef blocks are ignored when finding namespaces."""
     infile = """\
 // Copyright 2010
 
 #include "foo.h"
 
 class Bar;
-///+namespace ns {
```
kimgr: I get a little suspicious when existing tests are modified. Can you give me a brief description of what happened here?

fix_includes_test.py
```diff
@@ -1197,6 +1193,40 @@ class Bar;  // lines 5-5
...
+add_fwd_decl_inside_namespace_without_compact_endings should remove these lines:
+- namespace ns { namespace ns1 { namespace ns2 { namespace ns3 { namespace ns4 { class Ns4Bye; } } } } } // lines 9-9
+
+The full include-list for add_fwd_decl_inside_namespace_without_compact_endings:
+namespace ns { namespace ns1 { namespace ns2 { class Ns2Bang; } } } // lines 4-4
+namespace ns { class NsBaz; } // lines 6-6
+---
+"""
+    self.RegisterFileContents({'add_fwd_decl_inside_namespace_without_compact_endings':
+                                 infile})
```
kimgr: Indentation looks off here

fix_includes_test.py
```diff
@@ -1229,13 +1258,248 @@ class NsFoo;
...
 namespace ns { template <typename T> class NsBaz; }
 namespace ns { namespace { template <typename T> class NsBar; } }  // lines 13-13
+namespace ns { namespace ns1 { class Ns1Foo; } }  // lines 18-18
+namespace ns { namespace ns1 { class Ns1Bar; } }
+namespace ns { namespace ns1 { class Ns1Baz; } }
 ---
 """
-    self.RegisterFileContents({'add_fwd_decl_inside_namespace_unnamed_ns':
+    self.RegisterFileContents({'add_fwd_decl_inside_allman_namespaces_with_existing_content':
+                                 infile})
```
kimgr: Indentation

fix_includes_test.py
```diff
@@ -1229,13 +1258,248 @@ class NsFoo;
...
+namespace ns { namespace ns1 { namespace ns2 { namespace { class NsaFoo; } } } } // lines 8-8
+namespace ns { namespace ns1 { namespace ns2 { namespace { template <typename T> class NsaBar; } } } } // lines 9-9
+namespace ns { namespace ns1 { namespace ns2 { namespace { class NsaBaz; } } } }
+namespace ns { namespace ns1 { namespace ns2 { namespace ns3 { class Ns3Bar; } } } }
+namespace ns { namespace ns1 { namespace ns2 { namespace ns3 { class Ns3Baz; } } } }
+namespace ns { namespace ns1 { namespace ns2 { namespace ns3 { class Ns3Foo; } } } } // lines 12-12
+---
+"""
+    self.RegisterFileContents({'add_fwd_decl_inside_mixed_namespaces_with_existing_content':
+                                 infile})
```
kimgr: Indentation

fix_includes_test.py
```diff
@@ -1229,13 +1258,248 @@ class NsFoo;
...
+namespace ns { class NsBang; }
+namespace ns { namespace ns1 { Ns1Bang; }
+namespace ns { namespace ns1 { namespace ns2 { class Ns2Bang; } } } // lines 4-4
+namespace ns { namespace ns1 { namespace ns2 { class Ns2Baz; } } }
+namespace ns { namespace ns1 { namespace ns2 { namespace ns3 { class Ns3Bang; } } } }
+namespace ns { namespace ns1 { namespace ns2 { namespace ns3 { class Ns3Baz; } } } } // lines 7-7
+---
+"""
+    self.RegisterFileContents({'add_fwd_decl_inside_nested_namespaces_and_top_level_for_complex_namespaces':
                                  infile})
```
kimgr: Indentation

kimgr:
Thank you! First round of comments inline.

fix_includes.py
kimgr: Sorry, no -- `sorted` can never return a generator, it (obviously) needs to materialize to sort. But the `list(namespace_reorder_spans.items())` looks unnecessary.

chri7325:
@kimgr sorry about the late response, I was out on vacation.  Thank you for the suggestions, I'm somewhat new to python so any education is appreciated.

fix_includes.py
chri7325: @kimgr a mixed namespace starts as a compact namespace but ends with an allman brace.  If there was only one namespace, it would be identified as either a regular or allman namespace which is handled elsewhere.  I used stack overflow to figure out how to treat the last element of a loop differently from https://stackoverflow.com/a/2429129.  Should I add a comment pointing to this?

fix_includes.py
chri7325: @kimgr this was actually a bug in this method.  Given an anonymous namespace, the regular expression would fail to properly symbolicate the namespace, leading to global symbol.  Using nm and c++filt on a class in an anonymous namespace shows the symbol namespace as (anonymous namespace).   The test to check this behavior is below where I include an anonymous namespace inside of a regular namespace and are still able to get the proper namespace span and include new forward declares where needed.

fix_includes.py
kimgr: And I would prefer the above, because if someone were to add a `break` in the loop body at some foggy point in the future, your necessary `else` would no longer run.

fix_includes.py
kimgr: Hmm, would it be nicer to extract that bugfix out into a separate commit? It would be much easier to see the change that way, with a smaller, more focused diff. Or did it only manifest with nested namespaces?

fix_includes_test.py
kimgr: Gotcha, thanks!

fix_includes_test.py
kimgr: I see now the cases that are misformatted on `master`. I'll update those and push, so you can rebase on something correct.

fix_includes.py
kimgr: Also, I still don't see any mention of `(anonymous namespace)` elsewhere in this patch or in `fix_includes_test.py`, so I'm still not convinced the normalization is covered correctly.

fix_includes_test.py
kimgr: Done.

fix_includes.py
```diff
@@ -1944,7 +2016,9 @@ def _GetSymbolNameFromForwardDeclareLine(line):
   """
   iwyu_namespace_re = re.compile(r'namespace ([^{]*) { ')
   symbolname_re = re.compile(r'([A-Za-z0-9_]+)')
-  namespaces_in_line = iwyu_namespace_re.findall(line)
+  # Turn anonymous namespaces into their proper symbol representation.
+  namespaces_in_line = iwyu_namespace_re.findall(line.replace(
```
chri7325: For this line, you had mentioned that there were no tests stressing this but if you remove this line and run the tests that I've written, you should see two failures where forward declares are not being added to anonymous namespaces, due to the regex not taking those into account, leading to a bug where a symbol is not able to be found.  This does not happen at global scope because the previous function only worked when the namespace was the first one found so I don't know if I can break this into a seperate commit with it's own test without the rest of the changes.  Thoughts?

kimgr: I had to twiddle with this for a while until I saw what was going on. `_GetSymbolNameFromForwardDeclareLine` is only used for uniquing. This reminds me of Clang's USR concept --
```
/**
 * Retrieve a Unified Symbol Resolution (USR) for the entity referenced
 * by the given cursor.
 *
 * A Unified Symbol Resolution (USR) is a string that identifies a particular
 * entity (function, class, variable, etc.) within a program. USRs can be
 * compared across translation units to determine, e.g., when references in
 * one translation refer to an entity defined in another translation unit.
 */
```
(from https://github.com/llvm-mirror/clang/blob/master/include/clang-c/Index.h#L4153).

Since the test has `NsBaz` in three different namespaces, when nested anonymous namespaces were omitted, the symbol names would come out identical.

Let's not worry about the separate commit, but I wonder if we could make it clearer that this symbol name is only used for deduplication and not for display or anything. I can try and remember to address this after merge. Let's leave this as-is.

fix_includes.py
```diff
@@ -1815,9 +1886,8 @@ def _DecoratedMoveSpanLines(iwyu_record, file_lines, move_span_lines, flags):
   if reorder_span is None:     # must be a new #include we're adding
     # If we're a forward-declare inside a namespace, see if there's a
     # reorder span inside the same namespace we can fit into.
-    if kind == _FORWARD_DECLARE_KIND:
```
kimgr: What happened to this check for `_FORWARD_DECLARE_KIND`? I'll admit I don't fully understand what's going on here... Does `_GetNamespaceLevelReorderSpans` somehow cover for this now?

fix_includes.py
```diff
@@ -1339,87 +1339,158 @@ def _GetFirstNamespaceLevelReorderSpan(file_lines):
...
+        namespace_reorder_spans[''.join(namespace_prefixes)] = (
+          line_number+1, line_number+1)
 
-    else:
-      # We should have handled all the cases above!
-      assert False, ('unknown line-info type',
-                     _LINE_TYPES.index(line_info.type))
+      elif line_info.type == _NAMESPACE_START_ALLMAN_RE:
+        for namespace in _GetNamespaceNames(line_info.line):
+          if not namespace:
```
kimgr: This conditional code keeps coming back. I wonder if you should define a local function to do this, something like:

    def _NamespaceDecl(nsname):
      namespace = ['namespace']
      if nsname:
        namespace += [nsname]
      return ' '.join(namespace)

That way the inline code would be simpler:

    for namespace in _GetNamespaceNames(line_info.line):
      namespace_prefix += _NamespaceDecl(namespace)

I'm not really happy with the name `_NamespaceDecl`, but I think this would flatten a lot of the already pretty complex code here.

fix_includes.py
```diff
@@ -1339,87 +1339,158 @@ def _GetFirstNamespaceLevelReorderSpan(file_lines):
...
-    if line_info.type in (_COMMENT_LINE_RE, _BLANK_LINE_RE, _INCLUDE_RE,
-                          _HEADER_GUARD_RE, _HEADER_GUARD_DEFINE_RE,
-                          _PRAGMA_ONCE_LINE_RE):
-      continue
+  namespace_reorder_spans = {}
+  try:
+    namespace_prefixes = []
+    namespace_prefix = ''
+    ifdef_depth = 0
+    namespace_depth = 0
```
kimgr: This variable is not necessary, it is always equal to `len(namespace_prefixes)`

fix_includes.py
```diff
@@ -1339,87 +1339,158 @@ def _GetFirstNamespaceLevelReorderSpan(file_lines):
...
+          line_number+1, line_number+1)
 
-    else:
-      # We should have handled all the cases above!
-      assert False, ('unknown line-info type',
-                     _LINE_TYPES.index(line_info.type))
+      elif line_info.type == _NAMESPACE_START_ALLMAN_RE:
+        for namespace in _GetNamespaceNames(line_info.line):
+          if not namespace:
+            namespace_prefix += 'namespace'
```
kimgr: At this point, it seems to be an invariant that `_GetNamespaceNames` only returns a single name... I wonder if you should assert that? Is `namespace_prefix` assumed to be empty here too, or could there be "residual" from a prior namespace, suffixed with `{`?

fix_includes.py
```diff
@@ -1339,87 +1339,158 @@ def _GetFirstNamespaceLevelReorderSpan(file_lines):
...
-    # header guard, which is (by definition) the only kind of #ifdef
-    # that we can be inside and still considered at the "top level".
-    if line_info.type in (_COMMENT_LINE_RE, _BLANK_LINE_RE, _INCLUDE_RE,
-                          _HEADER_GUARD_RE, _HEADER_GUARD_DEFINE_RE,
-                          _PRAGMA_ONCE_LINE_RE):
-      continue
+  namespace_reorder_spans = {}
+  try:
+    namespace_prefixes = []
+    namespace_prefix = ''
```
kimgr: I think this should be called `pending_namespace_prefix` to indicate that it's being built. It's a little confusing that sometimes we add to `namespace_prefixes`, sometimes to `namespace_prefix` and sometimes both :)

fix_includes.py
```diff
@@ -1339,87 +1339,158 @@ def _GetFirstNamespaceLevelReorderSpan(file_lines):
...
+
+      elif line_info.type == _FORWARD_DECLARE_RE:
+        # If we're not in a namespace, keep going.  Otherwise, this is
+        # just the situation we're looking for!  Update the dictionary
+        # with the better reorder span
+        if namespace_depth > 0:
+          namespace_reorder_spans[''.join(namespace_prefixes)] = (
+            line_info.reorder_span)
+
+      elif line_info.type == None:
```
kimgr: This should never happen, right? Let's remove this and crash if it does.

fix_includes.py
chri7325: This was actually an oversight on my part.  I must've removed it by accident.  It didn't change the way the script worked but adding it back should greatly reduce the calls made to _GetNamespaceLevelReorderSpans

fix_includes.py
chri7325: I'm not too sure about this one.  There's a lot of repeating if/else statements that are a bit hard to digest but every case needs to deal with anonymous namespaces as well as regular namespaces in all three types of situations.  In some cases, we can just add right to the namespace_prefixes while in others, we need to build the pending namespace prefix.  Some cases need to add the { while others need to wait for the actual bracket to add it.  I'd have to add an additional argument to have the bracket appended to the return type I believe.  I can still add it but I just don't know if it'll be cleaner in the end, given all the special cases.

fix_includes.py
chri7325: Good idea on the assert.  I'll reset the prefix here instead of waiting for the end to reset it.

fix_includes.py
```diff
@@ -1339,87 +1339,160 @@ def _GetFirstNamespaceLevelReorderSpan(file_lines):
...
+        if not namespaces[-1]:
+          pending_namespace_prefix += 'namespace'
+        else:
+          pending_namespace_prefix += 'namespace %s' % namespaces[-1]
+
+      elif line_info.type == _NAMESPACE_CONTINUE_ALLMAN_MIXED_RE:
+        # Append to the simplified allman namespace.
+        if pending_namespace_prefix == '':
+          raise FixIncludesError('Namespace bracket found without an associated '
+                                 'namespace name at line: %s', line_number)
```
kimgr: `line_info.line` or `line_number + 1`?

chri7325: I was going to make it line_info.lin but I started thinking that it would only print out `{` in most cases which might not be very helpful.  I opted for line number so you'd be able to see what line number the bracket in question would be.  I don't think I'd need a +1 since line 0 is marked as deleted on the script and just gets skipped over immediately so line_number actually corresponds to the actual line number (context shift for a computer scientist haha).

kimgr: Oh, I thought `line_info.line` *was* the line number, not the line content.

As for `line_number`, it's consistently used with +1 through the function, should that also be removed, then? (I suggested `enumerate(lines, start=1)` to get rid of this earlier.

chri7325: @kimgr the line_number+1 is actually meant to include the next line in the output as the possible reorder span.  For example
```
namespace A // Begin Allman if statement
{ // Begin continue bracket and add the next line as the possible reorder span, not this line
class Forwared_declared_class; // line_number+1 from line above as the possible reorder span
```
If I did enumerate with start=1, I'd be shifting the actual line numbers than from what the script initially did when it registered each line to a regex.  I'll add comments to the sections where I'm adding line_number+1.

fix_includes.py
```diff
@@ -1339,87 +1339,160 @@ def _GetFirstNamespaceLevelReorderSpan(file_lines):
...
-        return (namespace_prefix, line_info.reorder_span)
+      elif line_info.type == _ENDIF_RE:
+        ifdef_depth -= 1
+
+      elif ifdef_depth != 0:
+        continue # skip lines until we're outside of an ifdef block
+
+      # Build the simplified namespace lists and namespace depths.  When
+      # any new namespace is encountered, add the namespace to the list
+      # using the next line to cover namespaces without forward declares.
```
chri7325: Oh, actually, it looks like I did add documentation for this behavior.  Should I add a bit more detail?

chri7325: The documentation was actually a bit off as I changed patterns a bit after code review so I updated it.

fix_includes.py
```diff
@@ -1407,12 +1407,14 @@ def _GetNamespaceNames(namespace_line):
...
-      # to the list.  Once a contentful line (None) has been found, add
-      # the final namespace if we're still inside one and stop searching.
+      # Build the simplified namespace dictionary.  When any new namespace is
+      # encountered, add the namespace to the list using the next line to cover
+      # namespaces without forward declarations.  When a forward declare is
+      # found, update the dictionary using the existing namespace span that the
+      # forward declare contains.  Once a contentful line (None) has been found
+      # or any exception occurs, return the results that have been found.  Any
+      # forward declare that wasn't able to have a proper namespace name found
+      # will still propigate to the top of the file.
```
kimgr: Typo: propigate

fix_includes.py
kimgr: Oh, I see! I had misunderstood the `line_number + 1` as an adjustment from zero-based to one-based. Case closed, thanks!

kimgr:
Actually, let me fix up the typo and squash so we don't have to pay the timezone tax. Would be nice to get this on master before other `fix_includes` changes.

kimgr:
Merged as b290cf5915302f85e0cbf606e09eb1bc66bf6e18 with the following changes:

- Fixed typo above
- Stripped trailing whitespace in both fix_includes.py and fix_includes_test.py
- Squashed all commits into the first one, and cleaned up the commit message a little.

Thank you so much!

fix_includes.py
chri7325: I should turn on my spell checker plugin for python ;)

chri7325:
Thank you for the comments!  The final version is much clearer.

PR #572, opened 2018-07-14
Use clang-format style namespace comments for fix_includes.py

chri7325:
The fix_includes.py script inserts namespace comments for any new
namespace inclusions.  These comments have two spaces in between
closing bracket like so:

```
namespace ns {
}  // namespace ns
```

Modify the inclusions to match clang-format style comments for new
namespace inclusions like so:

```
namespace ns {
} // namespace ns
```

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1863,11 +1863,11 @@ def _NormalizeNamespaceForwardDeclareLines(lines):
      namespace bar {
      class A;
      class B;
-     }  // namespace bar
+     } // namespace bar
      namespace bang {
      class C;
-     }  // namespace bang
-     }  // namespace foo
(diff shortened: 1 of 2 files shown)


kimgr:
Is this true for all clang-format styles? I wouldn't want to break someone else's style :-/

chri7325:
@kimgr sorry about the late response, I just came back from vacation.  Looking at this again, this was actually a failed understanding on my part.  I had wrongfully assumed that setting `FixNamespaceComments: true` in clang format would automatically fix namespaces in the style I had on this PR.  It looks like this actually comes from `SpacesBeforeTrailingComments: 1` that we have set.  Closing this PR.  Apologies for the noise.

PR #574, opened 2018-07-20
Check for builtin templates

### LINKED ISSUE #150: Assertion failed: class_decl && "TemplateSpecializationType is not a TplSpecDecl?"
Originally reported on Google Code with ID 150

```
I'm trying to use iwyu with http://github.com/mapbox/mapbox-gl-native on OS X 10.9.
I compiled iwyu against a clang 3.4.2 final stable.

I'm running this command:

CXX="include-what-you-use -Xiwyu --verbose=3 -isystem\"`xcode-select -p`/Toolchains/XcodeDefault.xctoolchain/usr/lib/c++/v1\""
make linux

and the first command gyp generates is

  include-what-you-use -Xiwyu --verbose=3 -isystem"/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/c++/v1"
'-DNDEBUG' -I../../include -I../Release/obj/gen/include  -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk
-fasm-blocks -mpascal-strings -O3 -mmacosx-version-min=10.9 -arch x86_64 -std=c++11
-stdlib=libc++ -fvisibility-inlines-hidden -Wall -Wextra -I/usr/local/Cellar/libpng/1.6.10/include/libpng16
-I/usr/local/include -I/Users/kkaefer/Code/gl/native/mapnik-packaging/osx/out/build-cpp11-libcpp-universal/include
-MMD -MF ../Release/.deps/../Release/obj.target/llmr-x86/src/clipper/clipper.o.d.raw
 -c -o ../Release/obj.target/llmr-x86/src/clipper/clipper.o ../../src/clipper/clipper.cpp

which fails with

/Users/kkaefer/Code/include-what-you-use/iwyu.cc:3081: Assertion failed: class_decl
&& "TemplateSpecializationType is not a TplSpecDecl?"
make[1]: *** [../Release/obj.target/llmr-x86/src/clipper/clipper.o] Abort trap: 6

```

Reported by `mail@kkaefer.com` on 2014-06-17 14:38:03

jru:
Fixes #150. Detail comments in the commits.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3306,9 +3306,18 @@ class InstantiatedTemplateVisitor
       return true;
     }
 
-    const ClassTemplateSpecializationDecl* class_decl
-        = DynCastFrom(TypeToDeclAsWritten(type));
-    CHECK_(class_decl && "TemplateSpecializationType is not a TplSpecDecl?");
+    const NamedDecl* named_decl = TypeToDeclAsWritten(type);
+    const ClassTemplateSpecializationDecl* class_decl = DynCastFrom(named_decl);
+
+    // Bail out if we are not a proper class
+    if (class_decl == nullptr) {
+      // If the template specialization decl is not sugar for a class, we
+      // expect it to be another kind of template decl, like a built-in.
+      CHECK_(llvm::isa<clang::TemplateDecl>(named_decl))
+          << "TemplateSpecializationType has no decl of type TemplateDecl?";
+      return true;
+    }
+
     if (ContainsKey(traversed_decls_, class_decl))
       return true;   // avoid recursion & repetition
     traversed_decls_.insert(class_decl);
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -216,7 +216,6 @@ using clang::ValueDecl;
 using clang::VarDecl;
 using llvm::cast;
 using llvm::dyn_cast;
-using llvm::dyn_cast_or_null;
 using llvm::errs;
 using llvm::isa;
 using std::make_pair;
@@ -2490,8 +2489,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreCurrentASTNode())  return true;
     // ... only if this NewExpr involves a constructor call.
     const Expr* initializer = expr->getInitializer();
-    if (const CXXConstructExpr* cce =
-            dyn_cast_or_null<CXXConstructExpr>(initializer)) {
+    if (const CXXConstructExpr* cce = DynCastFrom(initializer)) {
       ReportIfReferenceVararg(cce->getArgs(),
                               cce->getNumArgs(),
                               cce->getConstructor());
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1045,7 +1045,11 @@ class AstFlattenerVisitor : public BaseAstVisitor<AstFlattenerVisitor> {
     NodeSet* node_set = &nodeset_decl_cache_[decl];
     if (node_set->empty()) {
       if (decl->isImplicit()) {
-        TraverseImplicitDeclHelper(dyn_cast_or_null<FunctionDecl>(decl));
+        // TODO: For now, it is only working for functions. Check if it could
+        // make sense for other implicit decls too (e.g. BuiltinTemplateDecl)
+        if (FunctionDecl* func = DynCastFrom(decl)) {
+          TraverseImplicitDeclHelper(func);
+        }
       } else {
         TraverseDecl(decl);
       }
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1120,16 +1120,27 @@ void ProcessFullUse(OneUse* use,
     }
   }
 
-  // (B3) Discard symbol uses for builtin symbols, including new/delete.
-  // TODO(csilvers): we could use getBuiltinID(), but it returns
-  // non-zero for things like malloc.  Figure out how to use it.
-  if (const FunctionDecl* fn_decl = DynCastFrom(use->decl())) {
-    if (StartsWith(use->symbol_name(), "__builtin_")) {
+  // (B3) Discard symbol uses for builtin symbols, including new/delete and
+  // template builtins.
+  if (isa<clang::BuiltinTemplateDecl>(use->decl())) {
+    VERRS(6) << "Ignoring use of " << use->symbol_name()
+             << " (" << use->PrintableUseLoc() << "): built-in template\n";
+    use->set_ignore_use();
+    return;
+  }
+  // A compiler builtin without a predefined header file (e.g. __builtin_..)
+  if (const clang::IdentifierInfo* iden = use->decl()->getIdentifier()) {
+    if (iden->getBuiltinID() != 0 &&
+        !clang::Builtin::Context::isBuiltinFunc(use->symbol_name().c_str())) {
       VERRS(6) << "Ignoring use of " << use->symbol_name()
                << " (" << use->PrintableUseLoc() << "): built-in function\n";
       use->set_ignore_use();
       return;
     }
+  }
+  // Special case for operators new/delete: Only treated as built-in if they
+  // are the default, non-placement versions.
+  if (const FunctionDecl* fn_decl = DynCastFrom(use->decl())) {
     const string dfn_file = GetFilePath(fn_decl);
     if (IsDefaultNewOrDelete(fn_decl, ConvertToQuotedInclude(dfn_file))) {
       VERRS(6) << "Ignoring use of " << use->symbol_name()
--- a/tests/cxx/built_ins_template.cc
+++ b/tests/cxx/built_ins_template.cc
@@ -1,0 +1,22 @@
+//===--- built_ins_template.cc - test input file for iwyu -----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Built-in templates have no declaration, thus they need no include.
+
+__type_pack_element<0, int> tp;
+
+template <class T, T...> struct A {};
+__make_integer_seq<A, int, 5> seq;
+
+
+/**** IWYU_SUMMARY
+
+(tests/cxx/built_ins_template.cc has correct #includes/fwd-decls)
+
+***** IWYU_SUMMARY */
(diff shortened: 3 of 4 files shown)


iwyu_output.cc
```diff
@@ -1123,17 +1123,22 @@ void ProcessFullUse(OneUse* use,
...
+  if (GetLocation(use->decl()).isInvalid()) {
+    VERRS(6) << "Ignoring use of " << use->symbol_name() << " ("
+             << use->PrintableUseLoc() << "): built-in symbol\n";
+    use->set_ignore_use();
+    return;
+  } else if (const FunctionDecl* fn_decl = DynCastFrom(use->decl())) {
     if (StartsWith(use->symbol_name(), "__builtin_")) {
-      VERRS(6) << "Ignoring use of " << use->symbol_name()
-               << " (" << use->PrintableUseLoc() << "): built-in function\n";
+      VERRS(6) << "Ignoring use of " << use->symbol_name() << " ("
```
kimgr: It would make the patch easier to parse if you didn't (let clang-format) reformat these unchanged lines.

tests/cxx/built_ins_template.cc
```diff
@@ -0,0 +1,20 @@
+//===--- built_ins_template.cc - test input file for iwyu -----------------===//
```
kimgr: "builtins" is usually written as one word, so maybe `builtin_templates.cc` instead?

tests/cxx/built_ins_template.cc
```diff
@@ -0,0 +1,20 @@
...
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+//Built-ins have no declaration, thus they need no include.
```
kimgr: Space after `//`

tests/cxx/built_ins_template.cc
```diff
@@ -0,0 +1,20 @@
...
+
+__type_pack_element<0, int> t;
+
+//TODO: Ignoring use of __type_pack_element ... built-in function
```
kimgr: Do you mean you'd like to check the IWYU output here? I wonder if it's possible at the high verbosity level it is right now... Maybe just skip the TODO.

iwyu.cc
```diff
@@ -1045,7 +1045,9 @@ class AstFlattenerVisitor : public BaseAstVisitor<AstFlattenerVisitor> {
     NodeSet* node_set = &nodeset_decl_cache_[decl];
     if (node_set->empty()) {
       if (decl->isImplicit()) {
-        TraverseImplicitDeclHelper(dyn_cast_or_null<FunctionDecl>(decl));
+        if (FunctionDecl* func = DynCastFrom(decl)) {
```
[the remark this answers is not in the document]
jru: As far as I could find, there really is nothing to traverse, since they are not defined in the source code. 
The BuiltinTemplateDecl contains only its ID and inherits TemplateDecl things, like parameters and so on.

iwyu_output.cc
jru: Yep. Sorry for that

iwyu_output.cc
```diff
@@ -1123,17 +1123,22 @@ void ProcessFullUse(OneUse* use,
   // (B3) Discard symbol uses for builtin symbols, including new/delete.
   // TODO(csilvers): we could use getBuiltinID(), but it returns
   // non-zero for things like malloc.  Figure out how to use it.
-  if (const FunctionDecl* fn_decl = DynCastFrom(use->decl())) {
+  if (GetLocation(use->decl()).isInvalid()) {
```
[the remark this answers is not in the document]
jru: True. But, I think that template builtins are special. They are not part of that table, I think. 
I will need to look into it, but it seems that checking for the dynamic type (BuiltinTemplateDecl) could be the only certain way to tell.

tests/cxx/built_ins_template.cc
jru: I was following the convention of the other two tests: `built_ins_no_includes.cc` and `built_ins_new_included.cc`. Personally, I prefer *builtins*, should I make them all consistent?

kimgr: Ah, I see. Let's leave it as-is for now, and we can follow up with a global rename later.

iwyu_output.cc
```diff
@@ -1120,22 +1120,32 @@ void ProcessFullUse(OneUse* use,
...
+  // (B3) Discard symbol uses for builtin symbols, including new/delete and
+  // template builtins.
+  if (isa<clang::BuiltinTemplateDecl>(use->decl())) {
+    VERRS(6) << "Ignoring use of " << use->symbol_name() << " ("
+             << use->PrintableUseLoc() << "): built-in template\n";
+    use->set_ignore_use();
+    return;
+  } else {
+    // A compiler builtin without a predefined header file (e.g. __builtin_..)
+    auto const iden = use->decl()->getIdentifier();
```
kimgr: Only use `auto` if the type is clearly visible on the line (or in extreme cases like `std::map` iteration :)). I think this should say `const IdentifierInfo* iden = use->decl->getIdentifier();`

iwyu_output.cc
```diff
@@ -1120,22 +1120,32 @@ void ProcessFullUse(OneUse* use,
...
-      VERRS(6) << "Ignoring use of " << use->symbol_name()
-               << " (" << use->PrintableUseLoc() << "): built-in new/delete\n";
+  // (B3) Discard symbol uses for builtin symbols, including new/delete and
+  // template builtins.
+  if (isa<clang::BuiltinTemplateDecl>(use->decl())) {
+    VERRS(6) << "Ignoring use of " << use->symbol_name() << " ("
+             << use->PrintableUseLoc() << "): built-in template\n";
+    use->set_ignore_use();
+    return;
+  } else {
```
kimgr: No need for `else` after return, here or below.

iwyu_output.cc
```diff
@@ -1120,22 +1120,32 @@ void ProcessFullUse(OneUse* use,
...
+    // A compiler builtin without a predefined header file (e.g. __builtin_..)
+    auto const iden = use->decl()->getIdentifier();
+    if (iden && iden->getBuiltinID() != 0 &&
+        !clang::Builtin::Context::isBuiltinFunc(use->symbol_name().c_str())) {
```
kimgr: This looks inverted -- shouldn't the `!` go away there? And does that mean there's a missing test?

iwyu_output.cc
```diff
@@ -1120,22 +1120,32 @@ void ProcessFullUse(OneUse* use,
...
+      VERRS(6) << "Ignoring use of " << use->symbol_name() << " ("
+               << use->PrintableUseLoc() << "): built-in function\n";
       use->set_ignore_use();
       return;
+      // Special case for operators new/delete: Only treated as built-in if they
+      // are the default, non-placement versions.
+    } else if (const FunctionDecl* fn_decl = DynCastFrom(use->decl())) {
+      const string dfn_file = GetFilePath(fn_decl);
+      if (IsDefaultNewOrDelete(fn_decl, ConvertToQuotedInclude(dfn_file))) {
```
kimgr: I think we should look into IsDefaultNewOrDelete and see if we can use some better technique to decide if it's placement new or not. But that can also wait for a follow-up.

tests/cxx/builtins_new_included.cc
```diff
@@ -1,4 +1,4 @@
-//===--- built_ins_new_included.cc - test input file for iwyu -------------===//
+//===--- builtins_new_included.cc - test input file for iwyu --------------===//
```
kimgr: Sorry I tricked you into this :)

Can you revert for now, and we can clean it up separately? Thanks

iwyu.cc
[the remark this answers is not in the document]
kimgr: Makes sense! Might be nice to add a comment to say that.

iwyu_output.cc
jru: I think it is a misnomer since in both cases they will be functions, but if `isBuiltinFunc` returns true the function is one of the special __builtin_* that have no header or declaration anywhere. That's why I check that it has a valid builtin ID but it isn't one of this special ones. I assume that means it is then a regular libc/libm function (possibly with associated header) like malloc. That is also the reason why the comment said before that checking whether getBuiltinID returns > 0 is not enough.

For more info look in clang/Basic/Builtins.def where they are [defined](https://github.com/llvm-mirror/clang/blob/a3ba348aeea36f4cd6155854364e088bfd66d14d/include/clang/Basic/Builtins.def#L76). 'F' attribute means special builtin, 'f' means regular function. Or so I understand it :)

Ideally, I would have used `isPredefinedLibFunction` which does exactly what I want but it requires us to get hold of the *Context* object from the *CompilerInstance* somehow. I thought it wouldn't be worth it to go that route, but I am open for suggestions!

iwyu_output.cc
jru: Yes, I agree.

iwyu.cc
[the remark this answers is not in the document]
jru: After experimenting for a while (trying to make it work for other decls) I went for the TODO comment.

jru:
Corrections done and some commits rephrased for clarity. Ready to merge if you are :)

iwyu.cc
```diff
@@ -3306,9 +3306,18 @@ class InstantiatedTemplateVisitor
...
-    const ClassTemplateSpecializationDecl* class_decl
-        = DynCastFrom(TypeToDeclAsWritten(type));
-    CHECK_(class_decl && "TemplateSpecializationType is not a TplSpecDecl?");
+    const NamedDecl* named_decl = TypeToDeclAsWritten(type);
+    const ClassTemplateSpecializationDecl* class_decl = DynCastFrom(named_decl);
+
+    // Bail out if we are not a proper class
+    if (class_decl == nullptr) {
+      // If the template specialization decl is not sugar for a class, we
+      // expect it to be other kind of template decl, like a built-in.
```
kimgr: other kind -> some other kind (or maybe "another kind")

iwyu_output.cc
kimgr: Either I'm misunderstanding your explanation or you've confused something above. As far as I can tell `isBuiltinFunc` returns true for any `f`-marked builtin, i.e. the ones that look like normal functions and require a header (e.g. memcpy). So your code is doing the right thing (if identifier has a non-zero builtin and name does not appear in libbuiltin-table, it must be a proper `__builtin_` builtin).

I think if we wanted to be fancy, we could do this lookup when reporting the use, and encode it as a use-flag, but I think this is fine as-is.

Issue #566, opened 2018-07-04
iwyu_tool and multiprocessing


pfaffe:
PR #444 introduced multiprocessing to iwyu_tool. Turns out this is problematic, since there's a bug (IMO) in python2.7's implementation: it swallows signals sent to the child. So if you hit ctrl-c, nothing happens (besides multiprocessing giving you the finger in the form of a nice stack trace).

I can think of four fixes, ordered by degree of radicalness:
- Bump the python version to 3
- Drop multiprocessing
- Use twisted
- Replace multiprocessing with a custom thread pool

Happy to file a PR, but would like to discuss which of these path's to go first.

pseyfert:
Sorry for that and thanks for reporting. Let's hear what the maintainers have to say. Obviously I have a preference against dropping multiprocessing. From what I googled, SO agrees with you that this is a bug in python. There are suggestions how to fix it within python 2.7 https://stackoverflow.com/a/35134329/ .

pfaffe:
The fix is worth a try, but looking at the multiprocessing code I don't see how it'll work: [process.py](https://github.com/python/cpython/blob/master/Lib/multiprocessing/process.py#L309). The catch-all except doesn't depend on the timeout or cache the signal.

kimgr:
A simple workaround is posted here: https://stackoverflow.com/a/2561809.

It doesn't appear to cover all scenarios, but maybe it's good enough for our purposes? A nice side quality is that it's portable, unlike explicit signal-handling.

pfaffe:
I'm not sure if they tested this solution, and it doesn't make much sense to me in general.

Multiprocesing wild-card catches all exceptions  except SystemExit, see the link I posted. As a hacky workaround we might catch KeyboardInterrupt and raise a SystemExit instead, but that seems icky and not incredibly stable.

kimgr:
It doesn't matter what multiprocessing does internally if the function it calls handles `KeyboardInterrupt` (unless signal handling is messed up, and maybe that's the root problem). It doesn't seem necessary to raise `SystemExit`, just returning from the function should be enough to terminate the worker process.

If you have a consistent repro, could you try it and report back?

Thanks!

pfaffe:
Multiprocessing essentially swallows the signal. The worker process observes and handles it, but the master doesn't terminate as you'd expect it to do, it just prints the stack trace.

[...]

PR #575, opened 2018-08-09
[iwyu_tool] Drop multiprocessing

pfaffe:
A potential fix for #566.

Drop multiprocessing, and instead manage the iwyu processes manually. iwyu outputs are redirected into temporary files and parsed upon completion.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -115,15 +115,25 @@ def find_include_what_you_use():
 IWYU_EXECUTABLE = find_include_what_you_use()
 
 class Process(object):
+    """ Manages an IWYU process in flight """
     def __init__(self, proc, outfile):
         self.proc = proc
         self.outfile = outfile
         self.output = None
 
     def poll(self):
+        """
+        Returns None if the IWYU process is still running,
+        the return code otherwise
+        """
         return self.proc.poll()
 
     def get_output(self):
+        """
+        Return the standard output of the IWYU process.
+
+        This call blocks until the process is complete, then returns the output.
+        """
         if not self.output:
             self.proc.wait()
             self.outfile.seek(0)
@@ -133,10 +143,11 @@ class Process(object):
         return self.output
 
     @classmethod
-    def start(cls, invocation, verbose):
-        if verbose:
-            print('# %s' % invocation)
-
+    def start(cls, invocation):
+        """
+        Start an IWYU process for the invocation, and return the Process
+        wrapper for it.
+        """
         outfile = tempfile.TemporaryFile(prefix='iwyu')
         process = subprocess.Popen(
             invocation.command,
@@ -176,12 +187,12 @@ class Invocation(object):
         return cls(command, entry['directory'])
 
 
-    def run(self, verbose):
+    def start(self, verbose):
         """ Run invocation and collect output. """
         if verbose:
             print('# %s' % self)
 
-        return Process.start(self, verbose)
+        return Process.start(self)
 
 
 def parse_compilation_db(compilation_db_path):
@@ -228,7 +239,7 @@ def slice_compilation_db(compilation_db, selection):
 def execute(invocations, verbose, formatter, jobs):
     if jobs == 1:
         for invocation in invocations:
-            print(formatter(invocation.run(verbose).get_output()))
+            print(formatter(invocation.start(verbose).get_output()))
         return
 
     pending = []
@@ -240,16 +251,16 @@ def execute(invocations, verbose, formatter, jobs):
             # handle iwyu output
             pending.remove(proc)
             print(formatter(proc.get_output()))
-            proc.close()
 
         n = jobs - len(pending)
-        pending.extend(e.run(verbose) for e in invocations[:n])
+        pending.extend(i.start(verbose) for i in invocations[:n])
         invocations = invocations[n:]
 
-        time.sleep(0) # yield cpu
+        time.sleep(0.0001) # yield cpu
 
 
-def main(compilation_db_path, source_files, verbose, formatter, jobs, iwyu_args):
+def main(compilation_db_path, source_files, verbose, formatter, jobs,
+         iwyu_args):
     """ Entry point. """
 
     try:
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -26,8 +26,10 @@ import os
 import re
 import sys
 import json
+import time
 import shlex
 import argparse
+import tempfile
 import subprocess
 import multiprocessing
 
@@ -112,6 +114,36 @@ def find_include_what_you_use():
 
 IWYU_EXECUTABLE = find_include_what_you_use()
 
+class Process(object):
+    def __init__(self, proc, outfile):
+        self.proc = proc
+        self.outfile = outfile
+        self.output = None
+
+    def poll(self):
+        return self.proc.poll()
+
+    def get_output(self):
+        if not self.output:
+            self.proc.wait()
+            self.outfile.seek(0)
+            self.output = self.outfile.read().decode("utf-8")
+            self.outfile.close()
+
+        return self.output
+
+    @classmethod
+    def start(cls, invocation, verbose):
+        if verbose:
+            print('# %s' % invocation)
+
+        outfile = tempfile.TemporaryFile(prefix='iwyu')
+        process = subprocess.Popen(
+            invocation.command,
+            cwd=invocation.cwd,
+            stdout=outfile,
+            stderr=subprocess.STDOUT)
+        return cls(process, outfile)
 
 class Invocation(object):
     """ Holds arguments of an IWYU invocation. """
@@ -143,17 +175,13 @@ class Invocation(object):
         command = [IWYU_EXECUTABLE] + extra_args + compile_args
         return cls(command, entry['directory'])
 
+
     def run(self, verbose):
         """ Run invocation and collect output. """
         if verbose:
             print('# %s' % self)
 
-        process = subprocess.Popen(
-            self.command,
-            cwd=self.cwd,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.STDOUT)
-        return process.communicate()[0].decode('utf-8')
+        return Process.start(self, verbose)
 
 
 def parse_compilation_db(compilation_db_path):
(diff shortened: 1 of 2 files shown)


pfaffe:
I'm taking suggestions how to test this :)

kimgr:
As chance would have it, we don't have a test plan for iwyu_tool.py. It started out as a minimal script to map compilation databases to IWYU invocations, and then it just grew.

I don't know if we should reconsider at some point, but for now I'm fine with having this tested by downstream users.

iwyu_tool.py
```diff
@@ -176,19 +178,26 @@ def main(compilation_db_path, source_files, verbose, formatter, jobs, iwyu_args)
 
     # Run analysis
     try:
+        pending = []
+        while entries or pending:
```
kimgr: I think this could do with a little more abstraction to separate the process scheduling from the IWYU details. I have a half-baked `ProcessPool` locally, I can try and clean that up and post it. Thanks for showing me a basic approach, though, that helped me get somewhere with what I had!

pfaffe: Since the whole thing was 20 LOC, i figured i'd just put it inline.

kimgr: Yeah, I think that makes sense. My attempt at this turned into an almost complete rewrite of `iwyu_tool.py`, and that's a little too revolutionary for my taste. Let me get a few refactors in to make your patch easier.

kimgr:
Hey @pfaffe, I just merged a series of refactors to `iwyu_tool.py`. The current state on master should make it pretty easy to get your change in place without the need for temporary files.

(While testing the process pooling in my Linux/Python2 environment I noticed that `time.sleep(0)` didn't yield, I needed at least `time.sleep(0.01)` for Python to back off. I found similar reports on Stack Overflow even if the general advice seems to be zero is enough.)

pfaffe:
I chose temporary files over pipes because pipes can deadlock if the output is large. On linux today the pipe buffer is 64KB, which probably is more than enough to hold the output of iwyu, but didn't like 'probably' :)

pfaffe:
Rebased onto master.

kimgr:
> I chose temporary files over pipes because pipes can deadlock if the output is large.
> On linux today the pipe buffer is 64KB, which probably is more than enough to hold
> the output of iwyu, but didn't like 'probably' :)

I was under the impression that `communicate` would solve that, see the note here: https://docs.python.org/2/library/subprocess.html#using-the-subprocess-module

pfaffe:
communicate does this, but it's synchronous, which is not what I want here.

kimgr:
Oh, so there's still a "race" -- if we `poll` until the process is completed and *then* call `communicate`, it's too late to save the potential deadlock?

pfaffe:
Yes. Should the pipe buffers run full the iwyu process will get stuck.

pfaffe:
Technically we could probably implement an asynchronous `communicate()`, but until the temporary file solution shows up in a profile I don't care enough :)

kimgr:
Yeah, I was mostly looking to get rid of the additional complexity of a temp file, but now I see why it's necessary, thanks!

iwyu_tool.py
```diff
@@ -197,6 +214,30 @@ def slice_compilation_db(compilation_db, selection):
 
     return new_db
 
+def process_invocations(invocations, verbose, formatter, jobs):
+    if jobs == 1:
```
kimgr: This special case seems unnecessary to me -- the algorithm below already handles that just fine.

iwyu_tool.py
```diff
@@ -197,6 +214,30 @@ def slice_compilation_db(compilation_db, selection):
...
+            # handle iwyu output
+            pending.remove(proc)
+            print(formatter(proc.get()))
+            proc.close()
+
+        n = jobs - len(pending)
+        pending.extend(e.run_iwyu(verbose) for e in invocations[:n])
+        invocations = invocations[n:]
+
+        from time import sleep
```
kimgr: Imports at the top, please

iwyu_tool.py
```diff
@@ -197,6 +214,30 @@ def slice_compilation_db(compilation_db, selection):
...
+        sleep(0) # yield cpu
```
kimgr: In my testing, I haven't seen `sleep(0)` yield -- the Python process would spin like crazy polling the pending processes. I'll re-run my tests with this patch to make sure I wasn't seeing something else. But I think you need to increase this sleep to yield effectively.

kimgr: Can't reproduce now, this seems to work as expected.

kimgr: Ah, this was because the special-casing for `jobs == 1` never executed the process pool. When I remove that, I'm back to 100% CPU for python.

iwyu_tool.py
```diff
@@ -107,12 +107,33 @@ def find_include_what_you_use():
         if os.path.isfile(full):
             return full
 
-    return None
-
 
 IWYU_EXECUTABLE = find_include_what_you_use()
 
 
+class InvocationResult(object):
```
kimgr: I experimented with a variation on this design -- I don't think `InvocationResult` conveys that this is a process in flight. Something like the below works much better for my brain:

```
class Process(object):
    def __init__(self, proc, outfile):
        self.proc = proc
        self.outfile = outfile
        self.output = None

    def poll(self):
        return self.proc.poll()

    def get_output(self):
        if not self.output:
            self.proc.wait()
            self.outfile.seek(0)
            self.output = self.outfile.read().decode("utf-8")
            self.outfile.close()

        return self.output

    @classmethod
    def start(cls, invocation, verbose):
        if verbose:
            print('# %s' % invocation)

        outfile = tempfile.TemporaryFile(prefix='iwyu')
        process = subprocess.Popen(
            invocation.command,
            cwd=invocation.cwd,
            stdout=outfile,
            stderr=subprocess.STDOUT)
        return cls(process, outfile)
```

This:
- removes the mutual dependency between this class and `Invocation`
- makes `poll` return the exit code if available -- at some point we may want to use that
- Renames get -> get_output and have it cache the resulting content and close the file

Rather than jumping hoops through `Invocation.run_iwyu`, I just had the processing loop call `Process.start(invocation, verbose)` for every invocation eligible for start.

iwyu_tool.py
```diff
@@ -107,12 +107,33 @@ def find_include_what_you_use():
...
+    def __init__(self, invocation):
+        from tempfile import TemporaryFile
```
kimgr: Imports at the top of the file, and prefer module imports.

kimgr:
Comments inline, please rebase on latest master.

iwyu_tool.py
pfaffe: It's not, it actually does something different. Instead of busy-waiting, it waits on the process to complete directly. Also it avoids all the queue shuffling and state checking required below.

iwyu_tool.py
```diff
@@ -143,17 +175,13 @@ def from_compile_command(cls, entry, extra_args):
         command = [IWYU_EXECUTABLE] + extra_args + compile_args
         return cls(command, entry['directory'])
 
+
     def run(self, verbose):
```
kimgr: This method can be deleted now (the verbose print currently happens twice)

iwyu_tool.py
```diff
@@ -197,41 +225,31 @@ def slice_compilation_db(compilation_db, selection):
...
-                    callback=lambda x: print(formatter(x))))
-        pool.close()
-        pool.join()
-        for r in results:
-            r.get()
-    except OSError as why:
-        print('ERROR: Failed to launch include-what-you-use: %s' % why)
-        return 1
+        n = jobs - len(pending)
+        pending.extend(e.run(verbose) for e in invocations[:n])
```
kimgr: Call `Process.start` here instead, it will be more obvious that that returns a Process object.

iwyu_tool_test.py
```diff
@@ -19,6 +19,23 @@
 except ImportError:
     from io import StringIO
 
+class MockProcess(object):
+    def __init__(self, block, content):
+        self.block = block
+        self.content = content
+    def poll(self):
```
kimgr: Blank line before here

iwyu_tool_test.py
```diff
@@ -19,6 +19,23 @@
...
+        if self.block > 0:
+            from time import sleep
+            sleep(self.block)
+        return 0
+
+    def get_output(self):
+        self.poll()
+        return self.content
+
+    def close(self):
```
kimgr: This is no longer used

iwyu_tool_test.py
```diff
@@ -19,6 +19,23 @@
 except ImportError:
     from io import StringIO
 
+class MockProcess(object):
+    def __init__(self, block, content):
+        self.block = block
+        self.content = content
+    def poll(self):
+        if self.block > 0:
+            from time import sleep
```
kimgr: Imports at the top

iwyu_tool.py
```diff
@@ -197,41 +225,31 @@ def slice_compilation_db(compilation_db, selection):
 
     return new_db
 
+def execute(invocations, verbose, formatter, jobs):
+    if jobs == 1:
```
kimgr: Please cut this block, it's not necessary as mentioned earlier.

iwyu_tool.py
kimgr: Oh, I see -- there's no seam for mocking if you do. Hmm. 

I'd like to see the following changes, then:

- Rename `run` -> `start`
- Rename `e` -> `i` for "invocation"

iwyu_tool.py
```diff
@@ -197,41 +225,31 @@ def slice_compilation_db(compilation_db, selection):
...
-        return 1
+        n = jobs - len(pending)
+        pending.extend(e.run(verbose) for e in invocations[:n])
+        invocations = invocations[n:]
 
-    return 0
+        time.sleep(0) # yield cpu
 
 
-def main(compilation_db_path, source_files, verbose, formatter, jobs,
```
kimgr: This line-wrap is kind-of ugly, but we try hard to stay within 80 char width, so please leave it alone.

iwyu_tool_test.py
kimgr: Also, this implementation seems weird -- `poll` should never block. Maybe it would be more "true" to keep track of time and return None until `self.block` has elapsed?

iwyu_tool.py
```diff
@@ -197,41 +225,31 @@ def slice_compilation_db(compilation_db, selection):
...
-    try:
-        pool = multiprocessing.Pool(jobs)
-        # No actual results in `results`, it's only used for exception handling.
-        # Details here: https://stackoverflow.com/a/28660669.
-        results = []
+        for proc in complete:
+            # handle iwyu output
+            pending.remove(proc)
+            print(formatter(proc.get_output()))
+            proc.close()
```
kimgr: Method no longer exists.

iwyu_tool.py
```diff
@@ -197,41 +225,31 @@ def slice_compilation_db(compilation_db, selection):
...
-            r.get()
-    except OSError as why:
-        print('ERROR: Failed to launch include-what-you-use: %s' % why)
-        return 1
+        n = jobs - len(pending)
+        pending.extend(e.run(verbose) for e in invocations[:n])
+        invocations = invocations[n:]
 
-    return 0
+        time.sleep(0) # yield cpu
```
kimgr: This doesn't work on my Ubuntu system at least. Needs to be a longer sleep to actually yield.

iwyu_tool.py
```diff
@@ -112,6 +114,36 @@ def find_include_what_you_use():
 
 IWYU_EXECUTABLE = find_include_what_you_use()
 
+class Process(object):
+    def __init__(self, proc, outfile):
```
kimgr: Add a doc string here and on all methods.

iwyu_tool.py
kimgr: Oops, I see now why it can't -- see below.

kimgr: So maybe the verbose print should happen here and not in `Process.start`.

iwyu_tool.py
pfaffe: I don't want busy-waiting when there's only a single job in flight though.

iwyu_tool.py
pfaffe: You're giving me mixed signals here ;)

kimgr: About what? I've (tried to) note several times that `sleep(0)` does not yield. When I run this script with `top` in another terminal, Python is competing with `include-what-you-use` for CPU. I wish there was a way in Python to just yield, but it appears to need a non-zero wait.

kimgr: It occurred to me that maybe you could use `subprocess.Popen.wait` instead of `poll` and provide a zero timeout. If that doesn't work, increase the timeout slightly. Waiting for processes is usually blocking.

kimgr:
Merged as dbab920 after some additional cleanup. Thank you!

PR #578, opened 2018-08-13
Treat explicit instantiations as full uses

jru:
Modify IsFowardDecl() to check that the decl is not an explicit template instantiation (declaration or definition) This check is also in its own function, IsExplicitInstantiation(), for convenience.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -109,6 +109,7 @@
 #include "iwyu_location_util.h"
 #include "iwyu_output.h"
 #include "iwyu_path_util.h"
+#include "iwyu_use_flags.h"
 // This is needed for
 // preprocessor_info().PublicHeaderIntendsToProvide().  Somehow IWYU
 // removes it mistakenly.
@@ -1632,12 +1633,13 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // Checkers, that tell iwyu_output about uses of symbols.
   // We let, but don't require, subclasses to override these.
 
-  // The comment, if not nullptr, is extra text that is included along
-  // with the warning message that iwyu emits.
+  // The comment, if not nullptr, is extra text that is included along with
+  // the warning message that iwyu emits. The extra use flags is optional
+  // info that can be assigned to the use (see the UF_* constants)
   virtual void ReportDeclUse(SourceLocation used_loc,
                              const NamedDecl* used_decl,
-                             const char* comment = nullptr) {
-
+                             const char* comment = nullptr,
+                             UseFlags extra_use_flags = 0) {
     const NamedDecl* target_decl = used_decl;
 
     // Sometimes a shadow decl comes between us and the 'real' decl.
@@ -1649,12 +1651,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreDecl(target_decl))
       return;
 
+    const UseFlags use_flags =
+        ComputeUseFlags(current_ast_node()) | extra_use_flags;
+
     // Canonicalize the use location and report the use.
     used_loc = GetCanonicalUseLocation(used_loc, target_decl);
     const FileEntry* used_in = GetFileEntry(used_loc);
     preprocessor_info().FileInfoFor(used_in)->ReportFullSymbolUse(
-        used_loc, target_decl, ComputeUseFlags(current_ast_node()),
-        comment);
+        used_loc, target_decl, use_flags, comment);
 
     // Sometimes using a decl drags in a few other uses as well:
 
@@ -1672,8 +1676,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         = GetUsingDeclarationOf(used_decl, 
               GetDeclContext(current_ast_node()))) {
       preprocessor_info().FileInfoFor(used_in)->ReportUsingDeclUse(
-          used_loc, using_decl, ComputeUseFlags(current_ast_node()),
-          "(for using decl)");
+          used_loc, using_decl, use_flags, "(for using decl)");
     }
 
     // For typedefs, the user of the type is sometimes the one
@@ -2910,19 +2913,20 @@ class InstantiatedTemplateVisitor
   // (if templates call other templates, we have to find the right
   // template).
   void ReportDeclUse(SourceLocation used_loc, const NamedDecl* decl,
-                     const char* comment = nullptr) override {
+                     const char* comment = nullptr,
+                     UseFlags extra_use_flags = 0) override {
     const SourceLocation actual_used_loc = GetLocOfTemplateThatProvides(decl);
     if (actual_used_loc.isValid()) {
       // If a template is responsible for this decl, then we don't add
       // it to the cache; the cache is only for decls that the
       // original caller is responsible for.
-      Base::ReportDeclUse(actual_used_loc, decl, comment);
+      Base::ReportDeclUse(actual_used_loc, decl, comment, extra_use_flags);
     } else {
       // Let all the currently active types and decls know about this
       // report, so they can update their cache entries.
       for (CacheStoringScope* storer : cache_storers_)
         storer->NoteReportedDecl(decl);
-      Base::ReportDeclUse(caller_loc(), decl, comment);
+      Base::ReportDeclUse(caller_loc(), decl, comment, extra_use_flags);
     }
   }
 
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3732,11 +3732,20 @@ class IwyuAstConsumer
   //    template <class T> struct Foo;
   //    template<> struct Foo<int> { ... };
   // we don't want iwyu to recommend removing the 'forward declare' of Foo.
+  //
+  // Additionally, this type of decl is also used to represent explicit template
+  // instantiations, in which case we want the full type, not only a forward
+  // declaration.
   bool VisitClassTemplateSpecializationDecl(
       clang::ClassTemplateSpecializationDecl* decl) {
     if (CanIgnoreCurrentASTNode())  return true;
     ClassTemplateDecl* specialized_decl = decl->getSpecializedTemplate();
-    ReportDeclForwardDeclareUse(CurrentLoc(), specialized_decl);
+
+    if (IsExplicitInstantiation(decl))
+      ReportDeclUse(CurrentLoc(), specialized_decl);
+    else
+      ReportDeclForwardDeclareUse(CurrentLoc(), specialized_decl);
+
     return Base::VisitClassTemplateSpecializationDecl(decl);
   }
 
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -217,6 +217,7 @@ using clang::ValueDecl;
 using clang::VarDecl;
 using llvm::cast;
 using llvm::dyn_cast;
+using llvm::dyn_cast_or_null;
 using llvm::errs;
 using llvm::isa;
 using std::make_pair;
@@ -2832,7 +2833,8 @@ class InstantiatedTemplateVisitor
 
     // As in TraverseExpandedTemplateFunctionHelper, we ignore all AST nodes
     // that will be reported when we traverse the uninstantiated type.
-    if (const NamedDecl* type_decl_as_written = TypeToDeclAsWritten(type)) {
+    if (const NamedDecl* type_decl_as_written =
+            GetDefinitionAsWritten(TypeToDeclAsWritten(type))) {
       AstFlattenerVisitor nodeset_getter(compiler());
       nodes_to_ignore_ = nodeset_getter.GetNodesBelow(
           const_cast<NamedDecl*>(type_decl_as_written));
@@ -3111,7 +3113,49 @@ class InstantiatedTemplateVisitor
     // We attribute all uses in an instantiated template to the
     // template's caller.
     ReportTypeUse(caller_loc(), actual_type);
+
+    // Also report all previous explicit instantiations (declarations and
+    // definitions) as uses of the caller's location.
+    ReportExplicitInstantiations(actual_type);
+
     return Base::VisitSubstTemplateTypeParmType(type);
+  }
+
+  bool VisitTemplateSpecializationType(TemplateSpecializationType* type) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+
+    CHECK_(current_ast_node()->GetAs<TemplateSpecializationType>() == type)
+        << "The current node must be equal to the template spec. type";
+
+    // Report previous explicit instantiations here, only if the type is needed
+    // fully.
+    if (!CanForwardDeclareType(current_ast_node()))
+      ReportExplicitInstantiations(type);
+
+    return Base::VisitTemplateSpecializationType(type);
+  }
+
+  void ReportExplicitInstantiations(const Type* type) {
+    const auto* decl = dyn_cast_or_null<ClassTemplateSpecializationDecl>(
+        TypeToDeclAsWritten(type));
+
+    if (decl == nullptr)
+      return;
+
+    // Go through all previous redecls and filter out those that are not
+    // explicit template instantiations or already provided by the template.
+    for (const NamedDecl* redecl : decl->redecls()) {
+      if (!IsExplicitInstantiation(redecl) ||
+          !GlobalSourceManager()->isBeforeInTranslationUnit(
+              redecl->getLocation(), caller_loc()) ||
+          IsProvidedByTemplate(decl))
+        continue;
+
+      // Report the specific decl that points to the explicit instantiation
+      Base::ReportDeclUse(caller_loc(), redecl, "(for explicit instantiation)",
+                          UF_ExplicitInstantiation);
+    }
   }
 
   // If constructing an object, check the type we're constructing.
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3207,7 +3207,7 @@ class InstantiatedTemplateVisitor
          ast_node != caller_ast_node_; ast_node = ast_node->parent()) {
       if (preprocessor_info().PublicHeaderIntendsToProvide(
               GetFileEntry(ast_node->GetLocation()),
-              GetFileEntry(decl)))
+              GetFileEntry(decl->getLocation())))
         return ast_node->GetLocation();
     }
     return SourceLocation();   // an invalid source-loc
@@ -3219,8 +3219,10 @@ class InstantiatedTemplateVisitor
   bool IsProvidedByTemplate(const Type* type) const {
     type = RemoveSubstTemplateTypeParm(type);
     type = RemovePointersAndReferences(type);  // get down to the decl
-    if (const NamedDecl* decl = TypeToDeclAsWritten(type))
+    if (const NamedDecl* decl = TypeToDeclAsWritten(type)) {
+      decl = GetDefinitionAsWritten(decl);
       return GetLocOfTemplateThatProvides(decl).isValid();
+    }
     return true;   // we always provide non-decl types like int, etc.
   }
 
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -105,6 +105,7 @@ using clang::TemplateArgumentLoc;
 using clang::TemplateDecl;
 using clang::TemplateName;
 using clang::TemplateParameterList;
+using clang::TemplateSpecializationKind;
 using clang::TemplateSpecializationType;
 using clang::TranslationUnitDecl;
 using clang::Type;
@@ -153,6 +154,13 @@ void DumpASTNode(llvm::raw_ostream& ostream, const ASTNode* node) {
   } else {
     CHECK_UNREACHABLE_("Unknown kind for ASTNode");
   }
+}
+
+TemplateSpecializationKind GetTemplateSpecializationKind(const Decl* decl) {
+  if (const auto* record = dyn_cast<CXXRecordDecl>(decl)) {
+    return record->getTemplateSpecializationKind();
+  }
+  return clang::TSK_Undeclared;
 }
 
 }  // anonymous namespace
@@ -942,12 +950,20 @@ bool IsFriendDecl(const Decl* decl) {
   return decl->getFriendObjectKind() != Decl::FOK_None;
 }
 
+bool IsExplicitInstantiation(const clang::Decl* decl) {
+  TemplateSpecializationKind kind = GetTemplateSpecializationKind(decl);
+  return kind == clang::TSK_ExplicitInstantiationDeclaration ||
+         kind == clang::TSK_ExplicitInstantiationDefinition;
+}
+
 bool IsForwardDecl(const NamedDecl* decl) {
-  if (const auto* record_decl = dyn_cast<RecordDecl>(decl)) {
+  if (const auto* record_decl = dyn_cast<CXXRecordDecl>(decl)) {
+
     return (!record_decl->getName().empty() &&
             !record_decl->isCompleteDefinition() &&
             !record_decl->isEmbeddedInDeclarator() &&
-            !IsFriendDecl(record_decl));
+            !IsFriendDecl(record_decl) &&
+            !IsExplicitInstantiation(record_decl));
   }
 
   return false;
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -237,12 +237,12 @@ string GetShortNameAsString(const clang::NamedDecl* named_decl) {
 
 // Holds information about a single full or fwd-decl use of a symbol.
 OneUse::OneUse(const NamedDecl* decl, SourceLocation use_loc,
-               OneUse::UseKind use_kind, UseFlags flags,
-               const char* comment)
+               SourceLocation decl_loc, OneUse::UseKind use_kind,
+               UseFlags flags, const char* comment)
     : symbol_name_(internal::GetQualifiedNameAsString(decl)),
       short_symbol_name_(internal::GetShortNameAsString(decl)),
       decl_(decl),
-      decl_loc_(GetInstantiationLoc(GetLocation(decl))),
+      decl_loc_(GetInstantiationLoc(decl_loc)),
       decl_file_(GetFileEntry(decl_loc_)),
       decl_filepath_(GetFilePath(decl_file_)),
       use_loc_(use_loc),
@@ -577,15 +577,23 @@ void IwyuFileInfo::ReportFullSymbolUse(SourceLocation use_loc,
                                        UseFlags flags,
                                        const char* comment) {
   if (decl) {
-    // Since we need the full symbol, we need the decl's definition-site.
-    // But only if we're not defining the function, in which case we want to use
-    // all its preceding declarations, and not try to canonicalize.
-    if (!(flags & UF_FunctionDfn)) {
-      decl = GetDefinitionAsWritten(decl);
+    const NamedDecl* report_decl;
+    SourceLocation report_decl_loc;
+
+    if ((flags & (UF_FunctionDfn | UF_ExplicitInstantiation)) == 0) {
+      // Since we need the full symbol, we need the decl's definition-site too.
+      // Also, by default we canonicalize the location, using GetLocation.
+      report_decl = GetDefinitionAsWritten(decl);
+      report_decl_loc = GetLocation(report_decl);
+    } else {
+      // However, if we're defining the function or we are targeting an explicit
+      // instantiation, we want to use it as-is and not try to canonicalize at all.
+      report_decl = decl;
+      report_decl_loc = decl->getLocation();
     }
 
-    symbol_uses_.push_back(OneUse(decl, use_loc, OneUse::kFullUse,
-                                  flags, comment));
+    symbol_uses_.push_back(OneUse(report_decl, use_loc, report_decl_loc,
+                                  OneUse::kFullUse, flags, comment));
     LogSymbolUse("Marked full-info use of decl", symbol_uses_.back());
   }
 }
@@ -630,8 +638,8 @@ void IwyuFileInfo::ReportForwardDeclareUse(SourceLocation use_loc,
   // combines friend decls with true forward-declare decls.  If that
   // happened here, replace the friend with a real fwd decl.
   decl = GetNonfriendClassRedecl(decl);
-  symbol_uses_.push_back(OneUse(decl, use_loc, OneUse::kForwardDeclareUse,
-                                flags, comment));
+  symbol_uses_.push_back(OneUse(decl, use_loc, GetLocation(decl),
+                                OneUse::kForwardDeclareUse, flags, comment));
   LogSymbolUse("Marked fwd-decl use of decl", symbol_uses_.back());
 }
 
--- a/tests/cxx/explicit_instantiation.cc
+++ b/tests/cxx/explicit_instantiation.cc
@@ -1,0 +1,48 @@
+//===-------- explicit_instantiation.cc - test input file for iwyu --------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "explicit_instantiation-template_direct.h"
+
+// Test that all explicit instantiations variants of the base template
+// require the full type:
+
+// - Declaration and definition
+// IWYU: Template is...*explicit_instantiation-template.h
+extern template class Template<int>;
+// IWYU: Template is...*explicit_instantiation-template.h
+template class Template<int>;
+
+// - Only declaration
+// IWYU: Template is...*explicit_instantiation-template.h
+extern template class Template<short>;
+
+// - Only definition
+// IWYU: Template is...*explicit_instantiation-template.h
+template class Template<double>;
+
+
+// The explicit instantiation of a specialization only needs a declaration
+// of the base template
+// IWYU: Template needs a declaration
+template<> class Template<char> {};
+extern template class Template<char>;
+
+
+/**** IWYU_SUMMARY
+
+tests/cxx/explicit_instantiation.cc should add these lines:
+#include "explicit_instantiation-template.h"
+
+tests/cxx/explicit_instantiation.cc should remove these lines:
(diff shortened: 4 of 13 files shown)


jru:
Well, in this case the header file doesn't play any role, that's why I didn't add any. 

The result of the test previous to the change suggests the instantiation definition for the *ExplicitlyDeclaredAndDefined* template to get removed: 
```
tests/cxx/explicit_template_instantiation.cc should add these lines:

tests/cxx/explicit_template_instantiation.cc should remove these lines:
- class ExplicitlyDeclaredAndDefined;  // lines 32-32

The full include-list for tests/cxx/explicit_template_instantiation.cc:
```
Honestly, I haven't analyzed completely why the third declaration and other two end up being treated as full uses (I am guessing something in the semantics changes when one follows the other), but in any case we don't want them to be confused with forward declarations, so I made sure to keep them.

iwyu_ast_util.cc
```diff
@@ -942,12 +942,21 @@ bool IsFriendDecl(const Decl* decl) {
   return decl->getFriendObjectKind() != Decl::FOK_None;
 }
 
+bool IsExplicitInstantiation(const clang::CXXRecordDecl* decl) {
+  const auto kind = decl->getTemplateSpecializationKind();
```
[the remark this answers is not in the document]
jru: The name of the getter contains the enum type, *TemplateSpecializationKind*. That's why I thought it would be clear. I can repeat myself, no problem ;)

kimgr:
I think a more typical layout/use of explicit instantiations is to have the template definition and explicit instantiation decl in a header file, and the explicit instantiation defn in a .cpp file.

So rather than a minimized test case to provoke this bug, we could put up some representative code. Something like:

```c++
// explicit_template_instantiation-template.h

// Base template
template<class T>
class Template {};

// Explicit instantiation declaration.
extern template class Template<int>;


// explicit_template_instantiation-short.h

// Another explicit instantiation declaration.
extern template class Template<short>;

// explicit_template_instantiation.cc
#include "explicit_template_instantiation-template.h"
#include "explicit_template_instantiation-short.h"

// Define explicit instantiation
template class Template<int>;

// Some uses
void f() {
    // Use the base template
    Template<bool> t0;

    // Use the locally-defined explicit instantiation
    Template<int> t1;

    // Use the declared but not defined explicit instantiation
    // (I'm guessing this should type-check, but not link?)
    // Should make sure the -short.h header is not removed
    Template<short> t2;
}
```

I left out the `// IWYU` assertions in the interest of brevity (and because I'm not entirely sure what might happen). Overall this seems like a more realistic test case to me.

jru:
I played with it a bit more and I found some other cases where IWYU is not doing the right thing. 
Specifically, explicit instantiation declarations of templated functions seem to be ignored. 

This needs more work than previously thought..

kimgr:
I think we should leave function templates as a known issue to make some headway with what we have here.

jru:
Yes, I don't think they can be tackled the same way since they are not visited directly, unless `shouldVisitTemplateInstantiations()` returns true, and that breaks currently. I'll create an issue to address this later.

Class and Var templates are almost working, but I got stuck for a while. I think I could finalize this tomorrow.

jru:
I finally managed to finalize this, for class templates so far. 

Not sure if we can add something else to the test case. What you mentioned in your example above about verifying that a header file with a explicit instantiation declaration is included is not really possible, since such declaration is not necessary for the usage of template itself.

kimgr:
I think the tests would benefit from separation like I suggested earlier. Maybe you need to be explicit with  `--check_also` to get proper output from headers.

As for the `short` case, that's really the most interesting to me -- see the review feedback.

jru:
> We included a header saying there's a singular instantiation for short and if that include is removed, we'll implicitly instantiate it, generating more code. Or if the base template has no definition, we'll fail to compile.

I guess it makes sense. But, it'll need some work :)
So, what should be expected behaviour? Which uses of that template should make the decl stay? 

For example, looking at the 7 cases below. Should we keep the include for all of them, or only the uses of the same specialization? And only full uses, or also fwd decl uses? What about implicit uses?

```cpp
// file template.h
template<class>
class Template {};

// file template-decl-short.h
extern template class Template<short>;

//file main.cc
#include "template.h"
#include "template-decl-short.h"

template <class T> class FullUseArg { T t; }
template <class T> class FwdDeclUseArg { T* t; }
template <class T = Template<short>> class TemplateAsDefaultFull { T t; };
template <class T = Template<short>> class TemplateAsDefaultFwd { T* t; };

Template<short> s; // 1
template Template<short>; // 2
FullUseArg<Template<short>> fu; // 3
FwdDeclUseArg<Template<short>> fd; // 4
TemplateAsDefaultFull<> td; // 5
TemplateAsDefaultFwd<> td; // 6
Template<int> s; // 7

```

I would go for 1-3 and 5, but maybe that misses other cases. For example, if this is itself an include and there are only forward decl uses of the template, but it is then fully used somewhere else.

Also note, if this is implemented the include 'template-decl-short.h' will be required in all compilation units that use that specialization. Do we want that?

kimgr:
> Also note, if this is implemented the include 'template-decl-short.h' will be required in all compilation
> units that use that specialization. Do we want that?

I think so... Or you mean that it could be transitively satisfied, so the header doesn't need to be included everywhere? From a strict IWYU perspective it would be nice to explicitly include that specialization everywhere it's used because it makes the code resilient to refactors -- if the specialization include is removed from an incidental dependency, our code will silently change meaning (using the base template instead).

I find it hard to reason about the seven cases when they're all in the same file, but I'll give it a shot...

```
template <class T = Template<short>> class TemplateAsDefaultFull { T t; };
template <class T = Template<short>> class TemplateAsDefaultFwd { T* t; };
```

I'm not sure this is 100% consistent, but it seems to me IWYU's general rule is that the file that spells out the type should include the type. So template argument defaults should be attributed to the file declaring the template. I can't play this out in my head for the templates above, so not sure how forward-decl vs full uses should behave.

`Template<short> s; // 1`

This seems like it should obviously require `template-decl-short.h`.

`template Template<short>; // 2`

Explicit template instantiation definition of `Template<short>`, right? That should require `template-decl-short.h`, otherwise it will create a reusable instantiation of the wrong template.

`FullUseArg<Template<short>> fu; // 3`

`Template<short>` is spelled out here and `fu` constructor will need to see the full type, so `template-decl-short.h` required.

`FwdDeclUseArg<Template<short>> fd; // 4`

`Template<short>` is spelled out but constructing `fd` does not need the full type, so a forward-decl or explicit instantiation declaration of `Template<short>` should do here.

```
TemplateAsDefaultFull<> td; // 5
TemplateAsDefaultFwd<> td; // 6
```

These are tricky.

The general rule says we shouldn't need `Template<short>` because it's not spelled out. But it is required for constructing `td`. I think this is one of the author-intent rules -- `Template<short>` should be attributed to the file declaring `TemplateAsDefaultFull` unless it has a forward-declare of `Template<short>` to signal that users should be responsible. And I think it should always require the complete type of the default template arguments, because I can't get my head around the analysis of what might happen when the templates are instantiated.

`Template<int> s; // 7`

No use of `Template<short>`, so this doesn't count toward that header.

kimgr:
I also think it's hard to reason about this, because I keep confusing "explicit instantiation" with "explicit specialization".

Specializations, whether partial or total, alter the behavior of the program if visible to the compilation unit. 

Explicit instantiations, on the other hand, seem to be more of a build-time optimization: we can use them to help the compiler do separate compilation of templates.

The challenge in IWYU is that we want to preserve the behavior of the program *and* not worsen build times. So it would seem we can never remove a specialization or explicit instantiation unless we can prove that they're unused.

I'm pretty sure this is not helping :), I just needed to type it out to understand it better. I'm really out of my depth when it comes to template handling in IWYU.

jru:
I just added some work-in-progress that already satisifies 1 and 2. For the rest, there is more work to do.

Also something I noticed is that IWYU tries very hard to normalize decls and their locations. :) For usual template instantiations that's not a problem (as we usually want the definition), but for explicit instantiation decls, I really had to fight it so that it would leave the decl I pass in to *ReportDeclUse* alone. 

I am not happy with the result so far, so I would like to gather more input before I continue with the other cases.

jru:
FYI, I am still working on this. It turned out to be much harder than I thought, but I think I start to see the light at the end of the tunnel :)

jru:
I finally managed to finish this. There are a few more commits that in the beginning, so let me explain briefly (more details in the commits themselves):

*These are the three main changes to make explicit template instantiations work:*
- Treat explicit template instantiations as full uses
- Require explicit inst. decls to use main template
- Use the definition to check if decl is provided by template

*These contain some refactoring work that simplifies the implementation:*
- Make 'comment' a default argument in Report* functions
- Refactor CanIgnoreType for clarity and reuse

*Supporting change to be able to report specific redecls:*
- Introduce an extra use flag UF_TargetRedecl

*Unrelated bugfix discovered during the implementation:*
- Fix GetLocation for templated functions

*Also unrelated to my changes. Seems to be failing after a new warning from clang:*
- Silence warning -Wsizeof-pointer-div in test

kimgr:
Very cool! I haven't had a chance to look at the body of the template instantiation logic, but it strikes me that many of the refactors would be nice on their own. Can you separate them out and post them as individual PRs? The problem with huge PRs like this is that they're both hard to review and blocked until all commits are ready to go in, and much of this looks useful even before trying to fix the explicit instantiation problems.

I'll actually start off by cherry-picking the `-Wnosizeof-pointer` commit, to get master green again. I'll come back with what I think is a logical order of changes afterwards.

kimgr:
Is `Fix GetLocation for templated functions` as free-standing as it looks? It's nice when you can remove code to fix a broken test :) Is there a test that should be written to capture the new behavior?

If it doesn't depend on anything else, I think this would be nice to get on master first.

kimgr:
`Introduce an extra use flag UF_TargetRedecl`: I feel there's more than this going on in this commit.

It seems suspicious that `UF_FunctionDfn` and `UF_TargetRedecl` control the same behavior, but I intended for use-flags to say something about the context in which the use was detected, so if they describe different contexts, they should be different. That said, I'm not sure what `UF_TargetRedecl` says...?

Overall, I feel like if we could break this commit into its constituent parts, maybe there are fragments that could go in earlier and solve smaller problems than the elusive explicit template instantiation :)

There's a lot of spurious reformatting in this commit (and maybe others) -- please tread lightly; it's so much harder to review (and blame in the future) when lines are reformatted as part of a functional change.

jru:
Hey,

I extracted the two refactor commits here: #607. I also rebased this branch on top of it to be sure there are no conflicts. Once it is merged I'll rebase again on top of master.


> Introduce an extra use flag UF_TargetRedecl

You were right about the whitespace, I messed it up somehow. Should be fixed now! Also made some small changes to ReportFullSymbolUse to hopefully make it clearer.

> That said, I'm not sure what UF_TargetRedecl says...?

The main idea here was to pass extra information to the reporting logic that says "please, treat this decl as-is". I went back and forth trying out different ideas and this seemed like the cleanest solution. Although, not perfect by any means :)

Right now, the flag is not used inside OneUse for anything else, but I can imagine this might turn out to be useful for other tings. In fact, during the implementation I had to use the UF_TargetRedecl to reimplement `DeclCanBeForwardDeclared` as a member of OneUse, but later on turned out not to be necessary for this case, so I removed it to make the patch smaller.

The existing UF_FunctionDfn is used to implement `OneUse::is_function_being_defined` so I cannot really mix the two. That said, it might not be the right thing to do to add another flag for this purpose. I am open for ideas :)

In general, it seems to me the reporting side of IWYU `iwyu_output.*` shouldn't know anything about Clang or the AST and instead work exclusively based on the interface provided by OneUse (tiny as of right now), so that it becomes easier to extend and maintain in the long run.

jru:
> Is Fix GetLocation for templated functions as free-standing as it looks? 

Yes, it is. I dropped it from the branch and will try to think of a test for it.

kimgr:
> The existing UF_FunctionDfn is used to implement OneUse::is_function_being_defined so I cannot
> really mix the two. That said, it might not be the right thing to do to add another flag for this purpose. I 
> am open for ideas :)

Ah, I see. No, I think a new flag is probably warranted.

> In general, it seems to me the reporting side of IWYU iwyu_output.* shouldn't know anything about 
> Clang or the AST and instead work exclusively based on the interface provided by OneUse (tiny as of 
> right now), so that it becomes easier to extend and maintain in the long run.

I've actually been considering going in the opposite direction -- capture and expose parts of the AST and expose to `iwyu_output.*`.

Some things are hard to classify correctly when traversing the AST because they require cross-examination of multiple structures (can't think of an example now, but `UF_FunctionDfn` came about to solve one of those problems). The idea of use-flags is to capture some information about the use-location and forward it to later analysis where we have more context. I've even played with the idea of refactoring `UseFlags` into a `UseContext` class to forward e.g. AST fragments into `iwyu_output` (which I think is necessary to solve #105, for example).

So I'm not sure I share the concern that `iwyu_output.*` should not know anything about prior phases. And I think the new use-flag should say something about the context in which the use occurred, not what the earlier phase thinks the later phase should do with it.

jru:
Hi, refactors are in and branch rebased.

> I've actually been considering going in the opposite direction -- capture and expose parts of the AST and expose to iwyu_output.*.

Yes, that might work too. The main point I was trying to make is that the visiting side is lower level, and as such, it doesn't have the complete picture. However, it does have access to all of the details about the AST. The analysis (reporting/iwyu_output) side, on the other hand, is much higher level, with much more context, but has lost much of the detail from the lower phase.

The main problem, as I see it, is that this separation of concerns is not well defined or maintained in the code. For example, the visiting side tries to make high-level decisions about what things to report, whereas the analysis will very often overrule those decisions at a later point after doing some low-level visiting of certain nodes. 

I think this interplay between the two should be represented with better abstractions. Maybe the analysis phase needs to use the AST directly after all, but I can imagine that a higher-level model of the AST where we expose what we need and hide the rest, could be more benefitial in the long run.

kimgr:
> The main point I was trying to make is that the visiting side is lower level, and as such, it doesn't have 
> the complete picture. However, it does have access to all of the details about the AST. The analysis 
> (reporting/iwyu_output) side, on the other hand, is much higher level, with much more context, but 
> has lost much of the detail from the lower phase.

Exactly, that's what I was trying to say too.

> The main problem, as I see it, is that this separation of concerns is not well defined or maintained in 
> the code.

Agreed!

> For example, the visiting side tries to make high-level decisions about what things to report, whereas 
> the analysis will very often overrule those decisions at a later point after doing some low-level visiting 
> of certain nodes.
> I think this interplay between the two should be represented with better abstractions. Maybe the 
> analysis phase needs to use the AST directly after all, but I can imagine that a higher-level model
> of the AST where we expose what we need and hide the rest, could be more benefitial in the long
> run.

Yes, I think there might be a phase missing to sort out stuff like author intent. That is;

- AST visiting to collect "raw" uses
- Refinement phase to take stuff like author intent into account and move responsibilities around
- Analysis phase to play out the IWYU rules

There are abstractions missing, for sure, and some amount of purity where phases don't make too many "forward" assumptions. I.e. visiting the AST should be mostly oblivious to whether a use will count for something (except possibly for `CanIgnore*` type things to avoid collecting too much data).

And the Clang AST data structures are mostly wonderful to work with, so it would be nice to use them as a lingua franca throughout the pipeline (with some additional decoration like parent maps, etc) instead of inventing our own representations.

I'm regularly missing an include graph abstraction, where it's possible to ask about the "path" between two files (e.g. between the main file and `GetFileEntry(some_symbol)`), and I'm sure there are others.

jru:
Yep, all good ideas! :) Sadly, it is not obvious where to start with it all while not breaking anything..

Btw, would you like me to rework the UF_TargetRedecl commit? Or is it clear enough?

tests/cxx/explicit_template_instantiation-template.h
```diff
@@ -0,0 +1,14 @@
+//=== explicit_template_instantiation_template.h - test input file for iwyu ===//
```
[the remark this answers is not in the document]
jru: Done

tests/cxx/explicit_template_instantiation-template.h
```diff
@@ -0,0 +1,14 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPLICIT_TEMPLATE_INSTANTIATION_TEMPLATE_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPLICIT_TEMPLATE_INSTANTIATION_TEMPLATE_H_
+
+template<typename T> struct TemplateClass {};
+
+#endif  //INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPLICIT_TEMPLATE_INSTANTIATION_TEMPLATE_H_
```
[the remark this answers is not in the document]
jru: Done

tests/cxx/explicit_template_instantiation-template_direct.h
```diff
@@ -0,0 +1,14 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===------------------------------------------------------------------------------===//
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPLICIT_TEMPLATE_INSTANTIATION_TEMPLATE_DIRECT_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPLICIT_TEMPLATE_INSTANTIATION_TEMPLATE_DIRECT_H_
+
+#include "explicit_template_instantiation-template.h"
+
+#endif  //INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPLICIT_TEMPLATE_INSTANTIATION_TEMPLATE_DIRECT_H_
```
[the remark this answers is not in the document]
jru: Done

kimgr:
> Btw, would you like me to rework the UF_TargetRedecl commit? Or is it clear enough?

Yes, I think it would be nice if it followed existing form and described the context of the use, rather than telling the next phase what to do.

iwyu_ast_util.cc
```diff
@@ -942,12 +942,28 @@ bool IsFriendDecl(const Decl* decl) {
   return decl->getFriendObjectKind() != Decl::FOK_None;
 }
 
+clang::TemplateSpecializationKind GetTemplateSpecializationKind(
```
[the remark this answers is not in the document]
jru: Done

jru:
> Yes, I think it would be nice if it followed existing form and described the context of the use

Right. The thing is that it is not really a flag for the use, but for the decl associated to it. I could name it *UF_ExplicitInstantiation* and explain that the use targets one.

Another option is to create another set of `DeclFlags`, but it seems like overdoing it a little.

kimgr:
`UF_ExplicitInstantiation` works for me!

jru:
Done. All other comments were addressed too.
If you are good with it I'll rebase the fixups and reword the commit.

kimgr:
I'd rather review the changes commit-by-commit, so please rewrite the commit history to look the way you want to ship it. I started looking at the total diff, and overall it looks really nice!

jru:
It is ready.

tests/cxx/explicit_instantiation.cc
```diff
@@ -0,0 +1,48 @@
...
+// - Only declaration
+// IWYU: Template is...*explicit_instantiation-template.h
+extern template struct Template<short>;
+
+// - Only definition
+// IWYU: Template is...*explicit_instantiation-template.h
+template struct Template<double>;
+
+
+// The explicit instantiation of a specialization needs a declaration
```
kimgr: "... *only* needs a declaration..."?

iwyu_ast_util.cc
```diff
@@ -942,12 +950,20 @@ bool IsFriendDecl(const Decl* decl) {
...
 
+bool IsExplicitInstantiation(const clang::Decl* decl) {
+  clang::TemplateSpecializationKind kind = GetTemplateSpecializationKind(decl);
+  return kind == clang::TSK_ExplicitInstantiationDeclaration ||
+         kind == clang::TSK_ExplicitInstantiationDefinition;
+}
+
 bool IsForwardDecl(const NamedDecl* decl) {
-  if (const auto* record_decl = dyn_cast<RecordDecl>(decl)) {
+  if (const CXXRecordDecl* record_decl = DynCastFrom(decl)) {
```
kimgr: Hmm, `DynCastFrom` predates `auto` and I think it's clearer to use LLVM's `dyn_cast` and friends with `auto` than leaning on this hack.

To be perfectly honest, I think it's slightly easier to read the `DynCastFrom` line, but I had to spend a considerable amount of time understand how it worked the first time I saw it. `auto` is well-understood.

iwyu_ast_util.cc
```diff
@@ -155,6 +155,14 @@ void DumpASTNode(llvm::raw_ostream& ostream, const ASTNode* node) {
   }
 }
 
+clang::TemplateSpecializationKind GetTemplateSpecializationKind(
+    const Decl* decl) {
+  if (const clang::CXXRecordDecl* record = DynCastFrom(decl)) {
```
kimgr: See below -- I think we should prefer `const auto* x = dyn_cast<T>(y)` over `const T* x = DynCastFrom(y)`.

iwyu.cc
```diff
@@ -3732,11 +3732,20 @@ class IwyuAstConsumer
   //    template <class T> struct Foo;
   //    template<> struct Foo<int> { ... };
   // we don't want iwyu to recommend removing the 'forward declare' of Foo.
+  //
+  // Additionally, this type of decl is also used to represent template
```
kimgr: I remember reading the Josuttis/Vandevoorde book on C++ templates about a decade ago, and they consistently used "specialization" instead of "instantiation", that is, they were the same thing in the standard text back then.

I think "explicit instantiation" is the primary context in which "instantiation" is used, even now. So maybe this comment should say "... is also used to represent *explicit* template instantiations...". Everything else is a specialization, not an instantiation.

Also, the "Additionally ... also" pair looks like a double-take. I think you can remove "also".

iwyu.cc
```diff
@@ -109,6 +109,7 @@
 #include "iwyu_location_util.h"
 #include "iwyu_output.h"
 #include "iwyu_path_util.h"
+#include "iwyu_use_flags.h"
```
kimgr: Commit message needs some updating after this use-flag was repurposed.

iwyu.cc
```diff
@@ -1632,29 +1633,32 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-
+                             const char* comment = nullptr,
+                             UseFlags extra_use_flags = 0) {
     const NamedDecl* target_decl = used_decl;
 
     // Sometimes a shadow decl comes between us and the 'real' decl.
     if (const UsingShadowDecl* shadow_decl = DynCastFrom(used_decl))
       target_decl = shadow_decl->getTargetDecl();
-    
+
```
kimgr: Let's do this cleanup as a separate no-functional-change commit.

iwyu.cc
```diff
@@ -1668,12 +1672,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // TODO(csilvers): maybe just insert our own using declaration
     // instead?  We can call it "Use what you use". :-)
     // TODO(csilvers): check for using statements and namespace aliases too.
-    if (const UsingDecl* using_decl
-        = GetUsingDeclarationOf(used_decl, 
-              GetDeclContext(current_ast_node()))) {
+    if (const UsingDecl* using_decl = GetUsingDeclarationOf(
```
kimgr: Spurious reformat

iwyu_output.cc
```diff
@@ -578,14 +578,19 @@ void IwyuFileInfo::ReportFullSymbolUse(SourceLocation use_loc,
...
-    // all its preceding declarations, and not try to canonicalize.
-    if (!(flags & UF_FunctionDfn)) {
-      decl = GetDefinitionAsWritten(decl);
+    // Also, by default we canonicalize the location, using GetLocation.
+    const NamedDecl* report_decl = GetDefinitionAsWritten(decl);
+    SourceLocation report_decl_loc = GetLocation(report_decl);
+
+    // However, if we're defining the function or we are targeting an explicit
+    // instantiation, we want to use it as-is and not try to canonicalize at all.
+    if ((flags & (UF_FunctionDfn | UF_ExplicitInstantiation)) != 0) {
```
kimgr: There should be one or more testcases enabled by this change... Any chance you can get them into the same commit?

kimgr: I wonder if it would be possible to turn this into an if/else block instead, something like:
```c++
SourceLocation decl_loc;

if (flags & (UF_FunctionDfn | UF_ExplicitInstantiation) == 0) {
    // Rationale here...
    decl = GetDefinitionAsWritten(decl);
    decl_loc = GetLocation(report_decl);
} else {
    // ... and here
    decl_loc = report_decl->getLocation();
}
```
That way it looks more balanced (even if it's annoying with the `decl_loc` declaration outside the block.

iwyu.cc
```diff
@@ -2548,12 +2548,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
```
kimgr: Commit message:  "when there is a explicit instantiation definition visible" -> "... *an* explicit..."

iwyu.cc
```diff
@@ -2548,12 +2548,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
     const NamedDecl* decl = TypeToDeclAsWritten(type);
 
+    const SourceLocation current_loc = CurrentLoc();
```
kimgr: No need to outline this into a variable, just put the inline `CurrentLoc` calls back

iwyu.cc
```diff
@@ -3110,9 +3113,53 @@ class InstantiatedTemplateVisitor
     // We attribute all uses in an instantiated template to the
     // template's caller.
     ReportTypeUse(caller_loc(), actual_type);
+
+    // Report also all previous explicit instantiations (declarations and
+    // defintions) as uses of the caller's location.
```
kimgr: Typo: defintions

iwyu.cc
```diff
@@ -3110,9 +3113,53 @@ class InstantiatedTemplateVisitor
     // We attribute all uses in an instantiated template to the
     // template's caller.
     ReportTypeUse(caller_loc(), actual_type);
+
+    // Report also all previous explicit instantiations (declarations and
```
kimgr: Report also -> Also report

iwyu.cc
```diff
@@ -3110,9 +3113,53 @@ class InstantiatedTemplateVisitor
...
+    // defintions) as uses of the caller's location.
+    ReportExplicitInstantiations(actual_type);
+
     return Base::VisitSubstTemplateTypeParmType(type);
   }
 
+  bool VisitTemplateSpecializationType(
+      clang::TemplateSpecializationType* type) {
```
kimgr: `TemplateSpecializationType` has been pulled into the namespace with `using` at the top of the file. If you remove the namespace specifier, the declaration fits on one line.

iwyu.cc
```diff
@@ -3110,9 +3113,53 @@ class InstantiatedTemplateVisitor
...
+    if (CanIgnoreCurrentASTNode())
+      return true;
+
+    CHECK_(current_ast_node()->GetAs<clang::TemplateSpecializationType>() ==
```
kimgr: Same here, lose `clang::`

iwyu.cc
```diff
@@ -3110,9 +3113,53 @@ class InstantiatedTemplateVisitor
...
+    // Report previous explicit instantiations here, only if the type is needed
+    // fully.
+    if (!CanForwardDeclareType(current_ast_node()))
+      ReportExplicitInstantiations(type);
+
+    return Base::VisitTemplateSpecializationType(type);
+  }
+
+  void ReportExplicitInstantiations(const Type* type) {
+    const ClassTemplateSpecializationDecl* decl =
```
kimgr: Prefer `auto` + `dyn_cast<>`

tests/cxx/explicit_instantiation2-template_helpers.h
```diff
@@ -0,0 +1,40 @@
...
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPLICIT_INSTANTIATION2_TEMPLATE_HELPERS_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPLICIT_INSTANTIATION2_TEMPLATE_HELPERS_H_
+
+template <class T>
```
kimgr: Is this a re-declaration of the other `Template` in explicit_instantiation-template.h? In that case, maybe use `struct` instead of `class`, so they're consistent -- some compilers warn when decls differ only by class/struct keyword.

If not, please choose a unique-r name.

tests/cxx/explicit_instantiation2.cc
```diff
@@ -0,0 +1,99 @@
...
+// 4: Fwd-decl use in a template, provided as an explicit parameter.
+// 5: Full use in a template, provided as an default parameter.
+// 7: Full use in a template, provided as a template template parameter.
+// 8: Fwd-decl use in a template, provided as a template template parameter.
+//
+// Negative scenarios, where the dependent template specialization is not
+// required, or it does not provide an explicit instantiation:
+// 6: Fwd-decl use in a template, provided as a default parameter.
+// 9: Implicit instantiation of Template<int>
+//10: Specialization of Template<T>
```
kimgr: The alignment looks nice, but I prefer to always have a space after `//`.

tests/cxx/explicit_instantiation2.cc
```diff
@@ -0,0 +1,99 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "explicit_instantiation-template_direct.h"
+#include "explicit_instantiation2-template_helpers.h"
+#include "explicit_instantiation2-template_short_direct.h"
+
+// These tests verify that a dependent explicit instantiation declaration is
+// always required when the main template is also needed, but not otherwise.
+//
+// Positive scenarios:
```
kimgr: Maybe these comments would have better locality next to the respective test case if possible?

tests/cxx/explicit_instantiation2.cc
```diff
@@ -0,0 +1,99 @@
...
+// Negative scenarios, where the dependent template specialization is not
+// required, or it does not provide an explicit instantiation:
+// 6: Fwd-decl use in a template, provided as a default parameter.
+// 9: Implicit instantiation of Template<int>
+//10: Specialization of Template<T>
+
+
+// IWYU: Template is...*explicit_instantiation-template.h
+// IWYU: Template is...*template_short.h.*for explicit instantiation
+Template<short> t1a;
```
kimgr: Ah, now I see you've used naming to refer back to the list... Maybe better to use comments consistently, like `// 2` below?

tests/cxx/explicit_instantiation2.cc
```diff
@@ -29,6 +29,8 @@
 // 6: Fwd-decl use in a template, provided as a default parameter.
 // 9: Implicit instantiation of Template<int>
 //10: Specialization of Template<T>
+//11: Explicit instantiation with a dependent instantiation declaration, but
```
kimgr: Prefer space after `//` to alignment of numbered list.

iwyu_ast_util.cc
jru: Alright. I actually thought we settled for `DynCasFrom`, but yeah, I also think `dyn_cast` is a better choice.

iwyu_output.cc
jru: > There should be one or more testcases enabled by this change
You mean, new ones I should write? Without this change the tests wouldn't pass, so it is technically covered.

iwyu.cc
jru: Yes, I meant to write *explicit* instantiations. Thanks!

Regarding the usage of the terms, I must admit I find the whole thing quite confusing:
Looking at how the [C++ standard](from http://eel.is/c++draft/temp.spec) uses them it seems to me that *instantiation* is the process of substitution and generation, and also the resulting types/methods (ClassTemplateSpecializationDecl in clang), whereas *specializations* are the template definitions before the process of instantiation (the ClassTemplateDecl in clang)

But now looking at clang, `ClassTemplateSpecializationDecl` *is* an instantiated `CXXRecordDecl` whereas `ClassTemplateDecl` *references* the uninstantiated `CXXRecordDecl`, which seems a bit backward to me :)

jru:
Thanks for the kind words! 
I think I addressed all the comments, except [this one](https://github.com/include-what-you-use/include-what-you-use/pull/578#discussion_r236791395). Could you clarify what you meant?

iwyu_output.cc
kimgr: Sorry, I didn't see this before commenting further. It seems to me this change came about for a reason, and it would be nice if that reason was expressed as tests and committed together with the change.

But maybe the tests were failing before this change? Or it makes more sense to commit them with later changes? I think this looks good as-is if it turns out weird with tests together with this.

iwyu.cc
```diff
@@ -1649,12 +1652,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreDecl(target_decl))
       return;
 
+    const UseFlags use_flags =
```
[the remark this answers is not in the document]
jru: Well, I think is good practice to prevent accidental modifications, especially in long functions. It also helps document the intent.

kimgr:
\o/

bolshakov-a:
Hey @jru, looking at the test case 11 added in edfc3be19932ee20e5b1d7606997a30f7994f87d, it seems to me that `TemplateAsDefaultFullProvided` doesn't actually provide `ProvidedTemplate<short>` explicit instantiation because `explicit_instantiation2-template_short.h` is not included into `explicit_instantiation2-template_helpers.h`. So the explicit instantiation declaration should be reported in `explicit_instantiation2.cc`, shouldn't it?

PR #580, opened 2018-08-17
Refactor iwyu_tool for cleaner composition

kimgr:
Inspired by @pfaffe's multiprocessing removal in PR #575, I went ahead and cleaned up everything that's bothered me about `iwyu_tool.py` for the past year or so.

The total diff is rather huge, so please look at commits in isolation to get a better idea of what's changed.

Still no real plan for testing, but at least the functionality is better factored for unit testing now.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -21,11 +21,11 @@ See iwyu_tool.py -h for more details on command-line arguments.
 """
 
 import os
+import re
 import sys
 import json
 import argparse
 import subprocess
-import re
 import multiprocessing
 
 
@@ -108,22 +108,22 @@ def get_output(cwd, command):
 
 
 def run_iwyu(dbentry, iwyu_args, verbose):
-    """ Rewrite compile_command to an IWYU command, and run it. """
+    """ Transform dbentry to an IWYU command, and run it. """
     cwd = dbentry['directory']
 
     if 'arguments' in dbentry:
-        # arguments is a command-line in list form
+        # arguments is a command-line in list form.
         arguments = dbentry['arguments']
         compile_command, compile_args = arguments[0], arguments[1:]
         compile_args = ' '.join(compile_args)
     elif 'command' in dbentry:
-        # command is a command-line in string form
+        # command is a command-line in string form.
         compile_command, _, compile_args = dbentry['command'].partition(' ')
     else:
         raise ValueError('Invalid compilation database entry: %s' % dbentry)
 
     if compile_command.endswith('cl.exe'):
-        # If the compiler name is cl.exe, let IWYU be cl-compatible
+        # If the compiler name is cl.exe, let IWYU be cl-compatible.
         clang_args = ['--driver-mode=cl']
     else:
         clang_args = []
@@ -174,9 +174,7 @@ def main(compilation_db_path, source_files, verbose, formatter, jobs, iwyu_args)
                 print('WARNING: \'%s\' not found in compilation database.' %
                       source)
 
-    # Run analysis
     try:
-
         pool = multiprocessing.Pool(jobs)
         # No actual results in `results`, it's only used for exception handling.
         # Details here: https://stackoverflow.com/a/28660669.
@@ -198,6 +196,7 @@ def main(compilation_db_path, source_files, verbose, formatter, jobs, iwyu_args)
 
 def _bootstrap():
     """ Parse arguments and dispatch to main(). """
+
     # This hackery is necessary to add the forwarded IWYU args to the
     # usage and help strings.
     def customize_usage(parser):
@@ -219,7 +218,7 @@ def _bootstrap():
 
         parser.format_help = custom_help
 
-    # Parse arguments
+    # Parse arguments.
     parser = argparse.ArgumentParser(
         description='Include-what-you-use compilation database driver.',
         epilog='Assumes include-what-you-use is available on the PATH.')
@@ -236,8 +235,9 @@ def _bootstrap():
     parser.add_argument('-p', metavar='<build-path>', required=True,
                         help='Compilation database path', dest='dbpath')
     parser.add_argument('source', nargs='*',
-                        help='Zero or more source files (or directories) to '
-                        'run IWYU on. Defaults to all in compilation database.')
+                        help=('Zero or more source files (or directories) to '
+                              'run IWYU on. Defaults to all in compilation '
+                              'database.'))
 
     def partition_args(argv):
         """ Split around '--' into driver args and IWYU args. """
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -165,36 +165,47 @@ def parse_compilation_db(compilation_db_path):
     return compilation_db
 
 
+def slice_compilation_db(compilation_db, selection):
+    """ Return a new compilation database reduced to the paths in selection. """
+    if not selection:
+        return compilation_db
+
+    # Canonicalize selection paths to match compilation database.
+    selection = [os.path.realpath(p) for p in selection]
+
+    new_db = []
+    for path in selection:
+        if not os.path.exists(path):
+            print('WARNING: \'%s\' not found on disk.' % path)
+            continue
+
+        matches = [e for e in compilation_db if e['file'].startswith(path)]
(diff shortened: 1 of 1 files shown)


iwyu_tool.py
```diff
@@ -163,36 +163,48 @@ def parse_compilation_db(compilation_db_path):
...
-        # Source files specified, analyze the ones appearing in compilation db,
-        # warn for the rest.
-        entries = []
-        for source in source_files:
-            if os.path.isdir(source) or os.path.isfile(source):
-                entries.extend([e for e in compilation_db if e['file'].startswith(source)])
-            else:
-                print('WARNING: \'%s\' not found in compilation database.' %
-                      source)
+    compilation_db = slice_compilation_db(compilation_db, source_files)
```
jru: Is this second call deliberate?

iwyu_tool.py
```diff
@@ -111,35 +111,44 @@ def find_include_what_you_use():
...
+    """ Holds arguments of an IWYU invocation. """
+    def __init__(self, command, cwd):
+        self.command = command
+        self.cwd = cwd
+
+    def __str__(self):
+        return ' '.join(self.command)
+
+    @staticmethod
+    def parse_compile_command(entry, extra_args):
```
jru: Misleading name. Maybe, `create_from_compile_command`?

jru: Or simply, `from_compile_command`?

jru:
Nice cleanup! Now we just need to figure out how to test this :)

iwyu_tool.py
```diff
@@ -91,9 +91,24 @@ def clang_formatter(output):
...
-if sys.platform.startswith('win'):
-    IWYU_EXECUTABLE += '.exe'
+def find_include_what_you_use():
+    """ Find IWYU executable and return its full pathname. """
+    executable_name = 'include-what-you-use'
+    if sys.platform.startswith('win'):
+        executable_name += '.exe'
+
+    search_path = [os.path.dirname(__file__)]
+    search_path += os.environ.get('PATH', '').split(os.pathsep)
```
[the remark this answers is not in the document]
kimgr: Nice, I didn't know this existed! We can't assume Python 3, though, and I still want to add `dirname(__file__)` as an additional search path entry.

iwyu_tool.py
```diff
@@ -69,20 +69,23 @@ def clang_formatter(output):
...
+                             (state[1], line_no))
+            formatted.append(match.group(1))
+
+    return os.linesep.join(formatted)
 
 
 DEFAULT_FORMAT = 'iwyu'
 FORMATTERS = {
-    'iwyu': iwyu_formatter,
+    'iwyu': lambda output: output,
```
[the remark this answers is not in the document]
kimgr: I don't think it did -- it joined all lines with `'\n'.join(output)`. The `join` function puts the delimiter string between all items, not before the first or after the end.

The current output looks good, in any case.

iwyu_tool.py
```diff
@@ -120,18 +138,8 @@ def run_iwyu(dbentry, extra_args, verbose):
...
 
-    # Environment dictionary handling from https://stackoverflow.com/a/4453495
-    # ensures that: The current environment is not altered, instead it is
-    # copied for the child process, then the directory of iwyu_tool.py is
-    # prepended to the PATH variable.
-    env = os.environ.copy()
-    path_variable = os.path.dirname(__file__)
-    if 'PATH' in env:
-        path_variable += os.pathsep + env['PATH']
-    env['PATH'] = path_variable
```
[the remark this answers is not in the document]
kimgr: Sure -- instead of relying on the `PATH` variable in the environment to help `subprocess` locate `include-what-you-use`, I just locate it up-front with `find_include_what_you_use()` and then use the absolute path.

iwyu_tool.py
kimgr: I have a tendency to call everything that translates from something less desirable to something more desirable "parse" :) But `from_` is shorter than `parse_`, so I'll take that -- good suggestion, thanks!

kimgr: Updated.

kimgr:
I'm more and more uncomfortable with this code being untested (it's much more popular than I thought it would be!), but I'd like to make a separate project of bringing up a test infrastructure for `iwyu_tool.py`, since it has a hard dependency on `include-what-you-use`.

iwyu_tool.py
kimgr: I don't see a second call -- first `parse_compilation_db`, then `slice_compilation_db` based on the source file selection from the user.

iwyu_tool.py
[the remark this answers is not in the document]
jru: Agh. I confused the join with an append. Disregard :)

iwyu_tool.py
jru: It was in commit baf53f1. Probably it was reworked afterwards?

iwyu_tool.py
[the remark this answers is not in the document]
jru: Maybe a TODO comment then, for when Python 2 is dropped?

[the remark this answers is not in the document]
kimgr: I still don't see a way to convince `shutil.which` to look at additional paths beyond `PATH`, so I would have to modify `PATH` before calling it, which is always a mess because of the str/list/pathsep divide.

But maybe I can trade that complexity for the `.exe` normalization. I'll add a TODO.

iwyu_tool.py
kimgr: Ah, thanks for catching that. Rebasing snafu.

kimgr: Should be fixed now.

iwyu_tool.py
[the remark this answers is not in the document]
kimgr: Done!

kimgr:
Oh, @jru -- could you take this for a spin on Windows at some point?

jru:
Yes, sure. I will report back if I find something!

kimgr:
Oh my, that's horrible! I'll try to look into this tomorrow.

I was so happy to get rid of the space-delimited string in favor of a list. :-/

As for the verbose print, I think that @pfaffe's removal of multiprocessing will fix that (as only one thread will be launching processes in that model), so maybe we should just call it a multiprocessing bug for now. I like the idea of printing what to launch before launching (and completing it), so we can more easily spot hanging commands, for example.

kimgr:
Do you have a Windows `compile_commands.json` you could e-mail to me for reference? Thanks!

jru:
I distilled the `compile_commands.json` I used into a very representative entry.
Let me know if it is not enough, I think you shouldn't need more than this to reproduce it.

The main problem is that the source file ends up without any backslashes. In this case `c:directorytofilesource.cpp` instead of `c:\directory\to\file\source.cpp`

```
  {
    "directory": "c:\\directory\\to\\file",
    "command": "cl.exe /nologo /FS /c /I.\\ /I..\\..\\relative1 /I..\\..\\dir1\\dir2 /Ic:\\absolute\\dir /W3 /WX /D MACRO_1 /Fo\"c:\\quoted\\doublequotes\\\\\" c:\\directory\\to\\file\\source.cpp",
    "file": "c:/directory/to/file/source.cpp"
  },
```

jru:
One more thing before I forget: The file selection is not case-insensitive on Windows anymore. 

Using the `compile_commands.json` above and doing `iwyu_tool.py -p . c:\directory\to\file\SOURCE.cpp` reports that the file doesn't exist in the compilation database. However doing `iwyu_tool.py -p . c:\directory\to\file\source.cpp` works.

kimgr:
Did that work before my changes, though? I think the prior filtering just silently discarded non-matches. The diagnostic when a file is not found is new.

Should be easy to fix now, though. Thanks!

jru:
> I think the prior filtering just silently discarded non-matches.

Just checked. You are completely right! Unless the exact case is used, it is ignored.

kimgr:
@jru Thanks! Do you have an actual `compile_commands.json` you could send to me in private? The `/Fo` quoting example looks wrong, so I suspect/hope that you've transcribed it incorrectly :-)

Issue #419, opened 2017-03-22
IWYU crashes on Windows for function templates calling builtins


zjturner:
I'd like to run IWYU on LLVM itself, but it seems there are some issues getting in my way.  First and foremost is that LLVM does not use `make`, which you seem to require for running over a whole codebase.  My first intuition was to just use `cmake` with the `Unix Makefiles` generator, but this produces makefiles with a layer of indirection which first shells out to CMake to actually run the dependency check, and from there I'm not sure what happens but it doesn't work.  It runs over the entire codebase and I just get no output at all.  Here are the steps I'm using.

1. Build LLVM with IWYU in-tree, release mode.
2. In separate tree, run `cmake -G "Unix Makefiles" "-DCMAKE_CXX_COMPILER=/bin/dir/from/step/1/clang++" "-DCMAKE_C_COMPILER=/bin/dir/from/step/1/clang" "/path/to/llvm/src"
3. Run `make -j32`  (this is necessary because LLVM builds tools like `llvm-tblgen` which other steps will fail if they are not present, so I do a full build once just to make sure all the required tools are present.
4. Run `make -k CXX=/bin/dir/from/step/1/include-what-you-use`

At this point it plows through from 0 to 100% in about 30 seconds, printing no output whatsoever.  If I run `make -n` instead, all the commands that it's running are just cmake commands.

Thoughts?

Bu11etmagnet:
LLVM *does* use make, if you ran  `cmake -G Unix Makefiles`. Unfortunately, CMake hardcodes the compiler in the makefiles it generates, so it's pointless to to try to override it on the command line: there's no difference between `make CXX=include-what-you-use` and `make`.

What you could try (warning, untested) is symlink your compiler somewhere (e.g. `ln -s /usr/bin/clang++ /tmp/clang++`) and then `cmake "-DCMAKE_CXX_COMPILER=/tmp/clang++"` . Then try to find out what tools beyond `tblgen` are needed and build them. Then, replace the symlink to point to IWYU (`ln -sf /usr/bin/include-what-you-use /tmp/clang++`) and run `make -k`.

purpleKarrot:
Try this:

```
cmake -GNinja -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE:STRING=include-what-you-use ..
ninja
```

Or, if you need to pass arguments to `include-what-you-use`, like a mapping file:

```
[...]

PR #586, opened 2018-08-29
Enable incremental processing

trixirt:
Fixes a panic building llvm/clang with clang 7 on Ubuntu 18.04

The issue is the Sema::TUScope is set to nullptr by
Sema::ActOnEndOfTranslationUnit unless incremental processing is enabled.

The file lib/TableGen/JSONBackend.cpp reduces to the testcase

template <int a> struct b { static constexpr int c = a; };
template <bool a> using d = b<a>;
template <bool, typename> struct aa;
template <typename...> struct e;
template <typename f, typename g> struct e<f, g> : aa<f::c, g>::h {};
template <typename i> i ab();
struct j {
  b<true> k;
};
template <typename> struct l : j { typedef decltype(k) h; };
template <typename i> struct m : e<d<!bool()>, l<i>> {};
template <typename i> struct n : m<i>::h {};
struct o {
  static b<false> p(...);
};
template <typename i> struct q : o { decltype(p(ab<i>())) h; };
template <typename i> struct r : e<n<i>, q<i>> {};
template <int> struct s;
template <bool, typename ac> struct aa { typedef ac h; };
class G;
struct H {
  template <typename t = G, typename u = int,
            typename s<e<r<t>, u>::c>::h = true>
  H();
};
class G {
  G(G &);
};
struct I {
  H ar;
};

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3975,6 +3975,7 @@ class IwyuAction : public ASTFrontendAction {
     compiler.getPreprocessor().addPPCallbacks(
         std::unique_ptr<PPCallbacks>(preprocessor_consumer));
     compiler.getPreprocessor().addCommentHandler(preprocessor_consumer);
+    compiler.getPreprocessor().enableIncrementalProcessing();
 
     auto* const visitor_state
         = new VisitorState(&compiler, *preprocessor_consumer);


kimgr:
Thanks for the root-cause analysis! This is captured in https://github.com/include-what-you-use/include-what-you-use/issues/419, but we hadn't gotten this far (see also the message to the Clang list).

Unfortunately it looks like your change breaks our test suite, so it's probably not the right one :) Unless the tests are covering for the bug, of course. Can you dig into it?

kimgr:
I wonder if the patch should go in Clang instead, so that TUScope isn't used for late parsing unless incremental processing is enabled? I'll try and dig into this and see what I can learn.

trixirt:
I am getting busy on another project so I will be delayed in getting to resolving the failures.
My thoughts generally about fixing in clang, is that it would be harder to convince clang folks, than changing iwyu.

trixirt:
I am withdrawing this patch but will have one shortly to reproduce the crash.

PR #589, opened 2018-09-16
Build to $build/bin instead of $build

kimgr:
This makes the layout the same for out-of-tree and in-tree build dirs
for IWYU.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,8 @@ endif()
 
 message(STATUS "IWYU: configuring for LLVM ${LLVM_VERSION}...")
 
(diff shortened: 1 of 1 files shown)


kimgr:
@pfaffe Here's the first refactor.

pfaffe:
When building shared libraries on Windows, the DLL part of the library is considered a RUNTIME target. Is that what you want?

kimgr:
There doesn't seem to be an output artifact type that's executables only. And I don't think we'll be building DLLs from IWYU for a long while, so following the default here is fine by me.

kimgr:
I realized I can set the output dir as a target property instead. Would that be better? It seems equivalent to me.

pfaffe:
If there aren't any shared libraries in IWYU, it doesn't matter to me.

PR #590, opened 2018-09-16
Remove dead synthesis of resource dir

kimgr:
HeaderSearchOpts.ResourceDir is never empty, so this code would never
execute.

Besides, it's too late to set the resource dir here; it's already been
used in Driver::BuildCompilation to build toolchain paths relative to
the compiler executable.


kimgr:
@pfaffe Another refactor. No functional change intended.

PR #591, opened 2018-09-16
Depend on clang-headers when built in-tree

kimgr:
Depends on PR #589.

As an initial refactor, this introduces a CMake variable for in-tree vs out-of-tree. I think this could be made cached, but I don't know enough CMake to make that happen in an idiomatic way. @pfaffe, do you know how to do that?


pfaffe:
Why would you cache this? There's nothing complex or expensive about it, so caching wouldn't give you anything, and at the same time this isn't a setting that should be configurable in any way.

kimgr:
Thanks! Merged after rebase.

PR #594, opened 2018-10-13
Add --tolerate_transitive mode

kimgr:
In this mode, IWYU does not suggest includes that are already
transitively available.

This makes IWYU much less intrusive, as it will only remove unused
includes, and add includes that are required to compile (which should be
a very rare occurrence -- IWYU typically runs on a compilable code
base).

It also makes IWYU weaker, it no longer works to simplify refactoring,
dependency cutting and improve self-documentation.

Replaces #529.


WSoptics:
@kimgr looks very good an clean to me.  Thank you so much for your efforts!

fiesh:
@kimgr looks very good and clean to me. Thank you so much for your efforts!

mvpossum:
I'm testing this feature but it stills suggests to add stuff, despite the code already compiling. Do you want me to dig more into this? What could be the reason?

fiesh:
(Just to clarify, I was purely talking about the implementation, I haven't tested it yet and I won't get around to it before November alas.)

fiesh:
@mvpossum , @Ede123 Could you compare your results with this patch to the results with the patch from #529?  Thank you.

vmartinv:
@fiesh @kimgr I compared both versions on my project. Unfortunately I cannot share the output. I do can say this version suggests more forward includes (I'm not very sure if they are accurate since I don't have time to test). On the other hand #529 suggests to add a lot of headers, it doesn't seem to be working as expected.

kimgr:
@Ede123 Thanks for testing.
> If I disable forward declarations using --no_fwd_decls it removes a lot of them that are actually required.

Interesting, does IWYU make this mistake without `--tolerate_transitive` also?

I wonder if `--tolerate_transitive` should take fwd decls into account, too. I think it should be possible to detect that a fwd decl (or full decl) is already provided via one of the present includes, and use that to discard redundant fwd decls.

kimgr:
@vmartinv Thank you for the feedback.

So does IWYU with `--tolerate_transitive` only suggest to add fwd decls, or includes too? It would be interesting to see a reduced example where it suggests to add an include that's already transitively available. There may be special cases (such as placement-new, which requires `<new>`), but generally includes should not be added in this mode.

vmartinv:
> @vmartinv Thank you for the feedback.
> 
> So does IWYU with `--tolerate_transitive` only suggest to add fwd decls, or includes too?
I don't get any suggestions about adding includes, only fwd decls.

kimgr:
@vmartinv OK, great! So if I manage to solve the forward-decl problem, IWYU should work for your project. I was afraid there were multiple bugs. I'll take a look at how feasible it would be.

vmartinv:
Could you upload the fixes so I can confirm you? Also I realized you can try running it against IWYU code itself to debug it, I saw some fwd-decls bugs there (in case you aren't already doing it).

kimgr:
Heh, I haven't had a chance to code up something that works yet. Will update when there's something to test.

vmartinv:
This test case may be related  #603

vmartinv:
@kimgr I found this test case, which I believe points to the main problem with this implementation.

- indirect.h
```
class Foo;
```

- direct.h
```
```

- use.cpp
```
#include "tests/cxx/direct.h"
Foo foo;
```

kimgr:
Then there's something else going on, iwyu should normally only analyze use.cpp and its associated header (we heuristic-check for headers named something like use.h). So direct.h should never be modified here unless explicitly asked for on the command line. Or there's some other quirk that's causing this.

kimgr:
You her the same behavior without `--tolerate_transitive`, right? In that case maybe move the comment to #603.

vmartinv:
@kimgr Sorry, I messed up with filenames, here it is

- indirect.h
```
class Foo {};
```

- use.h
```
```

- use.cpp
```
#include "tests/cxx/use.h"
Foo foo;
```

__IWYU Output__
```
tests/cxx/use.h should remove these lines:
- #include "tests/cxx/indirect.h"
```

kimgr:
@vmartinv I would argue that's correct --

* `use.h` does not use `Foo`
* `use.cpp` does not use `Foo`, but forward-declares it without using it

so I would expect IWYU to recommend the fwd-decl in `use.cpp` should be removed.

kimgr:
@vmartinv Ah, right you are! I didn't use my eyes enough when reading :)

vmartinv:
Problem is that `FileTransitivelyIncludes` should also take into account the suggestions done previously. In the example it says that `use.h` includes `indirect.h` but that is wrong because IWYU previously suggested to delete that include.

kimgr:
Thank you, that makes sense. I'm not sure how the algorithm could/should change to accommodate that, I need to give it some more thought.

kimgr:
This does not appear to be going anywhere. Closing.

PR #599, opened 2018-10-31
Update libstdc++ mappings

loskutov:
Add mappings from explicit `@headername` annotations


kimgr:
Thank you!

This would also need to go in iwyu_include_picker.cc, which has a built-in mapping for libstdc++. Could you use a similar shell pipeline to produce aggregate initializers instead of JSON?

That way we should be able to see whether the tests pass with these changes.

The .imp files for GCC were temporarily the single point of truth, but we reverted to the built-in mappings to reduce deployment complexity, so now the .imp files are mostly complementary.

loskutov:
Also, if .imp files are useless, why maintain them at all?

kimgr:
I'm still hoping that we can remove policy from the IWYU binary and dynamically select the right mappings depending on compiler environment. But that work hasn't seen much progress in the last number of years, so it could be a pipe dream.

Maybe you have a point that the redundant .imp files should be removed eventually, until we can use them for something, but let's them in sync for now.

loskutov:
OK, I believe they're currently in sync.

kimgr:
This looks good to me. I spot-checked a few recent manual changes and they are still intact. I'm a little worried that the update will break some legacy libstdc++, but so far I've found no obvious problems.
Tests all pass.

Can you squash the commits into one so I can merge it?

kimgr:
Thanks! The build is currently red due to a change in Clang, working on that separately.

PR #600, opened 2018-11-01
Look for iwyu binary from environment variable in iwyu_tool.py

vmartinv:
I have different binaries for different platforms and I need to get more specific when using this tool. This will come handy for sure.


vmartinv:
@kimgr Fixed it!

kimgr:
Thanks!

PR #601, opened 2018-11-02
A testcase for iwyu crash while analyzing llvm 7+

trixirt:
This testcase reproduces a problem with iwyu-ing llvm.
The problem still exists.
I am testing a fix on the clang side.
Irregardless of where the fix happens, a reproducing test is needed.

### DIFF
--- a/tests/cxx/llvm_tablegen_jsonbackend.cc
+++ b/tests/cxx/llvm_tablegen_jsonbackend.cc
@@ -1,0 +1,78 @@
+//===--- llvm_tablegen_jsonbackend.cc - test input file for iwyu ----------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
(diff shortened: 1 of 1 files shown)


trixirt:
A link to the clang change.
https://reviews.llvm.org/D54047

kimgr:
I ran this crash through a debugger, and this actually seems to hit `InstantiateImplicitMethods` in iwyu.cc, not `ParseFunctionTemplates`.

So I don't think this has anything to do with late-parsed templates, but rather is a more pressing problem with us trying to force-instantiate implicit ctors/dtors.

kimgr:
This is most likely related to #419 (it's crashing on a NULL `TUScope` with a similar backtrace).

kimgr:
This is a good testcase, but we don't have any XFAIL facility in the current infrastructure, so we can never merge it. It's still referenced from the open bug issue, so available for reference. Closing.

PR #603, opened 2018-11-05
test case where --no_fwd_decls suggest deleting important fwd_decls

vmartinv:
With the test case I added currently I get this output:
```
tests/cxx/fwd_decl_without_full_use.cc should add these lines:

tests/cxx/fwd_decl_without_full_use.cc should remove these lines:
- class Foo;  // lines 16-16
```
Which is wrong because if I delete this line obviously it doesn't compile. I think the expected behavior of `--no_fwd_decls` would be to not suggest forward declarations. If a forward declaration is already there it should keep it that way.

### DIFF
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -85,6 +85,7 @@ class OneIwyuTest(unittest.TestCase):
       'prefix_header_includes_remove.cc': ['--prefix_header_includes=remove'],
       'prefix_header_operator_new.cc': ['--prefix_header_includes=remove'],
       'quoted_includes_first.cc': ['--pch_in_code', '--quoted_includes_first'],
++     'fwd_decl_without_full_use.cc': ['--no_fwd_decls'],
     }
     prefix_headers = [self.Include('prefix_header_includes-d1.h'),
                       self.Include('prefix_header_includes-d2.h'),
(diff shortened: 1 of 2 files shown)


vmartinv:
Thanks for the response. What if I want that `--no_fwd_decls` leave the current existing forward declarations as they are?

kimgr:
This is an iwyu bug, so there's not a lot to do, as far as I can see. I think it should be possible to use a technique similar to the one from #594 to fix it. Then again, I still haven't checked a verbose log to see where it's being discarded, so I could be wrong.

kimgr:
So I ran this with higher log verbosity:
```
$ include-what-you-use -Xiwyu --verbose=7 -Xiwyu --no_fwd_decls -I . tests/cxx/fwd_decl_without_full_use.cc
...
--- Calculating IWYU violations for tests/cxx/fwd_decl_without_full_use.cc ---
Ignoring use of Foo (tests/cxx/fwd_decl_without_full_use.cc:15:19): definition is present: tests/cxx/fwd_decl_without_full_use.cc:13:7

tests/cxx/fwd_decl_without_full_use.cc should add these lines:

tests/cxx/fwd_decl_without_full_use.cc should remove these lines:
- class Foo;  // lines 13-13

The full include-list for tests/cxx/fwd_decl_without_full_use.cc:
---
```
So there seems to be an unintended side-effect of `--no_fwd_decls`:
- First the presence of `--no_fwd_decls` reclassifies all fwd-decl uses as full uses
- Then there's a policy in `ProcessFullUse` that ignores all uses for which there is a declaration earlier in the same file

I'm not quite sure how the second point makes sense, so I tried removing it. But then we stumble on another failure, and when I remove that, another.

I don't have any clear idea for how this should be solved yet.

kimgr:
This is now fixed on master.

PR #604, opened 2018-11-06
Failing test case when there's a full template use

vmartinv:
IWYU currently suggests to remove the include line even though this is a full use.

### DIFF
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -191,6 +191,7 @@ class OneIwyuTest(unittest.TestCase):
       'using_aliased_symbol_unused.cc': ['.'],
       'varargs_and_references.cc': ['.'],
       'virtual_tpl_method.cc': ['.'],
+      'template_use.cc': ['.'],
     }
     # Internally, we like it when the paths start with rootdir.
     self._iwyu_flags_map = dict((posixpath.join(self.rootdir, k), v)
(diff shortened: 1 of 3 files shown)


kimgr:
I finally came around to testing this. Either I'm missing something, or you're not seeing what you think you're seeing :)

```
$ ./run_iwyu_tests.py template_use
...
>>> Running ../build/bin/include-what-you-use -Xiwyu --verbose=3 -I . tests/cxx/template_use.cc

(tests/cxx/template_use.h has correct #includes/fwd-decls)

(tests/cxx/template_use.cc has correct #includes/fwd-decls)

F
FAIL: runTest (__main__.template_use)
  File "./run_iwyu_tests.py", line 259, in <lambda>
    {'runTest': lambda self, f=filename: self.RunOneTest(f),
  File "./run_iwyu_tests.py", line 230, in RunOneTest
    iwyu_flags, clang_flags, verbose=True)
  File "/home/kgrasman/code/iwyu/include-what-you-use/iwyu_test_util.py", line 455, in TestIwyuOnRelativeFile
    test_case.assertTrue(not failures, ''.join(failures))
AssertionError: 
Unexpected summary diffs for tests/cxx/template_use.h:
+++ 
@@ -0,0 +1 @@
+(tests/cxx/template_use.h has correct #includes/fwd-decls)
---


Ran 1 test in 0.047s

FAILED (failures=1)
```

So the test runner is complaining that `template_use.h` does not have any test expectations, not that there's anything wrong with the template.

If possible, IWYU automatically analyzes an `x.cc` file and its associated header `x.h` together, so if the stem of the filenames matches, they both need to have the IWYU test framework markup. If I rename `template_use.h` to `template_use-direct.h`, the test passes here.

kimgr:
Closing because this does not appear to be a real problem. Please reopen if you have more information to the contrary.

PR #605, opened 2018-11-08
When include-what-you-use if found, return the real path

trixirt:
If iwyu_tool.py is call by a relative path, include-what-you-use will
also be called with a static, relative path. Which breaks once the
cwd changes.  So force the tool path to be absolute.


trixirt:
can not reproduce this error, going to fuz the change and try again.

PR #606, opened 2018-11-08
When include-what-you-use if found, return the real path

trixirt:
If iwyu_tool.py is call by a relative path, include-what-you-use will
also be called with a static, relative path. Which breaks once the
cwd changes.  So force the tool path to be absolute.


kimgr:
Please rebase on top of latest master and get rid of the "Silence compiler warning" commit.

No need to use both `os.path.abspath` and `os.path.realpath` -- `os.path.realpath` is enough (it makes the resolved path absolute).

trixirt:
Changes made
Thanks for the review.

kimgr:
Looks good to me, thanks!

PR #607, opened 2018-11-12
Explicit instantiation - refactor

jru:
Refactor commits from #578


kimgr:
Both look good to me!

You've changed the order of the checks in `CanIgnoreType` so that the resugar map is checked before the `IsDefaultTemplateParameter", but I guess that's fine as no tests are complaining.

I find the new code much more readable, so LGTM.

Issue #583, opened 2018-08-21
iwyu_tool.py does not work on Windows


kimgr:
After PR #580, iwyu_tool.py no longer works on Windows.

This is because `shlex.split` does not handle Windows path delimiters and quoting correctly, so the command-line becomes jumbled.

kimgr:
@jru Could you try this patch on your Windows system?

```diff
commit 80fe4d8331229afc87850733b40c9dd4524cbfd1 (HEAD -> iwyu-tool-windows)
Author: Kim Grasman <kim.grasman@gmail.com>
Date:   Wed Aug 22 18:05:44 2018 +0200

    [iwyu_tool] Fix command splitting for Windows
    
    Turns out shlex.split does not do the right thing on Windows by
    default. But with the posix=False flag, it comes close enough to work
    with JSON compilation database input.
    
    Add an extra step to strip quotes for quoted arguments, because the
    command list is passed on to subprocess, which quotes anything
    containing whitespace, whether it's already quoted or not.
    
    No behavior change for non-Windows systems.

diff --git a/iwyu_tool.py b/iwyu_tool.py
index 9445293..084d5fd 100755
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -91,6 +91,27 @@ FORMATTERS = {
 }
 
 
+def split_command(cmdstr):
+    """ Split a command string into a list, respecting shell quoting. """
+    def strip_quotes(arg):
+        if arg.startswith('"') and arg.endswith('"'):
+            return arg[1:-1]
+        return arg
+
+    # shlex.split with posix=False is known to not work perfectly for Windows
+    # command-lines, but it seems good enough in the general case for
+    # compilation databases.
+    windows = sys.platform.startswith('win')
+    cmd = shlex.split(cmdstr, posix=not windows)
+
+    # It does, however, retain quotes around arguments, which causes
+    # double-quoting and escaping when passed on to subprocess. Strip them.
+    if windows:
+        cmd = [strip_quotes(a) for a in cmd]
+
+    return cmd
+
+
 def find_include_what_you_use():
     """ Find IWYU executable and return its full pathname. """
 
@@ -130,7 +151,7 @@ class Invocation(object):
             command = entry['arguments']
         elif 'command' in entry:
             # command is a command-line in string form, split to list.
-            command = shlex.split(entry['command'])
+            command = split_command(entry['command'])
         else:
             raise ValueError('Invalid compilation database entry: %s' % entry)
```

[...]

PR #609, opened 2018-11-19
Fix iwyu tool for non posix systems

alacasta:
# What?
When dealing with non-posix systems, iwyu_tool run from a compilation commands database handles incorrectly the paths that can appear in the commands (Typically, the -I followed by corresponding paths). This is because of an inaccurate shlex.split call. 

**Issue**: iwyu_tool.py does not work on Windows #583

# Details
When splitting the command from a compile command database, specification on posix or non-posix standard is needed to avoid incorrect parsing on non posix based OS. When running on Windows, a command with complex includes
```
clang -Iinclude\\a\\directory\\using\\os\\separators file.cc
``` 
may be composed by something like
```
[iwyu_tool.IWYU_EXECUTABLE, '-Iincludeadirectoryusingosseparators', 'file.cc']
```
instead of
```
[iwyu_tool.IWYU_EXECUTABLE, ' -Iinclude\\a\\directory\\using\\os\\separators', 'file.cc']
```
As a result, iwyu_tool when running on Windows fails because these paths do not exist.
# Fix
If specifying the non-posix flag to shlex.split, the commands are adequately splitted both in windows and linux/Mac

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -169,7 +169,7 @@ class Invocation(object):
             command = entry['arguments']
         elif 'command' in entry:
             # command is a command-line in string form, split to list.
-            command = shlex.split(entry['command'])
+            command = shlex.split(entry['command'], posix=not sys.platform.startswith('win'))
         else:
             raise ValueError('Invalid compilation database entry: %s' % entry)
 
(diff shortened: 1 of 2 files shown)


alacasta:
Thanks for the comment. Let me work on merging both and i'll come with a suggestion to be merged into yours. I think that looks more standard and covers more cases.
If you prefer to close this, I can just try to open a newer one combining both. 👍

kimgr:
Closing this; I'm about to post fixes for all (:crossed_fingers:) Windows issues as separate pull requests.

PR #613, opened 2018-11-27
 Use utf-8-sig encoding with utf-8 bom files

trixirt:
With utf8 bom files, fix_includes.py sets the encoder to utf-8.
This is mostly correct.
The problem is the bom header is not removed on decoding.
Instead it is transformed into the utf-16 bom (guess) which messes up the diff-ing of files for headers unlucky enough to be near the first line.
The recommended encoder is utf-8-sig
https://docs.python.org/3/library/codecs.html#module-encodings.utf_8_sig

Start pytests dir, a place for python unit tests.


kimgr:
Does `utf-8-sig` exist in Python 2 too? We need to support both.

Not convinced about the `pytests/` directory, I'm guessing that needs some further test framework? We try to avoid extra dependencies for our Python code (and C++, mostly).

kimgr:
When I wrote that encoding-heuristic code, I had some local scrap files that I used for testing. I would have liked to build a real test harness out of them, but I ran out of steam before getting that far.

I really try my best to avoid third-party dependencies for support tools, because I want to keep it simple and obvious to hack on IWYU.

If you can describe any concrete changes you'd like to make to `fix_includes_test.py`, maybe we can sketch out a plan before you start making code changes?

trixirt:
Dropped unit test commit.

kimgr:
Looks good to me. Could someone try this on Windows to make sure the `utf-8-sig` is represented there too? Thanks!

trixirt:
I had verified this is ok with windows, the log was captured in the unit test that i dropped.

PR #614, opened 2018-11-27
Create proper patch files

bmanga:
This patch adds an additional command line argument to `fix_includes.py` to optionally specify the directory the files in the diff should be relative to.

This way, the script can generate an actual patch that can then be applied to the codebase.


kimgr:
Thanks for the patch! We need test coverage for this in `fix_includes_test.py`. Related, it looks like these changes break the current test coverage in `fix_includes_test.py`. Could you take a look?

bmanga:
Hi!
Sure, I will create a testcase!

kimgr:
The whole `a/` and `b/` prefix thing looks unnecessary to me -- I see it in git diffs, but not in normal `diff -u` output.

And why do we want to remove the path prefix? Is it a privacy thing? If not, I guess most patch tools accept the `-pN` argument to strip N path components from the beginning of the paths in the patch.

bmanga:
I didn't know of the existence of `-pN`. Thanks for introducing me to it! :)

My use case is to generate a patch, edit some parts and then apply it to a codebase, so I am not personally concerned about the privacy aspect.

I suppose if someone cared about it and wanted to share the patch they could simply apply it and generate a patch via git/svn. I didn't want to do this in my case because it triggers the recompilation of all the edited files, even those that I want to remove the edits from.

I wonder if at this point the tool could just print a patch-friendly output by default instead of relying on another cl argument then. Do you have any preference on the matter?

kimgr:
> I wonder if at this point the tool could just print a patch-friendly output by default
> instead of relying on another cl argument then. Do you have any preference on the matter?

This sounds good to me, let's just include the filenames. I'm not sure if we should make an effort to make them relative -- it would be nice if we didn't include absolute paths, but I guess any choice here will be unworkable from some other perspective. Let's just add the filenames as they appear in the iwyu_record.

No need for `a/` and `b/` prefixes either, as far as I can see.

kimgr:
This would be nice, but it doesn't seem to be moving forward. Closing for now.

PR #615, opened 2018-11-29
Nested wrapper header support

jbytheway:
I ran into a problem attempting to use IWYU on a project with the following pattern:

* a 3rd-party library private header, included by
* a 3rd party library public header, included by
* a local project header intended to be the only way to include headers from that 3rd party library (and which also uses symbols from that library), included by
* a local project source file.

The existing special case code for handling private headers forces the last header to include the first one, even when a mapping exists from the first to the second.

This PR fixes that, by tweaking the logic so that we will retain private includes when they exist, but not necessarily add them whenever we are a potential public header for them.

The test I've added piggy-backs on an existing test to take advantage of its mapping file.  Let me know if that's inappropriate and I should make this test independent of existing test files.

### DIFF
--- a/tests/cxx/export_nesting.h
+++ b/tests/cxx/export_nesting.h
@@ -1,0 +1,26 @@
+//===--- export_nesting.h - test input file for iwyu ----------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPORT_NESTING_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPORT_NESTING_H_
+
+// We export a header which already re-exports things to verify that nested
+// exports are acceptable.
+
+#include "tests/cxx/export_nesting-d1.h" // IWYU pragma: export
+
+Nested_Enum f;
+
+#endif  // INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPORT_NESTING_H_
+
+/**** IWYU_SUMMARY
+
+(tests/cxx/export_nesting.h has correct #includes/fwd-decls)
+
+***** IWYU_SUMMARY */
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -754,8 +754,10 @@ set<string> CalculateMinimalIncludes(
     // this is a file that the use-file is re-exporting symbols for,
     // and we should keep the #include as-is.
     const string use_file = ConvertToQuotedInclude(GetFilePath(use.use_loc()));
-    if (use.PublicHeadersContain(use_file)) {
-      use.set_suggested_header(ConvertToQuotedInclude(use.decl_filepath()));
+    const string decl_file = ConvertToQuotedInclude(use.decl_filepath());
+    if (use.PublicHeadersContain(use_file) &&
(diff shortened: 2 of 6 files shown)


kimgr:
Thank you, this looks interesting. I can't claim to understand the premise of the original code, but the improvement as described sounds useful.

I would much prefer if the test was free-standing -- is it possible to create this double re-export without mappings, e.g. with nested `IWYU pragma: export`?

I'll try and get my head around this in the meantime.

jbytheway:
> > i.e. if the decl was in a private header and we export that header then we are forced to include it.
> 
> I think this may be include-what-you-use policy -- IWYU does its best to flatten the header dependency graph. What if we included `<sstream>` and only used `std::string` -- wouldn't we want to get rid of `<sstream>` in favor of `<string>`?

Yes, and we would, because we don't export `<string>`.  This logic only applies in the specific case where we are a public header for some decl we use in a private header.  This is a niche situation.

> I guess explicit exports/mappings could override that, but I guess I don't understand how your change takes them into account.

The explicit mappings are taken account inside the call to `use.PublicHeadersContain(use_file)`.

> Current behavior: if a mapping leads back from decl-file to use-file, attribute use to decl-file.
> New behavior: if a mapping leads back from decl-file to use-file _and use-file directly includes decl-file_, attribute use to decl-file.
> 
> This implies there's some case where a mapping can occur from decl-file to use-file, _without_ use-file directly including decl-file. Why can we assume this happened because of nested re-exports?

Well, there are other ways.  It can happen via exports, mapping files, `pragma: private`, or some of the IWYU heuristics.  But I think all should be handled similarly.

> Sorry for being slow, I just want to make sure we don't regress some other common scenario.

In any code that was previously IWYU-compliant, this logic (before the change) will have ensured that use-file already includes decl-file.  And in that situation the new behaviour matches the current behaviour.  So, this change ought not to have any impact on code already using IWYU; it only matters for new code.

(It's possible there's some corner case e.g. when including decl-file through a symlink that something might go wrong; I don't understand IWYU's path canonicalization well enough to be certain what will happen there)

kimgr:
Thanks for the extensive explanation, now I understand.

I have some style comments on the tests, I'll get back to that later today. Otherwise, looks good!

tests/cxx/export_nesting_1.h
```diff
@@ -0,0 +1,15 @@
+//===--- export_nesting_1.h - test input file for iwyu --------------------===//
```
kimgr: Additional includes are usually called `-d1.h` and `-i1.h`, I think for "direct" and "indirect".

So I think these two supporting headers should be renamed:
- `export_nesting_1.h` -> `export_nesting-d1.h`
- `export_nesting_2.h` -> `export_nesting-i1.h`

tests/cxx/export_nesting.h
```diff
@@ -0,0 +1,26 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPORT_NESTING_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_EXPORT_NESTING_H_
+
+// We export a header which already re-exports things to verify that nested
+// exports are acceptable
```
kimgr: Full stop at end of sentence

kimgr:
Left a few nits inline, otherwise I think this is ready to go.

kimgr:
Oh, and please squash the commits together.

kimgr:
Bingo, thanks!

PR #616, opened 2018-11-29
A POC on how to refactor fix_includes_test.py into pytests

trixirt:
A followup on how to refactor the python test setup to pytest.
https://docs.pytest.org/en/latest/

Input files are <the-source>.input,  this is the starting file that will change.
Test copies input to <the-source>
Then runs fix_includes.py <  <the-source>.iwyu
Then generates a diff of the input with the fixed.
The test passes if the diff agrees with a golden copy <the-source>.diff

test_simple.py is the rough hack poc.
The common bits of test_simple.py were exported to harness.py's simple_run function
test_simple_2.py uses the harness to do what test_simple.py did by calling simple_run

Having files as input will make adding tests easier and  allow greater flexibility on the encodings 

To run, run py.test in the pytests dir.

### DIFF
--- a/pytests/__init__.py
+++ b/pytests/__init__.py
@@ -1,0 +1,6 @@
+#
+#                     The LLVM Compiler Infrastructure
+#
+# This file is distributed under the University of Illinois Open Source
+# License. See LICENSE.TXT for details.
+#
--- a/pytests/harness.py
+++ b/pytests/harness.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python
+#
+#                     The LLVM Compiler Infrastructure
+#
+# This file is distributed under the University of Illinois Open Source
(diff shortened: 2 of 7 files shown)


kimgr:
Thank you for showing this off.

Our current `fix_includes.py` tests are all unittests, but these seem to be more end-to-end in nature.

Some of the current unit tests would probably make more sense in this form, and there's a battery of tests we could write more easily if we had an end-to-end framework.

But I don't think this is it, it looks pretty fragmented.

We've discussed using LLVM's `lit` and `OutputCheck` for IWYU testing at large, and I think that combination would work for `fix_includes.py` as well. There are a couple of problems to iron out, but I'd rather spend energy on that than trying to bring in pytest.

trixirt:
That's fine

PR #617, opened 2018-12-06
Define and use UF_Macro to determine when a use is a macro

trixirt:
Added this in another development effort to track which uses were macros.
Turned out that it wasn't needed for that effort.
However it may be useful in the future so I am posting it.


kimgr:
We only use use-flags to transport context information about the use that's otherwise been lost and is needed for analysis. Since there appears to be no need anymore, I don't see a reason to take this patch. I think you can also recognize macro uses in that their decl is null.

PR #618, opened 2018-12-06
Inherit the symbol uses of included files

trixirt:
See the posted test.
iwyu fails to include another includes dependencies.

There are two problems.
1. uses are ignored if they don't happen in the main files.
2. adding an include to the main file does not also add the include files dependencies.

This change fixes these problem but break many of the regression tests.
Which isn't surprising because removing includes has become more restrictive.

I believe this would be solved in a cleaner fashion if iwyu did multiple runs overs the source and modified source until a stable, compilable version was heuristically settled on.   This is a big job so it was not what i used.  If there is an interest, i'll give a try.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -2033,15 +2033,71 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
   // On the other hand, if foo.h used to have it but is removing it,
   // we *do* need to add it.
   set<string> associated_desired_includes = AssociatedDesiredIncludes();
-
-  CalculateIwyuViolations(&symbol_uses_);
-  EmitWarningMessages(symbol_uses_);
+  auto SU = symbol_uses_;
+  CalculateIwyuViolations(&SU);
   internal::CalculateDesiredIncludesAndForwardDeclares(
-      symbol_uses_, associated_desired_includes, kept_includes_,  &lines_);
+      SU, associated_desired_includes, kept_includes_, &lines_);
+
+  //
+  // If a #include file is kept, then all of it's symbol uses are kept.
+  // All of the included file symbols may not be in the current file.
+  // So for each desired include file add it's symbols to the
+  // current file's and recalculate the needed includes.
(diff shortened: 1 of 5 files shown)


kimgr:
This is exactly the kind of implicit dependency iwyu tries to kill off. I wonder if iwyu would already do what you want if you `--check_also=symbol_uses_a.h`

kimgr:
At my previous employer, somebody built a nice IWYU toolchain that ran IWYU on every header file and .cpp file in the code base, and produced a web report for inspection/statistics.

We settled on a process where we would fix free-standing headers first (there's more bang for the buck here, as they're usually used by many .cpp files and their associated headers), and then move on to each component (.cpp/.h pair), fix the header first and then the .cpp file.

This was a nice systematic process that we could run over a number of months to slowly remove incidental coupling and excess includes without too many surprises.

I think it would be nice if the IWYU tool ecosystem supported that process (e.g. by making it easier to derive compile commands for individual headers).

I'm not so sure, on the other hand, that trying to convince IWYU to do this automatically is feasible -- one header through the lens of one compilation unit can be entirely different when analyzed as part of another compilation unit.

So with that preconception (and I could be wrong!), I'm having a hard time getting my head around the soundness of this change. I suspect it adds complexity that will ultimately generate more complexity.

trixirt:
My assumption is that the includes will not change.
For my usecase, the ultimate users of iwyu will be external developers.
They will have there code base and several sdk's.
Whereas they could change their own includes, they would not be able change the sdk's.

This change allows iwyu to fail less often for some non ideal includes.  
The no reorder change is similar in that it handles non ideal includes.

Some wags at effectiveness.
+50% success using no reorder change
+10% success using inherit includes change.

Is there some generalizing infra I could add to help?
Maybe something like a general compiler pass and these could be passes ?

kimgr:
I don't think this is something we should pursue, it seems to go directly against IWYU's philosophy.

kimgr:
Closing.

PR #621, opened 2018-12-09
[iwyu_tool] Model blocking more faithfully in tests

kimgr:
No functional change, but the test suite is twice as fast.

### DIFF
--- a/iwyu_tool_test.py
+++ b/iwyu_tool_test.py
@@ -22,7 +22,6 @@ except ImportError:
 
 class MockProcess(object):
     def __init__(self, block, content):
-        self.block = block
         self.content = content
         self.complete_ts = time.time() + block
 
@@ -32,7 +31,9 @@ class MockProcess(object):
         return 0
 
     def get_output(self):
-        self.poll()
+        remaining = self.complete_ts - time.time()
+        if remaining > 0:
+            time.sleep(remaining)
         return self.content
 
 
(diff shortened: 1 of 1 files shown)


kimgr:
@pfaffe Please take a look, this is a re-take on your process mocks.

iwyu_tool_test.py
```diff
@@ -32,7 +31,9 @@ def poll(self):
         return 0
 
     def get_output(self):
-        self.poll()
+        remaining = self.complete_ts - time.time()
```
pfaffe: Instead of waiting here, why not return something unexpected, like `None`? Simplifying a bit, the process semantics are that you only get the proper output once it's done. If you sleep here you're not matching that, i.e. calling get_output on a MockProcess always gets you the full output, whereas on a real process it does not.

kimgr: `None` is probably not going to work -- `Process.get_output` always returns something, if only an empty string.

But the way the process is used/called in the main `execute` loop, `get_output` is only ever called if `proc.poll()` returns 0, which the `MockProcess` only does after the block time has expired.

So in the current test suite and for the current implementation, the `sleep` is pointless, but it does mimic the behavior of `Process` (block until output is available, then return it).

kimgr: > calling get_output on a MockProcess always gets you the full output, whereas on a real process it does not.

I don't really understand this. `Process.get_output` will wait until the process exits and then return all its output, no?

pfaffe: Actually you're right here. I must have misremembered how I implemented the Process.

kimgr:
Thank you, merged!

PR #622, opened 2018-12-09
[iwyu_tool] Improve MSVC-style driver detection

kimgr:
Match cl.exe without case on Windows to allow for e.g. CL.EXE.

Also match 'clang-cl', which could presumably be used as a cross-tool on
non-Windows systems.

Add tests.

(Note: this conflicts slightly with PR #620, but I'll figure that out when merging.)

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -120,6 +120,21 @@ def is_subpath_of(path, parent):
     return suffix.startswith(os.path.sep)
 
 
+def is_msvc_driver(compile_command):
+    """ Return True if compile_command matches an MSVC CL-style driver. """
+    compile_command = normcase(compile_command)
+
+    if compile_command.endswith('cl.exe'):
+        # Native MSVC compiler or clang-cl.exe
+        return True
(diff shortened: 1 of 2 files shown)


iwyu_tool.py
```diff
@@ -89,6 +89,30 @@ def clang_formatter(output):
...
+
+    if compile_command.endswith('cl.exe'):
+        # Native MSVC compiler or clang-cl.exe
+        return True
+
+    if compile_command.endswith('clang-cl'):
+        # Cross clang-cl on non-Windows
+        return True
+
+    return False
```
jru: It can get simpler by changing it to `return compile_command.endswith(('cl.exe', 'clang-cl'))`

kimgr: Neat! I didn't know `endswith` accepted a tuple. But that form would make it hard to place the explanatory comments, I think I prefer the current form (which explains how cl.exe also matches clang-cl.exe, and why clang-cl without extension is relevant).

PR #624, opened 2018-12-11
Track use of namespace in using directive decl

trixirt:
a.h:
namespace a { ... };

b.cpp:
include "a.h"
using namespace a;
...

Add a test that header is not removed because of using directive

Adjust badinc.cc test for new diagnostic output.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -212,6 +212,7 @@ using clang::TypedefNameDecl;
 using clang::TypedefType;
 using clang::UnaryExprOrTypeTraitExpr;
 using clang::UsingDecl;
+using clang::UsingDirectiveDecl;
 using clang::UsingShadowDecl;
 using clang::ValueDecl;
 using clang::VarDecl;
@@ -3797,6 +3798,20 @@ class IwyuAstConsumer
       ReportDeclForwardDeclareUse(CurrentLoc(), specialized_decl);
 
     return Base::VisitClassTemplateSpecializationDecl(decl);
+  }
+
+  // Track use of namespace in using directive decl
+  // a.h:
+  //   namespace a { ... };
+  //
+  // b.cpp:
+  //   include "a.h"
+  //   using namespace a;
+  //   ...
+  bool VisitUsingDirectiveDecl(clang::UsingDirectiveDecl *decl) {
+    if (CanIgnoreCurrentASTNode())  return true;
+    ReportDeclUse(CurrentLoc(), decl->getNominatedNamespaceAsWritten());
+    return Base::VisitUsingDirectiveDecl(decl);
   }
 
   // If you say 'typedef Foo Bar', then clients can use Bar however
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -136,6 +136,7 @@ int i1_macro_symbol_with_value_and_value2_var2;
 // re-opened in many different files.  So let this go, even though all
 // uses of this namespace are in badinc-i1.h; we'll get the iwyu
 // violation later when we try to use symbols from i1_ns2.
+// IWYU: i1_ns2 is defined in...*which isn't directly #included.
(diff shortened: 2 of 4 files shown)


kimgr:
Very nice, thank you! I was sincerely surprised that this wasn't already supported. One micro-nit inline, otherwise looks good!

iwyu.cc
```diff
@@ -3799,6 +3800,20 @@ class IwyuAstConsumer
...
+  // Track use of namespace in using directive decl
+  // a.h:
+  //   namespace a { ... };
+  //
+  // b.cpp:
+  //   include "a.h"
+  //   using namespace a;
+  //   ...
+  bool VisitUsingDirectiveDecl(clang::UsingDirectiveDecl *decl) {
+    if (CanIgnoreCurrentASTNode())  return true;
```
kimgr: One space too many before `return`

kimgr: Actually, nevermind. I see this is almost consistently wrong. Let's merge this as-is, and I can follow up with a refactor commit to fix all lines.

kimgr:
Nevermind, this looks good to me! Merging as-is.

jru:
It is a bit too late now, but still, I am not convinced that this change is a good idea.

I can see how tracking using directives can help have a more complete analysis, but I don't agree that the outcome of it should be to require we include files that contain the namespace definitions.

For example, look at what happens if the same namespace is in different files:

```cpp
// file namespace-A1.h
#include "namespace-A2.h"
namespace A
{
    void func1() {
        func2();
    }
}

//file namespace-A2.h
namespace A
{
    void func2();
}

// file using-namespace.cc

#include "namespace-A1.h"
using namespace A;
```
```
using-namespace.cc should add these lines:
#include "namespace-A2.h"  // for A

using-namespace.cc should remove these lines:
- #include "namespace-A1.h"  // lines 2-2

The full include-list for using-namespace.cc:
#include "namespace-A2.h"  // for A
---
```
For some reason it chooses the file included indirectly (A2) to satisfy the "using namespace A". But, it would be equally bad if it said that both (A1 and A2) were required. 

In my opinion, the solution is not to require that using directives include anything, but rather the opposite: 
If after the IWYU analysis there is no include left that provides namespace A, *then* all the using directives that refer to A should also be removed. 

That was, by the way, what IWYU did before this change, breaking the compile. And maybe what motivated it in the first place?

kimgr:
@jru Thank you, I can agree to that analysis.

That is, there's no use keeping a `using` directive if no decl use depends on it, and there's no use keep an include that's only there to satisfy a `using` directive.

So it seems like the right thing to do would be to remove unused `using` directives?

@trixirt I think I'd like to revert this commit to get us back to the status quo, and we can consider other approaches. OK?

trixirt:
The test should be kept and the tests for the other case added.

kimgr:
@trixirt @jru I'm not sure what to do about the new testcase. There are basically two cases here:

```
// decl.h
namespace ns {
   void foo();
}

// unused.h
namespace unused {
  void bar();
}

// use.cc
#include "decl.h"
#include "unused.h"

using namespace ns;
using namespace unused;

void use() {
  foo();  // usincode g ns::foo, so decl.h needs to stay
}

// No use of anything inside `unused`, so remove unused.h
```

The problem, as indicated by @trixirt, is that removing unused.h will make the code not compile (because the using directive is now referencing an undeclared namespace).

So I'm starting to think @trixirt's patch is right -- if users have excess using directives, they will have to live with excess includes (otherwise we need to perform analysis on using directives, and we don't really do that yet.)

Thoughts?

jru:
Hi,

Yes, breaking the compile is obviously wrong, and this change avoids that. 
My point was just that in order to really fix it we should have taken a diferent approach. However, if recomending the extra headers is preferable for the time being, I am good to keep it until a better solution is found.

PR #625, opened 2018-12-12
Fix a link errror by adding libClangSerialization to link

trixirt:
clang::PCHContainerOperations::PCHContainerOperations() moved to this
library with this commit in clang

commit 6be5408471f9a5f03e45f0cfe3c5bcdbfab4d707
Author: Richard Trieu <rtrieu@google.com>
Date:   Wed Dec 12 02:53:59 2018 +0000

    Move PCHContainerOperations from Frontend to Serialization

    Fix a layering violation.  Frontend depends on Serialization, so anything used
    by both should be in Serialization.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,7 @@ target_link_libraries(include-what-you-use
   clangSema
   clangFrontend
   clangDriver
+  clangSerialization
   )
 
 # Platform dependencies.


kimgr:
I installed the Clang package from SVN r348915 and can't reproduce the problem. Looks like there were a series of follow-up commit after the one quoted above that fixed this dependency Clang-side. Try and build/install the latest Clang and let me know if that resolves.

trixirt:
yes. today's clang is fine, no need for this change now.

PR #626, opened 2018-12-13
POC to preserve the original include order

trixirt:
Once a working list of includes if found both iwyu and fix_includes
reorders the list in roughly alpha order.

Reording is wrong.

This change preserves the original order.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1793,51 +1793,6 @@ OutputLine PrintableIncludeOrForwardDeclareLine(
     GetSymbolsSortedByFrequency(line.symbol_counts()));
 }
 
-enum class LineSortOrdinal {
-  PrecompiledHeader,
-  AssociatedHeader,
-  AssociatedInlineDefinitions,
-  QuotedInclude,
-  CHeader,
-  CppHeader,
-  OtherHeader,
-  ForwardDeclaration
-};
-
-using LineSortKey = pair<LineSortOrdinal, string>;
-
-LineSortOrdinal GetLineSortOrdinal(const OneIncludeOrForwardDeclareLine& line,
-                                   const set<string>& associated_quoted_includes,
-                                   const IwyuFileInfo* file_info) {
-  if (!line.IsIncludeLine())
-    return LineSortOrdinal::ForwardDeclaration;
-  if (file_info && file_info->is_pch_in_code())
-    return LineSortOrdinal::PrecompiledHeader;
-
-  const std::string quoted_include = line.quoted_include();
-  if (ContainsKey(associated_quoted_includes, quoted_include)) {
-    if (EndsWith(quoted_include, "-inl.h\""))
-      return LineSortOrdinal::AssociatedInlineDefinitions;
-    return LineSortOrdinal::AssociatedHeader;
-  }
-
-  if (GlobalFlags().quoted_includes_first && EndsWith(quoted_include, "\""))
-    return LineSortOrdinal::QuotedInclude;
-  if (EndsWith(quoted_include, ".h>"))
-    return LineSortOrdinal::CHeader;
-  if (EndsWith(quoted_include, ">"))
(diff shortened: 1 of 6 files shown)


trixirt:
Changing the include order is a critical problem.

But the current sorted order is entrenched.
So I am putting this change out as starting point to reach a consensus.
Maybe need a cmdline arg to pick one behavior over the other ?
Maybe just ditching the sorted order in iwyu because fix_includes.py does this ?

I also see a need to iwyu to have a in-place editing of the file like the -i that clang-format.
Anyone working on this ?

trixirt:
I am working on the no_reorder option and have run into a design issue.
The output of no_reorder will be similar to the default but have a different meaning.
The changes to fix_includes.py would be ugly and hard to maintain.

What i propose is instead of working directly with stdout, that iwyu outputs changes to a file that fix_includes.py consumes.  Other s2s programs like clang-format and clang-tidy do this.  Is there any preferences to which format is used ?

My preference is xml similar to clang-format.
xml is general enough that it would fit the existing output as well what i am cooking up for no reorder.

kimgr:
> The output of no_reorder will be similar to the default but have a different meaning.
> The changes to fix_includes.py would be ugly and hard to maintain.

Can you expand on this a little?

kimgr:
`fix_includes.py` already has a `--noreorder` switch. Would it be workable to require that users specify that when they're processing output from `include-what-you-use` with `--no_reorder`?

I'm hesitant to commit to a new file format, especially XML-based, as output from `include-what-you-use`. I would much prefer to make `include-what-you-use` generate a plain diff and remove the need for `fix_includes.py` entirely. I'm not entirely sure what the current division of responsibilities is.

trixirt:
If i read it correctly, fix_includes' --noreorder does not reorder within the big sets of types of includes.  This is not the same as doing no reording at all. 

I do not have strong feelings one way or the other about xml, it was just the easiest to implement output file format that could support both the original and the noreorder output. I looked into a diff output, while it would work for noreorder output, it would not work for normal output.  Normal output depends on sets of include files and heuristics within fix_includes. 

The basic thing i think iwyu needs do is to output its changes to a file.  Then have fix_includes reads that file.  This would simplify fix_includes as it would not have do any parsing of iwyu's stderr. 

Or with a lot more work, the output file format matched another clang tools, ex/ clang-tidy & clang-apply-replacements  then fix_includes could go away.

kimgr:
I've been going back and forth on this -- but I think ultimately I agree that this is the right thing to do: IWYU should not impose an include order.

The argument for sorting includes with associated header first, then project includes and then system includes at the end is that it helps smoke out implicit include dependencies. But if IWYU is doing a good enough job, those dependencies should already be fixed.

So off the top of my head, I think we should leave `#include` sorting to a separate tool (like e.g. `clang-format` or LLVM's `sort_includes.py`). Let's see if @jru feels differently.

If we can agree on a general direction, we'd need to somehow fix up the test suite not to assume the current order (your current patch breaks 20% of the regression suite).

Assuming we go with some variation on your original patch, we don't need to reinvent the protocol between `include-what-you-use` and `fix_includes.py`, right?

Flamefire:
> So off the top of my head, I think we should leave #include sorting to a separate tool 

:+1: Keep this tool focused on what its job is :)

In my imagination the output of IWYU would be a machine readable but good-enough-for-humans file. Some structure like:
```
- File
    ToAdd:
      - Header, Reason
      - ...
   ToRemove:
     - File1
     - File2
     ...
- File2...
```

This can be trivially represented in C++ and e.g. YAML (clang-apply-replacements already uses yaml and yaml is way more concise than XML). Serialization YAML<->C++ should also be easy (as support is somewhere in clang already). No (manual) parsing is then required in fix_includes and the like and a (better) human readable version (e.g. current format) can also be generated from that if required.

kimgr:
No forward motion on this for a while. Closing.

PR #628, opened 2018-12-25
Update readme - on some platform, it may be lib64

mykaul:
The path to include files may not be /usr/lib but /usr/lib64 and accordingly you should create the dirs for IWYU to run properly
(Example, Fedora 29)


kimgr:
Trying with #631 instead.

Issue #612, opened 2018-11-21
iwyu_tool.py does not attend to clang flags


alacasta:
# Enhancement proposal
It may be required in some cases to enable the possibility of using additional clang flags in combination with iwyu_tool. One example is when needing the following flag
```
-target i686-w64-mingw32
```
to be explicetly used by include-what-you-use.

# Background
include-what-you-use is usually called with the following arguments

```
include-what-you-use [-Xiwyu --iwyu_opt]... <clang opts> 
```

However, `iwyu_tool` limits its usage only to `-Xiwyu` assuming that calls like
```
ìwyu_tool.py -p . -- arg1
```
automatically translates the information into

```
-Xiwyu arg1
```
arguments to the `include-what-you-use` call. This is a design decission in favor of simplifying the calls of the tool:
```python
    # Force -Xiwyu prefix to iwyu_args so users don't have to provide prefix
    # explicitly.
    prefixes = ['-Xiwyu'] * len(iwyu_args)
    iwyu_args = list(sum(zip(prefixes, iwyu_args), ()))
```

# Possible modifications
The enhancement that i would like to be considered can be implemented respecting the current implementation.

By keeping `--` as an implicit separator of `-Xiwyu` flags and then include a `-clang` separator for clang specific flags, both can live at the same time. Pros include the non-breaking front-end specification enabling at the same time the possibility of postfixing the clang flags.

kimgr:
Yes, I think this was the wrong decision on my part, to assume that trailing flags would always be IWYU flags. I think we should make a breaking change for next version where everything after `--` is passed verbatim to `include-what-you-use`, so IWYU flags need to be explictly specified with `-Xiwyu`.

alacasta:
👍 IMO this would be the best option if not needed to keep backwards comptibility as suggested from my side.

Is it okay if I prepare something in that way?

kimgr:
That would be great, thank you!

PR #629, opened 2018-12-26
[Iwyu_tool] Remove -Xiwyu implicit prefix

alacasta:
# What?

iwyu_tool included `-Xiwyu` prefix to all those arguments comming after the separator `--`. While this had the advantage of not requiring users to provide this flag explicitely, the assumption was that all those arguments were `iwyu` arguments. 

Considering the current iwyu interface 

```
include-what-you-use [-Xiwyu --iwyu_opt]... <clang opts> <source file>
```

The change purposed here goes in the direction of removing this implicitely setting in favor of being able to provide also other type of arguments that iwyu support like `<clang opts>` 

Issue: iwyu_tool.py does not attend to clang flags #612

# How?

The previous implicit handling of the arguments

```
    # Force -Xiwyu prefix to iwyu_args so users don't have to provide prefix
    # explicitly.
    prefixes = ['-Xiwyu'] * len(iwyu_args)
    iwyu_args = list(sum(zip(prefixes, iwyu_args), ()))
```

has been removed so that `iwyu_args` are passed verbatim

Additional tests covering how the argumentes are processed have been also included

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -381,8 +381,8 @@ def main(compilation_db_path, source_files, verbose, formatter, jobs,
     return execute(invocations, verbose, formatter, jobs)
 
 
-def _bootstrap():
-    """ Parse arguments and dispatch to main(). """
+def _bootstrap(sys_argv):
+    """ Parse sys_argv arguments and dispatch to main(). """
 
     # This hackery is necessary to add the forwarded IWYU args to the
     # usage and help strings.
@@ -433,11 +433,13 @@ def _bootstrap():
             return argv[:double_dash], argv[double_dash+1:]
         except ValueError:
             return argv, []
-    argv, extra_args = partition_args(sys.argv[1:])
+    argv, extra_args = partition_args(sys_argv[1:])
     args = parser.parse_args(argv)
-    sys.exit(main(args.dbpath, args.source, args.verbose,
-                  FORMATTERS[args.output_format], args.jobs, extra_args))
+    main(
+        args.dbpath, args.source, args.verbose,
+        FORMATTERS[args.output_format], args.jobs, extra_args
+        )
 
 
 if __name__ == '__main__':
-    _bootstrap()
+    sys.exit(_bootstrap(sys.argv))
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -362,7 +362,7 @@ def execute(invocations, verbose, formatter, jobs):
 
 
 def main(compilation_db_path, source_files, verbose, formatter, jobs,
-         iwyu_args):
+         extra_args):
     """ Entry point. """
 
     try:
@@ -375,7 +375,7 @@ def main(compilation_db_path, source_files, verbose, formatter, jobs,
 
     # Transform compilation db entries into a list of IWYU invocations.
     invocations = [
-        Invocation.from_compile_command(e, iwyu_args) for e in compilation_db
+        Invocation.from_compile_command(e, extra_args) for e in compilation_db
     ]
 
     return execute(invocations, verbose, formatter, jobs)
@@ -433,10 +433,10 @@ def _bootstrap():
             return argv[:double_dash], argv[double_dash+1:]
         except ValueError:
             return argv, []
-    argv, iwyu_args = partition_args(sys.argv[1:])
+    argv, extra_args = partition_args(sys.argv[1:])
     args = parser.parse_args(argv)
     sys.exit(main(args.dbpath, args.source, args.verbose,
-                  FORMATTERS[args.output_format], args.jobs, iwyu_args))
+                  FORMATTERS[args.output_format], args.jobs, extra_args))
 
 
 if __name__ == '__main__':
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -329,12 +329,6 @@ def _bootstrap():
             return argv, []
     argv, iwyu_args = partition_args(sys.argv[1:])
     args = parser.parse_args(argv)
-
-    # Force -Xiwyu prefix to iwyu_args so users don't have to provide prefix
-    # explicitly.
-    prefixes = ['-Xiwyu'] * len(iwyu_args)
-    iwyu_args = list(sum(zip(prefixes, iwyu_args), ()))
-
     sys.exit(main(args.dbpath, args.source, args.verbose,
                   FORMATTERS[args.output_format], args.jobs, iwyu_args))
 
--- a/iwyu_tool_test.py
+++ b/iwyu_tool_test.py
@@ -12,6 +12,7 @@ import sys
 import time
 import random
 import unittest
+import inspect
 import iwyu_tool
 
 try:
@@ -86,18 +87,30 @@ class MockIwyuToolMain(object):
         the arguments used in the called within the section tested.
     """     
     def __init__(self):
+        self._iwyu_tool_main_args = inspect.getargspec(iwyu_tool.main).args
         self.real_iwyu_tool_main = iwyu_tool.main
         iwyu_tool.main = self._iwyu_tool_main_mock
-        self._call_args = []
+        self._call_args = {}
+        for arg in self._iwyu_tool_main_args:
+            self._call_args[arg] = ''
 
     def reset(self):
         iwyu_tool.main = self.real_iwyu_tool_main
 
     def get_call_args(self):
+        """ Returns a dictionary built with the iwyu_tool.main API keys filled
+            with the arguments used in the last call.
+        """
         return self._call_args   
 
-    def _iwyu_tool_main_mock(self, *args):
-        self._call_args = args
+    def _iwyu_tool_main_mock(self, *args, **kwargs):
+        assert len(args) + len(kwargs) <= self._iwyu_tool_main_args
+        index = 0
+        for idx, arg in enumerate(args):
+            key_idx = self._iwyu_tool_main_args[idx]
+            self._call_args[key_idx] = arg
+        for key, arg in kwargs.iteritems():
+            self._call_args[key] = arg
 
 class IWYUToolTests(unittest.TestCase):
     def _execute(self, invocations, verbose=False, formatter=None, jobs=1):
@@ -266,10 +279,10 @@ class BootstrapTests(unittest.TestCase):
 
     def test_argument_parser_sets_argument_correctly(self):
         """ Check that arguments are injected verbatim to iwyu. """
-        argv_stub = StubSysArgv(["iwyu_tool.py", "-p", ".", "--", "arg1"])
+        argv_stub = StubSysArgv(['iwyu_tool.py', '-p', '.', '--', 'arg1'])
         iwyu_tool._bootstrap()
-        iwyu_call_args = self.iwyu_tool_main_mock.get_call_args()
-        self.assertIn(["arg1"], iwyu_call_args)
+        call_args = self.iwyu_tool_main_mock.get_call_args()
+        self.assertEqual(['arg1'], call_args['iwyu_args'])
 
     def test_argument_parser_does_include_iwyu_flags(self):
         """ Check that arguments that appear after the '--' delimiter are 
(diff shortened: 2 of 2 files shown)


alacasta:
CI is not able to run these tests because of #630

alacasta:
Ci is now running fine once the Mock dependency has been removed. An alternative monkey patch implementation is proposed.

iwyu_tool.py
```diff
@@ -435,12 +435,6 @@ def partition_args(argv):
             return argv, []
     argv, iwyu_args = partition_args(sys.argv[1:])
```
kimgr: It would be nice to rename `iwyu_args` to `extra_args` now that they're not strictly -Xiwyu args.

But please make that refactor a separate commit, as it touches several places in this file.

iwyu_tool_test.py
```diff
@@ -52,6 +52,57 @@ def will_return(self, content):
...
+        self._will_return = ''
+        self._call_args = []
+
+    def reset(self):
+        iwyu_tool.main = self.real_iwyu_tool_main
+
+    def get_call_args(self):
+        return self._call_args
+
+    def will_return(self, content):
```
kimgr: This is unused, and can be removed

iwyu_tool_test.py
```diff
@@ -52,6 +52,57 @@ def will_return(self, content):
...
+
+class MockIwyuToolMain(object):
+    """ Provides an object that configures a mock'd behaviour
+        of the iwyu_tool.main method, being possible to obtain
+        the arguments used in the called within the section tested.
+    """     
+    def __init__(self):
+        self.real_iwyu_tool_main = iwyu_tool.main
+        iwyu_tool.main = self._iwyu_tool_main_mock
+        self._will_return = ''
```
kimgr: Unused

iwyu_tool_test.py
```diff
@@ -208,6 +259,40 @@ def test_consecutive_spaces(self):
...
+        self.iwyu_tool_main_mock =  MockIwyuToolMain()
+        self.sys_exit_stub = StubSysExit()
+        self.sys_exit_stub.will_return(0)
+
+    def tearDown(self):
+        self.iwyu_tool_main_mock.reset()
+        self.sys_exit_stub.reset()
+
+    def test_argument_parser_sets_argument_correctly(self):
+        """ Check that arguments are verbatim injected to iwyu. """
```
kimgr: verbatim injected -> injected verbatim

iwyu_tool_test.py
```diff
@@ -208,6 +259,40 @@ def test_consecutive_spaces(self):
...
+        argv_stub = StubSysArgv(["iwyu_tool.py", "-p", ".", "--", "arg1"])
+        iwyu_tool._bootstrap()
+        iwyu_call_args = self.iwyu_tool_main_mock.get_call_args()
+        self.assertIn(["arg1"], iwyu_call_args)
+
+    def test_argument_parser_does_include_xiwyu_flag(self):
+        """ Check that -Xiwyu and --iwyu_opt arguments are passed to the iwyu call. """
```
kimgr: `--iwyu_opt` has no special meaning for IWYU, so it probably doesn't need to be called out in the test here. The important quality is that `-Xiwyu --something` is forwarded.

iwyu_tool_test.py
```diff
@@ -52,6 +52,57 @@ def will_return(self, content):
...
+        iwyu_tool.main = self.real_iwyu_tool_main
+
+    def get_call_args(self):
+        return self._call_args
+
+    def will_return(self, content):
+        self._will_return = content      
+
+    def _iwyu_tool_main_mock(self, *args):
+        self._call_args = args
```
kimgr: This would be stronger if you could remember *which* arg got which value.

I can't see an elegant way to do that, so the easiest is probably to repeat `main`'s argument list here, and store the args in individual named fields;

```
def main(self, compilation_db_path, source_files, verbose, formatter, jobs, iwyu_args):
    self.compilation_db_path = compilation_db_path
    self.source_files = source_files
    self.verbose = verbose
    self.formatter = formatter
    self.jobs = jobs
    self.iwyu_args = iwyu_args
    return 0
```

That way the tests become much more specific, e.g.

    self.assertEqual(['-Xiwyu', '--arg'], mock.iwyu_args);

iwyu_tool_test.py
```diff
@@ -208,6 +259,40 @@ def test_consecutive_spaces(self):
...
+        self.assertIn(["arg1"], iwyu_call_args)
+
+    def test_argument_parser_does_include_xiwyu_flag(self):
+        """ Check that -Xiwyu and --iwyu_opt arguments are passed to the iwyu call. """
+        argv_stub = StubSysArgv(["iwyu_tool.py", "-p", ".", "--", "arg1", "-Xiwyu", "arg2", "--iwyu_opt", "arg3"])
+        iwyu_tool._bootstrap()
+        iwyu_call_args = self.iwyu_tool_main_mock.get_call_args()
+        self.assertIn(["arg1", "-Xiwyu", "arg2", "--iwyu_opt", "arg3"], iwyu_call_args)        
+
+    def test_argument_parser_uses_the_first_separator_for_splitting_arguments(self):
```
kimgr: Try shortening this method name so it fits within 80-column width

alacasta:
Hi @kimgr 
Thanks for the review, i will be working on it. 
Before all, i'd like to know if there is a pylint-like, PEP-8 or whatever configuration file with all the requirements about format like this you talk about. It's just to be sure about what style guidelines are required before commiting my code to be reviewed.

iwyu_tool_test.py
alacasta: Done ✔️

iwyu_tool_test.py
alacasta: Done ✔️

iwyu_tool_test.py
alacasta: Done ✔️

iwyu_tool_test.py
alacasta: Done ✔️ 
Absolutely right. I misunderstood the meaning of 'iwyu_opt' flag.

iwyu_tool_test.py
alacasta: Done ✔️

iwyu_tool_test.py
alacasta: Done ✔️ 
# Details:
I like the idea but i think that finding a mechanism for respecting the (probably changing in time) API may allow others not to fall into waste of time trying to find why tests may fail. 
The solution for this has been to inspect the iwyu_tool.main API and from that point, to create two sepparated attributes for the mock, one containing all these arguments and another one (a dictionary) that will be filled with the arguments of the call. 
Hence, you can get the arguments with something like
```python
       call_args = self.iwyu_tool_main_mock.get_call_args()
       self.assertEqual(['arg1'], call_args['iwyu_args'])
```
I liked the possibility of adding the attributes directly to the mock, but i think that accessing the mock attributes with the name of the arguments considering that they are the arguments used in the last call is less standard.
Let's see how do you see this option but i think that is an elegant approach.

iwyu_tool.py
alacasta: Done ✔️ 
I like the suggestion. Ii've applied this into the whole iwyu_tool.

iwyu_tool.py
```diff
@@ -435,12 +435,6 @@ def partition_args(argv):
...
     args = parser.parse_args(argv)
-
-    # Force -Xiwyu prefix to iwyu_args so users don't have to provide prefix
-    # explicitly.
-    prefixes = ['-Xiwyu'] * len(iwyu_args)
-    iwyu_args = list(sum(zip(prefixes, iwyu_args), ()))
-
     sys.exit(main(args.dbpath, args.source, args.verbose,
```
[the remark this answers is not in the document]
alacasta: Not applied ❌ (See comment)

iwyu_tool_test.py
```diff
@@ -52,6 +52,57 @@ def will_return(self, content):
...
+        sys.exit = lambda x: self._will_return
+        self._will_return = 0
+
+    def reset(self):
+        sys.exit = self.real_sys_exit
+
+    def will_return(self, content):
+        self._will_return = content   
+
+class StubSysArgv(object):
```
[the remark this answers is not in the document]
alacasta: Not applied ❌ (See comment)

alacasta:
Hello @kimgr 
Thanks for the review. I have applied all the changes in three different commits:

1. Typos, style format and unused variables removed
2. Modification on how the iwyu_tool.main() mock register the arguments from a callee has been introduced so that checks are made using directly the API naming instead of 'implicit' order on the call. (further details can be found in the conversation).
3. Refactor of the iwyu_args to extra_args has been applied

However, i consder that the first two changes you propose (bascally, moving sys calls to main) are out of the scope of the current change. From my side, it is not a problem to create this stubs and if in a future is considered better to move sys to main, then we can just remove all this stubs. 
One of my main concerns about this is how the `__main__` is built. There are spare logic in _bootstrap and in main(). IMHO, I'd propose to introduce the _bootstrap logic into main() perhaps splitting all the argument logint in different parts, but at the end, conclding in something like

```python
if __name__ == '__main__':
    sys.exit(main(sys.argv))
``` 
So, I'd like to know your opinion on this:
- Should i definitively remove the sys logic from _bootstrap just not to need to mocked this components as i already did?
- Alternatively, should we keep everything as it and perhaps (if you think that this is needed, i can open an issue for this) this can be handle in a future task

Thanks!

kimgr:
Thanks! I won't have time to look at the patch today because New Year's Eve :fireworks:, but I thought I'd answer some of your questions briefly:

> Before all, i'd like to know if there is a pylint-like, PEP-8 or whatever configuration file with all the
> requirements about format

`iwyu_tool.py` should follow standard PEP8 (with an exception for line-width -- PEP8 defaults to 79, we use 80). `pep8 iwyu_tool.py` finds a few complaints but mostly long lines.

> However, i consder that the first two changes you propose (bascally, moving sys
> calls to main) are out of the scope of the current change. 

Sort of. But they're trivial changes that improve testability, and the diff would be comparatively tiny.

> if __name__ == '__main__':
>    sys.exit(main(sys.argv))

I wouldn't want `main` to take an argv list. As it currently works, you can `import iwyu_tool` as a module and invoke `main` as a regular function. If you change it to take a list, callers would need to compose that list before calling `main`, which would be a weird interface for a function.

> Should i definitively remove the sys logic from _bootstrap just not to need to mocked this components > as i already did?

Yes, I think that would be better. All you need to do is re-shape it as:
```
if __name__ == '__main__':
    sys.exit(_bootstrap(sys.argv))
```

and testability would be much improved. The only reason it wasn't like this already is because `iwyu_tool.py` was originally intended as more of an example/minimal util and didn't have tests. Then it grew unexpectedely.

Happy new year!

alacasta:
Thanks!
Ok, i see the point; i will apply that changes in the implementation now that we have made modifications on this part (tests will keep clearer!).

One additional question:

> `iwyu_tool.py` should follow standard PEP8 (with an exception for line-width -- PEP8 defaults to 79, we use 80). `pep8 iwyu_tool.py` finds a few complaints but mostly long lines.

is then PEP8 used as the official standard for IWYU? If so, should not we set checks on this on CI? I think this would help developers to not ping-ponging due to style related mistakes. 

Happy new year!

iwyu_tool_test.py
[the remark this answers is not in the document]
alacasta: Done ✔️

iwyu_tool.py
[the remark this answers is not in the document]
alacasta: Done ✔️

alacasta:
Hello again, 
I've applied the modifications suggested. Basically, all the `sys` dependency has been moved into the `__main__` `_boostrap` call so that it can be easier tested. Moreover, all the stubs created in this PR have been also removed.

kimgr:
I would like a clean commit history before merging this, but I also see several nits in the code itself. Let me try and adopt your patch and shape it to my liking, so we can get this in place. Thanks for the hint on `inspect`, we'll see if it becomes necessary.

alacasta:
Maybe I can try to create one new PR with a couple of commits containing the final implementation. Just let me know.

kimgr:
Thanks, I'd rather build off your excellent work to finalize something without bandying back and forth with review comments. I'll make sure to attribute the work to you.

alacasta:
Looks good! 
The complete `test_argparse_args` use case was missing and is convinient. 

Two minor comments related to style.
- L243 - Missing white-space before 'src1' 
- L262 - Missing white-space before '-Xiwyu'

Thanks!

kimgr:
Thanks, good catch! Please see PR #634 for an updated version. I'll close this in the meantime.

PR #632, opened 2018-12-30
bail on error

trixirt:
A fairly serious problem.
Iwyu will provide output to change includes when basic parsing fails.

The link error is the same as 
https://github.com/include-what-you-use/include-what-you-use/pull/625
It happens on the debug build.


iwyu.cc
```diff
@@ -3596,6 +3596,11 @@ class IwyuAstConsumer
     ParseFunctionTemplates(context.getTranslationUnitDecl());
 
     TraverseDecl(context.getTranslationUnitDecl());
+
+    // Check if any unrecoverable errors have occurred
```
kimgr: Comments should be sentences including full stop.

iwyu.cc
```diff
@@ -3596,6 +3596,11 @@ class IwyuAstConsumer
...
+    // For ex/ No sense continuing if an #include-ed file isn't found
```
kimgr: Unusual abbreviation there, `For ex/` -- how about just `There's no sense continuing [...]`

iwyu.cc
```diff
@@ -3596,6 +3596,11 @@ class IwyuAstConsumer
...
+    if (compiler()->getDiagnostics().hasUnrecoverableErrorOccurred())
+      exit(EXIT_INVALIDARGS);
```
kimgr: I'm not so sure about the `EXIT_INVALIDARGS` here, that seems to be used to signal invalid arguments to IWYU.

But apparently we never reserved room for more failure exit codes :-(, so I think this is the best we can do.

kimgr:
The link error looks like an upstream bug, we don't depend directly on libClangSerialization, do we? I'd rather not work around bugs in Clang's build system, because that quickly turns into an arms race :)

I can't think of a way to test this with our current infrastructure, so let's leave it untested.

iwyu.cc
trixirt: Fixed in latest change.

iwyu.cc
trixirt: Fixed in latest change.

iwyu.cc
trixirt: Noted in latest change that this a least-bad option and todo-ed when the exit codes get reworked.

trixirt:
It is the nature of working clang's master that there will be churn.  As this problem happens on the debug build, it is likely not directly our problem. However our debug build fails to build and folks needing to use the debug build will need this change.

kimgr:
@trixirt I can't reproduce the link error on my Linux system. Can you provide more details? This is not something I think we should work around in IWYU, it's better to just fix it upstream if at all possible. We don't *have* a dependency on `clangSerialization`, so let's not say we do.

trixirt:
Everyday I pull/rebuild release llvm/clang/lld master and then pull/rebuild release iwyu master against that.
When I am debugging a problem i rebuild the debug version of both and have the linking problem.
The building is on ubuntu 18.04 with clang 7 from the release_70 branch that hasn't changed in a while.

kimgr:
@trixirt Thanks, I thought I understood at first, but now everything you say sounds contradictory :)

Is it branch master/master or branch release_70/clang_7.0 that's broken for you? I've built the debug version of llvm/clang:master and built the debug version of iwyu:master against those outputs, and it seems fine.

trixirt:
Note '...' replaces the fullpath to my test install area.

.../install.Linux.x86_64.debug/bin/clang++  -fPIC -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -std=c++11 -w -fdiagnostics-color -g  -Wl,-rpath-link, CMakeFiles/include-what-you-use.dir/iwyu.cc.o CMakeFiles/include-what-you-use.dir/iwyu_ast_util.cc.o CMakeFiles/include-what-you-use.dir/iwyu_cache.cc.o CMakeFiles/include-what-you-use.dir/iwyu_driver.cc.o CMakeFiles/include-what-you-use.dir/iwyu_getopt.cc.o CMakeFiles/include-what-you-use.dir/iwyu_globals.cc.o CMakeFiles/include-what-you-use.dir/iwyu_include_picker.cc.o CMakeFiles/include-what-you-use.dir/iwyu_lexer_utils.cc.o CMakeFiles/include-what-you-use.dir/iwyu_location_util.cc.o CMakeFiles/include-what-you-use.dir/iwyu_output.cc.o CMakeFiles/include-what-you-use.dir/iwyu_path_util.cc.o CMakeFiles/include-what-you-use.dir/iwyu_preprocessor.cc.o CMakeFiles/include-what-you-use.dir/iwyu_verrs.cc.o  -o bin/include-what-you-use  -Wl,-rpath,"\$ORIGIN/../lib" .../install.Linux.x86_64.debug/lib/libLLVMOption.so.8svn .../install.Linux.x86_64.debug/lib/libLLVMSupport.so.8svn .../install.Linux.x86_64.debug/lib/libLLVMX86AsmParser.so.8svn .../install.Linux.x86_64.debug/lib/libLLVMX86Desc.so.8svn .../install.Linux.x86_64.debug/lib/libLLVMX86Info.so.8svn -lpthread .../install.Linux.x86_64.debug/lib/libclangBasic.so.8svn .../install.Linux.x86_64.debug/lib/libclangLex.so.8svn .../install.Linux.x86_64.debug/lib/libclangAST.so.8svn .../install.Linux.x86_64.debug/lib/libclangSema.so.8svn .../install.Linux.x86_64.debug/lib/libclangFrontend.so.8svn .../install.Linux.x86_64.debug/lib/libclangDriver.so.8svn

/usr/bin/ld: CMakeFiles/include-what-you-use.dir/iwyu_driver.cc.o: undefined reference to symbol '_ZN5clang22PCHContainerOperationsC1Ev'
.../install.Linux.x86_64.debug/lib/../lib/libclangSerialization.so.8svn: error adding symbols: DSO missing from command line
clang-8: error: linker command failed with exit code 1 (use -v to see invocation)

kimgr:
Oh, it's a `BUILD_SHARED_LIBS` build? Can you also share your `cmake` line?

kimgr:
Anyway, I'm not interested in taking the CMake patch, but the fatal error reporting is very welcome. Can you split them apart so we can argue about them separately? :)

trixirt:
link fix moved out of this branch and into
https://github.com/trixirt/include-what-you-use/commit/db268293e3bc8db5e228199b1176d17fdecbbf8b

trixirt:
Is anything else needed for this change ?

iwyu.cc
```diff
@@ -3596,7 +3596,18 @@ class IwyuAstConsumer
...
 
     TraverseDecl(context.getTranslationUnitDecl());
-   
+
+    // Check if any unrecoverable errors have occurred.
+    // There is no point in continuing when the AST is in a bad state.
+    //
+    // EXIT_INVALIDARGS is not a great choice for the return status
+    // because a compile error will not have a strong connection to the
+    // command line arguements, but there are only 2 error codes and
```
kimgr: typo: arguements

trixirt: fixed

kimgr:
Thank you!

PR #635, opened 2019-01-05
In fix_includes, treat associated headers as main-CU headers for sorting

jbytheway:
`fix_includes.py`, when sorting headers, will put main-CU #include lines first.

It should respect `IWYU pragma: associated` as a way for the user to specify what the main-CU files are when its heuristics fail (this pragma is [already defined](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-associated) by IWYU for this purpose).

Make it so, and add a test.


jbytheway:
Test failure appears unrelated.

kimgr:
Looks good to me! Can you squash the commits into one -- no need to commit code without tests or vice versa. Thanks!

jbytheway:
Done.

kimgr:
Looks good to me, thanks!

PR #637, opened 2019-01-13
Stop ignoring uses for builtins with mappings

jbytheway:
Currently IWYU will ignore all uses of buitlins.

This makes sense normally.  However, when a symbol has an explicit mapping defined, we should not ignore uses of it.

In particular, for example, one might require a particular header to be included when certain builtins are used, and this change allows that by defining a mapping for that builtin (whereas before all uses of builtins were ignored).

This change is originally inspired by the fact that some `cmath` functions (`std::pow`, `std::round`, perhaps more) are considered a builtin by clang with certain compiler settings, and IWYU was suggesting removing the include of `cmath`, which breaks the build.  I wanted to override this with a mapping, but it did not work.  With this change, that will now work.

Moreover, I think there is wider application to mappings for buitlins.  For example, if a buitin exists on clang but not some other compiler and you have a header which emulates it for that other compiler, you want to make sure that IWYU enforces inclusion of that header, even though it's not needed on clang.

The solution I've implemented affects `ProcessFullUse`, which contains most of the code that will ignore a use under some situation or another.  For three of the situations therein, I make it not ignore the use when the symbol is a builtin that has a mapping to a file that does not (transitively) include the file containing the use.

The three situations are:
* When it's a builtin.
* When it's defined in this file.
* When it's defined in a file including this one.
All three changes are required to make this work for buitins, because clang considers a builtin to be defined at first use, so the second and third cases can apply for uses of builtins.  

This solution does not work for all builtins; some don't get noticed as uses by IWYU at all.  But I think that's a separate issue.

I've only changed builtin functions.  There is also code in `ProcessFullUse` for builtin templates, but I have no idea when those arise, so I'm taking the conservative option and not changing any behaviour for them.

Similarly, there are probably other times when IWYU ignores uses where it might be better to respect a mapping when it exists.  I actually wrote a broader version of this change which would affect some, but I don't really have evidence that it's necessary, so again I'm being conservative and just making the changes needed for builtins.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -39,6 +39,7 @@
 #include "clang/AST/TemplateName.h"
 #include "clang/AST/Type.h"
 #include "clang/AST/TypeLoc.h"
+#include "clang/Basic/Builtins.h"
 #include "clang/Basic/SourceLocation.h"
 #include "clang/Basic/SourceManager.h"
 
@@ -1077,6 +1078,15 @@ bool DeclsAreInSameClass(const Decl* decl1, const Decl* decl2) {
   if (decl1->getDeclContext() != decl2->getDeclContext())
     return false;
   return decl1->getDeclContext()->isRecord();
+}
+
+bool IsBuiltinFunction(const clang::NamedDecl* decl,
+                       const std::string& symbol_name) {
+  if (const clang::IdentifierInfo* iden = decl->getIdentifier()) {
+    return iden->getBuiltinID() != 0 &&
+           !clang::Builtin::Context::isBuiltinFunc(symbol_name.c_str());
+  }
+  return false;
 }
 
 // --- Utilities for Type.
--- a/iwyu_ast_util.h
+++ b/iwyu_ast_util.h
@@ -655,6 +655,10 @@ const clang::NamedDecl* GetNonfriendClassRedecl(const clang::NamedDecl* decl);
 // same, and it's a class (or struct).
 bool DeclsAreInSameClass(const clang::Decl* decl1, const clang::Decl* decl2);
 
+// Returns true if the given decl/name is a builtin function
+bool IsBuiltinFunction(const clang::NamedDecl* decl,
+                       const std::string& symbol_name);
+
 // --- Utilities for Type.
 
 const clang::Type* GetTypeOf(const clang::Expr* expr);
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1111,12 +1111,25 @@ void ProcessForwardDeclare(OneUse* use,
   }
 }
 
+// Returns true if the given symbol has a mapping defined to a file.
+static bool HasMapping(const string& symbol) {
+  return !GlobalIncludePicker().GetCandidateHeadersForSymbol(symbol).empty();
+}
+
 void ProcessFullUse(OneUse* use,
                     const IwyuPreprocessorInfo* preprocessor_info) {
   CHECK_(use->decl() && "Must call ProcessFullUse on a decl");
   CHECK_(use->is_full_use() && "Must not call ProcessFullUse on fwd-decl");
   if (use->ignore_use())   // we're already ignoring it
     return;
+
+  // We normally ignore uses for builtins, but when there is a mapping defined
+  // for the symbol, we should respect that.  So, we need to determine whether
+  // the symbol has any mappings.
+  bool is_builtin_function = IsBuiltinFunction(use->decl(), use->symbol_name());
+
+  bool is_builtin_function_with_mappings =
+      is_builtin_function && HasMapping(use->symbol_name());
 
   // (B1) If the definition is after the use, re-point to a prior decl.
   // If iwyu followed the language precisely, this wouldn't be
@@ -1162,7 +1175,7 @@ void ProcessFullUse(OneUse* use,
   // All this is moot when FunctionDecls are being defined, all their redecls
   // are separately registered as uses so that a definition anchors all its
   // declarations.
-  if (!use->is_function_being_defined()) {
+  if (!use->is_function_being_defined() && !is_builtin_function_with_mappings) {
     set<const NamedDecl*> all_redecls;
     if (isa<RecordDecl>(use->decl()) || isa<ClassTemplateDecl>(use->decl()))
       all_redecls.insert(use->decl());  // for classes, just consider the dfn
@@ -1188,7 +1201,7 @@ void ProcessFullUse(OneUse* use,
     return;
   }
   // A compiler builtin without a predefined header file (e.g. __builtin_..)
-  if (IsBuiltinFunction(use->decl(), use->symbol_name()) {
+  if (is_builtin_function && !is_builtin_function_with_mappings) {
     VERRS(6) << "Ignoring use of " << use->symbol_name()
              << " (" << use->PrintableUseLoc() << "): built-in function\n";
     use->set_ignore_use();
@@ -1254,7 +1267,8 @@ void ProcessFullUse(OneUse* use,
   // the language requires).
   // TODO(csilvers): remove this when we resolve the bugs with macros/typedefs.
   if (preprocessor_info->FileTransitivelyIncludes(
-          GetFileEntry(use->decl()), GetFileEntry(use->use_loc()))) {
+          GetFileEntry(use->decl()), GetFileEntry(use->use_loc())) &&
+      !is_builtin_function_with_mappings) {
     VERRS(6) << "Ignoring use of " << use->symbol_name()
              << " (" << use->PrintableUseLoc() << "): 'backwards' #include\n";
     use->set_ignore_use();
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1188,14 +1188,11 @@ void ProcessFullUse(OneUse* use,
     return;
   }
   // A compiler builtin without a predefined header file (e.g. __builtin_..)
-  if (const clang::IdentifierInfo* iden = use->decl()->getIdentifier()) {
-    if (iden->getBuiltinID() != 0 &&
-        !clang::Builtin::Context::isBuiltinFunc(use->symbol_name().c_str())) {
-      VERRS(6) << "Ignoring use of " << use->symbol_name()
-               << " (" << use->PrintableUseLoc() << "): built-in function\n";
-      use->set_ignore_use();
-      return;
-    }
+  if (IsBuiltinFunction(use->decl(), use->symbol_name()) {
+    VERRS(6) << "Ignoring use of " << use->symbol_name()
+             << " (" << use->PrintableUseLoc() << "): built-in function\n";
+    use->set_ignore_use();
+    return;
   }
   // Special case for operators new/delete: Only treated as built-in if they
   // are the default, non-placement versions.
--- a/tests/cxx/built_ins_with_mapping.cc
+++ b/tests/cxx/built_ins_with_mapping.cc
@@ -1,0 +1,40 @@
+//===--- built_ins_with_mapping.cc - test input file for iwyu -------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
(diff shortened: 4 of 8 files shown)


kimgr:
Typo in commit message:
> This allows mapping to be defined to those to IWYU does not remove #include <cmath>.

to those to -> to those so

Overall this sentence reads a little funny. How about `This makes IWYU respect mappings for such builtins, so as not to remove #include <cmath>.`?

iwyu_output.cc
```diff
@@ -1069,6 +1069,34 @@ void ProcessFullUse(OneUse* use,
   if (use->ignore_use())   // we're already ignoring it
     return;
 
+  // We normally ignore uses for builtins, but when there is a mapping defined
+  // for the symbol, we should respect that.  So, we need to determine whether
+  // the symbol has any mappings.
+  bool is_builtin_function = false;
+  if (const clang::IdentifierInfo* iden = use->decl()->getIdentifier()) {
```
kimgr: I think it would be nice to break this code out to a helper in `iwyu_ast_util` -- the flow in `ProcessFullUse` is overwhelming enough without these details. Something like `bool IsBuiltinFunction(const clang::NamedDecl* decl, const std::string& symbol_name);`

kimgr:
I think this may have been introduced by e31cd2adda44729ad17df8446008935ae3f8b69e -- before that we did strict name checking on `__builtin_` to detect builtins. I wonder if we could just relax the builtin detection to look at compiler builtins and not library builtins? I haven't looked at/fully understand the details and the difference between the two kinds, but I think there's some distinction that we could possibly detect.

iwyu_output.cc
```diff
@@ -1069,6 +1069,34 @@ void ProcessFullUse(OneUse* use,
...
+    is_builtin_function = iden->getBuiltinID() != 0 &&
+        !clang::Builtin::Context::isBuiltinFunc(use->symbol_name().c_str());
+  }
+
+  bool is_builtin_function_with_mappings = false;
+
+  if (is_builtin_function) {
+    for (const std::string& public_file :
```
kimgr: There's some conceptual mismatch between the variable name `is_builtin_function_with_mappings` and what this code is doing. I don't yet understand the details perfectly, but spontaneously it feels like this could be triggered even if there are not strictly any mappings involved.

I'm having a hard time getting my head around the algorithm here; could you spell out an example where the cyclic detection would trigger?

Thanks!

kimgr: I think maybe this is a variant of a more general problem/algorithm, that might as well live in `iwyu_include_picker`, but I want to make sure I understand what's going on before suggesting anything potentially stupid :)

iwyu_output.cc
```diff
@@ -1069,6 +1069,34 @@ void ProcessFullUse(OneUse* use,
...
+        GlobalIncludePicker().
+          GetCandidateHeadersForSymbol(use->symbol_name())) {
+      // To avoid cyclic includes we must only count files that do not
+      // include this one.
+      const FileEntry* public_file_entry =
+        preprocessor_info->IncludeToFileEntry(public_file);
```
kimgr: Continued lines should be indented four spaces from the line above. Clang-format on the changed lines might help here.

iwyu_output.cc
```diff
@@ -1069,6 +1069,34 @@ void ProcessFullUse(OneUse* use,
...
+      if (public_file_entry == nullptr ||
+          !preprocessor_info->FileTransitivelyIncludes(
+            public_file_entry, GetFileEntry(use->use_loc()))) {
```
kimgr: See above about indentation

tests/cxx/built_ins_with_mapping.imp
```diff
@@ -0,0 +1,9 @@
+# Header mappings for IWYU tests.
+# Note that some headers listed here don't actually exist; we just want
+# IWYU to suggest using them
+[
+  { symbol: ["__builtin_expect", "private", "\"tests/cxx/built_ins_with_mapping-d2.h\"", "public"] },
+  { symbol: ["__builtin_strlen", "private", "\"tests/cxx/built_ins_with_mapping-d3.h\"", "public"] },
+  { symbol: ["__builtin_strcmp", "private", "\"tests/cxx/built_ins_with_mapping.h\"", "public"] },
```
kimgr: There's all sorts of special rules for the associated header, so maybe the test would be stronger if it used a non-associated header for this mapping/test scenario?

I'm fully prepared to accept it doesn't make a difference, but I thought I'd ask.

jbytheway: As mentioned above, I'm specifically using the associated header because there are potential issues there.

jbytheway:
> I think this may have been introduced by [e31cd2a](https://github.com/include-what-you-use/include-what-you-use/commit/e31cd2adda44729ad17df8446008935ae3f8b69e) -- before that we did strict name checking on `__builtin_` to detect builtins. I wonder if we could just relax the builtin detection to look at compiler builtins and not library builtins? I haven't looked at/fully understand the details and the difference between the two kinds, but I think there's some distinction that we could possibly detect.

I know very little of clang internals, so I wasn't sure whether this was possible.  I considered looking into it, but decided that supporting mappings for all builtins would be less surprising to users.

If you disagree, then I can look into that alternative.  It would almost certainly be a simpler change to IWYU than this one.

iwyu_output.cc
jbytheway: So, I tried moving the logic to `iwyu_include_picker` and it seemed inelegant, because it needs the proprocessor_info as well, and currently that file doesn't use any preprocessor info at all.  So instead I just pulled it out into a static function in `iwyu_output.cc`.

jbytheway:
OK, I have pushed a few changes:
* Fixed bad commit message.
* Factored out two functions.
* Ran `clang-format`.
I've left things as separate commits for now to facilitate review; can squash later as desired.

iwyu_output.cc
```diff
@@ -1062,13 +1062,43 @@ void ProcessForwardDeclare(OneUse* use,
   }
 }
 
+// Returns true if the given symbol has a mapping defined to a file, and that
```
kimgr: Thank you for this! Now I understand what this snippet is really doing.

iwyu_output.cc
```diff
@@ -1062,13 +1062,43 @@ void ProcessForwardDeclare(OneUse* use,
...
+static bool HasMappingSuitableForFile(
+    const string& symbol, const FileEntry* file,
+    const IwyuPreprocessorInfo* preprocessor_info) {
+  for (const std::string& public_file :
+       GlobalIncludePicker().GetCandidateHeadersForSymbol(symbol)) {
+    // To avoid cyclic includes we must only count files that do not
+    // include this one.
+    const FileEntry* public_file_entry =
+        preprocessor_info->IncludeToFileEntry(public_file);
+    if (public_file_entry == nullptr ||
```
kimgr: This nullptr check feels cowardly :) Let me see if I can understand what it means:

If `IncludeToFileEntry` returns `nullptr`, it means we haven't seen the public mapped header anywhere in our include graph, which means we cannot have included it. Is that it?

tests/cxx/built_ins_with_mapping.cc
```diff
@@ -0,0 +1,40 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/built_ins_with_mapping.h"
+#include "tests/cxx/built_ins_with_mapping-d1.h"
+
+// Noramlly if we use a builtin function IWYU will ignore uses of it.
```
kimgr: Typo: Noramlly

iwyu_output.cc
```diff
@@ -1062,13 +1062,43 @@ void ProcessForwardDeclare(OneUse* use,
...
+        !preprocessor_info->FileTransitivelyIncludes(public_file_entry, file)) {
```
kimgr: Hmm. I wonder if it's legit to use `FileTransitivelyIncludes` in the IWYU output stage. I see that it's been used before, but I remember running into problems in PR #594 because I couldn't be sure that the include I was basing decisions on wouldn't be added/removed later in the analysis. Or something like that.

So in this example, what if `public_file_entry` does not include `file`, but a later analysis pass adds it? Will that invalidate the analysis?

It does seem unlikely to me that we would be editing files we have mappings for, though... Thoughts?

iwyu_output.cc
jbytheway: Yes.  This happens for example in the test case for `built_ins_with_mapping-d2.h`.

tests/cxx/built_ins_with_mapping.cc
jbytheway: Fixed.

iwyu_output.cc
jbytheway: ...of course, by adding a `#include` of an unseen header we could perhaps be creating a cyclic include, but I don't think there's anything that can be done about that.

iwyu_output.cc
kimgr: Thanks for walking me through this. Let's start with the simpler same-file check, and we can circle back to this if it becomes a problem. Don't forget to update the doc comment for the function.

jru:
Hi,

I was briefly looking into this and realized that there is `unsigned FunctionDecl::getBuiltinID()` in clang. It handles a couple of special cases. Maybe it also works here?

In any case, the code in iwyu_output.cc should probably be rewritten to make use of it instead.

jbytheway:
> I was briefly looking into this and realized that there is `unsigned FunctionDecl::getBuiltinID()` in clang. It handles a couple of special cases. Maybe it also works here?
> 
> In any case, the code in iwyu_output.cc should probably be rewritten to make use of it instead.

It looks like that might indeed do a better job than what is currently being used.  That might well be a simpler fix for the `cmath` problem.  I might try it when I get the chance.  I think it's somewhat orthogonal to this issue, though.

iwyu_output.cc
jbytheway: I've had an idea for an alternative solution that might be neater.  I'm going to investigate that first.

iwyu_ast_util.cc
```diff
@@ -1079,6 +1080,15 @@ bool DeclsAreInSameClass(const Decl* decl1, const Decl* decl2) {
   return decl1->getDeclContext()->isRecord();
 }
 
+bool IsBuiltinFunction(
```
kimgr: This is correct formatting, but I think it would look nicer as:
```
bool IsBuiltinFunction(const clang::NamedDecl* decl,
                       const std::string& symbol_name) {
    ...
}
```

Does clang-format leave that alone?

iwyu_ast_util.h
```diff
@@ -655,6 +655,10 @@ const clang::NamedDecl* GetNonfriendClassRedecl(const clang::NamedDecl* decl);
 // same, and it's a class (or struct).
 bool DeclsAreInSameClass(const clang::Decl* decl1, const clang::Decl* decl2);
 
+// Returns true if the given decl / name is a builtin function
```
kimgr: No space around / in "decl/name"

jbytheway:
Some more updates:
* Rebased on master.
* Given the changes in #652 it's no longer a problem for the mapping to point to a header using the symbol, so now more files are "safe" and I've replaced `HasMappingSuitableForFile` with a simpler `HasMapping`.
* Did the two formatting changes you suggested above, and a couple more arising out of the simplification of `HasMapping`.

Questions:
* Is `HasMapping` still worth being a separate function?  I think yes; it's only one line, but it's not obvious what it's doing and giving it a name helps.
* Would you like me to squash everything down to a single commit?

kimgr:
Looks great!

I agree that `HasMapping` is warranted as you suggested.

I would prefer the `IsBuiltinFunction` refactor as one commit, and the rest as a single commit.

jbytheway:
Rebased into two commits as suggested.

kimgr:
Thank you for pushing this through, looks great!

PR #639, opened 2019-01-13
Simplify handling of defined() operator

kimgr:
Clang r350891 changed the raw lexer so it asserts for IWYU's use
case.

Instead of re-lexing to catch symbols inside defined(), use a
PPCallbacks::Defined override to capture the tokens directly
(presumably, Defined did not exist when the open-coded parser was
added).

This removes a lot of code, including the PPCallbacks for If and Elif,
which are no longer necessary.

No functional change.

### DIFF
--- a/iwyu_preprocessor.cc
+++ b/iwyu_preprocessor.cc
@@ -609,24 +609,6 @@ void IwyuPreprocessorInfo::MacroDefined(const Token& id,
   }
 }
 
-void IwyuPreprocessorInfo::If(SourceLocation loc, SourceRange condition_range,
-                              ConditionValueKind condition_value) {
-  ERRSYM(GetFileEntry(condition_range.getBegin()))
-      << " [ #if         ] "
-      << PrintableSourceRange(condition_range) << "\n";
-  CheckIfOrElif(condition_range);
-}
-
-void IwyuPreprocessorInfo::Elif(SourceLocation loc,
-                                SourceRange condition_range,
-                                ConditionValueKind condition_value,
-                                SourceLocation if_loc) {
-  ERRSYM(GetFileEntry(condition_range.getBegin()))
-      << " [ #elif       ] "
-      << PrintableSourceRange(condition_range) << "\n";
-  CheckIfOrElif(condition_range);
-}
-
 void IwyuPreprocessorInfo::Ifdef(SourceLocation loc,
                                  const Token& id,
                                  const MacroDefinition& /*definition*/) {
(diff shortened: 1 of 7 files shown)


iwyu_preprocessor.h
```diff
@@ -177,18 +177,17 @@ class IwyuPreprocessorInfo : public clang::PPCallbacks,
...
-          ConditionValueKind condition_value) override;
-  void Elif(clang::SourceLocation loc,
-            clang::SourceRange condition_range,
-            ConditionValueKind condition_value,
-            clang::SourceLocation if_loc) override;
   void Ifdef(clang::SourceLocation loc, const clang::Token& id,
              const clang::MacroDefinition& definition) override;
   void Ifndef(clang::SourceLocation loc, const clang::Token& id,
               const clang::MacroDefinition& definition) override;
+  void Defined(const clang::Token &MacroNameTok,
```
kimgr: Use IWYU style for argument names

iwyu_preprocessor.cc
```diff
@@ -645,6 +627,18 @@ void IwyuPreprocessorInfo::Ifndef(SourceLocation loc,
...
 }
 
+// Clang will give a MacroExpands() callback for all macro-tokens
+// used inside an #if or #elif, *except* macro-tokens used within a
+// 'defined' operator. They produce a Defined() callback.
+void IwyuPreprocessorInfo::Defined(const Token& id,
+                                   const MacroDefinition& /*definition*/,
+                                   SourceRange /*range*/) {
+  ERRSYM(GetFileEntry(id.getLocation()))
+      << "[ #if defined ] " << PrintableLoc(id.getLocation())
```
jru: It seems the old method used a `SourceRange` instead of a `SourceLocation`. The Token type has `getEndLoc`/`getLastLoc`. Maybe you can use it to get a range like before?

jru:
Hey, this is an awesome cleanup! You didn't specifically ask for a review, but I just went through it anyway :)

kimgr:
@jru Great, thanks!

The Clang patch was reverted and reapplied in the meantime, and I tested the new solution without r350891 applied. It works in that case too, so it seems strictly superior.

I'll address both of our nits and merge. Thanks!

kimgr:
@jru Actually it turns out this was the last user of `PrintableSourceRange`, and using it comes out a little erratic:
```
[ #define     ] tests/cxx/badinc.cc:79:9: MACRO_CALLING_I1_MACRO
[ #ifdef      ] tests/cxx/badinc.cc:82:8: I1_MACRO_SYMBOL_WITHOUT_VALUE
[ #ifndef     ] tests/cxx/badinc.cc:87:9: I1_MACRO_SYMBOL_WITHOUT_VALUE
[ Use macro   ] tests/cxx/badinc.cc:92:5: I1_MACRO_SYMBOL_WITH_VALUE (from tests/cxx/badinc-i1.h:25:9)
[ Use macro   ] tests/cxx/badinc.cc:97:5: I1_MACRO_SYMBOL_WITH_VALUE0 (from tests/cxx/badinc-i1.h:26:9)
[ Use macro   ] tests/cxx/badinc.cc:100:7: I1_MACRO_SYMBOL_WITH_VALUE2 (from tests/cxx/badinc-i1.h:27:9)
[ #if defined ] tests/cxx/badinc.cc:105:5 - tests/cxx/badinc.cc:105:43: I1_MACRO_SYMBOL_WITHOUT_VALUE
[ #if defined ] tests/cxx/badinc.cc:111:7 - tests/cxx/badinc.cc:111:45: I1_MACRO_SYMBOL_WITHOUT_VALUE
```

I think I'll just have it print the start location to match the other preprocessor directives, and remove `PrintableSourceRange` too (yay!)

kimgr:
> I think I'll just have it print the start location to match the other preprocessor directives, and
> remove `PrintableSourceRange` too

And as further justification to this: before, the range was critically important, because it was the input to the custom lexer. Now the macro token is already parsed by Clang, so the full range is less relevant.

PR #640, opened 2019-01-16
Ignore compile_commands.json, .gdb_history

jbytheway:
Add a couple of convenient things to `.gitignore`.

* `.gdb_history`, the command history file for gdb (which gets put in $PWD when enabled).
* `compile_commands.json`, which needs to be in the source dir for some tools to find it automatically.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ __pycache__
 
 # Visual Studio Code
 .vscode
+
+.gdb_history
+compile_commands.json


jbytheway:
I'm pretty confident all those test failures are not the result of this branch.

kimgr:
Yeah, I think the recent Clang lexer changes that broke IWYU are now manifesting as segfaults. That's fixed on latest master. The gitignore changes look good, thanks!

PR #641, opened 2019-01-16
Better fix_includes handling of template lines

jbytheway:
When multiple forward declarations were present of the form:
```
template<typename T>
class Foo;
```
`fix_includes.py` would consider the multiple `template<typename T>` lines to be duplicates and delete all but one.

Fix that, and add a test for it.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1124,6 +1124,8 @@ def _DeleteDuplicateLines(file_lines, line_ranges):
   particular, it should not cover lines that may have C literal
   strings in them.
 
+  We only delete whole move_spans, not lines within them.
+
   Arguments:
     file_lines: an array of LineInfo objects.
     line_ranges: a list of [start_line, end_line) pairs.
@@ -1131,13 +1133,21 @@ def _DeleteDuplicateLines(file_lines, line_ranges):
   seen_lines = set()
   for line_range in line_ranges:
     for line_number in range(*line_range):
-      if file_lines[line_number].type in (_BLANK_LINE_RE, _COMMENT_LINE_RE):
+      line_info = file_lines[line_number]
+      if line_info.type in (_BLANK_LINE_RE, _COMMENT_LINE_RE):
         continue
-      uncommented_line = _COMMENT_RE.sub('', file_lines[line_number].line)
-      if uncommented_line in seen_lines:
-        file_lines[line_number].deleted = True
-      elif not file_lines[line_number].deleted:
-        seen_lines.add(uncommented_line)
+      if line_number != line_info.move_span[0]:
+        continue
+      span_line_numbers = range(line_info.move_span[0], line_info.move_span[1])
+      line_infos_in_span = [file_lines[i] for i in span_line_numbers]
+      uncommented_lines = [
+          _COMMENT_RE.sub('', inf.line.strip()) for inf in line_infos_in_span]
+      uncommented_span = ' '.join(uncommented_lines)
+      if uncommented_span in seen_lines:
+        for info in line_infos_in_span:
+          info.deleted = True
+      elif not line_info.deleted:
+        seen_lines.add(uncommented_span)
 
 
 def _DeleteExtraneousBlankLines(file_lines, line_range):
--- a/fix_includes_test.py
+++ b/fix_includes_test.py
@@ -2810,7 +2810,7 @@ The full include-list for comments_with_includes:
     self.RegisterFileContents({'comments_with_includes': infile})
     self.ProcessAndTest(iwyu_output)
 
-  def testRemoveDuplicates(self):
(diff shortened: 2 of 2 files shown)


fix_includes_test.py
```diff
@@ -2841,6 +2841,31 @@ def testRemoveDuplicates(self):
...
+#include <notused.h>  ///-
+template<typename T>
+class A;
+template<typename T>
+class B;
+
+f(A&, B&);
+"""
+    iwyu_output = """\
+remove_duplicates should add these lines:
```
kimgr: Maybe name this something unique to avoid confusing it with the test before (the diff will become more obvious, too).

Also, I wonder if there's a test missing to make sure we *do* treat two identical multi-line template fwd decls as dupes?

jbytheway: Good idea.  Done both.

(CI still struggling, it looks like).

fix_includes.py
```diff
@@ -1124,20 +1124,30 @@ def _DeleteDuplicateLines(file_lines, line_ranges):
...
         continue
-      uncommented_line = _COMMENT_RE.sub('', file_lines[line_number].line)
-      if uncommented_line in seen_lines:
-        file_lines[line_number].deleted = True
-      elif not file_lines[line_number].deleted:
-        seen_lines.add(uncommented_line)
+      span_line_numbers = range(line_info.move_span[0], line_info.move_span[1])
+      line_infos_in_span = [file_lines[i] for i in span_line_numbers]
+      uncommented_lines = [
+              _COMMENT_RE.sub('', info.line) for info in line_infos_in_span]
```
kimgr: Indentation looks off here, I think it should be four spaces in from `uncommented` on the preceding line. Or two, I can never tell with Python :). Try `pep8` if you have it/can install it, and see which it prefers (there are existing issues, but let's address those separately).

fix_includes_test.py
```diff
@@ -2830,19 +2830,73 @@ def testRemoveDuplicates(self):
...
+  def testDontRemoveTemplateLines(self):
+    """Tests we don't accidentally think repeated template lines are dupes."""
+    infile = """\
+#include <notused.h>  ///-
+template<typename T>
+class A;
+template<typename T>
+class B;
+
+f(A&, B&);
```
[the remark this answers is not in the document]
jbytheway: Done.

fix_includes.py
jbytheway: `pep8` was no help because it just complains everywhere that you're not using 4-space indents.  But I found elsewhere in the file 4 spaces were being used similarly, so I made it 4.

fix_includes.py
```diff
@@ -1124,20 +1124,30 @@ def _DeleteDuplicateLines(file_lines, line_ranges):
...
+      uncommented_span = '\n'.join(uncommented_lines)
```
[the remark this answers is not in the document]
jbytheway: It doesn't matter what the correct line ending for the file is; this joined string is only being used for comparison with other strings.  It just has to be consistent within this function.  It could be `'***'.join(...)` and it would work the same.

We don't even have to join the strings at all; could just store it as a tuple of lines.

Or, we could join with spaces, which would dedupe more cases, like this one:
```
template<typename T> class A;
template<typename T>
class A;
```
Any preference?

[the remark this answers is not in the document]
kimgr: Ah, I see. I never looked at the function as a whole, I thought `seen_lines` was returned. Cool! Yeah, let's use a space, then.

fix_includes.py
kimgr: :+1:

fix_includes.py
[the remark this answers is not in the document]
jbytheway: Note that this logic may conflict with the changes in #643.  If I understand that correctly, when it is merged, this code will also need to strip line endings; should be a simple fix.

fix_includes_test.py
```diff
@@ -2830,19 +2830,75 @@ def testRemoveDuplicates(self):
...
+    iwyu_output = """\
+remove_duplicate_forward_declarations should add these lines:
+
+remove_duplicate_forward_declarations should remove these lines:
+- #include <notused.h>  // lines 1-1
+
+The full include-list for remove_duplicate_forward_declarations:
+class A;  // lines 2-2
+template <typename T> class B;  // lines 3-4
+class A;  // lines 5-5
```
kimgr: Why are these duplicates not removed now?

jbytheway: They are removed.  That's what the `///-` markings signify in the lines above (in `infile`).

What you're quoting here is IWYU output.  IWYU doesn't do deduping, so it's expected to see the dupes here.

kimgr:
Thanks for driving this to completion!

PR #643, opened 2019-01-23
Reduce spurious line seperator changes.

trixirt:
A diff of the original and change file should show only changes related to the includes and/or foreward declarations.  In files that have line separators that do not match the FileInfo.linesep, those line will be
changed.  So instead of stripping a file's line separators, reuse then. For any new line, use the FileInfo.linesep.

Also remove adding a line separator at the end of a file.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -581,7 +581,12 @@ def _ReadFile(filename, fileinfo):
   try:
     with open(filename, 'rb') as f:
       content = f.read()
-      return content.decode(fileinfo.encoding).splitlines()
+      # Call splitlines with True to keep the original line
+      # endings.  Later in WriteFile, they will be used as-is.
+      # This will reduce spurious changes to the original files.
+      # The lines we add will have the linesep determined by
+      # FileInfo.
+      return content.decode(fileinfo.encoding).splitlines(True)
   except (IOError, OSError) as why:
     print("Skipping '%s': %s" % (filename, why))
   return None
@@ -591,7 +596,8 @@ def _WriteFile(filename, fileinfo, file_lines):
   """Write the given file-lines to the file."""
   try:
     with open(filename, 'wb') as f:
-      content = fileinfo.linesep.join(file_lines) + fileinfo.linesep
+      # file_lines already have line endings, so join with ''.
+      content = ''.join(file_lines)
       content = content.encode(fileinfo.encoding)
       f.write(content)
   except (IOError, OSError) as why:
@@ -2038,7 +2044,7 @@ def _GetSymbolNameFromForwardDeclareLine(line):
   return symbol_name
 
 
-def FixFileLines(iwyu_record, file_lines, flags):
+def FixFileLines(iwyu_record, file_lines, flags, fileinfo):
   """Applies one block of lines from the iwyu output script.
 
   Called once we have read all the lines from the iwyu output script
@@ -2057,6 +2063,7 @@ def FixFileLines(iwyu_record, file_lines, flags):
     flags: commandline flags, as parsed by optparse.  We use
        flags.safe_headers to turn off deleting lines, and use the
        other flags indirectly (via calls to other routines).
+    fileinfo: FileInfo for the current file.
 
   Returns:
     An array of 'fixed' source code lines, after modifications as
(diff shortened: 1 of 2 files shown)


fix_includes.py
```diff
@@ -2146,20 +2153,26 @@ def FixFileLines(iwyu_record, file_lines, flags):
       #    'namespace foo { class Bar; }\n' -> 'namespace foo {\nclass Bar;\n}'
       # along with collecting multiple classes in the same namespace.
       new_lines = _NormalizeNamespaceForwardDeclareLines(new_lines)
+
+    # Add line seperators to the new_lines.
```
kimgr: `new_lines` -> `new lines`. I think you can leave out the `Strip the ends...` part of this comment, that detail is obvious from the code.

I would prefer this whole thing as a list comprehension:

    new_lines = [nl.rstrip() + fileinfo.linesep for nl in new_lines]

fix_includes.py
```diff
@@ -2057,6 +2063,7 @@ def FixFileLines(iwyu_record, file_lines, flags):
     flags: commandline flags, as parsed by optparse.  We use
        flags.safe_headers to turn off deleting lines, and use the
        other flags indirectly (via calls to other routines).
+    fileinfo: FileInfo class for the current file.
```
kimgr: Remove `class`

fix_includes_test.py
```diff
@@ -107,10 +107,12 @@ def RegisterFileContents(self, file_contents_map):
     for (filename, contents) in file_contents_map.items():
       before_contents = []
       expected_after_contents = []
-      for line in contents.splitlines():
+      for line in contents.splitlines(True):
         m = remove_re.search(line)
         if m:
-          before_contents.append(line[:m.start()])
+          # The trailing line seperator is stripped
```
kimgr: Typo: seperator -> separator.

Collapse this to a single sentence, including closing period:

    # The trailing line separator is stripped, so append a '\n'.

kimgr:
Commit message:
> Reduce spurious line separator changes.
>
> A diff of the original and change file should show only changes related
> to the includes and/or foreward declarations.  In files that have line

foreward -> forward

> separators that do not match the FileInfo.linesep, those line will be
> changed.  So instead of stripping a file's line separators, reuse then.

then -> them

> For any new line, use the FileInfo.linesep.
>
> Also remove adding a line separator at the end of a file.

I was concerned about the last sentence here, but I see now that we only avoid adding a linesep if the last line is otherwise unchanged, right? You may want to clarify that.

kimgr:
This is very nice, thanks!

Would it be possible to add a testcase with mixed line endings and see that they are maintained?

fix_includes.py
trixirt: Fixed in latest

fix_includes.py
trixirt: Fixed in latest

fix_includes_test.py
trixirt: Fixed in latest

trixirt:
Tried to add a test but failed, the string version of input make reproducing actual files difficult.
The failure happens with binutils/gas/ecoff.c

```
diff -up /a/pub-iwyu/src/binutils-fsf-git/gas/ecoff.c /tmp/24eb-030b-a804-7329.c 
--- /a/pub-iwyu/src/binutils-fsf-git/gas/ecoff.c	2019-01-22 11:49:30.129024189 -0800
+++ /tmp/24eb-030b-a804-7329.c	2019-01-28 15:07:04.057825171 -0800
@@ -22,8 +22,6 @@
    Software Foundation, 51 Franklin Street - Fifth Floor, Boston, MA
    02110-1301, USA.  */
 
-#include "as.h"
-
 /* This file is compiled conditionally for those targets which use
    ECOFF debugging information (e.g., MIPS ELF, Alpha ECOFF).  */
 
@@ -617,7 +615,8 @@
     #25             36  0x00000024  unsigned long
     #26             48  0x00000030  struct no name { ifd = -1, index = 1048575 }
 */
-
+
+
```

<shameless-plug>
binutils is one of the projects i am running through iwyu with my s2s project to test the effectiveness of iwyu.
https://github.com/trixirt/s2s/blob/master/README.txt
</shameless-plug>

kimgr:
This looks good to me, thank you! I'll merge this and look into a reduced testcase once this is on master.

> binutils is one of the projects i am running through iwyu with my s2s project to test the effectiveness of iwyu.
> https://github.com/trixirt/s2s/blob/master/README.txt

Cool project!

PR #644, opened 2019-01-27
Make --no_fwd_decls respect local forward decls

kimgr:
Before this, the --no_fwd_decls mode would force-remove existing forward
decls:

    class Foo;
    void Bar(const Foo*);

Add an additional rule that we avoid promoting uses of
forward-declarations in the same file to full uses. This retains forward
decls like the above.

@vmartinv As far as I can see, this fixes your testcase in #603. Let me know how it works out in practice.


iwyu_output.cc
```diff
@@ -1056,8 +1056,11 @@ void ProcessForwardDeclare(OneUse* use,
     }
   }
 
-  // (A7) If --no_fwd_decls has been passed, recategorize as a full use.
-  if (GlobalFlags().no_fwd_decls) {
+  // (A7) If --no_fwd_decls has been passed, recategorize as a full use unless
+  // the decl is in this file (in which case it must be a self-sufficient decl
+  // being used, so we can just let IWYU do its work).
+  if (GlobalFlags().no_fwd_decls &&
+      GetFileEntry(use->decl_loc()) != GetFileEntry(use->use_loc())) {
```
jru: This is not the right solution, in my opinion. What if the forward decl is in a different file?
I think we should instead look for the definition of the associated decl, and otherwise leave it as a fwd-decl use. 

Haven't tested it, but something like this should do it:
```cpp
if (GlobalFlags().no_fwd_decls &&
    GetDefinitionForClass(use->decl()) != nullptr) {
...
```

kimgr: We're trying to keep types that *don't* have a definition as forward-declares, but I'm guessing maybe you meant `== nullptr`?

It's an interesting take; you're proposing that if we've said "no fwd decls", we force all uses to full except the ones for which there is no definition available? That sounds good to me, but I think there's a risk that IWYU will suggest to add redundant forward decls to the current file for name-only decls that have leaked into the current translation unit. But it still beats the current behavior. I'll see if I can implement that.

jru: I really meant that: The definition for the decl must exist for the fwd-decl use to be convertible to a full use. 
Otherwise, what sense would it make to forcefully convert it to a full use, if the complete type doesn't exist?

Or maybe I am missing something (I havent tried the switch myself)?

kimgr: I applied your suggestion and it passes the current test. I'm not sure what happens with fwd-decls in other files, though. Something like this might go awry:
```
// foohash.h
class Foo;

uintptr_t hash(const Foo* p) {
    return (uintptr_t)p;
}

// foo.h
class Foo {
   int lo;
   int hi;
};

// main.cc
#include "foohash.h"
#include "foo.h"

bool equal(const Foo* a, const Foo* b) {
    return hash(a) == hash(b);
}
```

* main.cc only uses `Foo` in name, and only needs a fwd-decl
* `--no_fwd_decls` stipulates that we don't want fwd-decls, and because `Foo`'s complete type is visible, we promote the uses in `equal` to full uses
* Now we keep foo.h because `Foo` is used, and foohash.h because `hash` is used.

This feels icky, but I think it makes sense. We asked for no fwd-decls. I'll have to test what happens if we remove the `hash` function and let foohash.h carry only the fwd-decl.

kimgr: The same-file check I suggested works better, for some reason. There is other logic earlier in `ProcessForwardDeclare` to ignore the use if the definition is present, I think that could be what helps.

There is logic similar to this in `CalculateIwyuForForwardDeclareUse`, but more sophisticated, but this basic check seems to cover all cases I've found.

I'll push a new version with one more example in the test.

kimgr: @jru Ping! I just rebased this.

I don't remember exactly now why your suggested solution didn't work out, but I couldn't get it to pass the testcase.

jru: Sorry, I forgot about this, sorry. I'll take another look!

kimgr: :+1:

jru:
After taking another look and trying out different cases, it really seems your solution is better. 
I thought it would break when the forward-decl is in a different file, but it doesn't. LGTM!

jru:
*Update* - I think I found some undesirable behaviour:

If I take your test and remove the use of the *hash()* function all we need now is the forward decl of the Foo, so in the regular mode, IWYU should suggest to remove it and add instead a `class Foo;`  

```cpp
#include "foohash.h"
bool equal(const Foo* a, const Foo* b) {
    //return hash(a) == hash(b);
    return true;
}
```
And that's exactly what happens:

```
example/main-foohash.cc should add these lines:
class Foo;

example/main-foohash.cc should remove these lines:
- #include "foohash.h"  // lines 5-5

The full include-list for example/main-foohash.cc:
class Foo;
```
But, if the `--no_fwd_decls` it will instead cling to the fwd-decl of Foo in `foohash.h` and instead suggest we keep the include (because we switch it to a full-use):
```
(example/main-foohash.cc has correct #includes/fwd-decls)
```

Not sure if that is acceptable, but it does seem to go against the IWYU principle.

jru:
I keep going in circles about this :) 

Just checked, and the solution I proposed handles well the special case I described above. 

The `no_fwd_decls.cc` test changes outcome slightly, but which one is the correct one I guess is a matter of perspective: If no fwd-decls are wanted, then expecting the full definition to be present seems like the right thing to do. For types where the definition is not present, IWYU would not honor the `--no_fwd_decls` switch, if we go for my solution.


Could you explain in more detail what made the same-file check work better with your proposed solution? Per your comment:

> The same-file check I suggested works better, for some reason. There is other logic earlier in ProcessForwardDeclare to ignore the use if the definition is present, I think that could be what helps.

kimgr:
Interesting, thanks for digging in!

> Could you explain in more detail what made the same-file check work better with
> your proposed solution? Per your comment:
>
> > The same-file check I suggested works better, for some reason. There is other logic earlier in
> > ProcessForwardDeclare to ignore the use if the definition is present, I think that could be what
> > helps.

I'm not sure what I meant by 'helps' above, but as I recall it a forward-declare-use is set as ignored if the definition is present [1], so I can imagine that trips something up.

I think I may have confused two qualities with regard to `--no_fwd_decls`:

1. IWYU must not do something that breaks compilation of the code
2. IWYU should not replace an `#include` by a forward-decl under `--no_fwd_decls`

The original defect was of category (1), but comparing our two solution alternatives, I think I've forgotten about that, and gotten stuck on (2). I'm going to have to revisit this with both in mind. I'm not even sure (2) needs to be so rigid -- if there's no complete definition available, as you say, we don't need to bend over backwards for (2). (1) on the other hand, is non-negotiable.

I'll take another look, thanks for elaborating!

[1] https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_output.cc#L1090

kimgr:
After experimenting with this, I also prefer the `GetDefinitionForClass` solution.

But I can't get it to suppress the forward-declare in a variation on the `foohash` example above (I tried to fold that into the testcase).

It now requires the include of `foohash.h`, but it also puts in a forward-declare for `Foo`, which seems pointless. I'll see if I can convince it to discard that if a forward-declare for `Foo` is already present through a desired header.

jru:
> I'm not sure what I meant by 'helps' above, but as I recall it a forward-declare-use is set as ignored if the definition is present [1], so I can imagine that trips something up.

The check (A6) is for definitions found earlier *in the same file*, which should still apply regardless. Shouldn't it?

But that brings a good point. I am not 100% sure now if using `GetDefinitionForClass` would be too restrictive. For example, C++11 enums are foward-declarable (even though we don't seem to support them)

> It now requires the include of foohash.h, but it also puts in a forward-declare for Foo, which seems pointless. I'll see if I can convince it to discard that if a forward-declare for Foo is already present through a desired header.

The fwd-decl of Foo is suggested in the regular mode, too. Isn't it what IWYU is supposed to do?

kimgr:
> The fwd-decl of Foo is suggested in the regular mode, too. Isn't it what IWYU is supposed to do?

I have a slightly different example;

```
// no_fwd_decls-nameonly.h

class NameOnly;

const void* AddressOf(const NameOnly& a) {
  return reinterpret_cast<const void*>(&a);
}

// no_fwd_decls.cc
#include "no_fwd_decls-nameonly.h"

bool AreSame(const NameOnly& a, const NameOnly& b) {
  return AddressOf(a) == AddressOf(b);
}
```

With `--no_fwd_decls`, I would guess that `NameOnly` should be covered by the include of `no_fwd_decls-nameonly.h`. But IWYU insists that `NameOnly` should be forward declared in `no_fwd_decls.cc`, even if the include is still necessary.

kimgr:
Here's a halfway hack that passes the new test.

Policy:
* If a forward-declare is enough, but a definition of the symbol is available from a header, prefer to include the header (`IndirectClass` in the test)
* If a forward-declare is in the same file as its use with no other redeclarations, use it (`LocalFwd` in the test)
* If a forward-declare is the only reason we're including a header, replace the header and forward-declare locally (`Fwd` in the test)
* If a forward-declare comes from a header, but we're also using something else from that header, attribute the forward-declare to the header (`NameOnly` in the test)

I think the implementation of this policy needs some work, but do you think the test scenarios are valid as described?

jru:
> With --no_fwd_decls, I would guess that NameOnly should be covered by the include of no_fwd_decls-nameonly.h. But IWYU insists that NameOnly should be forward declared in no_fwd_decls.cc, even if the include is still necessary for the full use of AddressOf.

Well, maybe it should in this case, but that's the result of how IWYU has been designed to work. In your `NameOnly` example, there are two fwd-declare uses of the type in the `AreSame()` prototype and so IWYU makes sure you either include the definition directly (if full use or in `--no_fwd_decls` mode) or that you forward declare it directly.

> I think the implementation of this policy needs some work, but do you think the test scenarios are valid as described?

I didn't have time yet to understand the policy you described in detail, but it seems to me that these are two orthogonal problems and should be tackled separately. Some of those points might end up making sense for full-uses too, so we could benefit from analysing the problem more holistically.

Isn't this related to the idea of re-exporting symbols? Meaning that under some conditions we could assume certain includes to be the source for symbols that we use directly, so that we don't have to include all of them.

kimgr:
> Well, maybe it should in this case, but that's the result of how IWYU has been designed to work.
> In your NameOnly example, there are two fwd-declare uses of the type in the AreSame() prototype 
> and so IWYU makes sure you either include the definition directly
> (if full use or in --no_fwd_decls mode) or that you forward declare it directly.

Yeah, `--no_fwd_decls` is an entirely different policy -- in normal IWYU mode we want to duplicate fwd-decls to decouple source files from each other. In `--no_fwd_decls` mode we want to reduce the risk that fwd-decls and actual decls mismatch/drift apart, so we push to prefer existing declarations over introducing new ones (i.e. the more redecls there are, the higher the risk is that they will eventually all need to be changed in concert). That line of reasoning could be extended to the `Fwd` example in the test too, and maybe it should.

> Isn't this related to the idea of re-exporting symbols? Meaning that under some conditions we
> could assume certain includes to be the source for symbols that we use directly, so that we
> don't have to include all of them.

I guess it's similar to re-export, but I don't see clearly what it would mean to unify the two ideas. It's not clear to me that solving that problem has higher priority than fixing this bug with `--no_fwd_decls`, or that the order in which we do those things make it particularly easier or harder.

jru:
> In --no_fwd_decls mode we want to reduce the risk that fwd-decls and actual decls mismatch/drift apart, so we push to prefer existing declarations over introducing new ones

Right, that makes sense. And I think the switch in the current form does it already quite good. It will suggest to remove the fwd-decl if a full use is availalbe.

Thinking back at what we are trying to solve here: We don't want to break the compile when the definition is not available, even in this mode. Other than that, there is little that can be done given that, in my opinion, an available definition should be a pre-requesite for this feature. 

Your policy looks reasonable, but actually all first three we already get for free. The last one, in my mind, is trying too hard at fixing an inconvenience for which this mode is not designed to work. 

If the users of this mode don't want fwd-decls poping up, they really need to make sure their definitions are reachable. Whenever they see one, though, they would immediately know that IWYU couldn't do without it and fix it themselves.

But, of course, I cannot speak for all of them :)

kimgr:
I did some more work on the test to have it reflect my current preference, which is: if there's a choice between a local forward-declare and an include, always choose the include.

And the only way I happened to make it pass was by reverting to my original solution. So I stand by it now; I think this is the best way to implement the perfect `--no_fwd_decls` :)

jru:
> if there's a choice between a local forward-declare and an include, always choose the include.

Sure. If that's your preference, then let's go for it. Both have trade-offs, it is more about which one you can live with :)

kimgr:
I polished comments and commit messages and added a little section in the docs to contrast the two policies. @jru I'm going to assume you're still fine with it; we can always adjust post-merge if not.

jru:
Perfect explanation. Thanks!

PR #647, opened 2019-02-04
Use forked llvm-toolchain-trusty-7 in TravisCI

eyal0:


kimgr:
Thanks! But this is the wrong general direction. IWYU master tracks llvm master, not the 7 release. So we would like it to install and use the snapshot builds of llvm/clang. 

And it looks like the -7 toolchain is no longer working in Travis either.

See the IWYU bug tracking this for more details. If you could figure out what to do, I'd be most grateful.

eyal0:
It didn't help anyway.  I hope that you find a solution soon.

PR #649, opened 2019-02-12
Specify version to install LLVM/Clang snapshots

jru:
Testing to see if IWYU would build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ addons:
      # TODO: These should really be the snapshots packages, but they are currently
      # broken. Remove the '-9' once they get fixed (see issue #642 for more info)
      - llvm-9-dev
+     - llvm-9-tools
      - libclang-9-dev
      - clang-9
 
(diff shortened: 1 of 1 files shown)


jru:
Well, the latest packages are installed succesfully, but it fails now with...
```
CMake Error at /usr/lib/llvm-9/lib/cmake/llvm/LLVMExports.cmake:1352 (message):
  The imported target "FileCheck" references the file
     "/usr/lib/llvm-9/bin/FileCheck"
```
Not sure what to make of this. I'll look into it.

jru:
So, after explicitly mentioning 9 in the package names and adding llvm-9-tools for FileCheck it builds again with the latest binaries (2 days old)

@kimgr Do you think we can get this in for the time being to have CI back? It is a bummer having to put the version in there, but otherwise I am not sure when/if will ever get fixed.

kimgr:
@jru Thanks for taking this on! Yes, definitely let's merge this to get back to sanity.

Could you just add a little TODO comment that the `-9` packages should really be snapshot? Feel free to merge with that fixed.

PR #652, opened 2019-03-02
Support headers using symbols mapped to themselves

jbytheway:
Currently if IWYU decides that the best header for a symbol is the same one that uses the symbol, it can cause that header to add a `#include` of itself.

This in particular can happen with mappings.

Detect this case and cause the use to be ignored rather than requiring a `#include`.

Furthermore, when choosing from multiple public mapping headers, prefer to use oneself where possible (since that minimizes `#include`s).

@kimgr This is the [alternative solution I mentioned](https://github.com/include-what-you-use/include-what-you-use/pull/637#discussion_r251594753) in #637.  Essentially, I realised that, rather than avoiding assignments of uses to the header containing them, we should be embracing them.  And mappings could already lead to similar problems as the one I encountered on that PR, even independently of the changes in that PR.  Here, I demonstrate such a problem in the test case, and solve it.

After this change, #637 becomes simpler (in particular, the `HasMappingSuitableForFile` function becomes just `HasMapping`).


kimgr:
👍 Cool, I'll take a look tomorrow!

iwyu_output.cc
```diff
@@ -803,18 +804,29 @@ set<string> CalculateMinimalIncludes(
...
-  // more than one public #include.  First choice: an include in
-  // associated_direct_includes (those are includes that are not going
-  // away, since we can't change associated files).  Second choice,
-  // includes in direct_includes that are also already in
-  // desired_headers.  Third choice, includes in desired_headers.
-  // Fourth choice, includes in direct_includes.  Picking in
-  // this order minimizes the number of #includes we add, while
-  // allowing us to remove #includes if need be.
+  // more than one public #include.  Use the following priority order:
+  // - Ourselves.
```
kimgr: I had trouble understanding why we prioritize the current header, until I read the rationale in the commit message/PR description. I think it would be nice to expand on the minimize-includes benefit here (if possible without too much verbiage).

iwyu_output.cc
```diff
@@ -1472,7 +1485,18 @@ void CalculateIwyuForFullUse(OneUse* use,
     return;
   }
 
-  // (E2) Mark iwyu violation unless in a current #include.
+  // (E2) Ignore use when the suggested header *is* the current file (this can
```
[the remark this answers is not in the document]
jbytheway: Done.

iwyu_output.cc
jbytheway: I feel that this is already covered by the sentence at the end of this comment block: "Picking in this order minimizes the number of #includes we add, while allowing us to remove #includes if need be."

Do you think more is needed?  Or perhaps it would be clearer if this were stated before the priority list?

kimgr: Yeah, reading it again, I think you're right. Ship it!

Issue #653, opened 2019-03-06
command line option to match 'IWYU pragma: keep'


trixirt:
In cases where it is impossible to edit a header ex/ stdio.h,  it would be useful to support the pragma keep as a command line option.  

So add the command line option
 --keep

Behaves like --also_needs in that it support glomming/heuristic file matching.
Will additional check likely in the pp handler InclusionDirective 
Reuses the existing kept_includes_ data handling.

i am starting work on this.

kimgr:
Fixed by PR #654

PR #654, opened 2019-03-07
Add command line option --keep=<glob>

trixirt:
--keep is the command line equivalent to the keep pragma with globbing.

Here is the issue link
https://github.com/include-what-you-use/include-what-you-use/issues/653


kimgr:
We've been reluctant to add command-line options in the past, but I'm trying to become better at embracing change :). This is nice and isolated and fills a need where external mappings are lacking.

So let's see if we can make it work. I'll add some comments inline, and this needs a test.

iwyu_globals.cc
```diff
@@ -405,6 +410,21 @@ bool ShouldReportIWYUViolationsFor(const clang::FileEntry* file) {
   return false;
 }
 
+void AddGlobToKeepIncludes(const string& glob) {
+  CHECK_(commandline_flags && "Call ParseIwyuCommandlineFlags() before this");
+  commandline_flags->keep.insert(NormalizeFilePath(glob));
+}
+
+bool ShouldKeepIncludeFor(const clang::FileEntry* file) {
+  if (GlobalFlags().keep.size()) {
```
kimgr: I'd prefer to early exit here, and use `empty()` instead of `size()`:

    if (GlobalFlags().keep.empty())
      return false;

iwyu_globals.h
```diff
@@ -89,6 +89,7 @@ struct CommandlineFlags {
 
   set<string> check_also;  // -c: globs to report iwyu violations for
   string howtodebug;       // -d: file to print gdb-invoking instruction for
+  set<string> keep;        // These globs keep includes
```
kimgr: Rephrase comment: `// -k: globs to force-keep includes for`

iwyu_globals.h
```diff
@@ -131,6 +132,11 @@ FullUseCache* ClassMembersFullUseCache();
 void AddGlobToReportIWYUViolationsFor(const string& glob);
 bool ShouldReportIWYUViolationsFor(const clang::FileEntry* file);
 
+// For the commanline option --keep.
```
kimgr: Typo: commanline -> commandline

iwyu_preprocessor.cc
```diff
@@ -400,6 +400,11 @@ void IwyuPreprocessorInfo::MaybeProtectInclude(
     protect_reason = "pragma_keep";
     FileInfoFor(includer)->ReportKnownDesiredFile(includee);
 
+  } else if (ShouldKeepIncludeFor(includee)) {
+    // The command line version of pragma keep.
+    protect_reason = "pragma_keep";
```
kimgr: This reason is used for logging, so maybe better not to attribute this to "pragma: keep". Let's say `--keep` instead.

trixirt:
I believe all of the current issues have been addressed and a test added.

run_iwyu_tests.py
```diff
@@ -69,7 +69,8 @@ def setUp(self):
                                     self.CheckAlsoExtension('-fnreturn.h'),
                                     self.CheckAlsoExtension('-typedefs.h'),
                                     self.CheckAlsoExtension('-d2.h')],
-      'keep_mapping.cc': [self.CheckAlsoExtension('-public.h'), 
+      'keep_includes.c' : [ '--keep=./tests/c/keep_b.h' ],
```
kimgr: I think you can get fancy here and use a glob -- how about `'--keep="*keep_b.h"'`?

kimgr: Also, in line with convention, the header should be called `keep_includes-d1.h` -- that makes it clear that it belongs to the `keep_includes` test, and the `d` indicates it's a direct include.

kimgr: Oh, and no space before colon, and no spaces inside list brackets.

tests/c/keep_includes.c
```diff
@@ -0,0 +1,18 @@
+//===--- keep_includes.c - test input file for iwyu -----------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+#include "keep_b.h"
+void hi() {
```
kimgr: This has no effect, so you might as well remove it.

tests/c/keep_includes.c
```diff
@@ -0,0 +1,18 @@
+//===--- keep_includes.c - test input file for iwyu -----------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+#include "keep_b.h"
```
kimgr: Can you add an explanatory comment before here that explains the feature being tested? It looks entirely pointless in isolation :)

Something like:
```
...

// Test that the --keep command-line argument forces us to keep an unused
// (and empty) header.

#include "keep_b.h"
...
```

trixirt:
Nearly all current issues addressed.
--keep=*keep_includes-d1.h and other similar did not work.
Test uses a glob that does.

tests/c/keep_includes-d1.h
```diff
@@ -0,0 +1,11 @@
+//===--- keep_includes-d1.h - test header for iwyu ----------------------------------===//
```
kimgr: Line runs too long now that the filename is longer, trim dashes to hit 80, please

tests/c/keep_includes.c
```diff
@@ -0,0 +1,23 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// keep_includes-d1.h is an empty file.
+// Normally it would be removed.
+// In this test we are exercising the --keep command line option.
+// IWYU should not remove keep_include-d1.h when --keep=*keep_includes-d1.h
```
kimgr: The glob expression here doesn't match the one actually used.

tests/c/keep_includes.c
```diff
@@ -0,0 +1,23 @@
...
+// is used.
+
+#include "keep_includes-d1.h"
+
+/**** IWYU_SUMMARY
+
+(tests/c/keep_includes.c has correct #includes/fwd-decls)
+
+***** IWYU_SUMMARY */
+
```
kimgr: No need for this blank line

run_iwyu_tests.py
```diff
@@ -69,7 +69,8 @@ def setUp(self):
                                     self.CheckAlsoExtension('-fnreturn.h'),
                                     self.CheckAlsoExtension('-typedefs.h'),
                                     self.CheckAlsoExtension('-d2.h')],
-      'keep_mapping.cc': [self.CheckAlsoExtension('-public.h'), 
+      'keep_includes.c': ['--keep=./*/*/keep_includes-d1.h'],
```
kimgr: Hm, I wonder if the Windows implementation of `GlobMatchesPath` can take this pattern...

Sorry to go back and forth on this, but maybe it's safer to use something `--keep=keep_includes-*.h`.

tests/c/keep_includes.c
```diff
@@ -0,0 +1,23 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+// keep_includes-d1.h is an empty file.
+// Normally it would be removed.
+// In this test we are exercising the --keep command line option.
+// IWYU should not remove keep_include-d1.h when --keep=*keep_includes-d1.h
+// is used.
+
+#include "keep_includes-d1.h"
```
kimgr: Most tests use a path from the header search path, i.e. `tests/c/keep_includes-d1.h` -- does that work here, or does it confuse the globbing?

trixirt:
Changes made.
glob-ing is not work as expected.
I am raising this issue to track
https://github.com/include-what-you-use/include-what-you-use/issues/660
And starting to work on it.

kimgr:
I don't think you should block this feature on fixing the globbing. Doing so would be nice, but we can worry about that separately. `--keep` is almost ready to go, and is useful even if its globbing is just as bad as for `--check_also`.

trixirt:
ok, stopping glob issue for now.
I have double checked the spacing on the test files, they appear to be ok.
Are there any other issues I need to fix?

kimgr:
Yes, see remaining open review comments:

* Stupify the globs in tests to be more likely to work, e.g. `--keep=keep_includes-*.h`
* Use `#include "tests/c/keep_includes-d1.h"` to exercise the `-I.` include path

I'm worried that the combination of those two will also break the glob, but maybe that highlights a way to do it better, e.g. only match on basename, or something?

trixirt:
I tried the suggestions, they both fail.
Then fiddled with the test some to be more globby.
I believe problems come down to, I am expecting
*/foo == <full-path>/foo or
*/foo == <non-trivial-relative-path>/foo
and both do not work.

kimgr:
Aargh. How about;

* `--keep=keep_includes*.h`
* No additional `.` include path
* Spell the test include `#include "keep_includes-d1.h"`

That must work, right?

I think for this to be truly useful, the globbing needs to be improved, but in the current state `--keep` can at least be used even if it's a little inconvenient.

Sorry, it looks like my recent getopt merge conflicts with your changes, but it should be easy to sort out, I hope.

trixirt:
There were no problems wrt your changes.

The arg for keep=<path> appears to be dependent on the cwd.

`> include-what-you-use tests/c/keep_includes.c -Xiwyu --keep=*/*/keep_includes*

(tests/c/keep_includes.c has correct #includes/fwd-decls)
> include-what-you-use tests/c/keep_includes.c -Xiwyu --keep=keep_includes*

tests/c/keep_includes.c should add these lines:

tests/c/keep_includes.c should remove these lines:
- #include "keep_includes-d1.h"  // lines 16-16

The full include-list for tests/c/keep_includes.c:
---
 > cd tests/c

include-what-you-use keep_includes.c -Xiwyu --keep=keep_includes*

(keep_includes.c has correct #includes/fwd-decls)
`

kimgr:
Hm, annoying. You might be able to work around that by using `#include "tests/c/keep_includes-d1.h"` -- that way it should always find it on the `-I` search path, and could be more consistent.

I was struggling yesterday with my local  IWYU crashing after applying your patch, but I think there was some problem in my build environment, so I'Il give it another shot tonight.

kimgr:
I experimented with this on your branch, and this appears to work:
```diff
diff --git a/run_iwyu_tests.py b/run_iwyu_tests.py
index aed2ffc..5cdba9c 100755
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -70,7 +70,7 @@ class OneIwyuTest(unittest.TestCase):
                                     self.CheckAlsoExtension('-fnreturn.h'),
                                     self.CheckAlsoExtension('-typedefs.h'),
                                     self.CheckAlsoExtension('-d2.h')],
-      'keep_includes.c': ['--keep=*/*/keep_includes*.h'],
+      'keep_includes.c': ['--keep=tests/c/keep_includes*.h'],
       'keep_mapping.cc': [self.CheckAlsoExtension('-public.h'),
                           self.MappingFile('keep_mapping.imp')],
       'macro_location.cc': [self.CheckAlsoExtension('-d2.h')],
diff --git a/tests/c/keep_includes.c b/tests/c/keep_includes.c
index b3c10b0..45bed9e 100644
--- a/tests/c/keep_includes.c
+++ b/tests/c/keep_includes.c
@@ -11,9 +11,9 @@
 // Normally it would be removed.
 // In this test we are exercising the --keep command line option.
 // IWYU should not remove keep_include-d1.h when
-// --keep=*/*/keep_includes*.h is used.
+// --keep=tests/c/keep_includes*.h is used.
 
-#include "keep_includes-d1.h"
+#include "tests/c/keep_includes-d1.h"
 
 /**** IWYU_SUMMARY
 ```

I think that's more obvious to understand, and fits better with convention. The glob behavior is not the most important part of this test -- the `--keep` switch is.

kimgr:
Also, I'm not sure why we have that `NormalizeFilePath` on the glob arguments, but let's save that for #660.

trixirt:
Latest update uses suggested change to the test.

kimgr:
Thank you!

PR #655, opened 2019-03-08
Refactor: Rename `built_ins*` to `builtins*`

kimgr:
I've always found the `built_ins_` spelling to be hard to read, so here's a set of commits to harmonize the spelling of 'builtin' without underscores.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -919,7 +919,7 @@ bool IsDefaultNewOrDelete(const FunctionDecl* decl,
                           const string& decl_loc_as_quoted_include) {
   // Clang will report <new> as the location of the default new and
   // delete operators if <new> is included. Otherwise, it reports the
-  // (fake) file "<built_in>".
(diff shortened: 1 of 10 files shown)


kimgr:
@jbytheway You may want to take a look, since I stepped all over your tests :)

kimgr:
> Also, I understand you didn't want to change the current 'built-in[s]' occurences in the code?
> There are a few in comments and strings.

I didn't think to look for them. I think "built-in" is fine in prose, I just thought "built_in" looked clunky. So I'd prefer to leave the remaining "built-in"s alone.

Thanks for spotting the include guards, I'll fix that up!

kimgr:
> Thanks for spotting the include guards, I'll fix that up!

Done.

jbytheway:
Thanks for rationalizing.  I was torn about what form to use :).

kimgr:
@jbytheway Yeah, nobody should have to live through that. Thanks!

PR #656, opened 2019-03-09
CastExpr improvements

### LINKED ISSUE #650: IWYU removes a necessary include if there are a derived-to-base cast and a user-defined conversion in the same position
In this example, no changes should be suggested, but iwyu suggests removing include cast_and_user_defined_conversion-d2.h.

```
// cast_and_user_defined_conversion-d1.h
class Derived;
  
class X {
public:
  operator Derived*() const { return nullptr; }
} gX;


// cast_and_user_defined_conversion-d2.h
class Base { };
class Derived : public Base { };


// cast_and_user_defined_conversion.cc
#include "cast_and_user_defined_conversion-d1.h"
#include "cast_and_user_defined_conversion-d2.h"

Base* hello() {
  return static_cast< Base* >( gX );
}
```

In `VisitCastExpr(clang::CastExpr* expr)` function, iwyu checks if the cast expression is ptr-to-ptr cast. But iwyu uses `GetTypeOf(expr->getSubExprAsWritten())` as `from_type`, and this is not a pointer type in this case. So iwyu doesn't report the definition of Derived class is necessary.

Changing `const Type* const from_type = GetTypeOf(expr->getSubExprAsWritten());` to `const Type* const from_type = GetTypeOf(expr->getSubExpr());` will work in this case, but I don't know this change is correct.

kimgr:
Here's a series of commits that eventually fix #650.

I think the simplifications are generally sound, but I've only had the test suite to guide me.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2009,6 +2009,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     const Type* from_type = GetTypeOf(expr->getSubExprAsWritten());
     const Type* to_type = GetTypeOf(expr);
 
+    // If this cast requires a user-defined conversion of the from-type, look up
+    // its return type so we can see through up/down-casts via such conversions.
+    const Type* converted_from_type = nullptr;
+    if (const NamedDecl* conv_decl = expr->getConversionFunction()) {
+      converted_from_type =
+          cast<FunctionDecl>(conv_decl)->getReturnType().getTypePtr();
+    }
+
     std::vector<const Type*> required_full_types;
 
     // The list of kinds: http://clang.llvm.org/doxygen/namespaceclang.html
@@ -2112,7 +2120,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       case clang::CK_DerivedToBase:
       case clang::CK_UncheckedDerivedToBase:
       case clang::CK_DerivedToBaseMemberPointer:
+        // Just 'from' type is enough: full type for derived gets base type too.
         required_full_types.push_back(from_type);
+
+        // If this derived-to-base cast had an associated conversion function,
+        // it's a user-defined conversion operator. The from-type in this cast
+        // is not necessarily related to the base type, but the converted
+        // from-type must be, so make sure we require both.
+        if (converted_from_type) {
+          required_full_types.push_back(converted_from_type);
+        }
         break;
       case clang::CK_Dynamic:
         // Usually dynamic casting is a base-to-derived cast, but it is
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2006,13 +2006,11 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // compiler a fighting chance of generating correct code.
   bool VisitCastExpr(clang::CastExpr* expr) {
     if (CanIgnoreCurrentASTNode())  return true;
-    const Type* const from_type = GetTypeOf(expr->getSubExprAsWritten());
-    const Type* const to_type = GetTypeOf(expr);
-    const Type* const deref_from_type = RemovePointersAndReferences(from_type);
-    const Type* const deref_to_type = RemovePointersAndReferences(to_type);
+    const Type* from_type = GetTypeOf(expr->getSubExprAsWritten());
+    const Type* to_type = GetTypeOf(expr);
 
-    bool need_full_deref_from_type = false;
-    bool need_full_deref_to_type = false;
+    std::vector<const Type*> required_full_types;
+
     // The list of kinds: http://clang.llvm.org/doxygen/namespaceclang.html
     switch (expr->getCastKind()) {
       // This cast still isn't handled directly.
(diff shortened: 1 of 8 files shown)


iwyu.cc
```diff
@@ -2006,28 +2006,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    const Type* from_type = GetTypeOf(expr->getSubExprAsWritten());
+    const Type* to_type = GetTypeOf(expr);
+
+    // If this cast requires a conversion function, look up its return type so
+    // we can see through up/down casts via user-defined conversions.
+    const Type* converted_to_type = nullptr;
+    if (const NamedDecl* conv_decl = expr->getConversionFunction()) {
+      converted_to_type =
+          cast<FunctionDecl>(conv_decl)->getReturnType().getTypePtr();
     }
```
jru: The *if* block can be moved inside the `clang::.+DerivedToBase.+` cases below. That's the only place where it is needed.

iwyu.cc
```diff
@@ -2006,28 +2006,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-    // a pointer to a pointer, or any type to a reference.
-    // Unfortunately, when casting to a reference, clang seems to
-    // strip the reference off of to_type, so we need a separate
-    // function call to tell.
-    if (expr->getCastKind() != clang::CK_UserDefinedConversion &&
-        expr->getCastKind() != clang::CK_ConstructorConversion) {
-      if (!((from_type->hasPointerRepresentation() &&   // pointer or reference
-             to_type->hasPointerRepresentation()) ||
-            IsCastToReferenceType(expr)))
-        return true;    // we only care about ptr-to-ptr casts for this check
```
jru: Is this no longer true? Or covered by the switch?

iwyu.cc
kimgr: I suspect that more cases will need the converted-to-type (e.g. the dynamic-casts), but I haven't bothered trying to come up with testcases for it.

It makes intuitive sense to me to keep the from-type, to-type and converted-to-type resolution in the same place, because they represent the triad of types we need to reason about a cast expression (depending on which kind of cast it is).

So I'd rather keep it out here, but can move it if you insist.

iwyu.cc
kimgr: I modified the switch/case to ignore instead of warn for non-pointer-to-pointer cast kinds. That seems to satisfy the test suite, so I optimistically assumed that was a valid way to simplify (otherwise I would have had to factor the converted-to-type into this already-hairy predicate).

iwyu.cc
```diff
@@ -2118,43 +2102,54 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        // Just 'from' type is enough: full type for derived gets base type too.
+        required_full_types.push_back(from_type);
+
+        // If this derived-to-base cast requires a conversion function, it's a
+        // user-defined conversion operator. The from-type in this cast is not
+        // necessarily related to the base type, but the converted-to type
+        // must be.
+        if (converted_to_type) {
+          required_full_types.push_back(converted_to_type);
+        }
```
jru: I think clang will unwrap the casts one be one and reach the `CK_UserDefinedConversion` eventually. This should be moved there to cover all cases, not just those inside a `DerivedToBase`

kimgr: There was some complication about that... Yes --

This cast: `static_cast<Derived*>(Convertible())` will only require the full type for `Convertible`, not the converted-to-type, `Derived*` -- a fwd decl use is enough there. This is the expression we see in `CK_UserDefinedConversion`.

However, this cast: `static_cast<Base*>(&derived)` will require the full type for both `Base` and `Derived` because the cast needs to know how they're related.

So the combination of the two -- `static_cast<Base*>(Convertible())` will only see a cast from `Convertible` to `Derived*` for `CK_UserDefinedConversion`, and a cast from `Derived*` to `Base*` for `CK_DerivedToBase`.

There's no single cast expression that sees the context of all these, that's why the conversion function is a nice indicator that the derived-to-base cast needs more than the standard stuff.

iwyu.cc
jru: I see, that makes sense. So, any unsupported cast kind will not enter the switch in the first place, and get simply ignored, right?

iwyu.cc
jru: Oh got it. I had the feeling I might be missing something :)

I intuitively thought that we would always want a full-use of the *to* type for a user-defined conversion, but yeah, I guess it depends a lot on what you actually do with it.

Maybe you could mention why we need the *to* type here? Might also help to explain above that the *converted_to_type* might be useful for other cast kinds in the future, so that it warrants a separate definition.

kimgr: The function doc comment goes into some detail, but I'll expand a little on these `Type`s too, because it took me far too long to get my head around it.

iwyu.cc
kimgr: In the old implementation that was the case. In the new implementation, all cast kinds enter the switch, but most of them just exit without action.

kimgr: Also, it was necessary/simpler to remove this because a cast from `Convertible` to `Base*` is *not* a pointer-to-pointer cast, but will need to be handled as one.

iwyu.cc
kimgr: Wait, you ask why we need the "to type". We don't -- we need the *from* type, both the actual from type, and the from type after any user-defined conversions.

I wonder if the name `converted_**to**_type` is misleading -- I meant for that to read "the type we user-defined-converted to", but it sounds like you're interpreting it as "converted to-type"?

I think the comments explain what's going on pretty explicitly (improvement suggestions welcome here, if you have them!), but I'll find a better name for `converted_to_type` to try and reduce the confusion.

kimgr:
OK, see if this makes more sense. Thank you!

iwyu.cc
jru: That's how I read it, yes. I think it is clearer now. Thanks!

kimgr: :+1:

PR #657, opened 2019-03-10
Remove Google-specific handling of third-party

kimgr:
This is part of Google's original open-sourcing of IWYU.

They keep third-party libraries in a directory called, well,
'third-party', so there was lots of special casing for code in
that directory.

Remove that code, unit tests covering it and explicit mappings.

Keep one special case for allowing include cycles for files with
'internal/' in the name, to avoid breaking the include_cycle test
case. Not sure what to do about that longer term, but I didn't want to
remove the test case right now.


kimgr:
Thanks!

PR #658, opened 2019-03-11
Correct Qt mapping file to take in account both include formats (system and local with double quotes)

troopa81:
In ``iwyu_path_util.cc`` file, ``ConvertToQuotedInclude`` method, the needed ``qstring.h`` include is returned in system format (with < and >). 

Problem, in ``iwyu_output.cc`` file, ``CalculateIwyuForForwardDeclareUse`` method, this include is used to know if a forwarddeclare is already existing in desiredincludes, and it fails because there is a mapping on ``"qstring.h"`` with quotes and not <>.

It leads to an issue where iwyu propose to include and forwarddeclare QString.

I presume the best way to correct this is to modify the qt mapping file.

Beware that I didn't modify the following lines (I tried but I didn't figure out how to start this command in the first place)

```shell
# Generated with
#    perl -le "use File::Find;use File::Basename; sub wanted { $x = lc $_. '.h'; print '    { include: [ -@\-('.basename($File::Find::dir).'/)?'.$x.'\--, -private-, -<'.$_.'>-, -public- ] },' if -e $x } find(\&wanted, '.')"
# on windows
```


haubourg:
ping @3nids !

kimgr:
@troopa81 Thank you for this, it seems a like nice (if brute-force) way to improve the situation for Qt!

Most of our mappings include something like that comment to ease automatic refresh (e.g. for a new version of Qt), so it would be nice if we could update it to generate the `[<"]` pattern directly.

I can take a stab at it if you feel out of your depth, but I'm not sure I have enough Qt here to test it properly.

troopa81:
I figure out how to use the perl script line but it doesn't generate the mappings the way they are in the file. Maybe It needs a specific version of perl. 

Other problem, the first part (symbol part) doesn't seem to be generated with the same Qt version. 

Maybe, it will be better to have one script file (python or shell) to generate all qt mappings according to a given Qt version. 

What do you think ? I can do this is you're OK

kimgr:
I couldn't get it to work as-was eiter.

Yes, I think it would be great to ship a script that could build the mappings on the target system instead, that way we're more future-proof (barring larger changes in Qt, but that bridge can be crossed when we get there).

Python is preferred, since we're also targetting Windows -- it's easier to port. Thanks!

troopa81:
I add a script python in order to generate mappings adb5d06  

However, I'm facing 2 issues : 
- There are several file pour the same class name (QAction in QtWidgets and QAction in Qt3DInputs for instance)
- There are several classes pointing to the same include file (QPointF and QPoint to QtCore/qtpoint.h for instance)

The script display warnings for this problems. Here is the generated output for Qt 5.11 [warnings.txt](https://github.com/include-what-you-use/include-what-you-use/files/3037716/warnings.txt)

I don't know how to deal with these issues. Could you tell me what would be the best option?

kimgr:
It should be possible to map the same symbol to multiple files, I think IWYU will accept any of them if they already exist, but suggest the first one if not. Something like that.

Same with the multiple-symbol-same-file problem.

troopa81:
OK, so generated mappings can stay that way. 

I just added an option in order to display the warnings or not.

troopa81:
@kimgr Is there any reason or something to be corrected that prevent this PR to be merged?

kimgr:
@troopa81 Sorry, I haven't had a chance to review this again; work and life got in the way. Hoping to find some time next week!

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python3
```
kimgr: Please leave this at `python`, and let's make sure the script is both Python2- and Python3-compatible.

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+
+"""
+This script generated the Qt mapping file according to given Qt include
```
kimgr: typo: generated -> generates

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+
+"""
+
+import argparse
+import glob
+import os
+import re
+
+if __name__ == "__main__":
+
```
kimgr: Excess blank line

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+   $ ./generate_qt_mappings.py /usr/include/x86_64-linux-gnu/qt5 qt5_11.imp
+
+"""
+
+import argparse
+import glob
+import os
+import re
+
```
kimgr: Some linters want two lines between imports and the first function/statement, so add an extra blank line here

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("qt_include_dir", help="Qt include directoy")
+    parser.add_argument("output_file", help="Generated output mapping file")
+    parser.add_argument("--warnings", action="store_true", help="Display possible warning messages")
+    args = parser.parse_args()
+
```
kimgr: I usually call out to a module-level `main`-function here, which makes it possible to import the module and call `main` with arguments. So I suggest adding:

```
def main(qt_include_dir, output_file, warnings):
    ...
```

and then doing: `sys.exit(main(args.qt_include_dir, args.output_file, args.warnings))` here.

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+    headers = glob.glob(os.path.join(args.qt_include_dir, '**/*[!\\.h]'))
```
kimgr: Move this to just before the `for` loop

kimgr: Also, do you think this glob expression works on Windows? Do you have any way to test?

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+    symbols_map = []
+    includes_map = []
+
+    class2includes = {}
+    include2classes = {}
+    for header in headers:
+
```
kimgr: Excess blank line

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+        if os.path.isdir(header):
+            continue
+
+        class_name = os.path.basename(header)
+        module_name = os.path.basename(os.path.dirname(header))
+
+        symbols_map += [f'{{ symbol: [ "{class_name}", "private", '
```
kimgr: Rather than building a long list of strings, could you build a list of dicts that you can then serialize to json?

Something like:

    # With quoting around the class_name
    symbols_map.append({'symbol': [class_name, "private", class_name, "public"]})

And:

    json.dump(symbols_map, f, indent=2)

That way you could also add includes and symbols to the same list, or if you want to separate them, concat the lists at the end (much like you're doing now).

kimgr: And it might be worth adding a `def symbol_record(class_name)` function that generates that dict with the right quoting, etc.

kimgr: Hmm. Why is the mapping from symbol to symbol? Shouldn't it be symbol to quoted-include? Ah, I see, you're back-generating the quoted-include from the class name. Hmm. Maybe that's the best we can do :)

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+                        f'"<{class_name}>", "public" ] }}']
+
+        content = open(header, 'r').read()
```
kimgr: with open(header, 'r') as f:
        content = f.read()

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+        includes = re.findall(r'#include "(.*)\.h"', content)
+        if len(includes) != 1:
+            continue
+
+        include = includes[0]
+        class2includes[class_name] = (class2includes.get(class_name, [])
```
kimgr: This is what `setdefault` is for:

    class2includes.setdefault(class_name, []).append(header)

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+                                      + [header])
+
+        full_include = module_name + "/" + include
+        include2classes[full_include] = (include2classes.get(full_include, [])
+                                         + [class_name])
+
+        includes_map += [
```
kimgr: Build proper data structure here and worry about json serialization later

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+        includes = re.findall(r'#include "(.*)\.h"', content)
+        if len(includes) != 1:
+            continue
+
+        include = includes[0]
+        class2includes[class_name] = (class2includes.get(class_name, [])
+                                      + [header])
+
+        full_include = module_name + "/" + include
+        include2classes[full_include] = (include2classes.get(full_include, [])
```
kimgr: `setdefault`

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+
+        class_name = os.path.basename(header)
+        module_name = os.path.basename(os.path.dirname(header))
+
+        symbols_map += [f'{{ symbol: [ "{class_name}", "private", '
+                        f'"<{class_name}>", "public" ] }}']
+
+        content = open(header, 'r').read()
+        includes = re.findall(r'#include "(.*)\.h"', content)
+        if len(includes) != 1:
```
kimgr: I don't understand this -- maybe a comment would help? Wouldn't there typically be more than one `#include "..."` in a given Qt header?

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+            f'{{ include: [ "@[\\"<]({module_name}/)?{include}\\\\.h[\\">]", '
+            f'"private", "<{class_name}>", "public" ] }}']
+
+    # Print some warnings
+    if args.warnings:
+        class2includes = [(class_name, includes) for class_name, includes in
+                          class2includes.items() if len(includes) > 1]
+
+        if class2includes:
+            print("Warning! These classes references several includes file:")
```
kimgr: typo: references -> reference

kimgr: typo: includes file -> include files

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+        include2classes[full_include] = (include2classes.get(full_include, [])
+                                         + [class_name])
+
+        includes_map += [
+            f'{{ include: [ "@[\\"<]({module_name}/)?{include}\\\\.h[\\">]", '
+            f'"private", "<{class_name}>", "public" ] }}']
+
+    # Print some warnings
+    if args.warnings:
+        class2includes = [(class_name, includes) for class_name, includes in
```
kimgr: Maybe not rebind `class2includes` here. I don't understand why it's warning, maybe the variable name could reflect the problem we're worried about?

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+                          class2includes.items() if len(includes) > 1]
+
+        if class2includes:
+            print("Warning! These classes references several includes file:")
+            for class_name, includes in class2includes:
+                print("  {} : {}".format(class_name, ",".join(includes)))
+
+        include2classes = [(include, classes) for include, classes in
```
kimgr: Again, don't reuse the `include2classes` name

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+                           include2classes.items() if len(classes) > 1]
+        if include2classes:
+            print("Warning! These includes references several classes:")
```
kimgr: typo: references

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+            for include, classes in include2classes:
+                print("  {} : {}".format(include, ",".join(classes)))
```
kimgr: I think `", "` with a space after comma produces nicer formatting

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # Print some warnings
+    if args.warnings:
+        class2includes = [(class_name, includes) for class_name, includes in
+                          class2includes.items() if len(includes) > 1]
+
+        if class2includes:
+            print("Warning! These classes references several includes file:")
+            for class_name, includes in class2includes:
+                print("  {} : {}".format(class_name, ",".join(includes)))
```
kimgr: I think `", "` with a space after comma produces nicer formatting

generate_qt_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+
+        include2classes = [(include, classes) for include, classes in
+                           include2classes.items() if len(classes) > 1]
+        if include2classes:
+            print("Warning! These includes references several classes:")
+            for include, classes in include2classes:
+                print("  {} : {}".format(include, ",".join(classes)))
+
+    f = open(args.output_file, 'w')
```
kimgr: with open(args.output_file, 'w') as f:
        ...

I'd also use the Python3 print function (with `from __future__ import print_function` at the top of the file):

    print("# This file has been generated by the script {}. Do not edit!".format(
            os.path.basename(__file__)), file=f)

and then use the `json` module to serialize the record list.

generate_qt_mappings.py
troopa81: The script now works on Windows f15e0dc

generate_qt_mappings.py
troopa81: > json.dump(symbols_map, f, indent=2)

Using json dump produce a more indented (and less readable?) result 

```json
{
    "symbol": [
      "QComputeCommand",
      "private",
      "<QComputeCommand>",
      "public"
    ]
  },
  {
    "include": [
      "\"@[\\\"<](Qt3DRender/)?qcomputecommand\\\\.h[\\\">]\"",
      "private",
      "<QComputeCommand>",
      "public"
    ]
  }
```
Are you sure you still want to use json module to write the symbols and includes records?

troopa81: > Why is the mapping from symbol to symbol? 

To be honest I try to make the symbol mapping the closer from the original qt5_4.imp file without really understanding what it implies :) Remember that my initial goal was to generate a mapping for include file who deals with quotes and <>

generate_qt_mappings.py
troopa81: See comment https://github.com/include-what-you-use/include-what-you-use/pull/658#discussion_r295170058

troopa81:
Thanks @kimgr for the review!

I made most of the modifications you report, but there is still points to discuss

> I don't quite understand why we're warning about 1-many relationships, can you expand on that?

and 

> I don't understand this -- maybe a comment would help? Wouldn't there typically be more than one #include "..." in a given Qt header?

I was not sure that IWYU was ok with 1-n relationship and I don't really see exactly how it dealt with. You seem confident on the subject and I made some tests. It looks like it works, so I remove this warning (7abece3 and ed52b72) and iterate on include (25998da)

generate_qt_mappings.py
troopa81: See https://github.com/include-what-you-use/include-what-you-use/pull/658#issuecomment-503576619

kimgr:
I didn't mean to sound confident, I genuinely didn't understand, and was hoping you could explain it to me :)

troopa81:
@kimgr Do you know if there is anyone who could enlighten us on the subject. 

It seems that the previous mapping file was only fill with 1-1 relationship, no 1-n relationship. I'm not sure it was made on purpose...

generate_qt_mappings.py
```diff
@@ -0,0 +1,59 @@
...
+            includes_map += [
+                '{{ include: [ "@[\\"<]({}/)?{}\\\\.h[\\">]", '.format(
+                    module_name, include)
+                + '"private", "<{}>", "public" ] }}'.format(class_name)]
+
+    with open(args.output_file, 'w') as f:
+        print("# This file has been generated by the script {},"
+              "do not edit this file!\n[\n\t{}\n]".format(
+                  os.path.basename(__file__),
+                  ",\n\t".join(symbols_map + includes_map)), file=f)
```
kimgr: I'd still recommend using the `json` module. Something like this:

```
import json

# when building the output
symbols.append({"symbol": [class_name, "private", "<%s>" % class_name, "public" ]})
includes_map.append({"include": ["@[\\"<]({}/)?{}\\\\.h[\\">]", '.format(module_name, include), "private", "<{}>".format(class_name), "public"]}  # this is probably wrong now, I copied/pasted/edited without checking.

# when dumping to file
with open(args.output_file, 'w') as f:
    print("Some stuff for the header, file=f)
    json.dump(symbols, f, indent=2, sort_keys=True)
    json.dump(include_map f, indent=2, sort_keys=True)
```

That _might_ produce excessive indentation and line wrapping, but it's worth trying out to reduce the risk of misformatting JSON

generate_qt_mappings.py
```diff
@@ -0,0 +1,59 @@
...
+
+    headers = glob.glob(os.path.join(args.qt_include_dir, '**/*[!.h]'))
+    for header in headers:
+        if os.path.isdir(header):
+            continue
+
+        class_name = os.path.basename(header)
+        module_name = os.path.basename(os.path.dirname(header))
+
+        symbols_map += ['{{ symbol: [ "{}", "private", '.format(class_name)
```
kimgr: I realize it's old-school and goes against current recommendations, but I much prefer the C-style format strings to `.format`, because it's more succinct and I more readable without all the curly braces.

kimgr:
> It seems that the previous mapping file was only fill with 1-1 relationship, no 1-n relationship.
> I'm not sure it was made on purpose...

I was only working on assumption, because the default IWYU mappings have several 1-n, n-1 mappings, and they seem to work OK most of the time.

generate_qt_mappings.py
```diff
@@ -0,0 +1,59 @@
...
+                        + '"<{}>", "public" ] }}'.format(class_name)]
+
+        includes = []
```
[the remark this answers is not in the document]
troopa81: But if the include file opening failed, the `includes` variable will be not undefined and the program will crash.

generate_qt_mappings.py
```diff
@@ -0,0 +1,59 @@
...
+        with open(header, 'r') as f:
+            content = f.read()
+            includes = re.findall(r'#include "(.*)\.h"', content)
```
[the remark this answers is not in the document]
troopa81: If file opening fails, content variable will be undefined and program will crash in an ugly way no?

generate_qt_mappings.py
troopa81: I tried using json module to serialize mapping like you describe above but I face a major issue: mapping file are not really json file.

Quote from the [documentation](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUMappings.md) 
> Single-word strings can be left un-quoted

Actually it HAS to be unquoted. If I use quote around `symbol` and `include` keyword (meaning the way json module generates file), mappings are not taking into account.

troopa81:
> I was only working on assumption, because the default IWYU mappings have several 1-n, n-1 mappings, and they seem to work OK most of the time.

Seems a good point for 1-n and n-1 relationship. 

Regarding json serialisation with json module, like I said in my comment (https://github.com/include-what-you-use/include-what-you-use/pull/658#discussion_r296691187) it won't work. So, for now, I let it the way it is.

generate_qt_mappings.py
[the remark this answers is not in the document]
kimgr: If file opening fails, `open` will crash in an ugly way :)

generate_qt_mappings.py
kimgr: @troopa81 Wow; I hadn't noticed that before. We use LLVM's YAML module to deserialize (JSON is allegedly a subset of YAML), and it has this strange behavior wrt quoting. It didn't occur to me that anything had to be done to unquote. I'll have to look into that; could you open a bug?

kimgr:
Aargh, GitHub's review UI drives me crazy. I hadn't seen your earlier responses. I'll take another look at this today.

generate_qt_mappings.py
troopa81: Done https://github.com/include-what-you-use/include-what-you-use/issues/697

generate_qt_mappings.py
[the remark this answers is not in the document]
troopa81: Done cf434b8
I had a bad understanding of `with` keyword

generate_qt_mappings.py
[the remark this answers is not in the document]
troopa81: Done cf434b8

generate_qt_mappings.py
kimgr: :+1:

generate_qt_mappings.py
```diff
@@ -0,0 +1,59 @@
...
+            includes_map += [
+                '{ include: [ "@[\\"<](%s/)?%s\\\\.h[\\">]", ' % (
+                    module_name, include)
+                + '"private", "<%s>", "public" ] }' % class_name]
+
+    with open(args.output_file, 'w') as f:
+        print("# Do not edit! This file was generated by the script %s." %
+              os.path.basename(__file__), file=f)
+        print("[", file=f)
+        print("\t%s" % ",\n\t".join(symbols_map + includes_map), file=f)
```
kimgr: Please prefer tabs to spaces; use two spaces instead of `\t`.

troopa81: Done

kimgr:
This looks great! Are you comfortable enough with Git to rewrite the history of this branch? I'd like it as three commits --

1) Script
2) New Qt mappings for 5.11
3) Update of existing Qt 5.4 mappings using script (or did you do them manually?)

I'd be happy to do it for you, but I'm not sure if I can force-push to your branch. There should be a little checkbox on the pull request, per: https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork. Not sure if it allows history rewriting, though.

troopa81:
I rewrite history accordingly.

The Qt 5.4 mapping file has been modified manually, I let it the way I change it for now, but I'm not sure this is interesting to keep it. There is a lot of Qt releases, do you plan to have several mapping file for each? If you ask me, I would just keep only one for the last stable version and let user generate their own with the script in case they use a different version.

kimgr:
> If you ask me, I would just keep only one for the last stable version
> and let user generate their own with the script in case they use a different version.

Yeah, maybe that makes sense. Or we can accumulate a set of tested/curated generated `.imp` files, so it's easier for Qt users to just pick and choose. We'll always be behind "latest", anyway.

troopa81:
Thanks for review and merge @kimgr

PR #665, opened 2019-03-20
Generalize include file mapping to handle different targets.

trixirt:
AddTripleMappings is here mapping specific handling is done.
If nothing is handled, then false is returned and the
AddDefaultMappings is called.

To exercise this interface, as very basic FreeBSD include file
mapping was done.  The external map file is freebsd.imp.

It was noticed that the internal c representation is close the
imp representation.  So instead of handcoding the c representation
the imp2array.py script was created and the c code generated from

imp2array.py freebsd.imp freebsd_include_map > freebsd_include_map.h


kimgr:
I like this!

There's too much going on at once, though:

* Script to generate .h from .imp
* Derive mappings from target triple
* FreeBSD as a motivating example

I would like to start with the first bullet, and integrate that into the build process instead, so we don't keep mappings both in JSON and .h form in the repo (only the JSON form would be necessary). Then we can think about the two others.

Kudos for putting this together as a whole, though -- it gets the point across clearly how it would work!

I'm also a little worried about keying primarily on the target triple, I think we need to take into account which libc and C++ stdlib we are using. I'm not sure if the mappings you've spec-ed here are part of the operating system FreeBSD, or the libc that ships with FreeBSD. I think this is less of a problem for FreeBSD and Windows than Linux, where there seem to be multiple libc providers.

But we can worry about that when we have a nice infrastructure for generating headers from `.imp` files :)

trixirt:
Agreed.
Moving this to issue

PR #666, opened 2019-03-20
 Report expanded loc for uses partially in a macro

### LINKED ISSUE #659: IWYU reports a wrong use location when a template function is called with macro
In this example, no changes should be suggested, but iwyu suggests changing `#include macro_and_template-d2.h` to `class Test;`.

```
// macro_and_template.cc
#include "macro_and_template-d1.h"
#include "macro_and_template-d2.h"

void hello() {
  MACRO().func<Test>();    // If you write MacroClass().func<Test>();, it is OK
}

// macro_and_template-d1.h
#define MACRO() MacroClass()

class MacroClass {
public:
  template <typename T> T func() { return T(); }
};

// macro_and_template-d2.h
class Test {};
```

Here is a verbose log:
```
[Uninstantiated template AST-node] T
tests/cxx/macro_and_template-d1.h:5:27: (5 in tpl) [ CXXMethodDecl ] 0x1090298f8 template<> Test func<Test>()
tests/cxx/macro_and_template-d1.h:5:25: (6 in tpl) [ FunctionProtoType ] 0x1090298b0 class Test (void)
tests/cxx/macro_and_template-d1.h:5:25: (6 in tpl) [ FunctionProtoTypeLoc ] class Test (void)
tests/cxx/macro_and_template-d1.h:5:25: (7 in tpl) [ SubstTemplateTypeParmType ] 0x109020330 class Test
(For type class Test):
Trying to determine use location for 'class Test {}'
Macro is defined in file 'tests/cxx/macro_and_template-d1.h'. Looking for fwd-decl hint...
Attributing use of 'class Test {}' to location at tests/cxx/macro_and_template-d1.h:1:17.
Marked full-info use of decl 0x10901f9a0 Test (from tests/cxx/macro_and_template-d2.h:1:7) at tests/cxx/macro_and_template-d1.h:1:17
```

So IWYU reports macro_and_template-d1.h uses the declaration of class Test but macro_and_template.cc does not.

I found `GetCanonicalUseLocation` says:
```
// 1) If the use_loc is in <scratch space>, we assume it's formed by macro
//    argument concatenation, and attribute responsibility to the expansion
//    location.
// 2) If the macro definition file forward-declares the used decl, that's a
//    hint that it wants the expansion location to take responsibility.
//
// Otherwise, the spelling loc is responsible.
```

However, in this case, the expansion location seems to be responsible for the use of Test.

jru:
This is a WIP to fix #659 and similar cases.

Not yet completely done, but I wanted to put it out there to a start a discussion and get the general feeling  :)

Some things that I still want to do:
 - [ ] Verify that no macro is expanded throughout the source range, not just at the ends.
 - [ ] Add more complex test scenarios
 - [ ] Check if other open issues would/could be easily fixed by this

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1391,6 +1391,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     SourceLocation spelling_loc = sm->getSpellingLoc(use_loc);
     SourceLocation expansion_loc = sm->getExpansionLoc(use_loc);
 
+    SourceRange use_range = use_node.GetSourceRange();
+
+    if (use_range.getBegin().isFileID() != use_range.getEnd().isFileID()) {
+      VERRS(4) << "Use of decl '" << PrintableDecl(decl)
+               << "' is partially inside a macro."
+                  " Attributing to expansion location at "
+               << PrintableLoc(expansion_loc) << ".\n";
+      return expansion_loc;
+    }
+
     // If the file defining the macro contains a forward decl, keep it around
     // and treat it as a hint that the expansion loc is responsible for the
     // symbol.
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -151,6 +151,11 @@ using clang::ASTConsumer;
 using clang::ASTContext;
 using clang::ASTFrontendAction;
 using clang::Attr;
+using clang::CallExpr;
+using clang::ClassTemplateDecl;
+using clang::ClassTemplateSpecializationDecl;
+using clang::CompilerInstance;
+using clang::ConstructorUsingShadowDecl;
 using clang::CXXConstructExpr;
 using clang::CXXConstructorDecl;
 using clang::CXXCtorInitializer;
@@ -160,11 +165,6 @@ using clang::CXXMethodDecl;
 using clang::CXXNewExpr;
 using clang::CXXOperatorCallExpr;
 using clang::CXXRecordDecl;
-using clang::CallExpr;
-using clang::ClassTemplateDecl;
-using clang::ClassTemplateSpecializationDecl;
-using clang::CompilerInstance;
-using clang::ConstructorUsingShadowDecl;
 using clang::Decl;
 using clang::DeclContext;
 using clang::DeclRefExpr;
@@ -179,22 +179,23 @@ using clang::FunctionProtoType;
 using clang::FunctionTemplateDecl;
 using clang::FunctionType;
 using clang::ImplicitCastExpr;
-using clang::LValueReferenceType;
 using clang::LinkageSpecDecl;
+using clang::LValueReferenceType;
 using clang::MemberExpr;
 using clang::NamedDecl;
 using clang::NestedNameSpecifier;
 using clang::NestedNameSpecifierLoc;
 using clang::OverloadExpr;
 using clang::ParmVarDecl;
-using clang::PPCallbacks;
 using clang::PointerType;
-using clang::QualType;
+using clang::PPCallbacks;
 using clang::QualifiedTypeLoc;
+using clang::QualType;
 using clang::RecordDecl;
 using clang::RecursiveASTVisitor;
 using clang::ReferenceType;
 using clang::SourceLocation;
+using clang::SourceRange;
 using clang::Stmt;
 using clang::SubstTemplateTypeParmType;
 using clang::TagDecl;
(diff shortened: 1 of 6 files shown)


kimgr:
Cool! I had thought about pushing `ASTNode` forward in the analysis chain, but it hadn't occurred to me that it's enough to push it into the `Report` methods to get better results. I was expecting we would have to keep them alive for `iwyu_output` to work on them.

I'm not so sure about the pointer-to-reference transform. It looks like HUGE diff with marginal returns. If `current_ast_node()` is indeed never NULL, I think we should just remove whatever null checks there are and leave it as-is.

kimgr:
If you could back that out (from this PR, at least), it will be much easier to review the awesome macro fix as a whole.

jru:
Yeah. I can revert that. I just thought it is a good time to clean it up, now that I am touching *ASTNode*.

FYI: I am out for the next couple of weeks, so I won't be able to push changes.

jru:
Got rid of the ponter-to-ref refactor commit.

iwyu.cc
```diff
@@ -1391,6 +1391,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     SourceLocation spelling_loc = sm->getSpellingLoc(use_loc);
     SourceLocation expansion_loc = sm->getExpansionLoc(use_loc);
 
+    SourceRange use_range = use_node.GetSourceRange();
+
```
kimgr: Can remove this blank line

iwyu.cc
```diff
@@ -151,6 +151,11 @@ using clang::ASTConsumer;
 using clang::ASTContext;
 using clang::ASTFrontendAction;
 using clang::Attr;
+using clang::CallExpr;
```
kimgr: This `using` sorting needs to go in a separate commit

iwyu.cc
```diff
@@ -1766,21 +1769,20 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        IwyuBaseAstVisitor<Derived>::ReportDeclUse(used_node, decl, comment);
       }
     }
   }
 
-  void ReportTypesUse(SourceLocation used_loc, const set<const Type*>& types) {
-    for (const Type* type : types)
-      ReportTypeUse(used_loc, type);
+  void ReportTypesUse(const ASTNode& used_node, const set<const Type*>& types) {
+    for (const Type* type : types) ReportTypeUse(used_node, type);
```
kimgr: No one-liners, please; some debuggers can't handle breakpoints like that.

iwyu.cc
```diff
@@ -1638,7 +1641,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // The comment, if not nullptr, is extra text that is included along with
   // the warning message that iwyu emits. The extra use flags is optional
   // info that can be assigned to the use (see the UF_* constants)
-  virtual void ReportDeclUse(SourceLocation used_loc,
+  virtual void ReportDeclUse(const ASTNode& used_node,
```
kimgr: Is it `used_node` or `use_node`? I think it denotes _where_ `used_decl` is actually used, so `use_node` seems better... Not sure. :man_shrugging:

iwyu.cc
```diff
@@ -2910,32 +2920,32 @@ class InstantiatedTemplateVisitor
...
                      UseFlags extra_use_flags = 0) override {
-    const SourceLocation actual_used_loc = GetLocOfTemplateThatProvides(decl);
-    if (actual_used_loc.isValid()) {
+    if (ASTNode const* actual_used_node =
+            GetAstNodeOfTemplateThatProvides(decl)) {
       // If a template is responsible for this decl, then we don't add
       // it to the cache; the cache is only for decls that the
       // original caller is responsible for.
-      Base::ReportDeclUse(actual_used_loc, decl, comment, extra_use_flags);
+      Base::ReportDeclUse(*actual_used_node, decl, comment);
```
kimgr: The flags? And likewise on the next call?

jru:
> Some comments inline. And I'm assuming you want to clean up the tests and make them real tests.

Thanks for the review! I'll take a look. 
There are some tests I still need to think about, so yeah, I will turn them into proper ones.

> .. we used to interpret forward-declares in macros as a signal that the expansion loc should be responsible ..

Good point! I forgot about the side-effects of the function; maybe one more reason to refactor it :) 

In general, I still need to think about a couple of cases that I'd like to cover, so the implementation will likely change. I just hope I can get back to it soon!

kimgr:
> I forgot about the side-effects of the function; maybe one more reason to refactor it :)

Yeah, the current behavior is no good. I just couldn't find a better way when I was in the middle of it, but I'm sure there's a nice factoring lurking somewhere :)

kimgr:
WIP for a year and a half. Closing now, hoping it will be reopened/repurposed soon!

Issue #663, opened 2019-03-15
iwyu crash when given option --max_line_length but no immediate =value


Guy-Shaw:
The command `iwyu -Xiwyu --max_line_length=120 file.c` works,
but the command `wyu -Xiwyu --max_line_length 120 file.c`  gets SIGSEGV
and drops core.  When parsing this option, atoi(optarg) can consume a NULL pointer.
The argument is optional.  This may seem like a contrived test, but option parsing
is not done the same way by all programs, and many programs allow options
with optional arguments specified using either `--option=value`  or  `--option value`.

kimgr:
Thanks for finding this! I think we've confused `optional_argument` to mean "this entire option is optional". `--max_line_length` and a few others should most likely be marked `required_argument`.

kimgr:
I started on this, and it turns out it's a little more complicated. Since IWYU only claims arguments prefixed by `-Xiwyu`, the splitting only picks up one `argv` after `-Xiwyu`. With `iwyu -Xiwyu --max_line_length 120`, the trailing `120` isn't even consumed by IWYU's arg parser.

That said, it shouldn't crash. I'll make it validate optarg for all argumented options, but it'll take me a little while.

kimgr:
I'm not sure what I ran into the other day; I must've tested with an outdated executable or something. With the suggested PR I get nicer behavior (help text on bad input). Let me know how it works out.

kimgr:
Thanks for the report, now fixed!

PR #667, opened 2019-03-21
Getopt fixes

kimgr:
Fix issue #663 by correctly marking options as `required_argument` and `no_argument` respectively.

Also remove unused and never-implemented switches `--how_to_debug` and `--cwd`.


jru:
Looks good!

PR #675, opened 2019-03-31
Re-export types to overridden method signatures

kimgr:
When a method overrides a virtual method in a base class, all types
mentioned in the signature will already be covered for by the base's
header.

As we must derive from a base to override methods, and derivation
requires a complete type, we already know that we've included the base
header to be able to override a method.

Exploit this and ignore all uses in overridden method signatures as they
must already be satisifed for the code to compile.

For example:

    // base.h
    #include <string>
    struct Base {
      std::string method() const;
    };

    // derived.h
    #include "base.h"
    struct Derived : public Base {
      // No warning about <string> here.
      std::string method();
    };

But this does not extend to the all uses of the type. For example, if we
use the type in the body of an overridden method, the definition file
must still include the type:

    // derived.cc
    #include "derived.h"

    // This mention of std::string is ignored, as above.
    std::string Derived::method() {
      // But this requires <string>.
      return std::string("abc");
    }

This reduces the number of required includes in headers significantly.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2559,23 +2559,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // declarations are forward-declarable.  This is true, *except*
     // for the exception (throw) types.  We clean that up here.
     // TODO(csilvers): figure out how to do these two steps in one place.
+    ASTNode* ast_node = current_ast_node();
     const FunctionProtoType* fn_type = nullptr;
-    if (!fn_type) {
-      fn_type = current_ast_node()->template GetParentAs<FunctionProtoType>();
+    if (const auto* fn_decl = ast_node->GetAncestorOfType<FunctionDecl>()) {
+      fn_type = dyn_cast<FunctionProtoType>(GetTypeOf(fn_decl));
     }
-    if (!fn_type) {
-      if (const FunctionDecl* fn_decl
-          = current_ast_node()->template GetParentAs<FunctionDecl>())
-        fn_type = dyn_cast<FunctionProtoType>(GetTypeOf(fn_decl));
-    }
+
     if (fn_type) {
-      for (FunctionProtoType::exception_iterator it =
-               fn_type->exception_begin();
-           it != fn_type->exception_end(); ++it)
-        if (it->getTypePtr() == type) {  // *we're* an exception decl
-          current_ast_node()->set_in_forward_declare_context(false);
+      for (const QualType& e : fn_type->exceptions()) {
+        if (e.getTypePtr() == type) {  // *we're* an exception decl
+          ast_node->set_in_forward_declare_context(false);
           break;
         }
+      }
     }
 
     return Base::VisitType(type);
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -86,6 +86,7 @@ using clang::NamedDecl;
 using clang::NestedNameSpecifier;
 using clang::ObjCObjectType;
 using clang::OverloadExpr;
+using clang::ParmVarDecl;
 using clang::PointerType;
 using clang::QualType;
 using clang::QualifiedTemplateName;
@@ -112,6 +113,7 @@ using clang::Type;
 using clang::TypeAliasTemplateDecl;
 using clang::TypeDecl;
 using clang::TypeLoc;
+using clang::TypeSourceInfo;
 using clang::TypedefNameDecl;
 using clang::TypedefType;
 using clang::UnaryOperator;
@@ -270,11 +272,80 @@ bool IsNodeInsideCXXMethodBody(const ASTNode* ast_node) {
   return false;
 }
 
+// This is subtle and a little complicated.
+//
+// In order to override a method, we must derive from a base class, and to do
+// that we must include the base class header. So we can use that as an include
+// optimization; there's no reason for derived class headers to include or
+// declare all the same types as their base class header for the purpose of
+// spelling out the override. We already know all those types are available
+// because the base class header _is_ included.
+
+// There's no canonical place in AST traversal to detect if a use is part of an
+// overridden method signature, VisitFunctionDecl reports full uses of return
+// type and 'autocast' parameters. All other uses are reported as descendant AST
+// nodes are visited (such as exception specifications, pointer and reference
+// type parameters, etc.)
+//
+// Instead, whenever a use is reported we walk back up the tree and see if
+// we're part of an override signature, which can then be ignored.
+//
+// Return true if the given ast_node is inside the signature of a C++ method
+// that overrides a virtual method in a base class.
+static bool IsNodeInsideCXXMethodOverrideSignature(const ASTNode* ast_node) {
+  // This analysis only makes sense for the signature, not the body.
+  // But the IsNodeInsideCXXMethodBody check is basically as expensive as this
+  // entire function, so let's not do it again.
+  // CHECK_(!IsNodeInsideCXXMethodBody(ast_node));
+
+  // Walk up and find a containing CXXMethodDecl, and make sure it overrides a
+  // base method.
+  const auto* method = ast_node->GetAncestorOfType<CXXMethodDecl>();
+  if (!method || !method->size_overridden_methods()) {
+    return false;
+  }
+
+  // Now we know this node is part of a C++ method signature overriding a base.
+
+  if (HasCovariantReturnType(method)) {
+    // Covariant return types require that the full return type is available,
+    // so don't special-case the return type for functions with covariance.
+  } else {
+    // See if this node is part of the return type.
+
+    // VisitFunctionDecl reports return type with the function's AST node, so
+    // ignore everything coming from the method itself.
+    if (ast_node->ContentIs(method)) {
+      return true;
+    }
+
+    if (ast_node->ContentIs(method->getReturnType().getTypePtr())) {
+      return true;
+    }
+
+    // The return type could be a type loc, though never seen this in practice.
+    const TypeSourceInfo* tsi = method->getTypeSourceInfo();
+    const TypeLoc retloc = tsi->getTypeLoc();
+    if (ast_node->ContentIs(&retloc)) {
+      return true;
+    }
+  }
+
+  // See if this node is part of the method's parameter list.
+  const auto* node_param = ast_node->GetAncestorOfType<ParmVarDecl>();
+  if (node_param) {
+    for (const ParmVarDecl* param : method->parameters()) {
+      if (param == node_param) {
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
 UseFlags ComputeUseFlags(const ASTNode* ast_node) {
   UseFlags flags = UF_None;
-
-  if (IsNodeInsideCXXMethodBody(ast_node))
-    flags |= UF_InCxxMethodBody;
 
   // Definitions of free functions are a little special, because they themselves
   // count as uses of all prior declarations (ideally we should probably just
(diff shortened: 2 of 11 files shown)


kimgr:
Should close #105.

jru:
Will this also work if the return type (or param) of a base method is only forward-declared in `base.h`? A test showing that would be good.

kimgr:
Thanks, I'll look into it. There's also the question of the meaning of "works" here 🙂. Forward-declaring the return type is a hint that users are responsible, so I need to make sure the transitivity works all the way through use of derived to derived to base.

kimgr:
I added some tests for forward-declares, which led to a slight improvement -- the compiler disallows a method definition if the return type is incomplete, even if the type is not mentioned in the body. So I adjusted the check to only cover pure declarations.

This seems to work according to expectations, please take another look.

iwyu.cc
```diff
@@ -1875,6 +1875,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (IsFriendDecl(decl))
       return true;
 
+    // We ignore all uses of types in method declarations overriding a virtual
+    // function; these types are already provided by the base class header,
+    // which must be included by the derived.
+    if (IsOverriddenMethodDecl(decl))
```
jru: I am wondering why we need this check. All types in a method override decl will be ignored anyway, right?

iwyu_ast_util.cc
```diff
@@ -1412,4 +1430,13 @@ TemplateArgumentListInfo GetExplicitTplArgs(const Expr* expr) {
   return explicit_tpl_args;
 }
 
+bool IsOverriddenMethodDecl(const clang::Decl* decl) {
+  if (const auto* method = dyn_cast_or_null<CXXMethodDecl>(decl)) {
+    return (!method->isThisDeclarationADefinition() &&
```
jru: I don't think this belongs here. There could be cases where knowing that a decl is an override is useful whether it has a body or not.

I think it should be moved to `IsPartOfOverrideDecl` where it could be made more specific to handle incomplete return types correctly, instead of bailing out completely.

If I am not mistaken, an inline override with unused arguments would need an include for them, as it is now:
```c++
class Override: public Base {
  void MyFunc(A& /*unused*/) override { // <-- need to include (forward-declare) A
  }
};
```

iwyu_output.cc
```diff
@@ -1529,6 +1529,17 @@ void IwyuFileInfo::CalculateIwyuViolations(vector<OneUse>* uses) {
   VERRS(6) << "--- Calculating IWYU violations for "
            << GetFilePath(file_) << " ---\n";
 
+  // Remove any uses part of an override decl -- they are covered for by the
+  // base class from where they were overridden, and the header for the base
+  // class must already be included because we derive from it.
+  for (OneUse& use : *uses) {
+    if (use.is_part_of_override_decl()) {
+      VERRS(5) << "Ignoring use of " << use.symbol_name() << " ("
+               << use.PrintableUseLoc() << "): re-exported from override\n";
```
jru: > re-exported from override

I find the term 'override' confusing. Depending on the context it might refer to the methods being overriden, or to the one that overrides them.

Why not 're-exported from base'?

iwyu.cc
kimgr: Good catch! It was necessary before for some half-broken removal of method bodies. Now that I do the is-definition check in the right place, it's no longer required. Thanks!

iwyu_ast_util.cc
kimgr: > I don't think this belongs here.

Quite right! And now that this method was only called from one place, I decided to inline it completely.

> If I am not mistaken, an inline override with unused arguments would need an include for
> them, as it is now:

That does look a little weird; maybe the special casing for definitions should only apply to the return type. I need to double-check what the compiler thinks about that one.

iwyu_output.cc
kimgr: I went with `re-exported from base class`

iwyu_ast_util.cc
```diff
@@ -270,12 +271,23 @@ bool IsNodeInsideCXXMethodBody(const ASTNode* ast_node) {
   return false;
 }
 
+static bool IsPartOfOverrideDecl(const ASTNode* ast_node) {
```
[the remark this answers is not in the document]
kimgr: > but that is surely not what you want.

That's exactly what I want. This function is not used from outside this source file, and I find the use of anonymous namespaces a terrible replacement for `static`. I've picked this up from LLVM's guidelines: https://llvm.org/docs/CodingStandards.html#static.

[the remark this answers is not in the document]
jru: Oh, my mistake. I thought I saw it declared in the header file.

iwyu_ast_util.cc
kimgr: I've spent the past few days trying to special case the return type of the function definition, but I haven't found anything that works well yet.

The challenge is that there's no obvious way to traverse the ASTNodes downwards after traversal. So phrasing the "ignore the return type of a user reported for CXXMethodDecl" logic is surprisingly hard.

I have some ideas for how to handle this better, but it requires larger-scale changes that we should be able to use for other "backward lookups", so I'd rather do that as a separate change altogether. 

OK to leave this so it requires a decl when unused args are part of the definition signature?

iwyu.cc
kimgr: I noticed while working on the improved handling of unused args that this was to suppress reporting of the return type in VisitFunctionDecl. It's also reported in VisitType. The new code semi-accidentally marks both cases ignored, but I'm starting to think it might be nicer to be explicit.

iwyu_ast_util.cc
jru: Yes, sure, I am fine with it. I thought it could be as simple as getting the return from the method decl and compare against the ast node, but I am likely missing something.

kimgr: I've tried 12 variations on that theme without getting to something that works reliably :)

kimgr:
Updated. I think this is even better:

* Definitions are now processed the same way, i.e. the _signature_ of an override is ignored, but any uses in the body are not
* Renamed tests
* Renamed and commented the most central function, `IsNodeInsideCXXMethodOverrideSignature`

If someone has time to take a look, I'd be delighted. Will probably merge this soonish as is if I don't hear anything.

jru:
Hey! I will take a look :)

iwyu.cc
```diff
@@ -2559,23 +2559,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-      fn_type = current_ast_node()->template GetParentAs<FunctionProtoType>();
-    }
-    if (!fn_type) {
-      if (const FunctionDecl* fn_decl
-          = current_ast_node()->template GetParentAs<FunctionDecl>())
-        fn_type = dyn_cast<FunctionProtoType>(GetTypeOf(fn_decl));
+    if (const auto* fn_decl = ast_node->GetAncestorOfType<FunctionDecl>()) {
+      fn_type = dyn_cast<FunctionProtoType>(GetTypeOf(fn_decl));
     }
+
```
jru, line 2567: It has been a while and not recall exactly, but could it be that there is only a `FunctionProtoType` without a corresponding `FunctionDecl`? Because in that case the new code won't find the prototype.

iwyu_ast_util.cc
```diff
@@ -270,12 +272,81 @@ bool IsNodeInsideCXXMethodBody(const ASTNode* ast_node) {
...
+  if (!method || !method->size_overridden_methods()) {
+    return false;
+  }
+
+  // Now we know this node is part of a C++ method signature overriding a base.
+
+  if (HasCovariantReturnType(method)) {
+    // Covariant return types require that the full return type is available,
+    // so don't special-case the return type for functions with covariance.
+  } else {
```
jru, line 313: I find this a bit confusing. Why not invert the condition and explain what we do for non-covariant return types?

iwyu_ast_util.cc
```diff
@@ -270,12 +272,81 @@ bool IsNodeInsideCXXMethodBody(const ASTNode* ast_node) {
...
+      return true;
+    }
+
+    if (ast_node->ContentIs(method->getReturnType().getTypePtr())) {
+      return true;
+    }
+
+    // The return type could be a type loc, though never seen this in practice.
+    const TypeSourceInfo* tsi = method->getTypeSourceInfo();
+    const TypeLoc retloc = tsi->getTypeLoc();
```
jru, line 328: Doesn't this refer to the type of the method itself?

tests/cxx/reexport_overridden.h
```diff
@@ -0,0 +1,37 @@
...
+  FwdRetType ReturnType() override;
+  FwdRetType ReturnTypeUnusedInBody() override;
+
+  // Unused argument available from Base.
+  void ArgumentUnused(const IndirectClass& x) override;
+
+  // Unused argument available from Base, in inline method.
+  void ArgumentUnusedInline(const IndirectClass& x) override {
+  }
+};
```
jru, line 31: Not sure if really needed, but it could be interesting to see that a method not overridden from the parent requires its types to be included directly.

mactalla:
I just tried out the branch and it works nicely.  Thanks @kimgr ! It doesn't seem to re-export the return type, but it does do the parameter types and that cuts down on quite a lot of noise form IWYU.

What's needed to get this PR merged?

kimgr:
@mactalla Thank you, good to hear! I've been experimenting with alternative solutions (lots of them!) offline, but still don't have anything I'm entirely happy with.

Maybe I should pick a favorite and get it merged 🙂.

I'm a little pressed for time right now, but I'll make an effort to cobble something together soon.

kimgr:
Just to make sure there's some update: I'm still working on/investigating solutions for this offline. It's a surprisingly tricky problem.

My current take is:

* Override `TraverseFunctionProtoTypeLoc` -- it has all the relevant information available (return type, parameter type, exception spec types, any associated `FunctionDecl` can be figured out)
* Traverse return type separately, taking IWYU's policy into account
* Traverse parameter types separately, taking IWYU's policy into account
* Traverse exception spec and noexcept expressions as normal

There's quite a bit of reporting machinery to get the `(for fn return type)` and `(for autocast)` comments in place, as well as maintaining forward-declare context, so unfortunately I haven't been able to make this agree with some of the template tests.

So I think there's some design shortcut for templates that needs to be resolved first, and then this might be easier.

I'll keep at it.

bolshakov-a:
@kimgr, you may take a look on [my variant](https://github.com/bolshakov-a/include-what-you-use/commit/c7ef7a57258469621498713c7bdfb85afee8867c). I block method declaration traversal except default arguments which may differ from declaration in base class (not sure about exception specifications). Covariant return types are handled in `VisitCXXMethodDecl` as before. Method definitions are traversed as before, hence return type is reported even if it isn't mentioned in the method body (your variant don't treat it properly, as your test [shows](https://github.com/include-what-you-use/include-what-you-use/pull/675/files#diff-4476e0e75321badf44d0ba46b72836581736bf2628b62859baf7d0774f9d8f28R38)).
The drawback is that not only `#include`s are pulled from .h into .cpp, but also forward declarations of unused parameters passed by reference or pointer. But I think that forward declarations could be suppressed in .cpp-files at all. You either need a full type, or you already should have (forward-)declaration from one of the headers included. Otherwise, you just have an unused garbage in your code, and it would be fine if compiler will point on it. I haven't succeeded in finding any counterexample.

bolshakov-a:
But, maybe, just adding base class headers into associated headers is better. It would also cover type reporting for public and protected members, which is also redundant for derived classes. On the other hand, private members are base class implementation details, and subclasses should not rely on presence their type info in base class header. Not sure, how acceptable it would be...

PR #676, opened 2019-04-04
Explicitly link to clangSerialization

tomKPZ:
This is necessary to fix the below linker error on Linux with clang 8.

/usr/bin/ld: CMakeFiles/include-what-you-use.dir/iwyu_driver.cc.o: undefined reference to symbol '_ZN5clang22PCHContainerOperationsC1Ev'

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,16 @@ target_link_libraries(include-what-you-use
   clangSema
   clangFrontend
   clangDriver
+
+  # Revision [1] in clang moved PCHContainerOperations from Frontend
+  # to Serialization, but this broke builds that set
+  # -DBUILD_SHARED_LIBS=on.  Revision [2] is a followup that works
+  # around the issue by adding an explicit dependency on Serialization
+  # wherever there was a dependency on Frontend.  Since we depend on
+  # Frontend, we need an explicit dependency on Serialization too.
+  # [1] https://llvm.org/viewvc/llvm-project?view=revision&revision=348907
+  # [2] https://llvm.org/viewvc/llvm-project?view=revision&revision=348915
+  clangSerialization
   )
 
 # Platform dependencies.


kimgr:
@tomKPZ Thank you! We've been through this patch before (https://github.com/include-what-you-use/include-what-you-use/pull/625), and I still think it's a workaround for a bug in LLVM/Clang's build system. If it should be fixed anywhere, it's there.

I suspect you're only seeing this with LLVM's build option `BUILD_SHARED_LIBS=True`, so try building Clang/LLVM without that, and you should be fine.

tomKPZ:
I'm using the system clang provided by my Linux distro.  Indeed they're building with BUILD_SHARED_LIBS=True https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/clang#n58 and they're obtaining the sources from the official release page https://releases.llvm.org/8.0.0/

In the bug you linked, you mentioned a series of followup commits that fixed the issue.  Would you happen to know which changes need to be cherry-picked?

kimgr:
I think the final conclusion was that the followup commits didn't really fix this problem. The one someone referred to was this: http://llvm.org/viewvc/llvm-project?view=revision&revision=348915

Oh, I see now that that commit message explains why CMake transitive dependencies somehow don't work with dylibs, and the commit is a workaround to explicitly add `clangSerialization` to the link list for all executables.

So I think that means we're at the mercy of a Clang bug they themselves have worked around. That tips the scales for me, and I'm willing to take this patch.

Could you make the commit message a little bit more explicit, e.g. `Explicitly link to clangSerialization` and add a comment referencing the link I found above on the line that adds clangSerialization to the link deps?

Thanks!

tomKPZ:
> Could you make the commit message a little bit more explicit, e.g. `Explicitly link to clangSerialization` and add a comment referencing the link I found above on the line that adds clangSerialization to the link deps?

Done!

kimgr:
Thank you, merged! And apologies to whoever suggested this last time when I said I didn't want to take that patch :-/

PR #677, opened 2019-04-04
[fix_includes] Simplify regexp

jwilk:


kimgr:
Thanks, looks good to me!

Issue #669, opened 2019-03-22
Generalize include file mapping to handle different targets.


trixirt:
Carry on from this pull request.
https://github.com/include-what-you-use/include-what-you-use/pull/665

Action plan from kimgr

-     Script to generate .h from .imp
-     Derive mappings from target triple
-     FreeBSD as a motivating example

I would like to start with the first bullet, and integrate that into the build process instead, so we don't keep mappings both in JSON and .h form in the repo (only the JSON form would be necessary). Then we can think about the two others.

Kudos for putting this together as a whole, though -- it gets the point across clearly how it would work!

I'm also a little worried about keying primarily on the target triple, I think we need to take into account which libc and C++ stdlib we are using. I'm not sure if the mappings you've spec-ed here are part of the operating system FreeBSD, or the libc that ships with FreeBSD. I think this is less of a problem for FreeBSD and Windows than Linux, where there seem to be multiple libc providers.

But we can worry about that when we have a nice infrastructure for generating headers from .imp files :)

-- 

Starting on the script.
Touchy part will be verifying nothing breaks.
I would like to commit the script and build infra and then the modified iwyu_include_picker so the map owners are not too surprised.

PR #680, opened 2019-04-10
Convert CreateCompilerInstance function to IwyuDriver class

trixirt:
Show use of compile state with compile specific stub method.
IncludePicker::AddTooChainMappings.  This method is where
triple specific libc mappings and libc++ vs libstdcpp mappings
will be done.

This is part 2 of issue 

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -104,6 +104,7 @@
 
 #include "iwyu_ast_util.h"
 #include "iwyu_cache.h"
+#include "iwyu_driver.h"
 #include "iwyu_globals.h"
 #include "iwyu_lexer_utils.h"
 #include "iwyu_location_util.h"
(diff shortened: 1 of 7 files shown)


kimgr:
This looks like it introduces significant ownership confusion and mutable global state. Can you say something about why this is necessary and what you wanted to achieve? There must be a simpler way to get there.

trixirt:
The target and c++ lib state need by the new target specific include mapper is held by the global compiler instance state.  The existing globals needed to be improved to make this visible to the lower levels. This could have been done by adding a few more global variables but it seemed like moving the effected variables into a class would be a better long term path.    A side effect of using the compiler instance late required some other global state to be saved in the IwyuDriver class. The other global were left as-is but should likely get a similar global-to-class makeover.

kimgr:
What would happen if, instead, you added a method to `IncludePicker` to add mappings after construction? It feels like there's probably a natural place in the setup to do this, where all the information is available.

I can take a stab at this when I have some time, but if you want to make a first attempt, please do.

Also, naming convention!

kimgr:
I'm pretty sure this is the wrong way to solve this problem. If/when I can find some time, I'll try to propose a simpler alternative as suggested above.

kimgr:
Closing this to clean up the list of open MRs. As mentioned earlier, I think there are better ways to do this. Will investigate at some point.

PR #671, opened 2019-03-25
Generate IncludeMapEntry tables from respective *.imp files.

trixirt:
iwyu_include_picker.cc's IncludeMapEntry tables duplicate in c form
several external imp file.  To remove this duplication, the c form
will be generated as part of the build process.  The generator is the
imp2array.py script.  It is called.

`imp2array.py -i <input-file> -o <output-file> -n <name-of-table>`

PR #686, opened 2019-05-04
FreeBSD target specific include mapping

trixirt:
Uses the imp to header generator
https://github.com/include-what-you-use/include-what-you-use/pull/671

Uses the target specific include mapping infra
https://github.com/include-what-you-use/include-what-you-use/pull/680

Adds FreeBSD as an example.

This should be all that is needed for this issue: 


trixirt:
Any comments ?

kimgr:
Sorry, spring-time is always a time of perfect work/life balance -- too much to do on both fronts :)

I'll circle back to this eventually. If you want to do prep in the meantime, please take some time to look over naming -- we should be using `lower_case_with_underscores`, not LLVM-style's `PascalCase` for variables.

Thanks!

kimgr:
I like how this patch builds on the other two, and does exactly one thing.

Just so I understand -- I'm an occasional FreeBSD user, so I should be able to test this for real at some point, but I'm also not intimately familiar with the platform -- are these libc headers for FreeBSD or the kernel headers?

It would be nice if we could be as specific as possible in the mapping names, regarding what they cover. For a "platform", I guess there are two-three mappings we care about: C stdlib, C++ stdlib and possibly kernel headers (?). Keeping these as separate mapping files should make it easier to (re-)generate. Does that make sense?

kimgr:
Closing this to clean up the list of open MRs. Platform-specific mappings are an interesting problem, but as mentioned above I'm not sure exactly how to key them. This branch will prove useful when I can circle back to it.

PR #687, opened 2019-05-07
Correct cmake command in README

gemfield:
Correct cmake command in README.md


kimgr:
I think the guide is correct as-is.

I typically place the build dir side by side with the source checkout, not inside it.

Are you sure you followed the instructions before your change? And have you tested them after? I'm not sure the whole procedure works after your update.

gemfield:
@kimgr  You are right.

PR #688, opened 2019-06-13
Don't crash if FileInfo cannot be found

ZachHenkel:
Discovered while running against my application's code.


ZachHenkel:
I'm skeptical that my change is responsible for the CI failure, is there a system issue?

iwyu.cc
```diff
@@ -3704,7 +3704,11 @@ class IwyuAstConsumer
     // though, because that will drag in every overload even if we're
     // only using one.  Instead, we keep track of the using decl and
     // mark it as touched when something actually uses it.
-    preprocessor_info().FileInfoFor(CurrentFileEntry())->AddUsingDecl(decl);
+    auto fileInfo = preprocessor_info().FileInfoFor(CurrentFileEntry());
```
kimgr: We typically don't use `auto` unless the type is already visible on the line.

Hacker-case for variables, so `fileInfo` -> `file_info`.

iwyu.cc
```diff
@@ -3704,7 +3704,11 @@ class IwyuAstConsumer
...
+
```
kimgr: This blank line seems unnecessary.

ZachHenkel:
I found a demonstration of the invalid source location by accident while working on a different issue.  Please see the AST output at https://godbolt.org/z/6MYLGr.  I'll make the requested changes including a test case and revise the PR very soon.

kimgr:
Thanks! But was that the right Godbolt link?

ZachHenkel:
I believe so, but it might only be correct because I still have it open in another tab :).  Here's a relevant portion of the AST dump, specifically the `<<invalid sloc>>`

`TranslationUnitDecl 0x563bfdd56e98 <<invalid sloc>> <invalid sloc>`
`|-TypedefDecl 0x563bfdd57770 <<invalid sloc>> <invalid sloc> implicit __int128_t '__int128'`
`| -BuiltinType 0x563bfdd57430 '__int128'`

kimgr:
Oh, I see what you're referring to now. That looks normal; builtin types usually don't have a source location, because they're, well, built into the compiler.

But if you have a using declaration refering to a builtin type, that might cause something like this.

kimgr:
@ZachHenkel I've fixed the CI breakdown on master, so please rebase.

kimgr:
@ZachHenkel Oh, and I meant to ask -- that Godbolt snippet does not cause trouble for IWYU, does it?

ZachHenkel:
@kimgr it looks like the issue is a using declaration in a PCH. 
main.cpp:

    int main()
    {
    A a;
    }

pch.h:

    struct A{};
    using ::A;

`clang++ -x c++-header pch.h -o pch.h.pch`
`include-what-you-use.exe -c -include pch.h main.cpp`

I'm uncertain if my proposed change is the correct fix for the crash.  Would you prefer that I close the PR and open an issue for further investigation?

ZachHenkel:
Extracting `struct A{};` into a file A.h that is #included in pch.h generates the following output:

	>include-what-you-use.exe -Xiwyu --verbose=3 -c -include pch.h main.cpp
	main.cpp:1:12: warning: A is defined in "A.h", which isn't directly #included.
	main.cpp:1:12: warning: A is defined in "pch.h", which isn't directly #included (for using decl).

	#include "A.h"    // for A
	#include "pch.h"  // for A

	#include "A.h"    // for A
	#include "pch.h"  // for A
	---

So the proposed fix doesn't inhibit later discovery of the using declaration in pch.h.  I'll rebase the PR and make the requested changes.

kimgr:
@ZachHenkel Ah, thanks! Is this true for any command-line-included (with `-include`) header, or only for PCHs?

kimgr:
It looks like IWYU skips over registering `-include`d files (see https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_preprocessor.cc#L479). I wonder if that should be fixed instead (and what the ramifications might be). I need to investigate this more closely.

In the meantime, could you clean up the branch's Git history so it's really based on master, and design a testcase for this? Thanks!

kimgr:
I just tested a variation on this, without precompiling pch.h, just adding it with an `-include` switch:
```
$ include-what-you-use -include pch.h main.cpp 
#include "pch.h"  // for A

#include "pch.h"  // for A
---
```

This is tricky. On the one hand we want to include-what-we-use, but the PCH is always special, and in this case it belongs in the buildsystem. 

It could be argued that it shouldn't contain any declarations, only other `#include`s, but your example shows that that doesn't work anyway.

(Aside: Back when I worked on a big C++ codebase, we settled on the convention to include-what-we-used everywhere, to make sure code was freestanding and dependencies explicit. Then we created PCHs with the most expensive and commonly used headers, and injected them using `-include` and `/FI` for MSVC. I think we saw indications that a header already contained in the PCH would be mostly free to `#include` again, so we opted for the double inclusion that this incurred.)

I wonder if IWYU should just never offer to add any includes that are already on the command-line? It won't work transitively, but that might be OK -- if you're `-include`-ing `X.h` for some reason, we can still suggest that `Y.h` (containing the actual type you're using) should be included.

kimgr:
Also, I wonder if we can short-circuit PCH use in IWYU overall. If it didn't search for the `pch.h.pch` file and attempt to load the binary format, everything would be more stable.

ZachHenkel:
How do you want to add the pch creation step to the testing framework?  It looks like a valid clang install isn't a requirement yet.

kimgr:
@ZachHenkel I'd really rather not require Clang (or even producing temporary output files like a .pch). I think it's problematic enough that we should just forget about the testcase if it's required. Instead, add a big fat comment that `file_info` can be `nullptr` when part of a PCH.

Actually, I think we could strengthen that even more -- if there's a PCH on the command-line, `PreprocessorOptions::ImplicitPCHInclude` will have a value, so you could do something like:

```c++
    IwyuFileInfo* file_info = preprocessor_info().FileInfoFor(CurrentFileEntry());
    if (file_info) {
      // ...
    } else {
      // Explanatory comment here...
      CHECK_(!compiler()->getInvocation().getPreprocessorOpts().ImplicitPCHInclude.empty());
    }
```

(untested). I think that will provide us with actionable information if we somehow get a null pointer for some other reason.

ZachHenkel:
Thanks @kimgr I'll give your suggestion a try against my application's code to make sure it meets our needs and update this thread with the results.

On another point that I missed earlier, not loading the binary pch would be problematic for our application.  On msvc the `#include <pch>` statement is used more like a marker than an actual include. 
 The compiler doesn't even attempt to find the file named "pch" on disk.  This behavior is retained in clang's msvc compatibility mode. We learned this lesson the hard way when first exploring clang static analysis.

kimgr:
@ZachHenkel 
> On another point that I missed earlier, not loading the binary pch would be problematic for our
> application. On msvc the #include <pch> statement is used more like a marker than an actual
> include.
> The compiler doesn't even attempt to find the file named "pch" on disk. This behavior is retained in 
> clang's msvc compatibility mode.

Yeah, but the binary pch is generated from the file in the `#include` directive anyway, right? So pulling in the binary PCH or including the source file it's generated from should have the same logical result from IWYU's point of view (even if the mechanics are different). Well, unless that source file is missing, but that seems like a strange setup. :)

ZachHenkel:
I hate to admit it, but we have quite a few cases where the pch file isn't discoverable by the compiler.  With the size project that I work on, fixing all the instances is a non-trivial effort.

kimgr:
@ZachHenkel Ah, I see! And I realized that from IWYU's point of view, the PCH is likely not at all the same as a normal include, since IWYU won't see the PCH source or the headers included from it.

I looked into it, but I don't see a way of avoiding PCH loading, so you're still safe in that regard :)

See also IWYU's `--pch_in_code` switch, which forces IWYU to keep the first included header, assuming it's the pch marker.

ZachHenkel:
The test against one of our most complicated projects succeeded.  I'm content that the fix will hold throughout our codebase.

iwyu.cc
```diff
@@ -3708,7 +3709,19 @@ class IwyuAstConsumer
     // though, because that will drag in every overload even if we're
     // only using one.  Instead, we keep track of the using decl and
     // mark it as touched when something actually uses it.
-    preprocessor_info().FileInfoFor(CurrentFileEntry())->AddUsingDecl(decl);
+    IwyuFileInfo* file_info =
+        preprocessor_info().FileInfoFor(CurrentFileEntry());
+    if (file_info) {
+      file_info->AddUsingDecl(decl);
+    } else {
+      // The precomiled header file will not (yet) be found in the
```
kimgr: The formatting is a little weird here, and the comment could be more to the point. I suggest:

```diff
@@ -3708,7 +3708,19 @@ class IwyuAstConsumer
     // though, because that will drag in every overload even if we're
     // only using one.  Instead, we keep track of the using decl and
     // mark it as touched when something actually uses it.
-    preprocessor_info().FileInfoFor(CurrentFileEntry())->AddUsingDecl(decl);
+    IwyuFileInfo* file_info =
+        preprocessor_info().FileInfoFor(CurrentFileEntry());
+    if (file_info) {
+      file_info->AddUsingDecl(decl);
+    } else {
+      // For using declarations in a PCH, the preprocessor won't have any
+      // location information. As far as we know, that's the only time the
+      // file-info will be null, so assert that we have a PCH on the
+      // command-line.
+      const string& pch_include =
+          compiler()->getInvocation().getPreprocessorOpts().ImplicitPCHInclude;
+      CHECK_(!pch_include.empty());
+    }
 
     if (CanIgnoreCurrentASTNode())  return true;
```

ZachHenkel:
Thanks I've updated to use the clearer wording/check.

kimgr:
Thanks! I'll squash and merge this in my local repo.

kimgr:
Merged as: https://github.com/include-what-you-use/include-what-you-use/commit/91c0b55f7f11c2852a0be75f1a598a72d26ab0a7.

Thank you!

PR #692, opened 2019-06-26
Travis build fixup

kimgr:
Our Travis build has been failing lately.

This is due to a bug in the Debian packaging of LLVM (see the `Horrible patchwork...` commit) that I've managed to work around.

While here, I also decided to bump the base dist of the build environment from the outdated Ubuntu Trusty (14.04) to Xenial, which enabled removal of some older workarounds.


kimgr:
I'm merging this to get the tree back to a state we can validate.

PR #696, opened 2019-06-30
Remove ".." from normalized file paths

kimgr:
This fixes a case where IWYU would suggest include paths with ".." where
a simpler non-".." path would suffice. E.g.

    // a.h
    struct Symbol {};

    // foo/b.h
    #include "../a.h"

    // main.c
    #include "foo/b.h"

    Symbol s;

IWYU would suggest we should include "foo/../a.h" from main.c here. This
is correct, but inelegant.

After this patch, we'll suggest "a.h" directly. Unfortunately, this is
broken in the face of symlinks (example cut from bug report) --

    akuli@Akuli-Desktop:~$ ln -s /usr a
    akuli@Akuli-Desktop:~$ file a
    a: symbolic link to /usr
    akuli@Akuli-Desktop:~$ file lib
    lib: cannot open `lib' (No such file or directory)
    akuli@Akuli-Desktop:~$ file a/../lib
    a/../lib: directory
    akuli@Akuli-Desktop:~$

-- but I suspect this is more unusual than relative includes.

Fixes issue 690.

(Please ignore the first few commits, they're part of another PR to rename `subfolder` to `subdir`)

### DIFF
--- a/iwyu_path_util.cc
+++ b/iwyu_path_util.cc
@@ -127,7 +127,7 @@ string GetCanonicalName(string file_path) {
 
 string NormalizeFilePath(const string& path) {
   llvm::SmallString<128> normalized(path);
-  llvm::sys::path::remove_dots(normalized);
+  llvm::sys::path::remove_dots(normalized, /*remove_dot_dot=*/true);
 
 #ifdef _WIN32
   // Canonicalize directory separators (forward slashes considered canonical.)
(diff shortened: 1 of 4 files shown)


Akuli:
Where is the changed code? I only see tests in the diff

kimgr:
See iwyu_path_util.cc

PR #704, opened 2019-07-11
Add license headers to Python scripts

kimgr:
Some of the Python scripts (both for development and production) had
fallen behind.


kimgr:
Abandon this, I'll let the script from #706 do the work once it's done.

PR #705, opened 2019-07-12
Fix wrong suggestions with --no_fwd_decls

### LINKED ISSUE #702: Wrong forward declaration removal suggestion with --no_fwd_decls
In this simple test case:
```
namespace NS {
  class Foo;

  template<class T>
  struct FWDOnly {
    T *t;
  };

  struct Bar {
    FWDOnly<Foo> ptr;
  };
}
```

IWYU with `--no_fwd_decls` suggests to remove the forward declaration of class Foo, however this causes a compilation error. This does not happen without the --no_fwd_decls option.

fabledpig:
This fixes 2 specific issues only happening with `--no_fwd_decls`:

* Removal of needed forward declaration from an associated header, when the
definition is in the source file. (Fixes #702)

* Replacing an include in an associated header file, containing a forward
declaration, with a local forward declaration, if the definition can be found in
the header.

Example:
```
// header.hpp
#pragma once

class Foo;

// test.hpp
#pragma once

#include "header.hpp"

Foo *foo;

class Foo {};

// test.cpp
#include "test.hpp"
```
IWYU suggests to replace `#include header.hpp` with `class Foo`. However, I think the correct suggestion (with --no_fwd_decls) would be to keep the include.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1105,13 +1105,29 @@ void ProcessForwardDeclare(OneUse* use,
     }
   }
 
-  // (A7) If --no_fwd_decls has been passed, recategorize as a full use unless
-  // the decl is in this file (in which case it must be a self-sufficient decl
-  // being used, so we can just let IWYU do its work).
-  if (!use->ignore_use() &&
-      GlobalFlags().no_fwd_decls &&
-      GetFileEntry(use->decl_loc()) != GetFileEntry(use->use_loc())) {
-    use->set_full_use();
+  // (A7) If --no_fwd_decls has been passed, and a decl can be found in one of
+  // the headers, suggest that header, and recategorize as a full use. If we can
+  // only find a decl in this file, it must be a self-sufficent decl being used,
+  // so we can just let IWYU do its work, and there is no need to recategorize.
+  if (!use->ignore_use() && GlobalFlags().no_fwd_decls) {
+    bool promote_to_full_use = true;
+    for (const Decl* decl = use->decl(); decl != nullptr;
+         decl = decl->getPreviousDecl()) {
+      if (IsBeforeInSameFile(decl->getLocation(), use->use_loc())) {
+        promote_to_full_use = false;
+      } else if (IsBeforeInTranslationUnit(decl->getLocation(),
+                                           use->use_loc())) {
+        // TODO: Choose a redecl that is already provided by a desired include, so we
+        // don't keep another include that is not necessary.
+        use->reset_decl(cast<NamedDecl>(decl));
+        promote_to_full_use = true;
+        break;
+      }
+    }
+
+    if (promote_to_full_use) {
+      use->set_full_use();
+    }
   }
 }
 
--- a/tests/cxx/no_fwd_decls.cc
+++ b/tests/cxx/no_fwd_decls.cc
@@ -29,6 +29,10 @@ IndirectClass* global;
 class LocalFwd;
 void ForwardDeclareUse(const LocalFwd*);
 
+// IWYU must not remove the forward declaration of this class even though its
+// definition can be found in the same file but after the use
+class LocalFwd {};
+
 // A forward-declare that also exists in an included header can be removed.
 // Normally IWYU would optimize for fewer includes, but in --no_fwd_decls mode
 // we optimize for fewer redeclarations instead.


fabledpig:
This still needs some test cases, but please have a look at it.

kimgr:
This looks good to me, conceptually. There are some details I'd like to poke at in the patch, but if you could extend the testing in `no_fwd_decls.cc` with these cases, I can review the whole thing. Thanks!

fabledpig:
Okay. I just added the test cases.

iwyu_output.cc
```diff
@@ -1106,11 +1106,27 @@ void ProcessForwardDeclare(OneUse* use,
...
-  if (!use->ignore_use() &&
-      GlobalFlags().no_fwd_decls &&
-      GetFileEntry(use->decl_loc()) != GetFileEntry(use->use_loc())) {
+  // there is a decl in this file before the use (in which case it must be a
+  // self-sufficient decl being used, so we can just let IWYU do its work).
+  if (!use->ignore_use() && GlobalFlags().no_fwd_decls) {
+    for (const Decl* curr_decl = use->decl(); curr_decl != nullptr;
+         curr_decl = curr_decl->getPreviousDecl()) {
+      if (IsBeforeInSameFile(curr_decl->getLocation(), use->use_loc()))
+        return;
```
kimgr: I'd like to avoid an early return here, because it's not unthinkable that we would like to add additional policies after (A7). Let's see if I can figure out a way to factor that...

```c++
bool promote_to_full_use = true;
for (const Decl* decl = use->decl(); decl != nullptr;
     decl = decl->getPreviousDecl()) {
  // Rationale comment here...
  if (IsBeforeInSameFile(decl->getLocation(), use->use_loc())) {
    promote_to_full_use = false;
    break;
  }

  // Rationale comment here...
  if (IsBeforeInTranslationUnit(decl->getLocation(), use->use_loc())) {
    use->reset_decl(cast<NamedDecl>(decl));
    use->set_suggested_header(
        ConvertToQuotedInclude(GetFilePath(decl)));
    break;
  }
}

if (promote_to_full_use) {
  use->set_full_use();
}
```

I think you can rationalize each of the cases in detail in the loop, and make the `// (A7)` comment a little more high-level.

iwyu_output.cc
```diff
@@ -1106,11 +1106,27 @@ void ProcessForwardDeclare(OneUse* use,
...
+
+      // If a decl can be found in the translation unit before the use but not in
+      // this file, we set that better decl as the use's decl and suggest
+      // the header where it was found (and finally recategorize it as a full
+      // use).
+      if (IsBeforeInTranslationUnit(curr_decl->getLocation(), use->use_loc())) {
+        const NamedDecl* better_decl = DynCastFrom(curr_decl);
```
kimgr: At this point, I think we're guaranteed that `curr_decl` is a `NamedDecl`, so just an unconditional cast seems better. And if we do that, we don't need `better_decl` for anything: `use->reset_decl(cast<NamedDecl>(curr_decl));`

iwyu_output.cc
```diff
@@ -1106,11 +1106,27 @@ void ProcessForwardDeclare(OneUse* use,
...
+        use->reset_decl(better_decl);
+        use->set_suggested_header(
+            ConvertToQuotedInclude(GetFilePath(curr_decl->getLocation())));
```
kimgr: There's a `GetFilePath` template that should be able to figure out the path from the Decl directly -- `GetFilePath(curr_decl)`

tests/cxx/no_fwd_decls.cc
```diff
@@ -41,6 +43,13 @@ bool AreSame(const NameOnly& a, const NameOnly& b) {
   return AddressOf(a) == AddressOf(b);
 }
 
+// ClassInSrc is forwad declared in no_fwd_decls-fwd2.h, so IWYU must keep that
```
kimgr: Typo: forwad

tests/cxx/no_fwd_decls.h
```diff
@@ -0,0 +1,22 @@
+//===--- no_fwd_decls.h - test input file for iwyu -----------------------===//
```
kimgr: Pad line with `-` to 80 chars

tests/cxx/no_fwd_decls-fwd2.h
```diff
@@ -0,0 +1,16 @@
+//===--- no_fwd_decls-fwd2.h - test input file for iwyu --------------------===//
```
kimgr: Overlong line

kimgr: I'm not entirely happy with the filename -- it would be nice if it reflected some quality of what it contains.

Something like `no_fwd_decls-order.h`, to indicate that it provides symbols for testing the order of declarations?

Same with the forward decls, I think the names could carry more weight: `class FwdDefinedAfter;` and `class FwdInHeader;`. I'm not 100% happy with these either, but I find they convey a little more meaning (assuming I understand what their purpose is :))

tests/cxx/no_fwd_decls.cc
```diff
@@ -41,6 +43,13 @@ bool AreSame(const NameOnly& a, const NameOnly& b) {
...
+// include even though the definition can be found in this source file, but
+// after the use.
+void ForwardDeclareUse(const ClassInSrc&);
+
+class ClassInSrc {};
```
kimgr: Actually, instead of introducing a new class and function, couldn't you just define `Fwd` below the first `ForwardDeclareUse`, and add a comment to explain why?

tests/cxx/no_fwd_decls.h
```diff
@@ -0,0 +1,22 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU should not replace the include with a local forward declaration
+#include "tests/cxx/no_fwd_decls-fwd2.h"
+
+// ClassInHeader is in no_fwd_decls-fwd2.h
+ClassInHeader *fwd;
```
kimgr: `*` binds to the left in IWYU style

kimgr: I don't understand how this test case is different from the `ClassInSrc` one? Include a header containing a forward-decl, use the forward-decl in name only, and then provide a definition of the decl?

There are some subtle rules around analysis of associated headers (in this case `no_fwd_decls.{h,cc}` are associated), they "share" more includes than what is typical, so the results could actually be skewed by this (since the `no_fwd_decls-fwd2.h` header is included both here and in the `.cc` file.

Or did you intend for this to be an associated header for some reason I'm missing?

iwyu_output.cc
kimgr: So if there's a longer chain of redeclarations, this will pick the first one outside of this file? This seems a little arbitrary, but I guess it kind-of makes sense.

I'm just worried there might be arcane cases where there are multiple headers containing forward-declarations for a symbol, and we choose the wrong one at random. But maybe mappings will figure that out after the fact, in that case.

iwyu_output.cc
fabledpig: In the changes I did the comments in a different way. I think it is easier to explain what's happening this way. Let me know if you are satisfied with it.

iwyu_output.cc
```diff
@@ -1105,13 +1105,28 @@ void ProcessForwardDeclare(OneUse* use,
...
+  if (!use->ignore_use() && GlobalFlags().no_fwd_decls) {
+    bool promote_to_full_use = true;
+    for (const Decl* decl = use->decl(); decl != nullptr;
+         decl = decl->getPreviousDecl()) {
+      if (IsBeforeInSameFile(decl->getLocation(), use->use_loc())) {
+        promote_to_full_use = false;
+      } else if (IsBeforeInTranslationUnit(decl->getLocation(),
+                                           use->use_loc())) {
+        use->reset_decl(cast<NamedDecl>(decl));
+        use->set_suggested_header(ConvertToQuotedInclude(GetFilePath(decl)));
```
kimgr: I suspect this would need to be run through the mapping machinery again -- for example, if symbol `X` is declared in private header `x_internal.h`, and documented to be used from `x.h`, we'll rewrite whatever the earlier analysis came up with to `"x_internal.h"`, because that's where the decl is.

I'm not sure how/when/where mapping happens now (I really should), so I don't know if there's a simple way to just remap to the public header. If you don't find an obvious way, I think it would be fine to leave a `TODO` here.

tests/cxx/no_fwd_decls.cc
```diff
@@ -29,6 +29,10 @@ IndirectClass* global;
 class LocalFwd;
 void ForwardDeclareUse(const LocalFwd*);
 
+// IWYU must not remove the forward declaration of this class even though its
+// definition can be found in the same file but after the use.
+class LocalFwd {};
+
```
kimgr: Does the existing test below with `Fwd` and `ForwardDeclareUse` cover the new "find-first-decl-not-in-this-file" logic, or does it need a separate test case? It seems to me this should cover it.

fabledpig: You are right, it is not necessary.

iwyu_output.cc
fabledpig: I think this can be solved by not setting a suggested header, so the header will be chosen in `CalculateMinimalIncludes`.

However there is another problem (they are maybe related, I'm not sure):
```
// foo.h
class Foo;

// fooandbar.h
class Foo;
class Bar;

// test.cpp
#include "fooandbar.h"
#include "foo.h"

Foo *foo;
Bar *bar;
```
Here IWYU will say that the includes are correct, but in fact `foo.h` could be removed. (If you change the order of the includes IWYU will correctly suggest the removal.)
I left a `TODO` for this one.

iwyu_output.cc
```diff
@@ -1105,13 +1105,29 @@ void ProcessForwardDeclare(OneUse* use,
...
+  // so we can just let IWYU do its work, and there is no need to recategorize.
+  if (!use->ignore_use() && GlobalFlags().no_fwd_decls) {
+    bool promote_to_full_use = true;
+    for (const Decl* decl = use->decl(); decl != nullptr;
+         decl = decl->getPreviousDecl()) {
+      if (IsBeforeInSameFile(decl->getLocation(), use->use_loc())) {
+        promote_to_full_use = false;
+      } else if (IsBeforeInTranslationUnit(decl->getLocation(),
+                                           use->use_loc())) {
+        // TODO: Choose a redecl that is already provided by an include, so we
```
kimgr: "provided by an include" -> "provided by a desired include"
"an other" -> "another"

iwyu_output.cc
kimgr: Interesting. This keeps coming up (see the tolerate-transitive PR), where decisions made at one point are invalidated by later steps. But I think a TODO is fine for this. If we manually remove `foo.h`, IWYU won't complain, so that's cool.

iwyu_output.cc
fabledpig: Yeah, I think I misunderstood you. That test case was indeed necessary and now it's there again.

And yes, the other test case should cover the case you're talking about.

kimgr:
Thank you, this looks great!

PR #706, opened 2019-07-12
Add script to check license headers

kimgr:
License headers in IWYU have a pretty strict general form:

```
 ##===--- filename - description -----------------------------------===##
 #
 #                     The LLVM Compiler Infrastructure
 #
 # This file is distributed under the University of Illinois Open Source
 # License. See LICENSE.TXT for details.
 #
 ##===--------------------------------------------------------------===##
```
(with 80 chars line width for the padded lines, and comment chars
depending on language).

This script recognizes C++ and Python and checks that the license header
lives up to the requirements. There are currently 18 files in violation
of this convention.

I'm planning to wire this into the CI infrastructure once all license
headers are fixed up, to get automated checks instead of relying on code
review for these.

### DIFF
--- a/iwyu-check-license-header.py
+++ b/iwyu-check-license-header.py
@@ -1,0 +1,277 @@
+#!/usr/bin/env python
+
+##===--- iwyu-check-license-header.py - check license headers -------------===##
+#
+#                     The LLVM Compiler Infrastructure
+#
+# This file is distributed under the University of Illinois Open Source
+# License. See LICENSE.TXT for details.
+#
+##===----------------------------------------------------------------------===##
+
+from __future__ import print_function
+import sys
+import os
+import re
+import argparse
+
+
+# This is used in selected functions to calculate the maximum length of filename
+# and filler dashes. Not otherwise useful.
+EMBELLISHMENTS = '//===---  ---===//'
+
+HDRFORMAT = """
+{c}
+{c}                     The LLVM Compiler Infrastructure
+{c}
+{c} This file is distributed under the University of Illinois Open Source
+{c} License. See LICENSE.TXT for details.
+{c}
+{t}===----------------------------------------------------------------------==={t}
+"""
+
+
+def make_hdrformat(one, two):
+    """ Materialize HDRFORMAT based on the one and two comment styles.
+
+    The one and two args should be '#' and '##' or '//' and '//', respectively.
+    Returns a list of lines.
+    """
+    r = HDRFORMAT.lstrip().format(c=one, t=two)
+    return r.splitlines()
+
+
+def truncated(filename):
+    """ Truncate the filename with ellipsis if too long """
+    maxlen = 80 - len(EMBELLISHMENTS)
+    trunclen = maxlen - 3  # ...
+    if len(filename) > maxlen:
+        filename = filename[:trunclen] + '...'
+    return filename
+
+
+def make_license_header(filename, one, two):
+    """ Build a valid license header from filename and comment styles.
+
+    The one and two args should be '#' and '##' or '//' and '//', respectively.
+    Returns a list of lines.
+    """
+    assert len(two) == 2
+    filename = os.path.basename(filename)
+    filename = truncated(filename)
+
+    def dashes():
+        c = 80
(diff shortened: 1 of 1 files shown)


jru:
This is an awesome idea! 

It could only be beaten by a tool that would additionally take a `-fix` argument, à la _clang-tidy_, to automatically add the license text :)

kimgr:
@jru Challenge accepted! 🙂

kimgr:
That took me a little longer than I had hoped, but please see the new rewritten version with a `--add` switch.

iwyu-check-license-header.py
```diff
@@ -0,0 +1,266 @@
...
+    def has_license_header(self):
+        """ Return True if a license header has been found """
+        return self.hdrindex != -1
+
+    def add_license_header(self):
+        """ Add license header to a file that doesn't have one. """
+        assert not self.has_license_header()
+
+        # Find insertion point
+        for p, line in enumerate(self.lines):
```
jru: This construct is used twice. I think a function that returns the expected line where the license should start would be clearer. 
e.g.
```py
p = find_expected_license_start_line(lines) # -1 < p <= len(lines)
```

iwyu-check-license-header.py
```diff
@@ -0,0 +1,266 @@
...
+    def dashes():
+        c = 80
+        c -= 2  # take off leading ##
+        c -= 6  # take off leading ===---
+        c -= len(filename) + 2  # take off filename with a space on either side
+        c -= 3  # take off trailing ===
+        c -= 2  # take off trailing ##
+        return '-' * c
+
+    firstline = '%s===--- %s %s===%s' % (two, filename, dashes(), two)
```
jru: If the filename is longer than 65 chars (80-2-6-2-3-2) this will fail.
Maybe better to calc the dashes count and truncate the filename by that amount in that case?

iwyu-check-license-header.py
```diff
@@ -0,0 +1,266 @@
...
+        """ Check that the header lines follow convention.
+
+        Returns True if everything is OK, otherwise returns False and populates
+        self.errors with all found errors.
+        """
+        if not self.has_license_header():
+            self.file_error('No license header found')
+            return False
+
+        hdrlines = self.lines[self.hdrindex:self.hdrindex+8]
```
jru: Not likely to change but the magic number can be easily calculated.

iwyu-check-license-header.py
```diff
@@ -0,0 +1,266 @@
...
+            self.line_error(1, 'Bad header line',
+                            " Expected: '%s'" % self.pattern.pattern,
+                            " Actual:   '%s'" % line)
+        else:
+            hfilename = m.group(1)
+            xfilename = os.path.basename(self.filename)
+            if hfilename != xfilename:
+                self.line_error(1, 'Bad header filename',
+                                " Expected:  '%s'" % xfilename,
+                                " Actual:    '%s'" % hfilename)
```
jru: This will fail for truncated filenames (longer than 65)

iwyu-check-license-header.py
kimgr: I've toyed with this, but I have yet to find a meaningful name for that function that's as clear as the simple open-coded implementation.

In the one place, it's used to detect if there is a license header, and in the other to figure out where to put it, so the semantics are subtly different.

iwyu-check-license-header.py
kimgr: I find Python slicing with variables becomes very hard to read, so I'd prefer to keep the magic.

iwyu-check-license-header.py
kimgr: Fixed

iwyu-check-license-header.py
kimgr: Fixed.

iwyu-check-license-header.py
kimgr: I left this unchanged. Not too worried about it.

kimgr:
This looks good to myself. And I want to get it onto master to get the CI step going to support existing PRs, so merging. Any further feedback can be handled as follow-ups.

PR #707, opened 2019-07-18
Add missing handling of ElaboratedTypes in VisitMemberExpr

fabledpig:
IWYU missed the use of typedefs, if they were 'hidden' behind elaborated types.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2167,6 +2167,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         = expr->isArrow() ? RemovePointerFromType(base_type) : base_type;
     if (CanIgnoreType(base_type) && CanIgnoreType(deref_base_type))
       return true;
+    deref_base_type = RemoveElaboration(deref_base_type);
     if (const TypedefType* typedef_type = DynCastFrom(deref_base_type)) {
       deref_base_type = DesugarDependentTypedef(typedef_type);
     }
--- a/tests/cxx/member_expr-elaborated.h
+++ b/tests/cxx/member_expr-elaborated.h
@@ -1,0 +1,13 @@
+//===--- member_expr-elaborated.h - test input file for iwyu --------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+template<class T>
+struct Typedef {
+  typedef T type;
+};
--- a/tests/cxx/member_expr.cc
+++ b/tests/cxx/member_expr.cc
@@ -11,6 +11,8 @@
 
 #include "tests/cxx/member_expr-d1.h"
 #include "tests/cxx/direct.h"
+#include "tests/cxx/member_expr-elaborated.h"
+#include "tests/cxx/struct-with-member.h"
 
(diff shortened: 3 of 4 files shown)


kimgr:
Thanks, this looks good!

I'm in the middle of formulating some guidelines for tests, because I suspect I've been doing it wrong for a while. I'll come back with some concrete feedback once I've thought it through properly.

kimgr:
I think there's a bit more to this than just elaborated types -- I've been experimenting with a reduced testcase on `master`, and there needs to be a template involved for this to be mischaracterized.

Assuming I'm not missing something, this is the smallest testcase I can come up with:
```
#include "tests/cxx/direct.h"

template<class T>
struct Wrapper {
  typedef T Type;
};

int foo() {
  return Wrapper<IndirectClass>::Type().a;
}
```

This suggests a forward declaration for `IndirectClass` instead of a full use.

So I suggest:
* Put the template in the main .cc file
* Use `IndirectClass` instead of creating a new type and header
* Use `direct.h` to force IWYU to generate diagnostics, and use the `  // IWYU: IndirectClass is...*indirect.h` directives to check that IWYU says the right thing at the right place

And it would be nice to clarify in the commit message what's going on -- there's something in the combination of templates and elaboration that requires this, but I can't quite make out what.

Incidentally, this also requires a forward-declaration instead of a full use:
```
// Now without the member expr
Wrapper<IndirectClass>::Type x;
```

So I wonder if there's some other place that needs the same fix (or if this fix is in the wrong place).

fabledpig:
I think the root of the problem is that in `GetCallerResponsibleTypesForTypedef` we don't cover the template arguments, so this is a deeper issue than it first looked, and is also present in GetCallerResponsibeTypesForFnReturn (probably present elsewhere too but I only looked into these two).

kimgr:
> I'm in the middle of formulating some guidelines for tests

Here's a draft: https://github.com/kimgr/include-what-you-use/blob/style-docs/docs/IWYUStyle.md#regression-tests

Hopefully it can provide some help and rationale for improved test design. It's up for review in PR #708 if you want to leave feedback.

Thanks!

kimgr:
@fabledpig How's it going with this? I don't remember the details, but it sounds like you had doubts about the solution?

fabledpig:
Yeah... I don't think this is the correct solution, this issue goes deeper. The problem is that template arguments are not taken into account in a lot of cases. 

But sadly I don't have time nowadays to contribute to this project. Maybe in the near future. :)

kimgr:
Thanks for the update, I'll leave this open in case I find time and inspiration to keep digging.

jbytheway:
Agreed that the root problem is the use of aliases of template parameters.  Here's a different minimal-ish example of the issue:

```c++
#include "direct.h"

template<typename T>
struct A {
  using TAlias = T;
  TAlias t;
};

A<IndirectClass> a;
```

Again, it suggests a forward declaration rather than a `#include`.

This affects standard library classes.  For example, gcc's `std::unordered_map` uses aliases and IWYU doesn't register the arguments as having full-uses.

This is the most common correctness issue I encounter with IWYU at the moment, so I will try looking into it, but I suspect it's a tricky thing, so we shall see what happens.

kimgr:
I haven't seen any activity on this for a while, so I'm going to close it. Feel free to reopen if there's new development, otherwise I'll focus on #775.

PR #712, opened 2019-09-13
iwyu_tool: use directory information from compilation database if pos…

vmiklos:
…sible

If the entry looks like this:

    {
        "arguments": [
            ....
        ],
        "directory": "/path/to/test",
        "file": "Test.cpp"
    },

And we invoked iwyu_tool.py in /path/to, then the constructed abs path was
/path/to/Test.cpp, not /path/to/test/Test.cpp, fix this.

If the change itself makes sense, I'm happy to look at covering this from `iwyu_tool_test.py`. (The above is an autotools project, compilation database generated by bear.)

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -298,6 +298,10 @@ class Invocation(object):
 def fixup_compilation_db(compilation_db):
     """ Canonicalize paths in JSON compilation database. """
     for entry in compilation_db:
+        # Convert relative paths to absolute ones if possible, based on the entry's directory.
+        if 'directory' in entry and not os.path.isabs(entry['file']):
+            entry['file'] = os.path.join(entry['directory'], entry['file'])
+
         # Expand relative paths and symlinks
         entry['file'] = os.path.realpath(entry['file'])
 
--- a/iwyu_tool_test.py
+++ b/iwyu_tool_test.py
@@ -281,7 +281,6 @@ class CompilationDBTests(unittest.TestCase):
         """ Compilation database path canonicalization. """
         compilation_db = [
             {
-                "directory": "/home/user/llvm/build/test",
                 "file": "Test.cpp"
             }
         ]
@@ -291,6 +290,21 @@ class CompilationDBTests(unittest.TestCase):
         # Check that file path is made absolute.
         entry = canonical[0]
         self.assertEqual(os.path.join(self.cwd, 'Test.cpp'), entry['file'])
+
+    def test_fixup_from_entry_dir(self):
+        """ Compilation database abs path is based on an entry's directory. """
+        compilation_db = [
+            {
+                "directory": "/home/user/foobar",
+                "file": "Test.cpp"
+            }
+        ]
+
+        canonical = iwyu_tool.fixup_compilation_db(compilation_db)
+
+        # Check that the file path is relative to the directory entry, not to the current directory.
+        entry = canonical[0]
+        self.assertEqual('/home/user/foobar/Test.cpp', entry['file'])
 
 
 if __name__ == '__main__':


iwyu_tool.py
```diff
@@ -308,6 +308,10 @@ def parse_compilation_db(compilation_db_path):
 
     # Expand symlinks.
     for entry in compilation_db:
+        # Convert relative paths to absolute ones if possible.
+        if 'directory' in entry.keys() and not os.path.isabs(entry['file']):
```
kimgr: Can remove `.keys()`, that's the default for `in` on `dict`s

vmiklos: I've updated the commit message, removed .keys() and also added a test.

I went for an on-disk compliation database, hope that's OK. The alternative would be mocking __builtins__.open, but this is python2, so I can't use python3 unittest.mock, and doing that manually sounds boring. :-)

iwyu_tool_test.py
```diff
@@ -97,6 +97,11 @@ def test_from_compile_command(self):
             [iwyu_tool.IWYU_EXECUTABLE, '-foo', '-Iinclude', 'file.cc'])
         self.assertEqual(invocation.cwd, '/home/user/llvm/build')
 
+    def test_compdb_dir(self):
+        compilation_db = iwyu_tool.parse_compilation_db('tests/tool/compdb-dir')
```
kimgr: I'm not a fan of test data files. It would be nice if we could make `parse_compilation_db` accept a file-object or a path. Or, since the function appears to do two thing (reading from file and canonicalizing), maybe move the file reading into `main` and add a `def fixup_compilation_db(compilation_db)` that does both this directory fixup and the realpath fixup. I chose `fixup` over `canonicalize_...` because the latter is too long :)

kimgr: I'm a little short on time, but let me know if you want me to do the refactor and push to master so you can rebase onto something that's simpler to test.

vmiklos: If you could do that, that would be great. :-) Failing that, I probably will have time to do this the next weekend.

kimgr: I had something like this in mind: https://github.com/include-what-you-use/include-what-you-use/pull/718

vmiklos: Thanks. I've subscribed to that, so I can rebase this on top of a new master once it's in.

kimgr: Should be in place now, see if you can retrofit this on top of the new testcase.

vmiklos: Rebasing on top of master was easy, let me look at the testing side now.

vmiklos:
Sorry, forgot to run the tests after rebase, let me fix that up first.

vmiklos:
Now with a test; indeed could do this without a test file this time. :-)

iwyu_tool.py
```diff
@@ -298,6 +298,11 @@ def start(self, verbose):
 def fixup_compilation_db(compilation_db):
     """ Canonicalize paths in JSON compilation database. """
     for entry in compilation_db:
+        # Convert relative paths to absolute ones if possible, based on the entry's directory.
+        if 'directory' in entry and not os.path.isabs(entry['file']):
+            if os.path.exists(entry['directory']):
```
kimgr: Hm, this seems speculative. If the directory doesn't exist, why is it in the compilation database?

I would expect something like:
```
filename = entry['file']
dirpath = entry.get('directory')
if dirpath and not os.path.isabs(filename):
    filename = os.path.join(dirpath, filename)

entry['file'] = os.path.realpath(filename)
```

And I think you need to adjust my cwd test to not have a `directory` in the entry.

iwyu_tool_test.py
```diff
@@ -292,6 +292,21 @@ def test_fixup_compilation_db(self):
         entry = canonical[0]
         self.assertEqual(os.path.join(self.cwd, 'Test.cpp'), entry['file'])
 
+    def test_fixup_from_entry_dir(self):
+        """ Compilation database abs path is based on an entry's directory. """
+        compilation_db = [
+            {
+                "directory": os.path.join(self.cwd, "tests"),
```
kimgr: I think you can use a synthetic path here -- looks like `realpath` does nothing if the path doesn't exist.

kimgr: Much easier correlate `/home/user/foobar` here and `self.assertEqual('/home/user/foobar/Test.cpp', entry['file'])` in the assertion if they're spelled out.

kimgr:
Some more minutiae inline.

vmiklos:
> And I think you need to adjust my cwd test to not have a `directory` in the entry.

Oh great, I somehow thought that touching existing tests is a red flag, so rather added the extra check instead to make both the test any my use-case happy. Let me tweak your test instead, that'll be then cleaner. :-)

iwyu_tool_test.py
vmiklos: Done, indeed the test is fine with a static path.

vmiklos:
I think I addressed all comments with this update, please let me know if something more is needed. Thanks!

kimgr:
Thank you!

PR #714, opened 2019-09-17
Fix final class forward-declaration bug

selat:
According to C++ standard, forward declaration of a final class [shouldn't have "final" specifier](https://en.cppreference.com/w/cpp/language/class).

IWYU didn't correctly handle this case for template classes and generated the following invalid code:
```c++
template<typename T> class final A;
```

This pull request fixes the issue by adjusting template name deduction logic in `MungedForwardDeclareLineForTemplates` function.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -452,10 +452,20 @@ string MungedForwardDeclareLineForTemplates(const TemplateDecl* decl) {
   line = Split(line, " :", 2)[0];
   // Get rid of the template body, if any (true if no superclasses).
   line = Split(line, " {", 2)[0];
+
+  // Remove "final" specifier which isn't needed for forward
+  // declarations.
+  const char kFinalSpecifier[] = " final ";
+  string::size_type final_pos = line.find(kFinalSpecifier);
+  if (final_pos != string::npos) {
+    line.replace(final_pos, sizeof(kFinalSpecifier), " ");
+  }
+
   // The template name is now the last word on the line.  Replace it
   // by its fully-qualified form.
   const string::size_type name = line.rfind(' ');
   CHECK_(name != string::npos && "Unexpected printable template-type");
+
   return PrintForwardDeclare(decl, line.substr(0, name), GlobalFlags().cxx17ns);
 }
 
--- a/tests/cxx/fwd_decl_final-d1.h
+++ b/tests/cxx/fwd_decl_final-d1.h
@@ -1,0 +1,19 @@
+//===--- fwd_decl_final-d1.h - test input file for iwyu -------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_FWD_DECL_FINAL_D1_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_FWD_DECL_FINAL_D1_H_
+
+
+template <typename T>
+class FinalTemplate final {};
+
+class FinalClass final {};
+
+#endif  // INCLUDE_WHAT_YOU_USE_TESTS_CXX_FWD_DECL_FINAL_D1_H_
--- a/tests/cxx/fwd_decl_final.cc
+++ b/tests/cxx/fwd_decl_final.cc
@@ -1,0 +1,31 @@
+//===--- fwd_decl_final.cc - test input file for iwyu ---------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/fwd_decl_final.h"
+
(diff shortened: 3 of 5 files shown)


kimgr:
Thanks for fixing this! I need some time to review the details, but the general approach looks good!

iwyu_output.cc
```diff
@@ -456,9 +456,16 @@ string MungedForwardDeclareLineForTemplates(const TemplateDecl* decl) {
   // The template name is now the last word on the line.  Replace it
   // by its fully-qualified form.  Apparently rfind's endpos
   // argument is inclusive, so substract one to get past the end-space.
-  const string::size_type name = line.rfind(' ', endpos - 1);
+  const char kFinalSpecifier[] = " final ";
+  string::size_type final_pos = line.find(kFinalSpecifier);
+  if (final_pos != string::npos) {
+    line.replace(final_pos, sizeof(kFinalSpecifier), " ");
+  }
+  string::size_type name = line.rfind(" ", endpos - 1);
```
kimgr: I was worried that `endpos` was no longer correct after the edits for `final`, but it seems it's never been relevant.

I tried removing it locally on master, and all tests still pass. I think I'll push a refactor commit to get rid of that, and maybe you can rebase?

kimgr: If you leave the `' '` space literal as a char instead of a string literal, the diff will be smaller, and it will probably generate better code.

iwyu_output.cc
```diff
@@ -456,9 +456,16 @@ string MungedForwardDeclareLineForTemplates(const TemplateDecl* decl) {
...
   CHECK_(name != string::npos && "Unexpected printable template-type");
-  return PrintForwardDeclare(decl, line.substr(0, name), GlobalFlags().cxx17ns);
```
kimgr: I think this was fine as-was (for some value of fine :)), let's not introduce the named temporary `template_name`.

tests/cxx/final_forward_declare.cc
```diff
@@ -0,0 +1,26 @@
+//===--- final_forward_declare.cc - test input file for iwyu --------------===//
```
kimgr: Please name this `fwd_decl_final.cc` instead.

tests/cxx/final_forward_declare.h
```diff
@@ -0,0 +1,30 @@
+//===--- final_forward_declare.h - test input file for iwyu ---------------===//
```
kimgr: Please name this `fwd_decl_final.h` instead.

tests/cxx/final_forward_declare.h
```diff
@@ -0,0 +1,30 @@
...
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
```
kimgr: Please use include guards instead of `#pragma once`, see other tests for naming.

tests/cxx/final_class_impl.h
```diff
@@ -0,0 +1,13 @@
+//===--- final_class_impl.h - test input file for iwyu --------------------===//
```
kimgr: Please name this `fwd_decl_final-d1.h`

tests/cxx/final_class_impl.h
```diff
@@ -0,0 +1,13 @@
...
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
```
kimgr: Please use include guards instead of `#pragma once`, see other tests for naming.

tests/cxx/final_class_impl.h
```diff
@@ -0,0 +1,13 @@
...
+
+template <typename T>
+class A final {};
```
kimgr: Could you also add a non-template class with `final`? It seems to work already, I just want to make sure it's covered.

If you name `A` as `FinalTemplate` instead, you can name the new class `FinalClass` and it's more obvious what their role is.

kimgr:
Pushed the `endpos` removal refactor.

selat:
I address your comments. Do I need to squash all commits into one?

tests/cxx/fwd_decl_final-d1.h
```diff
@@ -0,0 +1,19 @@
+//===--- final_class_impl.h - test input file for iwyu --------------------===//
```
kimgr: Please also update filename in file headers

tests/cxx/fwd_decl_final.cc
```diff
@@ -0,0 +1,31 @@
+//===--- final_forward_declare.cc - test input file for iwyu --------------===//
```
kimgr: Please also update filename in file headers

tests/cxx/fwd_decl_final.h
```diff
@@ -0,0 +1,36 @@
+//===--- final_forward_declare.h - test input file for iwyu ---------------===//
```
kimgr: Please also update filename in file headers

selat:
Done :)

kimgr:
Beautiful, many thanks!

PR #720, opened 2019-10-01
Support make test

Flamefire:
After taking a break for a while from IWYU I struggled again to find what to do to execute the tests and `make test` failed with inexisting target.

I therefore added the necessary CMake commands to create such a target and use them in the travis file.

This is tested with the clang/llvm packages ('out-of-tree configuration') and should work for 'in-tree configuration'  too.

Note:
- You can start individual tests with `ctest -R IWYU_Tests`
- To show output on failure use `ctest --output-on-failure`
- To show all output use `ctest --extra-verbose`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ __pycache__
 
 .gdb_history
 compile_commands.json
+
+/build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,3 +125,17 @@ install(PROGRAMS fix_includes.py iwyu_tool.py DESTINATION bin)
 # Install mapping files
 file(GLOB MAPPING_FILES *.imp)
 install(FILES ${MAPPING_FILES} DESTINATION share/include-what-you-use)
+
+find_package(PythonInterp)
+if(PYTHONINTERP_FOUND)
+  enable_testing()
+  add_test(NAME iwyu_tests
+    COMMAND ${PYTHON_EXECUTABLE} run_iwyu_tests.py -- $<TARGET_FILE:include-what-you-use>
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+  add_test(NAME fix_includes_test
+    COMMAND ${PYTHON_EXECUTABLE} fix_includes_test.py
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+  add_test(NAME iwyu_tool_test
+    COMMAND ${PYTHON_EXECUTABLE} iwyu_tool_test.py
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+endif()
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -125,3 +125,17 @@ install(PROGRAMS fix_includes.py iwyu_tool.py DESTINATION bin)
 # Install mapping files
 file(GLOB MAPPING_FILES *.imp)
 install(FILES ${MAPPING_FILES} DESTINATION share/include-what-you-use)
+
+find_package(PythonInterp)
+if(PYTHONINTERP_FOUND)
+  enable_testing()
+  add_test(NAME IWYU_Tests
```
kimgr: Let's keep target names lower-case -- `iwyu_tests`

CMakeLists.txt
```diff
@@ -125,3 +125,17 @@ install(PROGRAMS fix_includes.py iwyu_tool.py DESTINATION bin)
...
+    COMMAND ${PYTHON_EXECUTABLE} run_iwyu_tests.py -- $<TARGET_FILE:include-what-you-use>
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+  add_test(NAME Fix_Includes_Test
```
kimgr: `fix_include_test`

CMakeLists.txt
```diff
@@ -125,3 +125,17 @@ install(PROGRAMS fix_includes.py iwyu_tool.py DESTINATION bin)
...
+    COMMAND ${PYTHON_EXECUTABLE} fix_includes_test.py
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+  add_test(NAME IWYU_Tool_Test
```
kimgr: iwyu_tool_test

.gitignore
```diff
@@ -6,3 +6,5 @@ __pycache__
 
 .gdb_history
 compile_commands.json
+
+/build
```
kimgr: Tip: don't create your build directory in the source dir, but rather one level out. I have the following structure:

```
~/code/iwyu
~/code/iwyu/build
~/code/iwyu/include-what-you-use
~/code/iwyu/include-what-you-use.org
```

Because that makes sure the source tree is always free of build artifacts. To each their own, I guess, so this ignore is fine :)

Flamefire: I found it to be common practice to create a 'build' folder in the source tree which is ignored and used for the same purpose: Delete it to start over. Common sequence:
```
cd foo-bar-git-folder
mkdir build && cd build
cmake ..
make
make test
```

Even the travis file here does this ;)

Just to be clear: I fully agree with your reasoning just with different structure. Reason: I usually have a "git" folder and all repos inside there and each has a build folder inside. Having to create another layer of directories would be cumbersome

Besides that it is easier to teach: Go to repo, paste the same 5 commands, done.

Flamefire:
Done. Note though that these are test names not target names. So you can't do `make iwyu_test`

These names are used by ctest (see initial post)

kimgr:
Ah, I see. Would it be possible to create phony targets with the same names? I'm not good enough with cmake to know how well it supports targets.

Flamefire:
> Ah, I see. Would it be possible to create phony targets with the same names?

You mean so it executes individual tests with e.g. `make iwyu_tests`? Why? What is wrong with `ctest -R iwyu_tests`. I could add a `make check` though (maybe) which would be `make && make test` but wouldn't recommend it.

kimgr:
Thanks, looks good! Can you squash the last fixup commit into the original "make test" commit, and I can merge.

Flamefire:
Done

kimgr:
Thanks!

PR #726, opened 2019-10-13
Fix copy-paste mistake

i-ky:
Contributing guidelines mention talks about "fixing bugs in clang". I guess it's a copy-paste mistake.


kimgr:
Thank you -- amazing that that was broken for so long!

PR #727, opened 2019-10-19
Support shortened includes for header-and-cpp file pairs

selat:
It's quite common code style to use a shorter name for a header directly related to cpp file:

```cpp
// important/module/main.hpp

#pragma once

class Main {
 public:
  Main();
};

// important/module/main.cpp

#include "main.hpp"

#include "important/module/other.hpp"

Main::Main() {
}
```

IWYU already supports it for simple use cases, but breaks when mapping files are used. If for the example above we have the following mapping file:

```json

[
  {
    "include": [
      "\"important/module/main.hpp\"",
      "private",
      "<ImportantModule/main.hpp>",
      "public"
    ]
  }
]
```

Then IWYU will try to replace `#include "main.hpp"` with `#include <ImportantModule/main.hpp>`.

This PR updates `ConvertToQuotedInclude` function to properly handle these situations.


kimgr:
I'm a little surprised the patch is so large (I haven't had a chance to think about the problem or look at the details). I wonder if you could use something like `BelongsToMainCompilationUnit`? The IWYU term for this is "associated header", maybe searching the code base for that can yield ideas?

kimgr:
@selat I don't quite understand what's going on in this patch...

The test wires a mapping file that maps `test/cxx/paired_header_and_cpp.h` to `<PairedHeaderAndCpp/paired_header_and_cpp.h>`. But since it never includes the former, I'm not sure why that mapping exists.

Also, it looks like IWYU now blindly forces path-less includes of the main file's header. I don't understand how that passes the existing test suite at all...?

I might be missing some detail that ties all this together, but it currently looks like `IsHeaderForMainFile` never returns true and your test fails to notice.

selat:
@kimgr if you run the test without the fix, IWYU will replace include with the one from mapping.

Yes, it indeed forces path-less includes, which is probably not very scalable behavior. What do you think about adding new IWYU flag, something like `--associated-header-format`, with 3 options:
* default - keep current format.
* short - force path-less includes.
* full - force full include path.

kimgr:
@selat I didn't have the current policy clear in my mind until I started looking at this more closely.

It seems like we currently;

* attempt to find the path of the included file along the header search path
* if not found, attempt to use the path of the includer as an implicit header search path

So you want to do the same, but institute a special rule for the associated header, that short-circuits before all this, right? And this associated-header rule is very similar to the second bullet above, just that the "includer" is always the main file.

I'm not so happy with `ConvertToQuotedInclude` -- it advertises a simple, pure (in the functional-programming sense) interface to translate a file path into an include-name. But behind the scenes it depends on user-provided global state (the header search path) to do its job.

And there's also several uses of `ConvertToQuotedInclude` that seem to boil down to "are these two files the same", which seems like using the wrong tool for the job (comparing `FileEntry*` would be more correct).

Introducing more global state into `ConvertToQuotedInclude` (i.e. implicitly knowing about the main file) seems bad, but I don't really have concrete ideas for what to do instead. I think maybe the path-munging should happen before calling `ConvertToQuotedInclude` in the case(s) where it matters. Not sure if that's workable...

zaufi:
@selat 

Does this PR would fix the #759?

kimgr:
No feedback on the latest set of review comments. Closing.

PR #729, opened 2019-10-21
Temp clang 9 test

kimgr:
Using this MR to test if coming clang_9.0 branch builds.


kimgr:
Purpose served, thanks for watching!

Issue #100, opened 2015-06-15
iwyu built against MacPorts LLVM-3.3 prelease can't find stdarg.h


vsapsai:
Originally reported on Google Code with ID 100

```
What steps will reproduce the problem?  Give the *exact* arguments passed
to include-what-you-use, and attach the input source file that gives the
problem (minimal test-cases are much appreciated!)
1. Run IWYU against the following file:
➜  iwyu  clear
➜  iwyu  cat ./cstdarg.cpp
#include <cstdarg>

int main(int argc, char* argv[]) {
    return 0;
}.

What is the expected output? What do you see instead?

I'd expect this file to compile without warnings or errors. Potentially, IWYU might
tell me that cstdarg is not needed. However, what I actually get is a compilation error:

➜  iwyu  include-what-you-use ./cstdarg.cpp
In file included from ./cstdarg.cpp:1:
/usr/include/c++/4.2.1/cstdarg:51:10: fatal error: 'stdarg.h' file not found
#include <stdarg.h>
         ^

/usr/include/c++/4.2.1/cstdarg should add these lines:

/usr/include/c++/4.2.1/cstdarg should remove these lines:
- #include <bits/c++config.h>  // lines 50-50

The full include-list for /usr/include/c++/4.2.1/cstdarg:

---

(cstdarg.cpp has correct #includes/fwd-decls)

What version of the product are you using? On what operating system?
I'm using r472 on OS X 10.8 built against MacPorts LLVM 3.3 prerelease:
➜  iwyu  port list clang-3.3 llvm-3.3
clang-3.3                      @3.3-r180025    lang/llvm-3.3
llvm-3.3                       @3.3-r180025    lang/llvm-3.3

Please provide any additional information below.





```

Reported by `andrew.c.morrow` on 2013-06-29 18:18:59

pfaffe:
There has been zero progress on this issue for years. Is there a roadmap?

pfaffe:
Sorry for unassigning, I have no idea why GH did this.

kimgr:
No definitive roadmap in sight.

The problem is that the headers belong to Clang and if it's not installed, the headers are not available.

Furthermore, if Clang *is* installed, IWYU needs to either know where or be installed in the same place, to be able to find them.

So in a sense, this is a packaging problem -- an IWYU package should depend on the Clang package, to ensure the headers are available. It also needs to install to the same bin path.

[...]

PR #730, opened 2019-10-22
Copy clang builtin headers to build directory

kimgr:
This is a second step for issue #100. It doesn't solve anything on its own, but here's some rationale:

* After IWYU is built to $build/bin, it can't be used on any interesting source files
* The reason is that it searches for the Clang builtin headers (`stddef.h`, `stdarg.h`, etc) in $build/lib/clang/<clang version>/include (this is imposed by the Clang libraries)
* In order to make IWYU fully usable, we need to make those headers available on that particular path

In 99f2eab, I added a build dependency on the `clang-headers` target, which copies the right headers into the build directory. Unfortunately that target is not exported in the Clang packages, and only available in the (unsupported) in-tree build.

Clang subsequently renamed the target to `clang-resource-headers`.

This PR takes care of both the new name, and puts in a home-brewn `clang-resource-headers` target for out-of-tree builds. With some tacit knowledge, we can collect the right files into the build directory.

I chose not to include the headers in the install target, because I think IWYU packaging should depend on Clang, and always be installed to the same prefix.

But for now, merging this PR will make the build directory self-consistent and usable for both in-tree and out-of-tree builds. This is a great win!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,10 +55,10 @@ add_llvm_executable(include-what-you-use
   )
 
 if (IWYU_IN_TREE)
-  # Add a dependency on clang-headers to ensure the builtin headers are
+  # Add a dependency on clang-resource-headers to ensure the builtin headers are
   # available when IWYU is executed from the build dir.
-  # The clang-headers target is only available in in-tree builds.
-  add_dependencies(include-what-you-use clang-headers)
+  # The clang-resource-headers target is only available in in-tree builds.
+  add_dependencies(include-what-you-use clang-resource-headers)
 endif()
 
 if (MINGW)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,12 +54,35 @@ add_llvm_executable(include-what-you-use
   iwyu_verrs.cc
   )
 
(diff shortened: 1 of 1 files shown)


kimgr:
Note also that these are necessary for the clang_9.0 branch. For some reason, the travis build is failing because these builtin headers are missing (I don't really understand why master/10 works).

sethrj:
This new piece of code makes the assumption that `${CMAKE_PREFIX_PATH}` is identical to the LLVM installation prefix. In general this won't be the case. Furthermore, since the arguments to `file` aren't quoted, if `${CMAKE_PREFIX_PATH}` is a list then it will be expanded before being passed to the globbing statement, so the result of this error is non-obvious:
```
make[2]: *** No rule to make target `/projects/spack/opt/spack/gcc-8.3.0/llvm/bgg33bm/lib/clang/llvm/bgg33bm/lib/clang/9.0.1/include/__clang_cuda_builtin_vars.h', needed by `lib/clang/llvm/bgg33bm/lib/clang/9.0.1/include/__clang_cuda_builtin_vars.h'.  Stop.
```

I will submit a PR to fix this.

PR #731, opened 2019-11-01
Link with clang-cpp if only that is available

aaronpuchert:
Since LLVM 9, the Clang component libraries are also linked into [clang-cpp](http://releases.llvm.org/9.0.0/tools/clang/docs/ReleaseNotes.html#build-system-changes), which provides the full C++ API. So we link with that if it is available and the component libraries are not.


kimgr:
Thank you, looks great!

PR #732, opened 2019-11-02
Add man page as a brief reference documentation

aaronpuchert:
The manual page is compiled from the builtin help and the extensive documentation, condensed down to the essence. We leave details and elaborated examples to the Markdown documentation.

Also takes installation paths from `GNUInstallDirs` (which provides the standard POSIX/FHS paths) instead of hardcoding them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,12 +147,13 @@ endif()
 add_definitions(-DIWYU_GIT_REV="${IWYU_GIT_REV}")
 
 # Install programs
-install(TARGETS include-what-you-use RUNTIME DESTINATION bin)
-install(PROGRAMS fix_includes.py iwyu_tool.py DESTINATION bin)
+include(GNUInstallDirs)
+install(TARGETS include-what-you-use RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+install(PROGRAMS fix_includes.py iwyu_tool.py DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 # Install mapping files
 file(GLOB MAPPING_FILES *.imp)
-install(FILES ${MAPPING_FILES} DESTINATION share/include-what-you-use)
+install(FILES ${MAPPING_FILES} DESTINATION ${CMAKE_INSTALL_DATADIR}/include-what-you-use)
 
 find_package(PythonInterp)
 if(PYTHONINTERP_FOUND)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,11 @@ install(PROGRAMS fix_includes.py iwyu_tool.py DESTINATION ${CMAKE_INSTALL_BINDIR
 file(GLOB MAPPING_FILES *.imp)
 install(FILES ${MAPPING_FILES} DESTINATION ${CMAKE_INSTALL_DATADIR}/include-what-you-use)
 
+# Install man page on Unix-like systems
+if (UNIX)
+  install(FILES include-what-you-use.1 DESTINATION ${CMAKE_INSTALL_MANDIR}/man1)
+endif()
+
 find_package(PythonInterp)
 if(PYTHONINTERP_FOUND)
   enable_testing()
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -1,0 +1,282 @@
+.\" t -*- coding: UTF-8 -*-
+.\" Man page for include-what-you-use
+.\"
+.\" This file is distributed under the University of Illinois Open Source
+.\" License. See LICENSE.TXT for details.
+.\"
+.TH INCLUDE-WHAT-YOU-USE 1 "2019-11-02" include-what-you-use "User Commands"
+.SH NAME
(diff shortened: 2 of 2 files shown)


aaronpuchert:
On GNU systems, the man page can be inspected with `groff -man -Tascii include-what-you-use.1` (or `-Tpdf`). If you don't want this to be installed on Windows (or Mac OS?), let's figure out the right condition for installation.

I've been maintaining this in openSUSE since 0.9, and given the moderate pace of interface changes I think I'll be able to keep this up to date.

kimgr:
@aaronpuchert Thanks, this looks good in principle!

I'm a little worried that the manpage will rot over time -- I've never gotten around to learning `troff`/`groff`, so I won't be able to keep it up to date easily. I wonder if we could look into generating the man source from markdown? I realize the manpage has a mostly fixed structure, and we'd need to somehow compose it from `include-what-you-use --help` and other documentation, so maybe it's a fool's errand...

>  If you don't want this to be installed on Windows (or Mac OS?), let's figure out the right condition
> for installation.

Yeah, I was just thinking about that. We probably don't want it for Windows (unless Cygwin or MinGW). I don't know what the story is with Mac and man pages. FreeBSD has `man`, so maybe the Mac does too?

Then again, an extra text file does no harm; packagers can remove it for the platforms where it doesn't make sense.

aaronpuchert:
> I'm a little worried that the manpage will rot over time

I understand your concern, and that's why I've been waiting to submit it until now (I wrote it almost two yours ago). I can only say that I [kept it up](https://build.opensuse.org/package/revisions/devel:tools:compiler/include-what-you-use) for the last few releases and so I think I should be able to do so in the future.

> I've never gotten around to learning `troff`/`groff`, so I won't be able to keep it up to date easily. I wonder if we could look into generating the man source from markdown? I realize the manpage has a mostly fixed structure, and we'd need to somehow compose it from `include-what-you-use --help` and other documentation, so maybe it's a fool's errand...

You're right, there are two issues here: the [grammar of troff/groff](http://man7.org/linux/man-pages/man7/groff_man.7.html), and the [conventions of man pages](http://man7.org/linux/man-pages/man7/man-pages.7.html). I'm actually not sure which took me longer. I'm not writing man pages every day, so I tend to forget things, but once a skeleton is there, it's easy to just copy-and-paste stuff around.

With Markdown I think the problem is that it doesn't have all the elements that man pages have. I wouldn't know how to write something like those “named lists” in Markdown:

```
       --keep=glob
              Always  keep  the  includes  matched by glob.  This flag may be
              used multiple times to specify more than one glob pattern.
```

One could perhaps use multiple levels of indentation, but then one needs to make sure that there is only one line before the second level starts...

> > If you don't want this to be installed on Windows (or Mac OS?), let's figure out the right condition
> > for installation.
> 
> Yeah, I was just thinking about that. We probably don't want it for Windows (unless Cygwin or MinGW). I don't know what the story is with Mac and man pages. FreeBSD has `man`, so maybe the Mac does too?

It seems that [Apple knows man pages](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man5/manpages.5.html). So maybe we can go with `NOT WIN32 OR CYGWIN OR MINGW`?

aaronpuchert:
> So maybe we can go with `NOT WIN32 OR CYGWIN OR MINGW`?

Or maybe just [UNIX](https://cmake.org/cmake/help/latest/variable/UNIX.html)? Seems to be set on Linux, Mac, Cygwin. Don't know about MinGW.

kimgr:
Thanks for the feedback. Let's give it a shot!

Re Markdown -- I think the bigger problem is that there's a sort-of template format for man pages (the whole `NAME`, `SYNOPSIS`, `DESCRIPTION`, `BUGS` format) and I guess the only reasonable way to be able to generate that is if the Markdown docs are structured the same way. And that in turn feels pretty "off" for web docs. So let's try it duplicated and see what happens :)

`UNIX` sounds like the right condition, we can make it more complicated if necessary later.

The only detail concern I had was the copyright notice -- we usually don't put individual contributor names in docs (we prefer something like `The IWYU maintainers`), but it's no big deal if you want it as-is.

aaronpuchert:
> Re Markdown -- I think the bigger problem is that there's a sort-of template format for man pages (the whole `NAME`, `SYNOPSIS`, `DESCRIPTION`, `BUGS` format) and I guess the only reasonable way to be able to generate that is if the Markdown docs are structured the same way. And that in turn feels pretty "off" for web docs. So let's try it duplicated and see what happens :)

You're right, the specific structure of man pages doesn't necessarily make sense for web documentation.

> `UNIX` sounds like the right condition, we can make it more complicated if necessary later.

Ok, then I'll use that.

> The only detail concern I had was the copyright notice -- we usually don't put individual contributor names in docs (we prefer something like `The IWYU maintainers`), but it's no big deal if you want it as-is.

The author is normally mentioned so that readers know whom they should send their complaints. But since I've added a link to the GitHub issue tracker, and git knows who wrote this, I think we can omit it.

kimgr:
I ran this through `groff` and it looks great! Thank you for putting this together.

PR #734, opened 2019-11-06
Disable forward-declares for decls in inline namespaces

kimgr:
See commit message for details.

Add a small commit to remove a method that already exists in the Clang API, to trade for the new `IsInInlineNamespace`.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -715,22 +715,34 @@ static void LogIncludeMapping(const string& reason, const OneUse& use) {
 
 namespace internal {
 
-bool DeclCanBeForwardDeclared(const Decl* decl) {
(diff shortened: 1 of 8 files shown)


iwyu_output.cc
```diff
@@ -730,6 +716,10 @@ static void LogIncludeMapping(const string& reason, const OneUse& use) {
 namespace internal {
 
 bool DeclCanBeForwardDeclared(const Decl* decl) {
```
kimgr: The logging on line 1016 is now a little misleading -- it basically says "no forward-declare, because not a class". Now there are more nuanced reasons for why a forward-declare might be disabled (and I think we should consider allowing forward-declaring free functions here at some point).

Issue #738, opened 2019-11-20
Assertion failed: desired_includes_have_been_calculated_ && "Must calculate desired includes before calling desired_includes()"


NickNick:
This assertion fails, I'm not sure what I can do to fix it. There's no other output besides that. 

Version:
```
/home/nick/.local/bin/include-what-you-use --version
include-what-you-use 0.13 (git:06e88ef) based on clang version 9.0.0 (tags/RELEASE_900/final)
```

kimgr:
Curious. Can you increase the verbosity level with `-Xiwyu - v7` and see if something interesting comes out?

NickNick:
Quite a bit. Do you have any hints or things to look for? 

These are the last few lines:

```
'const class SphereUi::Model::ViewportModel *' implicit this

/mnt/data3/code/circlelabsnl/client/src/SphereUi/Model/ViewportModel.cpp:428:12: (4) [ FunctionCall ] 0x7f0b1e211510 /mnt/data3/code/circlelabsnl/client/src/SphereUi/Model/ViewportModel.cpp:428:12: (4) [ FunctionCall ] 0x7f0b1e211510 inline constexpr ThemeModel(const SphereUi::Model::ThemeModel &) noexcept
inline constexpr ThemeModel(const SphereUi::Model::ThemeModel &) noexcept
Marked full-info use of decl 0x7f0b1e211510 Marked full-info use of decl 0x7f0b1e211510 SphereUi::Model::ThemeModel::ThemeModel (from ../client/src/SphereUi/Model/Theme/ThemeModel.h:18:7) at /mnt/data3/code/circlelabsnl/client/src/SphereUi/Model/ViewportModel.cpp:428:12SphereUi::Model::ThemeModel::ThemeModel (from ../client/src/SphereUi/Model/Theme/ThemeModel.h:18:7) at /mnt/data3/code/circlelabsnl/client/src/SphereUi/Model/ViewportModel.cpp:428:12

/mnt/data3/code/circlelabsnl/client/src/SphereUi/Model/ViewportModel.cpp:428:12: (4) [ Destruction ] /mnt/data3/code/circlelabsnl/client/src/SphereUi/Model/ViewportModel.cpp:428:12: (4) [ Destruction ] class SphereUi::Model::ThemeModel
[...]

PR #740, opened 2019-11-23
Add associated headers' glob to report violations for.

NickNick:
See github issue #738.

One block below, the same call exists to `AddGlobToReportIWYUViolationsFor`. It seems to have been forgotten in this branch.


kimgr:
I think the missing thing for the testcases is to add `.` to the include path for the new test in `run_iwyu_tests.py`. It's a little roundabout, but it should work.

Please also take a look at the style guide for tests, https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#regression-tests.

If you feel it's too overwhelming, I can try and polish up the changes and get it merged.

Thank you for doing all the legwork!

NickNick:
> It's a little roundabout, but it should work.

Hehehe.

I made a new pull request (741) for master. I'll take a look at the docs, thanks for pointing them out.

PR #741, opened 2019-11-23
#738 Add associated headers also via FileSkipped

NickNick:


NickNick:
I'm not sure what happened here... sorry for the mess. I don't do these pull requests often. I just tried to update it, not close it.. but I can't reopen it either.

NickNick:
Okay, I think we're all set.

1 remaining issue: the test fails because 'it' says roundabout.cc has correct includes/fwd-decls, but the order is 'wrong'... I'm not sure what the behavior of the program should be in that case however, so please change that to whatever is expected in this case. Thanks!

kimgr:
I think you may be falling victim to https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#force-diagnostics -- I suspect if IWYU finds nothing to do, it's not going to reorder your includes either.

kimgr:
I pushed a branch to make the tests more idiomatic: https://github.com/kimgr/include-what-you-use/tree/738-tests

But I noticed when playing with this that neither your original test nor my new one actually reproduce the problem. I'm not sure why that is, I think your original root cause analysis made sense. Any ideas? Can you repro the assertion failure with your test if your remove the `AddGlobToReportIWYUViolationsFor` call?

kimgr:
I suspect it's some path canonicalization issue surfacing like this. Your original bug report had absolute paths in the log output, and the only way I've managed to reproduce this is with:

* `include-what-you-use -I . /an/absolute/path/associated_skipped.cc`
* `associated_skipped.cc` has `#include "associated_skipped.h"`
* `associated_skipped-i1.h` has `#include "tests/cxx/associated_skipped.h"`

So it seems the combination of absolute input paths and including the associated header by two different names triggers this.

Applying your patch fixes the crash, but the resulting IWYU analysis seems pretty confused:
```
$ include-what-you-use -I. /abs/path/include-what-you-use/tests/cxx/associated_skipped.cc

(/abs/path/include-what-you-use/tests/cxx/associated_skipped.h has correct #includes/fwd-decls)

/abs/path/include-what-you-use/tests/cxx/associated_skipped.cc should add these lines:
#include "tests/cxx/associated_skipped.h"
#include "tests/cxx/associated_skipped-i1.h"  // for quad

/abs/path/include-what-you-use/tests/cxx/associated_skipped.cc should remove these lines:
- #include "associated_skipped-d1.h"  // lines 19-19

The full include-list for /abs/path/include-what-you-use/tests/cxx/associated_skipped.cc:
#include "tests/cxx/associated_skipped.h"
#include "associated_skipped.h"               // for twice
#include "tests/cxx/associated_skipped-i1.h"  // for quad
---
```

NickNick:
Thanks. Haven't had time to look at it today.

The confused bit is that it doesn't 'see' it's the same header?

kimgr:
Yeah. It's complicated.

* `AddDirectInclude` does `BelongsToMainCompilationUnit` analysis and marks the header as associated
* `FileChanged_EnterFile` first calls `AddDirectInclude` and then does its own `BelongsToMainCompilationUnit` analysis, which calls `AddGlobToReportIWYUViolationsFor`
* `FileSkipped` calls `AddDirectInclude`, but does not eventually call `AddGlobToReportIWYUViolationsFor` for associated headers

As Clang internally uses something more canonical than path names (probably `FileEntry*`) to know whether a file has already been visited, we'll get a `FileChanged_EnterFile` first for the path name that fails the `BelongsToMainCompilationUnit` test. On processing the next include, Clang realizes that this file has already been entered, and calls `FileSkipped` instead. `AddDirectInclude` does not call `AddGlobToReportIWYUViolationsFor` when `BelongsToMainCompilationUnit` succeeds, so now the header has been marked as associated, but we don't have it in our files-to-process set.

`FileSkipped` does not follow up with `AddGlobToReportIWYUViolationsFor`, and so the assert will fail later.

The trick to reproducing this is that the first `FileChanged_EnterFile` -> `AddDirectInclude` flow must not classify the file as an associated header, but the second `FileSkipped` -> `AddDirectInclude` must do so.

And the only way to get to `FileSkipped` is to have a successful `FileChanged_EnterFile` before.

kimgr:
This means your patch fixes a bug, but I can't think of a way to test it with the current infrastructure.

NickNick:
Can we merge this 'fix' in the meanwhile? I understand you are not 100% happy with the end result, but I do think it's a net improvement and I hope you do too. Maybe a follow-up issue to make includes are unique? 

Thanks a bunch!

kimgr:
I think the bugfix is fine, it just bugs me that I can't think of a way to test it.

But I do think it's better to get it merged without a test than not. However, can you add a `TODO` comment before the line you added, so we know it's not covered by tests. I'd like something like:
```diff
@@ -533,6 +533,11 @@ void IwyuPreprocessorInfo::AddDirectInclude(
     VERRS(4) << "Marked " << GetFilePath(includee)
              << " as associated header of " << GetFilePath(includer) << ".\n";
 
+    // All associated headers need to be included in IWYU analysis.
+    // TODO: This AddGlob... cannot be covered with our current test framework.
+    // It is required if IWYU is invoked with an absolute source path and its
+    // associated header is included by two different path names (e.g.
+    // "rel/path/assoc.h" and "assoc.h") in different files.
     AddGlobToReportIWYUViolationsFor(GetFilePath(includee));
   }
```

kimgr:
Hey @NickNick, do you mind if I take over this, clean it up and get it merged? I'll make sure you're maintained as author of the fix.

kimgr:
Now merged with some extensive commentary to try and explain what was wrong, and why the test is not really testing it. Thank you for doing the legwork on this!

PR #742, opened 2019-11-26
Correctly handle enums which use typedefed types

selat:
Previously enum declarations based on integers types weren't traversed by IWYU and it was suggesting to remove header with types definitions:

```cpp
#include <cstdint> // IWYU removed this header

enum class Test : int32_t {
    VALUE1,
    VALUE2
};
```

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1800,6 +1800,17 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  bool VisitEnumDecl(clang::EnumDecl* decl) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+
+    clang::QualType integer_type = decl->getIntegerType();
+    if (const clang::Type* type = integer_type.getTypePtrOrNull()) {
+      ReportTypeUse(CurrentLoc(), type);
+    }
+    return Base::VisitEnumDecl(decl);
+  }
+
   // If you say 'typedef Foo Bar', C++ says you just need to
   // forward-declare Foo.  But iwyu would rather you fully define Foo,
   // so all users of Bar don't have to.  We make two exceptions:
--- a/tests/cxx/enum_base.cc
+++ b/tests/cxx/enum_base.cc
@@ -1,0 +1,29 @@
+//===--- enum_base.cc - test input file for iwyu --------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/enum_base-d1.h"
+
+// IWYU: int_t is...*tests/cxx/enum_base-i1.h
+enum class Test : int_t {
+    VALUE1,
+    VALUE2
+};
+
+/**** IWYU_SUMMARY
+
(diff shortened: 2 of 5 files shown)


tests/cxx/enum_int_type.cc
```diff
@@ -0,0 +1,20 @@
+//===--- enum_int_type.cc - test input file for iwyu ----------------------===//
```
kimgr: Consider naming this `enum_base` instead -- slightly shorter.

tests/cxx/enum_int_type.cc
```diff
@@ -0,0 +1,20 @@
...
+
+#include <cstdint>
+
+enum class Test : int32_t {
+    VALUE1,
+    VALUE2
+};
+/**** IWYU_SUMMARY
+
+(tests/cxx/enum_int_type.cc has correct #includes/fwd-decls)
```
kimgr: Please see https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#regression-tests, and particularly https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#force-diagnostics.

I'd even go so far as suggest you shouldn't use `<cstdint>` here, because we're not in control of its contents -- it could (and probably does) include `stdint.h`, which in turn could include `bits/inttypes.h` or something.

So I'd make this three files:
```
// enum_base-i1.h
typedef int int_t;

// enum_base-d1.h
#include "tests/cxx/enum_base-i1.h"

// enum_base.cc
#include "tests/cxx/enum_base-d1.h"

...
```

It's significantly more complicated, but it gives a diagnostic on the use of `int_t` in `enum_base.cc`, which you can use to verify that it's finding that specific use.

selat:
Thanks for you suggestions! I addressed code review comments.

iwyu.cc
```diff
@@ -3933,6 +3933,16 @@ class IwyuAstConsumer
     return Base::VisitTypedefType(type);
   }
 
+  bool VisitEnumDecl(clang::EnumDecl* decl) {
```
kimgr: This wants to move closer to the other `Visit` methods for `Decl`s (there's a comment somewhere saying `// Visitors of types derived from clang::Decl.`).

Also, I think this should have the `CanIgnoreCurrentASTNode()` guard clause like the others.

kimgr: (but as a two-liner, see other PR)

iwyu.cc
```diff
@@ -3933,6 +3933,16 @@ class IwyuAstConsumer
...
+    clang::QualType integer_type = decl->getIntegerType();
+    if (!integer_type.isNull()) {
```
kimgr: You can say `if (const clang::Type* type = integer_type.getTypePtrOrNull())` to get both of these conditions in a single expression.

tests/cxx/enum_base.cc
```diff
@@ -0,0 +1,29 @@
...
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/enum_base-d1.h"
+
+// IWYU: int_t is defined in "tests/cxx/enum_base-i1.h", which isn't directly #included.
```
kimgr: Pattern-matching assert

kimgr:
LGTM, thank you!

PR #743, opened 2019-11-26
Support hpp extension for main CU files in fix_includes.py

selat:
Quite common header file extension for C++ projects.


kimgr:
Thank you! This looks good to me as-is.

PR #744, opened 2019-11-26
Ignore file extension case when detecting main CU in fix_includes.py

selat:
With this approach there's no need in explicitly specifying all cases in a replacement regex `(-inl\.h|\.h)`.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1571,8 +1571,8 @@ def _IsMainCUInclude(line_info, filename):
     return True
   # First, normalize the includee by getting rid of -inl.h and .h
   # suffixes (for the #include) and the "'s around the #include line.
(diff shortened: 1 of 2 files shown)


kimgr:
Not so sure about this, since it makes the `-inl.h` case-insensitive too. Convince me that that's good? :)

selat:
It makes the regex simpler and I don't think there are setups when developers would want "`-inl.h` to be considered main CU and `-INL.H` not the main one.

kimgr:
Yeah, makes sense. Can you rebase and add a test for mixed-case endings?

selat:
Addressed code review comments.

kimgr:
LGTM, thanks!

PR #745, opened 2019-11-26
Handle binary type traits

selat:
Clang has 4 binary type traits

```
__is_base_of
__is_convertible_to
__is_convertible
__is_same
```

and all of them require full type information even for pointers, so the following code would work correctly:

```cpp
#include "Derived.hpp"
#include "Base.hpp"

static_assert(__is_convertible_to(Derived*, Base*), "Should be convertible!");
```

**NOTE:** Replacing includes with forward declarations won't break the build for this case, but will change the behavior: type trait will return `false` instead of `true`.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2224,6 +2224,29 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  bool VisitTypeTraitExpr(clang::TypeTraitExpr* expr) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+
+    // We assume that all type traits with >= 2 arguments require
+    // full type information even for pointer types. For example,
+    // this is the case for `__is_convertible_to` trait.
+    if (expr == nullptr || expr->getNumArgs() < 2)
+      return true;
+
+    for (const clang::TypeSourceInfo* arg : expr->getArgs()) {
+      clang::QualType qual_type = arg->getType();
+      const Type* type = qual_type.getTypePtr();
+      const Type* deref_type = RemovePointersAndReferencesAsWritten(type);
+
+      if (!CanIgnoreType(deref_type)) {
+        ReportTypeUse(CurrentLoc(), deref_type);
+      }
+    }
+
+    return true;
+  }
+
   // Mark that we need the full type info for the thing we're taking
   // sizeof of.  Sometimes this is double-counting: for
   // sizeof(some_type), RecursiveASTVisitor will visit some_type and
--- a/tests/cxx/binary_type_trait.cc
+++ b/tests/cxx/binary_type_trait.cc
@@ -1,0 +1,48 @@
+//===--- binary_type_trait.cc - test input file for iwyu ------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/binary_type_trait-d1.h"
+
+int main() {
+    // IWYU: BinaryTypeTraitBase is...*binary_type_trait-i1.h
+    // IWYU: BinaryTypeTraitBase needs a declaration
+    // IWYU: BinaryTypeTraitDerived is...*binary_type_trait-i2.h
+    // IWYU: BinaryTypeTraitDerived needs a declaration
+    static_assert(__is_convertible_to(BinaryTypeTraitDerived*, BinaryTypeTraitBase*),
+        "Derived should be convertible to the Base class");
+
+    // IWYU: BinaryTypeTraitBase is...*binary_type_trait-i1.h
+    // IWYU: BinaryTypeTraitBase needs a declaration
+    // IWYU: BinaryTypeTraitDerived is...*binary_type_trait-i2.h
+    // IWYU: BinaryTypeTraitDerived needs a declaration
+    static_assert(!__is_convertible_to(BinaryTypeTraitDerived**, BinaryTypeTraitBase**),
+        "Indirect pointers shouldn't be convertible");
+
+    // IWYU: BinaryTypeTraitBase is...*binary_type_trait-i1.h
+    // IWYU: BinaryTypeTraitBase needs a declaration
+    // IWYU: BinaryTypeTraitDerived is...*tests/cxx/binary_type_trait-i2.h
+    // IWYU: BinaryTypeTraitDerived needs a declaration
+    static_assert(__is_convertible_to(BinaryTypeTraitDerived&, BinaryTypeTraitBase&),
+        "Derived should be convertible to the Base class");
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/binary_type_trait.cc should add these lines:
+#include "tests/cxx/binary_type_trait-i1.h"
+#include "tests/cxx/binary_type_trait-i2.h"
+
+tests/cxx/binary_type_trait.cc should remove these lines:
+- #include "tests/cxx/binary_type_trait-d1.h"  // lines XX-XX
+
+The full include-list for tests/cxx/binary_type_trait.cc:
+#include "tests/cxx/binary_type_trait-i1.h"  // for BinaryTypeTraitBase
+#include "tests/cxx/binary_type_trait-i2.h"  // for BinaryTypeTraitDerived
+
+***** IWYU_SUMMARY */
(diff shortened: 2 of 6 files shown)


iwyu.cc
```diff
@@ -2224,6 +2224,32 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+  bool VisitTypeTraitExpr(clang::TypeTraitExpr* expr) {
+    if (CanIgnoreCurrentASTNode())  return true;
+
+    // We assume that all type traits with >= 2 arguments require
+    // full type information even for pointer types. For example,
+    // this is the case for `__is_convertible_to` trait.
+    if (expr == nullptr || expr->getNumArgs() <= 1)  return true;
+
+    for (const clang::TypeSourceInfo* arg : expr->getArgs()) {
+      auto qual_type = arg->getType();
```
kimgr: Don't use `auto` unless the type is already visible in the expression.

iwyu.cc
```diff
@@ -2224,6 +2224,32 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      const Type* type = qual_type.getTypePtr();
+      if (const PointerType* pointer_type =
```
kimgr: Here you _could_ use `const auto*` to shorten line length (as ` PointerType` is clearly visible in the cast expr)

iwyu.cc
```diff
@@ -2224,6 +2224,32 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+              dyn_cast<PointerType>(type)) {
+        const Type* deref_type = pointer_type->getPointeeType().getTypePtr();
```
kimgr: It feels like this conditional chain would be slightly simplified if you did:

    if (const auto* pointer_type = dyn_cast<PointerType>(type)) {
      type = pointer_type->getPointeeType().getTypePtr();
    }

    ReportTypeUse(CurrentLoc(), type);

iwyu.cc
```diff
@@ -2224,6 +2224,32 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        ReportTypeUse(CurrentLoc(), deref_type);
+      } else {
+        // This isn't really required (underlying types will be
+        // traversed anyway), by makes type traits handling logic
```
kimgr: typo: by -> but

I'm not sure about the "more self-contained" bit. It sounds like you don't need to do anything extra anywhere for the else-leg, right, it's just redundant reporting?

See also `VisitBinaryOperator`, https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L2214. It seems to me like the arguments here could be `T**` in which case `getPointeeType()` will return `T*`? 

You could add a testcase for double indirection (and C++ references!) and see what happens. It could be that you need to strip off all pointers and references here to get to the underlying type. There should be a support function for that in `iwyu_ast_util.h`.

Also the `CanIgnoreType` call might be worth adding. I haven't fully understood what that does, but it seems like if we _can_ ignore a type, we probably should.

tests/cxx/binary_type_trait.cc
```diff
@@ -0,0 +1,22 @@
+//===--- binary_type_trait.cc - test input file for iwyu ------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/binary_type_trait-d1.h"
```
kimgr: Please use a `-d1`, `-i1` pair of headers to force diagnostics.

tests/cxx/binary_type_trait.cc
```diff
@@ -0,0 +1,22 @@
...
+#include "tests/cxx/binary_type_trait-d2.h"
+
+int main() {
+    static_assert(__is_convertible_to(BinaryTypeTraitDerived*, BinaryTypeTraitBase*),
```
kimgr: ... and if you do, you can actually detect diagnostics for all variations here:
```
static_assert(__is_convertible_to(BinaryTypeTraitDerived&, BinaryTypeTraitBase&), "...");
static_assert(__is_convertible_to(BinaryTypeTraitDerived**, BinaryTypeTraitBase**), "...");
// other traits...
```

iwyu.cc
selat: Thanks for the suggestion about `iwyu_ast_util.h`. I ended up using `RemovePointersAndReferencesAsWritten` function. Also I added `CanIgnoreType` check.

selat:
Addressed code review comments.

TIL: pointers to convertible pointers aren't convertible.

iwyu.cc
```diff
@@ -2224,6 +2224,27 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  bool VisitTypeTraitExpr(clang::TypeTraitExpr* expr) {
+    if (CanIgnoreCurrentASTNode())  return true;
+
+    // We assume that all type traits with >= 2 arguments require
+    // full type information even for pointer types. For example,
+    // this is the case for `__is_convertible_to` trait.
+    if (expr == nullptr || expr->getNumArgs() <= 1)  return true;
```
kimgr: It's unfortunate that the `CanIgnoreCurrentASTNode()` checks everywhere are packed on a single line, because I think that's really poor style (it's much more ergonomic to use debuggers when return stmts are on their own lines).

Also `getNumArgs() < 2` makes it clearer that 2 is the magic number.

iwyu.cc
```diff
@@ -2224,6 +2224,27 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  bool VisitTypeTraitExpr(clang::TypeTraitExpr* expr) {
+    if (CanIgnoreCurrentASTNode())  return true;
```
kimgr: Actually, let's break this into two lines too

tests/cxx/binary_type_trait.cc
```diff
@@ -0,0 +1,48 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/binary_type_trait-d1.h"
+
+int main() {
+    // IWYU: BinaryTypeTraitBase is defined in "tests/cxx/binary_type_trait-i1.h", which isn't directly #included.
```
kimgr: Please use the pattern-matching form of these diagnostic assertions, see e.g. `funcptrs.cc` for examples.

selat:
I agree, packed checks aren't really ergonomical. It would be nice to setup `.clang-format` configuration for this project, then it would be much easier to enforce single indentation style 😃

kimgr:
There is a clang-format config. Maybe I've just overlooked this setting...

kimgr:
Looks great, thank you!

bolshakov-a:
@selat could you remember please the specific problem which you solved with this patch? Seems like the following code is fully standard-conformant:
```cpp
#include <type_traits>
struct Base;
struct Derived;
static_assert(!std::is_convertible_v<Derived*, Base*>);
```
despite it stops compiling when `Base.hpp` and `Derived.hpp` are included. Yes, the behavior may change, but when a type trait is used, the checked property should normally be used somewhere else in the translation unit, and IWYU should report the type anyway, shouldn't it?

`VisitTypeTraitExpr` in the present form causes some unpleasant bugs like #1615, requiring `#include`s which are not actually needed. What can be the consequences of reverting this?

bolshakov-a:
@kimgr, wdyt? It would be nice to fix #1615 before the release because it looks like a regression. But the patch handling the majority of the type trait builtins more-or-less completely appears to be huge: the test is already more than 800 lines. Would it be OK, or it would be better to do it step-by-step, starting e.g. with `__is_assignable`?

kimgr:
@bolshakov-a Thanks for looking into this. Reviewing an 800-line patch trying to correlate with the spec does sound like hard work. Would it be possible to slice it by kinds of traits, e.g. simple unary, simple binary, insert with implicit conversions, binary with implicit conversions, or something? Maybe if you know a direction, refactor the current implementation to pass the current test suite, but be more amenable to improvements?

bolshakov-a:
Ok, let's start with binary type traits. Most of them requires the types to be complete, so `CanForwardDeclareType` function (which causes at now reporting the types when they're not a pointer or a reference) may be still leaved as-is.

bolshakov-a:
@kimgr, FYI: that's the proposal introducing the cited paragraph about the UB: [P1285R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1285r0.pdf). We are trying to solve that the C++ committee members were unable to solve. (Maybe, just to not over-complicate the standard, tbh).

Issue #735, opened 2019-11-15
Forward declaration suggested despite specialization being available


aaronpuchert:
Suppose I declare a template, but never define it, only specializations thereof. So we have template.hpp with content

```c++
template<typename>
struct T;

template<> struct T<int> { int x; };
```

and source.cpp with

```c++
#include "template.hpp"

int f(T<int>& t) { return t.x; }
```

This is a full use of the specialization, so removing the header is not suggested, but I'm told to declare the original template again:

```
source.cpp should add these lines:
template <typename> struct T;

source.cpp should remove these lines:

The full include-list for source.cpp:
#include "template.hpp"  // for T
template <typename> struct T;
```

Generally we want to repeat forward declarations, but here we have a full use of a specialization, so I think it's pretty clear that the additional declaration cannot be necessary.

kimgr:
Hi @aaronpuchert, thanks for the report!

Based on your example alone, your conclusion makes sense. I don't know exactly how the IWYU template analysis works to say why it doesn't work as expected, but there were some improvements made in this area recently about a year ago: https://github.com/include-what-you-use/include-what-you-use/pull/578

Maybe you can use the discussion and resulting testcases to find motivation for the current behavior?

Or turn up the log level (`-Xiwyu -v7`) to see if you can figure out in which phase IWYU makes the decision.

aaronpuchert:
> Based on your example alone, your conclusion makes sense. I don't know exactly how the IWYU template analysis works to say why it doesn't work as expected, but there were some improvements made in this area recently about a year ago: #578

If I'm not mistaken, this was talking about explicit instantiations, not specializations, of templates.

> Or turn up the log level (`-Xiwyu -v7`) to see if you can figure out in which phase IWYU makes the decision.

We have a fwd-decl use of the specialization in the parameter list of `f`, and then a full-info use of the same specialization in `t.x`. Interestingly though, on the parameter list we have a "warning: T needs a declaration, but does not provide or directly #include one." That could point to the problem.

aaronpuchert:
This is the full log, it's not so long:

```
Setting verbose-level to 7
Search path: /usr/include/c++/7/x86_64-suse-linux/ (system)
Search path: /usr/include/c++/7/backward/ (system)
Search path: /usr/include/c++/7/ (system)
Search path: /usr/local/include/ (system)
Search path: /usr/include/ (system)
Found include: <built-in>:-1 -> source.cpp
[...]

PR #765, opened 2020-03-07
Don't second-guess template specializations

kimgr:
Before this patch, IWYU would skip past template specializations and go
for the base template when trying to find the definition for a class
template.

This led to misbehavior as reported in issue #735 so that:

    // template.hpp
    template<typename>
    struct T;

    template<> struct T<int> { int x; };

    // source.cpp
    #include "template.hpp"

    int f(T<int>& t) { return t.x; }

would require a forward declaration of the base template in source.cpp.

Remove this step on the basis that we _are_ in fact using the
specialization, and the file containing the specialization will probably
have some declaration (or even definition) of the base template.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1389,14 +1389,10 @@ void CalculateIwyuForForwardDeclareUse(
 
   // If this record is defined in one of the desired_includes, mark that
   // fact.  Also if it's defined in one of the actual_includes.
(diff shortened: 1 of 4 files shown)


kimgr:
cc @aaronpuchert

aaronpuchert:
Looks good to me, thanks!

Issue #767, opened 2020-03-08
Segmentation fault when compiled with MinSizeRel (-Os)


EmilGedda:
Branch: clang_9.0
Distro: Alpine Linux
Description: Segfault when running IWYU compiled with -Os

So I'm currently embedding IWYU inside a docker image for running inside a CI.
While testing, I noticed that `iwyu_tool.py` with a MinSizeRel out-of-tree built IWYU only outputted a single empty line and then exited cleanly. When running IWYU directly through CMake, IWYU segfaults. When building IWYU without `-DCMAKE_BUILD_TYPE=MinSizeRel` IWYU (and `iwyu_tool.py`) works as expected. 

A crash due to change in optimization flags stinks of UB. If I have the time I'll try to compile IWYU with some sanitizers and `-g` to see if that reveals anything.

The cpp file which crashes IWYU is as follows:
```cpp
#include <iterator>
#include <array>

int main () {
  std::string foo;
}
```
I can provide a Dockerfile which shows the crashing and non-crashing behavior if wanted.

I expect `iwyu_tool.py` to inform me or crash itself if `include-what-you-use` segfaults, or at least exit with exit code 1.

EmilGedda:
Here is the log from `env IWYU_VERBOSE=7 ninja` when compiling the above file.
[log.txt](https://github.com/include-what-you-use/include-what-you-use/files/4303555/log.txt)

kimgr:
I usually see random segfaults when I mismatch IWYU and LLVM/Clang config flags.

I suspect the LLVM/Clang package you link with is not built with `MinSizeRel` (but something more like `RelWithDebInfo`), causing some kind of ABI mismatch and subsequent death.

For what it's worth I can reproduce on my regular Ubuntu rig, I'll see if I can debug it.

kimgr:
Hah, I figured I would run a sanitized build and see what came out:
```
$ cmake -G Ninja -DCMAKE_BUILD_TYPE=MinSizeRel -DCMAKE_PREFIX_PATH=/usr/lib/llvm-11 -DCMAKE_CXX_FLAGS="-g -fsanitize=undefined" ../../include-what-you-use
$ ninja
$ bin/include-what-you-use segv.cc                                                
.../include-what-you-use/iwyu.cc:330:48: runtime error: member call on null pointer of type 'struct A
STNode'
...
```

It turns out line 330 is `TraverseDecl`, which starts like this:
```
[...]

PR #769, opened 2020-03-09
Check for nullptr before calling member function

EmilGedda:
Calling a member function on a nullpointer object is Undefined
Behaviour. When IWYU was compiled with -Os it crashed with a segmentation
fault due to the compiler exploiting this UB for performance reasons.

This patch simply checks current_ast_node_ for nullptr before calling
member functions.

Fixes include-what-you-use/include-what-you-use#767


kimgr:
Nice, thanks!

PR #770, opened 2020-03-09
Check for nullptr before calling member function [clang_9.0]

EmilGedda:
Calling a member function on a nullpointer object is Undefined
Behaviour. When IWYU was compiled with -Os it crashed with a segmentation
fault due to the compiler exploiting this UB for performance reasons.

This patch simply checks current_ast_node_ for nullptr before calling
member functions.

Backported from 147c2b3.


kimgr:
I'd prefer not to take changes for older releases, we have enough to do keeping `master` healthy. Any chance you could carry this as a local patch?

EmilGedda:
Understandable, no problem! Will keep running my fork parallel to upstream then.

PR #772, opened 2020-03-13
Fix handling of alias templates

### LINKED ISSUE #703: IWYU includes the wrong header for alias templates
Reduced test case:

test.cc:
```
#include "foo.hh"
#include <string>

int main() {
  Foo<std::string> foo;
}
```

foo.hh:
```
#pragma once

#include "bar.hh"

template <typename T>
using Foo = Bar<T, int>;
```

bar.hh:
```
#pragma once

template<typename T, typename U>
class Bar {
};
```

IWYU reports:
```
test.cc should add these lines:
#include "bar.hh"  // for Bar

test.cc should remove these lines:
- #include "foo.hh"  // lines 1-1

The full include-list for test.cc:
#include <string>  // for string
#include "bar.hh"  // for Bar
---
```

But following this advice would break compilation as Foo would no longer be found.

jbytheway:
When an alias template was used, IWYU would resolve it to the underlying aliased type rather than the template for the purposes of determining what header was required.  This led to incorrect header removals in some cases.  The test added in this PR demonstrates such a situation.

Fixes #703.

Fix that by splitting the `TypeToDeclAsWritten` into two functions: `TypeToDeclAsWritten` and `TypeToDeclForContent`.  The former is used for determining uses of the `Decl` itself, the latter is used for determining uses of the arguments passed to the template.

`TypeToDeclForContent` has the old behaviour of `TypeToDeclAsWritten` wheras the new `TypeToDeclAsWritten` prefers to report the alias template rather than the underlying `Decl`.

I only switched one call location to `TypeToDeclForContent`.  It's possible more should be changed, but this was the only one required to make the tests pass.  If you'd like me to try to review the other call sites then I can do so.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -110,6 +110,7 @@ using clang::TemplateSpecializationKind;
 using clang::TemplateSpecializationType;
 using clang::TranslationUnitDecl;
 using clang::Type;
+using clang::TypeAliasTemplateDecl;
 using clang::TypeDecl;
 using clang::TypeLoc;
 using clang::TypedefNameDecl;
@@ -1207,7 +1208,7 @@ const Type* RemovePointersAndReferences(const Type* type) {
   return type;
 }
 
-const NamedDecl* TypeToDeclAsWritten(const Type* type) {
+static const NamedDecl* TypeToDeclImpl(const Type* type, bool as_written) {
   // Get past all the 'class' and 'struct' prefixes, and namespaces.
   type = RemoveElaboration(type);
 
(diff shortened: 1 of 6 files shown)


iwyu_ast_util.cc
```diff
@@ -1207,7 +1208,7 @@ const Type* RemovePointersAndReferences(const Type* type) {
   return type;
 }
 
-const NamedDecl* TypeToDeclAsWritten(const Type* type) {
+static const NamedDecl* TypeToDeclImpl(const Type* type, bool AsWritten) {
```
kimgr: `AsWritten` is LLVM style, IWYU uses `hacker_case` -- so please make this `as_written`

iwyu_ast_util.cc
```diff
@@ -1245,6 +1255,14 @@ const NamedDecl* TypeToDeclAsWritten(const Type* type) {
   }
 }
 
+const NamedDecl* TypeToDeclAsWritten(const Type* type) {
+  return TypeToDeclImpl(type, true);
```
kimgr: You could add a `/*as_written=*/true` here to make it clearer what the flag means. Cheapest form of named arguments.

iwyu_ast_util.cc
```diff
@@ -1245,6 +1255,14 @@ const NamedDecl* TypeToDeclAsWritten(const Type* type) {
...
+}
+
+const NamedDecl* TypeToDeclForContent(const Type* type) {
+  return TypeToDeclImpl(type, false);
```
kimgr: Named-arg comment for the Boolean literal.

iwyu_ast_util.cc
jbytheway: Fixed.

iwyu_ast_util.cc
jbytheway: Done.

iwyu_ast_util.cc
jbytheway: Done.

kimgr:
Most appreciated!

PR #773, opened 2020-03-14
Ignore tags file

jbytheway:
Add `tags` to `.gitignore`.

Tags files (a la [ctags](https://en.wikipedia.org/wiki/Ctags)) ought to be common enough that this is a reasonable entry to have.  In particular, I find them useful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,7 @@ __pycache__
 
 .gdb_history
 compile_commands.json
+tags
+TAGS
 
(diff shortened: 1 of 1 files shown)


kimgr:
Thanks! Could you also include `TAGS` in the same commit? I happen to have that spelling here.

jbytheway:
Done.

kimgr:
Thank you.

PR #775, opened 2020-03-18
Aliases in templates

### LINKED ISSUE #431: Typedefs of template arguments aren't required anywhere

### LINKED ISSUE #780: iwyu disagrees that std::optional needs a full definition
This is a simplified example of an issue I ran into this week. What is going on?

#### Steps to Reproduce

1. Create the following files

```c++
// date.h
class Date {
public:
  int hour = 0;
  int minute = 0;
  int second = 0;
};
```

```c++
// functions.h
#include <optional>

class Date;

std::optional<Date> get_default_date();
```
```c++
// functions.cpp
#include "functions.h"
#include "date.h"

std::optional<Date> get_default_date() {
  return std::optional<Date>(Date());
}
```
```c++
// thing.h
#include "date.h"
#include <optional>

std::optional<Date> get_date();
```
```c++
// thing.cpp
#include "thing.h"
#include "functions.h"

std::optional<Date> get_date() {
  return get_default_date();
}
```
2. Compile
```
clang++ -std=c++17 -c thing.cpp -c functions.cpp
```
3. IWYU
```
include-what-you-use -std=c++17 thing.cpp

...
thing.h should remove these lines:
- #include "date.h"  // lines 1-1
...
```
#### Analysis

Clang and GCC fail to compile if `thing.h` doesn't include `date.h`. If `date.h` is removed as suggested by IWYU the compiler yields

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/type_traits:3697:69: error: 
      incomplete type 'Date' used in type trait expression
    : public integral_constant<bool, is_destructible<_Tp>::value && __has_trivial_destructor(_Tp)> {};
                                                                    ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/optional:539:5: note: 
      in instantiation of template class 'std::__1::is_trivially_destructible<Date>' requested here
    is_trivially_destructible<_Tp>::value &&
    ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/optional:586:15: note: 
      in instantiation of default argument for '__optional_move_assign_base<Date>' required here
    : private __optional_move_assign_base<_Tp>
              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
thing.cpp:5:10: note: in instantiation of template class 'std::__1::optional<Date>' requested here
  return get_default_date();
```

jbytheway:
I've been working to fix the issues around uses of typedefs / aliases in templates.  This PR replaces #707, fixes #431 and fixes #780.

However, it's not a complete solution yet.  In particular, it's still not correctly handling `libstdc++`'s `std::unordered_map`, which was the issue that originally inspired me to work on this.

So, I hope to continue to improve the solution until it does handle that case correctly.

For now, I'm posting the PR to get feedback on the general approach taken, which is as follows:

The existing code detected uses of template arguments via `VisitSubstTemplateTypeParmType` which appears in the AST whenever a template argument is used in a template instantiation.

However, when an alias of a template argument is used, no such node appears.  Instead, we get `VisitTypedefType`.

The approach of this PR is to track all aliases of template arguments in a `map`, and when `VisitTypedefType` is called on an expression using such an alias, treat it similarly to how `VisitSubstTemplateTypeParmType` used to work (the two now use a common helper function for the bulk of their functionality).

The test I've rewritten demonstrates a few cases where IWYU used to (incorrectly) suggest forward declarations were sufficient, and now treats as full-uses.

There are more cases which still fail, so (as mentioned above) I hope to understand them and improve this further.  (However, this is nevertheless an improvement on the status quo, so if you're reading this a few months after I wrote it and I haven't made further progress then it might be worth merging anyway.)

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3075,19 +3075,11 @@ class InstantiatedTemplateVisitor
     return TraverseSubstTemplateTypeParmTypeHelper(typeloc.getTypePtr());
   }
 
-  // These do the actual work of finding the types to return.  Our
-  // task is made easier since (at least in theory), every time we
-  // instantiate a template type, the instantiation has type
-  // SubstTemplateTypeParmTypeLoc in the AST tree.
-  bool VisitSubstTemplateTypeParmType(clang::SubstTemplateTypeParmType* type) {
-    if (CanIgnoreCurrentASTNode() || CanIgnoreType(type))
-      return true;
-
-    // Figure out how this type was actually written.  clang always
-    // canonicalizes SubstTemplateTypeParmType, losing typedef info, etc.
-    const Type* actual_type = ResugarType(type);
-    CHECK_(actual_type && "If !CanIgnoreType(), we should be resugar-able");
-
+  // This helper is called on every use of a template argument type in an
+  // instantiated template.  Its goal is to determine whether that use should
+  // constitute a full-use by the template caller, and perform other necessary
+  // bookkeeping.
+  void AnalyzeTemplateTypeParmUse(const Type* type) {
     // TODO(csilvers): whenever we report a type use here, we want to
     // do an iwyu check on this type (to see if sub-types are used).
 
@@ -3098,8 +3090,16 @@ class InstantiatedTemplateVisitor
     // in_forward_declare_context.  I think this will require changing
     // in_forward_declare_context to yes/no/maybe.
     if (current_ast_node()->ParentIsA<NestedNameSpecifier>()) {
-      ReportTypeUse(CurrentLoc(), actual_type);
-      return Base::VisitSubstTemplateTypeParmType(type);
+      ReportTypeUse(CurrentLoc(), type);
+      return;
+    }
+
+    // If the immediate parent node is a typedef, then register the new type as
+    // a new name for the template argument.
+    if (const TypedefNameDecl* typedef_decl =
+            current_ast_node()->GetParentAs<TypedefNameDecl>()) {
+      const Type* typedef_type = typedef_decl->getTypeForDecl();
+      template_argument_aliases_.emplace(typedef_type, type);
     }
 
     // If we're inside a typedef, we don't need our full type info --
@@ -3118,40 +3118,70 @@ class InstantiatedTemplateVisitor
     for (const ASTNode* ast_node = current_ast_node();
          ast_node != caller_ast_node_; ast_node = ast_node->parent()) {
       if (ast_node->IsA<TypedefNameDecl>())
-        return Base::VisitSubstTemplateTypeParmType(type);
+        return;
     }
 
     // sizeof(a reference type) is the same as sizeof(underlying type).
     // We have to handle that specially here, or else we'll say the
     // reference is forward-declarable, below.
     if (current_ast_node()->ParentIsA<UnaryExprOrTypeTraitExpr>() &&
-        isa<ReferenceType>(actual_type)) {
-      const ReferenceType* actual_reftype = cast<ReferenceType>(actual_type);
+        isa<ReferenceType>(type)) {
+      const ReferenceType* actual_reftype = cast<ReferenceType>(type);
       ReportTypeUse(CurrentLoc(),
                     actual_reftype->getPointeeTypeAsWritten().getTypePtr());
-      return Base::VisitSubstTemplateTypeParmType(type);
+      return;
     }
 
     // If we're used in a forward-declare context (MyFunc<T>() { T* t; }),
     // or are ourselves a pointer type (MyFunc<Myclass*>()),
     // we don't need to do anything: we're fine being forward-declared.
     if (current_ast_node()->in_forward_declare_context())
-      return Base::VisitSubstTemplateTypeParmType(type);
+      return;
 
     if (current_ast_node()->ParentIsA<PointerType>() ||
         current_ast_node()->ParentIsA<LValueReferenceType>() ||
-        IsPointerOrReferenceAsWritten(actual_type))
-      return Base::VisitSubstTemplateTypeParmType(type);
+        IsPointerOrReferenceAsWritten(type))
+      return;
 
     // We attribute all uses in an instantiated template to the
     // template's caller.
-    ReportTypeUse(caller_loc(), actual_type);
+    ReportTypeUse(caller_loc(), type);
 
     // Also report all previous explicit instantiations (declarations and
     // definitions) as uses of the caller's location.
-    ReportExplicitInstantiations(actual_type);
+    ReportExplicitInstantiations(type);
+  }
+
+  // Our task is made easier since (at least in theory), every time we
+  // instantiate a template type, the instantiation has type
+  // SubstTemplateTypeParmTypeLoc in the AST tree.
+  bool VisitSubstTemplateTypeParmType(clang::SubstTemplateTypeParmType* type) {
+    if (CanIgnoreCurrentASTNode() || CanIgnoreType(type))
+      return true;
+
+    // Figure out how this type was actually written.  clang always
+    // canonicalizes SubstTemplateTypeParmType, losing typedef info, etc.
+    const Type* actual_type = ResugarType(type);
+    CHECK_(actual_type && "If !CanIgnoreType(), we should be resugar-able");
+
+    AnalyzeTemplateTypeParmUse(actual_type);
 
     return Base::VisitSubstTemplateTypeParmType(type);
+  }
+
+  bool VisitTypedefType(clang::TypedefType* type) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+
+    // Typedefs of template arguments require special handling to ensure that
+    // we record full-uses of those arguments where appropriate.  Those
+    // typedefs are stored in the template_argument_aliases_ map.
+    if (const Type* template_arg_type =
+            GetOrDefault(template_argument_aliases_, type, nullptr)) {
+      AnalyzeTemplateTypeParmUse(template_arg_type);
+    }
+
+    return Base::VisitTypedefType(type);
   }
 
   bool VisitTemplateSpecializationType(TemplateSpecializationType* type) {
@@ -3214,6 +3244,7 @@ class InstantiatedTemplateVisitor
   void Clear() {
     caller_ast_node_ = nullptr;
     resugar_map_.clear();
+    template_argument_aliases_.clear();
     traversed_decls_.clear();
     nodes_to_ignore_.clear();
     cache_storers_.clear();
@@ -3542,6 +3573,12 @@ class InstantiatedTemplateVisitor
   // value, it's a default template parameter, that the
   // template-caller may or may not be responsible for.
   map<const Type*, const Type*> resugar_map_;
+
+  // When we see a full-use of a template argument we need to assign that full
+  // use to the template-caller.  Sometimes those uses are hidden behind
+  // type aliases (typedefs).  This maps those aliases to the underlying
+  // template arguments.
+  map<const Type*, const Type*> template_argument_aliases_;
 
   // Used to avoid recursion in the *Helper() methods.
   set<const Decl*> traversed_decls_;
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3075,14 +3075,8 @@ class InstantiatedTemplateVisitor
     return TraverseSubstTemplateTypeParmTypeHelper(typeloc.getTypePtr());
   }
 
-  // This helper is called on every use of a template argument type in an
-  // instantiated template.  Its goal is to determine whether that use should
-  // constitute a full-use by the template caller, and perform other necessary
-  // bookkeeping.
-  void AnalyzeTemplateTypeParmUse(const Type* type) {
-    // TODO(csilvers): whenever we report a type use here, we want to
-    // do an iwyu check on this type (to see if sub-types are used).
-
+  // Check whether a use of a template parameter is a full use.
+  bool IsTemplateTypeParmUseFullUse(const Type* type) {
     const ASTNode* node = MostElaboratedAncestor(current_ast_node());
 
     // If we're a nested-name-specifier class (the Foo in Foo::bar),
@@ -3092,16 +3086,7 @@ class InstantiatedTemplateVisitor
     // in_forward_declare_context.  I think this will require changing
     // in_forward_declare_context to yes/no/maybe.
     if (node->ParentIsA<NestedNameSpecifier>()) {
-      ReportTypeUse(CurrentLoc(), type);
-      return;
-    }
-
-    // If the immediate parent node is a typedef, then register the new type as
-    // a new name for the template argument.
-    if (const TypedefNameDecl* typedef_decl =
-            node->GetParentAs<TypedefNameDecl>()) {
-      const Type* typedef_type = typedef_decl->getTypeForDecl();
-      template_argument_aliases_.emplace(typedef_type, type);
+      return true;
     }
 
     // If we're inside a typedef, we don't need our full type info --
@@ -3120,7 +3105,7 @@ class InstantiatedTemplateVisitor
     for (const ASTNode* ast_node = node; ast_node != caller_ast_node_;
          ast_node = ast_node->parent()) {
       if (ast_node->IsA<TypedefNameDecl>()) {
-        return;
+        return false;
       }
       if (ast_node->IsA<TemplateSpecializationType>()) {
         // If we hit a template specialization node before the typedef then we
@@ -3134,22 +3119,70 @@ class InstantiatedTemplateVisitor
     // reference is forward-declarable, below.
     if (node->ParentIsA<UnaryExprOrTypeTraitExpr>() &&
         isa<ReferenceType>(type)) {
-      const ReferenceType* actual_reftype = cast<ReferenceType>(type);
-      ReportTypeUse(CurrentLoc(),
-                    actual_reftype->getPointeeTypeAsWritten().getTypePtr());
-      return;
+      return true;
     }
 
     // If we're used in a forward-declare context (MyFunc<T>() { T* t; }),
     // or are ourselves a pointer type (MyFunc<Myclass*>()),
     // we don't need to do anything: we're fine being forward-declared.
     if (node->in_forward_declare_context())
-      return;
+      return false;
 
     if (node->ParentIsA<PointerType>() ||
         node->ParentIsA<LValueReferenceType>() ||
         IsPointerOrReferenceAsWritten(type))
+      return false;
+
+    return true;
+  }
+
+  // This helper is called on every use of a template argument type in an
+  // instantiated template.  Its goal is to determine whether that use should
+  // constitute a full-use by the template caller, and perform other necessary
+  // bookkeeping.
+  void AnalyzeTemplateTypeParmUse(const Type* type) {
+    const ASTNode* node = MostElaboratedAncestor(current_ast_node());
+
+    // If the immediate parent node is a typedef, then register the new type as
+    // a new name for the template argument.
+    if (const TypedefNameDecl* typedef_decl =
+            node->GetParentAs<TypedefNameDecl>()) {
+      const Type* typedef_type = typedef_decl->getTypeForDecl();
+      VERRS(6) << "Registering " << PrintableType(typedef_type)
+               << " as an alias for " << PrintableType(type)
+               << " in the context of this instantiation\n";
+      template_argument_aliases_.emplace(typedef_type, type);
+    }
+
+    // Figure out how this type was actually written.  clang always
+    // canonicalizes SubstTemplateTypeParmType, losing typedef info, etc.
+    const Type* actual_type = ResugarType(type);
+    CHECK_(actual_type &&
+           "Type passed to AnalyzeTemplateTypeParmUse should be resugar-able");
+
+    VERRS(6) << "AnalyzeTemplateTypeParmUse: type = " << PrintableType(type)
+             << ", actual_type = " << PrintableType(actual_type) << '\n';
+
+    if (!IsTemplateTypeParmUseFullUse(actual_type)) {
+      // Non-full uses will already have been reported when they were used as
+      // template arguments, so nothing to do here.
       return;
+    }
+
+    if (isa<ReferenceType>(actual_type)) {
+      // If the argument type is a reference type, then we actually care about
+      // the referred-to type.
+      const ReferenceType* actual_reftype = cast<ReferenceType>(actual_type);
+      type = actual_reftype->getPointeeTypeAsWritten().getTypePtr();
+    }
+
+    // At this point we know we are looking at a full-use of type.  However,
+    // there are still two cases.  If this is a template param that was written
+    // by the user, then we report a use of it.  However, it might also be a
+    // parameter of some implementation detail template which just happens top
+    // involve the user's template parameter somehow.  In this case, we need to
+    // recursively explore the instantiation of *that* template to see if the
+    // user's type is full-used therein.
 
     // We attribute all uses in an instantiated template to the
     // template's caller.
@@ -3167,12 +3200,7 @@ class InstantiatedTemplateVisitor
     if (CanIgnoreCurrentASTNode() || CanIgnoreType(type))
       return true;
 
-    // Figure out how this type was actually written.  clang always
-    // canonicalizes SubstTemplateTypeParmType, losing typedef info, etc.
-    const Type* actual_type = ResugarType(type);
-    CHECK_(actual_type && "If !CanIgnoreType(), we should be resugar-able");
-
-    AnalyzeTemplateTypeParmUse(actual_type);
+    AnalyzeTemplateTypeParmUse(type);
 
     return Base::VisitSubstTemplateTypeParmType(type);
   }
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1248,13 +1248,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   using Base::PrintableCurrentLoc;
   using Base::current_ast_node;
 
+  enum class IgnoreKind {
+    ForUse,
+    ForExpansion,
+  };
+
   //------------------------------------------------------------
   // Pure virtual methods that a subclass must implement.
 
   // Returns true if we are not interested in iwyu information for the
   // given type, where the type is *not* the current AST node.
   // TODO(csilvers): check we're calling this consistent with its API.
-  virtual bool CanIgnoreType(const Type* type) const = 0;
+  virtual bool CanIgnoreType(const Type* type,
+                             IgnoreKind = IgnoreKind::ForUse) const = 0;
 
   // Returns true if we are not interested in doing an iwyu check on
   // the given decl, where the decl is *not* the current AST node.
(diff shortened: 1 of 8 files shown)


iwyu.cc
```diff
@@ -3543,6 +3572,12 @@ class InstantiatedTemplateVisitor
   // template-caller may or may not be responsible for.
   map<const Type*, const Type*> resugar_map_;
 
+  // When we see a full-use of a template argument we need to assign that full
+  // use to the template-caller.  Sometimes those uses are hidden behind
+  // type aliases (typedefs).  This maps those aliases to the underlying
+  // template arguments.
+  map<const Type*, const Type*> template_argument_aliases_;
```
kimgr: Does this need to be separate from the resugar map? I've tried a few times to get my head around the resugaring logic, without success, so I may be missing some subtlety, but they look closely related.

iwyu.cc
```diff
@@ -3118,42 +3120,68 @@ class InstantiatedTemplateVisitor
...
+    // canonicalizes SubstTemplateTypeParmType, losing typedef info, etc.
+    const Type* actual_type = ResugarType(type);
+    CHECK_(actual_type && "If !CanIgnoreType(), we should be resugar-able");
+
+    TemplateTypeParmTypeHelper(actual_type);
 
     return Base::VisitSubstTemplateTypeParmType(type);
   }
 
+  bool VisitTypedefType(clang::TypedefType* type) {
```
kimgr: I have a really hard time with the template visitation. This is the `InstantiatedTemplateVisitor`, right? And I'm guessing it only traverses template instantiations. So why are all typedefs subject to resugaring in instantiated templates?

I'm thinking something like:

``` 
template<class T>
size_t GetTheSize(const T& t) {
  return sizeof(t);
}

int main() {
    float val = 0;
    return (int)GetTheSize(val);
}
```

`size_t` in the `GetTheSize` instantiation is visited here, right?

I wonder if `TemplateTypeParmTypeHelper` should be split into the part that registers the resugar entry, and the lookup? I don't know, it's all very confusing :)

jbytheway: Yes, this is `InstantiatedTemplateVisitor`.

The need to track typedefs comes from examples like this:
```c++
template<typename T>
size_t GetTheSize(const T& t) {
  using myT = T;
  return sizeof(myT);
}
```
Without these changes, that template argument would be considered a forward-declare-only use.

kimgr: Right, but I mean that even `size_t` is visited with this logic, even though it doesn't really need to be, right?

iwyu.cc
kimgr: Hmm, yeah, that sounds weird. I would expect the resugar map to do exactly what you're attempting here + maybe some  magic for dependent types.

jbytheway: I haven't tested this idea yet, but I *think* that `resugar_map` is designed to help with situations where template arguments don't match with template parameters.  For example,

```
template<typename T>
struct array_traits;

template<typename T, size_t N>
struct array_traits<T[N]> {
  static size_t size = N;
};

int foo(int) {
  char my_string[] = "...";
  return array_traits<decltype(my_string)>::size;
}
```

Here when walking the specialization of `array_traits`, there are two template parameters (`T` and `N`) corresponding to a single template argument (`decltype(my_string)`) and we need some way to match them up.  I've certainly seen some code that appears to try to deal with this sort of situation, but I've not experimented with it myself.

iwyu.cc
jbytheway: I don't think `size_t` will be passed to `TemplateTypeParmTypeHelper`.  Which use of `size_t` are you thinking might cause that?

kimgr: I'm really just guessing, but if `InstantiatedTemplateVisitor` traverses instantiated templates, it seems to me a use like:
```
int main() {
    float val = 0;
    return (int)GetTheSize(val);
}
```
Should visit components of `GetTheSize` and find a `typedef` in the return type. But I'll experiment a bit locally and try to bring some facts rather than theoretical worry.

I think my original trigger was that it looked like `TemplateTypeParmTypeHelper` was doing too much (the over-generic name also indicates this), and that maybe it could be split into a populate and lookup pair of methods. But let's hold off on my ideas until I've experimented a little and understand more of what's really going on :)

jbytheway: Do you mean because `size_t` is itself a typedef?  It shouldn't cause a call to `TemplateTypeParmTypeHelper` because that typedef will not have been put in the `template_argument_aliases_` map.

I agree the function needs some refactoring.  I want to separate the "is is a full use" part from what is done in response to that, because currently it's a bit inconsistent.  For example, some places therein report a use at `CurrentLoc` and others at `caller_loc`.  I'm pretty sure the former is a mistake.  And only one type of uses leads to a call to `ReportExplicitInstantiations`, which also seems unlikely to be correct.

I also agree that the name isn't great.  Perhaps if I split it up better names will present themselves.

jbytheway:
I've pushed some further updates.  This is now feature-complete in the sense that it is sufficient to allow for use of `libstdc++`'s `std::unordered_map` to be handled correctly, which was my goal with this PR.

The major behavioural change in these new commits is that IWYU now performs more recursive template visitation.  Specifically, when an template being visited uses another template in a full-use context, passing down to it an argument that originates from the user's top-level template instantiation, it now recursively visits that other template specialization.

Currently this is unnecessarily slow, because it does so every time.  I still need to hook this into the `traversed_decls_` set so that it doesn't repeat traversals unnecessarily.

I've refactored and renamed `TemplateTypeParmTypeHelper`.  It's now two functions: `IsTemplateTypeParmUseFullUse` and `AnalyzeTemplateTypeParmUse`.

The commit history still needs cleaning up, but in terms of the overall change, this is (I think) close to what I want, so your thoughts are welcome.

kimgr:
Impressive! This is a lot to chew through, so it will take me a while. But just looking at the new succeeding test cases makes me really hopeful.

kimgr:
I can recommend https://github.com/kimgr/git-format if you want to clean up your commit series wrt clang-format.

tests/cxx/typedef_in_template.cc
```diff
@@ -7,51 +7,54 @@
...
-template<class T>
-class Container {
- public:
-  // Should not be an iwyu violation for T
-  typedef T value_type;
-
-  // C++11 alias declaration, should not be an iwyu violation for T
-  using alias_type = T;
-
-  // IWYU: Pair is...*typedef_in_template-i1.h
```
kimgr, line 21: I think we lost coverage for this example. The reason this was interesting was because it proves that the template is not exempt from analysis -- the instantiator of this template should not be responsible for `Pair`, it's an implementation detail of the template.

tests/cxx/typedef_in_template.cc
```diff
@@ -7,51 +7,54 @@
...
+A<IndirectClass>::TAlias a2;
 
-  // TODO: This is almost certainly wrong, see bug #431
-  // We should not require the full definition of Class for passing it as a
-  // template argument, but we must require it when the typedef it's aliasing
-  // is full-used.
-  // The bug has instructions for how to provoke the error more obviously.
-
-  // IWYU: Class needs a declaration
-  Container<Class>::value_type vt;
```
kimgr, line 36: Same here, we need to make sure that `value_type` requires `Container`, but does not force `Container` to require `Class`. I think your new testcases demonstrate this, but it's not as easily mappable back to STL-like types.

tests/cxx/typedef_in_template.cc
```diff
@@ -7,51 +7,54 @@
...
-  using alias_type = T;
-
-  // IWYU: Pair is...*typedef_in_template-i1.h
-  typedef Pair<T,T> pair_type;
+#include "direct.h"
+
+// Verify that a full-use of an alias of a template parameter is treated as a
+// full-use of that parameter.
+template<typename T>
+struct A {
```
kimgr, line 15: I think `A`, `B` and `C` could use some more evocative names to signal their role

tests/cxx/typedef_in_template.cc
```diff
@@ -7,51 +7,54 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "tests/cxx/typedef_in_template-d1.h"
-
-template<class T>
-class Container {
```
kimgr, line 13: One thing I really liked with this testcase was how it showed how something STL-container-like might behave. Granted, it was too simplified to be a real example, but it told a nice story.

It might be nice to add little breadcrumb comments to describe what real-life libstdc++ or libc++ constructs your example classes map to, because that makes it easier to to see how the testcase maps to the real world, and notice when they drift apart (e.g. if we get bug reports for libstdc++ template stuff in the future, it's nice to know where to look for related test coverage)

run_iwyu_tests.py
```diff
@@ -206,6 +207,7 @@ def setUp(self):
       'typedef_chain_no_follow.cc': ['.'],
       'typedef_in_template.cc': ['.'],
       'typedefs_and_resugaring.cc': ['.'],
+      'typedefs_of_template_arguments.cc': ['.'],
```
kimgr, line 210: This file is not included in the patch set?

iwyu_ast_util.cc
```diff
@@ -1299,12 +1325,20 @@ GetTplTypeResugarMapForClassNoComponentTypes(const clang::Type* type) {
     return retval;
 
   // Get the list of template args that apply to the decls.
-  const NamedDecl* decl = TypeToDeclForContent(tpl_spec_type);
-  const ClassTemplateSpecializationDecl* tpl_decl = DynCastFrom(decl);
-  if (!tpl_decl)   // probably because tpl_spec_type is a dependent type
-    return retval;
-  const TemplateArgumentList& tpl_args
-      = tpl_decl->getTemplateInstantiationArgs();
+  const NamedDecl* decl = TypeToDeclAsWritten(tpl_spec_type);
```
kimgr, line 1328: I'd like to see this move down to just before its use.

iwyu_ast_util.cc
```diff
@@ -1299,12 +1325,20 @@ GetTplTypeResugarMapForClassNoComponentTypes(const clang::Type* type) {
...
+
+  const TemplateArgument* tpl_args = tpl_spec_type->getArgs();
+  unsigned num_args = tpl_spec_type->getNumArgs();
+
+  // When this is a ClassTemplateSpecializationDecl specifically, we want to
```
kimgr, line 1333: And maybe move this comment up above the "default" case where we pull the args/num_args from the tpl_spec_type, to make it clearer that it all belongs together? Something like:
```
// Pull the template arguments out of the specialization type. If this is
// a ClassTemplateSpecializationDecl specifically, we want to
// get the arguments therefrom to correctly handle default arguments.
const TemplateArgument* tpl_args = tpl_spec_type->getArgs();
unsigned num_args = tpl_spec_type->getNumArgs();
const NamedDecl* decl = TypeToDeclAsWritten(tpl_spec_type);
if (const ClassTemplateSpecializationDecl* cls_tpl_decl = DynCastFrom(decl)) {
  const TemplateArgumentList& tpl_arg_list = cls_tpl_decl->getTemplateInstantiationArgs();
  tpl_args = tpl_arg_list.data();
  num_args = tpl_arg_list.size();
}
```

or something like that.

kimgr: I find the loop below a little confusing when it's _not_ a `ClassTemplateSpecializationDecl` -- it seems like it's just running all types against themselves. But on the whole, I don't understand what this function does, so maybe that's something I just need to work on :)

iwyu.cc
```diff
@@ -3115,45 +3128,126 @@ class InstantiatedTemplateVisitor
...
-    // template's caller.
-    ReportTypeUse(caller_loc(), actual_type);
+    return true;
+  }
+
+  // This helper is called on every use of a template argument type in an
+  // instantiated template.  Its goal is to determine whether that use should
+  // constitute a full-use by the template caller, and perform other necessary
+  // bookkeeping.
+  void AnalyzeTemplateTypeParmUse(const Type* type)
```
kimgr, line 3169: How about just `ReportTemplateTypeParmUse`?

Hmm, I see it's also registering `template_argument_aliases_`, I guess that's the "other necessary bookkeeping".

I wonder if you could make that more explicit by e.g. having callers pass in the `template_argument_aliases_` container? On balance, I'm not sure if it's an improvement, but worth playing around with.

iwyu_ast_util.cc
```diff
@@ -234,6 +234,14 @@ bool IsElaborationNode(const ASTNode* ast_node) {
   return elaborated_type && elaborated_type->getKeyword() != clang::ETK_None;
 }
 
+const ASTNode* MostElaboratedAncestor(const ASTNode* ast_node) {
```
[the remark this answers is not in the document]
kimgr: No, wait. I see now that it actually _is_ the most elaborated. I think I was confused by the comment that says `Read _past_ elaborations`. I mean, it does, but ... Yeah. I think this is the right name as-is.

iwyu.cc
```diff
@@ -2934,6 +2953,11 @@ class InstantiatedTemplateVisitor
     return ContainsKey(resugar_map_, type);
   }
 
+  bool InvolvesKnownTemplateParam(const Type* type) const {
+    return InvolvesTypeForWhich(
```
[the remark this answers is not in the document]
jbytheway: Indeed, it seems that way.

iwyu.cc
```diff
@@ -3075,31 +3100,19 @@ class InstantiatedTemplateVisitor
...
-    // Figure out how this type was actually written.  clang always
-    // canonicalizes SubstTemplateTypeParmType, losing typedef info, etc.
-    const Type* actual_type = ResugarType(type);
-    CHECK_(actual_type && "If !CanIgnoreType(), we should be resugar-able");
-
-    // TODO(csilvers): whenever we report a type use here, we want to
-    // do an iwyu check on this type (to see if sub-types are used).
+  // Check whether a use of a template parameter is a full use.
+  bool IsTemplateTypeParmUseFullUse(const Type* type)
+  {
```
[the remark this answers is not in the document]
jbytheway: I've used git-format to (hopefully) fix all of these and similar issues.

iwyu.cc
```diff
@@ -3115,45 +3128,126 @@ class InstantiatedTemplateVisitor
...
+        isa<ReferenceType>(type)) {
+      return true;
     }
 
     // If we're used in a forward-declare context (MyFunc<T>() { T* t; }),
     // or are ourselves a pointer type (MyFunc<Myclass*>()),
     // we don't need to do anything: we're fine being forward-declared.
-    if (current_ast_node()->in_forward_declare_context())
-      return Base::VisitSubstTemplateTypeParmType(type);
+    if (node->in_forward_declare_context())
```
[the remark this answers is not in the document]
jbytheway: I think it shouldn't matter.  Elaboration nodes shouldn't affect forward-declaredness.

iwyu.cc
```diff
@@ -3115,45 +3128,126 @@ class InstantiatedTemplateVisitor
...
+  // instantiated template.  Its goal is to determine whether that use should
+  // constitute a full-use by the template caller, and perform other necessary
+  // bookkeeping.
+  void AnalyzeTemplateTypeParmUse(const Type* type)
+  {
+    const ASTNode* node = MostElaboratedAncestor(current_ast_node());
+
+    // If the immediate parent node is a typedef, then register the new type as
+    // a new name for the template argument.
+    if (const TypedefNameDecl* typedef_decl
```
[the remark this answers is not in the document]
jbytheway: Fixed via git-format.

iwyu_ast_util.cc
```diff
@@ -1171,6 +1179,24 @@ const Type* RemoveSubstTemplateTypeParm(const Type* type) {
     return type;
 }
 
+bool InvolvesTypeForWhich(const Type* type,
+                          std::function<bool(const Type*)> pred) {
+  type = RemoveSubstTemplateTypeParm(type);
+  if (pred(type)) return true;
```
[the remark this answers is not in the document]
jbytheway: Fixed via git-format.

iwyu_ast_util.cc
```diff
@@ -1299,12 +1325,20 @@ GetTplTypeResugarMapForClassNoComponentTypes(const clang::Type* type) {
...
-  const TemplateArgumentList& tpl_args
-      = tpl_decl->getTemplateInstantiationArgs();
+  const NamedDecl* decl = TypeToDeclAsWritten(tpl_spec_type);
+
+  const TemplateArgument* tpl_args = tpl_spec_type->getArgs();
+  unsigned num_args = tpl_spec_type->getNumArgs();
+
+  // When this is a ClassTemplateSpecializationDecl specifically, we want to
+  // get the arguments therefrom to correctly handle default arguments.
+  const ClassTemplateSpecializationDecl* cls_tpl_decl = DynCastFrom(decl);
```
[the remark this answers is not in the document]
jbytheway: Ah, I wondered why there were two different styles.  Thanks for clarifying.

run_iwyu_tests.py
jbytheway: It was added and later removed again, but I missed this line.  Good catch.

iwyu_ast_util.cc
jbytheway: I don't really understand the function either, so this is something of the blind leading the blind...

iwyu_ast_util.cc
```diff
@@ -1299,12 +1325,20 @@ GetTplTypeResugarMapForClassNoComponentTypes(const clang::Type* type) {
     return retval;
 
   // Get the list of template args that apply to the decls.
-  const NamedDecl* decl = TypeToDeclForContent(tpl_spec_type);
```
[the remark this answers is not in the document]
jbytheway: Can I do that in a separate PR?

iwyu_ast_util.cc
jbytheway: Code rearranged.

iwyu_ast_util.cc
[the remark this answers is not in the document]
kimgr: Of couse.

iwyu.cc
```diff
@@ -1248,13 +1248,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   using Base::PrintableCurrentLoc;
   using Base::current_ast_node;
 
+  enum class IgnoreType {
```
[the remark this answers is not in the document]
jbytheway: Renamed as suggested.

iwyu_ast_util.cc
```diff
@@ -1171,6 +1179,24 @@ const Type* RemoveSubstTemplateTypeParm(const Type* type) {
     return type;
 }
 
+bool InvolvesTypeForWhich(const Type* type,
```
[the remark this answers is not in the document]
jbytheway: Not sure that's better.  I feel like that's suggestive of something related to source locations.

iwyu_ast_util.cc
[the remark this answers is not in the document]
jbytheway: Code changed accordingly (one of them had to be a `dyn_cast_or_null`, and indeed they no longer fit on one line).

tests/cxx/typedef_in_template.cc
jbytheway: I've restored the deleted test cases.

tests/cxx/typedef_in_template.cc
jbytheway: I've restored the deleted test cases.

tests/cxx/typedef_in_template.cc
jbytheway: Renamed.

tests/cxx/typedef_in_template.cc
jbytheway: I've added a comment here.  There's less direct correlation than in the previous testcase, but it might help explain what's going on.

iwyu.cc
jbytheway: I don't think it helps.  Etiehr way it's just member functions using a member variable.  And moving it to the callers means one more reference to that variable.

jbytheway:
OK, I think I've actioned or responded to all your comments.  I've also rebased the commit history into something which is hopefully somewhat logical and comprehensible.

Reinstating the old test cases actually revealed a case which is not currently handled correctly, but it's a non-full-use being treated as a full-use, which is less bad than the opposite, so I've left it as is, with a TODO in the test case for now.

Ready for you to take another look now.

tests/cxx/typedef_in_template.cc
```diff
@@ -24,34 +25,87 @@ class Container {
...
+// Verify that a full-use of an alias of a template parameter is treated as a
+// full-use of that parameter.
+template <typename T>
+struct UsesAliasedParameter {
+  using TAlias = T;
+  TAlias t;
+};
+
+// IWYU: IndirectClass is...*indirect.h
+// IWYU: IndirectClass needs a declaration
```
kimgr, line 69: At first I found it suspicious that there was both a full- and a forward-decl-use here, but I think it kind-of makes sense. The forward-decl use is for the `using TAlias = T;`, and the full use is for the `TAlias t;`?

I guess I could argue that the forward-decl use does not belong here, because it's an internal detail of the template, but it doesn't really do any harm either, since it will be short-circuited by the full use.

tests/cxx/alias_template.cc
```diff
@@ -33,6 +38,21 @@ AliasChain2<IndirectClass> aliasChain;
...
 
+// Test the used class being nested deeper in the alias
+template <typename T>
+using AliasNested = FullUseTemplateArg<FullUseTemplateArg<T>>;
+
+// IWYU: IndirectClass needs a declaration
+// IWYU: IndirectClass is...*indirect.h
+AliasNested<IndirectClass> aliasNested;
+
+// Test the used class being nested deeper in the alias
```
kimgr, line 49: I'm not sure I understand how this is different from `AliasNested` above.
The template argument is used in an evaluated context, as opposed to the unevaluated `sizeof` -- but both lead to the same IWYU result?
I wonder if the `sizeof` template should be renamed to clarify that, or something. Now it just looks like there's two identical tests after each other.

tests/cxx/typedef_in_template.cc
kimgr: After having seen a number of other template testcases during review, I think I now understand that the forward-decl is for the use of the template argument. Nothing to see here.

iwyu_ast_util.h
```diff
@@ -376,6 +376,10 @@ class CurrentASTNodeUpdater {
 // uses ElaboratedType for namespaces ('ns::Foo myvar').
 bool IsElaborationNode(const ASTNode* ast_node);
 
+// Walk up to parents of the given node so long as each parent is an
+// elaboration node (in the sense of IsElaborationNode).
```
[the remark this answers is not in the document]
jbytheway: The case I originally actually needed to catch was the `struct X` case.  But it makes sense for the other type of elaboration also.  I've expanded the doc comment.

iwyu.cc
```diff
@@ -3115,45 +3125,137 @@ class InstantiatedTemplateVisitor
...
+      if (const auto* cts_decl =
+              dyn_cast<ClassTemplateSpecializationDecl>(decl)) {
+        if (ContainsKey(traversed_decls_, decl))
+          return;  // avoid recursion & repetition
+        traversed_decls_.insert(decl);
+
+        VERRS(6)
+            << "Recursively traversing " << PrintableDecl(cts_decl)
+            << " which was full-used and involves a known template param\n";
+        TraverseDecl(const_cast<ClassTemplateSpecializationDecl*>(cts_decl));
```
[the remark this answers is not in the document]
jbytheway: There is precedent for doing this already.  There's a call to `TraverseDecl` in `TraverseDataAndTypeMembersOfClassHelper`.

iwyu_ast_util.h
```diff
@@ -694,6 +698,11 @@ bool IsClassType(const clang::Type* type);
 // However, vector<T> is *not* converted to vector<int>.
 const clang::Type* RemoveSubstTemplateTypeParm(const clang::Type* type);
 
+// Returns true if any type involved (recursively examinging template
```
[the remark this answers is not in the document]
jbytheway: Fixed.

iwyu.cc
```diff
@@ -3115,45 +3125,137 @@ class InstantiatedTemplateVisitor
...
     return Base::VisitSubstTemplateTypeParmType(type);
   }
 
+  bool VisitTypedefType(clang::TypedefType* type) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+
+    if (const Type* template_arg_type =
+            GetOrDefault(template_argument_aliases_, type, nullptr)) {
+      AnalyzeTemplateTypeParmUse(template_arg_type);
```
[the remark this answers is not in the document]
jbytheway: Added a comment.

tests/cxx/typedef_in_template.cc
```diff
@@ -7,6 +7,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "direct.h"
```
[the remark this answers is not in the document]
jbytheway: Fixed.

tests/cxx/alias_template.cc
jbytheway: They are roughly identical; I suspect they are redundant.  But I felt like it didn't hurt to have an extra test case.  Certainly it's important to test that both `sizeof` and regular use-as-a-variable count as full uses, and the deeper nesting was also important to test.  There was no particular reason to choose one over the other for the deeper nesting test, so I just did both.

I've renamed the templates for extra clarity and deleted a redundant comment.

jbytheway:
> > Fix another alias-in-template issue
> 
> I wish this commit message had a small example with real code. The fix is so small (good thing) that it's hard to understand where and how it has effect (bad thing) :)

You have good instincts.  This is actually the most dubious change, and isn't always correct.  If you're concerned, we can drop this commit and the remainder still works.  But I think it's a step in the right direction.  In any case, I expanded the commit message to clarify.

I think I've addressed your other comments.  Over to you again.

tests/cxx/sizeof_in_template_arg.cc
```diff
@@ -0,0 +1,41 @@
+//===--- sizeof_in_template_arg.cc - test input file for iwyu -------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "direct.h"
```
kimgr, line 10: `tests/cxx/direct.h`

iwyu.cc
```diff
@@ -2897,10 +2903,22 @@ class InstantiatedTemplateVisitor
 
   string GetSymbolAnnotation() const override { return " in tpl"; }
 
-  bool CanIgnoreType(const Type* type) const override {
-    if (!IsTypeInteresting(type) || !IsKnownTemplateParam(type))
+  bool CanIgnoreType(const Type* type, IgnoreKind ignore_type =
```
kimgr, line 2906: `ignore_type` -> `ignore_kind`

iwyu.cc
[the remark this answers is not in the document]
kimgr: OK, that makes my heart lighter :)

tests/cxx/sizeof_in_template_arg.cc
jbytheway: Fixed.

iwyu.cc
jbytheway: Fixed.

jbytheway:
OK, nits fixed.

kimgr:
Thank you!

Issue #777, opened 2020-03-22
IWYU should not recommend <new> when using size-aware operator delete


adembo:
According to https://en.cppreference.com/w/cpp/memory/new/operator_delete, the C++14 size-aware operator delete functions are in the set of functions that are implicitly declared in each translation unit even if the \<new\> header is not included. And yet, if you use them and run IWYU, it recommends including \<new\>.

The incorrect recommendation appears to be rooted in iwyu_ast_util.cc's IsDefaultNewOrDelete function, which includes this code:
```
// Placement-new/delete has 2 args, second is void*.  The only other
// 2-arg overloads of new/delete in <new> take a const nothrow_t&.
if (decl->getNumParams() == 2 &&
    !decl->getParamDecl(1)->getType().isConstQualified())
  return false;
```

For the size-aware operator delete functions, there are two parameters and the second is a (non-const) size_t, which causes IWYU to return false here and recommend the inclusion of \<new\>.

My guess is that this function was written pre-C++14 and hasn't been revised since. I've been working on a patch to update it; will submit a separate PR.

kimgr:
We did lots of research in the context of PR #778, and learned the following:

* `new` and `delete` have two family trees: one as overloadable operators, and one as expressions, whose semantics are built into the language.
* the semantics of `new X;` is to first call `operator new` for X, whether built-in or user-defined, to allocate memory, and then execute `X`'s constructor.
* `operator new(size_t)` is declared in `<new>`, but its declaration is made implicitly available by the compiler, as well as a definition, if none is provided by the user. This is called a 'replaceable allocation function'.
* `operator new(size_t, void*)` is the non-allocating allocation function for placement syntax, i.e. `new (<some pointer>) X;` -- again, it calls the (noop) allocation function, and then executes `X`'s constructor.
* `operator new(size_t, void*)` is _defined_ in `<new>`, and so can't be replaced/overloaded.
* There are also placement-like operators provided to enable syntax `new (std::nothrow) X;` and `new (std::align_val_t(32)) X;`, but these _are_ replaceable (so that the user can provide custom allocation with alignment no-throw guarantees)
* Finally, while you can't _replace_ the global placement allocation function, you can overload it for more specific types, e.g. `void* operator new(size_t sz, Arena& arena);`. Used like this `new (arena) X;`, it could delegate allocation to an arena that's more efficient than built-in `malloc`.
[...]

PR #778, opened 2020-03-23
Update implicitly defined operator new/delete check for newer C++ standards

adembo:
IsDefaultNewOrDelete attempts to identify when a particular usage of
operator new or operator delete is one of a small set of functions that are
implicitly defined in every translation unit by the compiler. Unfortunately,
it doesn't appear to honor newer implicitly defined variants added in C++14
and C++17.

This patch updates it accordingly, and adds a test that fails without the
patch but passes with it.

Fixes tracking issue #777.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -20,7 +20,6 @@
 #include "iwyu_path_util.h"
 #include "iwyu_port.h"  // for CHECK_
 #include "iwyu_stl_util.h"
-#include "iwyu_string_util.h"
 #include "iwyu_verrs.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/Support/Casting.h"
@@ -40,6 +39,7 @@
 #include "clang/AST/Type.h"
 #include "clang/AST/TypeLoc.h"
 #include "clang/Basic/Builtins.h"
+#include "clang/Basic/OperatorKinds.h"
 #include "clang/Basic/SourceLocation.h"
 #include "clang/Basic/SourceManager.h"
 
@@ -946,18 +946,58 @@ bool IsDefaultNewOrDelete(const FunctionDecl* decl,
       !IsBuiltinFile(GetFileEntry(decl)))
     return false;
 
-  const string decl_name = decl->getNameAsString();
-  if (!StartsWith(decl_name, "operator new") &&
-      !StartsWith(decl_name, "operator delete"))
+  // The following variants of operator new[1] are implicitly defined in every
+  // translation unit and should not require including <new>.
+  //
+  // void* operator new  ( std::size_t count );
+  // void* operator new[]( std::size_t count );
+  // void* operator new  ( std::size_t count, std::align_val_t al ); (since C++17)
+  // void* operator new[]( std::size_t count, std::align_val_t al ); (since C++17)
+  //
+  // Likewise, the following variants of operator delete[2] are implicitly
+  // defined in every translation unit and should not require including <new>.
+  //
+  // void operator delete  ( void* ptr ) throw(); (until C++11)
+  // void operator delete  ( void* ptr ) noexcept; (since C++11)
+  // void operator delete[]( void* ptr ) throw(); (until C++11)
+  // void operator delete[]( void* ptr ) noexcept; (since C++11)
+  // void operator delete  ( void* ptr, std::align_val_t al ) noexcept; (since C++17)
+  // void operator delete[]( void* ptr, std::align_val_t al ) noexcept; (since C++17)
+  // void operator delete  ( void* ptr, std::size_t sz ) noexcept; (since C++14)
+  // void operator delete[]( void* ptr, std::size_t sz ) noexcept; (since C++14)
+  // void operator delete  ( void* ptr, std::size_t sz,
+  //                         std::align_val_t al ) noexcept; (since C++17)
+  // void operator delete[]( void* ptr, std::size_t sz,
+  //                         std::align_val_t al ) noexcept; (since C++17)
+  // void operator delete  ( void* ptr, const std::nothrow_t& tag ) throw(); (until C++11)
+  // void operator delete  ( void* ptr, const std::nothrow_t& tag ) noexcept; (since C++11)
+  // void operator delete[]( void* ptr, const std::nothrow_t& tag ) throw(); (until C++11)
+  // void operator delete[]( void* ptr, const std::nothrow_t& tag ) noexcept; (since C++11)
+  //
+  // FunctionDecl::isReplaceableGlobalAllocationFunction() does a great job of
+  // parsing the variants of operator new/delete and returning true if a
+  // particular variant is "replaceable". It just so happens that all implicitly
+  // defined variants are also replaceable. However, there are four variants of
+  // operator new that are replaceable but not implicitly defined; we must look
+  // for these explicitly and filter them out.
+  //
+  // 1. https://en.cppreference.com/w/cpp/memory/new/operator_new
+  // 2. https://en.cppreference.com/w/cpp/memory/new/operator_delete
+  bool is_nothrow = false;
+  if (!decl->isReplaceableGlobalAllocationFunction(nullptr, &is_nothrow))
     return false;
 
-  // Placement-new/delete has 2 args, second is void*.  The only other
-  // 2-arg overloads of new/delete in <new> take a const nothrow_t&.
-  if (decl->getNumParams() == 2 &&
-      !decl->getParamDecl(1)->getType().isConstQualified())
-    return false;
-
-  return true;
+  // These variants are replaceable but not implicitly defined.
+  //
+  // void* operator new  ( std::size_t count, const std::nothrow_t& tag );
+  // void* operator new[]( std::size_t count, const std::nothrow_t& tag );
+  // void* operator new  ( std::size_t count,
+  //                       std::align_val_t al, const std::nothrow_t& ); (since C++17)
+  // void* operator new[]( std::size_t count,
+  //                       std::align_val_t al, const std::nothrow_t& ); (since C++17)
+  return decl->getDeclName().getCXXOverloadedOperator() == clang::OO_Delete ||
+         decl->getDeclName().getCXXOverloadedOperator() == clang::OO_Array_Delete ||
+         !is_nothrow;
 }
 
 bool IsFriendDecl(const Decl* decl) {
--- a/tests/cxx/builtins_new_included_cxx14.cc
+++ b/tests/cxx/builtins_new_included_cxx14.cc
@@ -1,0 +1,31 @@
+//===--- builtins_new_included_cxx14.cc - test input file for iwyu --------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Test that iwyu suggests the include for <new> be removed if only
+// built-in functions are used, C++14 edition (i.e. with -fsized-deallocation).
+
+#include <new>
+
+void foo() {
+  char* ch = new char;
+  delete ch;
+  int* int_array = new int[10];
+  delete[] int_array;
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/builtins_new_included_cxx14.cc should add these lines:
+
+tests/cxx/builtins_new_included_cxx14.cc should remove these lines:
+- #include <new>  // lines XX-XX
+
+The full include-list for tests/cxx/builtins_new_included_cxx14.cc:
+
+***** IWYU_SUMMARY */
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -100,6 +100,8 @@ class OneIwyuTest(unittest.TestCase):
     clang_flags_map = {
       'alias_template.cc': ['-std=c++11'],
       'auto_type_within_template.cc': ['-std=c++11'],
+      'builtins_new_included_cxx14.cc': ['-std=c++14', '-fsized-deallocation'],
+      'builtins_new_included_cxx17.cc': ['-std=c++17'],
       # MSVC targets need to explicitly enable exceptions, so we do it for all.
       'catch.cc': ['-fcxx-exceptions', '-fexceptions'],
       'clmode.cc': ['--driver-mode=cl', '/GF', '/Os', '/W2'],
--- a/tests/cxx/builtins_new_included_cxx17.cc
+++ b/tests/cxx/builtins_new_included_cxx17.cc
@@ -1,0 +1,35 @@
+//===--- builtins_new_included_cxx17.cc - test input file for iwyu --------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Test that iwyu suggests the include for <new> be removed if only
+// built-in functions are used, C++17 edition (i.e. with aligned allocations).
+
+#include <new>
+
+void foo() {
+  class alignas(32) test_struct {
+    float value[8];
+  };
+
+  test_struct* t = new test_struct;
+  delete t;
+  test_struct* t_array = new test_struct[10];
+  delete[] t_array;
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/builtins_new_included_cxx17.cc should add these lines:
+
+tests/cxx/builtins_new_included_cxx17.cc should remove these lines:
+- #include <new>  // lines XX-XX
+
+The full include-list for tests/cxx/builtins_new_included_cxx17.cc:
+
+***** IWYU_SUMMARY */


kimgr:
Thanks for this update! Looks like `IsDefaultNewOrDelete` really needed some attention in the presence of C++14 and C++17.

There's a lot of stuff going on here now... I'm a little concerned that the C++14 and 17 rules are quietly seeping into analysis of earlier standards too -- does this need to be gated behind a langopts check or something?

Since you're referring to `FunctionDecl::isReplaceableGlobalAllocationFunction`, couldn't we use that and `FunctionDecl::isReservedGlobalPlacementOperator` to implement `IsDefaultNewOrDelete`? You have some commentary on `no_throw_t`, but that seems to be covered by `FunctionDecl::isRGAF`...?

Also, I feel like there should be testing for all the variants listed in the source code.

adembo:
> I'm a little concerned that the C++14 and 17 rules are quietly seeping into analysis of earlier standards too -- does this need to be gated behind a langopts check or something?

I ended up reimplementing this based on `FunctionDecl::isReplaceableGlobalAllocationFunction` as you suggested. The implementation is much cleaner, and I don't think there's a need for a langopts check (as the IWYU piece isn't assuming anything about C++ language support.

> Since you're referring to FunctionDecl::isReplaceableGlobalAllocationFunction, couldn't we use that and FunctionDecl::isReservedGlobalPlacementOperator to implement IsDefaultNewOrDelete? You have some commentary on no_throw_t, but that seems to be covered by FunctionDecl::isRGAF...?

OK, I tried this out, and it's not so bad looking. Didn't need to use isRGPO though; did I miss something?

> Also, I feel like there should be testing for all the variants listed in the source code.

Agreed. Right now we have coverage for the basic variants as well as sized-delete (via the cxx14 test). I can add coverage for the aligned allocations, but I don't know how to cover the nothrow_t ones. Thoughts?

iwyu_ast_util.cc
```diff
@@ -951,13 +953,57 @@ bool IsDefaultNewOrDelete(const FunctionDecl* decl,
       !StartsWith(decl_name, "operator delete"))
     return false;
 
-  // Placement-new/delete has 2 args, second is void*.  The only other
-  // 2-arg overloads of new/delete in <new> take a const nothrow_t&.
-  if (decl->getNumParams() == 2 &&
-      !decl->getParamDecl(1)->getType().isConstQualified())
+  // The following variants of operator new[1] are implicitly defined in every
```
kimgr, line 956: I wonder if the initial string-checking at the beginning of this function is even necessary any more?

iwyu_ast_util.cc
```diff
@@ -951,13 +953,57 @@ bool IsDefaultNewOrDelete(const FunctionDecl* decl,
...
+  // particular variant is "replaceable". It just so happens that all implicitly
+  // defined variants are also replaceable. However, there are four variants of
+  // operator new that are replaceable but not implicitly defined; we must look
+  // for these explicitly and filter them out.
+  //
+  // 1. https://en.cppreference.com/w/cpp/memory/new/operator_new
+  // 2. https://en.cppreference.com/w/cpp/memory/new/operator_delete
+  Optional<unsigned> alignmentParam;
+  bool isNoThrow = false;
+  if (!decl->isReplaceableGlobalAllocationFunction(&alignmentParam, &isNoThrow))
```
kimgr, line 995: Looks like you can pass `nullptr` for `AlignmentParam`, so you don't need the dummy argument.

iwyu_ast_util.cc
```diff
@@ -951,13 +953,57 @@ bool IsDefaultNewOrDelete(const FunctionDecl* decl,
...
+  // parsing the variants of operator new/delete and returning true if a
+  // particular variant is "replaceable". It just so happens that all implicitly
+  // defined variants are also replaceable. However, there are four variants of
+  // operator new that are replaceable but not implicitly defined; we must look
+  // for these explicitly and filter them out.
+  //
+  // 1. https://en.cppreference.com/w/cpp/memory/new/operator_new
+  // 2. https://en.cppreference.com/w/cpp/memory/new/operator_delete
+  Optional<unsigned> alignmentParam;
+  bool isNoThrow = false;
```
kimgr, line 994: IWYU uses `hacker_case`, so this should be `is_nothrow`.

iwyu_ast_util.cc
```diff
@@ -951,13 +953,57 @@ bool IsDefaultNewOrDelete(const FunctionDecl* decl,
...
-  return true;
+  // These variants are replaceable but not implicitly defined.
+  //
+  // void* operator new  ( std::size_t count, const std::nothrow_t& tag );
+  // void* operator new[]( std::size_t count, const std::nothrow_t& tag );
+  // void* operator new  ( std::size_t count,
+  //                       std::align_val_t al, const std::nothrow_t& ); (since C++17)
+  // void* operator new[]( std::size_t count,
+  //                       std::align_val_t al, const std::nothrow_t& ); (since C++17)
+  return StartsWith(decl_name, "operator delete") || !isNoThrow;
```
kimgr, line 1006: I see `isReplaceable...` uses `(getDeclName().getNameKind() != DeclarationName::CXXOperatorName)` and `getDeclName().getCXXOverloadedOperator()` to use enums instead of string comparisons. That does seem a little nicer, can you see if that's applicable here too, outside the function?

tests/cxx/builtins_new_included_cxx14.cc
```diff
@@ -0,0 +1,31 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+// Test that iwyu suggests the include for <new> be removed if only
+// built-in functions are used, C++14 edition (i.e. with -fsized-deallocation).
+
+#include <new>
+
+void foo() {
+  char* ch = new char;
```
kimgr, line 16: It would be nice if this file demonstrated the use of sized deallocation.

iwyu_ast_util.cc
adembo: Probably not, no. I've removed it.

iwyu_ast_util.cc
adembo: Thanks for the catch. Fixed.

iwyu_ast_util.cc
adembo: Whoops. Too much staring at clang code. Fixed.

iwyu_ast_util.cc
adembo: Yeah I can do that, though I don't think we need the `getDeclName().getNameKind()` check, as having returned true from `isReplaceable...`, we're guaranteed that this is a `CXXOperatorName`.

tests/cxx/builtins_new_included_cxx14.cc
adembo: Hmm, I thought the compiler does so transparently by replacing the call site with the sized variant of `operator delete`.

FWIW, I observed that if I reverted my patch to `iwyu_ast_util.cc` and ran just the new test, it fails as IWYU sees the two-arg variant of `operator delete` with a non-const second argument (i.e. the `std::size_t`) and retains the recommendation to include `<new>`.

adembo:
> So, isn't the case that _only_ placement new requires `#include <new>`? I wonder if there should be a testcase for that too, now it's just phrased as a negative testcase that `<new>` is removed for lots of other allocation functions.

Isn't this already covered in `badinc.cc`?
```
  // Make sure we only report an iwyu for <new> because of placement-new.
  // We also need to check the argument to new.
  int* newed_int = new int;
  // IWYU: operator new is...*<new>
  new(newed_int) int(4);
```

tests/cxx/builtins_new_included_cxx14.cc
kimgr: @adembo Oh, nice! I'm not sure what my perception of sized deallocation was, but that makes sense.

Hmm, OK... But this introduces another annoying thing, I guess -- IWYU's detection of default-new-or-delete now depends on language mode. So if you run IWYU with `-fsized-deallocation`, it will remove an include that's required for C++11 builds (assuming for example a portable library for C++11 to C++14).

I guess the answer to that is to always run IWYU with the lowest supported language version. That should go in our docs at some point.

I've convinced myself that this is the right solution, thanks for listening :)

iwyu_ast_util.cc
```diff
@@ -946,18 +946,58 @@ bool IsDefaultNewOrDelete(const FunctionDecl* decl,
       !IsBuiltinFile(GetFileEntry(decl)))
     return false;
 
-  const string decl_name = decl->getNameAsString();
```
kimgr, line 949: Can you even get rid of the `decl_loc_as_quoted_include` parameter, as well as the code looking at declaration location spelling? Or does `isReplaceableGlobalAllocationFunction` return true for actual operator new overloads in user code as well? It looks like it might.

We want this predicate to only return true if the decl after overload resolution actually refers to a builtin new/delete, right?

I think the code as-is is right, but it would be nice if we _could_ get rid of the string munging on filenames.

kimgr:
> > So, isn't the case that _only_ placement new requires `#include <new>`? I wonder if there should be a testcase for that too, now it's just phrased as a negative testcase that `<new>` is removed for lots of other allocation functions.
> 
> Isn't this already covered in `badinc.cc`?
> 
> ```
>   // Make sure we only report an iwyu for <new> because of placement-new.
>   // We also need to check the argument to new.
>   int* newed_int = new int;
>   // IWYU: operator new is...*<new>
>   new(newed_int) int(4);
> ```

Ah, I've given up on trying to find things in badinc.cc. Maybe this is the push I need to move that testcase into a dedicated test module.

tests/cxx/builtins_new_included_cxx14.cc
adembo: Yeah it's not perfect, but I'm honestly not sure how IWYU can avoid stuff like that writ large. For example, if your portable library built with one header on one architecture and a second on another, you may end up with different IWYU results depending on which architecture you run it.

This is actually a problem we're having with IWYU in our large codebase, and we work around it via prolific use of `// IWYU pragma: keep`.

iwyu_ast_util.cc
adembo: In general I don't understand how these user-defined overloads work w.r.t. `<new>`: if a user-defined overload uses a weird assortment of expected parameters, how is `<new>` supposed to provide a declaration for it? I looked at `<new>` on my system and it doesn't look like it declares anything for user-defined overloads.

Anyway, I don't understand this particular aspect of C++ (or IWYU) well enough to feel comfortable removing the code you mentioned. But, since you're the expert, I'll defer to your judgment and will remove it if you think it's safe to do so.

kimgr: Off the top of my head, I think it would be the other way around: user-defined overloads need their own declaring header instead of `<new>`.

I realized it didn't really know what was going on here, especially in the light of C++14 and -17 additions, so I read up a bit and ran some experiments, and I think I have a better understanding now, if not entirely complete :)

These two SO answers both make an illuminating argument:
* https://stackoverflow.com/a/6784138/96963
* https://stackoverflow.com/a/9691364/96963
* and an accompanying rant: http://www.scs.stanford.edu/~dm/home/papers/c++-new.html

As I understand it, they says that `operator new` and `new`-expressions are two different things. The former is a function responsible for memory allocation, the latter is a language construct where the compiler generates code to call `operator new`, and then run the constructor on the allocated memory.

Because the language always does both things -- call the allocator, and run the constructor -- a placement-new allocation function becomes necessary, which just roundtrips the pointer that was passed in.

The placement-new (and -delete) functions are _defined_ in `<new>` and thereby aren't recognized by the compiler unless `<new>` is included, and they're also not replaceable. This seems to me to be a poor choice on behalf of the language designers, but I think there's historically been a tendency to avoid magic imbued-by-the-compiler-declarations if a library declaration is possible.

Of course there are subtle exceptions to this rule that placement allocator functions are not replaceable -- they can be overloaded as long as the "placement arguments" are anything other than a single `void*`. Such placement functions can be used (with limited convenience and safety) for arena allocators, etc. It looks like the `std::align_val_t` and `std::nothrow_t` overloads are examples of such special placement functions that happen to have defined semantics.

Anyway, all my research so far indicates that only the "raw" placement-new functions with a `void*` placement parameter are non-replaceable. They also require `<new>` to be included to make that dummy implementation available.

The other placement-like allocation functions (for alignment and nothrow) both use types that are exported from `<new>`, and so IWYU should track these types back to `<new>` without additional handling.

And finally, the remaining allocation functions are somehow magically built into the language, and no header is required. A surprising example is that this compiles without including `<new>`:
```
class Foo {};

int main() {
    // Allocates memory, but does not run a constructor.
    // Surprising because the declaration of void* operator new(size_t) is in <new>.
    Foo* a = (Foo*)::operator new(sizeof(Foo));
    return 0;
}
```

kimgr: I'm going to try and capture this in one or more testcases extracted from badinc.cc, so I can play around with removing the is-declaration-in-header-`<new>` guard clause there.

kimgr:
Please take a look at this, I think it builds a better foundation for the testing here: https://github.com/include-what-you-use/include-what-you-use/pull/782/files

Not sure if it's even necessary to duplicate the test across C++03, 14 and 17. There are some quirks around sized allocation that we could try and trigger, but I don't think it's necessarily worth it, now that I understand more of the mechanics.

kimgr:
I had a crazy idea: https://github.com/kimgr/include-what-you-use/commits/kimgr-collapse-replaceable

Looks like based on these testcases, we can just replace `IsDefaultNewOrDelete` with `isReplaceableGlobalAllocationFunction`, and everything pans out nicely. Let me know what you think.

adembo:
So if I'm understanding your crazy idea correctly:

The file location checks are, as you suspected, not necessary and can be removed.

The `std::nothrow_t` variants of `operator new` that are replaceable but not implicitly defined aren't worth special-casing, because to actually define `std::nothrow_t` you're gonna need to include `<new>` anyway. Are there any circumstances where the compiler would replace a bog-standard `new` expression with `operator new(std::size_t count, const std::nothrow_t& tag)`, such that you wouldn't expect to see `<new>` included? I don't see any evidence of this actually happening; intuitively I might expect to see it when calling `new` in a destructor (i.e. where one should not throw an exception). The key is: the code winds up with a call to a non-throwing variant of `operator new` despite the code author never having called that, and never having interacted with or defined `std::nothrow_t`.

It's worth mentioning that this _does_ happen with `std::align_val_t`: the compiler detects that the `new` allocation requires an alignment greater than `__STDCPP_DEFAULT_NEW_ALIGNMENT__` and converts the expression into one that uses `std::align_val_t`. Thus, we end up using the alignment-based variants of `operator new` without having interacted with `std::align_val_t` at all.

Anyway, I'm happy if the tests pass and the coverage is good, and it looks like it is.

kimgr:
> The file location checks are, as you suspected, not necessary and can be removed.

Right.

> The std::nothrow_t variants of operator new that are replaceable but not implicitly defined
> aren't worth special-casing, because to actually define std::nothrow_t you're gonna need
> to include <new> anyway.

Exactly. We don't need to trick IWYU into requiring `<new>` for `operator new`, when it's already required for `std::nothrow`. I'm not sure, though, if it would be prudent to require it for `operator new` itself, not that it makes much different in practice.

>  Are there any circumstances where the compiler would replace a bog-standard new expression
> with operator new(std::size_t count, const std::nothrow_t& tag), such that you wouldn't expect
> to see <new> included? 

Not that I've found. You could imagine some `-fnothrow-new` flag to control the default, but I haven't found a flag like this. I tried this snippet to try and capture the example you gave:
```
void foo() noexcept {
    new int;
}
```
but it generates code to call normal `operator new` sans `nothrow`.

> It's worth mentioning that this does happen with std::align_val_t: the compiler detects that the
> new allocation requires an alignment greater than __STDCPP_DEFAULT_NEW_ALIGNMENT__
> and converts the expression into one that uses std::align_val_t. Thus, we end up using the 
> alignment-based variants of operator new without having interacted with std::align_val_t at all.

Right, see the `ImplicitAlignedAllocation` in `operator_new.cc` vs `ExplicitAlignedAllocation` in `placement_new.cc` (in a couple commits preceding the `HACK` one).

When the implicit aligned allocation happens, everything is synthesized behind the scenes and no `<new>` is required.

If you're OK with it, I can try and push a PR with all these test improvements and the destructuring fix of just using `isReplaceableGlobalAllocationFunction` as a whole. What's missing compared to your PR:

* No longer any restricted testing of `operator new` in C++03, 11 or 14
* No longer any testing of `-fsized-allocation`

But since behavior is the same for the base functionality under all standard versions 03-17, I think it's OK to just test with `-std=c++17` to be able to cover explicit aligned allocation.

Not sure about `-fsized-allocation`. Based on your testcase, it doesn't look like it makes a difference for `<new>` vs not, right? If that's the case, I think we can skip it.

kimgr:
Ugh. https://travis-ci.org/github/kimgr/include-what-you-use/builds/674173294

The libstdc++ in our Travis environment is too old to support `std::align_val_t`. May want to reconsider that.

adembo:
Thanks for the detailed explanations.

> If you're OK with it, I can try and push a PR with all these test improvements and the destructuring fix of just using `isReplaceableGlobalAllocationFunction` as a whole. What's missing compared to your PR:
> 
> * No longer any restricted testing of `operator new` in C++03, 11 or 14
> * No longer any testing of `-fsized-allocation`
> 
> But since behavior is the same for the base functionality under all standard versions 03-17, I think it's OK to just test with `-std=c++17` to be able to cover explicit aligned allocation.

I don't have a horse in this race provided IWYU does the right thing for sized-deallocation (i.e. either `-std=c++14` or `-fsized-deallocation`), and I think it will.

> Not sure about `-fsized-allocation`. Based on your testcase, it doesn't look like it makes a difference for `<new>` vs not, right? If that's the case, I think we can skip it.

`-fsized-deallocation` is actually enabled by default for `-std=c++14` or above (see gcc manpage), so you're getting the same coverage if your test runs with `-std=c++17`. In my project we use `-fsized-deallocation` because we're still on C++11, and that combination is apparently legal.

> The libstdc++ in our Travis environment is too old to support std::align_val_t. May want to reconsider that.

Could you conditionalize the tests such that either `-std=c++17` or `-std=c++14` is passed in, depending on what the host compiler supports, and the tests will pass either way? Then when the Travis environment is upgraded, you could hardcode the usage of `-std=c++17`.

kimgr:
We have no mechanisms in the testing infra to run tests conditionally, or evaluate parts of a testcase conditionally. I'm hoping we can switch to `lit` at some point in the future to help with stuff like this.

kimgr:
@adembo This is no longer necessary, right?

adembo:
> @adembo This is no longer necessary, right?

Right, this has been rendered obsolete by #784.

kimgr:
Thanks! I wasn't sure anymore :o)

PR #781, opened 2020-03-26
Do not assume CMAKE_PREFIX_PATH is LLVM_ROOT

sethrj:
It is not atypical for a user's CMake prefix path to include multiple paths. (This is the case when installing a package with Spack, for example.) Instead of assuming that `CMAKE_PREFIX_PATH` is identically `LLVM_ROOT`, use the LLVM prefix path defined by `find_package(LLVM)`.

See https://github.com/include-what-you-use/include-what-you-use/pull/730


kimgr:
Thank you! It didn't occur to me that `CMAKE_PREFIX_PATH` is a search path, I've only ever used it with a single entry.

Can I be super-picky and ask you to split this into two commits -- one for replacing `CMAKE_PREFIX_PATH` with `LLVM_INSTALL_PREFIX` (with quoting), and one for adding quotes everywhere else?

The `LLVM_INSTALL_PREFIX` change currently drowns in all the other changes.

sethrj:
No problem!

kimgr:
Thank you!

PR #784, opened 2020-04-13
Modern operator new handled more correctly

kimgr:
This builds on @adembo's work in PR #778.

First improves testing of both placement new and regular operator new, and then simplifies detection of placement new so that sized deallocation and aligned allocation/deallocation no longer require `<new>`.


adembo:
Looks great, but how did you get this to pass with `-std=c++17` in Travis?

kimgr:
Upgraded the environment from Ubuntu 16.04 to 18.04. I'm not above cheating, occasionally 🙂

PR #803, opened 2020-06-06
Report use of TemplateSpecializationType

### LINKED ISSUE #801: template specialization in header is mishandled
Given the following input, IWYU produces an incorrect output:

**main.cpp**:
```C++
struct Foo
{
	static constexpr int num = 10;
};

#include "NumGetter.h"

#include "IntegerWrapperFoo.h"

int dummy()
{
	NumGetter<Foo> x;
	return x.GetNum();
}
```

**NumGetter.h**:
```C++
template<typename T>
struct False
{
	static constexpr bool value = false;
};

template<typename T>
struct IntegerWrapper
{};

template<typename T>
struct NumGetter
{
	int GetNum()
	{
		return IntegerWrapper<T>::num;
	}
};
```

**IntegerWrapperFoo.h**:
```C++
template<>
struct IntegerWrapper<Foo>
{
	static constexpr int num = Foo::num;
};
```

**IWYU output**:
```text
>include-what-you-use.exe main.cpp

- #include "IntegerWrapperFoo.h"  // lines 8-8

#include "NumGetter.h"  // for NumGetter
---
```

I am on the `master` branch of IWYU at commit `a8cd91a62beae51379109f194d7a6f988fa4cc2a` from May 4 2020.

shivanshu3:
This change aims to fix #801

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3044,6 +3044,8 @@ class InstantiatedTemplateVisitor
 
   bool TraverseTemplateSpecializationTypeLoc(
       TemplateSpecializationTypeLoc typeloc) {
+    if (!CanIgnoreCurrentASTNode())
+      ReportTypeUse({}, typeloc.getTypePtr());
     if (!Base::TraverseTemplateSpecializationTypeLoc(typeloc))  return false;
     return TraverseTemplateSpecializationTypeHelper(typeloc.getTypePtr());
   }


kimgr:
> Let me know what you think about this change. I applied this change on our internal fork of
> IWYU and it seems to work well and fixes the issue I was seeing originally.
>
> If you think I'm on the right track, I will add a test and also look at some test failures.
> 22 tests were failing before my change, and 28 are failing after this change.

It's hard to judge in isolation from the test results, and for some reason the Travis integration with GitHub pull requests seems to have broken. I found the build on travis-ci.org eventually.

Most of the failures seem to be of the nature that new code is seen, and I suppose that's the intent. But a lot of it looks like internal template details are being attributed to the instantiation, which doesn't seem so good.

For example the `iterator.cc` test goes to great lengths to explain that `<iterator>` should never be required, and yet the suggested changes are:
```
tests/cxx/iterator.cc should add these lines:
+#include <ext/aligned_buffer.h>
+#include <ext/alloc_traits.h>
+#include <ext/type_traits.h>
+#include <ext/type_traits>
+#include <iterator>
+#include <memory>
+#include <type_traits>
```
Also, reporting the instantiation location as a default-initialized, invalid location can't be right. I'm not sure how that's resolved to report in the location it is.
 
Could you try forcing `-fno-delayed-template-parsing` onto the IWYU command-line and see if the results change?

kimgr:
> Could you try forcing -fno-delayed-template-parsing onto the IWYU command-line and see if the
> results change?

In your local testing, I mean.

shivanshu3:
> But a lot of it looks like internal template details are being attributed to the instantiation, which doesn't seem so good.

Ah, I'll have to take a look at why that's happening.

> Also, reporting the instantiation location as a default-initialized, invalid location can't be right. I'm not sure how that's resolved to report in the location it is.

`used_loc` in `InstantiatedTemplateVisitor::ReportTypeUse` is unused. It passes `caller_loc()` to `Base::ReportTypeUse`. All other users of this function pass in a location to this function, which shouldn't be needed. A couple users even pass is `CurrentLoc()` which is misleading because internally it uses `caller_loc()`. Maybe I should log a bug saying that we should `CHECK_` that the location passed to this function is the default `SourceLocation` object to avoid confusion in the future?

> Could you try forcing -fno-delayed-template-parsing onto the IWYU command-line and see if the results change?

I tried doing that in the small test case which I added in the linked issue, and that still passes.

Also, on a kind of unrelated note, the way we use IWYU in my team at work is kind of different. We have a forked version of IWYU which dumps out all reported symbol uses to a file. We run it on all .cpp files and then we use a different tool to merge all of these symbol uses into a single graph and then analyze the usage of #includes on that one graph. This way we can tell which files use which symbols from a given .h file, which allows us to remove unused #includes from header files. This tool only removes #includes, it doesn't add any.

IWYU doesn't work as-is on our code because of various reasons, and that is why I couldn't test my change from this PR on our real code.

kimgr:
@shivanshu3 Ah, right you are, `InstantiatedTemplateVisitor::ReportTypeUse` slipped away when I was looking at this.

> Maybe I should log a bug saying that we should CHECK_ that the location passed to this function
> is the default SourceLocation object to avoid confusion in the future?

Yeah, that might be a good idea.

> Also, on a kind of unrelated note, the way we use IWYU in my team at work is kind of different. We 
> have a forked version of IWYU which dumps out all reported symbol uses to a file. 

I've been thinking about a mode like that too; given a mapping from symbol -> include, a complete include graph, and a symbol use -> location mapping, I think IWYU policy should be pretty easily scriptable. Do you have anything you could share, for playing around with?

There are probably some challenges with policy decisions made too early (i.e. during AST visitation) or requiring more context to make the right call (e.g. use-flags, etc), but it might be a nice tool for analysis without action.

shivanshu3:
> Yeah, that might be a good idea.

I've logged the bug #804. I will send out a PR soon.

> Do you have anything you could share, for playing around with?

I'd love to share the tool with you and get some feedback on it. Currently it's tied to our internal build system. I'm waiting on my manager to make a decision about when and how we will be able to talk more about the tool. I'll keep you updated!

kimgr:
This does not seem like the right fix, and no feedback lately. Closing for now.

Issue #755, opened 2020-01-29
False "incomplete type" compile errors


shivanshu3:
We have similar code like the example I have below where Clang is able to compile it just fine without any errors or warnings, but IWYU isn't happy with it. It emits many compile errors, and therefore doesn't run the analysis on such files.

Why is this happening? Shouldn't all code that compiles with Clang, be able to work with IWYU? Note that I tried compiling this in MSVC as well, and it's happy with it too.

I ran clang-check on this file, and it doesn't complain about it either.

Contents of `main.cpp`:
```c++
#include <vector>

struct Foo;

struct Bar
{
   std::vector<Foo> foos;
};
```

Errors:
```
$>include-what-you-use.exe main.cpp -I"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include"
In file included from main.cpp:1:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\vector:6:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\xmemory:6:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\xmemory0:9:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\new:6:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\exception:8:
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\type_traits:1010:19: error: incomplete type 'Foo' used in type trait expression
                : bool_constant<__is_trivially_destructible(_Ty)>
                                ^
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\type_traits:74:18: note: in instantiation of template class 'std::is_trivially_destructible<Foo>' requested here
                : _Conjunction<_First::value, _First, _Rest...>::type
                               ^
[...]

PR #818, opened 2020-08-02
Instantiate implicit methods in a separate pass

kimgr:
Before this we would force-instantiate implicit methods (constructors,
destructors, assignment operators) while traversing the AST for IWYU
analysis. This approach is a little suspect as it modifies the AST
mid-traversal.

Instead, collect all implicit methods in a separate traversal, and
instantiate them before running IWYU analysis.

Quite a lot of code moves around here, but it should be mostly
functionally equivalent. The CanIgnore logic is somewhat dumbed down,
as the visitor is no longer part of the BaseAstVisitor hierarchy, but
it appears to behave the same.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -532,17 +532,10 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
 
   //------------------------------------------------------------
   // (4) Add implicit text.
-
-  // When we see an object that has implicit text that iwyu
-  // wants to look at, we make callbacks as if that text had
-  // been explicitly written.  Here's text we consider:
   //
-  //    * CXXDestructorDecl: a destructor call for each non-POD field
-  //      in the dtor's class, and each base type of that class.
-  //    * CXXRecordDecl: a CXXConstructorDecl for each implicit
-  //      constructor (zero-arg and copy).  A CXXDestructor decl
-  //      if the destructor is implicit.  A CXXOperatorCallDecl if
-  //      operator= is explicit.
+  // This simulates a call to the destructor of every non-POD field and base
+  // class in all classes with destructors, to mark them as used by virtue of
+  // being class members.
   bool TraverseCXXDestructorDecl(clang::CXXDestructorDecl* decl) {
     if (!Base::TraverseCXXDestructorDecl(decl))  return false;
     if (CanIgnoreCurrentASTNode())  return true;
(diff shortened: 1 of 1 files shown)


kimgr:
@shivanshu3 I was looking at https://github.com/include-what-you-use/include-what-you-use/issues/755 and noticed I proposed to move `InstantiateImplicitMethods` out of the main IWYU analysis pass.

I took a stab at it here, and I think it would make it easier to fix #755 for you.

I have two additional patches I would like you to try on top of this PR:

1) Disable all implicit instantiation if -fdelayed-template-parsing is enabled:
```diff
diff --git a/iwyu.cc b/iwyu.cc
index 51ea7ce..9b090bb 100644
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3592,7 +3592,13 @@ class IwyuAstConsumer
     // But we need to be non-lazy: IWYU depends on analyzing what future
     // code *may* call in a class, not what current code *does*.  So we
     // force all the lazy evaluation to happen here.
-    InstantiateImplicitMethods(tu_decl);
+    //
+    // Skip this if -fdelayed-template-parsing is on -- such code bases
+    // typically assume that template code is only complete at the point of
+    // instantiation.
+    if (!compiler()->getLangOpts().DelayedTemplateParsing) {
+      InstantiateImplicitMethods(tu_decl);
+    }
 
     // Run IWYU analysis.
     TraverseDecl(tu_decl);
```

2) Disable forced template instantiation of destructors if -fdelayed-template-parsing is enabled:
```diff
diff --git a/iwyu.cc b/iwyu.cc
index 51ea7ce..d1a29f4 100644
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3730,10 +3730,13 @@ class IwyuAstConsumer
       }
 
       if (auto* dtor = dyn_cast<CXXDestructorDecl>(method)) {
-        if (!dtor->hasBody() && dtor->isImplicit())
+        if (!dtor->hasBody() && dtor->isImplicit()) {
           sema.DefineImplicitDestructor(loc, dtor);
-        if (!dtor->isDefined() && dtor->getTemplateInstantiationPattern())
+        }
+        if (!compiler()->getLangOpts().DelayedTemplateParsing &&
+            !dtor->isDefined() && dtor->getTemplateInstantiationPattern()) {
           sema.PendingInstantiations.emplace_back(dtor, loc);
+        }
       }
     } 
```
It would be interesting to hear how these changes affect your results on Windows. (You may need to explicitly pass `-fdelayed-template-parsing` to `include-what-you-use`.)

kimgr:
@shivanshu3 Also, can I ask that you build and run tests for

* PR as posted
* With first patch
* With second patch

and attach the results here so I can compare? Thanks.

kimgr:
@shivanshu3 Any chance you could take this for a spin? I don't have a Windows rig to test on, so it's hard for me to forecast how this will work out in real life.

shivanshu3:
Hi @kimgr I'm really sorry for the late reply! I have recently been very busy with some other work stuff so I couldn't get around to testing your changes. I'm currently planning on getting back to working on IWYU in January/February 2021.

kimgr:
@shivanshu3 No worries, thanks! It would be interesting to try this in a Windows setting if somebody else wants to volunteer for testing, but otherwise we'll just let it sit until you have time. Thank you!

kimgr:
Rebased.

kimgr:
The separate implicit method instantiation pass is now on master. If we want to revisit behavior on Windows, we can do that on top of latest instead.

PR #819, opened 2020-08-02
Move node logging to Traverse methods

kimgr:
This provides more granular logging of the AST traversal.

Having logging in the Visit methods would only print for node kinds that
were actually visited; some nodes are only traversed.

Logging in Traverse* means we get log output _before_ nodes are visited,
where we would previously see logging only when derived Visit* methods
called their base.

Notably, we now get log output for FunctionProtoTypeLoc under FunctionDecl,
for example.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -324,6 +324,12 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
       return true;               // avoid recursion
     ASTNode node(decl, *GlobalSourceManager());
     CurrentASTNodeUpdater canu(&current_ast_node_, &node);
+
+    if (ShouldPrintSymbolFromCurrentFile()) {
+      errs() << AnnotatedName(string(decl->getDeclKindName()) + "Decl")
+             << PrintablePtr(decl) << PrintableDecl(decl) << "\n";
+    }
+
     return Base::TraverseDecl(decl);
   }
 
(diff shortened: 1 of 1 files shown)


kimgr:
I _think_ this is a strict improvement. The only drawback I can think of is that we might log traversal of nodes that are then omitted from visitation in the base class. But I'd argue that that's less confusing than not seeing all AST elements being visited.

kimgr:
Another way to think about this is that with the patch we log the traversal order. Without the patch we log the structure of the AST. The two are not necessarily the same, see e.g. TraverseFunctionDecl in RecursiveASTVisitor.h.

kimgr:
Unfortunately this segfaults in `RecursiveASTVisitor`, I think when traversing the body of an implicitly-generated constructor in a template. Hmm.

kimgr:
I now understand why this segfaults. In `RecursiveASTVisitor` the `Traverse` methods all early-exit if the node is null, but IWYU's base-most `Traverse` methods don't (because they never use the node directly. When I introduced logging in `Traverse`, we were suddenly calling methods on null Stmts, for example.

Furthermore, it seems like I was misguided about the improved reporting -- contrariwise it looks like we get reporting for more nodes on master.

Closing this.

kimgr:
(One potential benefit would be the reporting before visitation, as that might make it easier to understand which node is being reported: most use reporting happens in `Visit` and some in `Traverse<Derived>`. I _think_ reporting in `Traverse` would make it more consistent. But let's revisit that at some later point.)

PR #822, opened 2020-08-18
Update table of clang versions in README.md

lkeegan:


kimgr:
Thanks for the patch! I fell back to `...` to denote "and so forth for versions after that", because I didn't want to build in something that I need to remember to update with every single release.

I wonder if there's a way to make it clearer _and_ not booby-trapping me for forgetting?

lkeegan:
I see - I don't have any good ideas to improve this so I've just put the `...` back in..

kimgr:
Nice, that's a clear improvement. Thank you!

PR #832, opened 2020-09-23
Add mappings for max_align_t, ptrdiff_t, and wchar_t

osandov:
According to C11, these are all provided by stddef.h, and wchar_t is
also provided by stdlib.h.


kimgr:
Thank you!

PR #836, opened 2020-10-09
docs: fix simple typo, proceded -> preceded

timgates42:
There is a small typo in fix_includes.py.

Should read `preceded` rather than `proceded`.


kimgr:
Thanks, merged!

PR #839, opened 2020-10-23
remove circular dependencies in boost-1.64-all-private.imp

pppyx:
remove circular dependencies in boost-1.64-all-private.imp


kimgr:
Thank you! There seems to be more than one thing going on at once here --

* Moving around/deduping `boost/numeric/odeint` mappings
* Removing `boost/python` mappings
* Removing self-mapping in `boost/variant` (and adding a comment)

Any chance you could do those as three commits, and for all that really are about circular mappings, add them to that comment?

I try to avoid mixing bug fixes (circular mappings) with other changes, to make it easier to revert/backport as fine-grainedly as possible.

pppyx:
I created PR #840 #841 #842

kimgr:
Thanks! I'm happy to take pull requests with multiple commits, as long as the total diff isn't too large. But this works too. I've left some comments in the new PRs, please post back there if anything is unclear.

Closing this one.

PR #841, opened 2020-10-26
Removing boost/python mappings

pppyx:
Removing boost/python mappings in boost-1.64-all-private.imp to avoid circular dependencies.


kimgr:
Since this is also manual cleanup on top of the command-line spelled out at the top, it would be nice if you could mention it in the "manual override" comment up here: https://github.com/include-what-you-use/include-what-you-use/blob/c3b9743085eca024940a94b538c3a5c7a12594d2/boost-1.64-all-private.imp#L4

pppyx:
I think it already had comments in boost-1.64-all-private.imp Line 9-12 ,it just doesn't delete the corresponding directives.

kimgr:
Oh, I see. I stopped reading when the lines looked like actual mappings, I didn't see that they were literal removed lines. Thanks!

PR #846, opened 2020-11-06
Do not provide arguments when calling Decl.isReplaceableGlobalAllocationFunction()

Flowdalic:
LLVM added a second boolean argument to
isReplaceableGlobalAllocationFunction() with 3dd5a298bfff ("[clang]
Annotating C++'s `operator new` with more attributes"). However both
parameters define a default argument. Hence there is no need to
explicitly state the argument, if it is already the default argument,
which is nullptr is isReplaceableGlobalAllocationFunction() case. This
also increases readability.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1249,7 +1249,7 @@ void ProcessFullUse(OneUse* use,
   // <new> will be required for them without us doing any magic for operator new
   // itself.
   if (const FunctionDecl* fn_decl = DynCastFrom(use->decl())) {
-    if (fn_decl->isReplaceableGlobalAllocationFunction(nullptr, nullptr)) {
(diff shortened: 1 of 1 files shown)


kimgr:
Thanks! Can you reference the `llvm` commit that made the change, so it's easier to know from which point this change is required?

This is the first breaking change vs llvm-11, so I might want to get some other changes in first before I branch, but I'll try to address this asap.

Flowdalic:
> Thanks! Can you reference the `llvm` commit that made the change, so it's easier to know from which point this change is required?

This change is never required and is not about forwards compatibility. As far as I can tell, all LLVM versions declare default arguments for this method.

kimgr:
Oh, I see, thanks. Yeah, it does look a little cleaner without the `nullptr`s, let's get this in place.

PR #848, opened 2020-11-11
Correctly detect link with LLVM/CLANG shared libraries.

Romain-Geissler-1A:
Hi,

I do try to build include-what-you-use outside of the LLVM build tree, and my LLVM/Clang installation was configured with LLVM_LINK_LLVM_DYLIB=On. It looks like include-what-you-use CMakeLists.txt fails to detect that correctly.

I am fixing it thusly (this is how @Maskray who is a big LLVM contributor does it in his own out of tree clang-based tool CCLS: https://github.com/MaskRay/ccls/blob/master/CMakeLists.txt#L73)

Cheers,
Romain

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,10 @@ endif()
 
 message(STATUS "IWYU: configuring for LLVM ${LLVM_VERSION}...")
 
+# The good default is given by the llvm toolchain installation itself, but still in
+# case both static and shared libraries are available, allow to override that default.
+option(IWYU_LINK_CLANG_DYLIB "Link against the clang dynamic library" ${CLANG_LINK_CLANG_DYLIB})
+
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
 add_definitions(${LLVM_DEFINITIONS})
@@ -106,7 +110,7 @@ if (MSVC)
 endif()
 
 # If only clang-cpp is available, we take that.
-if (TARGET clang-cpp AND NOT TARGET clangBasic)
+if (IWYU_LINK_CLANG_DYLIB)
   target_link_libraries(include-what-you-use PRIVATE clang-cpp)
 else()
   target_link_libraries(include-what-you-use


kimgr:
Nice, thanks! Does that also mean users can override dylib vs static link?

I tried cmake with `-DCLANG_LINK_CLANG_DYLIB=Off`, but I couldn't seem to override the config from the Clang cmake-config module.

The current hack was probably there to avoid failing if _only_ dylib was available. On my Ubuntu rig, the llvm/clang packages seem to come with both static and dynamic libraries. It would be nice to be able to choose. 

Any thoughts on that? Would we have to introduce a `IWYU_LINK_LLVM_DYLIB` flag to be able to choose?

Romain-Geissler-1A:
I have update the pul request to introduce an option spelled `IWYU_LINK_CLANG_DYLIB` (slightly different from the name you suggested). For me the good default value shall neither be ON nor OFF, but really be what the LLVM toolchain packager choose to use. I tested all 3 cases where `IWYU_LINK_CLANG_DYLIB` is unset, explicitly ON or explicitly OFF and it works as expected for me.

kimgr:
Very nice, thank you!

It looks good to me, I'll just test this quickly to see how it works.

kimgr:
Yes, works without a hitch in all modes (default, `-DIWYU_LINK_CLANG_DYLIB=on`, `-DIWYU_LINK_CLANG_DYLIB=off`). Merging.

MaskRay:
Yeah, this is correct. `CLANG_LINK_CLANG_DYLIB` means that dependent libraries should link against `libclang-cpp.so`
while `CLANG_BUILD_CLANG_DYLIB` means whether `libclang-cpp.so` is produced.

CMakeLists.txt
```diff
@@ -106,7 +110,7 @@ if (MSVC)
 endif()
 
 # If only clang-cpp is available, we take that.
```
aaronpuchert, line 112: That comment is no longer relevant then.

aaronpuchert (commented):
Nice, I didn't know that `CLANG_LINK_CLANG_DYLIB` is exposed to the outside. Then it's of course better to use that.

CMakeLists.txt
kimgr: @aaronpuchert Ouch! Thanks for noticing, I'll clean that out.

PR #854, opened 2020-11-23
Fix build on non-X86 targets where the LLVM x86 backend is not built.

Romain-Geissler-1A:
Hi,

Today I started a built of a full gcc+clang toolchain with PGO + LTO (so taking quite a very long time), and it happens that at the very end the job failed when trying to build include-what-you-use. It happens that I was doing all that on a aarch64 target using native compilers and where the X86 backend was deactivated on purpose. I didn't expect include-what-you-use is X86 only, but it seems I was wrong ;)

I got:

```
FAILED: CMakeFiles/include-what-you-use.dir/iwyu.cc.o
/opt/1A/toolchain/aarch64-v19.0.53/bin/c++  -DGTEST_HAS_RTTI=0 -DIWYU_GIT_REV=\"\" -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -I/opt/1A/toolchain/aarch64-
v19.0.53/include -O2 -I/opt/1A/toolchain/aarch64-v19.0.53/build-pack/19.0.53.0/include -I/workdir/build/build-pack/build-pack-temporary-static-dependencies/install/include -I/workdi
r/build/build-pack/build-pack-temporary-static-dependencies/install/include/ncursesw -I/workdir/build/build-pack/build-pack-temporary-static-dependencies/install/lib/libffi-3.3/include -I/opt/1A/toolchain/aarch64-v19.0.53/build-pack/19.0.53.0/internal-python-only-for-build-pack/include -Wno-error -fPIC -fvisibility-inlines-hidden -Werror=date-time -w -fdiagnostics-color -ffunction-sections -fdata-sections -O3 -DNDEBUG   -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS  -fno-exceptions -fno-rtti -MD -MT CMakeFiles/include-what-you-use.dir/iwyu.cc.o -MF CMakeFiles/include-what-you-use.dir/iwyu.cc.o.d -o CMakeFiles/include-what-you-use.dir/iwyu.cc.o -c /workdir/src/include-what-you-use-0.16/iwyu.cc
/workdir/src/include-what-you-use-0.16/iwyu.cc: In function 'int main(int, char**)':
/workdir/src/include-what-you-use-0.16/iwyu.cc:4142:3: error: 'LLVMInitializeX86TargetInfo' was not declared in this scope; did you mean 'LLVMInitializeAArch64TargetInfo'?
 4142 |   LLVMInitializeX86TargetInfo();
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      |   LLVMInitializeAArch64TargetInfo
/workdir/src/include-what-you-use-0.16/iwyu.cc:4143:3: error: 'LLVMInitializeX86TargetMC' was not declared in this scope; did you mean 'LLVMInitializeAArch64TargetMC'?
 4143 |   LLVMInitializeX86TargetMC(); 
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~
      |   LLVMInitializeAArch64TargetMC
/workdir/src/include-what-you-use-0.16/iwyu.cc:4144:3: error: 'LLVMInitializeX86AsmParser' was not declared in this scope; did you mean 'LLVMInitializeAArch64AsmParser'?
 4144 |   LLVMInitializeX86AsmParser();
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~   
      |   LLVMInitializeAArch64AsmParser
[166/167] Building CXX object CMakeFiles/include-what-you-use.dir/iwyu_ast_util.cc.o
ninja: build stopped: subcommand failed.
```

Unfortunately, I have lost the rest of the toolchain with the job, so right now I can't test this on aarch64. I will re-start a build with this patch and let you know if it worked. I did however test this on a x86_64 machine where I already had saved a working toolchain. On x86_64, this seems to compile (I have not tried to run, but normally it shall work).

Cheers,
Romain

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,11 @@ message(STATUS "IWYU: configuring for LLVM ${LLVM_VERSION}...")
 # case both static and shared libraries are available, allow to override that default.
 option(IWYU_LINK_CLANG_DYLIB "Link against the clang dynamic library" ${CLANG_LINK_CLANG_DYLIB})
 
+# Try to guess the architecture as much as possible from the underlying toolchain.
+set(IWYU_ARCH ${LLVM_NATIVE_ARCH} CACHE STRING "The architecture IWYU will run on.")
+
+add_compile_definitions(IWYU_ARCH=${IWYU_ARCH})
+
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
 add_definitions(${LLVM_DEFINITIONS})
@@ -41,9 +46,9 @@ include_directories(
 set(LLVM_LINK_COMPONENTS
   Option
   Support
-  X86AsmParser
-  X86Desc
-  X86Info
+  ${IWYU_ARCH}AsmParser
+  ${IWYU_ARCH}Desc
+  ${IWYU_ARCH}Info
   )
 
 add_llvm_executable(include-what-you-use
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -4136,12 +4136,15 @@ using include_what_you_use::IwyuAction;
 using include_what_you_use::CreateCompilerInstance;
 
 int main(int argc, char **argv) {
-  // Must initialize X86 target to be able to parse Microsoft inline
-  // assembly. We do this unconditionally, because it allows an IWYU
-  // built for non-X86 targets to parse MS inline asm without choking.
-  LLVMInitializeX86TargetInfo();
-  LLVMInitializeX86TargetMC();
-  LLVMInitializeX86AsmParser();
+  // Must initialize the arch target to be able to parse Microsoft inline
+  // assembly.
+#define IWYU_INTERNAL_LLVM_INITIALIZE_ARCH_TARGET_HELPER2(arch, target) LLVMInitialize##arch##target()
+  // This second macro will expand IWYU_ARCH correctly.
+#define IWYU_INTERNAL_LLVM_INITIALIZE_ARCH_TARGET_HELPER1(arch, target) IWYU_INTERNAL_LLVM_INITIALIZE_ARCH_TARGET_HELPER2(arch, target)
+#define IWYU_INTERNAL_LLVM_INITIALIZE_ARCH_TARGET(target) IWYU_INTERNAL_LLVM_INITIALIZE_ARCH_TARGET_HELPER1(IWYU_ARCH, target)
+  IWYU_INTERNAL_LLVM_INITIALIZE_ARCH_TARGET(TargetInfo);
+  IWYU_INTERNAL_LLVM_INITIALIZE_ARCH_TARGET(TargetMC);
+  IWYU_INTERNAL_LLVM_INITIALIZE_ARCH_TARGET(AsmParser);
 
   // The command line should look like
   //   path/to/iwyu -Xiwyu --verbose=4 [-Xiwyu --other_iwyu_flag]... CLANG_FLAGS... foo.cc


Romain-Geissler-1A:
@kimgr After looking at https://github.com/include-what-you-use/include-what-you-use/commit/8f648796746a30fd48031e05435282654e191b4b I am actually unsure this is what shall be done. To be honest, I don't understand this whole X86 Microsoft inline assembly, I am totally not interested by any non-Linux plateform. What I know however is that I definitely don't want to build a X86 backend when creating native toolchains for ARM. What do you suggest I do ? Go back to calling the Native C function like you used to do it before https://github.com/include-what-you-use/include-what-you-use/commit/8f648796746a30fd48031e05435282654e191b4b ? And what shall I do with the X86 mentions in the CMakefile lists then ? Just drop them entirely and hope the native backend will include them already ?

The other option is see is adding some IWYU_I_ABSOLUTELY_DONT_CARE_ABOUT_WINDOWS configure option to conditionally drop these X86 Windows specific things ;)

Cheers,
Romain

Romain-Geissler-1A:
FYI, I checked my delivered lib/cmake/llvm/LLVMConfig.cmake file, and in addition to defining LLVM_NATIVE_ARCH, it also defines `set(LLVM_TARGETS_TO_BUILD X86)` which seems specified here https://llvm.org/docs/CMake.html as a list of targets. So in the suggestion in my previous comments, we could define an hypethical IWYU_I_ABSOLUTELY_DONT_CARE_ABOUT_WINDOWS (with obviously a more politically correct wording :D) defined by checking the presence of `X86` in `${LLVM_TARGETS_TO_BUILD}`.

Romain-Geissler-1A:
Yet another more option which I think is even more simple is to just drop the explicit dependencies on the X86 libraries, and just call `LLVMInitializeAllAsmParsers` and the likes for all targets LLVM knows, and if it happens that it contains X86, people using Windows won't have issues.

Romain-Geissler-1A:
This following (untested) patch shows what I meant in my latest comment:

```
commit dd974ceb669a86938870337268aa9ceb30f27168 (HEAD -> fix-aarch64-build)
Author: Romain Geissler <romain.geissler@amadeus.com>
Date:   Mon Nov 23 22:12:01 2020 +0000

    Fix build on non-X86 targets where the LLVM x86 backend is not built.

diff --git a/CMakeLists.txt b/CMakeLists.txt
index 3c41384..abb981f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,9 +41,9 @@ include_directories(
 set(LLVM_LINK_COMPONENTS
   Option
   Support
-  X86AsmParser
-  X86Desc
-  X86Info
+  AllTargetsAsmParsers
+  AllTargetsDescs
+  AllTargetsInfos
   )

 add_llvm_executable(include-what-you-use
index 931d57b..22583e1 100644
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -4136,12 +4136,10 @@ using include_what_you_use::IwyuAction;
 using include_what_you_use::CreateCompilerInstance;

 int main(int argc, char **argv) {
-  // Must initialize X86 target to be able to parse Microsoft inline
-  // assembly. We do this unconditionally, because it allows an IWYU
-  // built for non-X86 targets to parse MS inline asm without choking.
-  LLVMInitializeX86TargetInfo();
-  LLVMInitializeX86TargetMC();
-  LLVMInitializeX86AsmParser();
+  // Initialize all targets, might be needed to parse inline assembly.
+  llvm::InitializeAllTargetInfos();
+  llvm::InitializeAllTargetMCs();
+  llvm::InitializeAllAsmParsers();

   // The command line should look like
   //   path/to/iwyu -Xiwyu --verbose=4 [-Xiwyu --other_iwyu_flag]... CLANG_FLAGS... foo.cc

```

kimgr:
Yeah, this hard dependency on `X86` is annoying. I wonder if we could hide all that behind a flag, like you did recently for clang dylib?

Something like:
```
option(IWYU_ENABLE_MS_INLINE_ASM "Support MS inline assembly" /* default: X86 target present */)
```

That way it's possible to:
- Build without support for ms inline asm even if x86 is available
- Have a working build by default if x86 is unavailable (albeit without support for ms inline asm)
- State the intent "I require support for ms inline asm", and have the build fail if x86 is unavailable

Unfortunately, I think it's a bit of work -- we'd need to generate an `iwyu-config.h` from CMake with a corresponding `#define` (useful infrastructure anyway).

Does this sound reasonable? I can pick it up and experiment if you want.

kimgr:
Hmm. I guess the `cxx/ms_inline_asm.cc` test needs to be conditionally disabled in that case, and we don't have facilities for that yet.

Romain-Geissler-1A:
Hi,

I know almost nothing about CMake, and each time I put my hands in it I remember how much I dislike CMake ;) My approach each was not through a `iwyu-config.h` file but by passing definitions directly on the command line through `add_compile_definitions`, which works as well, but I agree is the poor man solution. I have no idea how to generate a `iwyu-config.h` and I also have no idea how to upgrade your tests (to be honest, I never ran them, I just build IWYU and ship them as part of our toolchain in our company. It looks like you are going to change the way you run tests anyway in https://github.com/include-what-you-use/include-what-you-use/pull/859 so I think I will let people who know cmake and IWYU better than me follow up on this one.

Cheers,
Romain

kimgr:
OK. I'm not much of a CMake wiz myself, and I don't have any concrete ideas for the testing, so I'll let this bake for a while. Maybe I'll find some time to start working on it during Christmas.

PR #855, opened 2020-11-25
Store file-specific IWYU and Clang arguments in test source files

storoj:
Some of source files require extra Clang or IWYU arguments like
`-std=c++17` or `--mapping_file=...`.

Before this change all these args were stored in `run_iwyu_tests.py`.

This change adds the concept of "ARGS lines" so now launch arguments
could be specified directly in a source file that represents a test case.

Inspired by RUN lines of `lit` in LLVM:
https://llvm.org/docs/TestingGuide.html, "Writing new regression tests".


tests/cxx/iwyu_stricter_than_cpp.cc
```diff
@@ -7,6 +7,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+// ARGS: -Xiwyu --check_also="tests/cxx/*-autocast.h" -Xiwyu --check_also="tests/cxx/*-fnreturn.h" -Xiwyu --check_also="tests/cxx/*-typedefs.h" -Xiwyu --check_also="tests/cxx/*-d2.h" -I .
```
storoj, line 10: It might be a good idea to support multi-line ARGS strings to avoid very long lines like this, but that feature seems to be a bit tricky to implement.

kimgr:
Very nice, thank you!

I've tried this before and never ended up with anything I liked, but this looks like a great increment. I think I was looking to solve too many problems at once.

I'd like to suggest a few simple high-level changes before looking into this more deeply:

* I'd like for the keyword to be `// IWYU-ARGS`, so everything IWYU-test-related has the same prefix
* Can you add support for in-file args as a separate commit first, and then port all the testcases separately? Easier to review the mechanism that way.

tests/cxx/iwyu_stricter_than_cpp.cc
kimgr: It should be possible to do something with a trailing `\` to indicate line continuation... Once the tactical changes I suggested are applied, I can experiment with that on top of your base.

storoj:
Hey @kimgr, thanks for your feedback

* renamed `ARGS` to `IWYU_ARGS` (you suggested `IWYU-ARGS` for consistency, however all other constants have `IWYU_` prefix with underscore. Did you _really_ mean `IWYU-ARGS`?);
* extracted the migration of testcases into a separate change to make it easier to review;
* added multiline support for `IWYU_ARGS`.

iwyu_test_util.py
```diff
@@ -410,11 +423,24 @@ def _CompareExpectedAndActualSummaries(expected_summaries, actual_summaries):
...
+
+        comments = itertools.takewhile(_IsCommentLine, fh)
+        
+        while len(args) > 0 and args[-1] == '\\':
+          args = args[:-1] # delete trailing slash
+          try:
+            line = next(comments)
+          except StopIteration:
+            raise SyntaxError('{}:{} syntax error in multiline IWYU_ARGS'.format(cc_file, lineno))
+          args += ' ' + _GetCommentLineContents(line)
```
kimgr, line 441: The `_GetCommentLineContents` functions seems a little over-abstracted, I'd probably just go with `line.lstrip('/').lstrip()`

iwyu_test_util.py
```diff
@@ -203,6 +204,18 @@ def _StripCommentFromLine(line):
     return line
 
 
+def _IsCommentLine(line):
```
kimgr, line 207: I think this can be spelled out inline, no need to have an abstraction for something as simple as checking `//`

kimgr (changes requested):
I think it makes sense to support multiline arguments from the get-go instead of rewriting all the tests twice. And let's use multiline `IWYU_ARGS` lines only when it's necessary, most tests should fit their args within 80 column width, right?

I'd like to suggest the following commit order:

* Some of source files require extra Clang or IWYU arguments like
* ... squashed with Add support for multiline IWYU_ARGS
* Migrate test launch arguments to IWYU_ARGS
* ... squashed with Use multiline IWYU_ARGS for long lists of arguments 
* Delete IWYU launch arguments editing from test runner

Overall, this is an amazing improvement to IWYU's test infra, thank you!

kimgr:
@storoj Did you want to continue working on this?

storoj:
@kimgr absolutely. Sorry for the delays, I just had pretty long winter holidays, and now I am trying to get back to work.

kimgr:
Nice! :snowflake:

storoj:
Hi @kimgr, I addressed your comments:
* squashed and reordered commits as you suggested;
* strictly one space in `_IWYU_TEST_RUN_ARGS_RE`;
* reordered the verbosity flags (--verbose=3; IWYU_ARGS; env var);
* avoided usage of `StopIteration`;

tests/cxx/relative_exported_mapped_include.cc
```diff
@@ -7,6 +7,9 @@
 //
 //===----------------------------------------------------------------------===//
 
+// IWYU_ARGS: -Xiwyu --mapping_file=tests/cxx/relative_exported_mapped_include.imp\
```
kimgr, line 10: Hmm, this line is actually too long. Unfortunately, I think you need to turn it around into:
```
// IWYU_ARGS: -I tests/cxx/subdir -Xiwyu \
//            --mapping_file=...
```
It looks a little atypical, but it works better. We're pretty strict with the 80-column limit.

kimgr (commented):
Beautifully done!

Some remaining suggestions, but I can't wait for this to go in.

kimgr:
Also, please rebase on top of master and the Travis job should go green (it's currently failing on a bug in the LLVM CMake packaging, so it never gets around to build+run the tests. Workaround is on master.).

iwyu_test_util.py
```diff
@@ -398,6 +404,35 @@ def _CompareExpectedAndActualSummaries(expected_summaries, actual_summaries):
   return failures
 
 
+def _GetLaunchArguments(cc_file):
```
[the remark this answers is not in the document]
storoj: Nice! TIL that `for` loop args are not scoped, and the last `lineno` is accessible after `break`-ing.

Thank you for the suggestion, that code is definitely easier to understand.

storoj:
I adopted the in-file args to 80-column rule, however I kept "smart" formatting in these files for better readability:
```
iwyu_stricter_than_cpp.cc
non_transitive_include.cc
prefix_header_includes_add.cc
prefix_header_includes_remove.cc
```
@kimgr Do you agree that when `IWYU_ARGS` has a list of visually grouped `-Xwyu ...` or `-include ...` args then any other following arg should be visually separated?

iwyu_test_util.py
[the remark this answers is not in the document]
kimgr: A cool thing about Python is that the smallest scoping level is function (afaik, at least). This opens up for all sorts of nice patterns:
```
def f1(flag):
    if flag:
        value = "this"
    else:
        value = "that"
    f2(value)
```
So unlike C++, no matter in which scope it's bound, `value` is visible inside the whole function `f1` (well, after it's bound :)).

kimgr:
> Do you agree that when IWYU_ARGS has a list of visually grouped -Xwyu ... or -include ... args
> then any other following arg should be visually separated?

Yeah, I think the current layout looks great, even if it's not strictly filling 80 columns. Nicely done!

kimgr (approved):

kimgr:
I was hoping for the travis build to run to completion, but it looks like LLVM packaging is broken after the recent bump to 13 on LLVM main. I pulled your branch out and ran all the build and test steps I could think of. It works, let's merge! :)

Thank you so much for pulling this completion, it's a really nice improvement.

PR #856, opened 2020-11-26
Use higher-level approach to build, test and install the project on CI

storoj:
`cmake --build`, `cmake --install` and `ctest` hide the implementation details about which exactly Generator was used to configure the build tree.


kimgr:
I'm more of a ninja zealot than a CMake one, but this looks neat and tidy. Thanks!

PR #857, opened 2020-11-26
Use version sort to determine the highest llvm version

storoj:
```
# install earlier version later
$ apt install llvm-11
$ apt install llvm-10

# current version
$ ls -t /usr/lib/ | grep '^llvm-' | head -n 1 | sed -E 's/llvm-(.+)/\1/'
10

# updated version
$ find /usr/lib -maxdepth 1 -name llvm-* -printf '%f\n' | sort -V -r | head -n1 | cut -c 6-
11
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ addons:
 
 before_install:
  # Extract the version number from the most-recently installed LLVM
- - VERSION=`ls -t /usr/lib/ | grep '^llvm-' | head -n 1 | sed -E 's/llvm-(.+)/\1/'`
+ - VERSION=`find /usr/lib -maxdepth 1 -name llvm-* -printf '%f\n' | sort -V -r | head -n1 | cut -c 6-`
 
  # Absolute paths to LLVM's root and bin directory
  - PREFIX_PATH=`llvm-config-$VERSION --prefix`


kimgr:
Right, but I don't think we want the highest version. We really do want the one that was latest installed (6 lines above).

Did you have any practical problems with the current solution, or did you just think it looked messy? (it does :))

storoj:
I did not have practical problems, it's just my curiosity.

Even if we want the latest _installed_, there is still a potential bug. For example, we might have `llvm-12` installed after `llvm-11`, and if the script requests installation of `llvm-11` (which is already installed, and it's mtime is not the latest), then the installation will result with noop, and `ls -t ...` will point to `llvm-12` (which was installed after `llvm-11`).

I don't think it's really necessary to fix anything here, I just wanted to share my thoughts and ideas here. Feel free to reject the PR :)

PS As a bit more reliable solution I could suggest using `apt-cache depends`:
```
$ apt-cache depends llvm
llvm
  Depends: llvm-runtime
  Depends: llvm-12
```

So we could get the exact llvm version that we were about to install.

kimgr:
Hmm, that looks a little harder to parse.

There's an additional wrinkle that you may not be aware of -- when we branch IWYU to match a specific clang version, we pin the LLVM/Clang packages to that version, see e.g. https://github.com/include-what-you-use/include-what-you-use/commit/5db414ac448004fe019871c977905cb7c2cff23f

So `master` always tracks the (version-less) snapshots, and the respective `clang_NN` branch always tracks the released `-NN` packages.

I was thinking of just specifying up-front something like `- LLVM_VER=12` and use that version throughout, but I don't think it plays so well with the snapshots. The nice thing with the current model is that IWYU master is guaranteed to always build against the latest Clang master, independent of version, and it just works.

So yeah, I'm inclined to close this, unless you can come up with something nice to cover all these strange combinations :)

kimgr:
> Even if we want the latest installed, there is still a potential bug. For example, we might have llvm-12 installed after
> llvm-11, and if the script requests installation of llvm-11 (which is already installed, and it's mtime is not the latest),
> then the installation will result with noop, and ls -t ... will point to llvm-12 (which was installed after llvm-11).

I think that's the behavior we want. For include-what-you-use's `master` branch, we want the latest snapshot of the LLVM packages (whatever most closely matches their `main`).

Since the travis job adds the apt.llvm.org deb source, there will _always_ be a newer package there than what's available in the base ubuntu bionic distribution. That snapshot package is unconditionally installed, and will have an mtime later than whatever was in the OS image to begin with.

I might be misunderstanding what you're saying, but I don't see a good reason to modify this policy.

kimgr:
Closing this, I don't think it adds any value.

PR #858, opened 2020-11-26
Fix warnings related to clang-12

storoj:


kimgr:
> Fix "loop variable ... binds to a temporary constructed" (-Wrange-loop-construct)

Commit subject too long. I usually try to avoid mentioning tools and warnings in the commit subject and focus on what was corrected. I suggest something like:

> Fix reference-to-temporary in header-search
>
> The map range iterator returns a 'const pair<const string, ...>&', not the
> expected 'const pair<string, ...>&', and the compiler helpfully converts
> to a temporary. Make the types match.
>
> Found by Clang's -Wrange-loop-construct 

But I agree that your original `auto` solution is probably nicer. Something like:
```
for (const auto& entry : ...)
```

would be less susceptible to bugs without losing readability, so if you want to go back to that (and adjust the commit message) that would probably be even better.

kimgr:
Oh, and for the new enumerators, could you mention in the commit message the Clang commit sha that introduced them, rather than clang major version? Thanks.

storoj:
Thanks @kimgr 

I addressed your comments:
* switched back to `const auto&`;
* updated the commit message for `-Wrange-loop-construct` fix;
* mentioned `sha`s from llvm-project that introduced enumeration values that were unhandled;

kimgr:
Nice, :+1:

PR #859, opened 2020-11-26
Use CMake to generate and run tests

storoj:
This PR is based on #855 and it introduces two commits with similar changes.

The idea of "Replace dynamically created test cases…" came to me while working on ObjC support, and I wanted to have simpler way to run ObjC tests only. So I replaced per-test-file TestCases with per-test-file test functions in scoped TestCases. It ended up with three "scopes": C, Cxx and ObjC so I could execute `./run_iwyu_tests.py ObjC` to only run ObjC tests.

Then I researched possible ways of tests parallelization in Python, but I could not find any suitable solution.

I analyzed `run_iwyu_tests.py` script and concluded that it should be easy to let CMake do the job while leaving `iwyu_test_util.py` as a test runner. It worked, and tests parallelization looks impressive to me.

The biggest benefits of cmake-based tests from my perspective:
* test parallelization (`ctest -j N`);
* rerun only failed tests (`ctest --rerun-failed`);
* potentially better integration with other tools (?);
* much faster on 12-core CPU:
** ~7.7s with `run_iwyu_tests.py`;
** ~11.2s with `ctest --progress`;
** ~1.3s with `ctest -j 12 --progress`;
** ~0.9s with `ctest -R 'Cxx::badinc' -j 12 --progress`;

Overall it gives about 8x speedup of local test runs on my machine, I think it might be a good productivity bonus.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,9 +177,23 @@ endif()
 find_package(PythonInterp)
 if(PYTHONINTERP_FOUND)
   enable_testing()
-  add_test(NAME iwyu_tests
-    COMMAND ${PYTHON_EXECUTABLE} run_iwyu_tests.py -- $<TARGET_FILE:include-what-you-use>
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+
+  function(ADD_IWYU_TEST NAME FILE)
+    add_test(NAME ${NAME}
+      COMMAND ${PYTHON_EXECUTABLE} run_iwyu_tests.py --run-test-file=${FILE} -- $<TARGET_FILE:include-what-you-use>
+      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+  endfunction()
+
+  execute_process(
+    COMMAND ${PYTHON_EXECUTABLE} run_iwyu_tests.py --list-test-files
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    OUTPUT_VARIABLE TEST_NAMES_AND_FILES)
+  string(REPLACE "\n" ";" TEST_NAMES_LIST ${TEST_NAMES_AND_FILES})
+  foreach (TEST_NAME_AND_FILE IN ITEMS ${TEST_NAMES_LIST})
+    string(REPLACE ":" ";" TEST_NAME_AND_FILE ${TEST_NAME_AND_FILE})
+    ADD_IWYU_TEST(${TEST_NAME_AND_FILE})
+  endforeach()
+
   add_test(NAME fix_includes_test
     COMMAND ${PYTHON_EXECUTABLE} fix_includes_test.py
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -429,12 +429,10 @@ def _GetLaunchArguments(cc_file):
   return args
 
 
-def TestIwyuOnRelativeFile(test_case, cc_file, cpp_files_to_check,
-                           verbose=False):
+def TestIwyuOnRelativeFile(cc_file, cpp_files_to_check, verbose=False):
   """Checks running IWYU on the given .cc file.
 
   Args:
-    test_case: A googletest.TestCase instance.
     cc_file: The name of the file to test, relative to the current dir.
     cpp_files_to_check: A list of filenames for the files
               to check the diagnostics on, relative to the current dir.
@@ -474,4 +472,5 @@ def TestIwyuOnRelativeFile(test_case, cc_file, cpp_files_to_check,
       _GetExpectedSummaries(cpp_files_to_check),
       _GetActualSummaries(output))
 
-  test_case.assertTrue(not failures, ''.join(failures))
+  if failures:
+    raise AssertionError(''.join(failures))
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -39,34 +39,31 @@ def Partition(l, delimiter):
   return l[:delim_index], l[delim_index+1:]
 
 
+def TestIwyuOnRelevantFiles(filename):
+  logging.info('Testing iwyu on %s', filename)
+  # Split full/path/to/foo.cc into full/path/to/foo and .cc.
+  (all_but_extension, extension) = os.path.splitext(filename)
+  (dirname, basename) = os.path.split(all_but_extension)
+  # Generate diagnostics on all foo-* files (well, not other
+  # foo-*.cc files, which is not kosher but is legal), in addition
+  # to foo.h (if present) and foo.cc.
+  all_files = (glob.glob('%s-*' % all_but_extension) +
+               glob.glob('%s/*/%s-*' % (dirname, basename)) +
+               glob.glob('%s.h' % all_but_extension) +
+               glob.glob('%s/*/%s.h' % (dirname, basename)))
+  
+  files_to_check = [f for f in all_files if not f.endswith(extension)]
+  files_to_check.append(filename)
+
+  # IWYU emits summaries with canonicalized filepaths, where all the
+  # directory separators are set to '/'. In order for the testsuite to
+  # correctly match up file summaries, we must canonicalize the filepaths
+  # in the same way here.
+  files_to_check = [PosixPath(f) for f in files_to_check]
+  iwyu_test_util.TestIwyuOnRelativeFile(filename, files_to_check, verbose=True)
+
+
 def GenerateTests(rootdir, pattern):
-  def _GetTestBody(filename):
-    def _test(self):
-      logging.info('Testing iwyu on %s', filename)
-
-      # Split full/path/to/foo.cc into full/path/to/foo and .cc.
-      (all_but_extension, _) = os.path.splitext(filename)
-      (dirname, basename) = os.path.split(all_but_extension)
-      # Generate diagnostics on all foo-* files (well, not other
-      # foo-*.cc files, which is not kosher but is legal), in addition
-      # to foo.h (if present) and foo.cc.
-      all_files = (glob.glob('%s-*' % all_but_extension) +
-                   glob.glob('%s/*/%s-*' % (dirname, basename)) +
-                   glob.glob('%s.h' % all_but_extension) +
-                   glob.glob('%s/*/%s.h' % (dirname, basename)))
-      files_to_check = [f for f in all_files if not fnmatch(f, pattern)]
-      files_to_check.append(filename)
-
-      # IWYU emits summaries with canonicalized filepaths, where all the
-      # directory separators are set to '/'. In order for the testsuite to
-      # correctly match up file summaries, we must canonicalize the filepaths
-      # in the same way here.
-      files_to_check = [PosixPath(f) for f in files_to_check]
-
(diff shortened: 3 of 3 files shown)


kimgr:
Interesting idea! I'm having an intuitive reaction of "No! Not more CMake magic!", that I'm fighting back against :)

I think the concept of using classes for directories and functions for tests is quite nice. I think they could possibly even be nested, to generate something like:
```
class cxx:
    class subdir:
        def test_something_down_here(self):
            pass

    def test_something_up_here(self):
        pass
```
It seems nicer to me to ignore any normalcy of class naming in Python and just let the names reflect the directory names.

I'm sad to see `run_iwyu_tests.py` go. I would prefer to be able to run tests without involving CMake. But it could be mostly for sentimental reasons. So maybe we can spare my heart and do this in increments... :)

Also, I wonder if it would be easier/more ergonomic to generate most of the CMake testing from python. Something like:
```
$ python run_iwyu_tests.py --list
c::keep_includes
c::elaborated_struct
cxx::alias_template
...
```
That way testcases are generated is contained on the python side, and CMake can focus on scheduling them.

Hmm. That turned out rambly. More actionable: could you put this on hold, and do the `run_iwyu_tests.py` refactoring separately? So it's easier to see and play with in isolation. Thanks!

storoj:
@kimgr thanks for the feedback

I extracted `run_iwyu_tests.py` refactoring to #863, now this PR only contains Cmake-related changes.

storoj:
>> Also, I wonder if it would be easier/more ergonomic to generate most of the CMake testing from python.
>> ... That way testcases are generated is contained on the python side, and CMake can focus on scheduling them.

IMO we need following data in order to create a test on the CMake side:
* "suite" / "group" name (c / cxx), currently it's a name of `tests/c` and `tests/cxx` directories;
* source file path (tests/cxx/using_unused.cc);
* test name (a filename without an extension, "using_unused";

If implemented in Python the script has to enumerate files using a glob, strip file names to delete extensions, and output the list of something-separated (comma/tab/semicolon/whatever) data so CMake could use it.

In fact the suggested implementation in CMake does exactly that: it just enumerates test files, strips extensions from file names, and adds tests using a directory name for grouping and stripped file names as test names. I am not quite sure if splitting this into Cmake and Python parts could simplify anything significantly.

I am not advocating CMake, I think I could feel your pain while I was learning the CMake syntax :) However, the parallelization benefit is pretty noticeable: 1.3s against 10.5s on my machine. With such an improvement I can literally run all tests on every save when I work with Python part, and after every build when I work with C++.

kimgr:
@storoj I'm proposing we put a mode into `run_iwyu_tests.py` that can just print all the tests.

That way this turns into something like (stubs for stuff that I don't know how to do in cmake and don't care to google):
```
  function(GenerateTests)
    # testnames = `${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/run_iwyu_tests.py --list`
    foreach (test ${testnames})
      add_test(NAME ${test}
                     `${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/run_iwyu_tests.py $test -- $<TARGET_FILE:include-what-you-use>`
    endforeach()
  endfunction()
```

That way:

* The policy for test discovery is in one place only (run_iwyu_tests.py)
* The CMake parallellization can be exploited like you described above
* run_iwyu_tests.py can still be used by old dogs like me

I see now that there's probably some overhead in that scenario -- run_iwyu_tests.py needs to discover all tests for every test it runs -- but that should be solvable once the basic infrastructure is in place. I'm not quite sure why the test runner even uses the unittest module.

storoj:
@kimgr ok I will give it a try

storoj:
> I'm not quite sure why the test runner even uses the unittest module
Do you mean that since we auto-generate the tests then we could also run the tests ourselves and verify the exit code of each of these runs? That might be true, but `unittest` also does extra little things:
* `./run_iwyu_tests cxx.test_using_unused` to run a single test;
* `./run_iwyu_tests.py cxx` to run only Cxx tests;
* reorders the results so the failures are on the bottom of the full output;
* maybe more, but I was pretty much actively using only the items above;

kimgr:
The updated patch looks like a nice distribution of responsibilities. I'm going to play around with it a little and see how it behaves in practice.

kimgr:
I think you could follow up with a separate commit (maybe before the new `--list` switch) to remove the annoying `Registering...` message:
```diff
-      logging.info('Registering %s.%s to test %s', cls.__name__, test_name,
-                   filename)
```

That's not useful, and especially not for the `--list` mode.

kimgr:
Another thought I had earlier was this:

* we could add another switch `--list-test-files`, that prints the path of all testcases
* then add yet another switch `--run-test-file=<path>` that given a path, registers it as a testcase, and runs it

That way we don't even need to start up the test loader to run a single testcase, and the cmake jobs should be on par with what you had before (or faster).

But I think I'd rather work on this as a completely separate followup if the current patch helps with your parallel cmake workflow.

run_iwyu_tests.py
```diff
@@ -87,26 +85,56 @@ def _AddTestFunctions(cls):
...
 
+  parser = argparse.ArgumentParser(add_help=False, usage=argparse.SUPPRESS)
+  group = parser.add_mutually_exclusive_group()
+  group.add_argument('--list', dest='list_tests', action='store_true')
+  group.add_argument('--list-test-files', action='store_true')
+  group.add_argument('--run-test-file')
+  (runner_args, _) = parser.parse_known_args(unittest_args)
+
+  if runner_args.run_test_file:
+    exit(TestIwyuOnRelativeFile(runner_args.run_test_file))
```
storoj, line 127: I am not sure what exactly produces exit code `1`, but it works. I think the `raise AssertionError` should force-exit with code 1, otherwise something like `exit(None)` is being called.

run_iwyu_tests.py
```diff
@@ -87,26 +85,56 @@ def _AddTestFunctions(cls):
...
+
+  @GenerateTests(rootdir='tests/c', pattern='*.c')
```
storoj, line 129: Moving `@GenerateTests` into `main` defers the creation of test classes, so in case when `--run-test-file` is set the initialization of test classes does not happen. IDK if it is a good practice in Python, maybe we should switch back to `RegisterFilesForTesting`.

CMakeLists.txt
```diff
@@ -172,9 +172,23 @@ endif()
 find_package(PythonInterp)
 if(PYTHONINTERP_FOUND)
   enable_testing()
-  add_test(NAME iwyu_tests
-    COMMAND ${PYTHON_EXECUTABLE} run_iwyu_tests.py -- $<TARGET_FILE:include-what-you-use>
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+
+  function(ADD_IWYU_TEST NAME FILE)
```
storoj, line 176: I use this helper function only to unpack `TEST_NAME_AND_FILE` list into the list of named arguments.

CMakeLists.txt
```diff
@@ -172,9 +172,23 @@ endif()
...
+    add_test(NAME ${NAME}
+      COMMAND ${PYTHON_EXECUTABLE} run_iwyu_tests.py --run-test-file=${FILE} -- $<TARGET_FILE:include-what-you-use>
+      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
```
zaufi, line 179:
```suggestion
      WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
```

CMakeLists.txt
```diff
@@ -172,9 +172,23 @@ endif()
...
+  endfunction()
+
+  execute_process(
+    COMMAND ${PYTHON_EXECUTABLE} run_iwyu_tests.py --list-test-files
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
```
zaufi, line 184:
```suggestion
    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
```

CMakeLists.txt
```diff
@@ -172,9 +172,23 @@ endif()
...
+    OUTPUT_VARIABLE TEST_NAMES_AND_FILES)
+  string(REPLACE "\n" ";" TEST_NAMES_LIST ${TEST_NAMES_AND_FILES})
+  foreach (TEST_NAME_AND_FILE ${TEST_NAMES_LIST})
```
zaufi, line 187:
```suggestion
  foreach (TEST_NAME_AND_FILE IN LISTS TEST_NAMES_LIST)
```

CMakeLists.txt
storoj: Hi @zaufi, could you please clarify why `SOURCE_DIR` should be replaced with `BINARY_DIR` here?
The expanded command should be something like:
`$ python3 run_iwyu_tests.py ...`
and `run_iwyu_tests.py` is in project's root source directory. I expect `BINARY_DIR` to be a directory where build artifacts are stored (the output iwyu binary).

In order for python interpreter to "see" `run_iwyu_tests.py` the working dir should be the root source dir.

kimgr: Yeah, that seems surprising to me too. `run_iwyu_tests.py` is definitely in the source dir.

zaufi, line 179: I meant that the current working directory better set to the binary one, so if tests produce any files they are not going to pollute the source tree.
```suggestion
      COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/run_iwyu_tests.py --run-test-file=${FILE} -- $<TARGET_FILE:include-what-you-use>
      WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
```

CMakeLists.txt
kimgr: It's not clear to me why this is better? The `IN LISTS` syntax mostly seems to be intended to deal with nested lists, and `TEST_NAMES_LIST` is just a plain list, isn't it?

iwyu_test_util.py
```diff
@@ -429,8 +429,7 @@ def _GetLaunchArguments(cc_file):
   return args
 
 
-def TestIwyuOnRelativeFile(test_case, cc_file, cpp_files_to_check,
-                           verbose=False):
+def TestIwyuOnRelativeFile(cc_file, cpp_files_to_check, verbose=False):
```
kimgr, line 432: There seems to be a lot going on at once in this commit. I wonder if you could separate it into a few separate refactors to set the stage, and then one or more distinct commits to add new functionality?

CMakeLists.txt
zaufi: It just gives more hints about semantic to the reader... The older syntax considered if not deprecated but at least is not a _Modern CMake_ approach. For ~plain lists~ literal items there is `foreach(... IN ITEMS ...)`.

iwyu_test_util.py
storoj: @kimgr I used `IN LISTS` syntax as @zaufi suggested, and I split the big commit into smaller atomic changes.

run_iwyu_tests.py
kimgr: Yeah, it looks a little funny, but I think it's fine.

kimgr:
LGTM! I've played with it in my local checkout, and it seems work really well. Let's not drag this out any longer. Thank you so much!

PR #863, opened 2020-11-30
Replace dynamically created test cases with test functions

storoj:
Depends on #855.

Currently each test source file `{dir}/{name}.{ext}` creates a test case with `{name}` name. Files in different root `{dir}`s but with same {name}s can end up with a name collision, so test runner just adds suffix '2' to the name of test suite to avoid collisions. If a developer wants to only run a specific test using `run_iwyu_tests.py {name}` it might be difficult to figure out which exactly test failed.

This change adds extra layer of hierarchy (scoping) for tests.

Having 'C' and 'Cxx' tests this change introduces 'c' and 'cxx' TestCase classes. As before, test runner walks down the `tests/` subtree, and populates test case classes with `test_{filename}` methods automatically.

Test name examples before and after this change:
File Name  | Test Before   | Test After
-|-|-
c/keep_includes.c   | keep_includes | c.test_keep_includes
cxx/using_unused.cc | using_unused  | cxx.test_using_unused

It also becomes possible to run C or Cxx tests only:
```
run_iwyu_tests.py c
run_iwyu_tests.py cxx
```

### DIFF
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -39,63 +39,71 @@ def Partition(l, delimiter):
   return l[:delim_index], l[delim_index+1:]
 
 
-class OneIwyuTest(unittest.TestCase):
-  """Superclass for tests.  A subclass per test-file is created at runtime."""
+def GenerateTests(rootdir, pattern):
+  def _GetTestBody(filename):
+    def _test(self):
+      logging.info('Testing iwyu on %s', filename)
 
-  def RunOneTest(self, filename):
-    logging.info('Testing iwyu on %s', filename)
-    # Split full/path/to/foo.cc into full/path/to/foo and .cc.
-    (all_but_extension, _) = os.path.splitext(filename)
-    (dirname, basename) = os.path.split(all_but_extension)
-    # Generate diagnostics on all foo-* files (well, not other
-    # foo-*.cc files, which is not kosher but is legal), in addition
-    # to foo.h (if present) and foo.cc.
-    all_files = (glob.glob('%s-*' % all_but_extension) +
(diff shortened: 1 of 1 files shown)


run_iwyu_tests.py
```diff
@@ -39,70 +39,75 @@ def Partition(l, delimiter):
...
-                      {'runTest': lambda self, f=filename: self.RunOneTest(f),
-                       'pattern': pattern})
-    setattr(module, test_class.__name__, test_class)
+def GenerateTests(rootdir, pattern):
+  def _GetTestBody(filename):
+    def _test(self):
+      logging.info('Testing iwyu on %s', filename)
+
+      # Split full/path/to/foo.cc into full/path/to/foo and .cc.
+      (all_but_extension, _) = os.path.splitext(filename)
```
kimgr, line 48: This is traditionally called the "stem" of the filename, so `stem` would be a good, short name.

kimgr: Ah, never mind. I see this code has just moved. I can follow up with a refactor after to improve naming here.

kimgr (commented):
This looks great! The commit message needs a little editing, it still mentions `C` and `Cxx` as class names, but otherwise this can land as soon as the flags-in-testfile changes have gone in.

storoj:
* lowercased `C` and `Cxx` in "Tests After" column;
* adopted the changes to 80-column rule;
* deleted unused `dirs` variable;

kimgr:
This looks and works great. Please rebase on top of latest master to get rid of the conflict, and I can merge. Thank you!

kimgr:
Thanks!

Issue #440, opened 2017-05-27
Please return different exit codes


LebedevRI:
A followup for #157
Looking at https://github.com/include-what-you-use/include-what-you-use/blob/13cd94804b140b06e0a58d2f949f13135a29e8d2/iwyu.cc#L3948-L3951, i'm guessing this was never actually merged, or was reverted.

It would be really nice if IWYU could return zero if it did not suggest anything, and anything else (e.g. 1) if there are any suggestions.
When used with https://cmake.org/cmake/help/v3.3/prop_tgt/LANG_INCLUDE_WHAT_YOU_USE.html, this would allow for compilation to be stopped if there are any issues.
This already works quite nicely with clang-tidy...

vsapsai:
I believe IWYU [exits with value 2 if there are no suggestions](https://github.com/include-what-you-use/include-what-you-use/blob/db2391c979ddd59e297f460011eb348cdcefd18f/iwyu.cc#L3525-L3526). Isn't it working for you?

LebedevRI:
`2`, exactly.
> It would be really nice if IWYU could return zero if it did not suggest anything, and anything else (e.g. 1) if there are any suggestions.

kimgr:
I'm beginning to think IWYU's original design, using a failure exit code to allow repeated re-analysis with `make -k`, has outlived its usefulness (see https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/InstructionsForUsers.md#plugging-into-make)

Does someone with more GNU Make chops see a way to simulate the above behavior without playing with the exit codes?

Bu11etmagnet:
I think that integration with the build system is desirable (not just make; it would work with `ninja -k0`). It is the easiest way to run iwyu with the necessary flags on all the sources of a project.

I see several possibilities:

- Always return a non-zero exit code; e.g. 1 for no suggestion and 2 if there were any suggestions.
- Return non-zero by default; return 0 or 1 if a certain command-line argument is supplied (e.g.  --descriptive-exit-codes). 

The we can have our cake and eat it too.

LebedevRI:
> It is the easiest way to run iwyu with the necessary flags on all the sources of a project.

Are you aware that it is possible to natively run IWYU from any cmake-generated project? (see link in first comment). I wonder how much people actually use it the recommended way (by replacing CC before calling make). I'm not sure this even works with auto-generated makefiles

kimgr:
[...]

PR #864, opened 2020-11-30
iwyu_tool.py: Notify success/failure via exit code

omeranson:
Modify the exit code of iwyu_tool.py so that it returns 0 only
when include-what-you-use does make any recommendations.

Related: #440

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -242,6 +242,10 @@ class Process(object):
         """
         return self.proc.poll()
 
+    @property
+    def returncode(self):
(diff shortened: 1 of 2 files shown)


kimgr:
Thank you, looks good!

Can I just ask that you use `[iwyu_tool]` instead of `iwyu_tool.py: ` in the commit message?

omeranson:
> SocksDirect: Datacenter Sockets can be Fast and Compatible

@kimgr Done

kimgr:
Thank you!

Issue #869, opened 2021-01-12
Add header mappings for Python(.h) / CPython


Flamefire:
For CPython you must only include <Python.h>, not <listobject.h>, <dictobject.h>, <pylifecycle.h> and the like.
Could this be added to the repo as a default?

kimgr:
Sure. If you have an environment up and running, could you sketch out a working `.imp`-file and open a PR?

Flamefire:
Not really firm with imp files so here is a list of headers pulled in when including `Python.h`, so maybe someone else could convert this to an imp file:

```
abstract.h
bltinmodule.h
boolobject.h
bytearrayobject.h
bytesobject.h
cellobject.h
ceval.h
classobject.h
codecs.h
code.h
compile.h
complexobject.h
context.h
descrobject.h
dictobject.h
dtoa.h
enumobject.h
eval.h
fileobject.h
fileutils.h
floatobject.h
funcobject.h
genobject.h
import.h
intrcheck.h
iterobject.h
listobject.h
longintrepr.h
longobject.h
memoryobject.h
methodobject.h
modsupport.h
moduleobject.h
namespaceobject.h
object.h
objimpl.h
odictobject.h
osmodule.h
patchlevel.h
pyarena.h
pyatomic.h
pycapsule.h
pyconfig.h
pyctype.h
pydebug.h
pyerrors.h
pyfpe.h
pyhash.h
pylifecycle.h
pymacconfig.h
pymacro.h
pymath.h
pymem.h
pyport.h
pystate.h
pystrcmp.h
pystrtod.h
Python.h
pythonrun.h
pythread.h
pytime.h
rangeobject.h
setobject.h
sliceobject.h
structseq.h
sysmodule.h
traceback.h
tupleobject.h
typeslots.h
unicodeobject.h
warnings.h
weakrefobject.h
```

Created with `echo '#include "Python.h"' | g++ -I /home/alex/.pyenv/versions/3.7.4/include/python3.7m/ -E - | grep "/home/alex/.pyenv/versions/3.7.4/include/python3.7m/" | awk '{print $3;}' | sort -u | xargs -n1 basename`

kimgr:
Thanks for the shell hack, that helps get me started.

I think what I want to do is;

- the equivalent of `grep -rh "#include \"" /usr/include/python3.8 | sort -u` to get a set of all Python headers included by other Python headers  (assuming `#include "xxx"` with quotes is only used for Python stuff)
[...]

PR #870, opened 2021-01-17
Add script to generate mappings for Python C API

kimgr:
For #869. @Flamefire could you test this in a real environment?

I see some semi-weird results when I run `include-what-you-use -Xiwyu --mapping_file=python3.8.imp -I /usr/include/python3.8 -xc /usr/include/python3.8/Python.h`, but I suspect it works better for a proper program using the Python C API.


Flamefire:
Tested and fails.

The file contains `{"include": ["@[\"<]patchlevel\\.h[\">]", "private", "<Python.h>", "public"]},`
But I get: 

```
/home/alex/git/scorep_binding_python/src/scorepy/events.cpp should add these lines:
#include <patchlevel.h>                    // for PY_MAJOR_VERSION

/home/alex/git/scorep_binding_python/src/scorepy/events.cpp should remove these lines:
- #include <Python.h>  // lines 2-2
```

Run with `iwyu_tool -p build src/scorepy/events.cpp --  --mapping_file=python3.7.imp` on https://github.com/score-p/scorep_binding_python after configuring with CMake (I hate setting up paths ;) )

kimgr:
If this is a reasonably recent `iwyu_tool.py`, you need to add `-Xiwyu`-prefix for suffix arguments:

    iwyu_tool -p build src/scorepy/events.cpp -- -Xiwyu --mapping_file=python3.7.imp

This was added to allow adding extra Clang arguments too.

Flamefire:
Ouch yes, it even says so: `error: unsupported option '--mapping_file=python3.7.imp'`

I'm surprised it doesn't just error out. Also the mapping_file param must be an absolute path: `iwyu_tool -p build src/scorepy/events.cpp -- -Xiwyu --mapping_file=$PWD/python3.7.imp` worked, so yes this works.
Hence I'd suggest to include a few of those generated ones with IWYU by default. I guess they won't vary to much by version if at all, so maybe 1 is enough.

For reference mine is attached: 
[python3.7.imp.txt](https://github.com/include-what-you-use/include-what-you-use/files/5921443/python3.7.imp.txt)

kimgr:
Thank you! Yeah, let's throw a python3 version in there, py2 is dead (as if :)).

I'll clean things up and push an updated branch.

kimgr:
This sets a new standard for mapping generators --

* Put them in directory mapgen/
* Name them `iwyu-mapgen-<library>.py`

These are not installed today, and I'm not sure we generally want to install them, so I'll just let the new one sit in this directory.

Will follow up with separate merge requests to move/rename existing mappers.

PR #877, opened 2021-02-04
Workaround missing libomptarget.so.13

storoj:
`llvm-13` is now the latest version, and the previous workaround does not work anymore.

It turned out that `libomptarget.so.12` is provided by `libomp-dev` package. That package is one of the recommended packages, however Travis uses `--no-install-recommends` option, and that package does not get installed.

Adding `libomp-dev` to the list of packages does not help because it seems that some version of `libomp-dev` (v5 probably) is pre-installed, so the recommended upgrade to v13 does not get installed.

This workaround explicitly installs `libomp-$VERSION-dev`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,6 @@ before_install:
  - PREFIX_PATH=`llvm-config-$VERSION --prefix`
  - BIN_PATH=`llvm-config-$VERSION --bindir`
 
- # Workaround for broken packaging.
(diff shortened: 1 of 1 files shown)


kimgr:
I think the problem is that there's no snapshot package for libomp-dev, just the versioned one called libomp-13-dev. That seems like an upstream bug. But I like this workaround, much nicer.

Issue #636, opened 2019-01-11
std::move does not require <utility> with template parameters


vmiklos:
Hi,

Consider the following reproducer:

````
#include <ostream>

template <typename> struct A {};

struct B
{
    template <typename T> B(A<T>&& a) {}
};

template <typename T> void f(A<T>&& a) { B(std::move(a)); }
````

Then:

````
$ include-what-you-use -std=gnu++17 bad.cxx 

bad.cxx should add these lines:

bad.cxx should remove these lines:
- #include <ostream>  // lines 1-1

The full include-list for bad.cxx:
---
````

Note how std::move() is used, but <utility> is not present and that missing include is not detected. But in case the parameter of f() is not a template:

````
#include <ostream>

struct A {};

struct B
{
    B(A&& a) {}
};

template <typename T> void f(A&& a) { B(std::move(a)); }
````

Then this is recognized correctly:

````
$ include-what-you-use -std=gnu++17 good.cxx 

good.cxx should add these lines:
#include <utility>  // for move

good.cxx should remove these lines:
- #include <ostream>  // lines 1-1

The full include-list for good.cxx:
#include <utility>  // for move
---
````

I.e. if you have <utility> in such an input by accident, then IWYU suggests to remove it, while it should not.

Version:

````
$ include-what-you-use --version            
include-what-you-use 0.11 based on clang version 7.0.1-svn348686-1~exp1~20181213084532.54 (branches/release_70)
````

Please let me know if any more information is needed. Also, if you have some code pointers, I might be able to try to fix this myself -- though from what I learned about the Clang AST so far suggests that anything related to templates is not easy. :-)

Thanks,

Miklos

**Comment by kimgr** [comment]
Sorry for not picking this up sooner.

It looks like the dependent template arg mangles the AST completely:
```
`-FunctionTemplateDecl 0x55bcfe7343f0 <line:10:1, col:59> col:28 f
  |-TemplateTypeParmDecl 0x55bcfe734138 <col:11, col:20> col:20 referenced typename depth 0 index 0 T
  `-FunctionDecl 0x55bcfe734358 <col:23, col:59> col:28 f 'void (A<T> &&)'
    |-ParmVarDecl 0x55bcfe7342a0 <col:30, col:37> col:37 referenced a 'A<T> &&'
    `-CompoundStmt 0x55bcfe734550 <col:40, col:59>
[...]

PR #883, opened 2021-02-25
Add libstdc++-10 related mappings

saki7:
Checked for my application using C++20 flags, with Clang 13.

---

__Issue described below has been moved to: #884__

~~Although most of the diff is trivial, I still get some error in my code:~~

> ~~#include <filesystem>~~
>
> ~~void clean_file(const std::filesystem::path& path)~~
> ~~{~~
>      ~~remove(path); // ADL~~
> ~~}~~

~~`std::filesystem::path` and `std::filesystem::remove` both sit in the same namespace, `std::filesystem`. So calling the `remove` function without namespace qualification should trigger ADL. My code correctly compiles without any compiler warning on my actual build.~~

~~However IWYU suggests including `<stdio.h>` for the function `remove`. It is obvious that [`remove(const char*)`](https://en.cppreference.com/w/c/io/remove) (in global namespace) should not be found even with ADL; the more matching signature is [`remove(const std::filesystem::path&)`](https://en.cppreference.com/w/cpp/filesystem/remove) (in namespace std::filesystem). In IWYU, where does this strange suggestion come from?~~

~~The ADL issue seem unrelated to this specific PR, so I omitted the workaround inside mapping update. I guess there exists some nasty pitfalls for ADL detection in IWYU. Maybe we could discuss that in another issue?~~

### DIFF
--- a/gcc.stl.headers.imp
+++ b/gcc.stl.headers.imp
@@ -1,7 +1,9 @@
 # GCC STL headers
 [
+  # Note: make sure to sync this setting with iwyu_include_picker.cc
+
   # Headers explicitly annotated with `@headername`
-  # ( cd /usr/include/c++/8 && grep -r headername | perl -nle 'm/^([^:]+).*@headername\{([^,]*)\}/ && print qq@  { include: ["<$1>", private, "<$2>", public ] },@' | sort -u )
+  # ( cd /usr/include/c++/10 && grep -r headername | perl -nle 'm/^([^:]+).*@headername\{([^,]*)\}/ && print qq@  { include: ["<$1>", private, "<$2>", public ] },@' | sort -u )
   { include: ["<backward/auto_ptr.h>", private, "<memory>", public ] },
   { include: ["<backward/backward_warning.h>", private, "<iosfwd>", public ] },
   { include: ["<backward/binders.h>", private, "<functional>", public ] },
@@ -17,6 +19,7 @@
   { include: ["<bits/basic_string.tcc>", private, "<string>", public ] },
   { include: ["<bits/boost_concept_check.h>", private, "<iterator>", public ] },
   { include: ["<bits/c++0x_warning.h>", private, "<iosfwd>", public ] },
+  { include: ["<bits/charconv.h>", private, "<charconv>", public ] },
   { include: ["<bits/char_traits.h>", private, "<string>", public ] },
   { include: ["<bits/codecvt.h>", private, "<locale>", public ] },
   { include: ["<bits/concept_check.h>", private, "<iterator>", public ] },
@@ -38,9 +41,11 @@
   { include: ["<bits/gslice.h>", private, "<valarray>", public ] },
   { include: ["<bits/hash_bytes.h>", private, "<functional>", public ] },
   { include: ["<bits/indirect_array.h>", private, "<valarray>", public ] },
+  { include: ["<bits/int_limits.h>", private, "<limits>", public ] },
   { include: ["<bits/invoke.h>", private, "<functional>", public ] },
   { include: ["<bits/ios_base.h>", private, "<ios>", public ] },
   { include: ["<bits/istream.tcc>", private, "<istream>", public ] },
+  { include: ["<bits/iterator_concepts.h>", private, "<iterator>", public ] },
   { include: ["<bits/list.tcc>", private, "<list>", public ] },
   { include: ["<bits/locale_classes.h>", private, "<locale>", public ] },
   { include: ["<bits/locale_classes.tcc>", private, "<locale>", public ] },
@@ -64,6 +69,10 @@
   { include: ["<bits/random.h>", private, "<random>", public ] },
   { include: ["<bits/random.tcc>", private, "<random>", public ] },
   { include: ["<bits/range_access.h>", private, "<iterator>", public ] },
+  { include: ["<bits/range_cmp.h>", private, "<functional>", public ] },
+  { include: ["<bits/ranges_algobase.h>", private, "<algorithm>", public ] },
+  { include: ["<bits/ranges_algo.h>", private, "<algorithm>", public ] },
+  { include: ["<bits/ranges_uninitialized.h>", private, "<memory>", public ] },
   { include: ["<bits/refwrap.h>", private, "<functional>", public ] },
   { include: ["<bits/regex_automaton.h>", private, "<regex>", public ] },
   { include: ["<bits/regex_automaton.tcc>", private, "<regex>", public ] },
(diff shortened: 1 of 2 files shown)


kimgr:
Thanks for the updated mappings!

Unfortunately we carry them in two places -- in gcc.stl.headers.imp and inline inside `iwyu_include_picker.cc`. Could you modify both?

Re the dual `remove`, I wonder if that's somehow related to:
https://github.com/include-what-you-use/include-what-you-use/issues/636

Let's open a separate issue for that, and maybe you can investigate the AST with `-Xiwyu -v7` and see if it's possible to identify where IWYU is going wrong?

saki7:
@kimgr I applied updates to both `gcc.stl.headers.imp` and `iwyu_include_picker.cc`.

Most of the standard headers should be backward-compatible, so I assume that hard-coding C++20 headers in IWYU does not lead to real world issues. If you are happy with this decision, please go ahead and merge this PR.

For now I've chosen C++20 because it's the latest published standard. I used the `c++20` branch from `cplusplus/draft` repo, to generate the list. If we are going to be pedantic then we expect IWYU to automatically detect user's C++ language version flags at the IWYU invocation, so it switches the header resolution dynamically. But that's more like a new feature request, so I'm just sharing my idea here.

p.s. I will split the ADL bug into different issue.

gcc.stl.headers.imp
```diff
@@ -157,15 +168,16 @@
   { include: ["<tr1/unordered_map.h>", private, "<tr1/unordered_map>", public ] },
   { include: ["<tr1/unordered_set.h>", private, "<tr1/unordered_set>", public ] },
   { include: ["<tr2/dynamic_bitset.tcc>", private, "<tr2/dynamic_bitset>", public ] },
-  # ( cd /usr/include/x86_64-linux-gnu/c++/8 && grep -r headername | perl -nle 'm/^([^:]+).*@headername\{([^,]*)\}/ && print qq@  { include: ["<$1>", private, "<$2>", public ] },@' | sort -u )
+  # ( cd /usr/include/x86_64-linux-gnu/c++/10 && grep -r headername | perl -nle 'm/^([^:]+).*@headername\{([^,]*)\}/ && print qq@  { include: ["<$1>", private, "<$2>", public ] },@' | sort -u )
   { include: ["<bits/basic_file.h>", private, "<ios>", public ] },
   { include: ["<bits/c++allocator.h>", private, "<memory>", public ] },
-  { include: ["<bits/c++config.h>", private, "<iosfwd>", public ] },
```
kimgr, line 163: This one is a little scary -- I wonder if we should map this to both `<iosfwd>` _and_ `<version>` (I'm guessing `<version>` is new in libstdc++10).

gcc.stl.headers.imp
```diff
@@ -300,4 +312,12 @@
   # The location of exception_defines.h varies by GCC version.  It should
   # never be included directly.
   { include: ["<exception_defines.h>", private, "<exception>", public ] },
+
+  # post libstdc++-10 stuffs which are not automatically caught by commands above
```
kimgr, line 316: stuffs -> stuff
which are -> which is

iwyu_include_picker.cc
```diff
@@ -473,9 +473,11 @@ const IncludeMapEntry stdlib_c_include_map[] = {
   // mapping.)  Here is how I identified the files to map:
   // $ for i in /usr/include/c++/4.4/c* ; do ls /usr/include/`basename $i | cut -b2-`.h /usr/lib/gcc/*/4.4/include/`basename $i | cut -b2-`.h 2>/dev/null ; done
   //
-  // These headers are defined in C++14 [headers]p3.  You can get them with
-  // $ sed -n '/begin{floattable}.*{tab:cpp.c.headers}/,/end{floattable}/p' lib-intro.tex | grep tcode | perl -nle 'm/tcode{<c(.*)>}/ && print qq@  { "<$1.h>", kPublic, "<c$1>", kPublic },@' | sort
-  // on https://github.com/cplusplus/draft/blob/master/source/lib-intro.tex
+  // These headers are defined in [headers.cpp.c].
+  // https://github.com/cplusplus/draft/blob/c+%2B20/source/lib-intro.tex
+  //
+  // $ wget https://raw.githubusercontent.com/cplusplus/draft/c%2B%2B20/source/lib-intro.tex
```
kimgr, line 479: Nice! Could you fold this into a one-liner so the instruction doesn't require disk for lib-intro.tex? E.g.
```
$ curl https://raw.githubusercontent.com/cplusplus/draft/c%2B%2B20/source/lib-intro.tex | sed -n '/begin{multicolfloattable}.*{headers.cpp.c}/,/end{multicolfloattable}/p' ...
```

iwyu_include_picker.cc
```diff
@@ -505,19 +507,31 @@ const IncludeMapEntry stdlib_c_include_map[] = {
 };
 
 const char* stdlib_cpp_public_headers[] = {
-  // These headers are defined in C++14 [headers]p2.  You can get them with
-  // $ sed -n '/begin{floattable}.*{tab:cpp.library.headers}/,/end{floattable}/p' lib-intro.tex | grep tcode | perl -nle 'm/tcode{(.*)}/ && print qq@  "$1",@' | sort
-  // on https://github.com/cplusplus/draft/blob/master/source/lib-intro.tex
+  // These headers are defined in [headers.cpp].
+  // https://github.com/cplusplus/draft/blob/c+%2B20/source/lib-intro.tex
+  //
+  // $ wget https://raw.githubusercontent.com/cplusplus/draft/c%2B%2B20/source/lib-intro.tex
```
kimgr, line 513: Use curl to avoid storing lib-intro.text per above

iwyu_include_picker.cc
```diff
@@ -721,10 +761,10 @@ const IncludeMapEntry libstdcpp_include_map[] = {
   { "<tr1/unordered_map.h>", kPrivate, "<tr1/unordered_map>", kPublic },
   { "<tr1/unordered_set.h>", kPrivate, "<tr1/unordered_set>", kPublic },
   { "<tr2/dynamic_bitset.tcc>", kPrivate, "<tr2/dynamic_bitset>", kPublic },
-  // cd /usr/include/x86_64-linux-gnu/c++/8 && grep -r headername | perl -nle 'm/^([^:]+).*@headername\{([^,]*)\}/ && print qq@  { "<$1>", kPrivate, "<$2>", kPublic },@' | sort -u
+  // cd /usr/include/x86_64-linux-gnu/c++/10 && grep -r headername | perl -nle 'm/^([^:]+).*@headername\{([^,]*)\}/ && print qq@  { "<$1>", kPrivate, "<$2>", kPublic },@' | sort -u
   { "<bits/basic_file.h>", kPrivate, "<ios>", kPublic },
   { "<bits/c++allocator.h>", kPrivate, "<memory>", kPublic },
-  { "<bits/c++config.h>", kPrivate, "<iosfwd>", kPublic },
```
kimgr, line 727: Again with mapping to both. I think that would make sense.

kimgr (commented):
Looks great. Some ideas for improvement inline.

saki7:
Force-pushed. I believe all issues have been resolved.

kimgr:
Thanks, LGTM!

PR #887, opened 2021-03-07
Encourage use of std::size_t in C++

saki7:
iwyu was previously suggesting `<iosfwd>` for `std::size_t` (which works because of transitive includes on standard headers, but due to the spec it's not guaranteed to work always). 

- <https://timsong-cpp.github.io/cppwp/n4861/support>
- <https://timsong-cpp.github.io/cppwp/n4861/iosfwd.syn#:%3ciosfwd%3e>

iwyu now suggests either `<iosfwd>` or `<version>` depending on the code's context, see #883

This pr resolves `std::size_t` to appropriate C++ headers. Same kind of workaround had already been implemented for C, but C++ support was lacking. Also iwyu's internal implementation is modified to use `std::size_t` instead of `size_t`.

### DIFF
--- a/gcc.symbols.imp
+++ b/gcc.symbols.imp
@@ -169,5 +169,13 @@
   # string/ostream/istream type.
   { symbol: [ "std::char_traits", private, "<string>", public ] },
   { symbol: [ "std::char_traits", private, "<ostream>", public ] },
-  { symbol: [ "std::char_traits", private, "<istream>", public ] }
+  { symbol: [ "std::char_traits", private, "<istream>", public ] },
+
+  { symbol: [ "std::size_t", private, "<cstddef>", public ] },
+  { symbol: [ "std::size_t", private, "<cstdio>", public ] },
+  { symbol: [ "std::size_t", private, "<cstdlib>", public ] },
+  { symbol: [ "std::size_t", private, "<cstring>", public ] },
+  { symbol: [ "std::size_t", private, "<ctime>", public ] },
+  { symbol: [ "std::size_t", private, "<cuchar>", public ] },
+  { symbol: [ "std::size_t", private, "<cwchar>", public ] }
 ]
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -252,6 +252,14 @@ const IncludeMapEntry libstdcpp_symbol_map[] = {
   { "std::char_traits", kPrivate, "<string>", kPublic },
   { "std::char_traits", kPrivate, "<ostream>", kPublic },
   { "std::char_traits", kPrivate, "<istream>", kPublic },
+
+  { "std::size_t", kPrivate, "<cstddef>", kPublic },  // 'canonical' location for std::size_t
+  { "std::size_t", kPrivate, "<cstdio>", kPublic },
+  { "std::size_t", kPrivate, "<cstdlib>", kPublic },
+  { "std::size_t", kPrivate, "<cstring>", kPublic },
+  { "std::size_t", kPrivate, "<ctime>", kPublic },
+  { "std::size_t", kPrivate, "<cuchar>", kPublic },
+  { "std::size_t", kPrivate, "<cwchar>", kPublic },
 };
 
 // Private -> public include mappings for GNU libc
--- a/tests/cxx/std_size_t.cc
+++ b/tests/cxx/std_size_t.cc
@@ -1,0 +1,39 @@
+//===--- std_size_t.cc - test input file for iwyu -------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// This tests a little quirk with the mappings for the new <cstd...> headers.
+// size_t is such a commonly-used type that it deserves a dedicated test to
+// demonstrate how this behaves in general. Note how the use of 'printf' is
+// attributed to <cstdio> and <stdio.h> is removed. This happens because
+// std::size_t is mapped to <cstdio> before 'printf' is seen and already-needed
+// includes take precedence if a symbol is available from multiple sources.
+//
+// (Note that if 'printf' was seen before std::size_t, <stdio.h> would still be
+// required.)
+
+#include <cstdio>
+#include <stdio.h>
+
+std::size_t f() {
+  const size_t x = 100;
+  printf("%zu\n", x);
+  return sizeof(int);
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/std_size_t.cc should add these lines:
+
+tests/cxx/std_size_t.cc should remove these lines:
+- #include <stdio.h>  // lines XX-XX
+
+The full include-list for tests/cxx/std_size_t.cc:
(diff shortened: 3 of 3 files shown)


kimgr:
Thank you!

I take issue with the wholesale switch to `std::size_t` in IWYU, but otherwise this looks good. Could you revert those changes (including the test suite), please?

Also, generally it makes sense to split that kind of editorial change into a separate commit from functional changes.

saki7:
No offense, but those changes are not intended to be treated like editorial changes. Those comment indicates that the iwyu code itself is also iwyu-ed, so I included them in single commit.

Also, I'm seeing an enigmatic test failures on Travis; do you really think the IWYU codebase modification should be reverted?

https://travis-ci.org/github/include-what-you-use/include-what-you-use/builds/761811439

p.s. why is Travis not showing up on GitHub PR? It's strange.

kimgr:
> Those comment indicates that the iwyu code itself is also iwyu-ed, so I included them in single commit.

Not regularly, but yeah, I'm trying to get to a point where that's possible again (some troubles with LLVM headers).

> Also, I'm seeing an enigmatic test failures on Travis; do you really think the IWYU codebase modification should be reverted?

It's a little hard to make out from that enormous results page, but I think the relevant portion is:
```
FAIL: test_badinc (__main__.cxx)
  File "run_iwyu_tests.py", line 67, in _test
    verbose=True)
  File "/home/travis/build/include-what-you-use/include-what-you-use/iwyu_test_util.py", line 477, in TestIwyuOnRelativeFile
    test_case.assertTrue(not failures, ''.join(failures))
AssertionError: 
tests/cxx/badinc.cc:1130: Unexpected diagnostic:
fprintf is defined in <cstdio>, which isn't directly #included.
tests/cxx/badinc.cc:1130: Unexpected diagnostic:
stdout is defined in <cstdio>, which isn't directly #included.
tests/cxx/badinc.cc:1130: Unmatched regex:
fprintf is...*<stdio.h>
tests/cxx/badinc.cc:1130: Unmatched regex:
stdout is...*<stdio.h>
tests/cxx/badinc.cc:1132: Unexpected diagnostic:
printf is defined in <cstdio>, which isn't directly #included.
tests/cxx/badinc.cc:1132: Unmatched regex:
printf is...*<stdio.h>
+++ 
@@ -2,6 +2,7 @@
 #include <ctype.h>
 #include <stdarg.h>
 #include <stddef.h>
+#include <cstdio>
 #include <list>
 #include "tests/cxx/badinc-i1.h"
 class D2_Class;
@@ -33,6 +34,7 @@
 #include <stdarg.h>  // for va_list
 #include <stddef.h>  // for offsetof
 #include <algorithm>  // for find
+#include <cstdio>  // for fprintf, printf, size_t, stdout
 #include <fstream>  // for fstream
 #include <list>  // for list
 #include <string>  // for basic_string, basic_string<>::iterator, operator+, string
---
```
So since `size_t` is now mapped into `<cstdio>`, the tests now attribute `fprintf`, `printf` and `stdout` to `<cstdio>` too.

It would be nice if we could allow both old C-style spelling (`<stdio.h>`) and new C++ spelling (`<cstdio>`) so that code that's ostensibly correct (including C headers) would continue to work without IWYU forcing the C++ wrapper headers.

> I'm having a feeling that the failures will increase if I revert the changes partially.

I don't think so, the tests don't run on IWYU's source code, only its test suite under `tests/`.

> p.s. why is Travis not showing up on GitHub PR? It's strange.

I know, I just noticed that too :-(. I've tried reconnecting travis with github, but so far failed.

saki7:
OK. I reverted those changes, Travis is happy. https://travis-ci.org/github/include-what-you-use/include-what-you-use/builds/761815023

I think you can merge this now, but if you want me to merge the 2 commits, then I'll force-push.

kimgr:
Maybe you could add something like this to help explain the expected behavior:
```
commit 51d07118f72bfa5b1403543df7d9db3ec0b4f49b
Author: Kim Grasman <kim.grasman@gmail.com>
Date:   Sun Mar 7 20:05:22 2021 +0100

    Add basic test for std::size_t mappings

diff --git a/tests/cxx/std_size_t.cc b/tests/cxx/std_size_t.cc
new file mode 100644
index 0000000..9f9824a
--- /dev/null
+++ b/tests/cxx/std_size_t.cc
@@ -0,0 +1,39 @@
+//===--- std_size_t.cc - test input file for iwyu -------------------------===//
+//
+//
+//
+
+// This tests a little quirk with the mappings for the new <cstd...> headers.
+// size_t is such a commonly-used type that it deserves a dedicated test to
+// demonstrate how this behaves in general. Note how the use of 'printf' is
+// attributed to <cstdio> and <stdio.h> is removed. This happens because
+// std::size_t is mapped to <cstdio> before 'printf' is seen and already-needed
+// includes take precedence if a symbol is available from multiple sources.
+//
+// (Note that if 'printf' was seen before std::size_t, <stdio.h> would still be
+// required.)
+
+#include <cstdio>
+#include <stdio.h>
+
+std::size_t f() {
+  const size_t x = 100;
+  printf("%zu\n", x);
+  return sizeof(int);
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/std_size_t.cc should add these lines:
+
+tests/cxx/std_size_t.cc should remove these lines:
+- #include <stdio.h>  // lines XX-XX
+
+The full include-list for tests/cxx/std_size_t.cc:
+#include <cstdio>  // for printf, size_t
+
+***** IWYU_SUMMARY */
```

And yeah, squash the two commits, they don't make sense without each other.

gcc.symbols.imp
```diff
@@ -169,5 +169,13 @@
...
-  { symbol: [ "std::char_traits", private, "<istream>", public ] }
+  { symbol: [ "std::char_traits", private, "<istream>", public ] },
+
+  { symbol: [ "std::size_t", private, "<cstddef>", public ] },
+  { symbol: [ "std::size_t", private, "<cstdio>", public ] },
+  { symbol: [ "std::size_t", private, "<cstdlib>", public ] },
+  { symbol: [ "std::size_t", private, "<cstring>", public ] },
+  { symbol: [ "std::size_t", private, "<ctime>", public ] },
+  { symbol: [ "std::size_t", private, "<cuchar>", public ] },
+  { symbol: [ "std::size_t", private, "<cwchar>", public ] }
```
kimgr, line 180: Could you add a trailing comma here in the interest of keep the next patch smaller? Thanks.

kimgr:
@saki7 Are you interested in continuing to work on this?

gcc.symbols.imp
kimgr: Actually, I think that's invalid in yaml/json, so that's probably a bad idea :)

saki7:
I am currently busy on developing duties. I have not abandoned this pr, but the fact is that I don't have enough time to work on this. I believe that the pr's concept is fairly simple, and remaining works are minimal (just squashing) so feel free to decide your next action as a maintainer.

I have checked the "Allow edits by maintainers" checkbox so a maintainer have write access to this specific branch.

kimgr:
@saki7 Thanks, I'll try to finish this up.

saki7:
@kimgr Thank you for taking care of this PR, and sorry for my absence.

PR #888, opened 2021-03-11
Add --output option to export report to file

insilications:
If tests are necessary, I could add them.

### DIFF
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -16,6 +16,7 @@
 #include <set>                          // for set
 #include <string>                       // for string, operator<, etc
 #include <utility>                      // for make_pair, pair
+#include <fstream>                      // for ofstream, basic_ostream<>::__...
 
 #include "iwyu_cache.h"
 #include "iwyu_include_picker.h"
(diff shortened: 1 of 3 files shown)


kimgr:
Why is this preferable to `include-what-you-use t.c > x.log 2>&1`?

insilications:
@kimgr Using IWYU with CMake, for example:

```
set(CMAKE_CXX_INCLUDE_WHAT_YOU_USE "/usr/bin/include-what-you-use;-Xiwyu;--verbose=1;-Xiwyu;--mapping_file=/usr/share/include-what-you-use/qt5_15.imp;-Xiwyu;--cxx17ns;>fixes.log2>&1")
cmake" --build /insilications/apps/Test/Test_Clang-Debug --target all --parallel 16
```

Doesn't work. The output is never redirected to fixes.log.

`set(CMAKE_CXX_INCLUDE_WHAT_YOU_USE "/usr/bin/include-what-you-use;-Xiwyu;--verbose=1;-Xiwyu;--mapping_file=/usr/share/include-what-you-use/qt5_15.imp;-Xiwyu;--cxx17ns;-Xiwyu;>fixes.log2>&1")`

Doesn't work either.

With my solution, the intended behavior is perfect:

`set(CMAKE_CXX_INCLUDE_WHAT_YOU_USE "/usr/bin/include-what-you-use;-Xiwyu;--verbose=1;-Xiwyu;--mapping_file=/usr/share/include-what-you-use/qt5_15.imp;-Xiwyu;--cxx17ns;-Xiwyu;--output=fixes.log")`

storoj:
Hi @insilications, if I understand it right `CMAKE_CXX_INCLUDE_WHAT_YOU_USE` command is being applied to every file in the project. If so then every `IWYU` run will overwrite the same file with an output for every source file. It seems that you open the output file for appending (`std::ios::app`), but how do you guarantee that the file will be empty before `cmake` starts calling IWYU? I am also wondering if appending the file from multiple IWYU processes can work reliably.

kimgr:
@insilications I think @storoj makes some good points wrt concurrency.

The patch is nice and simple, but on plain gut feeling it seems weird to me that IWYU would need file-output if compilers in general don't, so I'd prefer not to do it this way.

As for running with cmake -- the question then becomes: why not just `cmake --build . > fixes.log 2>&1`?

insilications:
Hi guys, sorry for the delay. I'm having some real life problems right now. I agree with the problems wrt concurrency, although in my current use case they haven't showed up. But they are definitely real and need to be addressed. I'll be answering you guys later.

kimgr:
Bundled into #1563

PR #891, opened 2021-03-30
[iwyu_tool] Add support for load average limiting via --load

Flowdalic:
If "--load N" is given, potentially limit the capacity of new jobs. If
the load_capacity is equal or less than one but there are no pending
jobs, then ensure that at least one job is running. Otherwise, cap the
capacity to load_capacity.


kimgr:
Nice, looks good to me. Thank you!

PR #894, opened 2021-04-07
Glob absolute

### LINKED ISSUE #893: Use absolute paths for glob comparison
Currently file names as used internally by llvm/clang are used for glob comparison

https://github.com/include-what-you-use/include-what-you-use/blob/6e6b06ed13bc5f592d89ad5d4e397f64b64a0bea/iwyu_location_util.h#L90-L93

It would be better to translate this to the absolute path. This way prefix matching would work (e.g. `--check_also=/path/to/project/*`).

darkdragon-001:
Fixes #893

### DIFF
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -383,9 +383,11 @@ void AddGlobToReportIWYUViolationsFor(const string& glob) {
 
 bool ShouldReportIWYUViolationsFor(const clang::FileEntry* file) {
   const string filepath = GetFilePath(file);
-  for (const string& glob : GlobalFlags().check_also)
-    if (GlobMatchesPath(glob.c_str(), filepath.c_str()))
+  for (const string& glob : GlobalFlags().check_also) {
+    std::string cmp_filepath = IsAbsolutePath(glob) ? MakeAbsolutePath(filepath) : filepath;
+    if (GlobMatchesPath(glob.c_str(), cmp_filepath.c_str()))
       return true;
+  }
   return false;
 }
 
--- a/iwyu_location_util.h
+++ b/iwyu_location_util.h
@@ -89,7 +89,7 @@ bool IsInScratchSpace(clang::SourceLocation loc);
 
 inline string GetFilePath(const clang::FileEntry* file) {
   return (IsBuiltinFile(file) ? "<built-in>" :
-          NormalizeFilePath(file->getName().str()));
+          file->getName().str());
 }
 
 //------------------------------------------------------------
--- a/iwyu_path_util.cc
+++ b/iwyu_path_util.cc
@@ -129,12 +129,7 @@ string NormalizeFilePath(const string& path) {
   llvm::SmallString<128> normalized(path);
   llvm::sys::path::remove_dots(normalized, /*remove_dot_dot=*/true);
 
-#ifdef _WIN32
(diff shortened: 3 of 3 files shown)


kimgr:
Your branch appears to be based on `clang_11`; can you rebase it on `master` instead? Thanks.

I'll see if I can get CI working again, Travis seems to be having trouble reporting back to pull requests.

darkdragon-001:
@kimgr I rebased on `master` branch.

kimgr:
Thanks! I managed to cook up a CI build native to GitHub, so if you rebase on latest master you should get that as part of your pull request as well.

darkdragon-001:
@kimgr I rebased on latest master.

darkdragon-001:
The tests are failing since output prints full paths now as well. I see it as a feature, since Ctrl+Click only opens files with full paths in editors like VSCode.

kimgr:
@darkdragon-001 Obviously I can't take any patches that break the test suite... Do you have any ideas for how to repair it?

kimgr:
I wonder if you're doing this absolute-path canonicalization in the wrong place -- is it possible to make it absolute where the glob matching happens instead?

I think what I'd like the most is something like:

```
bool ShouldReportIWYUViolationsFor(const clang::FileEntry* file) {
  const string filepath = GetFilePath(file);
  for (const string& glob : GlobalFlags().check_also) {
    std::string comparable_filepath
    if (IsAbsolutePath(glob)) {
        cmp_filepath = MakeAbsolute(filepath);
    } else {
        cmp_filepath = filepath;
    }

    if (GlobMatchesPath(glob.c_str(), cmp_filepath.c_str()))
      return true;
  }
  return false;
}
```

(just to give an idea, there are more elegant ways of doing this, I'm sure)

So if the glob is absolute, make the filepath absolute, otherwise compare them as relative.

Not sure about the viability of that approach, but it seems less invasive to me.

darkdragon-001:
@kimgr I first tried to do it there, but then noticed that it doesn't work beause `GetFilePath()` calls `NormalizeFilePath()` which in terms calls `llvm::sys::path::remove_dots(normalized, /*remove_dot_dot=*/true)` aka removing preceding `..` which are needed for the correct absolute paths.

When calling the `MakieAbsolutePath()` within `GetFilePath()`, I noticed that the output changed from relative to absolute paths (that's why the tests are failing). Actually I see this as an improvement instead of a regression since all tools working with the files should still work but additionally IDEs like VSCode allow Ctrl+Click to directly open the file in the editor. This is really helpful and speeds up the process for manual fixing based on the output of the tool. That's why I would prefer this behavior. If this is fine, I will think of a way to modify the tests accordingly. Or do you see any problems with the changed output?

darkdragon-001:
When removing `NormalizeFilePath()` from `GetFilePath()`, your suggestion should work for the glob. I think it is better to do one thing at a time anyways. What do you think?

Let's solve the output of absolute file names to the console for VSCode (as described in https://github.com/include-what-you-use/include-what-you-use/issues/893#issuecomment-831278560) in a separate issue.

darkdragon-001:
@kimgr Can you have a look at why the tests fail?

darkdragon-001:
I do not plan to continue this PR as we are no longer using IWYU as it didn't work for our use case.

kimgr:
@darkdragon-001 Thanks for the heads-up!

PR #899, opened 2021-04-12
[iwyu_tool] Allow float as -j argument

Flowdalic:
Most tools, e.g. GNU Make or ninja, allow the -j argument to be
provided as floating point value. Hence it is sensible that we allow
this too.


kimgr:
@Flowdalic Did you mean `-l`? That's what the diff does, and that seems to match other tools.

Flowdalic:
> @Flowdalic Did you mean `-l`? That's what the diff does, and that seems to match other tools.

Right, of course I mixed the two up. At least I was consistent doing so. :)

Fixed

kimgr:
Nice, thanks!

Issue #790, opened 2020-05-04
Feature request: iwyu_tool.py should exit with a non-zero exit value if it has detected include issues


Flowdalic:
As of now I have to check `iwyu_tool.py`'s output to determine if there are any issues with the include. Which not only requires extra work, it is also fragile, as the output the tool emits in case there are include issues may change.

It would be great, if  `iwyu_tool.py` would simply exit with a non-zero exit value if it has detected include issues.

kimgr:
I think it boils back down to the confusion in `include-what-you-use` itself -- by design it always returns a failure exit code, so that `make -k` can be used to re-run it.

I'm not convinced that decision was a good one, so maybe `iwyu-tool.py` should at least try and report success/failure consistently.

I think we should do something about `include-what-you-use` itself too, but I'm not sure what, especially not to maintain backward compat. Maybe a `--fail-always` switch to produce the current behavior.

Flowdalic:
> Maybe a `--fail-always` switch to produce the current behavior.

If you are concerned about backwards compatibility, which is always good, then a `--fail-only-on-iwyu-error` switch is probably a better idea.

kimgr:
Yeah, but I think the current behavior is weird enough that it warrants explicit user action to preserve :)

mortie:
I just decided to check out IWYU, and this bit me too. I would love to require `make iwyu` to return 0 in order to commit, and check that sort of stuff in CI, but that's really hard due to the current design.

I use GNU make, so the `make -k` hack is unnecessary to me since `-B` exists which seems to do the job better. It would probably be best to just require `--fail-always` for the special case of BSD make users who want to run `make CC=iwyu` ad-hoc from the command line, since it makes the tool so much more useful for everyone (including BSD make users).

PR #904, opened 2021-04-20
Add flag to exit with exit code 0 on success

mortie:
As discussed in https://github.com/include-what-you-use/include-what-you-use/issues/440 and https://github.com/include-what-you-use/include-what-you-use/issues/790, there's a need for IWYU to exit with code 0 if there are no suggested edits, and a non-zero exit code if there are suggested edits. This PR adds the `--exit_code` flag to do just that.

Here are some considerations:

* I don't know if the name `--exit_code` is ideal. I used it because that's what other tools, such as `git diff`, uses to make the exit code useful. Maybe it should be called `--fail_only_on_error` or something, I don't know. I'm open to suggestions.
* I don't like having both `EXIT_SUCCESSFULLY` and `EXIT_SUCCESS_OFFSET`, but I'm having trouble coming up with better names.
* It might be better to switch the default state of the option, so that the default behavior is to exit with code 0 if there are no suggested edits, and passing `--always_fail` makes it exit with code `2 + num_edits`. That would certainly be more in line with what people expect, since people expect a tool to exit with code 0 when there were no errors. I can imagine people spending a long time trying to figure out what's wrong with their makefile, without realizing that it's actually IWYW that behaves weirdly, not their makefile that's wrong. On the other hand, it would break backwards compatibility.

Also, I think I noticed a bug in the existing exit handling code: `exit(EXIT_SUCCESS_OFFSET + num_edits);` will exit with exit code 0 if `num_edits + 2` is a multiple of 256. I don't know if that's realistically ever going to happen.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3695,8 +3695,10 @@ class IwyuAstConsumer
     // unless --exit_code is specified
     if (GlobalFlags().exit_code && num_edits == 0)
       exit(EXIT_SUCCESSFULLY);
-    else
+    else if (EXIT_SUCCESS_OFFSET + num_edits <= EXIT_MAX)
       exit(EXIT_SUCCESS_OFFSET + num_edits);
+    else
+      exit(EXIT_MAX);
   }
 
   void ParseFunctionTemplates(TranslationUnitDecl* decl) {
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3691,8 +3691,12 @@ class IwyuAstConsumer
     num_edits += preprocessor_info().FileInfoFor(main_file)
         ->CalculateAndReportIwyuViolations();
 
-    // We need to force the compile to fail so we can re-run.
-    exit(EXIT_SUCCESS_OFFSET + num_edits);
+    // We need to force the compile to fail so we can re-run,
+    // unless --exit_code is specified
+    if (GlobalFlags().exit_code && num_edits == 0)
+      exit(EXIT_SUCCESSFULLY);
+    else
+      exit(EXIT_SUCCESS_OFFSET + num_edits);
   }
 
   void ParseFunctionTemplates(TranslationUnitDecl* decl) {
@@ -4162,5 +4166,7 @@ int main(int argc, char **argv) {
   // We always return a failure exit code, to indicate we didn't
   // successfully compile (produce a .o for) the source files we were
   // given.
+  // Note: If --exit_code was passed, we want to exit with code 0
+  // if there are no edits. That's handled in HandleTranslationUnit.
   return 1;
 }
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -96,6 +96,7 @@ static void PrintHelp(const char* extra_msg) {
          "   --quoted_includes_first: when sorting includes, place quoted\n"
          "        ones first.\n"
          "   --cxx17ns: suggests the more concise syntax introduced in C++17\n"
+         "   --exit_code: Return 0 if no issues were found\n"
          "\n"
          "In addition to IWYU-specific options you can specify the following\n"
          "options without -Xiwyu prefix:\n"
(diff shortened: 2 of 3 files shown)


iwyu.cc
```diff
@@ -3691,8 +3691,12 @@ class IwyuAstConsumer
...
         ->CalculateAndReportIwyuViolations();
 
-    // We need to force the compile to fail so we can re-run.
-    exit(EXIT_SUCCESS_OFFSET + num_edits);
+    // We need to force the compile to fail so we can re-run,
+    // unless --exit_code is specified
+    if (GlobalFlags().exit_code && num_edits == 0)
+      exit(EXIT_SUCCESSFULLY);
+    else
+      exit(EXIT_SUCCESS_OFFSET + num_edits);
```
EdSchouten, line 3699: Even though `exit()` takes an `int`, functions like `wait()` and `waitpid()` only guarantee that you can get the bottom eight bits. `waitid()` has addressed this, but is only rarely used.

This means that if `EXIT_SUCCESS_OFFSET + num_edits % 256 == 0`, most shells/launchers may interpret the result as success. :-)

mortie: Yeah, I pointed that out in my PR description. It's an existing bug, I just moved the existing `exit(EXIT_SUCCESS_OFFSET + num_edits)` into an `else`. I decided to not touch it in order to avoid changing the behavior of the code when running without `--exit_code`, but that might have been the wrong choice.

EdSchouten: Whoops! Sorry for not reading the description properly. Great work!

mortie:
I made https://github.com/include-what-you-use/include-what-you-use/pull/904/commits/ef76947aa66be44473037cf45e5f34d01019a412 to cap the exit code at 100. The reason I didn't cap it at 255 is that high exit codes are used by shells to signal various errors; 126 means the binary couldn't be found, 127 means the binary wasn't marked executable, 128 + N indicates that the process exited due to signal N. I chose 100 as the cap rather than 125 in order to be consistent with fix_includes.py: https://github.com/include-what-you-use/include-what-you-use/blob/master/fix_includes.py#L2453

This is arguably slightly out of the scope of this PR, but I think it fits. `--exit_code` is significantly less useful if there's a 1/256 chance that the exit code will just be incorrect.

iwyu.cc
mortie: This should be resolved by https://github.com/include-what-you-use/include-what-you-use/commit/ef76947aa66be44473037cf45e5f34d01019a412

kimgr:
Thank you! I have some review feedback that I'm hoping to get in around this weekend.

I feel terrible having let the % 256 and >100 bugs slip through, that's just not very nice.

I think the time has come to break default behavior of IWYU to succeed by default and only fail if a special flag is passed. Everything will make more sense that way. I'll think of some concrete changes when I look at your code more closely.

kimgr:
I started reviewing, but I kept coming back to the idea that we should probably just stop messing around with all this. Rather than adding lots of review feedback, I hacked up a branch to demonstrate what I think we should do: https://github.com/kimgr/include-what-you-use/commits/exit-code.

There's probably some work to do on commit messages/switch names/help text -- oh! man page! -- etc, but I think the general direction is good. Let me know what you think.

kimgr:
@mortie @EdSchouten What do you think of the more drastic approach of moving to a normal exit-code strategy for IWYU? ^

alokpr:
@kimgr I like your drastic approach. Are you still interested in pushing it?

kimgr:
@alokpr @Warchant Thanks. I'll polish that up after the next release to give it some time to bake on mainline before it hits a public release.

firewave:
Being able to return a custom exitcode on error does make sense so you are able to differentiate it from other failures. Especially if this tool is used in a chain which passes on the exit code of underlying processes. Cppcheck and the GCC/Clang sanitizers have such options.

And the wording should probably be changed to indicate that `1` also indicates a finding (i.e. something to fix) and not just an "error" as in something went wrong with the analysis.

The custom exitcode could be used for just the findings case though. That's what Cppcheck does - so by default on a finding or anything goes wrong you get `1`. If nothing is found you get `0`. If you set the custom one to `0` you will always get that unless there's an actual error like a wrong option. And if you change it to something like `33` that would indicate a finding for you to fix and `1` indicates the analysis went wrong.

PR #906, opened 2021-04-23
Minor updates

storoj:
CK_MatrixCast added in 71ab6c98a

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1998,6 +1998,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       case clang::CK_IntegralToFixedPoint:
       case clang::CK_IntegralToFloating:
       case clang::CK_IntegralToPointer:
+      case clang::CK_MatrixCast:
       case clang::CK_MemberPointerToBoolean:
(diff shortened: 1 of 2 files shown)


kimgr:
Nice, thanks!

Can you link to the Clang commit using a full GitHub URL, now that everything is available on the web? https://github.com/llvm/llvm-project/commit/71ab6c98a0d1b5590ed12f955dea6dfc714e9490

Otherwise LGTM!

storoj:
@kimgr I added the full github url to the commit.

I am wondering if linking the diff https://reviews.llvm.org/D99037 could be even better idea? Phabricator is hosted at llvm.org which is probably more "reliable" link because the repo could potentially migrate from github to anywhere else. The diff page also contains all the discussion, including the actual patches.

kimgr:
The phabricator revision is linked from their commit message, so I think the commit sha is the most important link. Even if github disappears, the sha is most likely going to be around one way or another. The URL is good for two reasons:

* Convenience, easy to follow and see what the corresponding change was
* Makes it clear(er) which repo the commit belongs to - there's nothing to say llvm doesn't go from monorepo to multiple again

Thanks, this looks good, I'll merge it as-is.

PR #907, opened 2021-04-24
Run tests in parallel on CI

storoj:
It currently takes ~18s to run tests on CI. Running tests in parallel can be much faster.


storoj:
Will give effect in combination with #859.

kimgr:
Thanks!

PR #911, opened 2021-05-08
Get a FileEntry for <new> imports

storoj:
Use preprocessor to lookup the `FileEntry` for `<new>` in the corner case that reports the "placement new" symbol usage.

This is an experimental change and I am looking for your feedback @kimgr.

I find the current implementation of the final steps of IWYU a bit overcomplicated. I mean the steps that actually find the intersection and the difference of two sets: existing imports and "desired" imports.

As far as I can see now IWYU works with "lines" (`OneIncludeOrForwardDeclareLine`), and IMO we can treat these as "strings".

I find it a bit inefficient and inaccurate, and my proposal is to use `FileEntries` for that. For every `IwyuFileInfo` we could collect the set of `FileEntries` being `#included`, as well as all the `SourceLocations` of these includes.

For example `IwyuFileInfo` could contain a map like:
```
std::map<const FileEntry*, std::set<SourceLocation>>
```
where the key is an import, and the value is all the locations of that import in the file relevant to that `IwuiFileInfo`. Then when we report a Full Use or a Forward Declaration we could also get the `FileEntry` for the corresponding symbol. In the end of AST traversal there will be two sets of `FileEntry` values and it should be much easier to calculate the deletions and the insertions.

Header mapping might be a good question, but as far as I can understand, mappings should only be applied to insertions (because IMO the existing imports that were used should not be modified - ?), and that operation could probably be made in the very end of the imports diffing algorithm.

I find the idea of migrating the logic to FileEntries very promising, and that pull request makes a first small step towards it. I am looking forward to get rid of as many of string-based operations as possible. One of the blockers was the `<new>` special case, and that change finds the FileEntry for `<new>` header.

What do you think about moving to FileEntry-based solution?

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1155,6 +1155,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   using Base::CurrentFileEntry;
   using Base::PrintableCurrentLoc;
   using Base::current_ast_node;
+  using Base::compiler;
 
   enum class IgnoreKind {
     ForUse,
@@ -2426,9 +2427,18 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         // parse it to '<new>' before using, so any path that does
         // that, and is clearly a c++ path, is fine; its exact
         // contents don't matter that much.
+        using clang::Optional;
+        using clang::DirectoryLookup;
+        using clang::FileEntryRef;
         const FileEntry* use_file = CurrentFileEntry();
-        preprocessor_info().FileInfoFor(use_file)->ReportFullSymbolUse(
-            CurrentLoc(), "<new>", "operator new");
+        const DirectoryLookup* curdir = nullptr;
+        Optional<FileEntryRef> file = compiler()->getPreprocessor().LookupFile(
+            CurrentLoc(), "new", true, nullptr, use_file, curdir, nullptr,
+            nullptr, nullptr, nullptr, nullptr, false);
+        if (file) {
+          preprocessor_info().FileInfoFor(use_file)->ReportFullSymbolUse(
+              CurrentLoc(), *file, "operator new");
+        }
       }
     }
 
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -631,9 +631,10 @@ void IwyuFileInfo::ReportFullSymbolUse(SourceLocation use_loc,
 }
 
 void IwyuFileInfo::ReportFullSymbolUse(SourceLocation use_loc,
-                                       const string& dfn_filepath,
+                                       const FileEntry* dfn_file,
                                        const string& symbol) {
-  symbol_uses_.push_back(OneUse(symbol, nullptr, dfn_filepath, use_loc));
+  symbol_uses_.push_back(OneUse(symbol, dfn_file, 
+                                GetFilePath(dfn_file), use_loc));
   LogSymbolUse("Marked full-info use of symbol", symbol_uses_.back());
 }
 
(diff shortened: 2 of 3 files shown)


kimgr:
Very cool! I'll try to look into this in more detail shortly. It would be nice if the commit message had more rationale as to why this is a valuable change. I no longer remember why I wanted to do this, but it's been mulling in the back of my head for ages without coming to anything, so if something was driving you to do this, the commit message is the right place to expand on it.

iwyu.cc
```diff
@@ -2426,9 +2427,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         // parse it to '<new>' before using, so any path that does
         // that, and is clearly a c++ path, is fine; its exact
         // contents don't matter that much.
+
```
kimgr, line 2430: Unrelated, unnecessary whitespace change

iwyu.cc
```diff
@@ -2426,9 +2427,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
         const FileEntry* use_file = CurrentFileEntry();
-        preprocessor_info().FileInfoFor(use_file)->ReportFullSymbolUse(
-            CurrentLoc(), "<new>", "operator new");
+        const clang::DirectoryLookup* TmpDir = nullptr;
```
kimgr, line 2432: We use `lowercase_with_underscores` naming for variables in IWYU, not LLVM's `PascalCase`.

This can probably be called `curdir` to signify what role it plays in the argument list.

iwyu.cc
```diff
@@ -2426,9 +2427,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        auto file = compiler()->getPreprocessor().LookupFile(
```
kimgr, line 2433: Don't use `auto` unless the type is clearly visible on the line already. `Optional<FileEntryRef>` provides relevant information here.

iwyu_output.cc
```diff
@@ -275,6 +275,22 @@ OneUse::OneUse(const string& symbol_name, const FileEntry* dfn_file,
     suggested_header_ = decl_filepath_;
 }
 
+OneUse::OneUse(const FileEntry* dfn_file, const string& dfn_filepath)
```
kimgr, line 278: I'm a little confused why this new overload is necessary. I played around with it on my local machine, but I didn't get my head around it.

Are there two changes in this commit -- one to use `FileEntry` and one to simplify construction at https://github.com/include-what-you-use/include-what-you-use/pull/911/files#diff-96c9906a158b909fda0f001cb5bd04615d46704df24173306894f3d228abbed1R669?

This one looks kind-of redundant, but I'm not entirely sure what's going on. One thing at a time is my motto :-)

iwyu_output.cc
```diff
@@ -298,8 +314,8 @@ void OneUse::SetPublicHeaders() {
   const IncludePicker& picker = GlobalIncludePicker();   // short alias
   const string use_path = GetFilePath(use_loc_);
   // If the symbol has a special mapping, use it, otherwise map its file.
-  public_headers_ = picker.GetCandidateHeadersForSymbolUsedFrom(
-      symbol_name_, use_path);
+  public_headers_ =
```
kimgr, line 317: Unrelated reformatting, please revert.

kimgr (changes requested):
Very nice. Hopefully we can use this as a first step to transition wholesale to `FileEntry*`. Comments inline.

iwyu.cc
storoj: reverted

iwyu.cc
storoj: used `curdir` as a variable name

iwyu.cc
storoj: replaced `auto` with the explicit type

storoj: replaced `auto` with the explicit type

iwyu_output.cc
storoj: Thanks for pointing that out, I actually had mixed feelings about adding another overload with less arguments. Reverted the change and reused the existing ctor.

kimgr (approved):

Issue #918, opened 2021-05-17
False positive: IWYU wants <cmath> removed when using std::pow


carlosgalvezp:
Hi,

The following code:

```
#include <cmath>

float f(float x)
{
    return std::pow(x, 2.0F);
}
```

leads to IWYU wrongly requesting us to remove `cmath`:

```
Ignoring use of std::pow (c.cpp:5:12): built-in function
c.cpp should remove these lines:
- #include <cmath>
```

Do note that the code will not compile if we follow IWYU's suggestion.

kimgr:
I don't fully understand all the nuances, but there appear to be a multitude of different kinds of built-ins, and some of them actually do require a header, such as `pow`. I think this is bug in how we try to determine what's a built-in function in IWYU. I've found some interesting classification methods in https://github.com/llvm/llvm-project/blob/main/clang/include/clang/Basic/Builtins.h#L66, but it looks like we have to thread a `Builtin::Context` object all the way into `iwyu_output.h`. Hopefully that does not turn out to be difficult :).

kimgr:
@carlosgalvezp Please give the change in #920 a shot, it seems to work well for your reduced testcase here.

kimgr:
Duplicate of #776, closing.

PR #920, opened 2021-05-22
Improve detection of builtin functions

kimgr:
The technique used before relied on linear search of all builtins by
name. This did not work with qualified C++ names of C library names such
as std::pow and std::round.

Use ASTContext to get to BuiltinInfo, and in turn use that to check more
specific traits of the builtin by ID instead of name.

Fixes issue #918.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1095,11 +1095,14 @@ bool DeclsAreInSameClass(const Decl* decl1, const Decl* decl2) {
   return decl1->getDeclContext()->isRecord();
 }
 
-bool IsBuiltinFunction(const clang::NamedDecl* decl,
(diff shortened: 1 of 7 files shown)


kimgr:
@carlosgalvezp @avostrik Could you take this for a spin and let me know if it solves your respective issues in context? Thanks!

carlosgalvezp:
@kimgr Just tested it and it works perfectly on my issue, thanks a lot! I wonder if it makes sense to add a unit test for this to make sure it doesn't pop up again?

kimgr:
@carlosgalvezp Thanks for testing. Oh, and thanks for keeping me honest -- I was so excited I got it to work that I forgot about regression testing  :clown_face:. Will update shortly.

PR #922, opened 2021-05-24
Update clang_12 branch to use LLVM 12 packages

kimgr:
Using an MR to see if Github CI seems to do the right thing.


kimgr:
Abandon, I'm working this out directly on the branch instead.

PR #934, opened 2021-07-14
Initialize all LLVM targets on startup

kimgr:
It used to be the case that the MS inline assembly parser in Clang crashed if an
X86 target was not registered and initialized.

The error handling there has been improved, so now Clang complains and says it
needs X86 target support to continue, and raises an error.

That's good news for IWYU, as the majority of code we analyze has no MS inline
assembly (fingers crossed!). So instead of requiring an X86 target to be
included, initialize _all_ registered LLVM targets and assume that X86 is
available in any configuration intended for use with MS inline assembly.

This makes it possible to build a fully non-X86 toolchain including IWYU.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -4150,7 +4150,6 @@ class IwyuAction : public ASTFrontendAction {
 
 #include "iwyu_driver.h"
 #include "clang/Frontend/FrontendAction.h"
-#include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/TargetSelect.h"
 
 using include_what_you_use::OptionsParser;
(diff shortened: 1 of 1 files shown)


kimgr:
@Romain-Geissler-1A I finally got around to thinking about this issue you reported in https://github.com/include-what-you-use/include-what-you-use/pull/854.

I resorted to just initializing all targets, hoping that X86 is available in environments where MS inline assembly is a thing. Error handling in LLVM/Clang has also been improved to no longer crash, but report an actionable error message (that no target information is available).

kimgr:
Replaces #854

Romain-Geissler-1A:
Cool, thank you !

vmiklos:
@kimgr Hmm, this breaks my build, I wonder why. Do you use -DBUILD_SHARED_LIBS=ON when building llvm? Or do you restrict the list of llvm targets in your config? In both cases, I don't specify anything explicitly, i.e. use the defaults.

The output I get is:

```
[ 92%] Built target clang-resource-headers
[ 93%] Linking CXX executable bin/include-what-you-use
ld: CMakeFiles/include-what-you-use.dir/iwyu.cc.o: in function `main':
iwyu.cc:(.text.main+0xc): undefined reference to `LLVMInitializeAArch64TargetInfo'
ld: iwyu.cc:(.text.main+0x11): undefined reference to `LLVMInitializeAMDGPUTargetInfo'
ld: iwyu.cc:(.text.main+0x16): undefined reference to `LLVMInitializeARMTargetInfo'
ld: iwyu.cc:(.text.main+0x1b): undefined reference to `LLVMInitializeAVRTargetInfo'
...
```

Everything is OK again if I revert this single commit locally. Thanks.

kimgr:
@vmiklos Oh no, thanks for the report!

I never use `BUILD_SHARED_LIBS`, as it's marketed as an unsupported llvm dev convenience. But I use the LLVM Debian packages, and I think they ship LLVM and Clang dylibs by default (based off a more supported configuration, not sure what).

Hmm. I wonder if there's a nice way to query what targets are available on startup -- i'd like to do something like:
```
if 'X86' in supported_targets:
    LLVMInitializeX86TargetInfo()
    ...
```

Issue #847, opened 2020-11-10
Wrongly suggested #include removal involving explicit template specialization and macro


stbergmann:
```
$ cat include1.h
#ifndef INCLUDE1
#define INCLUDE1
template<typename> void f() = delete;
#define M f<int>()
#endif

$ cat include2.h
#include "include1.h"
template<> void f<int>() {}

$ cat test.cc
#include "include1.h"
#include "include2.h"
void g() { M; }

$ include-what-you-use test.cc

test.cc should add these lines:

test.cc should remove these lines:
- #include "include2.h"  // lines 2-2

The full include-list for test.cc:
#include "include1.h"  // for M
---
```
but actually removing line 2 from `test.cc` breaks:
```
$ clang++ test.cc
test.cc:2:12: error: call to deleted function 'f'
void g() { M; }
           ^
./include1.h:4:11: note: expanded from macro 'M'
#define M f<int>()
          ^~~~~~
./include1.h:3:25: note: candidate function [with $0 = int] has been explicitly deleted
template<typename> void f() = delete;
                        ^
1 error generated.
```

kimgr:
Nice repro, thanks!

This actually works as designed, it's just that the design doesn't do what's intuitive 🙂.

The rules for macros are basically:

Uses inside a macro are attributed to the macro definition unless:

* the symbol used comes from a macro argument
* there is a forward-declare in the macro dfn file of the used symbol (indicating author intent that macro users are responsible) 

I wonder if a base template should count as a forward declaration for a specialization.

kimgr:
Fix merged.

PR #936, opened 2021-07-20
Use primary function template declaration for macro author intent

kimgr:
We already have a rule that if a decl is used entirely inside a macro, and it's
forward-declared in the macro file, the use is attributed to the expansion
location. This, however, did not cover template specializations, as described in
issue 847.

So now we treat a primary function template declaration in the macro file as a
regular forward-declaration, which means that function template specializations
used in macros will be attributed to the expansion location iff the primary
template has a declaration in the macro file.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1259,7 +1259,26 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     for (const NamedDecl* redecl : GetClassRedecls(decl)) {
       if (GetFileEntry(redecl) == macro_def_file && IsForwardDecl(redecl)) {
         fwd_decl = redecl;
+        break;
+      }
+    }
 
+    if (!fwd_decl) {
+      if (const auto* func_decl = dyn_cast<FunctionDecl>(decl)) {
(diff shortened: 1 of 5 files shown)


kimgr:
@stbergmann I think this fixes #847 -- could you try out the patch in the real world? Thanks.

stbergmann:
> @stbergmann I think this fixes #847 -- could you try out the patch in the real world? Thanks.

I don't use iwyu myself, and only filed #847 following the discussion in the comments at LibreOffice commit <https://gerrit.libreoffice.org/c/core/+/105315/2#message-f5292b496cf1f5c78d6381bc3e2a59320e8ffbb9> "tdf#42949 Fix new IWYU warnings in directories c*", but @vmiklos might be interested.

vmiklos:
Yes, I just tried this PR on libreoffice/core.git, the original cppu/qa/test_any.cxx had this problem that o3tl/cppunittraitshelper.hxx was suggested for removal and this is now gone. As I understand it, that was the source of the stripped down sample in the issue. Thanks!

Issue #826, opened 2020-09-11
Template class taking an std::function, usage not recognized


vmiklos:
Hi,

Steps to reproduce the problem:

1. Create a `header.h`:

````
#include <functional>

template <class Func> struct Defer {
    Defer(Func&& func)
        : m_Func(std::move(func)) {
    }

    ~Defer() { m_Func(); }

    Func m_Func;
};
````

2. Create a `test.cc`:

````
#include "header.h"

void f() {}

int g() {
    Defer defer([] { f(); });
    return 0;
}
````

3. Run IWYU on this:

````
include-what-you-use -std=c++17 -c test.cc
````

Expected result: IWYU says there is nothing to do.

Actual result:

````

test.cc should add these lines:

test.cc should remove these lines:
- #include "header.h"  // lines 1-1

The full include-list for test.cc:
---
````

But if I do what it says, I get:

````
test.cc:4:5: error: unknown type name 'Defer'
    Defer defer([] { f(); });
    ^
````

Note that in case `defer` is created using the older pre-C++17 syntax:

````
    Defer<std::function<void()>> defer([] { f(); });
````

Then the problem goes away.

Version:

````
include-what-you-use 0.15 (git:bff9327) based on clang version 12.0.0 (git://github.com/llvm/llvm-project d2c69c2f4947b38832a34cab14fe32c6b94dd4d2)
````

i.e. this is meant to be ~latest iwyu master vs llvm-project.git master (sorry, first posted wrong version, which was outdated, due to incremental build).

(The above is a simplified version of LibreOffice's `comphelper::ScopeGuard` class, that's how I ran into this.)

As usual, if there are code pointers on what should be fixed here, I'm not against attempting to fix this myself, but I'm unsure where to start. :-)

Also, feel free to rename the issue, I'm just guessing what may be the root cause here.

Thanks,

Miklos

kimgr:
I wonder if there's an even simpler repro without `std::function`. It sounds like we're missing handling of https://en.cppreference.com/w/cpp/language/class_template_argument_deduction.

Looking at your example, it's not clear to me why `<functional>` is required in `header.h`. It feels like `Defer` could do just as well with a function pointer.

vmiklos:
Sorry, my `<functional>` include in the header is misleading. Actually just `<utility>` is enough, needed for `std::move()`. And the problem happens without `std::move()` as well, so here is a reproducer that doesn't include any STL header. header.h:

````
[...]

PR #937, opened 2021-07-20
Support class template argument deduction

kimgr:
CTAD is a C++17 feature that allows template arguments to be deduced from
constructor expressions. For example:

    A a(10, 20);

is shorthand for:

    A<int, int> a(10, 20);

This should fix issue #826.


kimgr:
@vmiklos Could you take a look at this attempt to fix #826?

vmiklos:
Yes, it fixes the problem I reported. Thanks! :-)

PR #941, opened 2021-08-07
C++11 type aliases should be treated similar to typedefs

bolshakov-a:
Seemingly useless line of code removed by the way.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1415,12 +1415,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   }
 
   set<const Type*> GetCallerResponsibleTypesForTypedef(
-      const TypedefDecl* decl) {
+      const TypedefNameDecl* decl) {
     set<const Type*> retval;
     const Type* underlying_type = decl->getUnderlyingType().getTypePtr();
     // If the underlying type is itself a typedef, we recurse.
     if (const TypedefType* underlying_typedef = DynCastFrom(underlying_type)) {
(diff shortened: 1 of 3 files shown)


iwyu.cc
```diff
@@ -3941,10 +3941,9 @@ class IwyuAstConsumer
   // TODO(csilvers): we can probably relax this rule in .cc files.
   // TODO(csilvers): this should really move into IwyuBaseASTVisitor
   // (that way we'll correctly identify need for hash<> in hash_set).
-  // This is a Traverse*() because Visit*() can't call HandleFunctionCall().
-  bool TraverseTypedefDecl(clang::TypedefDecl* decl) {
-    if (!Base::TraverseTypedefDecl(decl))
-      return false;
+  // This is called from Traverse*() because Visit*()
+  // can't call HandleFunctionCall().
+  bool CheckAliasedClassMethods(TypedefNameDecl* decl) {
```
kimgr, line 3946: I'd prefer `Handle...` to `Check...` here. "Check" implies "try-and-fail-hard" in IWYU, given the `CHECK_` macros.

kimgr (commented):
Looks great! One naming nit, otherwise lgtm.

iwyu.cc
bolshakov-a: Thanks, fixed.

bolshakov-a: I`m sorry, actually, isn't.

bolshakov-a: Fixed at now.

kimgr:
Wonderful, thank you!

PR #952, opened 2021-09-20
Don't force aliased type method callability

bolshakov-a:
Don't require aliased type method callability if explicitly forward declared
according to the IWYU forward declaration policy.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3944,7 +3944,10 @@ class IwyuAstConsumer
   // This is called from Traverse*() because Visit*()
   // can't call HandleFunctionCall().
   bool HandleAliasedClassMethods(TypedefNameDecl* decl) {
-    if (CanIgnoreCurrentASTNode())  return true;
+    if (CanIgnoreCurrentASTNode())
+      return true;
+    if (current_ast_node()->in_forward_declare_context())
+      return true;
 
     const Type* underlying_type = decl->getUnderlyingType().getTypePtr();
     const Decl* underlying_decl = TypeToDeclAsWritten(underlying_type);
--- a/tests/cxx/no_forced_alias_callability.cc
+++ b/tests/cxx/no_forced_alias_callability.cc
@@ -1,0 +1,29 @@
+//===--- no_forced_alias_callability.cc - test input file for iwyu --------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --check_also="tests/cxx/no_forced_alias_callability-d2.h" \
+//            -I .
+
+// Tests that IWYU doesn't require inclusion of an aliased class header
+// (...-d1.h) into a header with the alias to provide callability of methods
+// of the aliased class if the aliased class is explicitly made forward declared
+// in accordance with the IWYU policy
+
+#include "tests/cxx/no_forced_alias_callability-d1.h"
+#include "tests/cxx/no_forced_alias_callability-d2.h"
+
+int main() {
+  Alias a;
+}
+
+/**** IWYU_SUMMARY
+
+(tests/cxx/no_forced_alias_callability.cc has correct #includes/fwd-decls)
+
+***** IWYU_SUMMARY */
--- a/tests/cxx/no_forced_alias_callability-d1.h
+++ b/tests/cxx/no_forced_alias_callability-d1.h
@@ -1,0 +1,14 @@
+//===--- no_forced_alias_callability-d1.h - test input file for iwyu ------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+struct ReturnType;
+
+struct Aliased {
+  ReturnType doSomething();
+};
--- a/tests/cxx/no_forced_alias_callability-d2.h
+++ b/tests/cxx/no_forced_alias_callability-d2.h
@@ -1,0 +1,18 @@
+//===--- no_forced_alias_callability-d2.h - test input file for iwyu ------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+struct Aliased;
+
+typedef Aliased Alias;
+
+/**** IWYU_SUMMARY
+
+(tests/cxx/no_forced_alias_callability-d2.h has correct #includes/fwd-decls)
+
+***** IWYU_SUMMARY */


kimgr:
Thanks! I'll look into this.

iwyu.cc
```diff
@@ -3975,14 +3975,18 @@ class IwyuAstConsumer
     if (!Base::TraverseTypedefDecl(decl))
       return false;
 
-    return HandleAliasedClassMethods(decl);
+    return current_ast_node()->in_forward_declare_context()
```
kimgr, line 3978: I think it would be nicer to move this check into `HandleAliasedClassMethods`, e.g.
```
  bool HandleAliasedClassMethods(TypedefNameDecl* decl) {
    if (current_ast_node()->in_forward_declare_context())
      return true;
    ...
```

tests/cxx/no_forced_alias_callability.cc
```diff
@@ -0,0 +1,29 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --check_also="tests/cxx/no_forced_alias_callability-d2.h" \
+//            -I .
+
+// Tests that IWYU don't require inclusion of an aliased class header (...-d2.h)
```
kimgr, line 13: typo: don't -> doesn't

tests/cxx/no_forced_alias_callability.cc
```diff
@@ -0,0 +1,29 @@
...
+#include "no_forced_alias_callability-d1.h"
+#include "no_forced_alias_callability-d2.h"
+
+int main() {
+  Alias a;
+}
+
+/**** IWYU_SUMMARY
+
+(tests/cxx/no_forced_alias_callability.cc has correct #includes/fwd-decls)
```
kimgr, line 27: I usually discourage success-only testcases like this, because it's hard to see _why_ they're successful: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#force-diagnostics

I'm not 100% sure how to achieve that in this case, but if you could give it a shot, I'd be grateful. I think what I want to achieve are inline IWYU diagnostics around the individual uses of `Aliased`, to verify that they are attributed to the right header.

kimgr (commented):
Thanks, some small feedback inline.

iwyu.cc
bolshakov-a: I surely saw the duplication, but I thought that such bypasses should not be in `HandleAliasedClassMethods` function. Maybe it should be renamed into `HandleAliasedClassMethodsIfNeeded` or something else?

tests/cxx/no_forced_alias_callability.cc
bolshakov-a: I thought about it, but I have no idea how to achieve negative scenario. Actually, we need to check that IWYU _doesn't complain_ that `-d1.h` isn't included into `-d2.h` when it isn't.
As far as I understand, IWYU diagnostics are only written when something goes wrong. But other combinations of header inclusions in `.cc` or `-d2.h` files either produce clang errors or break the test case.

iwyu.cc
kimgr: `CanIgnoreCurrentASTNode` also potentially short-circuits `HandleAliasedClassMethods`, so I wouldn't worry about it too much. It makes sense to me for it to silently pass if it doesn't need to do anything.

kimgr:
Please shorten commit subject to <=50 chars and commit body to <=72 chars according to https://chris.beams.io/posts/git-commit/. Short lines are easier to read, and helps side-by-side reading.

tests/cxx/no_forced_alias_callability.cc
kimgr: Something like this appears to work:
```diff
164e8b2e7112cfcb1681101fa99aaae816a7d72c
Author:     Kim Gräsman <kim.grasman@gmail.com>
AuthorDate: Mon Sep 27 20:36:09 2021 +0200
Commit:     Kim Gräsman <kim.grasman@gmail.com>
CommitDate: Mon Sep 27 20:36:09 2021 +0200

Experimental failing testcase

5 files changed, 50 insertions(+), 18 deletions(-)
tests/cxx/no_forced_alias_callability-d1.h |  6 +-----
tests/cxx/no_forced_alias_callability-d2.h | 10 +---------
tests/cxx/no_forced_alias_callability-i1.h | 14 ++++++++++++++
tests/cxx/no_forced_alias_callability-i2.h | 18 ++++++++++++++++++
tests/cxx/no_forced_alias_callability.cc   | 20 ++++++++++++++++----

modified   tests/cxx/no_forced_alias_callability-d1.h
@@ -7,8 +7,4 @@
 //
 //===----------------------------------------------------------------------===//
 
-struct ReturnType;
-
-struct Aliased {
-  ReturnType doSomething();
-};
+#include "tests/cxx/no_forced_alias_callability-i1.h"
modified   tests/cxx/no_forced_alias_callability-d2.h
@@ -7,12 +7,4 @@
 //
 //===----------------------------------------------------------------------===//
 
-struct Aliased;
-
-typedef Aliased Alias;
-
-/**** IWYU_SUMMARY
-
-(tests/cxx/no_forced_alias_callability-d2.h has correct #includes/fwd-decls)
-
-***** IWYU_SUMMARY */
+#include "tests/cxx/no_forced_alias_callability-i2.h"
new file   tests/cxx/no_forced_alias_callability-i1.h
@@ -0,0 +1,14 @@
+//===--- no_forced_alias_callability-i1.h - test input file for iwyu ------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+struct ReturnType;
+
+struct Aliased {
+  ReturnType doSomething();
+};
new file   tests/cxx/no_forced_alias_callability-i2.h
@@ -0,0 +1,18 @@
+//===--- no_forced_alias_callability-i2.h - test input file for iwyu ------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+struct Aliased;
+
+typedef Aliased Alias;
+
+/**** IWYU_SUMMARY
+
+(tests/cxx/no_forced_alias_callability-i2.h has correct #includes/fwd-decls)
+
+***** IWYU_SUMMARY */
modified   tests/cxx/no_forced_alias_callability.cc
@@ -7,7 +7,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-// IWYU_ARGS: -Xiwyu --check_also="tests/cxx/no_forced_alias_callability-d2.h" \
+// IWYU_ARGS: -Xiwyu --check_also="tests/cxx/no_forced_alias_callability-i2.h" \
 //            -I .
 
 // Tests that IWYU doesn't require inclusion of an aliased class header
@@ -15,15 +15,27 @@
 // of the aliased class if the aliased class is explicitly made forward declared
 // in accordance with the IWYU policy
 
-#include "no_forced_alias_callability-d1.h"
-#include "no_forced_alias_callability-d2.h"
+#include "tests/cxx/no_forced_alias_callability-d1.h"
+#include "tests/cxx/no_forced_alias_callability-d2.h"
 
 int main() {
+  // IWYU: Alias is...*no_forced_alias_callability-i2.h
+  // IWYU: Aliased is...*no_forced_alias_callability-i1.h
   Alias a;
 }
 
 /**** IWYU_SUMMARY
 
-(tests/cxx/no_forced_alias_callability.cc has correct #includes/fwd-decls)
+tests/cxx/no_forced_alias_callability.cc should add these lines:
+#include "tests/cxx/no_forced_alias_callability-i1.h"
+#include "tests/cxx/no_forced_alias_callability-i2.h"
+
+tests/cxx/no_forced_alias_callability.cc should remove these lines:
+- #include "tests/cxx/no_forced_alias_callability-d1.h"  // lines XX-XX
+- #include "tests/cxx/no_forced_alias_callability-d2.h"  // lines XX-XX
+
+The full include-list for tests/cxx/no_forced_alias_callability.cc:
+#include "tests/cxx/no_forced_alias_callability-i1.h"  // for Aliased
+#include "tests/cxx/no_forced_alias_callability-i2.h"  // for Alias
 
 ***** IWYU_SUMMARY */
```

But I'm not sure it's unequivocally better. It does shine the light on one thing I didn't pick up from your testcase: both `Aliased` and `Alias` need to be visible to the user of `Alias`. That kind of makes sense in retrospect, but it's always nice if the testcase can tell the story.

More concrete note: please `#include "tests/cxx/..."` to mirror most other testcases. The includes without `tests/cxx/` namespacing are exercising a special mode.

bolshakov-a: Thank you! It looks really better. On the other side, these diagnostics don't fail even without changes suggested in the PR. Consequently, it occurs that the test case doesn't test single thing...

bolshakov-a:
> Please shorten commit subject to <=50 chars and commit body to <=72 chars according to https://chris.beams.io/posts/git-commit/. Short lines are easier to read, and helps side-by-side reading.

tests/cxx/no_forced_alias_callability.cc
kimgr: Hm. You mean the testcase as a whole succeeds with or without the change? I tested that, and saw failure before and success after. But it took me a couple of attempts to successfully revert the change, so maybe I ended up testing the wrong rev. I'll try again.

bolshakov-a: No, not as a whole, but checks in `.cc`-file only. So I suggest that they test already working functionality, and it may be misleading.

kimgr: Ah, I see more clearly now that the interesting part is in `tests/cxx/no_forced_alias_callability-[di]2.h`. I think you're right that the success-only test is better -- my suggestion makes a lot of noise about non-essential uses.

```
include-what-you-use$ git grep "has correct #includes" tests/cxx/ | wc -l
72
```

so it's not exactly unusal, and I think this is a situation where it makes sense. Could you change it back, but keep the `tests/cxx` qualification on `#include`s? Thanks!

bolshakov-a: Of course. Done.

kimgr: :+1:

PR #953, opened 2021-09-25
Add --update_comments option

### LINKED ISSUE #75: fix_includes.py --comments should add comments even if the actual includes didn't change

### LINKED ISSUE #494: option to generate full include-list
It would be nice if there was an option to generate the full include-list in the output even if the file is considered to be correct.

Rationale is that the comments may be wrong and I want to obtain the updated list.

salman-javed-nz:
…clude comments, regardless of whether any include lines need to be added or removed.

Fixes #75 and fixes #494.

With this change, IWYU will update the "// for xyz" comments as the code changes, preventing the comments from becoming stale.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -2309,11 +2309,12 @@ def ProcessIWYUOutput(f, files_to_process, flags, cwd):
   # seen for them.  (We have to wait until we're all done, since a .h
   # file may have a contentful change when #included from one .cc
   # file, but not another, and we need to have merged them above.)
-  for filename in iwyu_output_records:
-    if not iwyu_output_records[filename].HasContentfulChanges():
-      print('(skipping %s: iwyu reports no contentful changes)' % filename)
-      # Mark that we're skipping this file by setting the record to None
-      iwyu_output_records[filename] = None
+  if not flags.update_comments:
+    for filename in iwyu_output_records:
+      if not iwyu_output_records[filename].HasContentfulChanges():
+        print('(skipping %s: iwyu reports no contentful changes)' % filename)
+        # Mark that we're skipping this file by setting the record to None
+        iwyu_output_records[filename] = None
 
   # Now do all the fixing, and return the number of files modified
   contentful_records = [ior for ior in iwyu_output_records.values() if ior]
@@ -2372,6 +2373,12 @@ def main(argv):
   parser.add_option('--comments', action='store_true', default=False,
                     help='Put comments after the #include lines')
   parser.add_option('--nocomments', action='store_false', dest='comments')
+
+  parser.add_option('--update_comments', action='store_true', default=False,
+                    help=('Update #include comments, even if no #include lines'
+                          ' are added or removed'))
+  parser.add_option('--noupdate_comments', action='store_false',
+                    dest='update_comments')
 
   parser.add_option('--safe_headers', action='store_true', default=True,
                     help=('Do not remove unused #includes/fwd-declares from'
@@ -2439,6 +2446,9 @@ def main(argv):
       not flags.separate_project_includes.endswith(os.path.sep) and
       not flags.separate_project_includes.endswith('/')):
     flags.separate_project_includes += os.path.sep
+
+  if flags.update_comments:
+    flags.comments = True
 
   if flags.sort_only:
     if not files_to_modify:
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -91,6 +91,8 @@ static void PrintHelp(const char* extra_msg) {
          "        the maximum line length can still be exceeded with long\n"
          "        file names (default: 80).\n"
          "   --no_comments: do not add 'why' comments.\n"
+         "   --update_comments: always add 'why' comments, even if no\n"
+         "        #include lines need to be added or removed.\n"
          "   --no_fwd_decls: do not use forward declarations.\n"
          "   --verbose=<level>: the higher the level, the more output.\n"
          "   --quoted_includes_first: when sorting includes, place quoted\n"
@@ -163,6 +165,7 @@ CommandlineFlags::CommandlineFlags()
       prefix_header_include_policy(CommandlineFlags::kAdd),
       pch_in_code(false),
       no_comments(false),
+      update_comments(false),
       no_fwd_decls(false),
       quoted_includes_first(false),
       cxx17ns(false) {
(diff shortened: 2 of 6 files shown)


salman-javed-nz:
Here is my first pass at implementing a `--force-comments` feature.
The reason for implementing this is because IWYU does not update the `// for xyz` include comments if there are no contentful changes (e.g. a #include being added or removed).
As the code changes, the reason stated in the comment for including a header may no longer be true.

For a while I have been deliberately adding "dummy" #includes to coax IWYU into removing them and printing an up-to-date list of include comments. While this workaround has worked fine for me, it's probably better if IWYU just had the proper support for it. So this pull request is how I see what that proper support would look like.

I would love to get feedback from others on what they would like from such a feature, and if there is another way to implement this that fits better with IWYU's established ways of doing things.

kimgr:
Very nice, thanks! I'll try to take some time tomorrow to work through the diff in detail, but on first pass it looks great.

fix_includes.py
```diff
@@ -2373,6 +2374,12 @@ def main(argv):
                     help='Put comments after the #include lines')
   parser.add_option('--nocomments', action='store_false', dest='comments')
 
+  parser.add_option('--update_comments', action='store_true', default=False,
```
salman-javed-nz, line 2377: Just realised that this arg doesn't do anything unless `--comments` is true as well. 
Should I change this code so that `--update_comments` implicitly sets `flags.comments`?

iwyu_globals.h
```diff
@@ -96,9 +96,10 @@ struct CommandlineFlags {
   int max_line_length;
   // Policy regarding files included via -include option.  No short option.
   PrefixHeaderIncludePolicy prefix_header_include_policy;
-  bool pch_in_code;   // Treat the first seen include as a PCH. No short option.
-  bool no_comments;   // Disable 'why' comments. No short option.
-  bool no_fwd_decls;  // Disable forward declarations.
+  bool pch_in_code;    // Treat the first seen include as a PCH. No short option.
```
kimgr, line 99: Please don't realign all comments here, just leave the existing lines alone.

iwyu_output.cc
```diff
@@ -1978,18 +1978,20 @@ size_t PrintableDiffs(const string& filename,
...
   // First, check if there are no adds or deletes.  If so, we print a
   // shorter summary line.
-  bool no_adds_or_deletes = true;
-  for (const auto& key_line : sorted_lines) {
-    const OneIncludeOrForwardDeclareLine* line = key_line.second;
-    if ((line->is_desired() && !line->is_present()) || // add
-        (line->is_present() && !line->is_desired())) { // delete
-      no_adds_or_deletes = false;
-      break;
+  if (!GlobalFlags().force_comments) {
```
kimgr, line 1981: I like the way this avoids extra work to determine if there are no adds or deletes, but it's annoying that the diff becomes so large due to reindentation. I think we should take the small perf hit and do:
```
if (no_adds_or_deletes && !GlobalFlags().force_comments) {
      output = "\n(" + filename + " has correct #includes/fwd-decls)\n";
      return 0;
}
```

iwyu_globals.cc
```diff
@@ -91,6 +91,8 @@ static void PrintHelp(const char* extra_msg) {
          "        the maximum line length can still be exceeded with long\n"
          "        file names (default: 80).\n"
          "   --no_comments: do not add 'why' comments.\n"
+         "   --force_comments: always add 'why' comments, even if no #include\n"
```
kimgr, line 94: Would it be worse to call this `--update_comments` too? Feels nice if there's some connection between the `fix_includes.py` args and the `include-what-you-use` args.

fix_includes.py
kimgr: I think that makes sense. It gets a little weird if you combine `--nocomments` and `--update_comments`, but either way one of them has to win, and I think it makes sense for `--update_comments` to take precedence.

kimgr (commented):
Please shorten the first line of the commit message to <50 characters, e.g. just `Add --force_comments option`, and expand on the details on the commit message body.

More feedback inline, but on the whole I think this looks great. Thank you.

salman-javed-nz:
- Shortened commit message
- Don't realign comments in code not directly related to this PR
- `fix_includes.py`: `--update_comments` now implies `--comments` as well
- Check `GlobalFlags().force_comments` after the check for adds and deletes
- Use the arg name `--update_comments` instead of `--force_comments`. Arg name is now consistent between `include-what-you-use` and `fix_includes.py`.

kimgr:
Thanks! Changes look good, but the commit message needs two more changes:

- No trailing period in subject
- Body line-wrapped to <=72 chars

A question unrelated to the patch: the `fix_includes.py` testcases both have a `// IWYU pragma: keep` include line. It looks like the pragma is not preserved in the summary, which could be problematic if users copy/paste the full include-list (I've used IWYU like this in the past).

Is that just a test artifact, or does IWYU's summary lose pragmas?

salman-javed-nz:
> A question unrelated to the patch: the `fix_includes.py` testcases both have a `// IWYU pragma: keep` include line. It looks like the pragma is not preserved in the summary, which could be problematic if users copy/paste the full include-list (I've used IWYU like this in the past).
> 
> Is that just a test artifact, or does IWYU's summary lose pragmas?

Yes, as you say, the IWYU summary loses the pragmas but it was already like this before my patch.
This behavior is already captured in `tests/cxx/comment_pragmas.cc`.

Notice how the IWYU pragma comment in https://github.com/include-what-you-use/include-what-you-use/blob/0997656913fc15d52b15715e275897b4f8de6d5c/tests/cxx/comment_pragmas.cc#L83 is scrubbed in the output in https://github.com/include-what-you-use/include-what-you-use/blob/0997656913fc15d52b15715e275897b4f8de6d5c/tests/cxx/comment_pragmas.cc#L234

The good thing is that `fix_includes.py` still does the right thing and does not remove IWYU pragmas as the IWYU summary would suggest. This is exemplified in the two unit tests that I added in this patch.

kimgr:
Cool, thanks. I'll make a note of that and see if there's anything to do about it later.

salman-javed-nz:
Commit message changed as per review:
- No trailing period in subject
- Body line-wrapped to <=72 char

kimgr:
Thanks! Sorry about the delay.

salman-javed-nz:
No problem! Thanks for the review!

PR #961, opened 2021-10-22
Undefine CALLBACK after including Shlwapi.h

cdglove:
clang/Analysis/CFG.h uses CALLBACK as a template parameter to the
function VisitBlockStmts. Since this is defined as a calling convention
in minwindef.h, we end up with an error similar to the following:

CFG.h(1380): error C2988: unrecognizable template declaration/definition

The fix is a continuation of the existing workaround to just undefine
the offending macro since it's not needed in our code anyway.

### DIFF
--- a/iwyu_port.h
+++ b/iwyu_port.h
@@ -66,10 +66,12 @@ class OstreamVoidifier {
 #define NOMINMAX 1 // Prevent Windows headers from redefining min/max.
 #include "Shlwapi.h"  // for PathMatchSpecA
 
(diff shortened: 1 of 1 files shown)


cdglove:
Building against LLVM 12.0.1
And yes, you're right, looks like it was fixed up stream in 
https://github.com/llvm/llvm-project/commit/54bff1522fc863329894d875d54c2fe4cd1b4f3f

Closing this one out.

kimgr:
Phew, thanks! 🙂

PR #967, opened 2021-11-10
Add export from <stdint.h> to <inttypes.h>

osandov:
IWYU currently warns for this example:

```
  $ cat test.c
  #include <inttypes.h>
  #include <stdio.h>

  int main(void)
  {
          uint64_t x = UINT64_C(1234);
          printf("%" PRIu64 "\n", x);
          return 0;
  }
  $ include-what-you-use test.c

  #include <stdint.h>    // for UINT64_C, uint64_t

  #include <inttypes.h>  // for PRIu64
  #include <stdint.h>    // for UINT64_C, uint64_t
  #include <stdio.h>     // for printf
  ---
```

However, the C standard states that "the header <inttypes.h> includes
the header <stdint.h>". So, this program shouldn't need to include
<stdint.h>.

Some mappings were previously added for individual symbols from
<stdint.h> (e.g., commit 24f9fdf0af94 ("Add more mappings for
[u]intptr_t"), but it's more accurate to export everything. Get rid of
the symbol mappings and add an include mapping.


alejandro-colomar:
`Acked-by: Alejandro Colomar <alx.manpages@gmail.com>`

osandov:
It seems like IWYU can find that definition just fine from the header file itself, but sure, I'll add back the mapping.

iwyu_include_picker.cc
```diff
@@ -360,8 +348,8 @@ const IncludeMapEntry libstdcpp_symbol_map[] = {
   { "std::size_t", kPrivate, "<cwchar>", kPublic },
 };
 
-// Private -> public include mappings for GNU libc
```
kimgr, line 363: This move looks incidental, and I think the comment looks better above. Please revert.

kimgr (commented):
One little tweak, otherwise looks good to me!

iwyu_include_picker.cc
osandov: I moved it because `libc_include_map` is no longer strictly private -> public mappings, so now the comment refers to those mappings, not including the reexported header, which is a different kind of mapping. But I'm happy to revert that if you disagree.

kimgr: Aah, I see. With that background it makes perfect sense. Nevermind this comment.

kimgr:
Thank you!

PR #963, opened 2021-10-30
Add some mappings for glibc symbols

alejandro-colomar:

PR #968, opened 2021-11-10
Add <bits/struct_stat.h> include mapping for glibc 2.33

osandov:
glibc commit [d8927238307b](https://sourceware.org/git/?p=glibc.git;a=commit;h=d8927238307b9df32319a34755ac36f6e92a0b7d) ("linux: Move the struct stat{64} to
struct_stat.h") moved the definition of struct stat from <bits/stat.h>
to <bits/struct_stat.h>. This results in the following IWYU warning:

```
  $ cat test.c
  #include <sys/stat.h>

  struct stat x;
  $ include-what-you-use test.c

  #include <bits/struct_stat.h>  // for stat

  - #include <sys/stat.h>  // lines 1-1

  #include <bits/struct_stat.h>  // for stat
  ---
```

Fix it by adding an include mapping.

### DIFF
--- a/gcc.libc.imp
+++ b/gcc.libc.imp
@@ -75,6 +75,7 @@
   { include: [ "<bits/socket_type.h>", private, "<sys/socket.h>", public ] },
   { include: [ "<bits/stab.def>", private, "<stab.h>", public ] },
   { include: [ "<bits/stat.h>", private, "<sys/stat.h>", public ] },
+  { include: [ "<bits/struct_stat.h>", private, "<sys/stat.h>", public ] },
   { include: [ "<bits/statfs.h>", private, "<sys/statfs.h>", public ] },
   { include: [ "<bits/statvfs.h>", private, "<sys/statvfs.h>", public ] },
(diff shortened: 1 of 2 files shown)


alejandro-colomar:
Could you please also add mapping similar to these:
<https://github.com/include-what-you-use/include-what-you-use/commit/dc2acabc587ddd0d604d61a0046031b9ea4537e9>

`POSIX.1-2008` says that `struct stat` shall be provided by the following headers:

- `<sys/stat.h>`
- `<ftw.h>`
I couldn't check if `POSIX.1-2001` has any difference, since the browser is failing.

alejandro-colomar:
Ahh, now I found it.  I knew I had done this before but couldn't find it in master.  It's because the PR is still pending.

See <https://github.com/include-what-you-use/include-what-you-use/pull/963>.

kimgr:
I prefer #963 for this as well, let's reevaluate once that is merged.

kimgr:
@osandov #963 is merged, could you double-check that it works with your original problem? Thanks!

osandov:
Yup, that worked, too.

Issue #975, opened 2021-11-30
Confusion about path/to/iwyu; Clang 11 and Clang 13 failed to use


zchrissirhcz:
Hi, iwyu developers

I use CMake for my C++ project, would like to integrate iwyu it with cmake. I found the usage in this project's readme page:

```bash
  mkdir build && cd build
  CC="clang" CXX="clang++" cmake -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE="path/to/iwyu;-Xiwyu;any;-Xiwyu;iwyu;-Xiwyu;args" ...
```

I'm in confusion with those.

- I don't know what exactly `path/to/iwyu` mean, is that the absolute path of the executable `iwyu` ?
- I build include-what-you-use from source, the executable I get is `include-what-you-use` instead of `iwyu`
- I don't know what exactly does `-Xiwyu;any;-Xiwyu;iwyu;-Xiwyu;args` mean, should I change them, or keep them anyway?
- I'm using Clang 13, and the build-from-latest-source include-what-you-use just failed for very simple code
```c++
#include <stdio.h> // this line is expected to be reported
int main()
{
    return 0;
}
```
- I switch to clang11.0, with include-what-you-use clang_11 branch, but it also failed to report

kimgr:
`/path/to/iwyu` and `-Xiwyu;any;-Xiwyu;args` are both brief placeholders, I think they assume you know what the `include-what-you-use` command line looks like.

> I don't know what exactly path/to/iwyu mean, is that the absolute path of the executable iwyu ?
> I build include-what-you-use from source, the executable I get is include-what-you-use instead of iwyu

Yes, make it the absolute path to the `include-what-you-use` executable.

> I don't know what exactly does -Xiwyu;any;-Xiwyu;iwyu;-Xiwyu;args mean, should I change them, or keep them anyway?

IWYU's command-line is a little strange. It accepts all compiler arguments just like the `clang` compiler, so for example `-I/usr/include/somelib` and `-DFOO=BAR` are accepted and passed directly onto the Clang front-end for parsing.

Any IWYU arguments need to be prefixed with `-Xiwyu`, e.g. `-Xiwyu --verbose=4` or `-Xiwyu --no_comments`.

What the example is trying to say is that you need to put any additional arguments as a semicolon-separated list, and if they are IWYU-specific arguments, they need to be `-Xiwyu`-prefixed.

[...]

PR #977, opened 2021-12-05
Clarify usage instructions for CMake integration

kimgr:
Based on excellent input from @zchrissirhcz in issue #975.


kimgr:
@zchrissirhcz Your additional examples for reusing IWYU command-lines across targets were also nice, but I figure they would be a better fit for CMake's documentation.

PR #992, opened 2022-01-09
[ci] Fix YAML syntax

kimgr:
In the previous commit, I merged overlooking the fact that it didn't work.

Apparently push and pull_request can be combined with schedule if they're all
dict forms; the former two can have empty values.


kimgr:
I know the build is still failing, but that's because of breaking upstream changes. I'll address those separately once the updated CI is on master.

PR #997, opened 2022-01-28
Fix compilation problem with latest clang

### LINKED ISSUE #996: Compile error with latest Clang
Getting the following compiling with Clang trunk:

```bash
external/include_what_you_use/iwyu.cc:2391:59: error: cannot initialize a parameter of type 'const clang::DirectoryLookup **' with an lvalue of type 'const clang::DirectoryLookup *'
            CurrentLoc(), "new", true, nullptr, use_file, curdir, nullptr,
                                                          ^~~~~~
```

Out of curiosity, have you considered merging the repo into the llvm-project? Seems like these issues could be caught more easily if it was all part of the same build?

carlosgalvezp:
Fixes #996


iwyu.cc
```diff
@@ -2386,7 +2386,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         using clang::DirectoryLookup;
         using clang::FileEntryRef;
         const FileEntry* use_file = CurrentFileEntry();
-        const DirectoryLookup* curdir = nullptr;
+        const DirectoryLookup** curdir = nullptr;
```
kimgr, line 2389: Let's just pass nullptr, no need for the named variable. That was the rationale for the change as described in the commit message here: https://github.com/llvm/llvm-project/commit/105c913156e94163c937d812992fe3a56210998c.

kimgr (changes requested):
See inline comment.

Speaking of which, could you reference the Clang commit (https://github.com/llvm/llvm-project/commit/105c913156e94163c937d812992fe3a56210998c) in the commit message instead of the IWYU issue (which has no relevant information).

Thanks!

iwyu.cc
kimgr: Oh, and you can also get rid of the `using clang::DirectoryLookup;` line in that case.

carlosgalvezp:
Thanks for the feedback! Will fix :)

carlosgalvezp:
Fixed now! Not quite sure how to mark "comments as done" so that the "Changes requested" text goes away, but should contain the changes you requested.

kimgr (approved):
Looks great, thanks!

PR #1009, opened 2022-02-14
Maintenance: redundant code removed

bolshakov-a:
Self-written using-decl lookup seems to became redundant after introducing
UsingType handling (PR #993)

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1051,16 +1051,6 @@ struct VisitorState {
   // instantiated calls, we can't store the exprs themselves, but have
   // to store their location.)
   set<SourceLocation> processed_overload_locs;
-
-  // When we see a using declaration, we want to keep track of what
-  // file it's in, because other files may depend on that using
-  // declaration to get the names of their types right.  We want to
-  // make sure we don't replace an #include with a forward-declare
-  // when we might need the #include's using declaration.
-  // The key is the type being 'used', the FileEntry is the file
-  // that has the using decl.  If there are multiple using decls
-  // for a file, we prefer the one that has NamedDecl in it.
-  multimap<const NamedDecl*, const UsingDecl*> using_declarations;
 };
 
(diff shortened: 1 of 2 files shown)


kimgr:
Interesting, I didn't see that coming! I'll see if I can dig out the original contributor, and they can test on the code base that triggered this solution.  Just to make sure our test suite hasn't missed something. Would be great if the AST now carries enough information.

kimgr:
So the relevant tests fail if you comment out the change from #993?

bolshakov-a:
> So the relevant tests fail if you comment out the change from https://github.com/include-what-you-use/include-what-you-use/pull/993?

Yes, badinc and include_with_using are failed in that case.

kimgr:
I'm not in front of the code right now, but there were more testcases added together with that using handling, wasn't there?

bolshakov-a:
No, include_with_using was seemingly the only test case added with the code under discussion:
https://github.com/include-what-you-use/include-what-you-use/commit/4813f2716ed407d374a8d78c984959159d2ec4f0

kimgr:
Sorry, I was thinking about 4c2e039e9dc2e308d960f25618b51345c3b600a6 (of which some disappears here too).

I think the one you cited covers `using` declarations themselves, and the one I was thinking of covers the actual dependencies pulled in via a `using` declaration. Or something to that effect :-)

@cdglove Would be great if you could swap in this change from 2015 and weigh in on whether this makes sense. Thanks!

kimgr:
@bolshakov-a Sorry it took me so long to circle back to this. I've read up and played around with this branch, and I think it makes sense. I have some ideas for further cleanups based on this work, but let's merge this first.

kimgr:
@bolshakov-a Sorry it took me so long to circle back to this. I've read up and played around with this branch, and I think it makes sense. I have some ideas for further cleanups based on this work, but let's merge this first.

PR #1026, opened 2022-04-12
Add additional <termios.h> mappings

fanquake:
```cpp
include <bits/termios-c_lflag.h>  // for ECHO
include <bits/termios-struct.h>   // for termios
include <bits/termios-tcflow.h>   // for TCSANOW

void SetStdinEcho(bool enable)
{
    struct termios tty;
    tcgetattr(STDIN_FILENO, &tty);
    if (!enable) {
        tty.c_lflag &= ~ECHO;
    } else {
        tty.c_lflag |= ECHO;
    }
    (void)tcsetattr(STDIN_FILENO, TCSANOW, &tty);
}
```


kimgr:
Unfortunately mappings need to go in two places -- where you made the change, and in `include_picker.cc`, these look like they belong in the `libc_include_map` array.

Could you also show a full example (including reduced source code) in the commit message, so `git log --grep` makes it easier to cross-reference? Thanks!

fanquake:
Thanks. Apologies for the slow follow up. I've added the mappings to `libc_include_map`, and added a minified example of the source code to the commit message.

kimgr:
Thank you! The CI failed due to an upstream Clang change, so I updated that in https://github.com/include-what-you-use/include-what-you-use/pull/1027. Please rebase once that's merged, so we can see that CI runs green before merge here. I'll merge that within half an hour.

fanquake:
> Please rebase once that's merged,

No worries. Have rebased on master.

kimgr:
Very nice, thank you!

alejandro-colomar:
Hi,

I'm not sure about the correctness of this change.  `termios` is a complex thing.

Please check some relevant threads in the linux-man@ mailing list:
https://lore.kernel.org/linux-man/20210730000915.d6ieqiuqah4tjjxf@pali/T/#u
https://lore.kernel.org/linux-man/20210730105353.10424-1-pali@kernel.org/T/

Maybe Pali Rohár <pali@kernel.org> <https://github.com/pali> can help.  (I ignore if github will notify him by mentioning his github URL.)

kimgr:
@alejandro-colomar Is that discussion for kernel-side `<termios.h>` or does it apply to userland `<termios.h>` as well, or are they one and the same?

alejandro-colomar:
> @alejandro-colomar Is that discussion for kernel-side `<termios.h>` or does it apply to userland `<termios.h>` as well, or are they one and the same?

I don't know if they are the same, I'm only a user-space programmer :)

But yes, it affects user space, AFAIR.

pali:
Hello!

There are two incompatible and different definitions of termios macros. One is in `<termios.h>` and used by glibc and second is `<asm/termbits.h>` (included also by `<linux/termios.h>`) and used by ioctls/syscalls.

So, if you are using libc functions like `tcgetattr()` or `tcsetattr()` then you must include `<termios.h>`. And if you are using `ioctl(TCGETS)` or `ioctl(TIOCMGET)` like in `ioctl_tty(2)` manpage (as mentioned in second linked patch) then you must include `<linux/termios.h>` or `<asm/termbits.h>`.

This _mess_ means that you cannot use both `tcgetattr()` and `ioctl(TIOCMGET)` in one source file as you cannot include both `<termios.h>` and `<asm/termbits.h>` at the same time.

kimgr:
@pali Thanks for the additional info!

So it seems to me this mapping from `<bits/termios.h>` to `<termios.h>` is safe and sound. It might be prudent to map all the ioctl IDs to `<asm/termios.h>`, but I suppose that can go in on top of this change if necessary.

alejandro-colomar:
> @pali Thanks for the additional info!
> 
> So it seems to me this mapping from `<bits/termios.h>` to `<termios.h>` is safe and sound.

Yes, I think this PR is safe.

> It might be prudent to map all the ioctl IDs to `<asm/termios.h>`, but I suppose that can go in on top of this change if necessary.

Not sure.  The IOCTL IDs come from `<sys/ioctl.h>`, if you mean for example `TCGETS` or `TIOCMGET`.

PR #994, opened 2022-01-11
[cmake] Include headers correctly for in-tree builds.

bc-lee:
When trying to build in-tree, I found that neither
`LLVM_INCLUDE_DIRS` nor `CLANG_INCLUDE_DIRS` were not defined.
Fix it by manually including the correct headers.

Also,  add a guide for building IWYU with LLVM source code.

PR #1029, opened 2022-04-23
CMake cleanup

kimgr:
Various and sundry cleanups for better consistency and style in CMakeLists.txt


kimgr:
In preparation for integrating something like #994.

PR #1032, opened 2022-04-24
Support LLVM external build of IWYU

kimgr:
Supersedes #994 after a bunch of cleanup refactoring in cmake and docs.

Thanks to @bc-lee for doing the original legwork!


kimgr:
Tested this with:

* IWYU build against installed LLVM-15 apt packages
* IWYU build against separately built LLVM mainline build tree
* LLVM build with IWYU as an external project

bc-lee:
I tested this to build using LLVM with IWYU as an external project, and it worked. Thank you!

Issue #981, opened 2021-12-14
Qt mapping file causes major slowdown


firewave:
I am using the `iwyu` package in version `8.17-1` on Kali Linux rolling (based on Debian - http://http.kali.org/pool/main/i/iwyu/) and generated a Qt mapping file using the latest script from the repo:

```
python3 iwyu-mapgen-qt.py /usr/include/x86_64-linux-gnu/qt5/ > qt5.imp
```

Without additional parameters:
```
real    0m1.843s
user    0m0.844s
sys     0m0.188s
```

With `--mapping_file=qt5.imp`:
```
real    0m37.874s
user    0m36.703s
sys     0m0.313s
```

The file is question is not even using Qt at all. The file is `tinyxml2.cpp` from https://github.com/danmar/cppcheck/tree/main/externals/tinyxml2. It is not related to the file it is just the first file which is being analyzed when using the compilation database and already shows the problem.

A few versions back I did not recognize such a slowdown.

kimgr:
Interesting. The Qt mappings are heavy on regex mappings to cover for `<>` and `""` transparency, and they are pretty exhaustive. Qt 5.11 mappings in tree contain about 2300 include mappings and 1500 symbol mappings.

I suspect the regex form causes a linear scan for each include mapping. I wonder if it's time to think through removing quotes from mappings. I've been putting it off because it feels like it might turn into an impossible problem.

firewave:
FYI It's so slow it's essentially broken. With other files it takes 10+ minutes until the first result even shows up whereas without the mappings the whole bunch is done in less than 3.

> Interesting. The Qt mappings are heavy on regex mappings to cover for `<>` and `""` transparency, and they are pretty exhaustive. Qt 5.11 mappings in tree contain about 2300 include mappings and 1500 symbol mappings.

In my case it is Qt 5.15.2 and the mappings files contains 2973 lines.
 
> I suspect the regex form causes a linear scan for ~each include mapping~ every time a header mapping is attempted. I wonder if it's time to think through removing quotes from mappings. I've been putting it off because it feels like it might turn into an impossible problem.

That's what I thought as well while looking at the file.

I am curious why this just started happening with the one of the previous versions. It was still working fine in April (I don't know which version it was since Kali/Debian is usually a bit behind - this time they updated extremely fast).

kimgr:
[...]

PR #1034, opened 2022-04-24
Add --regex=dialect option

kimgr:
@firewave I decided to give this a try to work around #981.

@zimin2000 This PR switches the default back to llvm::Regex for reasons discussed in the commit message, but adds an override command-line switch. I trust that works for you as well?

### DIFF
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -1425,8 +1425,8 @@ void IncludePicker::ExpandRegexes() {
     const string& hdr = incmap.first;
     for (const string& regex_key : filepath_include_map_regex_keys) {
       const vector<MappedInclude>& map_to = filepath_include_map_[regex_key];
-      // Enclose the regex in ^(...)$ for full match.
-      std::string regex("^(" + regex_key.substr(1) + ")$");
+      // Enclose the regex in ^...$ for full match.
+      std::string regex("^" + regex_key.substr(1) + "$");
       if (RegexMatch(regex_dialect, hdr, regex) &&
           !ContainsQuotedInclude(map_to, hdr)) {
         Extend(&filepath_include_map_[hdr], filepath_include_map_[regex_key]);
@@ -1435,7 +1435,7 @@ void IncludePicker::ExpandRegexes() {
       }
     }
     for (const string& regex_key : friend_to_headers_map_regex_keys) {
-      std::string regex("^(" + regex_key.substr(1) + ")$");
+      std::string regex("^" + regex_key.substr(1) + "$");
       if (RegexMatch(regex_dialect, hdr, regex)) {
         InsertAllInto(friend_to_headers_map_[regex_key],
                       &friend_to_headers_map_[hdr]);
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -16,7 +16,6 @@
 // not hash_map: it's not as portable and needs hash<string>.
 #include <map>                          // for map, map<>::mapped_type, etc
 #include <memory>
-#include <regex>
 #include <string>                       // for string, basic_string, etc
 #include <system_error>                 // for error_code
 #include <utility>                      // for pair, make_pair
@@ -25,6 +24,7 @@
 #include "iwyu_location_util.h"
 #include "iwyu_path_util.h"
 #include "iwyu_port.h"
+#include "iwyu_regex.h"
 #include "iwyu_stl_util.h"
 #include "iwyu_string_util.h"
 #include "iwyu_verrs.h"
@@ -1424,17 +1424,16 @@ void IncludePicker::ExpandRegexes() {
     for (const string& regex_key : filepath_include_map_regex_keys) {
       const vector<MappedInclude>& map_to = filepath_include_map_[regex_key];
       // Enclose the regex in ^(...)$ for full match.
-      std::regex regex(std::string("^(" + regex_key.substr(1) + ")$"));
-      if (std::regex_match(hdr, regex) &&
-          !ContainsQuotedInclude(map_to, hdr)) {
+      std::string regex("^(" + regex_key.substr(1) + ")$");
+      if (RegexMatch(hdr, regex) && !ContainsQuotedInclude(map_to, hdr)) {
         Extend(&filepath_include_map_[hdr], filepath_include_map_[regex_key]);
         MarkVisibility(&include_visibility_map_, hdr,
                        include_visibility_map_[regex_key]);
       }
     }
     for (const string& regex_key : friend_to_headers_map_regex_keys) {
-      std::regex regex(std::string("^(" + regex_key.substr(1) + ")$"));
-      if (std::regex_match(hdr, regex)) {
+      std::string regex("^(" + regex_key.substr(1) + ")$");
+      if (RegexMatch(hdr, regex)) {
         InsertAllInto(friend_to_headers_map_[regex_key],
                       &friend_to_headers_map_[hdr]);
       }
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -1425,9 +1425,7 @@ void IncludePicker::ExpandRegexes() {
     const string& hdr = incmap.first;
     for (const string& regex_key : filepath_include_map_regex_keys) {
       const vector<MappedInclude>& map_to = filepath_include_map_[regex_key];
-      // Enclose the regex in ^...$ for full match.
-      std::string regex("^" + regex_key.substr(1) + "$");
-      if (RegexMatch(regex_dialect, hdr, regex) &&
+      if (RegexMatch(regex_dialect, hdr, regex_key.substr(1)) &&
           !ContainsQuotedInclude(map_to, hdr)) {
         Extend(&filepath_include_map_[hdr], filepath_include_map_[regex_key]);
         MarkVisibility(&include_visibility_map_, hdr,
@@ -1435,8 +1433,7 @@ void IncludePicker::ExpandRegexes() {
       }
     }
     for (const string& regex_key : friend_to_headers_map_regex_keys) {
-      std::string regex("^" + regex_key.substr(1) + "$");
-      if (RegexMatch(regex_dialect, hdr, regex)) {
+      if (RegexMatch(regex_dialect, hdr, regex_key.substr(1))) {
         InsertAllInto(friend_to_headers_map_[regex_key],
                       &friend_to_headers_map_[hdr]);
       }
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -1219,8 +1219,10 @@ bool MappedInclude::HasAbsoluteQuotedInclude() const {
   return IsAbsolutePath(path);
 }
 
-IncludePicker::IncludePicker(bool no_default_mappings)
-    : has_called_finalize_added_include_lines_(false) {
+IncludePicker::IncludePicker(bool no_default_mappings,
+                             RegexDialect regex_dialect)
+    : has_called_finalize_added_include_lines_(false),
+      regex_dialect(regex_dialect) {
   if (!no_default_mappings) {
     AddDefaultMappings();
   }
@@ -1425,7 +1427,8 @@ void IncludePicker::ExpandRegexes() {
       const vector<MappedInclude>& map_to = filepath_include_map_[regex_key];
       // Enclose the regex in ^(...)$ for full match.
       std::string regex("^(" + regex_key.substr(1) + ")$");
-      if (RegexMatch(hdr, regex) && !ContainsQuotedInclude(map_to, hdr)) {
+      if (RegexMatch(regex_dialect, hdr, regex) &&
+          !ContainsQuotedInclude(map_to, hdr)) {
         Extend(&filepath_include_map_[hdr], filepath_include_map_[regex_key]);
         MarkVisibility(&include_visibility_map_, hdr,
                        include_visibility_map_[regex_key]);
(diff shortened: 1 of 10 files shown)


kimgr:
Btw, if anyone has ideas for how to solve this _without_ additional command-line switches, I'd love to hear them.

The higher-level mapping flow in `ExpandRegexes` looks pretty crazy, but I'm a little low on brain cycles to come up with something better that preserves the current behavior. It's no wonder it's slow now that it evalutes N headers x M regexes where both N and M are potentially large.

firewave:
I would suggest to use a less explicit name for the switch. That would allow you to switch to another implementation which provides the additional features and is  also performant in the future without any changes.

I think having such a switch is fine. Seems similar to `grep -E` / `egrep`.

On a side note - I saw by accident that starting with version 1.76 Boost.Regex is header-only and "standalone".

kimgr:
@firewave Thanks for taking a look! I wanted to avoid the "extended"/"-E" phrasing, because this is not Basic vs Extended Regex -- those words have a very specific meaning as described here: https://www.gnu.org/software/grep/manual/html_node/Basic-vs-Extended.html. I never found any good metaphorical spelling, so I decided to make it more explicit.

But if you have other suggestions around what to call the dialect, I'd love to hear them. I'm putting off merging this because I'm not happy about supporting a sloppy switch forever.

firewave:
Sorry for the late reply. I thought I already did...

> @firewave Thanks for taking a look! I wanted to avoid the "extended"/"-E" phrasing, because this is not Basic vs Extended Regex -- those words have a very specific meaning as described here: https://www.gnu.org/software/grep/manual/html_node/Basic-vs-Extended.html. I never found any good metaphorical spelling, so I decided to make it more explicit.

This is not a `grep` drop-in so that should not be a problem. I see no mention of "extended" applying to some specification so it seems they just defined it in a way and you should be able to do the same in your application.

> But if you have other suggestions around what to call the dialect, I'd love to hear them. I'm putting off merging this because I'm not happy about supporting a sloppy switch forever.

I think having the name tied to a specific implementation is "sloppy". Having that option to actually have it work properly in a timely fashion is not sloppy at all. Having an unusable implementation in `std` forcing people to resort to third-party solution is beyond sloppy...

I would really like to have usable Qt include support again in the upcoming version.

kimgr:
@firewave Here we go with a revised version. Please review.

iwyu_regex.cc
```diff
@@ -0,0 +1,47 @@
...
+bool RegexMatch(RegexDialect dialect, const std::string& str,
+                const std::string& pattern) {
+  switch (dialect) {
+    case RegexDialect::LLVM: {
+      llvm::Regex r(pattern);
+      return r.match(str);
+    }
+
+    case RegexDialect::ECMAScript: {
+      std::regex r(pattern, std::regex_constants::ECMAScript);
```
firewave, line 40: I wasn't aware it supports different grammars.

https://en.cppreference.com/w/cpp/regex/basic_regex

Does using `optimize` increase the performance? Can we use `basic` instead of using the LLVM implementation?

firewave: https://www.regular-expressions.info/stdregex.html

This goes a bit more into detail what the different grammars do.

kimgr: With this patch unchanged:
```
$ cat tests/981/t.cc 
#include <string>

$ time include-what-you-use -Xiwyu --regex=ecmascript -Xiwyu --mapping_file=qt5_11.imp tests/981/t.cc 

tests/981/t.cc should add these lines:

tests/981/t.cc should remove these lines:
- #include <string>  // lines 1-1

The full include-list for tests/981/t.cc:
---

real    0m5,865s
user    0m5,860s
sys     0m0,004s

$ time include-what-you-use -Xiwyu --regex=llvm -Xiwyu --mapping_file=qt5_11.imp tests/981/t.cc 

tests/981/t.cc should add these lines:

tests/981/t.cc should remove these lines:
- #include <string>  // lines 1-1

The full include-list for tests/981/t.cc:
---

real    0m0,425s
user    0m0,417s
sys     0m0,008s
```
Then I applied the following diff, built and ran again:
```
$ git diff
diff --git a/iwyu_regex.cc b/iwyu_regex.cc
index dca92be..c469fe5 100644
--- a/iwyu_regex.cc
+++ b/iwyu_regex.cc
@@ -32,8 +32,10 @@ bool RegexMatch(RegexDialect dialect, const std::string& str,
                 const std::string& pattern) {
   switch (dialect) {
     case RegexDialect::LLVM: {
-      llvm::Regex r(pattern);
-      return r.match(str);
+      std::regex r(pattern, std::regex_constants::extended);
+      return std::regex_match(str, r);
+      // llvm::Regex r(pattern);
+      // return r.match(str);
     }
 
     case RegexDialect::ECMAScript: {

$ time include-what-you-use -Xiwyu --regex=llvm -Xiwyu --mapping_file=qt5_11.imp tests/981/t.cc 

tests/981/t.cc should add these lines:

tests/981/t.cc should remove these lines:
- #include <string>  // lines 1-1

The full include-list for tests/981/t.cc:
---

real    0m5,986s
user    0m5,952s
sys     0m0,012s
```
(note `llvm` is now `std::regex::extended`, LLVM's Regex supports extended regex, so this is a fair comparison)

Just to make sure, I tested with `basic` too:
```
$ git diff
diff --git a/iwyu_regex.cc b/iwyu_regex.cc
index dca92be..5728831 100644
--- a/iwyu_regex.cc
+++ b/iwyu_regex.cc
@@ -32,8 +32,10 @@ bool RegexMatch(RegexDialect dialect, const std::string& str,
                 const std::string& pattern) {
   switch (dialect) {
     case RegexDialect::LLVM: {
-      llvm::Regex r(pattern);
-      return r.match(str);
+      std::regex r(pattern, std::regex_constants::basic);
+      return std::regex_match(str, r);
+      // llvm::Regex r(pattern);
+      // return r.match(str);
     }
 
     case RegexDialect::ECMAScript: {

$ time include-what-you-use -Xiwyu --regex=llvm -Xiwyu --mapping_file=qt5_11.imp tests/981/t.cc 

tests/981/t.cc should add these lines:

tests/981/t.cc should remove these lines:
- #include <string>  // lines 1-1

The full include-list for tests/981/t.cc:
---

real    0m5,930s
user    0m5,910s
sys     0m0,016s
```

So `std::regex` seems objectively worse. :-)

kimgr: I would expect `optimize` to actually make things worse, since we create a new `regex` object every time we evaluate a pattern -- `optimize` trades construction time for evaluation time.

I tested it both with `basic` and `ECMAScript`, but could not observe any difference in runtime.

kimgr: (I have a pretty cool experiment here with a runtime-polymorphic regex class where we could potentially cache the compiled regex to make some gains, but I wanted to start with this simpler variation first. Take a look here if you're interested: https://github.com/include-what-you-use/include-what-you-use/compare/master...kimgr:include-what-you-use:polyregex-variant)

firewave: > I would expect `optimize` to actually make things worse, since we create a new `regex` object every time we evaluate a pattern -- `optimize` trades construction time for evaluation time.
> 

☹
Just for giggles give it a spin (since I can't). The evaluation time is bad - maybe the construction time is not as bad. Unfortunately I see no `constexpr` anywhere in the signatures.

firewave: > (I have a pretty cool experiment here with a runtime-polymorphic regex class where we could potentially cache the compiled regex to make some gains, but I wanted to start with this simpler variation first. Take a look here if you're interested: [master...kimgr:include-what-you-use:polyregex-variant](https://github.com/include-what-you-use/include-what-you-use/compare/master...kimgr:include-what-you-use:polyregex-variant))

Yeah - I don't get it 🥴. Some patterns I haven't used yet and modern C++ are things that make my head hurt.

kimgr: I already did give it a spin, as mentioned up-thread
> I tested it both with basic and ECMAScript, but could not observe any difference in runtime.

firewave: > I already did give it a spin, as mentioned up-thread
> 
> > I tested it both with basic and ECMAScript, but could not observe any difference in runtime.

Ah, sorry. I misunderstood that line.

iwyu_include_picker.cc
```diff
@@ -1424,17 +1426,17 @@ void IncludePicker::ExpandRegexes() {
     for (const string& regex_key : filepath_include_map_regex_keys) {
       const vector<MappedInclude>& map_to = filepath_include_map_[regex_key];
       // Enclose the regex in ^(...)$ for full match.
-      std::regex regex(std::string("^(" + regex_key.substr(1) + ")$"));
-      if (std::regex_match(hdr, regex) &&
+      std::string regex("^(" + regex_key.substr(1) + ")$");
+      if (RegexMatch(regex_dialect, hdr, regex) &&
           !ContainsQuotedInclude(map_to, hdr)) {
```
firewave, line 1431: Assuming the regular expression is slower we should switch the expressions.

And if we do that we could inline `regex` so it is only constructed if actually needed.

kimgr: I have a followup planned to move the regex construction into RegexMatch itself, but I didn't want to complicate this PR with even more details.

By "switch the expressions", did you mean the map lookup and the RegexMatch call?

firewave: > By "switch the expressions", did you mean the map lookup and the RegexMatch call?

Yes.

kimgr: 👍 I'll look into it tomorrow. Possibly as a separate patch set. Thanks!

firewave:
The patch looks fine overall. Looking forward to finally enabling the Qt mappings again. 🤞

iwyu_include_picker.cc
kimgr: > Assuming the regular expression is slower we should switch the expressions.

I tried both, and could not see any consistent difference. At first it looked like `llvm::Regex` was faster than the `Contains...` lookup and `std::regex` was _slower_ than the `Contains...` lookup, forming a nice dilemma :)

But with some more test repetitions it was mostly a draw, so I'll leave it as-is.

Added two other optimizations/fixups on top, though. Please take another look.

firewave: > But with some more test repetitions it was mostly a draw, so I'll leave it as-is.

Exactly what I expected.

> Added two other optimizations/fixups on top, though. Please take another look.

Nice - I didn't look at those since I don't know what the code actually does. I also assumed they are "correct".

iwyu_include_picker.cc
```diff
@@ -1425,18 +1425,15 @@ void IncludePicker::ExpandRegexes() {
...
         Extend(&filepath_include_map_[hdr], filepath_include_map_[regex_key]);
         MarkVisibility(&include_visibility_map_, hdr,
                        include_visibility_map_[regex_key]);
       }
     }
     for (const string& regex_key : friend_to_headers_map_regex_keys) {
-      std::string regex("^" + regex_key.substr(1) + "$");
-      if (RegexMatch(regex_dialect, hdr, regex)) {
+      if (RegexMatch(regex_dialect, hdr, regex_key.substr(1))) {
```
firewave, line 1436: Since c++17 is the default now, `regex_key` could be a `std::string_view` thus saving the temporary object from `std::string::substr()`. But the `std::regex` interface doesn't support that so there would be still be a copy for that. Same applies above. But because of the anchoring there would still be a copy for the LLVM dialect as well. Still it would save a copy...

Not having `constexpr` or `std::string_view` support in `std::regex` makes me wonder if this was just dumped into the standard and abandoned afterwards since it is unusable anyways. Or maybe it is just too complex to support those.

And you call `std::string::substr()` in both cases. So maybe it could already be stored as that value.

kimgr: Now that that `substr` copy is an rvalue, I believe C++'s copy elision will take care of eliminating temporaries better.

The ultimate goal has to be to act on that `TODO` a few lines up; compile the regexes once in `ExtractKeysMarkedAsRegexes`, and then evaluate them on the include tree separately. But that's only possible once we can store a compiled regex, and that's where surgery like the `PolyRegex` example I posted earlier is necessary.

So I think we can just leave this alone for now, since `std::regex` is opt-in and `llvm::Regex` is back as the fast default.

firewave: > Now that that `substr` copy is an rvalue, I believe C++'s copy elision will take care of eliminating temporaries better.

Yes, but `std::string::substr()` will still result in the creation of a `std::string` object whereas `std::string_view::substr()` will put a range on the existing allocation for the `RegexMatch()` parameter. We cannot get rid of the new object inside that call.

kimgr: As you note, I guess no matter what, a fresh string is required for both `llvm::Regex` (for anchoring) and `std::regex` (because the constructor takes a `std::string` and not a `std::string_view`).

So I'm not sure what you're suggesting...?

At any rate, it would be nice to get this regex machinery merged. Any optimizations on top of that would be welcome if they make a significant improvement for `std::regex`.

firewave: > As you note, I guess no matter what, a fresh string is required for both `llvm::Regex` (for anchoring) and `std::regex` (because the constructor takes a `std::string` and not a `std::string_view`).
> 
> So I'm not sure what you're suggesting...?
> 
> At any rate, it would be nice to get this regex machinery merged. Any optimizations on top of that would be welcome if they make a significant improvement for `std::regex`.

Yes. It's micro-optimizations not necessary for this to land. But while you are at it. If I would build it myself I could just test it myself and post a PR but I am not gonna burden myself with another project right now. Still got way too much backlog. 😓

Go ahead and merge this.

kimgr: I tried some half-hearted attempt at having `ExtractKeysMarkedAsRegexes` return a vector of `std::string_view`, but the after-effects caused even more temporaries for the subsequent map lookups and regex eval. And I couldn't see any impact on execution time one way or the other.

So I think reducing temporaries here is a laudable goal, but it's probably not super-trivial.

Thanks for the review!

Issue #1039, opened 2022-05-01
Map htons(3) to <arpa/inet.h> as defined by POSIX


alejandro-colomar:
```
$ make lint-iwyu
LINT (iwyu)	tmp/src/man2/recvmmsg.2.d/recvmmsg.lint.iwyu.touch

tmp/src/man2/recvmmsg.2.d/recvmmsg.c should add these lines:
#include <bits/types/struct_iovec.h>  // for iovec
#include <netinet/in.h>               // for sockaddr_in, htonl, htons, INAD...
#include <time.h>                     // for timespec

tmp/src/man2/recvmmsg.2.d/recvmmsg.c should remove these lines:
- #include <netinet/ip.h>  // lines 3-3

The full include-list for tmp/src/man2/recvmmsg.2.d/recvmmsg.c:
#include <bits/types/struct_iovec.h>  // for iovec
#include <netinet/in.h>               // for sockaddr_in, htonl, htons, INAD...
#include <stdio.h>                    // for perror, printf
#include <stdlib.h>                   // for exit, EXIT_FAILURE, EXIT_SUCCESS
#include <string.h>                   // for memset
#include <sys/socket.h>               // for mmsghdr, bind, recvmmsg, socket
#include <time.h>                     // for timespec
---
make: *** [Makefile:388: tmp/src/man2/recvmmsg.2.d/recvmmsg.lint.iwyu.touch] Error 1
```

But POSIX defines `htons()` and `htonl()` in `<arpa/inet.h>`:
<https://pubs.opengroup.org/onlinepubs/9699919799/functions/htonl.html>

See also the manual page:

```
SYNOPSIS
       #include <arpa/inet.h>

       uint32_t htonl(uint32_t hostlong);
       uint16_t htons(uint16_t hostshort);

       uint32_t ntohl(uint32_t netlong);
       uint16_t ntohs(uint16_t netshort);

```
<https://www.man7.org/linux/man-pages/man3/htons.3.html>

kimgr:
Fixed by https://github.com/include-what-you-use/include-what-you-use/commit/fb7f8557a15310ded2e4a399810791124693c9b5

PR #1041, opened 2022-05-02
Add mappings for byteorder(3) functions

alejandro-colomar:
Solves https://github.com/include-what-you-use/include-what-you-use/issues/1039


kimgr:
This breaks the test suite. There's a test that reimplements `htons` using builtins, and the mappings force that use to use the system `<arpa/inet.h>`, which is exactly what we were trying to avoid.

I think a fix might be to run the test with `--no_default_mappings`. Or rename the symbols in the testcase to be more private.

Please follow up with a fix in the same commit, and describe your reasoning in the commit message.

alejandro-colomar:
> This breaks the test suite. There's a test that reimplements `htons` using builtins, and the mappings force that use to use the system `<arpa/inet.h>`, which is exactly what we were trying to avoid.

Okay.

> 
> I think a fix might be to run the test with `--no_default_mappings`. Or rename the symbols in the testcase to be more private.
> 

I very much prefer renaming the internal function.

I see the following:

```c
$ grepc '\w*[hn]to[nh]\w+'


./tests/cxx/macro_location-inet.h:12:

#define htons(x) bswap(x)


./tests/cxx/macro_location.cc:37:

int my_htons(int x) {
  return htons(x);
}
```

So, we can call it `iwyu_htons()`?

> Please follow up with a fix in the same commit, and describe your reasoning in the commit message.

Sure.

I'll put it for now in a separate commit, and when you confirm it's okay, I'll squash it into a single one.

```diff
diff --git a/tests/cxx/macro_location-inet.h b/tests/cxx/macro_location-inet.h
index f19e3fb..2728fe5 100644
--- a/tests/cxx/macro_location-inet.h
+++ b/tests/cxx/macro_location-inet.h
@@ -9,4 +9,4 @@
 
 #include "tests/cxx/macro_location-byteswap.h"
 
-#define htons(x) bswap(x)
+#define iwyu_htons(x) bswap(x)
diff --git a/tests/cxx/macro_location.cc b/tests/cxx/macro_location.cc
index bc91ec8..ebb3cc1 100644
--- a/tests/cxx/macro_location.cc
+++ b/tests/cxx/macro_location.cc
@@ -35,7 +35,7 @@ class ToBeDeclaredLater2 { };
 // Its nesting provokes something that isn't covered by the rest of the
 // macro_location suite.
 int my_htons(int x) {
-  return htons(x);
+  return iwyu_htons(x);
 }
 
 /**** IWYU_SUMMARY
```

kimgr:
Thanks, looks good! Can you squash, and make sure the commit message still provides a clear rationale for why the test symbol needed to be renamed? It's always a little suspicious when tests need to change when making unrelated changes, so it helps if it's documented somewhere. Cheers.

alejandro-colomar:
Sorry, I had squashed it a few days ago, but forgot to push it :)

kimgr:
> Can you squash, and make sure the commit message still provides a clear rationale for why the test symbol needed to be 
> renamed? It's always a little suspicious when tests need to change when making unrelated changes, so it helps if it's
> documented somewhere. 

The commit message is still completely opaque as to why the test changed, though. I'll fix it up and merge as a separate commit.

kimgr:
> The commit message is still completely opaque as to why the test changed, though. I'll fix it up and merge as a separate commit.

Ah, not _completely_ opaque, I just didn't read "internal" as having to do with the test suite. I'll clarify a little.

kimgr:
Done, merged as https://github.com/include-what-you-use/include-what-you-use/commit/fb7f8557a15310ded2e4a399810791124693c9b5

alejandro-colomar:
Thanks!

PR #1048, opened 2022-05-15
Do not warn on unnamed struct fields

jiixyj:
I've been getting warnings on unnamed struct fields inside structs (<https://gcc.gnu.org/onlinedocs/gcc/Unnamed-Fields.html>). The warning wants me to forward declare the anonymous struct type. Of course, this is not possible.

Weirdly, the warnings only happen when the unnamed struct field gets expanded from a macro. I haven't dug in too deep to find out why, but I found that changing the condition that decides if a struct type is forward declarable helps. Previously, it only checked if the type is a lambda. With this PR, it just checks if `getIdentifier()` returns `NULL`. I'm not sure if this check is more correct, but at least no tests seem to get broken.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -727,10 +727,10 @@ bool DeclCanBeForwardDeclared(const Decl* decl, string* reason) {
   if (isa<ClassTemplateDecl>(decl)) {
     // Class templates can always be forward-declared.
   } else if (const auto* record = dyn_cast<RecordDecl>(decl)) {
-    // Record decls can be forward-declared unless they denote a lambda
-    // expression; these have no type name to forward-declare.
-    if (record->isLambda()) {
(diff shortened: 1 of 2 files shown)


kimgr:
Very nice, thanks!

I suspect the macro-specific behavior is somehow due to the special processing in `GetCanonicalUseLocation` -- https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L1229 -- but this looks like the right fix no matter what.

kimgr:
Very nice, thanks!

I suspect the macro-specific behavior is somehow due to the special processing in `GetCanonicalUseLocation` -- https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L1229 -- but this looks like the right fix no matter what.

Issue #989, opened 2022-01-03
Segmentation fault after upgrade to 0.17 on macOS 11.6.2


srpgilles:
After upgrading to IWYU 0.17 on my BigSur installation (within Homebrew), IWYU stopped working on my codebase due to some pesky segmentation fault.

By trial and error on one file that was segmenting I was able to reduce to a small snippet that cause the segmentation fault:

````
#include <algorithm>
#include <cassert>
#include <vector>

int main()
{
    constexpr auto unused = static_cast<std::size_t>(-1);
    const auto Ndof = 100ul;

    std::vector<std::size_t> mapping(Ndof, unused);

    assert(std::none_of(mapping.cbegin(),
                        mapping.cend(),
                        [](auto value)
                    {
                        return value == unused;
                    }));

    return EXIT_SUCCESS;
}
````

This  segmentation fault is highly specific:

- If the algorithm is not inside the `assert` command, no segmentation fault.
- If I declare the lambda as a variable `auto is_unused = [](auto value) {... } ` and then use `is_unused` in the `std::none` algorithm, no segmentation fault.

Before the upgrade to newest IWYU (LLVM was also upgraded - I did it through `brew upgrade`), IWYU was working on my entire codebase without issue.

srpgilles:
I tried reinstalling 0.16 to check: the code abode works perfectly fine.

kimgr:
Thanks for the reduced repro! Unfortunately it seems to work fine on my Ubuntu/libstdc++-gcc9.3 setup. The main difference is that macOS has libc++ by default, so the standard library implementation probably triggers some path not seen with libstdc++.

Can you get a backtrace for the segfault?

kimgr:
@srpgilles This seems like a serious problem, can you please provide more details? Thanks.

srpgilles:
I concur it is a serious problem, but I must admit I am a bit at a loss to provide you more...

I build the snippet with the following CMakeLists.txt:

````
cmake_minimum_required(VERSION 3.12)

project(Debug_IWYU)

set(CMAKE_CXX_STANDARD 20 CACHE STRING "C++ standard; at least 20 is expected.")
set(CMAKE_C_COMPILER /usr/local/Cellar/llvm/13.0.0_2/bin/clang)
set(CMAKE_CXX_COMPILER /usr/local/Cellar/llvm/13.0.0_2/bin/clang++)
[...]

PR #1050, opened 2022-05-19
Robustify logging in the face of null pointers

kimgr:
The example described in #989 triggers an assertion failure (or segfault if built without assertions) in IWYU's main logic.
Unfortunately, if we up the log level, it also segfaults in logging, because we attempt to convert a null pointer to a string using `PrintableDecl`.

It seems prudent to teach the `Printable...` helpers to deal with null pointers, because it's nice if logging doesn't mask other bugs.

Clean up a bunch of small things, and print null values as `<null xyz>` consistently.


kimgr:
Oops, also add a new workaround for broken llvm-apt packaging.

PR #1052, opened 2022-05-19
create ```--comment_style``` arg for IWYU to allow people to adjust verbosity of 'why' comments

danielh2942:
In a project that I am working on it is required that I have the namespace that a function/class belongs to in the ```\\ for xyz``` comment. I figured that somebody else somewhere may need to do the same.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -2373,8 +2373,8 @@ def main(argv):
   parser.add_option('--nocomments', action='store_false', dest='comments')
 
   parser.add_option('--update_comments', action='store_true', default=False,
-                    help=('Update #include comments, even if no #include lines'
-                          ' are added or removed'))
+                    help=('Replace \'why\' comments with the ones provided by'
+                          ' IWYU'))
   parser.add_option('--noupdate_comments', action='store_false',
                     dest='update_comments')
 
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -76,6 +76,10 @@ length can still be exceeded with long filenames (default: 80).
 Do not add comments after includes about which symbols the header was required
 for.
 .TP
+.BI \-\-update_comments
+Print full include list with comments, even if there are no
+\(lqinclude-what-you-use\(rq violations.
+.TP
 .B \-\-no_default_mappings
 Do not use the default mappings.
 .TP
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -72,6 +72,24 @@ Maximum line length for includes.
 Note that this only affects the comments and their alignment, the maximum line
 length can still be exceeded with long filenames (default: 80).
 .TP
+.BI \-\-comment_style= verbosity
+Controls the style and verbosity of \(lqwhy\(rq comments at the end of
+suggested includes.
+.IR verbosity
+options are:
+.RS
+.TP
+.B none 
+No \(lqwhy\(rq comments.
+.TP 
+.B short 
+\(lqWhy\(rq comments include symbol names, but no namespaces.
+This is the default.
+.TP 
+.B long  
+\(lqWhy\(rq comments include function/class names with namespaces. 
+.RE
+.TP
 .B \-\-no_comments
 Do not add comments after includes about which symbols the header was required
 for.
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -74,9 +74,9 @@ length can still be exceeded with long filenames (default: 80).
 .TP
 .BI \-\-comment_style= verbosity
 Controls the style and verbosity of \(lqwhy\(rq comments at the end of
-suggested includes.
-.IR verbosity
-options are:
+suggested includes. Options for
+.I verbosity
+are:
 .RS
 .TP
 .B none 
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -91,7 +91,7 @@ static void PrintHelp(const char* extra_msg) {
          "        the maximum line length can still be exceeded with long\n"
          "        file names (default: 80).\n"
          "   --no_comments: do not add 'why' comments.\n"
-         "   --update_comments: always add 'why' comments, even if no\n"
+         "   --update_comments: update and insert 'why' comments, even if no\n"
          "        #include lines need to be added or removed.\n"
          "   --no_fwd_decls: do not use forward declarations.\n"
          "   --verbose=<level>: the higher the level, the more output.\n"
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -90,6 +90,12 @@ static void PrintHelp(const char* extra_msg) {
          "        Note that this only affects comments and alignment thereof,\n"
          "        the maximum line length can still be exceeded with long\n"
          "        file names (default: 80).\n"
+         "   --comment_style=<level> set verbosity of 'why' comments to\n"
+         "        one of the following values:\n"
+         "          none:  do not add 'why' comments.\n"
+         "          short: 'why' comments do not include namespaces.\n"
+         "          long:  'why' comments include namespaces.\n"
+         "        Default value is 'short'.\n"
          "   --no_comments: do not add 'why' comments.\n"
          "   --update_comments: update and insert 'why' comments, even if no\n"
          "        #include lines need to be added or removed.\n"
@@ -185,6 +191,7 @@ CommandlineFlags::CommandlineFlags()
       pch_in_code(false),
       no_comments(false),
       update_comments(false),
+      comments_with_namespace(false),
       no_fwd_decls(false),
       quoted_includes_first(false),
       cxx17ns(false),
@@ -205,6 +212,7 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
     {"prefix_header_includes", required_argument, nullptr, 'x'},
     {"pch_in_code", no_argument, nullptr, 'h'},
     {"max_line_length", required_argument, nullptr, 'l'},
+    {"comment_style", required_argument, nullptr, 'i'},
     {"no_comments", no_argument, nullptr, 'o'},
     {"update_comments", no_argument, nullptr, 'u'},
     {"no_fwd_decls", no_argument, nullptr, 'f'},
@@ -225,6 +233,18 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
       case 'n': no_default_mappings = true; break;
       case 'o': no_comments = true; break;
       case 'u': update_comments = true; break;
+      case 'i':
+        if (strcmp(optarg, "none") == 0) {
+          no_comments = true;
+        } else if (strcmp(optarg, "short") == 0) {
+          comments_with_namespace = false;
+        } else if (strcmp(optarg, "long") == 0) {
+          comments_with_namespace = true;
+        } else {
+          PrintHelp("FATAL ERROR: unknown comment style.");
+          exit(EXIT_FAILURE);
+        }
+        break;
       case 'f': no_fwd_decls = true; break;
       case 'x':
         if (strcmp(optarg, "add") == 0) {
--- a/tests/cxx/comment_style-d1.h
+++ b/tests/cxx/comment_style-d1.h
@@ -1,0 +1,17 @@
+//===--- comment_style-d1.h - test input file for iwyu --------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// this file exists to verify --comment_style works
+#include "tests/cxx/comment_style-i2.h"  // Transitive include
+
+namespace Foo {
+  int bar(int x) {
+    return x;
+  }
+};
--- a/tests/cxx/comment_style-i2.h
+++ b/tests/cxx/comment_style-i2.h
@@ -1,0 +1,14 @@
+//===--- comment_style-i2.h - test input file for iwyu --------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+namespace Bar {
+  int foo(int x) {
+    return x;
+  }
+};
--- a/tests/cxx/comment_style_update_long.cc
+++ b/tests/cxx/comment_style_update_long.cc
@@ -1,0 +1,34 @@
+//===--- comment_style_update_long.cc - test input file for iwyu ----------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --update_comments -Xiwyu --comment_style=long -I .
+
+// Test that passing --update_comments respects comment style.
+
+#include "tests/cxx/comment_style-d1.h"
+
+int main() {
+  // IWYU: Bar::foo is...*"tests/cxx/comment_style-i2.h"
+  Bar::foo(123);
+  Foo::bar(456);
+  return 0;
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/comment_style_update_long.cc should add these lines:
+#include "tests/cxx/comment_style-i2.h"
+
+tests/cxx/comment_style_update_long.cc should remove these lines:
+
+The full include-list for tests/cxx/comment_style_update_long.cc:
+#include "tests/cxx/comment_style-d1.h"  // for Foo::bar
+#include "tests/cxx/comment_style-i2.h"  // for Bar::foo
+
+***** IWYU_SUMMARY */
--- a/iwyu_globals.h
+++ b/iwyu_globals.h
@@ -90,6 +90,7 @@ struct CommandlineFlags {
   bool pch_in_code;   // Treat the first seen include as a PCH. No short option.
   bool no_comments;   // Disable 'why' comments. No short option.
   bool update_comments; // Force 'why' comments. No short option.
+  bool comments_with_namespace; // Show namespace in 'why' comments.
   bool no_fwd_decls;  // Disable forward declarations.
   bool quoted_includes_first; // Place quoted includes first in sort order.
   bool cxx17ns; // -C: C++17 nested namespace syntax
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1747,7 +1747,11 @@ void CalculateDesiredIncludesAndForwardDeclares(
       auto range = include_map.equal_range(use.suggested_header());
       for (auto it = range.first; it != range.second; ++it) {
         it->second->set_desired();
-        it->second->AddSymbolUse(use.short_symbol_name());
+        if (GlobalFlags().comments_with_namespace) {
+          it->second->AddSymbolUse(use.symbol_name());
+        } else {
+          it->second->AddSymbolUse(use.short_symbol_name());
+        }
       }
     }
   }
--- a/tests/cxx/comment_style_long.cc
+++ b/tests/cxx/comment_style_long.cc
@@ -1,0 +1,32 @@
+//===--- comment_style_long.cc - test input file for iwyu -----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --comment_style=long -I .
+
+// Test behavior is right with long comments
+
+#include "tests/cxx/comment_style-d1.h" // for bar
+
+int main() {
+  Foo::bar(1);
+  // IWYU: Bar::foo is...*"tests/cxx/comment_style-i2.h"
+  Bar::foo(2);
+  return 0;
+}
+
+/**** IWYU_SUMMARY
+tests/cxx/comment_style_long.cc should add these lines:
+#include "tests/cxx/comment_style-i2.h"
+
+tests/cxx/comment_style_long.cc should remove these lines:
+
+The full include-list for tests/cxx/comment_style_long.cc:
+#include "tests/cxx/comment_style-d1.h"  // for Foo::bar
+#include "tests/cxx/comment_style-i2.h"  // for Bar::foo
+***** IWYU_SUMMARY */
--- a/tests/cxx/comment_style_none.cc
+++ b/tests/cxx/comment_style_none.cc
@@ -1,0 +1,32 @@
+//===--- comment_style_none.cc - test input file for iwyu -----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --comment_style=none -I .
+
+// check that --comment_style=none adds no comments
+
+#include "tests/cxx/comment_style-d1.h"
+
+int main() {
+  Foo::bar(1);
+  // IWYU: Bar::foo is...*"tests/cxx/comment_style-i2.h"
+  Bar::foo(2);
+  return 0;
+}
+
+/**** IWYU_SUMMARY
+tests/cxx/comment_style_none.cc should add these lines:
+#include "tests/cxx/comment_style-i2.h"
+
+tests/cxx/comment_style_none.cc should remove these lines:
+
+The full include-list for tests/cxx/comment_style_none.cc:
+#include "tests/cxx/comment_style-d1.h"
+#include "tests/cxx/comment_style-i2.h"
+***** IWYU_SUMMARY */
--- a/tests/cxx/comment_style_short.cc
+++ b/tests/cxx/comment_style_short.cc
@@ -1,0 +1,32 @@
+//===--- comment_style_short.cc - test input file for iwyu ----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --comment_style=short -I .
+
+// check that --comment_style=short adds short comments
+
+#include "tests/cxx/comment_style-d1.h" // some Comment
+
+int main() {
+  Foo::bar(1);
+  // IWYU: Bar::foo is...*"tests/cxx/comment_style-i2.h"
+  Bar::foo(2);
+  return 0;
+}
+
+/**** IWYU_SUMMARY
+tests/cxx/comment_style_short.cc should add these lines:
+#include "tests/cxx/comment_style-i2.h"
+
+tests/cxx/comment_style_short.cc should remove these lines:
+
+The full include-list for tests/cxx/comment_style_short.cc:
+#include "tests/cxx/comment_style-d1.h"  // for bar
+#include "tests/cxx/comment_style-i2.h"  // for foo
+***** IWYU_SUMMARY */
--- a/tests/cxx/comment_style_update_none.cc
+++ b/tests/cxx/comment_style_update_none.cc
@@ -1,0 +1,34 @@
+//===--- comment_style_update_none.cc - test input file for iwyu ----------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --update_comments -Xiwyu --comment_style=none -I .
+
+// Test that --update_comments respects comment style.
+
+#include "tests/cxx/comment_style-d1.h" // for foo, bar
+
+int main() {
+  // IWYU: Bar::foo is...*"tests/cxx/comment_style-i2.h"
+  Bar::foo(123);
+  Foo::bar(456);
+  return 0;
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/comment_style_update_none.cc should add these lines:
+#include "tests/cxx/comment_style-i2.h"
+
+tests/cxx/comment_style_update_none.cc should remove these lines:
+
+The full include-list for tests/cxx/comment_style_update_none.cc:
+#include "tests/cxx/comment_style-d1.h"
+#include "tests/cxx/comment_style-i2.h"
+
+***** IWYU_SUMMARY */


danielh2942:
Added Tests

kimgr:
Thanks! An additional boolean flag, while convenient, seems a little confusing. It's not clear to me that `--show_namespaces` is intimately related to `--comments` and `--update_comments`.

I wonder if it would be better to let `--comments` take an argument, like `--comments={none,short,long}` and then likewise for `--update_comments`?

danielh2942:
Yeah that'd probably be the better call, I'll rework it.

danielh2942:
First pass at rework should, hopefully pass all the tests (Unless I missed something, currently compiling llvm from source on my machine so I can test an up-to-date IWYU version)

kimgr:
Thanks, looks nice! I'll jump into reviewing the details.

Can you rebase on latest master to work around the upstream packaging issue, so the build+test cycle runs? Thanks!

danielh2942:
done! :)

fix_includes_test.py
```diff
@@ -1901,25 +1902,25 @@ def testNocommentsFlag(self):
     iwyu_output = """\
 subdir/include_comments.cc should add these lines:
 #include "subdir/include_comments.h"
-#include <stdio.h>   // for printf(), etc.
-#include "used2.h"   // for Used2, Used2::Used2, Used2::~Used2, Used2::Used_Enum, operator==()
+#include <stdio.h>
```
kimgr, line 1905: Hmm, it looks like the semantics of `--nocomments` has changed from "don't touch the comments" to "remove the comments". I hadn't noticed that before, I wonder if that affects the design of the command-line arguments. I'd like to preserve existing behavior as much as possible, while extending it to support the longer comments with namespace.

kimgr: I don't have any specific action here, I need to think about this a little bit.

fix_includes_test.py
```diff
@@ -1965,6 +1966,52 @@ def testNoUpdateCommentsFlag(self):
...
+show_namespaces.cc should remove these lines:
+
+The full include-list for show_namespaces.cc:
+#include "first.h"  // for First::foo
+#include "second.h" // for Second::bar
+---
+"""
+    self.RegisterFileContents({'first.h': header1})
+    self.RegisterFileContents({'second.h': header2})
+    self.RegisterFileContents({'show_namespaces.cc': infile})
```
kimgr, line 2012: Filename should probably be something other than `show_namespace.cc` now

fix_includes_test.py
```diff
@@ -1965,6 +1966,52 @@ def testNoUpdateCommentsFlag(self):
     self.ProcessAndTest(iwyu_output,
                         unedited_files=['subdir/include_comments.cc'])
 
+  def testUpdateCommentsLong(self):
```
kimgr, line 1969: Similar here, I feel like we're missing testcases for `--comments`

iwyu_globals.cc
```diff
@@ -90,9 +90,15 @@ static void PrintHelp(const char* extra_msg) {
          "        Note that this only affects comments and alignment thereof,\n"
          "        the maximum line length can still be exceeded with long\n"
```
kimgr, line 91: We need to get the command-line arg changes into the manpage as well (https://github.com/include-what-you-use/include-what-you-use/blob/master/include-what-you-use.1). Let me know if you don't have a good way to test that (or the patience to do it :)), and I can try and push a patch when we've settled on the command-line design.

kimgr (commented):
Very nice! Please find questions and suggestions inline.

fix_includes.py
```diff
@@ -1906,7 +1906,8 @@ def _DecoratedMoveSpanLines(iwyu_record, file_lines, move_span_lines, flags):
     # to do the comment-replacing first.
     sort_key = firstline.line
     iwyu_version = iwyu_record.full_include_lines.get(m.group(1), '')
-    if _COMMENT_LINE_RE.search(iwyu_version):  # the iwyu version has comments
+    if _COMMENT_LINE_RE.search(iwyu_version) or (flags.update_comments == 'none'):
```
kimgr, line 1909: There's an 80-char line width limitation in python code too, so this overruns.

Luckily, python code usually doesn't use parentheses around `if` predicates, so if you remove the parens, it fits perfectly :)

fix_includes.py
```diff
@@ -1906,7 +1906,8 @@ def _DecoratedMoveSpanLines(iwyu_record, file_lines, move_span_lines, flags):
     # to do the comment-replacing first.
     sort_key = firstline.line
     iwyu_version = iwyu_record.full_include_lines.get(m.group(1), '')
-    if _COMMENT_LINE_RE.search(iwyu_version):  # the iwyu version has comments
+    if _COMMENT_LINE_RE.search(iwyu_version) or (flags.update_comments == 'none'):  
+                                               # the iwyu version has comments
```
kimgr, line 1910: I'd rephrase this and align it with `sort_key = ...` so you can remove the `# replace the comments` fragments:
```
if ...:
  # there are comments that need to be replaced
  sort_key = iwyu_version
```

fix_includes.py
```diff
@@ -2368,15 +2369,16 @@ def main(argv):
                           ' [default]'))
   parser.add_option('--noblank_lines', action='store_false', dest='blank_lines')
 
-  parser.add_option('--comments', action='store_true', default=False,
-                    help='Put comments after the #include lines')
-  parser.add_option('--nocomments', action='store_false', dest='comments')
+  parser.add_option('--comments', default='short',
+                    help='Put comments after the #include lines'
```
kimgr, line 2373: Python trick: use parentheses around multiline strings to help indent them correctly -- this is such common practice that most python editors do the right thing:
```
parser.add_option('--comments', default='short',
                    help=('Put comments after the #include lines'
                              ' options: {none, short, long} default: short'))
```

fix_includes.py
```diff
@@ -2368,15 +2369,16 @@ def main(argv):
...
+                         ' options: {none, short, long} default: short')
+  parser.add_option('--nocomments', action='store_const', const='none', dest='comments')
```
kimgr, line 2375: Overlong line, please stay under 80 here and throughout.

fix_includes.py
```diff
@@ -2445,8 +2447,15 @@ def main(argv):
       not flags.separate_project_includes.endswith('/')):
     flags.separate_project_includes += os.path.sep
 
-  if flags.update_comments:
-    flags.comments = True
+  if flags.comments not in ('long','short', 'none'):
```
kimgr, line 2450: Space after comma

fix_includes_test.py
```diff
@@ -33,8 +33,9 @@
 class FakeFlags(object):
   def __init__(self):
     self.blank_lines = False
-    self.comments = True
-    self.update_comments = False
+    self.comments = "short"
+    self.nocomments = False
```
kimgr, line 37: This seems redundant since `comments` is already `short`. I don't think `nocomments` is used directly, right?

fix_includes_test.py
```diff
@@ -33,8 +33,9 @@
 class FakeFlags(object):
   def __init__(self):
     self.blank_lines = False
-    self.comments = True
-    self.update_comments = False
+    self.comments = "short"
```
kimgr, line 36: Single quotes

fix_includes_test.py
```diff
@@ -1884,9 +1885,76 @@ def testIncludeComments(self):
...
+#include "used.h"
+#include "used2.h"
+---
+"""
+    self.RegisterFileContents({'subdir/include_no_comments.cc': infile})
+    self.ProcessAndTest(iwyu_output)
+
+  def testIncludeLongComments(self):
+    """Tests that we properly include long comments on #include lines."""
+    self.flags.comments='long'
```
kimgr, line 1923: Space around =

fix_includes_test.py
```diff
@@ -1884,9 +1885,76 @@ def testIncludeComments(self):
...
+#include "subdir/include_no_comments.h"
+#include <stdio.h>
+#include "used.h"
+#include "used2.h"
+---
+"""
+    self.RegisterFileContents({'subdir/include_no_comments.cc': infile})
+    self.ProcessAndTest(iwyu_output)
+
+  def testIncludeLongComments(self):
```
kimgr, line 1921: Wait a minute. Sorry I'm asking this at this point, but it hadn't occurred to me before: does the value of `--comments` affect the processing in `fix_includes.py` at all?

It looks like it's now:
* `--comments=none` - strip comments from IWYU output
* `--comments=short`, `--comments=long` - use comments for append-lines from IWYU output
* `--update_comments=none` - strip comments if they exist, even if IWYU suggests no changes
* `--update_comments=short`, `--update_comments=long` - use comments from IWYU output, even if IWYU suggests no changes

Maybe we/I need to take a step back again and think about what the switches really mean.

Intuitively, this makes me think that the `fix_includes.py` behavior should not have needed patching in order to add longer comments? The original behavior was (correct me if I'm wrong):

* `--comments` - use comments for append-lines from IWYU output
* `--update_comments` - use comments for all include-lines from IWYU output

which seems simple enough...

Sorry about the late realization, hopefully I've overlooked something :-)

fix_includes_test.py
```diff
@@ -1884,9 +1885,76 @@ def testIncludeComments(self):
...
+#include "used2.h"   // for NS::Used2, NS::Used2::Used2, NS::Used2::~Used2, NS::Used2::Used_Enum, operator==()
+---
+"""
+    self.RegisterFileContents({'subdir/include_long_comments.cc': infile})
+    self.ProcessAndTest(iwyu_output)
+
   def testNocommentsFlag(self):
     """Tests we properly don't include/modify comments with --nocomments."""
-    self.flags.comments = False
+    self.flags.nocomments = True
```
kimgr, line 1957: This test isn't very useful -- I don't think `nocomments` affects the execution at all (does it?).

So I suspect this just runs with the default value for `comments`.

fix_includes_test.py
```diff
@@ -1901,24 +1969,24 @@ def testNocommentsFlag(self):
...
+#include "used.h"  // my favorite #include!
+#include "used2.h"
 ---
 """
     self.RegisterFileContents({'subdir/include_comments.cc': infile})
     self.ProcessAndTest(iwyu_output)
 
   def testUpdateCommentsFlag(self):
-    """Tests we update comments with --update_comments."""
+    """Tests we update comments with --update_comments="short"."""
```
kimgr, line 1989: Let's remove quotes around "short"

include-what-you-use.1
```diff
@@ -72,10 +72,33 @@ Maximum line length for includes.
 Note that this only affects the comments and their alignment, the maximum line
 length can still be exceeded with long filenames (default: 80).
 .TP
+.BI \-\-comments [=Verbosity]
```
kimgr, line 75: We have optargs spelled all-lower for `--prefix_header_includes`, so I think this could just be `verbosity`. Here and below

include-what-you-use.1
```diff
@@ -72,10 +72,33 @@ Maximum line length for includes.
...
+short - \(lqwhy\(rq comments include function/class names but no namespaces
+.br 
+long  - \(lqwhy\(rq comments include function/class names with namespaces  
+.TP
 .B \-\-no_comments
 Do not add comments after includes about which symbols the header was required
 for.
 .TP
+.BI \-\-update_comments [=Verbosity]
+(disabled by default) updates prexisting comments with generated \(lqwhy\(rq comments
```
kimgr, line 92: `pre-existing`

iwyu_globals.cc
```diff
@@ -90,9 +90,15 @@ static void PrintHelp(const char* extra_msg) {
...
          "        file names (default: 80).\n"
+         "   --comments=<level> add 'why' comments of given verbosity to includes\n"
```
kimgr, line 93: Could you use the same formatting for values as `--prefix_header_includes`?

iwyu_globals.cc
```diff
@@ -90,9 +90,15 @@ static void PrintHelp(const char* extra_msg) {
...
+         "        level options: {none,\n"
+         "                        short,(just the function/class name)\n"
+         "                        long,(show namespaces)}\n"
+         "        default: short\n"
          "   --no_comments: do not add 'why' comments.\n"
-         "   --update_comments: always add 'why' comments, even if no\n"
-         "        #include lines need to be added or removed.\n"
+         "   --update_comments=<level>: update and insert 'why' comments with\n"
```
kimgr, line 99: ... and mirror whatever `--comments`ends up being here

iwyu_globals.cc
```diff
@@ -224,7 +232,31 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
       case 'm': mapping_files.push_back(optarg); break;
       case 'n': no_default_mappings = true; break;
       case 'o': no_comments = true; break;
-      case 'u': update_comments = true; break;
+      case 'u': 
+        // As update_comments overrides comment verboseness, there is no break here
```
kimgr, line 236: Let's rewrite this section as simply:
```
        update_comments = true;

        // Fall through to comment style parsing unless no value was provided
        if (!optarg)
          break;

        LLVM_FALLTHROUGH;
```

I don't think we need to explain fallthrough, just why we're sometimes avoiding it.

But looking at this it makes me wonder -- do we really need separate processing? `comments_with_namespace` is initialized to `false` in the constructor so if we removed the overwriting below if optarg is NULL, wouldn't it just work? Maybe I'm missing some other weird interaction :-)

iwyu_globals.cc
```diff
@@ -224,7 +232,31 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
...
+        }
+
+        if (strcmp(optarg, "none") == 0) {
+          no_comments = true;
+        } else if (strcmp(optarg, "short") == 0) {
+          comments_with_namespace = false;
+        } else if (strcmp(optarg, "long") == 0) {
+          comments_with_namespace = true;
+        } else {
+          PrintHelp("FATAL ERROR: unknown comment verboseness specifier");
```
kimgr, line 256: verbosity

tests/cxx/show_comments-d1.h
```diff
@@ -0,0 +1,23 @@
...
+#include "tests/cxx/show_comments-d2.h"  // Transitive include
+
+namespace Foo {
+    int bar(int x)
+    {
+        return x;
+    }
+};
+
+#endif
```
kimgr, line 23: Missing trailing newline

tests/cxx/show_comments-d1.h
```diff
@@ -0,0 +1,23 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// this file exists to verify show_comments works
+#ifndef SHOW_COMMENTS_D1_H
+#define SHOW_COMMENTS_D1_H
+
+#include "tests/cxx/show_comments-d2.h"  // Transitive include
```
kimgr, line 14: Ah, if you want to do transitive includes, that header should be called `-i1.h`, `i` for "indirect"

tests/cxx/show_comments-d1.h
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Foo {
+    int bar(int x)
```
kimgr, line 17: Let's clang-format tests too (2-space indentation, opening brace always on same line)

tests/cxx/show_comments_short.cc
```diff
@@ -0,0 +1,32 @@
...
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --comments=short -I .
+
+// check that --comments=shor adds short comments
```
kimgr, line 12: typo: shor -> short

tests/cxx/update_comments_long.cc
```diff
@@ -0,0 +1,32 @@
...
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --update_comments=long -I .
+
+// Test that passing --update_comments="long" provides appropriate
```
kimgr, line 12: No need for double-quotes there

tests/cxx/show_comments-d2.h
```diff
@@ -0,0 +1,20 @@
+//===--- show_comments-d2.h - test input file for iwyu --------------------===//
```
kimgr, line 1: I wonder if these test files would be better named `comment_style*`?

kimgr (commented):
Hey, thank you so much for being so forthcoming with accepting changes.

I'm beginning to lose my footing now, I'm no longer sure the `--comments=xx` + `--update_comments=xx` switch pair is the best design.

Let's follow up with a little design discussion in the PR comments (separate from the review, which is harder to track) and see if we can agree on a direction before I ask you to redo everything again :grimacing:

kimgr:
So it seems I didn't really fully understand what these switches did. It's pretty subtle how they mean one thing in `include-what-you-use` and another in `fix_includes.py`.

Here's how I currently understand it:

* `include-what-you-use` adds comments to all `#include` and forward-declare lines unless no changes are suggested
    * `--no_comments` can be added to disable comments
    * `--update_comments` can be added to force output including comments even if no changes are needed
* `fix_includes.py` applies changes to source files based on `include-what-you-use` output
    * `--comments` causes it to include the comments (if any) on added lines (if any)
    * `--nocomments` causes it to suppress comments even if they exist in the output (I think?)
    * `--update_comments` causes it to rewrite all remaining `#include`/forward-declare lines with comments from the output

So it's curious how the switches cooperate nicely when used together, but do entirely different things, e.g.

* `include-what-you-use --no_comments` + `fix_includes.py --comments` => no comments in IWYU output, nothing for fix_includes to do
* `include-what-you-use --update_comments` + `fix_includes.py --nocomments` => IWYU generates full output, fix_includes ignores it
* `include-what-you-use --update_comments` + `fix_includes.py --update_comments` => IWYU generates full output, fix_includes uses it to rewrite all `#include`/forward-declare lines
* `include-what-you-use --update_comments` + `fix_includes.py --comments` => IWYU generates full output, fix_includes only applies it to added lines
* etc.

Curious to hear your thoughts on this and how best to tweak it to include a comment style (which I think was what you really wanted to do)

danielh2942:
I realized I greatly misunderstood where in the overall process ```fix_includes.py``` is, I convinced myself ProcessIWYUOutput called IWYU without ever thinking of reading the function body for it. Thank you for clarifying. As such I could remove all  changes to it without affecting comment length whatsoever (one of the changes I added allows you to erase existing 'why' comments altogether and that was probably a bit of a strange addition in hindsight). 

I could however update the output of ```fix_includes.py --help``` to potentially remove some ambiguity I.E
```
--update_comments: replaces 'why' comments with the ones provided by IWYU
```

As for ```include-what-you-use``` itself, the optional parameters for ```--comments``` and ```--update_comments``` seem to be the best port of call as ```--verbose``` serves another purpose and adding a flag like ```--detail``` would make it confusing

kimgr:
> As for include-what-you-use itself, the optional parameters for --comments and --update_comments seem to be the best port of > call [...]

Unless I've misunderstood everything again :smile_cat:, I think it can be made even simpler for `include-what-you-use`:

* new switch `--comments=<style>`, where `<style>` defaults to `short`
* leave `--update_comments` as a bool value to indicate whether to print all include/fwddecl lines with comments (true) or just added lines (false)

I think this is exactly how `--update_comments` already works, so the only new behavior is the comment style indicated by `--comments`.

One drawback of that approach is that `--comments` now also means something different between `include-what-you-use` and `fix_includes.py`, just like `--update_comments` and `--no[_]comments` :shrug:.

So maybe it would be better to use a very explicit name, like `--comment_style=`, where `--comment_style=none` is a synonym for `--no_comments` and the latter could be deprecated over time.

I'm mostly just throwing out ideas, I haven't really had a chance to play this out in my head all the way.

danielh2942:
```--comment_style=``` would probably make the most sense yeah, I'll do that

**Comment by danielh2942** [comment]
summary of second wave changes
- Addressed feedback
- undid most changes to ```fix_includes.py``` and ```test_fix_includes.py``` as they didn't do what I initally thought they did (or anything for that matter)
- changed ```--comments=[verbosity]``` to ```--comments_style[=verbosity]``` 
- removed optional argument from ```--update_comments```
- adjusted documentation to reflect these changes

File: iwyu_globals.cc [hunk 28]
```diff
@@ -90,9 +90,15 @@ static void PrintHelp(const char* extra_msg) {
...
+         "        includes and are controlled with the following values:\n"
+         "          none:  no 'why' comments are added\n"
+         "          short: 'why' comments do not include namespaces\n"
+         "          long:  'why' comments include namespaces\n"
+         "        Default value is 'short'\n"
          "   --no_comments: do not add 'why' comments.\n"
-         "   --update_comments: always add 'why' comments, even if no\n"
-         "        #include lines need to be added or removed.\n"
+         "   --update_comments: update and insert 'why' comments\n"
+         "        disabled by default\n"
```
kimgr, line 101: I don't think we need this `disabled by default` line.

iwyu_globals.cc
```diff
@@ -90,9 +90,15 @@ static void PrintHelp(const char* extra_msg) {
...
+         "   --comments_style=<level> add 'why' comments of given verbosity to\n"
+         "        includes and are controlled with the following values:\n"
+         "          none:  no 'why' comments are added\n"
+         "          short: 'why' comments do not include namespaces\n"
+         "          long:  'why' comments include namespaces\n"
+         "        Default value is 'short'\n"
          "   --no_comments: do not add 'why' comments.\n"
-         "   --update_comments: always add 'why' comments, even if no\n"
-         "        #include lines need to be added or removed.\n"
+         "   --update_comments: update and insert 'why' comments\n"
```
kimgr, line 100: Let's end with a full stop like the other switches

iwyu_globals.cc
```diff
@@ -205,6 +212,7 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
     {"prefix_header_includes", required_argument, nullptr, 'x'},
     {"pch_in_code", no_argument, nullptr, 'h'},
     {"max_line_length", required_argument, nullptr, 'l'},
+    {"comments_style", optional_argument, nullptr, 'i'},
```
kimgr, line 215: As I mentioned in passing above, I think this can be `required_argument` now

iwyu_globals.cc
```diff
@@ -225,6 +233,23 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
...
+        }
+
+        if (strcmp(optarg, "none") == 0) {
+          no_comments = true;
+        } else if (strcmp(optarg, "short") == 0) {
+          comments_with_namespace = false;
+        } else if (strcmp(optarg, "long") == 0) {
+          comments_with_namespace = true;
+        } else {
+          PrintHelp("FATAL ERROR: unknown comment verbosity specifier");
```
kimgr, line 249: `comment verbosity specifier` -> `comment style`

iwyu_globals.cc
```diff
@@ -225,6 +233,23 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
       case 'n': no_default_mappings = true; break;
       case 'o': no_comments = true; break;
       case 'u': update_comments = true; break;
+      case 'i':
+        if (!optarg) {
```
kimgr, line 237: Once optarg is required, I think this can be removed.

tests/cxx/comment_style-d1.h
```diff
@@ -0,0 +1,22 @@
...
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// this file exists to verify show_comments works
+#ifndef COMMENT_STYLE_D1_H
```
kimgr, line 11: Include guards in IWYU tests are usually: `INCLUDE_WHAT_YOU_USE_TESTS_CXX_COMMENT_STYLE_H_` (see other test cases). But there's also a tendency to avoid them entirely if they're not needed, and I don't think they are here.

tests/cxx/comment_style-i2.h
```diff
@@ -0,0 +1,19 @@
+//===--- comment_style-i2.h - test input file for iwyu --------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef COMMENT_STYLE_D2_H
```
kimgr, line 10: `D2` instead of `I2`. But consider removing, per earlier comment.

tests/cxx/update_comments_long-d1.h
```diff
@@ -0,0 +1,27 @@
+//===--- update_comments_long-d1.h - test input file for iwyu -------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// This header exists to verify that --update_comments=long works
```
kimgr, line 10: This comment is no longer correct. If possible, I think you should:

* Remove the `update_comments*.h" header files
* Rename the new `update_comments*.cc` test cases to `comment_style_update*.cc`
* Reuse the `comment_style` headers for these tests

That way it's a little clearer how the two are related.

kimgr: I guess the trick is to avoid IWYU diagnostics for the update tests, but that's possible by including `comment_style-d1.h` and only using the `Foo::bar` symbol.

danielh2942:
Addressed new wave of feedback
summary:
- changed ```-comments_style``` to ```--comment_style```
- renamed some tests
- made verbosity specifier for ```--comment_style``` required 
- reworded parts of manpage

fix_includes.py
```diff
@@ -2373,8 +2373,7 @@ def main(argv):
   parser.add_option('--nocomments', action='store_false', dest='comments')
 
   parser.add_option('--update_comments', action='store_true', default=False,
-                    help=('Update #include comments, even if no #include lines'
-                          ' are added or removed'))
+                    help=('replace \'why\' comments with the ones provided by IWYU'))
```
kimgr, line 2376: This runs over 80 column limit, needs to be wrapped.
All other help texts use sentence case, so this should probably start with an upper-case R.

include-what-you-use.1
```diff
@@ -72,10 +72,29 @@ Maximum line length for includes.
 Note that this only affects the comments and their alignment, the maximum line
 length can still be exceeded with long filenames (default: 80).
 .TP
+.BI \-\-comment_style= verbosity
+Controls the style and verbosity of \(lqwhy\(rq comments at the end of suggested includes.
```
kimgr, line 76: This runs over the 80 column limit, please wrap.

include-what-you-use.1
```diff
@@ -72,10 +72,29 @@ Maximum line length for includes.
...
+.IR verbosity
+options are:
+.RS
+.TP
+.B none 
+No \(lqwhy\(rq comments.
+.TP 
+.B short 
+\(lqWhy\(rq comments include function/class names but no namespaces. This is the default.
```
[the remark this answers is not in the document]
kimgr: And keep an eye on line length, this is >80 now.

kimgr (commented):
I think the overall looks good now. Lots of small cosmetic and wording suggestions to harmonize with prevalent style. Thank you!

kimgr:
Also, let's talk about commit partitioning when this is done. I have a few ideas for how to split things so the history becomes obviously nice.

danielh2942:
Sure!

fix_includes.py
```diff
@@ -2373,8 +2373,8 @@ def main(argv):
   parser.add_option('--nocomments', action='store_false', dest='comments')
 
   parser.add_option('--update_comments', action='store_true', default=False,
-                    help=('Update #include comments, even if no #include lines'
-                          ' are added or removed'))
+                    help=('Replace \'why\' comments with the ones provided by'
+                    ' IWYU'))
```
kimgr, line 2377: Indent `' IWYU'` to align with `'Replace ...`

tests/cxx/comment_style_update_long.cc
```diff
@@ -0,0 +1,35 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --update_comments -Xiwyu --comment_style=long -I .
+
+// Test that passing --update_comments respects comment style.
+// namespaces in "// for XYZ" comments
```
kimgr, line 13: Leftover comment line

kimgr (commented):
Nice! A few small fixups remaining, but then I'd like to suggest to squash this into two commits:

* One to clarify the help text for `--update_comments` in `fix_includes.py` and `iwyu_globals.cc` and add `--update_comments` to manpage (something like `Improve documentation for --update_comments`)
* One to add support for `--comment_style`, with tests and manpage changes

Let me know if you need guidance with `git rebase` or if you want me to do it (will have to clone your branch where I can push to it, though).

danielh2942:
rebased it into two commits (+ commit for upstream fetch)

include-what-you-use.1
```diff
@@ -72,10 +72,32 @@ Maximum line length for includes.
 Note that this only affects the comments and their alignment, the maximum line
 length can still be exceeded with long filenames (default: 80).
 .TP
+.BI \-\-comment_style= verbosity
+Controls the style and verbosity of \(lqwhy\(rq comments at the end of
+suggested includes.
+.IR verbosity
+options are:
```
aaronpuchert, line 79:
```suggestion
suggested includes. Options for
.I verbosity
are:
```

To make sure we start the sentence with a capital letter. Also, to quote [Knuth](https://jmlr.csail.mit.edu/reviewing-papers/knuth_mathematical_writing.pdf): “Don’t start a sentence with a symbol.”

Also `.I` means italic, `.IR` means alternating italic and regular (“roman”). We have just one word here.

aaronpuchert (commented):
Changes to man page look good, just one minor suggestion.

include-what-you-use.1
danielh2942: Committed, thank you :)

kimgr:
> rebased it into two commits (+ commit for upstream fetch)

Nice! Can you also rebase on top of upstream instead of merging it in, so we don't have gratuitous merge commits on master after this is fast-forward merged? And squash the manpage fixup from @aaronpuchert into the `clarify purpose...` commit, we don't keep review feedback separate. It's more important that the history on master is clear.

danielh2942:
rebased but for some reason @aaronpucherts change would not go into mine so I had to add it separately again

kimgr:
Thank you! I can squash it and merge it locally. Will close this manually when it's pushed to master.

kimgr:
I found a few more things I wanted to tweak, so this is going in with the following additional changes:

* Keep switches alpha-sorted in manpage
* Strip trailing whitespace
* Use `symbol` instead of `function/class` for `long` too
* Remove full stops in value lists

I think these are all still in the spirit of the patch, so I'm hoping you're fine with squashing them into your commits in your name.

Will merge shortly.

```
diff --git a/include-what-you-use.1 b/include-what-you-use.1
index 20ea3d5..3132d92 100644
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -44,6 +44,24 @@ given glob pattern (in addition to the default of reporting for the input
 source file and associated header files).
 This flag may be specified multiple times to specify multiple glob patterns.
 .TP
+.BI \-\-comment_style= verbosity
+Controls the style and verbosity of \(lqwhy\(rq comments at the end of
+suggested includes. Options for
+.I verbosity
+are:
+.RS
+.TP
+.B none
+No \(lqwhy\(rq comments.
+.TP
+.B short
+\(lqWhy\(rq comments include symbol names, but no namespaces. This is the
+default.
+.TP
+.B long
+\(lqWhy\(rq comments include symbol names with namespaces.
+.RE
+.TP
 .B \-\-cxx17ns
 Suggest the more concise syntax for nested namespaces introduced in C++17.
 .TP
@@ -72,32 +90,10 @@ Maximum line length for includes.
 Note that this only affects the comments and their alignment, the maximum line
 length can still be exceeded with long filenames (default: 80).
 .TP
-.BI \-\-comment_style= verbosity
-Controls the style and verbosity of \(lqwhy\(rq comments at the end of
-suggested includes. Options for
-.I verbosity
-are:
-.RS
-.TP
-.B none 
-No \(lqwhy\(rq comments.
-.TP 
-.B short 
-\(lqWhy\(rq comments include symbol names, but no namespaces.
-This is the default.
-.TP 
-.B long  
-\(lqWhy\(rq comments include function/class names with namespaces. 
-.RE
-.TP
 .B \-\-no_comments
 Do not add comments after includes about which symbols the header was required
 for.
 .TP
-.BI \-\-update_comments
-Print full include list with comments, even if there are no
-\(lqinclude-what-you-use\(rq violations.
-.TP
 .B \-\-no_default_mappings
 Do not use the default mappings.
 .TP
@@ -143,6 +139,10 @@ Do not suggest that a file should add
 .IR foo.h " unless " foo.h
 is already visible in the file's transitive includes.
 .TP
+.BI \-\-update_comments
+Print full include list with comments, even if there are no
+\(lqinclude-what-you-use\(rq violations.
+.TP
 .BI \-\-verbose= level
 Set verbosity. At the highest level, this will dump the AST of the source file
 and explain all decisions.
diff --git a/iwyu_globals.cc b/iwyu_globals.cc
index 2096e96..8a32bda 100644
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -90,11 +90,11 @@ static void PrintHelp(const char* extra_msg) {
          "        Note that this only affects comments and alignment thereof,\n"
          "        the maximum line length can still be exceeded with long\n"
-         "   --comment_style=<level> set verbosity of 'why' comments to\n"
-         "        one of the following values:\n"
-         "          none:  do not add 'why' comments.\n"
-         "          short: 'why' comments do not include namespaces.\n"
-         "          long:  'why' comments include namespaces.\n"
+         "   --comment_style=<level> set verbosity of 'why' comments to one\n"
+         "        of the following values:\n"
+         "          none:  do not add 'why' comments\n"
+         "          short: 'why' comments do not include namespaces\n"
+         "          long:  'why' comments include namespaces\n"
          "        Default value is 'short'.\n"
          "   --update_comments: update and insert 'why' comments, even if no\n"
diff --git a/tests/cxx/comment_style-d1.h b/tests/cxx/comment_style-d1.h
index 6ff6043..a104aca 100644
--- a/tests/cxx/comment_style-d1.h
+++ b/tests/cxx/comment_style-d1.h
@@ -7,8 +7,7 @@
 //
 
-// this file exists to verify --comment_style works
-#include "tests/cxx/comment_style-i2.h"  // Transitive include
+#include "tests/cxx/comment_style-i2.h"
 
 namespace Foo {
   int bar(int x) {
diff --git a/tests/cxx/comment_style_long.cc b/tests/cxx/comment_style_long.cc
index 75b3feb..9296067 100644
--- a/tests/cxx/comment_style_long.cc
+++ b/tests/cxx/comment_style_long.cc
@@ -9,7 +9,7 @@
 
 // IWYU_ARGS: -Xiwyu --comment_style=long -I .
 
-// Test behavior is right with long comments
+// Test behavior is right with long comments.
 
 #include "tests/cxx/comment_style-d1.h" // for bar
 
diff --git a/tests/cxx/comment_style_none.cc b/tests/cxx/comment_style_none.cc
index ab8e09f..5e071ec 100644
--- a/tests/cxx/comment_style_none.cc
+++ b/tests/cxx/comment_style_none.cc
@@ -9,7 +9,7 @@
 
 // IWYU_ARGS: -Xiwyu --comment_style=none -I .
 
-// check that --comment_style=none adds no comments
+// Test that --comment_style=none adds no comments.
 
 #include "tests/cxx/comment_style-d1.h"
 
diff --git a/tests/cxx/comment_style_short.cc b/tests/cxx/comment_style_short.cc
index 6c3bf04..4879fdc 100644
--- a/tests/cxx/comment_style_short.cc
+++ b/tests/cxx/comment_style_short.cc
@@ -9,7 +9,7 @@
 
 // IWYU_ARGS: -Xiwyu --comment_style=short -I .
 
-// check that --comment_style=short adds short comments
+// Test that --comment_style=short adds short comments.
 
 #include "tests/cxx/comment_style-d1.h" // some Comment
```

kimgr:
Now on master as of https://github.com/include-what-you-use/include-what-you-use/commit/3f456f66c22767a19477d189deea6fdc169de90b. Thanks for working on this!

PR #1055, opened 2022-05-23
[fix_includes] Recognize namespace aliases

### LINKED ISSUE #1022: fix_includes erroneously removes namespace alias
When I run `fix_includes.py` on the following file (piping IWYU's stderr to it) with the `--nosafe_headers` option, it removes not just the unused header, but also the namespace alias near the bottom:
```c++
#pragma once

#include <stdint.h>

namespace outer {
namespace middle {
namespace inner {

enum Foo {
  FOO,
  BAR,
  BAZ,
};

}  // namespace inner
}  // namespace middle

namespace inner = middle::inner;

}  // namespace outer
```
It also removes the line with the closing brace for the `outer` namespace, which is even more problematic.

I ran the following command on the above file (named `foo.h` below) to reproduce the problem:
```console
include-what-you-use -x c++ foo.h 2>&1 | fix_includes.py --nosafe_headers
```

kimgr:
... as different from Allman-style namespace definitions.

Fixes #1022.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -104,7 +104,7 @@ _NAMESPACE_START_RE = re.compile(r'\s*(namespace\b[^{]*{\s*)+(//.*)?$|'
                                  r'\s*(HASH_NAMESPACE_DECLARATION_START)')
 # Also detect Allman and mixed style namespaces.  Use a continue regex for
 # validation and to correctly set the line info.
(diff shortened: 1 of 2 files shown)


gedankenexperimenter:
Looks good to me.  It works for me on my example, and on the original code that triggered the problem in the first place.  Thank you.

kimgr:
Thanks, let's get this on mainline!

PR #1057, opened 2022-05-25
Class member reporting improvements

bolshakov-a:
- Don't redundantly report all class/struct/union member uses, not only methods.
- Obsolete code removed, which led to ignoring typedef-prefixed names.
In fact, typedef-defining files may or may not provide full aliased type info. This change fixes a bug with absense of reporting a type under a nested typedef, enclosing class of which is referred to with another typedef (test case added).

Besides visual noise suppression and bugfix, these changes are needed for further improving of typedef handling. (I intend to make a PR after this one).

### DIFF
--- a/tests/cxx/iwyu_stricter_than_cpp-d3.h
+++ b/tests/cxx/iwyu_stricter_than_cpp-d3.h
@@ -1,0 +1,17 @@
+//===--- iwyu_stricter_than_cpp-d3.h - test input file for iwyu -----------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/iwyu_stricter_than_cpp-i3.h"
+#include "tests/cxx/iwyu_stricter_than_cpp-i4.h"
+
+typedef IndirectStruct3 IndirectStruct3ProvidingTypedef;
+typedef IndirectStruct4 IndirectStruct4ProvidingTypedef;
+
+using IndirectStruct3ProvidingAl = IndirectStruct3;
+using IndirectStruct4ProvidingAl = IndirectStruct4;
--- a/tests/cxx/iwyu_stricter_than_cpp-i4.h
+++ b/tests/cxx/iwyu_stricter_than_cpp-i4.h
@@ -1,0 +1,16 @@
+//===--- iwyu_stricter_than_cpp-i4.h - test input file for iwyu -----------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+class IndirectClass;
+
+struct IndirectStruct4 {
+  typedef IndirectClass IndirectClassNonProvidingTypedef;
+
+  using IndirectClassNonProvidingAl = IndirectClass;
+};
--- a/tests/cxx/iwyu_stricter_than_cpp.cc
+++ b/tests/cxx/iwyu_stricter_than_cpp.cc
@@ -36,6 +36,7 @@
 // This test tests that the iwyu requirement is correctly suppressed
 // when these two conditions are met, and not otherwise.
 
+#include "tests/cxx/direct.h"
 #include "tests/cxx/iwyu_stricter_than_cpp-typedefs.h"
 #include "tests/cxx/iwyu_stricter_than_cpp-type_alias.h"
 #include "tests/cxx/iwyu_stricter_than_cpp-autocast.h"
@@ -44,6 +45,7 @@
 // is visible in the translation unit (but not by -d2.h)
 #include "tests/cxx/iwyu_stricter_than_cpp-autocast2.h"
 #include "tests/cxx/iwyu_stricter_than_cpp-d2.h"
+#include "tests/cxx/iwyu_stricter_than_cpp-d3.h"
 
 typedef DoesEverythingRight DoubleTypedef;
 
@@ -78,6 +80,16 @@ void TestTypedefs() {
   // ...at least until we dereference the pointer
   // IWYU: IndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
   (void) dor_ptr->a;
+
+  // Nested name testing
+  IndirectStruct3ProvidingTypedef::IndirectClassProvidingTypedef pp;
+  // IWYU: IndirectClass is...*indirect.h
+  IndirectStruct4ProvidingTypedef::IndirectClassNonProvidingTypedef pn;
+  // IWYU: IndirectStruct3 is...*iwyu_stricter_than_cpp-i3.h
+  IndirectStruct3NonProvidingTypedef::IndirectClassProvidingTypedef np;
+  // IWYU: IndirectStruct4 is...*iwyu_stricter_than_cpp-i4.h
+  // IWYU: IndirectClass is...*indirect.h
+  IndirectStruct4NonProvidingTypedef::IndirectClassNonProvidingTypedef nn;
 
   // TODO(csilvers): test template types where we need some (but not
   // all) of the template args as well.
@@ -114,6 +126,16 @@ void TestTypeAliases() {
   // ...at least until we dereference the pointer
   // IWYU: IndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
   (void) dor_ptr->a;
+
+  // Nested name testing
+  IndirectStruct3ProvidingAl::IndirectClassProvidingAl pp;
+  // IWYU: IndirectClass is...*indirect.h
+  IndirectStruct4ProvidingAl::IndirectClassNonProvidingAl pn;
+  // IWYU: IndirectStruct3 is...*iwyu_stricter_than_cpp-i3.h
+  IndirectStruct3NonProvidingAl::IndirectClassProvidingAl np;
+  // IWYU: IndirectStruct4 is...*iwyu_stricter_than_cpp-i4.h
+  // IWYU: IndirectClass is...*indirect.h
+  IndirectStruct4NonProvidingAl::IndirectClassNonProvidingAl nn;
 }
 
 void TestAutocast() {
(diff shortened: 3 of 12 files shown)


kimgr:
Cool, thanks!

I feel like we're missing a little granularity here now that nested types are no longer reported. But it's nice that we won't automatically report twice for every nested type (nested + parent).

What happens in the following case:

```
#include "somethingelse_that_includes_vector"

// contrived, but makes the point
bool next_equals(std::vector<int>::const_iterator i, int v) {
   return (*(++i) == v);
}
```

Will we get a diagnostic only for `vector`, or only for `vector::const_iterator`, or something else?

kimgr:
Oops, accidental close

tests/cxx/iwyu_stricter_than_cpp.cc
```diff
@@ -61,6 +62,15 @@ void TestTypedefs() {
   // IWYU: IndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
   DoubleTypedef dt(6);
 
+  // "indirect.h" reporting is not needed for IndirectStruct1 typedef,
```
kimgr, line 65: I've chased around for quite a while trying to unravel the typedef chain here :-)

Is it possible to expand in this comment on which typedef maps to which type and _why_ reporting is needed/expected, without it turning into a mega-comment?

I think the one that stands out to me is `DoesNotForwardDeclareAndIncludes::Type ic2;` -- the general rule is that if the file declaring typedef included the type, it takes responsibility so the user of the typedef doesn't need to. Why is this nested type a special case?

kimgr:
> Avoid class member reporting
>
> Don't redundantly report all class/struct/union member uses,
> not only methods

There's a double-negative there in the commit body that makes it hard to understand what has changed. I think it should be something like:
```
Avoid double reporting of all class members

In addition to C++ methods, silence reports of all declarations inside a
class (data members, types, etc) if the parent type has already been
reported.
```

kimgr:
> Obsolete hindering code removed
>
> typedefs may or may not provide full aliased type info;
>
> this change fixes a bug with absense of reporting a type under a nested
> typedef, enclosing class of which is referred to with another typedef
> (test case added)

Would be nice if the commit subject said more about the essence of the change. To me this looks like an improvement that moves suppression from analysis (`CanIgnore...`) into reporting (that already happens to handle this), in a way that allows full analysis even for nested typedefs. But I could be thinking about it wrong :-)

So maybe something like:
```
Perform full analysis of typedef components

...
```

But let's make sure we don't screw up the author-intent policy (https://github.com/include-what-you-use/include-what-you-use/pull/1057#discussion_r883557446) with that change before spending energy on the commit message.

tests/cxx/iwyu_stricter_than_cpp.cc
bolshakov-a: > Is it possible to expand in this comment on which typedef maps to which type and why reporting is needed/expected, without it turning into a mega-comment?

I can try to do it. It would make sense.
> Why is this nested type a special case?

No, it isn't a special case. `DoesNotForwardDeclareAndIncludes` typedef still provides its underlying type info (`DirectStruct2`), but `DirectStruct2` itself (more precisely, `DirectStruct2::Type`) doesn't provide `IndirectClass` type info.

bolshakov-a:
> What happens in the following case:
> 
> ```
> #include "somethingelse_that_includes_vector"
> 
> // contrived, but makes the point
> bool next_equals(std::vector<int>::const_iterator i, int v) {
>    return (*(++i) == v);
> }
> ```
> 
> Will we get a diagnostic only for `vector`, or only for `vector::const_iterator`, or something else?

With Visual Studio, IWYU reports `std::vector` for the line with the function signature and `std::_Vector_const_iterator` for the next line (with the expression with `i`):
```
../../../../Project4/Source.cpp:3:23: warning: std::vector is defined in <vector>, which isn't directly #included.
../../../../Project4/Source.cpp:4:11: warning: std::_Vector_const_iterator is defined in <vector>, which isn't directly #included.

../../../../Project4/Source.cpp should add these lines:
#include <vector>  // for _Vector_const_iterator, vector

../../../../Project4/Source.cpp should remove these lines:
- #include "indirect-vector.h"  // lines 1-1

The full include-list for ../../../../Project4/Source.cpp:
#include <vector>  // for _Vector_const_iterator, vector
---
```
With libstdc++, IWYU reports only `std::vector` (most probably, because iterator type mapping is hard-coded [here](https://github.com/include-what-you-use/include-what-you-use/blob/0.18/iwyu.cc#L1151)).

bolshakov-a:
> > Avoid class member reporting
> > Don't redundantly report all class/struct/union member uses,
> > not only methods
> 
> There's a double-negative there in the commit body that makes it hard to understand what has changed. I think it should be something like:
> 
> ```
> Avoid double reporting of all class members
> 
> In addition to C++ methods, silence reports of all declarations inside a
> class (data members, types, etc) if the parent type has already been
> reported.
> ```

It looks better, thanks!

tests/cxx/iwyu_stricter_than_cpp.cc
kimgr: > No, it isn't a special case. DoesNotForwardDeclareAndIncludes typedef still provides its underlying type info 
> (DirectStruct2), but DirectStruct2 itself (more precisely, DirectStruct2::Type) doesn't provide
>  IndirectClass type info.

Ah, I didn't get that it mapped to `DirectStruct2`. That file has a forward-decl of `IndirectClass` that pushes responsibility to the use location, so it behaves as expected.

But the typedef name there is confusing everything: `DoesNotForwardDeclareAndIncludes` -- it _does_ forward-declare this new used type.

Is there another `Direct` type you could put the forward-declared use of `IndirectClass`, or does that shake everything up?

bolshakov-a: Maybe, it's really worth to introduce separate structs and typedefs, not trying to reuse types from other test cases.

kimgr:
Nice! Much easier to understand those tests. I'll take another review pass tomorrow, but this mostly looks really good. Thank you.

tests/cxx/iwyu_stricter_than_cpp.cc
```diff
@@ -113,7 +125,17 @@ void TestTypeAliases() {
   TplDoesEverythingRightAgainAl* tdora_ptr = 0;
   // ...at least until we dereference the pointer
   // IWYU: IndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
-  (void) dor_ptr->a;
+  (void)dor_ptr->a;
+
+  // Nested name testing
```
kimgr, line 130: Very nice. I had several drafts for name change suggestions, but this segment really drives home the full matrix of outer/nested type providing or not. This looks right.

tests/cxx/iwyu_stricter_than_cpp-i4.h
```diff
@@ -0,0 +1,16 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+class IndirectClass;
+
+struct IndirectStruct4 {
+  typedef IndirectClass IndirectClassNonprovidingTypedef;
```
kimgr, line 13: Could you use `NonProviding` instead of `Nonproviding` throughout?

kimgr (commented):
Nice! It took me quite some effort to unravel through all the typedefs and name aliasing, but I've convinced myself that this is a great improvement.

I had some trouble following the commit message, so I'd like to suggest something like:
```
Perform full analysis of typedef components

We used to ignore all analysis of AST nodes inside a `typedef`, but that
caused us to miss author-intent analysis of nested typedefs.

Remove the special casing for member-of-typedef now that reporting as a
whole is less granular (we now report only the parent type instead of all
nested components).

Add test cases to demonstrate that nested typedefs observe the author-intent
rules.
```

Some nits inlines, but I think this is mostly ready to go.

tests/cxx/iwyu_stricter_than_cpp-d3.h
```diff
@@ -0,0 +1,17 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/iwyu_stricter_than_cpp-i3.h"
+#include "tests/cxx/iwyu_stricter_than_cpp-i4.h"
+
+typedef IndirectStruct3 IndirectStruct3ProvidingTypedef;
```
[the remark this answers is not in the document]
bolshakov-a: >OK, I'm going to pick some more on naming. Please have patience

You are welcome!
> ... but I think the new names introduced here should be `Direct...` rather than `Indirect...`.

Do you mean typedef names? But they contain `Indirect...` just because they refer to `IndirectStruct*`.

tests/cxx/iwyu_stricter_than_cpp-i4.h
bolshakov-a: Of course, if you insist. But [here](https://github.com/include-what-you-use/include-what-you-use/blob/0.18/iwyu_ast_util.h#L619) `Nonclass` spelling is used.

kimgr: Yeah, I was hoping we could clean that up as part of renaming to `Tag` :-)

tests/cxx/iwyu_stricter_than_cpp-d3.h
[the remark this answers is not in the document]
kimgr: Yeah, it's not obvious how to best make it work. Hmm.

I think `DirectProvidingTypedefForIndirectStruct3` captures everything I'd want in the name, but it's extremely long.

I'm not overly enthusiastic about the `DoesEverythingRight` typedefs/aliases, but they do communicate their purpose with very little detail.

I've thought about this for about an hour off and on, and I can't think of anything better. Let's go with what we have.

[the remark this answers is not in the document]
kimgr: > Let's go with what we have.

I mean: Let's go with what's in the patch.

kimgr:
Yep, looks good to me! Merging.

PR #1058, opened 2022-05-26
Minor improvements to man page formatting

aaronpuchert:
In [=N] only N is a non-terminal, the = is a terminal and [] is neither.
When referring to another man page, only the name should be bold, not
the number in parentheses.

For referring to the example we don't need special fonts. The order and
naming of sections in a man page is standardized and examples are always
under this heading towards the end.


kimgr:
Excellent, thanks! I don't know my way around the markup language for `man` and was purely googling and imitating when I wrote that.

Could I also ask you to take a look at the manpage changes for https://github.com/include-what-you-use/include-what-you-use/pull/1052? A few new switches made their way in there. Cheers.

Issue #1070, opened 2022-06-22
IWYU seems to treat headers included in quotes and in brackets differently


storypku:
## How to Reproduce:
Here is my `file_size.cc` adapted from https://www.boost.org/doc/libs/1_78_0/libs/filesystem/example/tut1.cpp . 

```
#include <iostream>

#include "boost/filesystem.hpp"

using boost::filesystem::file_size;

int main(int argc, char* argv[]) {
  if (argc < 2) {

    std::cout << "Usage: tut1 path\n";
    return 1;
  }
  std::cout << argv[1] << " " << file_size(argv[1]) << '\n';
  return 0;
}
```

On a Ubuntu 18.04 system with libboost-all-dev installed, running include-what-you-use with `Xiwyu --mapping_file=boost-all.imp` gives the following output:

```
experimental/story/toolchain/file_size.cc should add these lines:
#include <boost/filesystem.hpp>  // for file_size

experimental/storypku/toolchain/file_size.cc should remove these lines:
- #include "boost/filesystem.hpp"  // lines 3-3

The full include-list for experimental/storypku/toolchain/file_size.cc:
#include <boost/filesystem.hpp>  // for file_size
#include <iostream>              // for operator<<, basic_ostream, cout, ostream, basic_ostream<>::__ostream_type, char_traits
---
```

I know that I cound workaround this issue by using bracket includes. However, on the IWYU side, can something be done to loosen this restriction ?

storypku:
Hi @kimgr,
 I found the following thread https://groups.google.com/g/include-what-you-use/c/ITG9mhL-Mt0 given by you regarding to whether to use quotes or brackets for header includes:

```
IWYU uses a simple heuristic to determine "" vs <> -- whether -I or
-isystem was used in the header search path.
For all headers found via -I it suggests "", and for -isystem it suggests <>.
So you can coerce it into using <> by adding your project-internal
include paths not as -I but -isystem.
```
In my opinion, IWYU can adhere to the decision on when to use brackets or quotes. However, when a header with a diffrent style exists, can IWYU just keep it as it is? Won't it be better to leave the choice of whether to use quotes or brackets to code base owners?

PR #1071, opened 2022-06-23
Fixes iwyu issue #1070

storypku:
This PR attempts to fix iwyu issue #1070.   
It should be noted that rewriting IWYU diff output is not the most efficient and elegant solution.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -2004,14 +2004,39 @@ size_t PrintableDiffs(const string& filename,
   // First, check if there are no adds or deletes.  If so, we print a
   // shorter summary line.
   bool no_adds_or_deletes = true;
+
+  set<string> to_add_headers;
+  set<string> to_rm_headers;
+
   for (const auto& key_line : sorted_lines) {
     const OneIncludeOrForwardDeclareLine* line = key_line.second;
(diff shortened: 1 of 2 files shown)


kimgr:
It does feel like the wrong place to address this, but thanks for coming up with a practical fix!

I wonder what it would take to iron this out sooner in the processing. We've discussed over the years to let the earlier stages of IWYU's analysis work on `FileEntry*` rather than strings to get rid of the quoted identity crisis. But I seem to recall that was much harder than expected at first glance.

On the one hand I would like reject this patch, but on the other hand, it would be nice to get rid of all the quoted/angled confusion even if the fix is in the wrong place. So maybe we should take this and mark it with a big `TODO` for future improvements. Not sure. Thoughts?

storypku:
Hi  @kimgr,
   Thank you for your reply. 

These day, I gave second thought on this PR, and I agree with you that it wasn't conceived throughly for the whole IWYU project. It should serve as a quick & dirty fix for my use case only. (It has bugs, too).

I am closing this PR now. Hope one day  I could spare some time to look into this issue. 

Thanks again for your ideas and IWYU. It really helps in my everyday work.

PR #1072, opened 2022-06-26
Fix autocast to reference

bolshakov-a:
CastExpr for constructor conversion is absent in the AST in such a case.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2002,19 +2002,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
       // Need the full to-type so we can call its constructor.
       case clang::CK_ConstructorConversion:
-        // 'Autocast' -- calling a one-arg, non-explicit constructor
-        // -- is a special case when it's done for a function call.
-        // iwyu requires the function-writer to provide the #include
-        // for the casted-to type, just so we don't have to require it
-        // here.  *However*, the function-author can override this
-        // iwyu requirement, in which case we're responsible for the
-        // casted-to type.  See IwyuBaseASTVisitor::VisitFunctionDecl.
-        if (!current_ast_node()->template HasAncestorOfType<CallExpr>() ||
-            ContainsKey(
-                GetCallerResponsibleTypesForAutocast(current_ast_node()),
-                RemovePointersAndReferences(to_type))) {
+        // 'Autocast' in function calls is handled in VisitCXXConstructExpr.
+        if (!current_ast_node()->template HasAncestorOfType<CallExpr>())
           required_full_types.push_back(to_type);
-        }
         break;
       // Need the full from-type so we can call its 'operator <totype>()'.
       case clang::CK_UserDefinedConversion:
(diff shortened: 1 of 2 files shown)


kimgr:
Nice, thanks!

I'm not sure about the approach -- the way I understand it, `Traverse` overrides are only necessary to customize the actual AST _traversal_ (e.g. visiting more nodes than the base RAV does, or skipping over nodes that it normally visits), which shouldn't be the case here.

The AST from the test looks like this:
```
tests/cxx/iwyu_stricter_than_cpp-d2.h:17:34: (2) [ CompoundStmt ] CompoundStmt 0x5572d77b9a28
|-ExprWithCleanups 0x5572d77b9860 'void'
| `-CallExpr 0x5572d77b94c0 'void'
|   |-ImplicitCastExpr 0x5572d77b94a8 'void (*)(struct IndirectStruct2)' <FunctionToPointerDecay>
|   | `-DeclRefExpr 0x5572d77b9458 'void (struct IndirectStruct2)' lvalue Function 0x5572d77b7710 'TwiceDeclaredFunction' 'void (struct IndirectStruct2)'
|   `-CXXConstructExpr 0x5572d77b9830 'struct IndirectStruct2' 'void (struct IndirectStruct2 &&) noexcept' elidable
|     `-MaterializeTemporaryExpr 0x5572d77b9708 'struct IndirectStruct2' xvalue
|       `-ImplicitCastExpr 0x5572d77b95e0 'struct IndirectStruct2' <ConstructorConversion>
|         `-CXXConstructExpr 0x5572d77b95b0 'struct IndirectStruct2' 'void (int)'
|           `-IntegerLiteral 0x5572d77b9118 'int' 1
`-ExprWithCleanups 0x5572d77b99e0 'void'
  `-CallExpr 0x5572d77b9940 'void'
    |-ImplicitCastExpr 0x5572d77b9928 'void (*)(const struct IndirectStruct2 &)' <FunctionToPointerDecay>
    | `-DeclRefExpr 0x5572d77b98e0 'void (const struct IndirectStruct2 &)' lvalue Function 0x5572d77b7858 'TwiceDeclaredRefFunction' 'void (const struct IndirectStruct2 &)'
    `-MaterializeTemporaryExpr 0x5572d77b99c8 'const struct IndirectStruct2' lvalue
      `-CXXConstructExpr 0x5572d77b9998 'const struct IndirectStruct2' 'void (int)'
        `-IntegerLiteral 0x5572d77b98c0 'int' 1 
```

so it seems to me there are two tree shapes we need to take into account:

* For `TwiceDeclaredFunction`: `CallExpr` / `CXXConstructExpr` / `MaterializeTemporaryExpr` / `ImplicitCastExpr(ConstructorConversion)` / `CXXConstructExpr` (handled today in `VisitCastExpr`)
* For `TwiceDeclaredRefFunction`: `CallExpr` / `MaterializeTemporaryExpr` / `CXXConstructExpr`

I wonder if it would make sense to work on all this in `VisitCXXConstructExpr` instead, and look back up the tree for both shapes? I'm not saying that's the way to go, I'm just thinking out loud. I don't have a good idea for how to look for tree shapes other than by looping up the `ASTNode` tree doing `IsA<NodeType>`, but it seems it would be nice to be able to express things like that as a single expression.

bolshakov-a:
Yes, I was in doubt whether `Traverse` should be used or `Visit` along with `isa<CXXConstructExpr>`. I can rewrite into the second variant.
I'm not sure whether checking actual tree structure makes sense. It seems like handling `CXXConstructExpr` inside `CallExpr` covers all the (test) cases. I thought that if there really are some cases when exact AST structure should be examined, such cases could be handled in a separate PR in the future.

bolshakov-a:
> Just note that there are two CXXConstructExprs in the subtree for the expr with the implicit conversion, and we probably don't want to signal both.

Yes, thanks, I see now that it's really a double reporting. But I wonder why it should be concerned? What consequences it has?

kimgr:
On further thought, I think it will be discarded anyway, since the outer construct-expr does not match the implicit-conversion-constructor rule. I think.

bolshakov-a:
Sorry, I haven't understood. The suggested code doesn't have such a rule, just checking of exact node class instead, and that is `CXXConstructExpr` in both cases, not `CXXTemporaryObjectExpr`.

bolshakov-a:
Yes, but `GetCallerResponsibleTypesForAutocast` unwinds the stack at first to get `CallExpr` and works with it, hence calls for inner and outer `CXXConstructExpr` should be equivalent.

kimgr:
@bolshakov-a Ah, I see that now. Makes sense. Let's get this merged.

bolshakov-a:
So, no problem with double-reporting?

kimgr:
Oops, maybe I was too quick. I lost track of why I asked.

I did get some ideas for simplifying refactors, so maybe I'll try and tackle that at the same time.

PR #1074, opened 2022-07-05
Add mappings for SYS_xxx syscall macros

alejandro-colomar:
But maybe there's a simpler way to fix this without adding hundreds of mappings.

See: <https://github.com/include-what-you-use/include-what-you-use/pull/1063>

### DIFF
--- a/gcc.libc.imp
+++ b/gcc.libc.imp
@@ -91,7 +91,7 @@
   { include: [ "<bits/string3.h>", private, "<string.h>", public ] },
   { include: [ "<bits/stropts.h>", private, "<stropts.h>", public ] },
   { include: [ "<bits/sys_errlist.h>", private, "<stdio.h>", public ] },
-  { include: [ "<bits/syscall.h>", private, "<sys/syscall.h>", private ] },
+  { include: [ "<bits/syscall.h>", private, "<sys/syscall.h>", public ] },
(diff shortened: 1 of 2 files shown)


alejandro-colomar:
This branch is on top of the other branch (referenced in the first message in this PR), and that's why it shows more than 1 commit, but as soon as that branch gets merged, this will only be 1 commit (the last one).

alejandro-colomar:
@kimgr Do you really require llvm-15?  I compiled iwyu a few weeks ago with Debian Sid's llvm-14, and it worked well.

kimgr:
IWYU master follows llvm-project main, but the Debian packaging we use for CI seems to be temporarily broken for mainline versions. It usually sorts itself out after a little while.

alejandro-colomar:
Huh, now it's asking for clang 16 :p

Please ping me when you expect it to be stable again.

kimgr:
You can follow https://github.com/include-what-you-use/include-what-you-use/issues/1092 if you're curious. I'll try to come to some conclusion and report there.

kimgr:
Please rebase on latest master to get the test suite in order. Thanks!

alejandro-colomar:
Hi Kim!  I'll send another round of patches now that the test suite works again.  Please merge this one, as it depends on this one.

Cheers,
Alex

kimgr:
Looks good to me; thanks for the informative commit message, it makes the case perfectly.

PR #1078, opened 2022-07-10
Drop -save-temps command-line args

kimgr:
IWYU doesn't produce any outputs, so -save-temps does not have any useful
effect.

Before this patch, passing -save-temps did, however, generate multiple
compilation jobs which caused a fatal error:

  error: unable to handle compilation, expected exactly one compiler job

Filter the arguments out, similar to what Clang tools do by default.


kimgr:
This seems like an obvious improvement, so I'll merge without confirmation that it solves the user problem.

PR #1079, opened 2022-07-10
Avoid redundant enum type template args reporting

bolshakov-a:
Reporting opaque declarations or full use of non-fixed enums at location where template specialization type is explicitly written should be sufficient, except the case of nested enumerator referencing (handled through NestedNameSpecifier).

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1543,19 +1543,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     used_loc = GetCanonicalUseLocation(used_loc, target_decl);
     const FileEntry* used_in = GetFileEntry(used_loc);
 
-    // Report EnumName instead of EnumName::Item.
-    // It supports for removing EnumName forward- (opaque-) declarations
-    // from output.
-    if (const auto* enum_constant_decl =
-            dyn_cast<EnumConstantDecl>(target_decl)) {
-      const auto* enum_decl =
-          cast<EnumDecl>(enum_constant_decl->getDeclContext());
-
-      // for unnamed enums, enumerator name is still preferred
-      if (enum_decl->getIdentifier())
-        target_decl = enum_decl;
-    }
-
     preprocessor_info().FileInfoFor(used_in)->ReportFullSymbolUse(
         used_loc, target_decl, use_flags, comment);
 
@@ -1653,6 +1640,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
                              const char* comment = nullptr) {
     // TODO(csilvers): figure out if/when calling CanIgnoreType() is correct.
     if (!type)
+      return;
+
+    // Enum type uses can be ignored. Their size is known (either implicitly
+    // 'int' or from a mandatory transitive inclusion of a non-fixed enum full
+    // declaration, or explicitly using a C++ 11 enum base). Only if an enum
+    // type or its enumerators are explicitly mentioned will they be reported
+    // by IWYU from VisitTagType or VisitDeclRefExpr correspondingly.
+    if (type->getAs<EnumType>())
       return;
 
     // Map private types like __normal_iterator to their public counterpart.
@@ -2434,6 +2429,26 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  bool VisitDeclRefExpr(clang::DeclRefExpr* expr) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+    // Report EnumName instead of EnumName::Item.
+    // It supports for removing EnumName forward- (opaque-) declarations
+    // from output.
+    if (const auto* enum_constant_decl =
+            dyn_cast<EnumConstantDecl>(expr->getDecl())) {
+      const auto* enum_decl =
+          cast<EnumDecl>(enum_constant_decl->getDeclContext());
+
+      // For unnamed enums, enumerator name is still preferred.
+      if (enum_decl->getIdentifier())
+        ReportDeclUse(CurrentLoc(), enum_decl);
+      else
+        ReportDeclUse(CurrentLoc(), enum_constant_decl);
+    }
+    return true;
+  }
+
   // When we call (or potentially call) a function, do an IWYU check
   // via ReportDeclUse() to make sure the definition of the function
   // is properly #included.
@@ -2990,9 +3005,6 @@ class InstantiatedTemplateVisitor
     if (node->ParentIsA<NestedNameSpecifier>()) {
       return true;
     }
-
-    if (const auto* enum_type = type->getAs<EnumType>())
-      return !CanBeOpaqueDeclared(enum_type);
 
     // If we're inside a typedef, we don't need our full type info --
     // in this case we follow what the C++ language allows and let
@@ -3985,7 +3997,7 @@ class IwyuAstConsumer
     // once we've tracked the UsingDecl use.
     if (const UsingShadowDecl* found_decl = DynCastFrom(expr->getFoundDecl())) {
       ReportDeclUse(CurrentLoc(), found_decl);
-    } else {
+    } else if (!isa<EnumConstantDecl>(expr->getDecl())) {
       ReportDeclUse(CurrentLoc(), expr->getDecl());
     }
     return Base::VisitDeclRefExpr(expr);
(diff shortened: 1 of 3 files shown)


kimgr:
I added that `sizeof` test to badinc.cc because I was curious what would actually happen. It diagnoses for both `I2_Enum` and `I1_Enum`, which makes sense to me, as `I1_Enum` is not an opaque declaration.

I also added an opaque enum forward decl in another testcase and verified that `sizeof` that one was fine with the forward-declaration itself.

I think this works as intended, and is good to go.

bolshakov-a:
> It diagnoses for both I2_Enum and I1_Enum, which makes sense to me, as I1_Enum is not an opaque declaration.

In fact, even that is redundant. If an enum cannot be opaque-declared, its definition _should_ be included in function-declaring header. In both cases, there isn't any need in declaring or `#include`-ing enum on call site, as long as it or its enumerators aren't named explicitly. Perhaps, it should be fixed some day too.

kimgr:
That makes sense. But in this case, `h_templateclass2.static_out_of_line(I22))` is a template, so it doesn't know it's instantiated with `I1_Enum`. On the other hand, it should be enough that the instantiation requires `I1_Enum`, it might be redundant to need it for the call-site return value.

bolshakov-a:
I've updated the PR with avoiding enum reporting in other unwanted places.

kimgr:
I'm not so sure about the extended direction.
```
// No need even in opaque declaration if enumeration or enumerator isn't named
// explicitly (it already should be elsewhere in the scope). No matter, whether
// enum has fixed type or not, used in sizeof operator, pointer arithmetics
// or whatever.
auto p1 = &ie1 + sizeof(ie1);
auto p2 = &ie3 + sizeof(ie3);
``` 
What exactly do you mean by "scope" here? `ie1` could be a declaration of an enum that's transitively included from some header. The fact that the declaration is visible here doesn't mean much for IWYU analysis.

Also, I wonder if this is over-optimizing. It would be nice to be quite straight-forward and stupid in AST traversal, and then leave policy decisions till IWYU analysis (in `iwyu_output.cc`). I know we don't exactly maintain a clean separation there now, but I feel like this is discarding information too early to get more comfortable results.

bolshakov-a:
> What exactly do you mean by "scope" here? ie1 could be a declaration of an enum that's transitively included from some header. The fact that the declaration is visible here doesn't mean much for IWYU analysis.

I mean "visibility scope" with all included header contents, indeed. I understand that it seems to be IWYU principle violation, but in fact, if you refer to global variable of enum type declared in some header, or call some function from some header which returns enum, you _should_ get underlying type info from that header transitively, so you can store enum value in variable, get its size, numeric value, transmit it somewhere. But if you explicitly refer to enumeration or enumerator name, IWYU suggests you to include the enum header or to add the opaque declaration, no problem. If you decide to refactor the variable or function declaration, e.g., by changing enum type, files including these declarations shouldn't be changed. (IWYU already has such behavior for function return types when function declaring header provides the header for return type info.) And if you, e.g. will decide to change enum header name, you should only change `#include` in the variable or function declaration header. There is no point in traversing through all the files that use the function or variable and changing enum header name there. So, IWYU principle goals occur to be achieved.

iwyu.cc
```diff
@@ -1654,6 +1641,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (!type)
       return;
 
+    if (isa<EnumType>(type))
```
kimgr, line 1644: You could add a comment here to rationalize why `enum` types are special, particularly since the function comment says "regardless of the type being explicitly written in the source code or not" (line 1647/1634).

I'd suggest something like:
```
    // Enum type uses can be ignored. Their size is known (either implicitly
    // 'int' or explicitly using a C++ 11 enum base). Only if an enum type or
    // its enumerators are explicitly mentioned (in a DeclRefExpr) will they be
    // required by IWYU.
```

Not 100% sure I captured that correctly, you can probably be more precise.

iwyu.cc
```diff
@@ -2434,6 +2424,26 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      return true;
+    // Report EnumName instead of EnumName::Item.
+    // It supports for removing EnumName forward- (opaque-) declarations
+    // from output.
+    if (const auto* enum_constant_decl =
+            dyn_cast<EnumConstantDecl>(expr->getDecl())) {
+      const auto* enum_decl =
+          cast<EnumDecl>(enum_constant_decl->getDeclContext());
+
+      // for unnamed enums, enumerator name is still preferred
```
kimgr, line 2438: Please use a full sentence including leading upper-case character and trailing full stop.

iwyu.cc
```diff
@@ -2434,6 +2424,26 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      if (enum_decl->getIdentifier())
+        ReportDeclUse(CurrentLoc(), enum_decl);
+      else
+        ReportDeclUse(CurrentLoc(), enum_constant_decl);
+    }
+    return Base::VisitDeclRefExpr(expr);
```
kimgr, line 2444: I just realized that these `Base::` calls are relatively recent mistakes (of mine). They are not necessary in `IwyuBaseAstVisitor`. See https://github.com/include-what-you-use/include-what-you-use/pull/1086.

tests/cxx/enums.cc
```diff
@@ -42,6 +42,13 @@ Struct1::IndirectEnum3 ie3;
 // IWYU: UnnamedEnumItem2 is...*enums-i4.h
 auto ie4 = UnnamedEnumItem2;
 
+// No need even in opaque declaration if enumeration or enumerator isn't named
```
kimgr, line 45: `No need even in` -> `No need even for`

tests/cxx/enums.cc
```diff
@@ -42,6 +42,13 @@ Struct1::IndirectEnum3 ie3;
...
+// explicitly (it already should be elsewhere in the scope). No matter, whether
```
kimgr, line 46: "Scope" is a slightly dangerous word, because it has such a precise definition in C and C++. I wonder if we could phrase this differently. Something like `in this case it is already declared earlier in the translation unit`?

tests/cxx/enums.cc
```diff
@@ -42,6 +42,13 @@ Struct1::IndirectEnum3 ie3;
...
+// enum has fixed type or not, used in sizeof operator, pointer arithmetics
```
kimgr, line 47: arithmetics -> arithmetic (it's uncountable in English)

kimgr (commented):
Some minor things inline, I get the gist of the change now, and agree that it makes sense.

Please avoid using "scope" in commit message as well.

Thanks!

tests/cxx/enums.cc
bolshakov-a: Thank you! In Russian too. Just a wrong word formation after "mathematics", "physics" etc.

bolshakov-a:
It seems like releasing the new LLVM version has broken the build somehow...

iwyu.cc
kimgr: Thanks!

kimgr:
@bolshakov-a I think the problem is the llvm/clang mainline packages have not been forwarded to llvm-16 yet, so the Debian package state is somewhat in flux. Let's give it a few days to sort itself out. (Unfortunately there's an IWYU-breaking change on LLVM mainline, so it will likely break in a different way, but one thing at a time).

kimgr:
Hey, sorry about the delay. That was a thorny issue. Please rebase on latest master and your tests should run green again.

bolshakov-a:
> That was a thorny issue. Please rebase on latest master and your tests should run green again.

Thanks a lot!

iwyu.cc
```diff
@@ -1655,6 +1642,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (!type)
       return;
 
+    // Enum type uses can be ignored. Their size is known (either implicitly
+    // 'int' or from a mandatory transitive inclusion of a non-fixed enum full
+    // decalration, or explicitly using a C++ 11 enum base). Only if an enum
```
kimgr, line 1647: oops, typo: decalration -> declaration

kimgr (commented):
Just a typo, otherwise LGTM

iwyu.cc
bolshakov-a: Thanks for attentiveness!

PR #1083, opened 2022-07-16
Avoid reporting type for "autocast" for function definition

bolshakov-a:
A function may just transmit passed-by-reference parameter to somewhere.
Requirement to explicitly write forward declaration in the same file (.cpp-file) to avoid `#include` suggestion is impractical when that type is already fwd-declared in the corresponding header.

I'm not sure about this change. Autocast to parameter of defined function still makes sense when the function definition is in header:
```cpp
inline void ProcessString(const StringWrapper& sw)
{
  stringProcessor.process(sw); // passed by reference
}
```
Maybe, to check whether the function is inline? Or there may be some better heuristics?

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1484,6 +1484,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
               GetFileEntry(call_expr), GetFileEntry(*fn_redecl))) {
         continue;
       }
+      if (fn_redecl->isThisDeclarationADefinition() && !IsInHeader(*fn_redecl))
+        continue;
       for (set<const Type*>::iterator it = retval.begin();
            it != retval.end(); ) {
         if (!CodeAuthorWantsJustAForwardDeclare(*it, GetLocation(*fn_redecl))) {
@@ -1767,6 +1769,11 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         FunctionDecl* redecl = decl;
         while ((redecl = redecl->getPreviousDecl()))
           ReportDeclUse(CurrentLoc(), redecl);
+      }
+      if (!IsInHeader(decl)) {
+        // No point in author-intent analysis of function definitions
+        // in source files.
+        return true;
       }
     } else {
       // Make all our types forward-declarable...
--- a/tests/cxx/implicit_ctor-d1.h
+++ b/tests/cxx/implicit_ctor-d1.h
@@ -22,6 +22,14 @@ int ImplicitCtorFn(IndirectWithImplicitCtor);
 // IWYU: IndirectWithImplicitCtor is...*implicit_ctor-i2.h.*for autocast
 int ImplicitCtorRefFn(const IndirectWithImplicitCtor&);
 
+// Reporting types for "autocast" for header-defined functions still makes sense
+// as opposed to function definitions in source files.
+// IWYU: IndirectWithImplicitCtor needs a declaration
+// IWYU: IndirectWithImplicitCtor is...*implicit_ctor-i2.h.*for autocast
+inline int InlineImplicitCtorRefFn(const IndirectWithImplicitCtor&) {
+  return 1;
+}
+
 // Test parameter type uses that do not require special handling for "autocast".
 int NoAutocastFn(
     // A subtle c++ point: forward-declaring is ok for nonconst, because
--- a/tests/cxx/implicit_ctor.cc
+++ b/tests/cxx/implicit_ctor.cc
@@ -22,6 +22,18 @@
 
 #include "tests/cxx/implicit_ctor-d1.h"
 
+// No reporting types for "autocast" for .cpp-file-local functions...
+// IWYU: IndirectWithImplicitCtor needs a declaration
+static int LocalFn(const IndirectWithImplicitCtor&) {
+  return 1;
+}
+
+// ... or for .cpp-file-definitions of functions declared in a header.
+// IWYU: IndirectWithImplicitCtor needs a declaration
+int ImplicitCtorRefFn(const IndirectWithImplicitCtor&) {
+  return 2;
+}
+
 // We don't need IndirectWithImplicitCtor even though we must convert to it.
 int a = ImplicitCtorFn(1);
 int b = ImplicitCtorRefFn(2);
@@ -31,6 +43,13 @@ int b = ImplicitCtorRefFn(2);
 int c = ImplicitCtorFn(IndirectWithImplicitCtor(3));
 // IWYU: IndirectWithImplicitCtor is...*implicit_ctor-i2.h
 int d = ImplicitCtorRefFn(IndirectWithImplicitCtor(4));
+
+// LocalFn doesn't provide IndirectWithImplicitCtor type info.
+// IWYU: IndirectWithImplicitCtor is...*implicit_ctor-i2.h
+int e = LocalFn(5);
+// InlineImplicitCtorRefFn should provide parameter type info,
+// hence no reporting.
+int f = InlineImplicitCtorRefFn(6);
 
 // Make sure we are responsible for the conversion when it's not for a
 // function call.
@@ -48,7 +67,7 @@ tests/cxx/implicit_ctor.cc should add these lines:
 tests/cxx/implicit_ctor.cc should remove these lines:
 
 The full include-list for tests/cxx/implicit_ctor.cc:
-#include "tests/cxx/implicit_ctor-d1.h"  // for ImplicitCtorFn, ImplicitCtorRefFn
+#include "tests/cxx/implicit_ctor-d1.h"  // for ImplicitCtorFn, ImplicitCtorRefFn, InlineImplicitCtorRefFn
 #include "tests/cxx/implicit_ctor-i2.h"  // for IndirectWithImplicitCtor
 
 ***** IWYU_SUMMARY */
--- a/tests/cxx/implicit_ctor.cc
+++ b/tests/cxx/implicit_ctor.cc
@@ -10,13 +10,13 @@
 // IWYU_ARGS: -Xiwyu --check_also="tests/cxx/*-d1.h" -I .
 
 // If you define an API that accepts an argument of class type or
-// const reference to a class type with an implicit constructor, you
-// must provide the definition for the class.
+// const reference to a class type with an implicit constructor which takes
+// exactly one argument, you must provide the definition for the class.
 //
-// If you use an API and that use triggers an implicit constructor
+// If you use an API and that use triggers such an implicit constructor
 // call, you are *not* considered to have used that constructor,
 // relying on the previous rule to make the code complete definition
-// available.
+// available. This is so called "autocast".
 //
 // This tests that logic.
 
--- a/tests/cxx/iwyu_stricter_than_cpp.cc
+++ b/tests/cxx/iwyu_stricter_than_cpp.cc
@@ -139,23 +139,30 @@ void TestTypeAliases() {
 }
 
 void TestAutocast() {
-  // We need full type of IndirectStruct2 because the declarer of FnValues and
-  // FnRefs didn't.
+  // We need full type of IndirectStruct2 because the declarer of the following
+  // functions didn't.
   FnValues(1, 2, 3, 4,
            // IWYU: IndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
            5);
   FnRefs(1, 2, 3, 4,
          // IWYU: IndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
          5);
+  HeaderDefinedFnRefs(1, 2, 3, 4,
+                      // IWYU: IndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
+                      5);
 
-  // We need full type of TplIndirectStruct2 because the declarer of TplFnValues
-  // and TplFnRefs didn't.
+  // We need full type of TplIndirectStruct2 because the declarer
+  // of the following functions didn't.
   TplFnValues(6, 7, 8, 9,
               // IWYU: TplIndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
               10);
   TplFnRefs(6, 7, 8, 9,
             // IWYU: TplIndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
             10);
+  HeaderDefinedTplFnRefs(
+      6, 7, 8, 9,
+      // IWYU: TplIndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
+      10);
 }
 
 void TestFunctionReturn() {
@@ -227,7 +234,7 @@ tests/cxx/iwyu_stricter_than_cpp.cc should remove these lines:
 
 The full include-list for tests/cxx/iwyu_stricter_than_cpp.cc:
 #include "tests/cxx/indirect.h"  // for IndirectClass
-#include "tests/cxx/iwyu_stricter_than_cpp-autocast.h"  // for FnRefs, FnValues, TplFnRefs, TplFnValues
+#include "tests/cxx/iwyu_stricter_than_cpp-autocast.h"  // for FnRefs, FnValues, HeaderDefinedFnRefs, HeaderDefinedTplFnRefs, TplFnRefs, TplFnValues
 #include "tests/cxx/iwyu_stricter_than_cpp-d3.h"  // for IndirectStruct3ProvidingAl, IndirectStruct3ProvidingTypedef, IndirectStruct4ProvidingAl, IndirectStruct4ProvidingTypedef
 #include "tests/cxx/iwyu_stricter_than_cpp-fnreturn.h"  // for DoesEverythingRightFn, DoesNotForwardDeclareAndIncludesFn, DoesNotForwardDeclareFn, DoesNotForwardDeclareProperlyFn, IncludesFn, TplDoesEverythingRightAgainFn, TplDoesEverythingRightFn, TplDoesNotForwardDeclareAndIncludesFn, TplDoesNotForwardDeclareFn, TplDoesNotForwardDeclareProperlyFn, TplIncludesFn
 #include "tests/cxx/iwyu_stricter_than_cpp-i2.h"  // for IndirectStruct2, TplIndirectStruct2
--- a/iwyu_location_util.cc
+++ b/iwyu_location_util.cc
@@ -10,6 +10,7 @@
 #include "iwyu_location_util.h"
 
 #include "iwyu_ast_util.h"
+#include "iwyu_port.h"
 #include "clang/AST/Decl.h"
 #include "clang/AST/DeclBase.h"
 #include "clang/AST/DeclCXX.h"
@@ -28,6 +29,7 @@ using clang::CXXMethodDecl;
 using clang::CXXOperatorCallExpr;
 using clang::ClassTemplateSpecializationDecl;
 using clang::ConditionalOperator;
+using clang::FileEntry;
 using clang::FunctionDecl;
 using clang::MemberExpr;
 using clang::SourceLocation;
@@ -168,6 +170,12 @@ SourceLocation GetLocation(const clang::TemplateArgumentLoc* argloc) {
 
 bool IsInScratchSpace(SourceLocation loc) {
   return StartsWith(PrintableLoc(GetSpellingLoc(loc)), "<scratch space>");
+}
+
+bool IsInHeader(const clang::Decl* decl) {
+  const FileEntry* containing_file = GetFileEntry(decl);
+  CHECK_(containing_file);
+  return !GlobalSourceManager()->isMainFile(*containing_file);
 }
 
 }  // namespace include_what_you_use
--- a/iwyu_location_util.h
+++ b/iwyu_location_util.h
@@ -228,6 +228,9 @@ inline bool IsBeforeInSameFile(const T& a, const U& b) {
   return IsBeforeInTranslationUnit(a, b);
 }
 
+// Returns true if the given declaration is located in a header file.
+bool IsInHeader(const clang::Decl*);
+
 }  // namespace include_what_you_use
 
 #endif  // INCLUDE_WHAT_YOU_USE_IWYU_LOCATION_UTIL_H_
--- a/tests/cxx/iwyu_stricter_than_cpp-autocast.h
+++ b/tests/cxx/iwyu_stricter_than_cpp-autocast.h
@@ -47,6 +47,16 @@ void FnRefs(
     const DirectStruct3& dc1, const struct DirectStruct4& dc2,
     const IndirectStruct2& ic2);
 
+inline void HeaderDefinedFnRefs(
+    // IWYU: IndirectStruct1 needs a declaration
+    // IWYU: IndirectStruct1 is...*iwyu_stricter_than_cpp-i1.h.*for autocast
+    const IndirectStruct1& ic1,
+    // IWYU: IndirectStructForwardDeclaredInD1 is...*iwyu_stricter_than_cpp-i1.h.*for autocast
+    const struct IndirectStructForwardDeclaredInD1& icfdid1,
+    const DirectStruct5& dc1, const struct DirectStruct6& dc2,
+    const IndirectStruct2& ic2) {
+}
+
 // --- Now do it all again, with templates!
 
 template <typename T>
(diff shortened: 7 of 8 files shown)


kimgr:
Hmm, yeah. This is confusing.

I don't like the short-circuiting of analysis in `VisitFunctionDecl` (though I happen to have some unfinished major rewrites of `FunctionDecl` handling somewhere that could result in something similar).

Did the last autocast change trigger warnings on simple forwarding functions? In that case, I wonder if we should maybe revert it and look for some other solution.

bolshakov-a:
> Did the last autocast change trigger warnings on simple forwarding functions?

No, they are unrelated changes. This one is about function declaration site handling, and that one — about call site. And changed tests are different, as you may see.

bolshakov-a:
>And changed tests are different, as you may see.

The previous version of the test case in `badinc` just test that autocast possibility triggers `#include`-suggestion _in function definition_ on parameter _passed by reference_. I don't like it because I often forget to mark my one-arg constructors as `explicit` (which is surely bad) and I often write parameter-transmitting factory methods. But I can wait if you intend to refactor here.

kimgr:
Thanks, that makes sense!

Hmm. So I guess those two work in tandem --

The callsite check tries to determine if the caller or the callee is responsible for auto-cast types. But it can't suggest any edits for the callee, because that's probably in an unrelated header associated with another translation unit.

The declaration-side check, then, needs to pick up the slack and guess that a parameter might be subject to auto-cast from some callsite, and take ownership.

I guess the latter is bound to have insufficient information, but I wonder if we could use some other heuristic to help it make better decisions. Off the top of my head, this patch is too heavy-handed.

> But I can wait if you intend to refactor here.

I don't think my planned changes will have immediate effects, but they might make it possible to use more information in the decision-making process. I'll see if I can breathe life into that again.

bolshakov-a:
> I guess the latter is bound to have insufficient information

I think that in the case when a function is declared in header and defined in .cpp (as it usually is), autocast analysis should be performed only on header (it is meaningless for .cpp in that case). When function is local for .cpp-file, autocast possibility assurance seems also redundant (at least, for me).
The only doubtful case is a function _defined_ in header. Such functions should usually be marked as `inline` (not doing so is almost always an error; methods defined inside a class declaration are also "inline" in fact). So, I think that `inline` checking should work (hoping that clang provides such possibility). Or, maybe, to check whether the file is .h or .cpp directly?

bolshakov-a:
Ah, it seems like I realize now, that if declaration-site analysis is changed, call-site analysis should be changed as well...

bolshakov-a:
But it doesn't seem a big problem to check visibility of the definition from the call site. Or, maybe, just test the first visible declaration whether it also is the definition.
This PR should probably be reworked, indeed, but I'm trying to find the right direction...

kimgr:
> Ah, it seems like I realize now, that if declaration-site analysis is changed, call-site analysis should be changed as well...

Can you explain this a bit more? Call-site analysis seems simpler, unless, I guess, the call-site is in the same file as the definition.

I think I understand what you want to achieve, but I don't have a clear idea how to do that. The challenge is to guess whether the declaration (possibly definition) we're processing should really obey the autocast rules, or if there's another declaration visible that already does, right?

I wonder if it would be possible to enumerate the redeclarations, and see if there's any one of them that has an author-intent forward decl hint...?

bolshakov-a:
> Can you explain this a bit more?

The problem is that for such code:
```cpp
#include "S.h" // S has implicit ctor accepting int

static void fn(const S&) {
}

int main() {
  fn(5);
}
```
IWYU from this branch suggests to remove `#include "S.h"`. This is an error not present on master branch, so this PR needs work.
> The challenge is to guess whether the declaration (possibly definition) we're processing should really obey the autocast rules

I think that if a declaration isn't a definition, it should provide autocast possibility, no problem with that. Of course, there are cases like
```cpp
void fn1(const S&); // prototype

void fn2(const S& s) {
 fn1(s);
}

void fn1(const S& s) { // definition
 ...
}
```
when one probably don't want to `#include "S.h"`, but these cases seem to be rare.

kimgr:
Thanks for the added detail, those examples are very helpful.

My question was more in regards to your mention of 'call-site analysis should be changed as well' -- I don't really see that; this looks to me like a decl/defn-side problem entirely.

I actually think searching the redecls might be a nice solution. Something like:
```
if current node is a function decl and has a parameter with implicit conversion ctor:
    for redecl in fn_decl.redecls():
        if code author wants just a forward declare(redecl):
            # yipee!
```

Oh, hm. `GetCallerResponsibleTypesForAutocast` already does almost that: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L1470. But it requires that _all_ declarations have the author-intent signal, rather than just one of them.

So let's say from a call-expr perspective, it needs to be all, but from a decl/defn perspective it could be any single one. I think that would make sense.

bolshakov-a:
>My question was more in regards to your mention of 'call-site analysis should be changed as well' -- I don't really see that; this looks to me like a decl/defn-side problem entirely.

But if you change the one, you must change the other, in order to keep decl/defn-handling logic and call site handling logic complementary. Otherwise you will get a bug like aforementioned bug in this branch. In fact, one can say that there isn't any problem at all on master branch. Just one more reason to mark constructors as `explicit`. This PR suggests just slight analysis improvement.

> But it requires that all declarations have the author-intent signal, rather than just one of them.

It seems reasonable, because if any of included headers with function declaration includes the header which contains parameter type, there is no need to include it in the calling file.
I just concern of local for .cpp-file functions, which usually have only one declaration coinciding with definition. I don't intend to change behavior for other cases.

bolshakov-a:
> I just concern of local for .cpp-file functions, which usually have only one declaration coinciding with definition

Sorry, I was wrong. When a function is declared in header and defined in .cpp, IWYU suggests `#include` for parameter autocast in the .cpp-file-definition too. Forward declaration in the corresponding header doesn't help. I already wrote about it, but forgot.
My idea is that function declaration handling shouldn't be changed, only definition handling.
All of that doesn't seem to be a big problem, but it's better to fix it somehow in the future.

kimgr:
Ah, yes. The `GetCallerResponsibleTypesForAutocast` function is almost a double-negative, so I always get confused about what it's trying to decide. Your naming makes more sense there, I think :-)

For some reason I feel uneasy short-circuiting `VisitFunctionDecl` for definitions, but I can't really see any obvious reason why it would be wrong. I have a bunch of scratch branches for #105 lying around (#605 among others, but that's not currently my favorite), I should probably try to finish that work; it seems like the same policy holds there, i.e. the special rules only apply to declarations.

Until that day comes, let's try this and see what falls out.

One thing: I try not to add to `badinc.cc`, since it's too large and unmaintainable already. Could you extract the new example to a separate test? Something that shows the difference between declaration and definition? `iwyu_stricter_than_cpp.*` are sort of related, but also a little bit too big and interconnected. A new testcase is probably/maybe better.

kimgr:
Please rebase.

bolshakov-a:
> Could you extract the new example to a separate test?

I think it could be placed into `implicit_ctor` test, isn't it?

bolshakov-a:
> (https://github.com/include-what-you-use/include-what-you-use/pull/605 among others, but that's not currently my favorite)

#605 seems like a wrong link.

kimgr:
> > (#605 among others, but that's not currently my favorite)
> 
> #605 seems like a wrong link.

Yeah, I think I meant #675, sorry!

kimgr:
I think you might be able to use `IsHeaderFile` heuristic with `GetFilePath` to determine if the definition is inside a header. It's pretty hacky, but I think it should play out OK:
```
if (IsHeaderFile(GetFilePath(decl))) {
   // declaration in header
}
``` 

It would be nice with a little comment on that `return true;` whether you go for header vs. source file checking, to describe that author-intent analysis of functions only apply to declarations, not definitions.

kimgr:
Oh, and yeah, I think `implicit_ctor` might be a good place for the testcase. But not that the change also changes semantics for the return type.

bolshakov-a:
> But note that the change also changes semantics for the return type.

I don't think so. Just note that [here](https://github.com/include-what-you-use/include-what-you-use/blob/0.18/iwyu.cc#L1768) forward-declare context is set only if this isn't a definition. Otherwise (for definition) `in_forward_declare_context()` returns `false` by default, hence `type_use_reported_in_visit_function_type` [here](https://github.com/include-what-you-use/include-what-you-use/blob/0.18/iwyu.cc#L1786) equals to `true`, and [reporting](https://github.com/include-what-you-use/include-what-you-use/blob/0.18/iwyu.cc#L1791) doesn't occur, as if with my short-circuiting.
In fact, my primary intent was to unify behavior for parameters and return type. Otherwise, that difference becomes more clearly seen with [my changes](https://github.com/bolshakov-a/include-what-you-use/commit/ca8032bda92184c5f250a00857527722d5e35790#diff-7ca107c8b836c1765e8b735b74f06833bec48f308a76017886cee1d5736002d9R2533).

kimgr:
Ah, I see. I missed the `in_forward_declare_context()` check.

bolshakov-a:
> I think you might be able to use `IsHeaderFile` heuristic with `GetFilePath` to determine if the definition is inside a header.

This is exactly what is needed, thank you!
I've updated the PR also with call site bug fix.

tests/cxx/implicit_ctor-d1.h
```diff
@@ -21,15 +22,44 @@ int ImplicitCtorFn(IndirectWithImplicitCtor);
 // IWYU: IndirectWithImplicitCtor is...*implicit_ctor-i2.h.*for autocast
 int ImplicitCtorRefFn(const IndirectWithImplicitCtor&);
 
+// Autocast assurance for header-defined functions still makes sense as opposed
+// to function definitions in source files.
+// IWYU: IndirectWithImplicitCtor needs a declaration
+// IWYU: IndirectWithImplicitCtor is...*implicit_ctor-i2.h.*for autocast
+inline int InlineImplicitCtorRefFn(const IndirectWithImplicitCtor&) {
```
bolshakov-a, line 29: Maybe, cover more test cases? E.g., when "autocast" assurance is inhibited by forward-declaring...

tests/cxx/implicit_ctor-d1.h
```diff
@@ -21,15 +22,44 @@ int ImplicitCtorFn(IndirectWithImplicitCtor);
...
+    // (implicit conversion involves creating a temporary, which
+    // doesn't bind to non-const references).
+    // IWYU: IndirectWithImplicitCtor needs a declaration
+    IndirectWithImplicitCtor& nonconst,
+    // Forward-declaring is ok because ptrref is a const reference
+    // to a *pointer*.
+    // IWYU: IndirectWithImplicitCtor needs a declaration
+    IndirectWithImplicitCtor* const& ptrref,
+    // IWYU: IndirectClass needs a declaration
+    IndirectClass);
```
bolshakov-a, line 49: Maybe, add "autocastable" parameter in order to test mixing it with non-"autocastable", like it was in `badinc`? Or, `iwyu_stricter_than_cpp` covers already that case?..

tests/cxx/implicit_ctor-d1.h
kimgr: I believe that's covered by `iwyu_stricter_than_cpp*`

bolshakov-a: But there is no function definitions in `iwyu_stricter_than_cpp-autocast.h`, only declarations...

iwyu.cc
```diff
@@ -1768,6 +1770,11 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         while ((redecl = redecl->getPreviousDecl()))
           ReportDeclUse(CurrentLoc(), redecl);
       }
+      if (!IsInHeader(decl)) {
```
kimgr, line 1773: I just ran into clang's `SourceManager::isMainFile(const FileEntry&)`, which I think might be an even better fit.

Unless I'm missing something, we only ever have one `.cpp` file in flight -- the main compilation unit. So anywhere `isMainFile` returns false, we're in some other file, and that must be a header (or some eccentric thing like included .cpp file).

I would generally prefer that to adding semantic weight to filenames, so please give it a shot and see if it works. There's a `GlobalSourceManager()` accessor available for the source manager.

tests/cxx/badinc.cc
```diff
@@ -307,27 +307,22 @@ template<> struct Cc_TemplateStruct<I1_Class, I12> { I1_Class* i; };
 // IWYU: I1_Enum is...*badinc-i1.h
 template<class T = I1_Class, I1_Enum E = I11> class Cc_DeclareOnlyTemplateClass;
 
-// I2_Class has a non-explicit constructor (actually, two), so we need
```
kimgr, line 310: This comment might actually help clarify what `No autocast assurance for function definitions` refers to. Consider reading the test after your change; there's no indication that `I2_Class` has implicit constructors and might trigger `autocast assurance`.

I suggest:
```
  // I2_Class has non-explicit constructors, so would normally trigger special
  // rules for autocast. But these rules do not apply to function definitions,
  // so uses of I2_Class below should behave just like I1_Class.
```

tests/cxx/badinc.cc
```diff
@@ -307,27 +307,22 @@ template<> struct Cc_TemplateStruct<I1_Class, I12> { I1_Class* i; };
...
-// the full type here even though it's a const reference.  I1_Class
-// has no implicit, one-argument constructor, so fwd declaring is ok.
 // IWYU: I2_Class needs a declaration
 const I2_Class& Cc_Function(
     // IWYU: I1_Class needs a declaration
     const I1_Class& i1,
-    // IWYU: I2_Class is...*badinc-i2.h.*for autocast
+    // No autocast assurance for function definitions.
```
kimgr, line 314: And actually, with the summary suggested above, maybe these `No autocast...` comments are superfluous?

I agree with the other comment cleanups.

tests/cxx/implicit_ctor-d1.h
```diff
@@ -21,15 +22,44 @@ int ImplicitCtorFn(IndirectWithImplicitCtor);
...
 
+// Autocast assurance for header-defined functions still makes sense as opposed
+// to function definitions in source files.
+// IWYU: IndirectWithImplicitCtor needs a declaration
+// IWYU: IndirectWithImplicitCtor is...*implicit_ctor-i2.h.*for autocast
+inline int InlineImplicitCtorRefFn(const IndirectWithImplicitCtor&) {
+  return 1;
+}
+
+// IndirectClass hasn't any implicit, one-argument constructor, so fwd declaring
```
kimgr, line 33: Let's not single out `IndirectClass` here:
```
// Test parameter type uses that do not require special handling for autocast.
```

tests/cxx/implicit_ctor-d1.h
kimgr: I'm not sure if that's explicitly covered, but I don't think it adds a lot of value. Possibly as a second function that just tests that interaction:
```
void AutocastAndNot(
    // IWYU: IndirectWithImplicitCtor needs a declaration
    // IWYU: IndirectWithImplicitCtor is...*implicit_ctor-i2.h.*for autocast
    const IndirectWIthImplicitCtor&,
    // IndirectClass does not have any implicit constructors.
    const IndirectClass&
);
```
But it feels like that's a testcase we can add if someone reports a problem :-)

tests/cxx/implicit_ctor-d1.h
```diff
@@ -21,15 +22,44 @@ int ImplicitCtorFn(IndirectWithImplicitCtor);
 // IWYU: IndirectWithImplicitCtor is...*implicit_ctor-i2.h.*for autocast
 int ImplicitCtorRefFn(const IndirectWithImplicitCtor&);
 
+// Autocast assurance for header-defined functions still makes sense as opposed
```
kimgr, line 25: I think instead of 'assurance', you may want to use 'requirement', here and throughout. 'Assurance' sounds a little wrong, somehow, but I'm not a native speaker, so I'm having trouble finding the perfect word :-)

tests/cxx/implicit_ctor.cc
```diff
@@ -10,18 +10,30 @@
...
 
 #include "tests/cxx/implicit_ctor-d1.h"
 
+// No 'autocast' assurance for .cpp-file-local functions...
+// IWYU: IndirectWithImplicitCtor needs a declaration
+static int LocalFn(const IndirectWithImplicitCtor&) {
+  return 1;
+}
+
+// ... and for .cpp-file-definitions of functions declared in a header.
```
kimgr, line 31: s/and/or/

tests/cxx/implicit_ctor-d1.h
kimgr: I see. I think you could add that example in `iwyu_stricter_than_cpp-autocast.h` -- the forward-decls are already available there, so it should just be a matter of adding a small inline function.

Ah, or copy the `Fn` there and make it an inline definition, so it covers all the same cases.

kimgr:
Nice! A bunch of new small comments, but I think this looks solid now.

tests/cxx/implicit_ctor-d1.h
bolshakov-a: I've added new cases into `iwyu_stricter_than_cpp`. But I'm not sure that they aren't just a redundant garbage, which multiplies possible subsequent test case changes by three...  Please take a look.

tests/cxx/implicit_ctor-d1.h
bolshakov-a: 'Providing'? 'Provision'? "Autocastability" provision requirement?

tests/cxx/implicit_ctor.cc
bolshakov-a: Sorry?

tests/cxx/implicit_ctor-d1.h
bolshakov-a: Maybe just "report types for 'autocast'". It seems to be more correct to say that 'autocastability' is provided by having an implicit constructor rather than by having the header included.

kimgr: "reporting types for autocast" sounds good to me.

tests/cxx/implicit_ctor.cc
kimgr: Oh, sorry! That's the sed (and vim, I believe) command sequence for search-and-replace:
https://www.gnu.org/software/sed/manual/html_node/The-_0022s_0022-Command.html

I really should have said "replace 'and' with 'or' here", but my fingers outran my empathetic brain :)

tests/cxx/badinc.cc
```diff
@@ -307,25 +307,20 @@ template<> struct Cc_TemplateStruct<I1_Class, I12> { I1_Class* i; };
 // IWYU: I1_Enum is...*badinc-i1.h
 template<class T = I1_Class, I1_Enum E = I11> class Cc_DeclareOnlyTemplateClass;
 
-// I2_Class has a non-explicit constructor (actually, two), so we need
-// the full type here even though it's a const reference.  I1_Class
-// has no implicit, one-argument constructor, so fwd declaring is ok.
+// I2_Class has non-explicit constructors, so would normally trigger special
```
kimgr, line 310: Do we even need to keep this construct in `badinc.cc` now that autocast is covered elsewhere?

kimgr: Not sure if the return-type is only covered here, but it seems unlikely that it wouldn't be tested elsewhere. Returning function static by reference might be worth keeping.

tests/cxx/implicit_ctor-d1.h
kimgr: It does get pretty complicated with all the different types. Is it really necessary to use distinct types for every parameter?

tests/cxx/implicit_ctor-d1.h
```diff
@@ -21,15 +22,44 @@ int ImplicitCtorFn(IndirectWithImplicitCtor);
...
+    // you can't do implicit conversion to a non-const reference
+    // (implicit conversion involves creating a temporary, which
+    // doesn't bind to non-const references).
+    // IWYU: IndirectWithImplicitCtor needs a declaration
+    IndirectWithImplicitCtor& nonconst,
+    // Forward-declaring is ok because ptrref is a const reference
+    // to a *pointer*.
+    // IWYU: IndirectWithImplicitCtor needs a declaration
+    IndirectWithImplicitCtor* const& ptrref,
+    // Test parameter type uses that do not require special handling for
```
kimgr, line 44: I meant to put this comment on the `NoAutocastFn` itself.

kimgr:
I think there might be a bit too much going on in this PR now -- it's hard to make progress because so much is changing at once.

Could we maybe split out the test cleanups/improvements into a separate PR? It would have been weird to start making test changes without purpose, but now that we agree the approach is workable, it's much easier to review test cleanups on the mainline separate from test changes required for new functionality.

I also noticed Clang mainline broke us again, so I'll try and fix that up. Looks easier than last time :crossed_fingers:

tests/cxx/implicit_ctor-d1.h
bolshakov-a: I think so, because various combinations of fwd-decl status and `#include` status should be tested. Or did you mean separate `DirectStruct*`s for every declaration? They must be different because requirement to `#include` them is only tested through the explanatory comment in "IWYU SUMMARY", because they are included directly. If it were just a single type, one could broke all reportings but one and still have the test passed.

tests/cxx/implicit_ctor-d1.h
bolshakov-a: Oh, I had already forgotten that the first two parameters don't require header inclusion for "autocast" too...

tests/cxx/implicit_ctor-d1.h
kimgr: Aha. That seems like a weakness in the test design; wouldn't it be better to use one type, but not have it directly included? So that there were line diagnostics? I'm not suggesting you should do it, I'm just curious if the test is really its best self.

bolshakov-a: No, the point is exactly in that the type definition is included directly. This is one of the two "type provision" rules (a type is considered as [should be] provided if it is not fwd-declared or is _directly_ included) which are tested in `iwyu_stricter_than_cpp`.

bolshakov-a:
Ok, opened #1108. Let's pause with this PR.

tests/cxx/badinc.cc
bolshakov-a: I've decided also to keep one passed-by-value parameter in order to test requirement of its type info in function definition. It's been made in #1108.

tests/cxx/implicit_ctor-d1.h
kimgr: Oh, right. I overlooked the context.

I'm almost inclined to reduce coverage to keep the tests less noisy here. I'll leave it to you to decide.

bolshakov-a:
@kimgr, should this PR be further divided? I think at least the first two commits should be merged together in order to keep call site - decl site consistence.

kimgr:
@bolshakov-a I agree, that sounds like a good idea. The additional test is related, so that's easier to review together with the change. Please squash the first two and do some commit message improvements there.

tests/cxx/iwyu_stricter_than_cpp.cc
```diff
@@ -139,23 +139,30 @@ void TestTypeAliases() {
...
+  // of the following functions didn't.
   TplFnValues(6, 7, 8, 9,
               // IWYU: TplIndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
               10);
   TplFnRefs(6, 7, 8, 9,
             // IWYU: TplIndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
             10);
+  HeaderDefinedTplFnRefs(
+      6, 7, 8, 9,
+      // IWYU: IndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
```
kimgr, line 164: Should be `TplIndirectStruct2`

kimgr:
Thanks! One remaining `Tpl` prefix missing due to the fuzzy diagnostic matching in IWYU's test tooling, but with that fixed, I think this is good to go.

tests/cxx/iwyu_stricter_than_cpp.cc
bolshakov-a: Oh, thanks!

kimgr:
Thank you for pulling through all that work!

bolshakov-a:
Thank you!

PR #1086, opened 2022-07-24
Remove fruitless Base::Visit calls

kimgr:
The IwyuBaseAstVisitor has a base class, BaseAstVisitor, but it only has a
small number of Visit functions implemented, for logging.

So there is no need to call Base::VisitX in general, only for the few with
an implementation that actually does something, and they are preserved
here.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1703,7 +1703,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (const clang::Type* type = integer_type.getTypePtrOrNull()) {
       ReportTypeUse(CurrentLoc(), type);
     }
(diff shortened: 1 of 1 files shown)


bolshakov-a:
What if some realization will be added to BaseAstVisitor in the future?

kimgr:
@bolshakov-a Yeah, it might happen, but the way I understand `BaseAstVisitor` it is very bare-bones and shouldn't contain any significant logic.

This was mostly for consistency -- all the places that called `Base::Visit...` were added later in IWYU's lifetime, after the original authors had jumped ship. So it looked like their original intent was _not_ to call it.

If we need some kind of base functionality in the future, I think we should consistently fix all `Visit` methods first.

PR #1089, opened 2022-07-26
Add another mapping for NULL

alejandro-colomar:
POSIX defines NULL in <unistd.h> too.

Signed-off-by: Alejandro Colomar <alx.manpages@gmail.com>


kimgr:
Thanks, looks good to me.

PR #1091, opened 2022-08-05
Remove llvm-15 packaging workarounds

kimgr:


kimgr:
This fails after my change, but for the right reasons -- upstream Clang has breaking AST changes that affect IWYU badly (see https://github.com/include-what-you-use/include-what-you-use/issues/1092 for more info). I am working on those separately, but I want to make sure that IWYU builds against latest LLVM.

alejandro-colomar:
Ahhh, this makes sense.

Issue #1092, opened 2022-08-06
Breaking change on Clang mainline


kimgr:
[The patch eventually committed as 15f3cd6bfc670b](https://github.com/llvm/llvm-project/commit/15f3cd6bfc670ba6106184a903eb04be059e5977) breaks IWYU in several different ways. I haven't isolated all of them yet, so I'll use this issue to keep a log of what's broken, and potential fixes.

Original review and some discussion around dealing with breakage: https://reviews.llvm.org/D112374.

This landed after the branch of llvm-15, so IWYU users can still build IWYU mainline with the llvm-14 or llvm-15 branches.

kimgr:
I just want to briefly describe what's happened here, and why it's taking so long to fix.

The Clang-side change modified the construction of the AST so that every `Type`-derived node in the tree is wrapped in an `ElaboratedType`. Elaborated type nodes were previously only inserted for namespace-qualified types (e.g. `ns::foo::Bar`) or elaborated type specifiers (e.g. `struct Foo` or `class Bar`).

The rationale is that an `ElaboratedType` with specifier `ETK_None` is a clearer indication that the type lacks an elaborated type specifier than just the absence of `ElaboratedType`, and type printing for Clang diagnostics has been improved significantly after this change (i.e. types are printed as they were written in source code, not accidentally elaborated or not).

Because `ElaboratedType` nodes were previously uncommon, we had ad-hoc handling of them where they might show up, and only some cases where we explicitly look at them to control use policy (e.g. the parameter in `void foo(class const Bar&);` counts as a forward-declaration in and of itself).

For individual nodes, this is not so much of a problem. But IWYU does much of its AST work looking at the _shape of subtrees_ in the AST, and the presence of a new `ElaboratedType` node breaks a lot of assumptions.

`ElaboratedType` like a few other AST node types, is colloquially called "sugar", and can mostly be ignored.

I'm making inventory to try and understand what the different cases are, and see if there's a systematic way to solve each one. Here's what I have so far:

* Getting a `Type*` from somewhere (e.g. an expression) and checking if it `isa<SomeType>` or can be `dyn_cast<SomeOtherType>` is no longer viable. It seems the safe way to look through sugar for cases like this is to do `auto x = type->getAs<SomeType>();`
[...]

PR #1097, opened 2022-08-28
[ci] Work around missing libclang-16.so.1

kimgr:


kimgr:
These two patches are required to get IWYU to even build with LLVM mainline. There are still failing testcases, but they are caused by #1092, and I'm starting work to push patches to fix that. Will merge these preliminary changes even though CI build says no.

kimgr:
Note: neither of these patches should go to the clang_15 branch once that is created.

PR #1098, opened 2022-08-28
Add some infrastructure for quick&dirty debug flags

kimgr:
The `printtypeclass` behavior was absolutely indispensable when troubleshooting #1092. But I don't feel comfortable adding that as default behavior, so add support for simple debug flags, and add a flag for toggling the extended printing.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -428,7 +428,12 @@ string PrintableType(const Type* type) {
   if (!type)
     return "<null type>";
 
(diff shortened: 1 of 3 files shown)


kimgr:
Intentionally not documented or tested, this is primarily for interactive one-offs. That said, I think the `printtypeclass` flag is useful for future endeavors too, so that will likely stay around.

kimgr:
Will merge this against a red tree, because it's necessary to fix the things that are breaking the build.

PR #1099, opened 2022-08-28
[clang compat] Use elaborated type specifier in CanForwardDeclareType

kimgr:
Preceding refactor to formalize terminology.


kimgr:
Will merge this with a failing build. It's the first set of fixes for #1092, and the test suite will not run green until they're all addressed.

PR #1100, opened 2022-08-28
[clang compat] Use getAs<> instead of dynamic cast for Types

kimgr:
Downcasting an AST Type* to a more specific type is increasingly unsafe in the
AST. In general any Type* picked out of the AST might contain so-called 'sugar',
i.e. visible or invisible structure that does not change the identity of the
underlying type, but adds information.

After Clang 15f3cd6bfc670ba6106184a903eb04be059e5977, almost every Type is
wrapped in an ElaboratedType with kind ETK_None (i.e. "no elaboration").

ElaboratedType counts as sugar, and thus needs to be stripped off before we do
dynamic type checking of Type nodes.

The recommended way to look through sugar downwards in the tree is
'type->getAs<MostDerivedType>()', which will recursively remove sugar until a
node of MostDerivedType is reached. If there is no MostDerivedType node it
returns nullptr, so it's safe to consider getAs a desugaring dynamic downcast.

There are still some places where we use DynCastFrom or isa<> on Types, I have
only modified the ones that had any effect on broken testcases. Planning to look
into the rest separately once the test suite is working.

This fixes 4 broken testcases, and is an improvement towards fixing #1092.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1421,9 +1421,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     set<const Type*> retval;
     const Type* underlying_type = decl->getUnderlyingType().getTypePtr();
     // If the underlying type is itself a typedef, we recurse.
-    if (const TypedefType* underlying_typedef = DynCastFrom(underlying_type)) {
-      if (const TypedefNameDecl* underlying_typedef_decl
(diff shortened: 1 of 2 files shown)


kimgr:
Merging this with broken tests; it's an increment towards getting the test suite green.

PR #1101, opened 2022-08-31
Desugaring refactors

kimgr:
This is a big one, to prepare for actual fixes for #1092.

https://github.com/llvm/llvm-project/commit/15f3cd6bfc670ba6106184a903eb04be059e5977 reshaped the AST quite significantly, by wrapping most `Type` nodes in an `ElaboratedType` node with `ETK_None`, to clearly indicate that the type does _not_ have an elaborated type specifier.

After careful analysis and help from the discussion on https://reviews.llvm.org/D112374, I have a much better understanding of type sugar in Clang and IWYU.

As far as Clang is concerned, all sugar is disposable, except for diagnostics. So `typedef int MyInt;` is functionally equivalent with `int`. So when Clang desugars a type, it sees through all the nodes that have no bearing on the type identity of the underlying type, to be able to do type checking on the desugared type. But in IWYU, we care deeply about the difference between typedefs/type aliases and underlying types, because they can live in different headers.

Thus:

* perform a type check for `SubstTemplateTypeParm` _before_ desugaring, as stronger desugaring would strip off that type
* institute a new `Desugar` function in IWYU, that desugars much like Clang, but stops at `TypedefType`, `UsingType` and `TemplateSpecializationType`, all of which are necessary for IWYU's more source-code oriented analysis.
* `sed` all uses of `RemoveElaboration` to `Desugar`, which is now a functional superset
* `sed` all uses of `RemoveSubstTemplateTypeParm` to `Desugar`, which is now a functional superset
* (fix up a test assertion in badinc.cc, where better sugar handling uncovered a new diagnostic, yay!)
* clean up a few oddities from the mechanical rewrites

This sets the scene for actually fixing a bunch of bugs where IWYU would overlook sugar and break with Clang 15f3cd6b. Coming in a later PR.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2820,7 +2820,7 @@ class InstantiatedTemplateVisitor
     // Among all subst-type params, we only want those in the resugar-map. If
     // we're not in the resugar-map at all, we're not a type corresponding to
     // the template being instantiated, so we can be ignored.
-    type = RemoveSubstTemplateTypeParm(type);
+    type = Desugar(type);
     return ContainsKey(resugar_map_, type);
   }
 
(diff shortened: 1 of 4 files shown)


kimgr:
No tests were fixed by this batch, will merge as a foundation for more fixes.

Issue #1001, opened 2022-01-30
additional implementation include suggested for siginfo_t::si_addr


firewave:
```cpp
#include <signal.h>

static void f(siginfo_t * info) {
	(void)info->si_addr;
}
```

```
/mnt/s/clion/example_lite_2/test.cpp should add these lines:
#include <bits/types/siginfo_t.h>  // for si_addr

/mnt/s/clion/example_lite_2/test.cpp should remove these lines:

The full include-list for /mnt/s/clion/example_lite_2/test.cpp:
#include <bits/types/siginfo_t.h>  // for si_addr
#include <signal.h>                // for siginfo_t
---
```

https://man7.org/linux/man-pages/man2/sigaction.2.html
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html

I am using `include-what-you-use 0.17 based on Debian clang version 13.0.1-+rc1-1~exp4`.

kimgr:
As @firewave noted, this was fixed by PR #1104.

PR #1104, opened 2022-09-01
More mappings for glibc

alejandro-colomar:


gcc.symbols.imp
```diff
@@ -43,6 +43,7 @@
   { symbol: [ "fsblkcnt_t", private, "<sys/statvfs.h>", public ] },
   { symbol: [ "fsfilcnt_t", private, "<sys/types.h>", public ] },
   { symbol: [ "fsfilcnt_t", private, "<sys/statvfs.h>", public ] },
+  { symbol: [ "getopt", private, "<unistd.h>", public ] },
```
kimgr, line 46: The getopt changes kind-of interfere with a pet project of mine -- https://github.com/kimgr/getopt_port. Someone using that compatibility port on Windows will be recommended `<unistd.h>`, which is not available.

That said, our mappings are already very POSIX-/glibc-oriented, so I guess that's fine for now. We should look into making mappings easier to override.

kimgr:
Thanks! Overall, this looks good. A little comment inline on `getopt` and friends, but I'm pretty sure we don't need to worry about it.

gcc.symbols.imp
alejandro-colomar: I guess the same issue would be triggered by libc or cc developers when running iwyu(1) on their code base.  I'd like that to be possible, so yes, there should be some way of telling iwyu(1) that it's running on a special case.

alejandro-colomar:
:-)

kimgr:
Merged, thanks!

firewave:
Looks like this fixed #1001.

PR #1106, opened 2022-09-09
More exact type loc for exceptions

jspam:
This adds the include for the exception type "Exc" when using macros such as BOOST_CHECK_THROW(..., Exc).

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1872,10 +1872,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreCurrentASTNode()) return true;
 
     if (const VarDecl* exception_decl = stmt->getExceptionDecl()) {
+      // Get the caught type from the decl via associated type source info to
+      // get more precise location info for the type use.
       TypeLoc typeloc = exception_decl->getTypeSourceInfo()->getTypeLoc();
       const Type* caught_type = typeloc.getType().getTypePtr();
 
-      // Strip off pointers/references to get to the 'base' type.
+      // Strip off pointers/references to get to the pointee type.
       caught_type = RemovePointersAndReferencesAsWritten(caught_type);
       ReportTypeUse(typeloc.getBeginLoc(), caught_type);
     } else {
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1871,10 +1871,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   bool VisitCXXCatchStmt(clang::CXXCatchStmt* stmt) {
     if (CanIgnoreCurrentASTNode()) return true;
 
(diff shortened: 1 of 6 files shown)


kimgr:
Thank you! I'm a little confused as to why this is necessary. I did some local logging and it looks like this check says the location is not in a macro: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L1235. And the spelling and expansion locations are the same for `CurrentLoc()` in the catch statement, which also indicates something is fishy.

I'd like to investigate this a bit more before taking the patch as-is. Have you done similar digging and found something?

jspam:
It's been a while since I wrote the patch (rebased over several iwyu versions), so I cannot say any more how exactly I arrived at it – sorry. Feel free to view it more as a bug report with attached test case than a ready-to-merge patch.

kimgr:
Cool, thanks for the context!

kimgr:
I think I've convinced myself why this is needed now :-)

* `CurrentLoc()` is inside the macro, and none of the `GetCanonicalUseLocation()` special-cases hit. I think there should be one more to try and detect if the use-location is the expansion of a macro argument (in which case the "caller" of the macro is responsible), but I have not found a way to implement that.
* So the type use from the catch-statement is attributed to `catch_macro.h`, which is not being analyzed, so nothing happens
* But then the AST traversal keeps going, and encounters the `RecordTypeLoc` corresponding to `Exception` inside the catch-statement, and it is evaluated again.
* `VisitTagType`, which does this, does not know that it's coming from `VisitCXXCatchStmt`, so it evaluates it according to all the normal rules for records; since it's a `const Exception&`, a forward-declare is enough, this time attributed to the main source file
* Using the actual exception type decl, as this patch does, gives access to typesource info, which gives a much better location for the type than `CurrentLoc()`, so `VisitCXXCatchStmt` gets a more proper location to base canonical use location analysis on, and it is attributed directly to the main source file.
* By virtue of a full use being needed for the catch, the forward-decl required by `VisitTagType` later is ignored.

kimgr:
I have some notes on the patch, do you want to hash them out here, or would you prefer I did some local cleanup and merged it in your name? Thanks!

jspam:
Thanks for the analysis! Both options are fine for me, please take the one that you think takes the least effort on your side.

kimgr:
@jspam Thanks! Is the `Allow edits from maintainers` option checked on your PR? In that case I can push a couple of fixups, and you can sign them off.

jspam:
Yes, the option is checked.

kimgr:
Thanks, done! I rebased on latest master as well, so had to force-push, but all content changes are in the individual fixups (with a little description in commit message body). Let me know if this is good to go, and I'll squash and edit the commit message a little bit.

jspam:
LGTM, thank you very much!

kimgr:
Thank you!

PR #1108, opened 2022-09-11
Move autocast testing from badinc

bolshakov-a:


kimgr:
Let's go even smaller. If you could do one commit at a time, we'll have CI support to make sure every commit works.

kimgr:
(I merged a test suite fix on master, btw)

bolshakov-a:
> Let's go even smaller. If you could do one commit at a time, we'll have CI support to make sure every commit works.

Did I understand correctly that you mean pushing commits separately? I've done.

kimgr:
I meant let's open one PR for every commit for these cleanup changes. It's easier to review that way.

The Apply formatting commit looks a little odd in the middle there, but I need to look into that and see if it should be squashed or further separated.

bolshakov-a:
Ok

bolshakov-a:
So, separate PR for "Apply formatting" too? Or maybe just one PR for it and its parent commit?

kimgr (approved):

PR #1119, opened 2022-10-03
Test 'autocast' call site handling more precisely

bolshakov-a:
Check that only the last parameters require full type info for autocast.


kimgr:
Very nice, I can really see the purpose of this change with the small increment.

Can you fold the PR description into the commit message:
> Check that only the last parameters require full type info for autocast.

Other than that, LGTM!

kimgr:
Ah, nevermind, I can fix that myself with squash&merge in GitHub's UI

bolshakov-a:
Thanks!

PR #1120, opened 2022-10-03
Better support CRLF line endings in custom lexer

kimgr:
Running the tests/cxx/comment_pragmas.cc test on Windows fails saying that <some_system_header_file> is not diagnosed correctly.

Turns out to be because our custom lexer for GNU @headername comment directives did not respect Windows line endings, and failed properly identify the @headername.

Match on both CR and LF when attempting to find end of line.


kimgr:
This was discovered in a git working copy with `core.autocrlf=true` on Windows. Given that Git can rewrite line endings like this, I don't want to add a test input with CRLF.

Time to shake life into the unittest suite again, I think, but I won't block this change on that work.

Issue #1047, opened 2022-05-12
Backreferences with regexes for include mappings


ingomueller-net:
I would like to do something like this:

```
{ include: ["@<(llvm/.*\.h)>", "public", "\"$1\"", "public"] }
```

That is, declare `"llvm/someheader.h"` as the preferred form for every `someheader.h` given in the form `<llvm/someheader.h>`.

Is something like that possible? If yes, how?

kimgr:
See also:

* https://github.com/include-what-you-use/include-what-you-use/issues/145
* https://github.com/include-what-you-use/include-what-you-use/issues/935#issuecomment-943930251
* https://github.com/include-what-you-use/include-what-you-use/issues/959

So to summarize, I don't think it's currently possible, but I suspect it's not too hard to make it work in `IncludePicker::ExpandRegexes`, https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_include_picker.cc#L1399, if you want to take a stab at it.

kimgr:
As of PR #1123, commit 6fb66575fba17039d1be6c1fc3dca08623163460, there's now support for replacements with backreferences for all regex dialects.

Issue #959, opened 2021-10-20
[Feature Request]: Support backreferences in include mapping's to-entry


4wrxb:
Inspired by #935 with suggestion from kimgr.

> Yeah, that would be nice. I think that logic is driven from `IncludePicker::FinalizeAddedIncludes`, from brief inspection it looks like it should be possible to use regex replacement.

It's going to be at least a month before I can make any attempt at this, but in the meantime I see one other area where I would like some guidance for IWYU's behavior.

In `IncludePicker::AddMappingsFromFile` include mappings are validated either as `IsQuotedFilepathPattern` for from-entry or  `IsQuotedInclude` for to-entry. `IsQuotedFilepathPattern` simply accepts any pattern beginning with '@'. Is it acceptable to use the same definition when checking the to-entry? Regex itself will behave predictably in terms of the substitution even if it's wrong from an `#include` standpoint, but it can be hard to check & validate all possible forms of backreference so I would rather err on the side of making the user figure it out.

kimgr:
Yes, I think it makes sense to loosen validation for @-prefixed entries.

kimgr:
As of PR #1123, commit 6fb66575fba17039d1be6c1fc3dca08623163460, there's now support for replacements with backreferences for all regex dialects.

Issue #145, opened 2015-06-15
Add support for regex for destination in mapping


vsapsai:
Originally reported on Google Code with ID 145

```
The mapping syntax supports a regex expression for the first entry of the mapping. It
would be nice if this could be extended to also be supported for the 3rd entry(=2nd
include). This would be useful if the same include is provided by a lot of other include
files.
```

Reported by `roland@rschulz.eu` on 2014-06-05 05:03:55

kimgr:
Hello, old issue from 2015! As of PR #1123, commit 6fb66575fba17039d1be6c1fc3dca08623163460, there's now support for replacements with backreferences for all regex dialects.

PR #1123, opened 2022-10-07
Add support for group/backreferences regex replacement in mapping files.

graveljp:
This is useful to generically remap files from one directory to another. The following mapping would for instance map all files under the "foo/" directory to "third_party/foo/", for instance, mapping "foo/bar.h" to "third_party/foo/bar.h".

[ { include: ['@"(foo/.*)"', private, '"third_party/\1"', public ] } ]

This is useful for large project where different third_party libraries are compiled together. In Chromium for instance, include paths would look like:
 -I. -Ithird_party/v8/include

This allows code in V8 to include its own headers directly. Other third_parties would include them via "third_party/v8/include/...". Because files are accessible from two different paths, IWYU can't know which should be used. Using a mapping file, adding or removing the "third_party/v8/include" prefix where needed, resolves this problem.

### DIFF
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -1436,10 +1436,14 @@ void IncludePicker::ExpandRegexes() {
   for (const auto& incmap : quoted_includes_to_quoted_includers_) {
     const string& hdr = incmap.first;
     for (const string& regex_key : filepath_include_map_regex_keys) {
+      const string regex = regex_key.substr(1);
       const vector<MappedInclude>& map_to = filepath_include_map_[regex_key];
-      if (RegexMatch(regex_dialect, hdr, regex_key.substr(1)) &&
+      if (RegexMatch(regex_dialect, hdr, regex) &&
           !ContainsQuotedInclude(map_to, hdr)) {
-        Extend(&filepath_include_map_[hdr], filepath_include_map_[regex_key]);
+        for (const MappedInclude& target : map_to) {
+          filepath_include_map_[hdr].push_back(MappedInclude(
+              RegexReplace(regex_dialect, hdr, regex, target.quoted_include)));
+        }
         MarkVisibility(&include_visibility_map_, hdr,
                        include_visibility_map_[regex_key]);
       }
--- a/iwyu_regex.cc
+++ b/iwyu_regex.cc
@@ -46,4 +46,26 @@ bool RegexMatch(RegexDialect dialect, const std::string& str,
   CHECK_UNREACHABLE_("Unexpected regex dialect");
 }
 
+std::string RegexReplace(RegexDialect dialect, const std::string& str,
+                         const std::string& pattern,
+                         const std::string& replacement) {
+  switch (dialect) {
+    case RegexDialect::LLVM: {
+      // llvm::Regex::sub  has search semantics. Enclose the pattern in ^...$
+      // for start/end anchoring.
+      llvm::Regex r("^" + pattern + "$");
+      return r.sub(replacement, str);
+    }
+
+    case RegexDialect::ECMAScript: {
+      // std::regex_replace has search semantics. Enclose the pattern in ^...$
+      // for start/end anchoring.
+      std::regex r("^" + pattern + "$", std::regex_constants::ECMAScript);
+      return std::regex_replace(str, r, replacement,
+                                std::regex_constants::format_first_only);
+    }
+  }
+  CHECK_UNREACHABLE_("Unexpected regex dialect");
+}
+
 }  // namespace include_what_you_use
--- a/iwyu_regex.h
+++ b/iwyu_regex.h
@@ -23,6 +23,12 @@ bool ParseRegexDialect(const char* str, RegexDialect* dialect);
 bool RegexMatch(RegexDialect dialect, const std::string& str,
                 const std::string& pattern);
 
+// Returns input string with the first match of pattern replaced, for the given
+// regex dialect.
+std::string RegexReplace(RegexDialect dialect, const std::string& str,
+                         const std::string& pattern,
+                         const std::string& replacement);
+
 }  // namespace include_what_you_use
 
 #endif  // INCLUDE_WHAT_YOU_USE_IWYU_REGEX_H_
--- a/tests/cxx/mapping_replace_regex_ecmascript.cc
+++ b/tests/cxx/mapping_replace_regex_ecmascript.cc
@@ -1,0 +1,38 @@
+//===--- mapping_replace_regex_ecmascript.cc - test input file for iwyu ---===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I . \
+//            -Xiwyu --regex=ecmascript \
+//            -Xiwyu --mapping_file=tests/cxx/mapping_replace_regex_ecmascript.imp
+
+// Generically map include paths to a different include directory:
+// * The include of tests/cxx/direct.h should nominally be replaced by
+//   tests/cxx/indirect.h, where IndirectClass is defined
+// * But we provide a mapping adding a "foobar" prefix to any include under
+//   "tests/cxx", resulting in suggesting foobar/tests/cxx/indirect.h.
+
+#include "tests/cxx/direct.h"
+
+void f() {
+  // IWYU: IndirectClass is defined in "foobar/tests/cxx/indirect.h"
+  IndirectClass i;
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/mapping_replace_regex_ecmascript.cc should add these lines:
+#include "foobar/tests/cxx/indirect.h"
+
+tests/cxx/mapping_replace_regex_ecmascript.cc should remove these lines:
+- #include "tests/cxx/direct.h"  // lines XX-XX
+
+The full include-list for tests/cxx/mapping_replace_regex_ecmascript.cc:
+#include "foobar/tests/cxx/indirect.h"  // for IndirectClass
+
+***** IWYU_SUMMARY */
--- a/tests/cxx/mapping_replace_regex_ecmascript.imp
+++ b/tests/cxx/mapping_replace_regex_ecmascript.imp
@@ -1,0 +1,5 @@
+# Maps headers to a different include path.
+# The foobar directory does not exist; we just want IWYU to suggest using it.
+[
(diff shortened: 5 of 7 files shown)


graveljp:
This is also very useful to prevent IWYU from suggesting private headers in libraries we don't control. For instance, this mapping prevents IWYU from suggesting internal headers from the standard library, instead suggesting the public official one:
 { include: ['@<__memory/.*>', 'private', '<memory>', 'public'] }

tests/cxx/mapping_replace_regex_ecmascript.imp
```diff
@@ -0,0 +1,5 @@
+# Maps headers to a different include path.
+# The foobar.h header does not exist; we just want IWYU to suggest using it.
```
kimgr, line 2: `The foobar directory [...]`

tests/cxx/mapping_replace_regex_llvm.imp
```diff
@@ -0,0 +1,5 @@
+# Maps headers to a different include path.
+# The foobar.h header does not exist; we just want IWYU to suggest using it.
```
kimgr, line 2: `The foobar directory [...]`

iwyu_regex.h
```diff
@@ -23,6 +23,13 @@ bool ParseRegexDialect(const char* str, RegexDialect* dialect);
 bool RegexMatch(RegexDialect dialect, const std::string& str,
                 const std::string& pattern);
 
+// Returns the result of replacing the first match of the `pattern` regular
```
kimgr, line 26: We don't have a tradition of backticks in comments in IWYU source, and that character is a little messy on international keyboards, so I'd like to avoid it. Could you phrase this as something like:
```
Returns input string with the first match of pattern replaced, for the given regex dialect.
```
to keep the markup to a minimum? Thanks

tests/cxx/mapping_replace_regex_ecmascript.cc
```diff
@@ -0,0 +1,38 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I . \
+//            -Xiwyu --regex=ecmascript \
+//            -Xiwyu --mapping_file=tests/cxx/mapping_replace_regex_ecmascript.imp
+
+// Generically map include paths to a different include folder:
```
kimgr, line 14: Please use "directory" instead of "folder"

tests/cxx/mapping_replace_regex_llvm.cc
```diff
@@ -0,0 +1,38 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I . \
+//            -Xiwyu --regex=llvm \
+//            -Xiwyu --mapping_file=tests/cxx/mapping_replace_regex_llvm.imp
+
+// Generically map include paths to a different include folder:
```
kimgr, line 14: Please use "directory" instead of "folder"

iwyu_regex.cc
```diff
@@ -46,4 +46,31 @@ bool RegexMatch(RegexDialect dialect, const std::string& str,
   CHECK_UNREACHABLE_("Unexpected regex dialect");
 }
 
+// Returns the result of replacing the first match of the `pattern` regular
+// expression in `str` with `replacement`. `dialect` specifies what regular
+// expression dialect to use.
+std::string RegexReplace(
```
kimgr, line 52: Formatting looks a little off here -- could you use clang-format on the new code? (I also have to recommend https://github.com/kimgr/git-format/, because that's my favorite way to clean up patch sets :smile:)

iwyu_include_picker.cc
```diff
@@ -1436,10 +1436,15 @@ void IncludePicker::ExpandRegexes() {
...
     for (const string& regex_key : filepath_include_map_regex_keys) {
+      const string regex = regex_key.substr(1);
       const vector<MappedInclude>& map_to = filepath_include_map_[regex_key];
-      if (RegexMatch(regex_dialect, hdr, regex_key.substr(1)) &&
+      if (RegexMatch(regex_dialect, hdr, regex) &&
           !ContainsQuotedInclude(map_to, hdr)) {
-        Extend(&filepath_include_map_[hdr], filepath_include_map_[regex_key]);
+        for (const MappedInclude& target : map_to) {
+          filepath_include_map_[hdr].push_back(
+            MappedInclude(RegexReplace(regex_dialect, hdr, regex,
```
kimgr, line 1445: Could you pull the `RegexReplace` result out into a temporary to keep line length down?

Ah, and I think the only misformat is that we usually indent continued lines by four spaces relative to previous line's indentation. Clang-format should fix that for you automatically.

kimgr (commented):
Very cool, thank you!

Some nits inline, and here is some commit message feedback:

> This is useful to generically remap files from one folder to another.
> The following mapping would for instance map all files under the "foo/"
> folder to "third_party/foo/", for instance, mapping "foo/bar.h" to

We prefer the posix term "directory" to "folder", so please use that in the two instance above too.

> "third_party/foo/bar.h".
>
>    [ { include: ['@"(foo/.*)"', private, '"third_party/\1"', public ] } ]
>
> This is useful for large project where different thrid_party libraries

Typo: thrid_party

> are compiled together. In Chromium for instance, include paths would
> look like:
>    -Ithird_party -Ithird_party/v8/include
>
> This allows code in V8 to include its own headers directly. Other
> third_parties would include them via "third_party/v8/include/...".
> Because files are accessible from two different paths, IWYU can't know
> which should be used. Using a mapping file, adding or removing the
> "third_party/v8/include" prefix where needed, resolves this problem.

Otherwise excellent commit message with a good motivation for the change. Thanks!

kimgr:
I ran a little perf test comparing `master` and your branch with
```
$ cat t.cc
#include <string>

$ include-what-you-use -Xiwyu --mapping_file=qt5_4.imp t.cc
```

`qt5_4.imp` is infamous for having MANY regex mappings. I could not see any particular difference between your branch and mainline here (for either of `llvm` or `ecmascript` dialects), so... regex replacements are free? That doesn't seem right to me. Am I missing something?

iwyu_include_picker.cc
graveljp: Clang-format made it fit on two lines, removing the need for a local variable. Using local variables isn't always useful or desired, as described in [abseil.io/tips/161](https://abseil.io/tips/161).

iwyu_regex.cc
graveljp: My favorite way is to hit a shortcut to invoke clang-format in my IDE ;)
It looks like lots of code isn't correctly formatted in this file. Perhaps a batch formatting of the code base would be useful. Unless my clang-format isn't formatting as it should?

iwyu_regex.h
graveljp: Sounds good. Short and simple!
Backticks are useful to allow code analysis tool to be able to find and index symbols in comments. Unlike other common delimiters ('|' for instance), backticks do not clash with the rest of the C++ syntax.

graveljp:
> I could not see any particular difference between your branch and mainline here (for either of llvm or ecmascript dialects), so... regex replacements are free? That doesn't seem right to me. Am I missing something?

Maybe the regex expansion is not a performance bottleneck? Regex replaces are certainly an order of magnitude faster than file IO.

iwyu_regex.cc
kimgr: That's why I prefer to do it patch-hunk-by-patch-hunk instead of file-by-file. The Emacs clang-format plugin does region formatting, so I typically just reformat what I just changed, and then use `git-format` to check/fix-up individual commits.

But yeah, at some point maybe we should take the plunge and reformat everything. I just don't like such big changes :)

iwyu_include_picker.cc
kimgr: Yeah, all else equal I prefer to avoid temporary locals, but long lines or lines with complex wrapping are usually harder for me to read, so I tend to err on the side of less complex expressions. But I think it looks good after the re-format, thanks!

kimgr:
The difference between regex dialects `llvm` and `ecmascript` came about because `std::regex` was an order of magnitude slower than `llvm::Regex` for simple matching. I have no reason to think replacement would be less costly. See #981 for some background on the regex performance.

I'll see if I can

iwyu_regex.cc
```diff
@@ -46,4 +46,29 @@ bool RegexMatch(RegexDialect dialect, const std::string& str,
   CHECK_UNREACHABLE_("Unexpected regex dialect");
 }
 
+// Returns the result of replacing the first match of the `pattern` regular
```
kimgr, line 49: I think we can ditch this comment; it's covered in the header

iwyu_regex.h
graveljp: Also, if I may venture in "programming wisdom" territory, consistency (or tradition as you put it) shouldn't be used as a blocker for improving the code style or quality. Any coding style or language improvement has to go through a phase where newer code breaks consistency with old one (unless you batch-apply style improvements to the whole code base at once, but that's not scalable, if at all possible). Or course, style improvements should be discussed and approved, so you are right in pointing inconsistencies where they occur.

With this in mind, note that backticked symbols in comments are understood by clangd, so if IWYU started using them, developers would see immediate improvements to code indexing in their IDE.

kimgr:
So with
```
$ cat t.cc
#include <QtCore/QSize>
$ include-what-you-use -I /usr/include/x86_64-linux-gnu/qt5 -Xiwyu --mapping_file=qt5_11.imp -Xiwyu --regex=ecmascript t.cc
$ include-what-you-use -I /usr/include/x86_64-linux-gnu/qt5 -Xiwyu --mapping_file=qt5_11.imp -Xiwyu --regex=llvm t.cc
```
I have exercised this code and noticed no significant difference between master and branch. It actually looks like branch is slightly faster for both `ecmascript` and `llvm`, which is a nice surprise :sweat_smile:

kimgr:
Merging this should close:

* https://github.com/include-what-you-use/include-what-you-use/issues/959
* https://github.com/include-what-you-use/include-what-you-use/issues/1047
* https://github.com/include-what-you-use/include-what-you-use/issues/145

Let me know if you want to squash branch history and fix the nits in commit message, or if you trust me to do a squash&merge here. Thank you for contributing this, it turned out much simpler than I thought it would!

graveljp:
I squashed the commits. Thanks!

kimgr:
Beautiful, thanks! Merging as soon as the CI build lets me.

PR #1128, opened 2022-10-29
Fixed IWYU builds by updating clang api

danielh2942:


kimgr:
Patch looks good. Commit message needs some work, though --

* we use `[clang compat]` to signal that a change is triggered by Clang API changes
* we always refer to the breaking Clang commit, to make it easier to reason about point compatibility

See e.g. https://github.com/include-what-you-use/include-what-you-use/commit/164b8fe7597805ae55f029ecf6580dc46a74c7ed and https://github.com/include-what-you-use/include-what-you-use/commit/046ca2238054eeb6817f8d6fa3735c8782e3a511 for examples.

kimgr:
Sorry for all the nit-picking, but could you try to shorten the subject line? When GitHub wraps it with `[...]`, there's a good chance it's too wide for many `git log` terminals too. The strictest convention limit the subject line to 50 chars, but around 70 is usually OK.

I suggest:
```
[clang compat] Use TemplateSpecializationType ArrayRef APIs

... body as-is
```

One nice thing I noticed is the `ArrayRef` APIs have been around since 2016, so this is not a breaking change vs. Clang/LLVM 15, for which I'm preparing a release now. We can get this merged and include it in the release without compatibility concerns.

danielh2942:
Done!

kimgr:
Thanks!

Issue #1095, opened 2022-08-12
begin_keep


FStefanni:
Hi,

I like to suggest to support a new couple of pragmas:
* `begin_keep`
* `end_keep`

The meaning should be to force keeping of includes between the two pragmas.
This is useful in cases like libraries development: usually there are some headers which just include other ones, to provide the implementation to the library users, e.g.:
``` c++
// header mylib.hpp
#include "MyClass1.hpp"
#include "MyClass2.hpp"
...
```
Currently the only way to suppress the warnings is to insert `// IWYU pragma: keep` in each line, which is not very ergonomic.
With the proposed feature we could instead write:
``` c++
// header mylib.hpp
// IWYU pragma: begin_keep
#include "MyClass1.hpp"
#include "MyClass2.hpp"
...
// IWYU pragma: end_keep
```
Regards

kimgr:
That sounds like a reasonable extension. I guess you could model it on `begin_exports` and `end_exports`. Do you want to take a stab at a patch?

FStefanni:
Hi,

sorry but I have no time to work on this.

Regards

danielh2942:
@FStefanni I implemented the request if you want to have a look at it (It currently won't build as there's another PR I have open that it depends on to fix the CI)

FStefanni:
Hi @danielh2942,

thank you for the pr, great work. 
Just giving a quick check it seems to do exactly what we need (I do not know IWYU internals).

I hope this will be included asap.

Regards.

kimgr:
#1129 is now merged, closing this.

PR #1129, opened 2022-10-29
Added begin_keep/end_keep pragma pair 

danielh2942:
I implemented the feature request from #1095 
I had to do some messing about to allow forward declares to follow the same behaviour

Wrote a test to verify and updated documentation to reflect the change

(This will not build without the fix from #1128 as a heads up)

### DIFF
--- a/docs/IWYUPragmas.md
+++ b/docs/IWYUPragmas.md
@@ -15,6 +15,20 @@ This pragma applies to a single `#include` directive or forward declaration. It 
 
 In this case, `std::vector` isn't used, so `<vector>` would normally be discarded, but the pragma instructs IWYU to leave it. Similarly the class `ForwardDeclaration` isn't used but is kept because of the pragma on it.
 
+## IWYU pragma: begin_keep/end_keep ##
+
+This pragma applies to a set of `#include` directives and forward declarations. It declares that the headers and forward declarations in between are to be left alone by IWYU.
+
+    main.cc:
+      // IWYU pragma: begin_keep
+      #include <vector>
+      #include <iostream>
+
+      class MyClass;
+      // IWYU pragma: end_keep
+
+In the provided case nothing within the bounds of `begin_keep` and `end_keep` will be discarded.
+
 ## IWYU pragma: export ##
 
 This pragma applies to a single `#include` directive. It says that the current file is to be considered the provider of any symbol from the included file.
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -243,6 +243,11 @@ Used after
 .B #include
 directives or forward declarations it ensures that they won't be removed.
 .TP
+.BR "// IWYU pragma: begin_keep" , " // IWYU pragma: end_keep"
+Has the same effect as the previous pragma comment, but applies to a range of
+.BR #include s
+and forward declarations instead of a single line.
+.TP
 .B // IWYU pragma: export
 Used after an
 .B #include
--- a/iwyu_preprocessor.cc
+++ b/iwyu_preprocessor.cc
@@ -177,6 +177,14 @@ bool IwyuPreprocessorInfo::HasOpenBeginExports(const FileEntry* file) const {
       GetFileEntry(begin_exports_location_stack_.top()) == file;
 }
 
+// Only call this when the given files is the one being processed
+// Only return true if there is an open begin_keep pragma in the current
+// state of the parse of the given file.
+bool IwyuPreprocessorInfo::HasOpenBeginKeep(const FileEntry* file) const {
+  return !begin_keep_location_stack_.empty() &&
+      GetFileEntry(begin_keep_location_stack_.top()) == file;
+}
+
 bool IwyuPreprocessorInfo::HandleComment(Preprocessor& pp,
                                          SourceRange comment_range) {
   HandlePragmaComment(comment_range);
@@ -209,6 +217,20 @@ void IwyuPreprocessorInfo::HandlePragmaComment(SourceRange comment_range) {
     return;
   }
 
+  if (HasOpenBeginKeep(this_file_entry)) {
+    if (MatchOneToken(tokens, "end_keep", 1, begin_loc)) {
+      ERRSYM(this_file_entry) << "end_keep pragma seen\n";
+      SourceLocation keep_loc_begin = begin_keep_location_stack_.top();
+      begin_keep_location_stack_.pop();
+      SourceRange keep_range(keep_loc_begin, begin_loc);
+      keep_location_ranges_.insert(std::make_pair(this_file_entry, keep_range));
+    } else {
+      // No pragmas allowed within "begin_keep" - "end_keep"
+      Warn(begin_loc, "Expected end_keep pragma");
+    }
+    return;
+  }
+
   if (MatchOneToken(tokens, "begin_exports", 1, begin_loc)) {
     ERRSYM(this_file_entry) << "begin_exports pragma seen\n";
     begin_exports_location_stack_.push(begin_loc);
@@ -217,6 +239,17 @@ void IwyuPreprocessorInfo::HandlePragmaComment(SourceRange comment_range) {
 
   if (MatchOneToken(tokens, "end_exports", 1, begin_loc)) {
     Warn(begin_loc, "end_exports without a begin_exports");
+    return;
+  }
+
+  if (MatchOneToken(tokens, "begin_keep", 1, begin_loc)) {
+    ERRSYM(this_file_entry) << "begin_keep pragma seen\n";
+    begin_keep_location_stack_.push(begin_loc);
+    return;
+  }
+
+  if (MatchOneToken(tokens, "end_keep", 1, begin_loc)) {
+    Warn(begin_loc, "end_keep without a begin_keep");
     return;
   }
 
@@ -397,7 +430,8 @@ void IwyuPreprocessorInfo::MaybeProtectInclude(
   // interpreted as a pragma. Maybe do "keep" and "export" pragma handling
   // in HandleComment?
   if (LineHasText(includer_loc, "// IWYU pragma: keep") ||
-      LineHasText(includer_loc, "/* IWYU pragma: keep")) {
+      LineHasText(includer_loc, "/* IWYU pragma: keep") ||
+      HasOpenBeginKeep(includer)) {
     protect_reason = "pragma_keep";
     FileInfoFor(includer)->ReportKnownDesiredFile(includee);
 
@@ -790,6 +824,12 @@ void IwyuPreprocessorInfo::FileChanged_ExitToFile(
          "begin_exports without an end_exports");
     begin_exports_location_stack_.pop();
   }
+  
+  if (HasOpenBeginKeep(exiting_from)) {
+    Warn(begin_keep_location_stack_.top(),
+         "begin_keep without an end_keep");
+    begin_keep_location_stack_.pop();
+  }
 }
 
 void IwyuPreprocessorInfo::FileChanged_RenameFile(SourceLocation new_file) {
@@ -1085,6 +1125,17 @@ bool IwyuPreprocessorInfo::ForwardDeclareIsInhibited(
       FindInMap(&no_forward_declare_map_, file);
   return (inhibited_forward_declares != nullptr) &&
       ContainsKey(*inhibited_forward_declares, normalized_symbol_name);
+}
+
+bool IwyuPreprocessorInfo::ForwardDeclareInKeepRange(SourceLocation loc) const {
+  const FileEntry* file = GetFileEntry(loc);
+  auto keep_ranges = keep_location_ranges_.equal_range(file);
+  for (auto it = keep_ranges.first; it != keep_ranges.second; ++it) {
+    if (it->second.fullyContains(loc)) {
+      return true;
+    }
+  }
+  return false;
 }
 
 }  // namespace include_what_you_use
--- a/iwyu_preprocessor.h
+++ b/iwyu_preprocessor.h
@@ -34,6 +34,8 @@
 //    f) // IWYU pragma: no_forward_declare foo::Bar
 //    g) // IWYU pragma: friend <regexp>
 //       // IWYU pragma: friend "<regexp>" -- needed if spaces in regexp.
+//    h) // IWYU pragma: begin_keep
+//    i) // IWYU pragma: end_keep
 // 'Annotation' constructs:
 //    h) #include "foo/bar/baz.h"  // IWYU pragma: export
 //    i) #include "foo/bar/baz.h"  // IWYU pragma: keep
@@ -86,6 +88,7 @@ using std::set;
 using std::stack;
 using std::string;
 using std::vector;
+using std::multimap;
 
 class IwyuPreprocessorInfo : public clang::PPCallbacks,
                              public clang::CommentHandler {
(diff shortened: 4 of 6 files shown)


danielh2942:
made modification to use a map of vectors as opposed to a plain vector to reduce search space

include-what-you-use.1
```diff
@@ -243,6 +243,11 @@ Used after
 .B #include
 directives or forward declarations it ensures that they won't be removed.
 .TP
+.BR "// IWYU pragma: begin_keep"  , " // IWYU pragma: end_keep"
```
kimgr, line 246: Two spaces between `"` and comma, I think one is sufficient (unless there's something in man/troff syntax that requires it)

iwyu_preprocessor.cc
```diff
@@ -209,6 +217,19 @@ void IwyuPreprocessorInfo::HandlePragmaComment(SourceRange comment_range) {
     return;
   }
 
+  if (HasOpenBeginKeep(this_file_entry)) {
+    if (MatchOneToken(tokens, "end_keep", 1, begin_loc)) {
+      ERRSYM(this_file_entry) << "end_keep pragma seen\n";
+      SourceLocation temp = begin_keep_location_stack_.top();
+	  begin_keep_location_stack_.pop();
+	  keep_location_ranges_[GetFileEntry(temp)].push_back(clang::SourceRange{temp,begin_loc});
```
kimgr, line 225: These two lines are overindented. And the second might possibly be too long (column limit is 80)

kimgr: Also we always have space after comma (`clang::SourceRange(temp, begin_loc)`).

Further, would be nice if you added a `using clang::SourceRange` to the top, so we don't need to qualify it here.

iwyu_preprocessor.cc
```diff
@@ -397,7 +429,8 @@ void IwyuPreprocessorInfo::MaybeProtectInclude(
   // interpreted as a pragma. Maybe do "keep" and "export" pragma handling
   // in HandleComment?
   if (LineHasText(includer_loc, "// IWYU pragma: keep") ||
-      LineHasText(includer_loc, "/* IWYU pragma: keep")) {
+      LineHasText(includer_loc, "/* IWYU pragma: keep") ||
+	  HasOpenBeginKeep(includer)) {
```
kimgr, line 433: Align this with `LineHasText` on the line above

iwyu_preprocessor.cc
```diff
@@ -1087,4 +1126,21 @@ bool IwyuPreprocessorInfo::ForwardDeclareIsInhibited(
       ContainsKey(*inhibited_forward_declares, normalized_symbol_name);
 }
 
+bool IwyuPreprocessorInfo::ForwardDeclareInKeepRange(
```
kimgr, line 1129: You should be able to drop the `clang::` qualifier throughout this method. If it's necessary for some symbol, prefer to add a using declaration.

iwyu_preprocessor.cc
```diff
@@ -1087,4 +1126,21 @@ bool IwyuPreprocessorInfo::ForwardDeclareIsInhibited(
...
+    const clang::SourceLocation& loc) const {
+  // Check if the entry exists, shortcircuit if not
+  const clang::FileEntry* ref = GetFileEntry(loc);
```
kimgr, line 1132: `ref` -> `file`

iwyu_preprocessor.cc
```diff
@@ -1087,4 +1126,21 @@ bool IwyuPreprocessorInfo::ForwardDeclareIsInhibited(
...
+  if (keep_location_ranges_.find(ref) == keep_location_ranges_.end()) {
+    return false;
+  }
+
+  const clang::SourceRange mRange{loc};
```
kimgr, line 1137: We don't use camelCase in IWYU, but rather hacker_case. Should this maybe be called `decl_range`?

iwyu_preprocessor.cc
```diff
@@ -1087,4 +1126,21 @@ bool IwyuPreprocessorInfo::ForwardDeclareIsInhibited(
...
+  for (auto const& range : keep_location_ranges_.at(ref)) {
```
kimgr, line 1138: We've already done this lookup with the find above, so consider caching find result in an iterator:
```
auto it = keep_location_ranges_.find(ref);
if (it == keep_location_ranges_.end()) {
  return false;
}

for (const auto& range : *it) {
   ...
}

File: iwyu_preprocessor.h [hunk 8]
```diff
@@ -345,6 +355,15 @@ class IwyuPreprocessorInfo : public clang::PPCallbacks,
   // per file in the current inclusion chain..
   stack<clang::SourceLocation> begin_exports_location_stack_;
 
+  // For processing pragmas. It is the current stack of open
+  // "begin_keep".     There should be at most one item in this stack
+  // per file in the current inclusion chain..
+  stack<clang::SourceLocation> begin_keep_location_stack_;
+
+  // For processing forward decls. It is a Vector of SourceRanges
```
kimgr, line 363: Consider making this a `std::multimap` instead.

tests/cxx/pragma_keep_multi.cc
```diff
@@ -0,0 +1,43 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+// This tests whether or not begin_keep/end_keep works
+
+// IWYU pragma: begin_keep
```
kimgr, line 14: I would like to see a few more variations:

* Empty `begin_keep`/`end_keep` block
* `begin_keep`/`end_keep` block with a single include line
* `begin_keep`/`end_keep` block with multiple include lines
* `begin_keep`/`end_keep` block with a single forward declaration
* `begin_keep`/`end_keep` block with multiple forward declarations
* `begin_keep`/`end_keep` block with include lines and forward declarations mixed

For the most part you can just regroup the existing includes, but you may need some new forward declarations.

iwyu_preprocessor.cc
```diff
@@ -209,6 +217,19 @@ void IwyuPreprocessorInfo::HandlePragmaComment(SourceRange comment_range) {
     return;
   }
 
+  if (HasOpenBeginKeep(this_file_entry)) {
+    if (MatchOneToken(tokens, "end_keep", 1, begin_loc)) {
+      ERRSYM(this_file_entry) << "end_keep pragma seen\n";
+      SourceLocation temp = begin_keep_location_stack_.top();
```
kimgr, line 223: `temp` is a little vague. Maybe `begin_keep_loc`?

Also, don't we know that `GetFileEntry(temp) == this_file_entry` by virtue of `HasOpenBeginExports` being true? So you can use `this_file_entry` for `keep_location_ranges` key below.

kimgr (commented):
Very nice, thank you!

General approach looks great, some stylistic notes inline.

Consider rephrasing commit message to play more with https://cbea.ms/git-commit/. Primarily, remove the issue reference from subject line (might be nice in the body, though).

danielh2942:
Addressed feedback and reworded commit message

tests/cxx/pragma_keep_multi.cc
```diff
@@ -0,0 +1,88 @@
...
+// IWYU_ARGS: -I .
+
+// This tests whether or not begin_keep/end_keep works
+
+// Check empty block
+
+// IWYU pragma: begin_keep
+// IWYU pragma: end_keep
+
+// Check Block with one include
```
kimgr, line 19: Block -> block

iwyu_preprocessor.cc
```diff
@@ -1087,4 +1129,25 @@ bool IwyuPreprocessorInfo::ForwardDeclareIsInhibited(
       ContainsKey(*inhibited_forward_declares, normalized_symbol_name);
 }
 
+bool IwyuPreprocessorInfo::ForwardDeclareInKeepRange(
+    const SourceLocation& loc) const {
+  // Check if the entry exists, shortcircuit if not
```
kimgr, line 1134: The style for IWYU comments follows Google style: https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar

Please make them full sentences with capitalization and punctuation. Here and below.

iwyu_preprocessor.cc
```diff
@@ -1087,4 +1129,25 @@ bool IwyuPreprocessorInfo::ForwardDeclareIsInhibited(
       ContainsKey(*inhibited_forward_declares, normalized_symbol_name);
 }
 
+bool IwyuPreprocessorInfo::ForwardDeclareInKeepRange(
+    const SourceLocation& loc) const {
```
kimgr, line 1133: `SourceLocation` is a value type, so you can pass it by val instead. That also makes the definition fit on one 80-column line.

iwyu_preprocessor.cc
```diff
@@ -1087,4 +1129,25 @@ bool IwyuPreprocessorInfo::ForwardDeclareIsInhibited(
...
+  const FileEntry* file = GetFileEntry(loc);
+
+  auto keep_ranges = keep_location_ranges_.equal_range(file);
+
+  // with multimaps first==second if and only if there's no matches
+  if (keep_ranges.first == keep_ranges.second) {
+    return false;
+  }
+
+  const SourceRange decl_range{loc};
```
kimgr, line 1144: We typically don't use `const` for local variables, or `{...}` initialization syntax. This would be easier to read as simply `SourceRange decl_range(loc);`

iwyu_preprocessor.cc
```diff
@@ -1087,4 +1129,25 @@ bool IwyuPreprocessorInfo::ForwardDeclareIsInhibited(
...
+  for (auto it = keep_ranges.first; it != keep_ranges.second; ++it) {
```
kimgr, line 1145: Looking at this loop now, I notice a few things --

* the short-circuit can be removed, it's the automatic behavior of the loop
* `SourceRange` implicitly converts from `SourceLocation`, so you don't need `decl_range`

So I think you could get away with:
```
const FileEntry* file = GetFileEntry(loc);
auto keep_ranges = keep_location_ranges_.equal_range(file);
for (auto it = keep_ranges.first; it != keep_ranges.second; ++it) {
  if (it->second.fullyContains(loc) {
    return true;
  }
}
return false;
```

iwyu_preprocessor.cc
```diff
@@ -209,6 +217,22 @@ void IwyuPreprocessorInfo::HandlePragmaComment(SourceRange comment_range) {
     return;
   }
 
+  if (HasOpenBeginKeep(this_file_entry)) {
+    if (MatchOneToken(tokens, "end_keep", 1, begin_loc)) {
+      ERRSYM(this_file_entry) << "end_keep pragma seen\n";
+      SourceLocation keep_loc_begin = begin_keep_location_stack_.top();
+      begin_keep_location_stack_.pop();
+      keep_location_ranges_.insert({this_file_entry,
```
kimgr, line 225: Initialization syntax is rare in IWYU, I would prefer
```
SourceRange keep_range(keep_loc_begin, begin_loc);
keep_location_ranges_.insert(std::make_pair(this_file_entry, keep_range));
```
here, to avoid implicit typing and line wrapping.

iwyu_preprocessor.h
```diff
@@ -345,6 +356,15 @@ class IwyuPreprocessorInfo : public clang::PPCallbacks,
   // per file in the current inclusion chain..
   stack<clang::SourceLocation> begin_exports_location_stack_;
 
+  // For processing pragmas. It is the current stack of open
+  // "begin_keep".     There should be at most one item in this stack
```
kimgr, line 360: The two spaces after period for `begin_exports_location_stack_` is some kind of editor setting from the original authors, I think. You can use two spaces or one, but this alignment looks funny. I think the most reasonable formatting would be:
```
  // For processing pragmas. It is the current stack of open "begin_keeps".
  // There should be at most one item in this stack per file in the current
  // inclusion chain.
```

iwyu_preprocessor.h
```diff
@@ -345,6 +356,15 @@ class IwyuPreprocessorInfo : public clang::PPCallbacks,
...
+  // per file in the current inclusion chain..
+  stack<clang::SourceLocation> begin_keep_location_stack_;
+
+  // For processing forward decls. It is a multimap
```
kimgr, line 364: Make sure you fill this comment to 80 columns e.g.
```
  // For processing forward decls. It is a multimap containing the bounds of
  // every keep_location.
```

iwyu_preprocessor.h
```diff
@@ -345,6 +356,15 @@ class IwyuPreprocessorInfo : public clang::PPCallbacks,
...
+  // containing the bounds of every keep_location.
```
kimgr, line 365: keep_location -> keep range?

kimgr:
Very nice! I took another pass and came up with a few more improvement suggestions, mostly around style.

Please rebase on latest master to get your clang-compat fix in too.

danielh2942:
Addressed feedback

iwyu_preprocessor.cc
```diff
@@ -209,6 +217,20 @@ void IwyuPreprocessorInfo::HandlePragmaComment(SourceRange comment_range) {
     return;
   }
 
+  if (HasOpenBeginKeep(this_file_entry)) {
+    if (MatchOneToken(tokens, "end_keep", 1, begin_loc)) {
+      ERRSYM(this_file_entry) << "end_keep pragma seen\n";
+      SourceLocation keep_loc_begin = begin_keep_location_stack_.top();
+      begin_keep_location_stack_.pop();
+	  SourceRange keep_range(keep_loc_begin, begin_loc);
```
kimgr, line 225: over-indented

iwyu_preprocessor.cc
```diff
@@ -1087,4 +1127,16 @@ bool IwyuPreprocessorInfo::ForwardDeclareIsInhibited(
       ContainsKey(*inhibited_forward_declares, normalized_symbol_name);
 }
 
+bool IwyuPreprocessorInfo::ForwardDeclareInKeepRange(SourceLocation loc) const {
+
```
kimgr, line 1131: Kill this blank line

kimgr (commented):
Looks great. Just two more formatting nits, then we're good to go.

danielh2942:
> `include-cleaner` from LLVM added capturing of IWYU pragmas yesterday in [llvm/llvm-project@9ab0827](https://github.com/llvm/llvm-project/commit/9ab0827f70a524aa9f94995662d37675cd042f30).
> 
> CC'ing the author @hokein to maybe get some feedback ahead of extending the existing pragmas.

If an extra pair of hands is wanted to add the new pragmas to clang-include-cleaner I don't mind writing it :)

danielh2942:
I butchered the git history, trying to fix

danielh2942:
fixed git history

docs/IWYUPragmas.md
```diff
@@ -15,6 +15,20 @@ This pragma applies to a single `#include` directive or forward declaration. It
...
+
+    main.cc:
+      // IWYU pragma: begin_keep
+      #include <vector>
+      #include <iostream>
+
+      class MyClass;
+      // IWYU pragma: end_keep
+
+In the provided case nothing within the bounds of of `begin_keep` and `end_keep` will be discarded.
```
kimgr, line 30: Gah! Typo: `of of`

danielh2942:
fixed typo

kimgr:
Thanks! I didn't want to delay this any longer, so I was disappointed to see I'd missed it in earlier review rounds. Thanks for the quick fixup, I'll merge as soon as CI has passed.

PR #1136, opened 2022-11-06
Fix template instantiation reporting

bolshakov-a:
Full template specialization type use requires full information about its template-argument-dependent fields and nested typedefs. But earlier, it wasn't reported in many cases when template specialization type isn't written explicitly in non-fwd-decl context.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1665,6 +1665,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
                                                                  comment);
       }
     } else {
+      if (const auto* template_spec_type =
+              dyn_cast<TemplateSpecializationType>(Desugar(type))) {
+        this->getDerived().ReportTplSpecComponentTypes(template_spec_type);
+      }
       if (const NamedDecl* decl = TypeToDeclAsWritten(type)) {
         decl = GetDefinitionAsWritten(decl);
         VERRS(6) << "(For type " << PrintableType(type) << "):\n";
@@ -2668,6 +2672,17 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
   void AddProcessedOverloadLoc(SourceLocation loc) {
     visitor_state_->processed_overload_locs.insert(loc);
+  }
+
+  // Report types needed for template instantiation in cases when template
+  // specialization type isn't explicitly written in a source code
+  // in a non-fwd-declarable context (otherwise, they should be reported from
+  // VisitTemplateSpecializationType), e.g.:
+  // template <class T> struct S { T t; };
+  // void fn(const S<Class>& s) { // Forward declarations are sufficient here.
+  //   (void)s.t; // Full 'Class' type is needed due to template instantiation.
+  // }
+  void ReportTplSpecComponentTypes(const TemplateSpecializationType*) {
   }
 
   // Do not add any variables here!  If you do, they will not be shared
@@ -4195,6 +4210,16 @@ class IwyuAstConsumer
         callee, parent_type,
         current_ast_node(), resugar_map);
     return true;
+  }
+
+  // --- Handler declared in IwyuBaseASTVisitor.
+
+  void ReportTplSpecComponentTypes(const TemplateSpecializationType* type) {
+    const map<const Type*, const Type*> resugar_map =
+        GetTplTypeResugarMapForClass(type);
+    ASTNode node(type);
+    node.SetParent(current_ast_node());
+    instantiated_template_visitor_.ScanInstantiatedType(&node, resugar_map);
   }
 
  private:
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2682,8 +2682,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // void fn(const S<Class>& s) { // Forward declarations are sufficient here.
   //   (void)s.t; // Full 'Class' type is needed due to template instantiation.
   // }
-  void ReportTplSpecComponentTypes(const TemplateSpecializationType*) {
-  }
+  void ReportTplSpecComponentTypes(const TemplateSpecializationType*) = delete;
 
   // Do not add any variables here!  If you do, they will not be shared
   // between the normal iwyu ast visitor and the
@@ -3234,6 +3233,12 @@ class InstantiatedTemplateVisitor
     CHECK_(actual_type && "If !CanIgnoreType(), we should be resugar-able");
     ReportTypeUse(caller_loc(), actual_type);
     return Base::VisitCXXConstructExpr(expr);
+  }
+
+  // --- Handler declared in IwyuBaseASTVisitor.
+
+  void ReportTplSpecComponentTypes(const TemplateSpecializationType* type) {
+    TraverseDataAndTypeMembersOfClassHelper(type);
   }
 
  private:
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -4037,32 +4037,6 @@ class IwyuAstConsumer
     return Base::VisitDeclRefExpr(expr);
   }
 
-  // This Expr is for sizeof(), alignof() and similar.  The compiler
-  // fully instantiates a template class before taking the size of it.
-  // So so do we.
-  bool VisitUnaryExprOrTypeTraitExpr(clang::UnaryExprOrTypeTraitExpr* expr) {
-    if (CanIgnoreCurrentASTNode())  return true;
-
-    const Type* arg_type = expr->getTypeOfArgument().getTypePtr();
-
-    // Calling sizeof on a reference-to-X is the same as calling it on X.
-    if (const auto* reftype = arg_type->getAs<ReferenceType>()) {
-      arg_type = reftype->getPointeeTypeAsWritten().getTypePtr();
-    }
-
-    if (const auto* arg_tmpl = arg_type->getAs<TemplateSpecializationType>()) {
-      // Special case: We are instantiating the type in the context of an
-      // expression. Need to push the type to the AST stack explicitly.
-      ASTNode node(arg_tmpl);
-      node.SetParent(current_ast_node());
-
-      instantiated_template_visitor_.ScanInstantiatedType(
-          &node, GetTplTypeResugarMapForClass(arg_type));
-    }
-
-    return Base::VisitUnaryExprOrTypeTraitExpr(expr);
-  }
-
   bool VisitConceptSpecializationExpr(ConceptSpecializationExpr* expr) {
     if (CanIgnoreCurrentASTNode())
       return true;
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -547,12 +547,35 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
   }
 
   // --- Methods on RecursiveASTVisitor
+  bool TraverseType(QualType type) {
+    TraverseArgumentsOfSugaredTemplates(type);
+    return Base::TraverseType(type);
+  }
+
+  bool TraverseTypeLoc(TypeLoc type_loc) {
+    TraverseArgumentsOfSugaredTemplates(type_loc.getType());
+    return Base::TraverseTypeLoc(type_loc);
+  }
+
   bool VisitType(Type* type) {
     seen_types_.insert(type);
     return true;
   }
 
  private:
+  // Clang doesn't traverse underlying type for most of the sugar types.
+  // If a TemplateSpecializationType occurs in a sugaring chain, traverse its
+  // arguments explicitly, otherwise they aren't put into resugar_map.
+  void TraverseArgumentsOfSugaredTemplates(QualType type) {
+    if (type.isNull())
+      return;
+
+    if (const auto* template_spec = type->getAs<TemplateSpecializationType>()) {
+      for (const TemplateArgument& arg : template_spec->template_arguments())
+        TraverseTemplateArgument(arg);
+    }
+  }
+
   set<const Type*> seen_types_;
 };
 
--- a/tests/cxx/array.cc
+++ b/tests/cxx/array.cc
@@ -24,8 +24,27 @@ class A {
     // IWYU: IndirectClass is...*indirect.h
     return &(b[i]);
   }
+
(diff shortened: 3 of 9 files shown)


iwyu.cc
```diff
@@ -2669,6 +2673,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     visitor_state_->processed_overload_locs.insert(loc);
   }
 
+  void ReportTemplateSpecTypeInternals(const TemplateSpecializationType*) =
```
bolshakov-a, line 2676: Maybe just `ReportTemplateInstantiation`? Or something else?

iwyu.cc
```diff
@@ -1664,6 +1664,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
                                                                  comment);
       }
     } else {
+      if (const auto* template_spec_type =
+              dyn_cast<TemplateSpecializationType>(Desugar(type))) {
+        this->getDerived().ReportTemplateSpecTypeInternals(template_spec_type);
```
kimgr, line 1669: I see the comment on line 1657 says "we want to be careful to call *our* ReportDeclUse" -- we can't really make `ReportTemplateSpecTypeInternals` uphold that constraint, can we? I wonder why, more specifically, it wants to avoid any subclass overrides...

iwyu.cc
```diff
@@ -2669,6 +2673,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     visitor_state_->processed_overload_locs.insert(loc);
   }
 
+  void ReportTemplateSpecTypeInternals(const TemplateSpecializationType*) {
```
kimgr: It's long, but I think `ReportTplSpecComponentTypes` matches terminology used in other parts of IWYU.

tests/cxx/array.cc
```diff
@@ -24,8 +24,27 @@ class A {
...
+    // IWYU: IndirectTemplate is...*indirect.h
+    // IWYU: IndirectClass is...*indirect.h
+    return &(_t[i]);
+  }
+
   // IWYU: IndirectClass needs a declaration
   IndirectClass *_b;
+  // IWYU: IndirectTemplate needs a declaration
+  // IWYU: IndirectClass needs a declaration
+  IndirectTemplate<IndirectClass> *_t;
```
kimgr, line 47: `_t` is a reserved name, per https://en.cppreference.com/w/cpp/language/identifiers ("[...] identifiers that begin with an underscore are reserved in the global namespace"), so let's pick a better name. Even `t` would work in this context.

Side note: `_b` has the same problem, and if you want to fix that up in a separate commit, that'd be great.

tests/cxx/indirect.h
```diff
@@ -23,5 +23,16 @@ class IndirectClass {
   static int statica;
 };
 
+template <typename T>
+class IndirectTemplate {
+ public:
+  void Method() const {};
```
kimgr, line 29: No semicolon after inline method body, here or below.

And same thing here, if you want to clean up `IndirectClass` by removing them there too, please do so in a separate commit.

tests/cxx/member_expr.cc
```diff
@@ -58,6 +58,65 @@ void ViaMacro(const IndirectClass& ic) {
 }
 
 
+// Test that referencing template specialization members needs full type info
```
kimgr, line 61: I'm having a hard time understanding this comment -- "[...] _all_ other members"? We only report anything for the member that was accessed, right? Or do you mean the template argument type?

iwyu_ast_util.cc
```diff
@@ -549,6 +549,11 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
   // --- Methods on RecursiveASTVisitor
   bool VisitType(Type* type) {
     seen_types_.insert(type);
+
+    if (const auto* template_spec = type->getAs<TemplateSpecializationType>()) {
```
kimgr, line 553: Hmm. `getAs` is a little deceptive here. If I recall correctly, it shaves off sugar until it reaches `TemplateSpecializationType`. So if you have a `TemplateSpecializationType` nested inside several layers of sugar, e.g. `ElaboratedType`, `TypedefType`, I think you might end up traversing the template argument multiple times.

For example (off the top of my head, might not be 100% in line with reality):
```
// Input
typedef class SomeTemplate<int> SomeSpecialization;

// AST:
// TypedefType: typedef class SomeTemplate<int> SomeSpecialization
// ElaboratedType: class SomeTemplate<int>
// TemplateSpecializationType: SomeTemplate<int>
```
In this case, it seems to me `VisitType` would see `TypedefType`, then `ElaboratedType`, then `TemplateSpecializationType`, and `getAs` will return the same node for all three.

Any chance you can provoke something like that and see if it actually happens?

So while `getAs` is often the right thing to do, if the above happens, I think `dyn_cast` is a better alternative, because we only want to trigger for the particular node that _is_ a `TemplateSpecializationType`. But I could definitely be wrong.

tests/cxx/template_args.cc
```diff
@@ -94,6 +98,23 @@ void NestedTemplateArguments() {
   // IWYU: IndirectClass needs a declaration
   Outer<Inner<IndirectClass> >* opi;
   (void)opi;
+
+  // Test that use of template specialization type template argument is not
+  // hidden by any sugar in the AST.
+
+  // IWYU: IndirectClass is...*indirect.h
+  Outer<decltype(StaticTemplateFieldStruct::tpl)> osi;
```
kimgr, line 106: Is `decltype` only used to introduce sugar here? I wonder if there's a lighter-weight mechanism we could use instead, like a plain typedef. I'm still unused to `decltype`, but I see it's been around since C++11, so maybe it's well-established :).

It just makes me a little uncomfortable that this is the only place it's used in IWYU's test suite, so if IWYU handling of `decltype` fails, e.g. due to some AST change in Clang, this is the only place we're going to notice, and it will be confusing.

kimgr: Unrelated note: really nice testcase showing the subtle differences between `Outer<T>`, `Outer<T*>` and `Outer<T>*`!

kimgr (commented):
Very nice, thank you! See some comments inline, but generally this lgtm.

tests/cxx/member_expr.cc
bolshakov-a: Member access constitutes full embracing type use, be it template specialization or just ordinary structure. Full use of template specialization type requires full type info for its components. It could be thought about as a compiler needs member sizes in order to compute the offset in bytes, but the standard and compilers require full type info for all members except functions, even for typedefs:
```c++
struct A;
struct B; // {};
struct C; // { typedef int InternalType; };

template <typename T1, typename T2, typename T3>
struct Template
{
  const T1& a;
  T2 b;
  typedef typename T3::InternalType CTypedef;
};

void fn(const Template<A, B, C>& t) // still compiles
{
  (void)t.a; // doesn't compile due to B and C; fwd-declaring A is still OK
}
```
And, of course, it concerns dependent types because non-dependent type info should already be available at the point of template definition.

iwyu_ast_util.cc
bolshakov-a: The original problem is that `RecursiveASTVisitor` doesn't traverse, e.g., typedef underlying type by default. This is why I had to write these lines of code. For the case of explicitly written template specialization type, it already worked fine and traversed its arguments. Hence, some redundancy actually present here, but I don't think it is critical, because `seen_types` is a `set`, and duplications are automatically removed.

tests/cxx/template_args.cc
bolshakov-a: > I wonder if there's a lighter-weight mechanism we could use instead, like a plain typedef.

I don't want to test it with typedef, because they have a special treatment in IWYU (they may provide underlying type or not).
>It just makes me a little uncomfortable that this is the only place it's used in IWYU's test suite, so if IWYU handling of decltype fails, e.g. due to some AST change in Clang, this is the only place we're going to notice, and it will be confusing.

I intend to commit a test with `decltype` some day:
https://github.com/bolshakov-a/include-what-you-use/commit/8ccd3ed102e0f86102efb1b17ac908aff79b0099#diff-91d0eae50131503e6661637be7ec19becc0ff3d65b0a3c88f5548776e9a63396R25

iwyu_ast_util.cc
kimgr: Ah, so `RecursiveASTVisitor` doesn't reach the `TemplateSpecializationType` from a `TypedefType` or a `UsingType`? Let's add a comment before the `getAs` call to clarify that. You _could_ probably override `TraverseTypedefType` and `TraverseUsingType` to also traverse the underlying type instead, but it seems like a weaker change (TST under other kinds of sugar wouldn't be handled).

One thing I would like, though, is to avoid `Visit` calling `Traverse`. Usually traversal tweaking happens in `TraverseX`, whereas the thing you want to do with the nodes should go in `VisitX`. There's some rationale in this comment: https://github.com/llvm/llvm-project/blob/main/clang/include/clang/AST/RecursiveASTVisitor.h#L87, particularly this part: https://github.com/llvm/llvm-project/blob/main/clang/include/clang/AST/RecursiveASTVisitor.h#L124.

So I would suggest a new `TraverseType` with the special traversal rule, and a comment to explain that it's necessary because some sugar nodes (I've found `TypedefType` and `UsingType`, but there may be more) do not visit the underlying type, and hence the template arguments would never be seen.

tests/cxx/template_args.cc
kimgr: Makes sense. Let's try to get `decltype` tests in place sooner rather than later.

tests/cxx/member_expr.cc
kimgr: But if we had a template;
```
class A;

template<typename T>
struct Template {
    void method() const {}
};

void test(const Template<A>& t) {
   t.method();
}
``` 
that would not require the definition of `A`? It's only if there's a full-use member of template argument type in the template, then?

Just tested locally, and yeah, that appears to be the case. Adding a `T t;` member to `Template` causes it to no longer compile because it can't instantiate the template.

I think I was confused by the "referencing ... members ... needs full type info for all other members" phrasing. Would it be fair to say something like:
```
// Any member expression instantiates the template, check that the specialization requires all
// template type arguments that are used for members (data members, nested types) whether those
// types are part of the member expression or not.
```

(I don't think that's a great comment, but maybe could serve as inspiration :) )

iwyu.cc
bolshakov-a: It is not fully clear to me. It is important for `ReportDeclUse` to call `IwyuBaseAstVisitor::ReportTypeUse` for typedef underlying type, and not to call `InstantiatedTemplateVisitor::ReportTypeUse`, because the latter resugars the underlying type back into the typedef type, and execution goes into an infinite recursion in that case.
But as for `IwyuBaseAstVisitor::ReportTypeUse`, I tried to remove class qualification from its internal `ReportDeclUse` call, have the tests run and figured out that IWYU stops to report template argument type usage if that type is "provided" by template. In fact, there is currently some difference between reportings of type and declaration usages in an instantiated template: type full uses are always reported for non-defaulted type template arguments, and declaration full uses are reported only if the declaration isn't "provided" by template (e.g., if it is placed in the template containing file or in a file directly included in it). I'm not sure whether types should not actually follow the logic for declarations, but maybe IWYU principle forces explicitly written template arguments to be reported in an instantiating file regardless of whether template defining file provides them or not?
Moreover, after aforementioned change, IWYU starts to report "provided" types in the location of a template. It matters when template defining file is included into analysis and, e.g., type declaration is placed _after_ template definition in the same file. But [I'm not sure if reporting a use into the template definition location is actually reasonable](https://github.com/bolshakov-a/include-what-you-use/commit/ef27dfa1f4fe49e9ef7f7d7f5eae0f4b08a83692).
As concerns `ReportTemplateSpecTypeInternals`, I don't expect any problems with infinite resugaring-desugaring or issues with declaration providing, because it should report types _different_ from original template specialization type.

kimgr: I've been looking at this for a good while, and I honestly can't get my head around it, neither your response above, nor the IWYU implementation :).

Avoiding infinite recursion is probably why the cautionary comment is there, so that makes sense.

So based on your analysis, I guess as long as `ReportTemplateSpecTypeInternals` doesn't call `ReportDeclUse` everything will be fine? I guess it's a little hard to maintain that invariant, but if we can put in comments to remind our future selves, that might be the best we can do right now.

bolshakov-a: The worst thing that can happen is some strange reporting of a template specialization type in location of its own definition, I think. I considered that it might be the case for such a code
```cpp
template <typename T>
struct Tpl
{
  typedef int Type;
  typename Tpl<T>::Type i; // senseless qualification, but compilers allow such a code
};
```
But I didn't succeed in obtaining any wrong behavior.
I think that infinite recursion might occur only on typedef underlying type reporting, but the possible recursion loop is already broken at the [line 1591](https://github.com/include-what-you-use/include-what-you-use/pull/1136/files/8c1f66ba57a3d251c01ceb47edca00128239b11d#diff-7ca107c8b836c1765e8b735b74f06833bec48f308a76017886cee1d5736002d9R1591).
Calling subclass' `ReportDeclUse` from this function (`ReportTypeUse`) is quite safe, as tests have shown.

tests/cxx/template_args.cc
bolshakov-a: BTW, I've just found out that `decltype(auto)` and `decltype(expression)` are different clang types of nodes. The former is `clang::AutoType` and the latter is `clang::DecltypeType`. Hence, a test which checks `decltype(auto)` doesn't check `decltype(expr)`, and vice versa.

kimgr: Good catch, that discovery will come in handy.

kimgr (commented):
Just a request for a comment, otherwise looks great! Thanks for taking the time to analyze all my questions.

tests/cxx/member_expr.cc
kimgr: Very informative now, thanks!

iwyu.cc
kimgr: Thank you for getting to the bottom of this. I can't think of a nice 100% safe construct, so let's not worry about this potential problem.

kimgr:
Excellent, thanks!

PR #1137, opened 2022-11-09
add a flag to workaround cyclic include issues

yuvalif:
this is an attemps to workaround this issue:
https://github.com/include-what-you-use/include-what-you-use/issues/424

Signed-off-by: yuval Lifshitz <ylifshit@redhat.com>

### DIFF
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -160,6 +160,9 @@ Print full include list with comments, even if there are no
 .BI \-\-verbose= level
 Set verbosity. At the highest level, this will dump the AST of the source file
 and explain all decisions.
+.TP
+.B \-\-ignore-cycles
+Do not assert on cyclic includes.
 .SH EXIT STATUS
 By default
 .B include-what-you-use
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -112,6 +112,7 @@ static void PrintHelp(const char* extra_msg) {
          "   --regex=<dialect>: use specified regex dialect in IWYU:\n"
          "          llvm:       fast and simple (default)\n"
          "          ecmascript: slower, but more feature-complete\n"
+         "   --ignore_cycles: don't assert on cyclic includes\n"
          "\n"
          "In addition to IWYU-specific options you can specify the following\n"
          "options without -Xiwyu prefix:\n"
@@ -202,7 +203,8 @@ CommandlineFlags::CommandlineFlags()
       cxx17ns(false),
       exit_code_error(EXIT_SUCCESS),
       exit_code_always(EXIT_SUCCESS),
-      regex_dialect(RegexDialect::LLVM) {
+      regex_dialect(RegexDialect::LLVM),
+      ignore_cycles(false) {
   // Always keep Qt .moc includes; its moc compiler does its own IWYU analysis.
   keep.emplace("*.moc");
 }
@@ -228,6 +230,7 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
     {"error_always", optional_argument, nullptr, 'a'},
     {"debug", required_argument, nullptr, 'd'},
     {"regex", required_argument, nullptr, 'r'},
+    {"ignore_cycles", no_argument, nullptr, 'y'},
     {nullptr, 0, nullptr, 0}
   };
   static const char shortopts[] = "v:c:m:d:nr";
@@ -308,6 +311,7 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
           exit(EXIT_FAILURE);
         }
         break;
+      case 'y': ignore_cycles = true; break;
       case -1: return optind;   // means 'no more input'
       default:
         PrintHelp("FATAL ERROR: unknown flag.");
--- a/iwyu_globals.h
+++ b/iwyu_globals.h
@@ -100,6 +100,7 @@ struct CommandlineFlags {
   int exit_code_always;  // Always exit with this exit code.
   set<string> dbg_flags; // Debug flags.
   RegexDialect regex_dialect;  // Dialect for regular expression processing.
+  bool ignore_cycles; // -y: don't assert on cyclic includes
 };
 
 const CommandlineFlags& GlobalFlags();
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -1101,11 +1101,17 @@ void MakeNodeTransitive(IncludePicker::IncludeMap* filename_map,
     }
   }
   if (status == kCalculating) {
-    VERRS(0) << "Cycle in include-mapping:\n";
-    for (const string& node : *node_stack)
-      VERRS(0) << "  " << node << " ->\n";
-    VERRS(0) << "  " << key << "\n";
-    CHECK_UNREACHABLE_("Cycle in include-mapping");  // cycle is a fatal error
+
+    if (GlobalFlags().ignore_cycles) {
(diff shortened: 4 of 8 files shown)


danielh2942:
Unit test?

yuvalif:
> Unit test?

will add.

yuvalif:
@danielh2942 added UTs in 2nd commit

danielh2942 (approved):
Looks good

bolshakov-a:
I haven't read the full conversation in the related issue, but I think the need to turn off the internal assertion suggests that the assertion is wrong... In the case of real cycle, clang isn't able to build an AST, and IWYU analysis just don't start.
There is a deeper problem: in general, it is impossible to unambiguously identify "included units" just by their filenames, because the same header may declare different names and may even include different set of other headers depending on environment (defined macros).
At now, I'd prefer to remove the assertion rather than introduce a new flag.

yuvalif:
> I haven't read the full conversation in the related issue, but I think the need to turn off the internal assertion suggests that the assertion is wrong... In the case of real cycle, clang isn't able to build an AST, and IWYU analysis just doesn't start.

in one of the tests, there is a real cycle, and IWYU detects it and asserts. so, it seems like it does start even if there is a real cycle

> There is a deeper problem: in general, it is impossible to unambiguously identify "included units" just by their filenames, because the same header may declare different names and may even include different sets of other headers depending on environment (defined macros). At now, I'd prefer to remove the assertion rather than introduce a new flag.

bolshakov-a:
You haven't any real cycle because you have guarding macros `INCLUDE_CYCLE_*1_H`. On the second entrance in a header, preprocessor reveals that macro defined, and the header effectively declares and includes nothing in that case. Hence, the second entrance effectively introduces another (empty) "included unit" in my terminology. If you remove the macros, you'll get a real cycle and the clang error.
Moreover, as far as I understand,  `CHECK_UNREACHABLE_` denotes an execution path that should never be really executed. To write a test that it really fires seems very strange...

kimgr:
Hey all, thanks for patch and reviews. A couple of thoughts:

* The assertion checks cycles in _include mappings_, not cycles in _includes_. The former can be triggered by the latter, but possibly by other inputs
* I'm curious what the follow-on effects are from ignoring the cycle like this. An assertion in general can guard from either already-broken invariants, or something that is known to break assumed invariants in some future path. I'm not sure which case this is, but I suspect the latter.
* With an include graph like this: `A <-> B -> C -> D`, won't we risk missing the `B -> C -> D` branch because traversal is aborted when the `A <-> B` cycle is detected?
* If this is the right fix (and unfortunately I suspect it isn't), I agree that it shouldn't be hidden behind a flag

So could we discuss the motivation for the solution, and what downstream problems it might or might not cause? Thanks!

yuvalif:
the original issue was submitted in 2017 and hasn't been addressed since.
there is a real need for running iwyu together with boost, where these false positives are aborting the tool.
therefore I suggested using a flag as a workaround, or stopgap until a better solution is implemented.

kimgr:
I'd prefer to try and get to the bottom of this, instead of adding workarounds that might cause subtle problems down the line. There's some more context here if you want to keep digging: https://github.com/include-what-you-use/include-what-you-use/issues/424#issuecomment-1312820812.

petersteneteg:
I think this should be merged now, as a workaround since no one has been able to fix this for several years, and it is a complete showstopper for many cases. boost, Qt, etc. 

But getting to the bottom if it would of course be good. From looking at this code to work around this my self I was somewhat depressed by the amount of hacks and hard coded google specific workarounds like 
https://github.com/include-what-you-use/include-what-you-use/blob/f9a6325326995e98b5b6f536092d4cd55937c860/iwyu_include_picker.cc#L1098 and the fact that included using "" and <> are treated completely differently, all without any mention in the documentation. 
So I think any attempt on making this work should be encouraged. Even if it is introducing an other hack. This is still much less of an hack then the current one ot just ignore all "internal" headers.

kimgr:
I can only be frank and say that I will not be taking any command-line flag patches to mask this bug.

Let's try and understand what the desired behavior is instead, and make it so.

yuvalif:
closing, given the above comment

kimgr:
@yuvalif @petersteneteg Could you guys see if https://github.com/include-what-you-use/include-what-you-use/pull/1150 does anything to improve the situation in your codebases? Thanks.

yuvalif:
> @yuvalif @petersteneteg Could you guys see if #1150 does anything to improve the situation in your codebases? Thanks.

i'm using your PR:
```
include-what-you-use 0.20 (git:d8d49b3) based on clang version 16.0.0 (https://github.com/llvm/llvm-project.git 98fa95492f3bbd5befdeb36c88a3ac5ef2740b4e)
```

but i can still see the boost issues. once difference is that it continues to run in-spite of the issue:
```
Cycle in include-mapping:                                                                                              
  <boost/preprocessor/iteration/detail/iter/limits/forward1_256.hpp> ->                    
  <boost/preprocessor/iteration/detail/iter/forward1.hpp> ->                                                           
  <boost/mpl/aux_/template_arity.hpp> ->                   
  <boost/preprocessor/iteration/detail/iter/limits/forward1_256.hpp> ->                       
  <boost/preprocessor/iteration/detail/iter/limits/forward1_256.hpp>
/home/ylifshit/projects/include-what-you-use/iwyu_include_picker.cc:1108: Assertion failed: Cycle in include-mapping
```
```
Cycle in include-mapping:                                  
  <boost/preprocessor/iteration/detail/iter/limits/forward1_256.hpp> ->                       
  <boost/preprocessor/iteration/detail/iter/forward1.hpp> ->        
  <boost/mpl/apply_wrap.hpp> ->                                                                                        
  <boost/mpl/apply_wrap.hpp>                                                                                           
/home/ylifshit/projects/include-what-you-use/iwyu_include_picker.cc:1108: Assertion failed: Cycle in include-mapping   
```
```
Cycle in include-mapping:                                                                                                                                                                                                                     
  <boost/preprocessor/iteration/detail/iter/limits/forward2_256.hpp> ->                                                                                                                                                                       
  <boost/preprocessor/iteration/detail/iter/forward2.hpp> ->                                                                                                                                                                                  
  <boost/mpl/aux_/advance_forward.hpp> ->                                                                                                                                                                                                     
  <boost/mpl/aux_/advance_forward.hpp>                                                                                                                                                                                                        
/home/ylifshit/projects/include-what-you-use/iwyu_include_picker.cc:1108: Assertion failed: Cycle in include-mapping    
```

petersteneteg:
@kimgr Hej, it seems my qt issues has been resolved somehow, I done a bunch of updates since this September, newer clang, newer qt, so my qt problems has disappeared. And currently I'm not able to reproduce the issue, so I can't really say what changes fixed it unfortunately.

petersteneteg:
And thanks for a great tool!!!

kimgr:
@yuvalif Thanks! So it is a reverse macro dependency, just with a longer cycle. My simplistic patch only handles X<=>X cycles. But I think I've convinced myself that we can remove the cycle diagnostics here entirely: https://github.com/include-what-you-use/include-what-you-use/issues/424#issuecomment-1336190336 -- let me know what you think?

yuvalif:
> @yuvalif Thanks! So it is a reverse macro dependency, just with a longer cycle. My simplistic patch only handles X<=>X cycles. But I think I've convinced myself that we can remove the cycle diagnostics here entirely: [#424 (comment)](https://github.com/include-what-you-use/include-what-you-use/issues/424#issuecomment-1336190336) -- let me know what you think?

makes perfect sense. thanks.

Issue #1130, opened 2022-11-02
Request: silence IWYU if correct includes


carlosgalvezp:
Hi!

I notice that IWYU is quite noisy when the code has correct includes:

```
(foo.h has correct #includes/fwd-decls)

(foo.cpp has correct #includes/fwd-decls)
```

This clutters our build logs and makes it harder to spot real issues. Is it possible to silence this somehow, so it only prints output on error? We are using `--verbose=3` to be able to see the explanation of why an include is needed (which line in code).

firewave:
Looks like a duplicate of #999.

kimgr:
Closed as duplicate.

Issue #999, opened 2022-01-30
add option to suppress output when all includes are correct


firewave:
Even if everything is fine a message that all includes are correct is being written. If you are running it in an environment like a CI job you are just interested in the suggested. Especially with big projects there's a lot of unnecessary output. So it would be helpful if there were an option which suppresses that message.

alejandro-colomar:
+1  I was using a sed filter, but an option would be nice.  `--quiet` would probably be a good name <https://www.gnu.org/prep/standards/html_node/Option-Table.html>.

Issue #998, opened 2022-01-30
add option to display full includes even if all are correct


firewave:
If you are manually analyzing the includes include-what-you-use could help with identifying why an include is being used. Unfortunately it does show the full include list only when there are suggested changes. It would be great if there were an option to always output that information.

firewave:
This would also help with identifying cases where multiple includes are suggested for the same symbol.

firewave:
And to keep stating the obvious. It would also help with analysis why things are included so you might be able to get rid of those dependencies.

firewave:
Closing as this functionality is already available via `-Xiwyu --update_comments`.

PR #1139, opened 2022-11-13
Add --silence_correct option for iwyu, tests, docs

danielh2942:
this was requested by #999 and #1130, it adds the
ability to suppress correct includes messages, to
reduce the noise of output generated by IWYU

wrote some tests and updated documentation to reflect the change
this nullifies all output for verbose levels <= 3

### DIFF
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -160,6 +160,9 @@ Print full include list with comments, even if there are no
 .BI \-\-verbose= level
 Set verbosity. At the highest level, this will dump the AST of the source file
 and explain all decisions.
+.TP
+.BI \-\-silence_correct
+Files with correct includes and forward declarations produce no output.
 .SH EXIT STATUS
 By default
 .B include-what-you-use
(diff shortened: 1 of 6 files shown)


firewave:
Very cool.

The option name feels a bit off though.

Could you maybe also look into implementing #998? The option could be renamed into something like `--output=` which takes `quiet` (this PR), `full` (my other suggestion) or `normal` (default - the name feels off, too).

firewave:
@danielh2942 
Thanks for pointing that out. Looks good.

It still prints the `should x these lines:` messages though. But it does this regardless of the findings. Maybe these could be suppressed if they contain no information - possibly via the proposed `--output=`.

danielh2942:
I'll have a look into this, I'll have to make sure that none of the affiliated python scripts break by the change in output and stuff so it might take a while.

kimgr:
Sorry to put a damper on this, but I really want to avoid messing with the IWYU output format. since it's being parsed by:

* Our tests
* iwyu_tool.py
* fix_includes.py
* CMake's IWYU mode

While I agree that it's horrendously ugly, forking into multiple output formats now will just create a lot of low-value complexity (more code to maintain, more tests to maintain, more difficult to integrate tools, etc, etc).

I can only recommend to wrap `include-what-you-use` with a script if you want to tweak output format.

danielh2942:
Cool, got it. closing.

firewave:
> While I agree that it's horrendously ugly, forking into multiple output formats now will just create a lot of low-value complexity (more code to maintain, more tests to maintain, more difficult to integrate tools, etc, etc).

In that case it should have a proper parsable output format instead (I prefer JSON but the world might disagree) - which could also be configured by `--output=`.

Issue #1115, opened 2022-09-23
IWYU pragma: export not effective for forward declared enum class


justusranvier:
IWYU adds forward declarations of enum classes to every file that references the enum even if those files already includes a header that has a forward declaration marked with // IWYU pragma: export

kimgr:
I don't think pragma export applies to forward declarations. Not sure how involved it would be to make that happen.

justusranvier:
I found this out because I was testing the develop branch of IWYU with clang-15 and found that it's now adding forward declarations for enum classes in a few hundred files that don't need the forward declarations because they already include the header that is intended for that purpose.

If we can't use the export then we'll need hundreds of no_forward_declare pragmas.

justusranvier:
Actually the feature of forward declaring an enum class the way IWYU is doing it is pretty suspect in the first place.

It's not like forward declaring a regular class where you're only making the name available - the forward declarations include the underlying type, which raises the possibility that it might get changed in one place and now all the forward declarations are wrong.

kimgr:
Sounds like something @bolshakov-a would be interested in. @justusranvier Do you see any pattern to the behavior? Can you reduce it to a freestanding testcase?

bolshakov-a:
Yes, it seems that IWYU don't have an option to export something but `#include`s at now. @justusranvier, the problem is mainly with source files or header files? [I have an idea](https://github.com/include-what-you-use/include-what-you-use/pull/675#issuecomment-1236344461) that forward declarations should not appear in source files at all. Maybe, it could be done though command line switch in order not to break tests...
> It's not like forward declaring a regular class where you're only making the name available - the forward declarations include the underlying type, which raises the possibility that it might get changed in one place and now all the forward declarations are wrong.

But you can forward-declare a regular class, then make it `struct` (or, likely, vice versa) and also have to change all the forward declarations. Or just change its name... My opinion is that the cases when one changes an underlying type are much less frequent as the cases when one adds a new enum item and recompiles many files which could not be recompiled. Maybe, IWYU didn't always report full enum type usages earlier, but there were inconsistencies with it.
I aggree that enum forward declarations could theoretically cause a problem: there are very specific cases when it is possible to link together files with different underlying types of single enum and to get a runtime crash. And I thought to make a command-line flag for this feature. But I think that the benefit outweights possible harm...

justusranvier:
> Can you reduce it to a freestanding testcase?

[...]

PR #1143, opened 2022-11-20
Fix opaque declaration of enum with bool underlying type

bolshakov-a:
Boolean type is internally represented in clang as `_Bool` builtin type. In order to display it correctly taking into account language options, actual printing policy should be used.
The bug was mentioned in this comment: https://github.com/include-what-you-use/include-what-you-use/issues/1115#issuecomment-1314727796


kimgr:
Nice, thanks!

PR #1154, opened 2022-12-15
Use python3 command for installed python scripts.

### LINKED ISSUE #1096: Python3 version of fix_includes.py
Some OSes are no longer shipping with Python2. Would a PR converting `fix_includes.py` to Python3 be welcome? If yes, is there a preferred minimum Python3 version to support?

sethrj:
The `python` command may not exist on a system with Python 3 installed. See [PEP 394](https://peps.python.org/pep-0394/) for a discussion of the commands expected to be installed.

Since Python2 was [officially sunsetted](https://www.python.org/doc/sunset-python-2/) almost three years ago, IWYU should prefer compatibility with newer systems over older ones.

Fixes #1096.


kimgr:
Love it! Could you do the same for all Python shebangs?

sethrj:
Absolutely! I didn't change those because it looks like they're unused in practice: they're testing functions invoked with `python blah.py` in the cmake. I'll update the rest anyway, though.

kimgr:
Thanks! You may as well take credit for the `mapgen/*-py` files too. Or I can fix it up after if you prefer.

kimgr:
Thanks! I squashed the commits into a single one, and touched up the commit message a little.

Issue #1162, opened 2022-12-21
[Help Wanted] IWYU 0.19 assertion failed


storypku:
## Description

Hi, when running iwyu release 0.19 for one of my use cases, it crashed with the following assertion failure:

```
 /home/story/projects/include-what-you-use-0.19/iwyu_location_util.cc:177: Assertion failed: containing_file
hello.sh: line 2:  2526 Aborted                 (core dumped) external/iwyu_prebuilt_pkg/bin/include-what-you-use -Xiwyu --cxx17ns -Xiwyu --error=1 -Xiwyu --max_line_length=127 -Xiwyu --no_fwd_decls -Xiwyu --transitive_includes_only -Xiwyu --mapping_file=bazel/iwyu/mappings/boost.imp -Xiwyu --mapping_file=bazel/iwyu/mappings/cuda.imp -Xiwyu --mapping_file=bazel/iwyu/mappings/eigen.imp -Xiwyu --mapping_file=bazel/iwyu/mappings/glibc.imp -Xiwyu --mapping_file=bazel/iwyu/mappings/googletest.imp -Xiwyu --mapping_file=bazel/iwyu/mappings/pcl.imp -Xiwyu --mapping_file=bazel/iwyu/mappings/qcraft.custom.imp -Xiwyu --mapping_file=bazel/iwyu/mappings/qt.imp 
....
```

Sorry I haven't figure out the minimal case to reproduce the bug. Built with RelWithDebugInfo and run gdb gives:

```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7a67859 in __GI_abort () at abort.c:79
#2  0x0000555555f0aee8 in include_what_you_use::FatalMessageEmitter::~FatalMessageEmitter (this=0x7ffffffeb24f, 
    __in_chrg=<optimized out>) at /home/story/projects/include-what-you-use-0.19/iwyu_port.h:31
#3  0x00005555563566fe in include_what_you_use::IsInHeader (decl=0x55555abcbef0)
    at /home/story/projects/include-what-you-use-0.19/iwyu_location_util.cc:177
[...]

PR #1165, opened 2022-12-29
Do not crash if IsInHeader is called for builtin

kimgr:
In very particular circumstances (see test case), VisitFunctionDecl would see a function definition that was an implicit constructor of a builtin (va_list_tag).

Rather than crashing for implicit code, just say it's not in a header.

This feels a little questionable, because depending on how IsInHeader is used, it might be misleading that it returns false for builtins/implicit code. But I think it makes sense for now.

Fixes issue #1162.


kimgr:
@bolshakov-a Please take a look.

bolshakov-a:
LGTM

Issue #927, opened 2021-06-02
False positive <type_traits> for enable_if_t and remove_reference<>::type


ptomato:
In two cases I get false positives of `<type_traits>` for files that don't use it. Here are minimal examples of both cases. In both cases, it seems to be the header file that's at fault; any implementation file that includes it will get the false positive.

## Example 1

**a.cpp**:
```c++
#include "b.h"

enum class F {
    VAL = 1,
};

struct A {
    F a;

    bool method() {
        return a & F::VAL;
    }
};
```

**b.h**:
```c++
#pragma once
#include <type_traits>

struct B {
    operator bool() { return true; }
};

template <class E>
std::enable_if_t<true, B> operator&(E const&, E const&) {
    return B();
}
```

Command line:
```
$ include-what-you-use a.cpp

a.cpp should add these lines:
#include <type_traits>  // for enable_if_t

a.cpp should remove these lines:

The full include-list for a.cpp:
#include <type_traits>  // for enable_if_t
#include "b.h"          // for operator&
---
```

## Example 2

**a.cpp**:
```c++
#include "b.h"

char a() {
    B<char> b;
    return b.b()[0];
}
```

**b.h**:
```c++
#pragma once
#include <type_traits>

template <typename T>
struct B {
    std::add_pointer_t<T> b();
};
```

Command line:
```
$ include-what-you-use a.cpp

a.cpp should add these lines:
#include <type_traits>  // for remove_reference<>::type

a.cpp should remove these lines:

The full include-list for a.cpp:
#include <type_traits>  // for remove_reference<>::type
#include "b.h"          // for B
---
```

kimgr:
Looks like `operator bool` triggers something in the first example:

* `std::enable_if_t<true, B>` resolves to `B`
* A call through `B::operator bool` is necessary to convert `B` to `bool` for return from `method`
* That call happens through `std::enable_if_t` even if it just happens to be a typedef/using decl for `B`

I think in a smaller, related example this would make more sense:
```
// a.h
#include <string>
typedef std::string MyString;

// t.cc
#include "a.h"
int f() {
    return MyString().size();
} 
```

In the above, it makes sense that the mention of `MyString` and call to `size` refer to `a.h`.

The complication with `enable_if` is that it's supposed to be transparent, so referring to `<type_traits>` looks off. I think `remove_reference` is similar -- it's a typedef that doesn't really want to introduce a new name, just massage types.

Not sure how to detect that more generally...

[...]

Issue #1149, opened 2022-11-26
`<tuple>` suggested for `std::array` - confused by `<map>`


firewave:
```cpp
#include <array>
#include <cstddef>
#include <string>
#include <map> // removing this will get rid of the messages
#include <utility>

template<std::size_t N>
static void f(const std::array<std::pair<std::string, std::string>, N>&)
{
    std::map<int, int> m; // added so <map> is not unused
}
```

```
test.cpp should add these lines:
#include <tuple>    // for array

test.cpp should remove these lines:
- #include <array>  // lines 1-1

The full include-list for test.cpp:
#include <cstddef>  // for size_t
#include <map>      // for map
#include <string>   // for string
#include <tuple>    // for array
#include <utility>  // for pair
---
```

I am using `include-what-you-use 0.18 based on Debian clang version 14.0.6-2`.

bolshakov-a:
I think it would be good if someone writes the symbol-to-header mapping file for the standard library, in addition to implementation-specific private-to-public header mappings. Or, even better, if someone writes a script to obtain it automatically from some standard library synopsis, if it is possible...

Bu11etmagnet:
Somebody has already written such a thing, because GCC is able to print suggestions like "strcmp is defined in <string.h>, did you forget to include <string.h>?"

kimgr:
@Bu11etmagnet Interesting. How comprehensive is it, e.g. does it suggest c++ library includes too? There may be licensing challenges with taking inspiration from GCC, I suppose.

Bu11etmagnet:
Yes, it does: https://godbolt.org/z/vM6b5ndes

kimgr:
Fixed on master as of https://github.com/include-what-you-use/include-what-you-use/commit/ffe6c3fe3a1dac9a4b577f4b07ef378ede933554 (and, as it turns out, before that).a

Jackaed:
I'm still getting suggestions for `#include <tuple> // for array` on 0.22 @kimgr. Are you certain this is fixed?

firewave:
I can confirm that the original example is not showing the issue with either libstdc++ or libc++.

@Jackaed this should be a new issue - best with a self-contained reproducer.

firewave:
> I can confirm that the original example is not showing the issue with either libstdc++ or libc++.

But I have no pre-0.20 to verify that I can actually reproduce the issue...so take it with a grain of salt.

Issue #1110, opened 2022-09-12
iwyu suggest to include std::map for std::regex


MarcoFalke:
Steps to reproduce:

```
$ cat /tmp/1.cpp 
#include <regex>
#include <string>

std::regex Reg(const std::string &r) { return std::regex{r}; }

int main() {}
```
```
$ ./bin/include-what-you-use  /tmp/1.cpp 

/tmp/1.cpp should add these lines:
#include <map>      // for map
#include <sstream>  // for basic_stringbuf<>::int_type, basic_stringbuf<>::p...
#include <vector>   // for vector

/tmp/1.cpp should remove these lines:

The full include-list for /tmp/1.cpp:
#include <map>      // for map
#include <regex>    // for regex, _NFA
#include <sstream>  // for basic_stringbuf<>::int_type, basic_stringbuf<>::p...
#include <string>   // for string
#include <vector>   // for vector
---
```
```
$ ./bin/include-what-you-use  --version 
include-what-you-use 0.18 (git:abd5d2b) based on Ubuntu clang version 14.0.0-1ubuntu1

**Comment by kimgr** [comment]
Interesting. Does it behave the same if you:

* only default-construct std::regex
* construct std::regex with a pattern, but don't copy it

?

We have some issues with template symbols being attributed to instatiation location, so I'm looking for a pattern.

Also, is this with libc++ or libstdc++?

**Comment by MarcoFalke** [comment]
Yeah, this was with libstdc++. `./bin/include-what-you-use -stdlib=libstdc++ /tmp/1.cpp` gives the same result. libc++ gives:


```
$ ./bin/include-what-you-use -stdlib=libc++ /tmp/1.cpp 

/tmp/1.cpp should add these lines:
#include <iosfwd>  // for string

/tmp/1.cpp should remove these lines:
- #include <string>  // lines 2-2

The full include-list for /tmp/1.cpp:
#include <iosfwd>  // for string
#include <regex>   // for regex
---
```

Switching to default-construction is fine:

```
$ ./bin/include-what-you-use -stdlib=libstdc++ /tmp/1.cpp 

(/tmp/1.cpp has correct #includes/fwd-decls)
```

Removing the copy keeps the bug:

```
$ cat /tmp/1.cpp && ./bin/include-what-you-use -stdlib=libstdc++ /tmp/1.cpp 
#include <regex>
#include <string>
int main() {std::regex{std::string{}};}

/tmp/1.cpp should add these lines:
[...]

Issue #1004, opened 2022-02-03
False Positive: include not needed but reported when using templates


carlosgalvezp:
Hi,

The following minimal example shows the issue:

```cpp
// foo.h
#include <array>

template <typename T>
struct Foo
{
    std::array<T, 2> x_{};
    int sum()
    {
        int v = 0;
        for (T x : x_)
        {
            v += x;
        }
        return v;
    }
};
```

```cpp
// main.cpp
#include "foo.h"

int main()
{
    Foo<int> f;
    int x = f.sum();
}
```

Gives:
```
include-what-you-use -Xiwyu --verbose=3 main.cpp 
main.cpp:6:13: warning: std::array is defined in <array>, which isn't directly #included.

main.cpp should add these lines:
#include <array>  // for array
```

The warning goes away if `Foo` is a regular class instead of a class template.

kimgr:
Fixed on master as of https://github.com/include-what-you-use/include-what-you-use/commit/ffe6c3fe3a1dac9a4b577f4b07ef378ede933554.

PR #1167, opened 2023-01-02
Maybe? Visit fewer node types in instantiated template visitor

kimgr:
A number of Visit... methods have been implemented in IwyuBaseASTVisitor over the years without any particular rationale.

This means they affect traversal and use analysis for both template instantiations (InstantiatedTemplateVisitor) and regular IWYU logic (IwyuAstConsumer).

Over time, we've seen several instances of confusing attribution with templates (particularly with the standard library), and it turns out this is the reason.

This patch moves CXXMethodDecl, BinaryOperator, CXXCatchStmt and CXXForRangeStmt down to IwyuAstConsumer, which means they don't see through template instantiations any more.

Fixes issues: #927, #944, #1003, #1004, #1110, #1149

format


kimgr:
@bolshakov-a If you have time and interest, please take a look at this. Pretty big impact for such a dumb change :-)

I can't quite reason about which nodes we should visit in `IwyuBaseASTVisitor` vs `InstantiatedTemplateVisitor` vs `IwyuAstConsumer` and why. We still do `VisitCXXConstructExpr` in the base visitor, which triggers e.g. https://github.com/include-what-you-use/include-what-you-use/issues/908.

I see you have a more localized fix for #1004, so I figure you might have built a more complete mental model of the visitor roles.

bolshakov-a:
Hi, Kim! Thanks for calling!
I can't check your branch right now because I don't yet have actual LLVM revision checked out and built, but your changes don't seem to be correct, at least fully. `InstantiatedTemplateVisitor` reports actual template argument types when they are fully used in the instantiated template, `IwyuAstConsumer` defines `Visit...` methods for nodes which cannot occur in a template, and `IwyuBaseASTVisitor` checks nodes which can occur both in non-template context and in templates and may cause full template argument type use.
I may be wrong, of course. You could test a sample like this:
```cpp
#include "Direct.h"

template <typename T>
void fn() {
  try {
    T* t = 0;
    (void)(t + 1);
  }
  catch (const T&) { }
}

int main() {
  fn<Indirect>();
}
```
Full `Indirect` struct/class use should be reported for both pointer arithmetic and exception catching.

kimgr:
@bolshakov-a Thanks! That makes sense (unfortunately :)). 

I wonder if it's a simple question of author-intent here too: if the template includes e.g. `<map>` for its implementation, we shouldn't report it in instantiation. Tricky.

bolshakov-a:
>  if the template includes e.g. <map> for its implementation, we shouldn't report it in instantiation.

Yes, it should work due to [that code](https://github.com/include-what-you-use/include-what-you-use/blob/0.19/iwyu.cc#L2882), as long as some `std::map` specialization isn't an explicit template argument.

kimgr:
Thanks, I hadn't understood that before. That `GetLocOfTemplateThatProvides` is only called from `ReportDeclUse` and not `ReportTypeUse` -- all the functions I moved from base to derived visitor use `ReportTypeUse`, so that's why we're seeing such an improvement.

I experimented with creating a `GetLocForTemplateThatProvides` overload for `Type*`, but that didn't really work out (suddenly every type is attributed to the instantiator.

Back to the drawing board, I guess.

bolshakov-a:
I can remember that [I've already written](https://github.com/include-what-you-use/include-what-you-use/pull/1136#discussion_r1038812918) about the difference between type and declaration reporting for instantiated templates and experimented with blocking of type use reports for types which are "provided by template".

kimgr:
This approach has no future. Closing.

PR #1173, opened 2023-01-07
Fwd decls for explicit types only

bolshakov-a:
There is no point in reporting forward-declaration of a type when just its pointer is used and the type is not explicitly written in the source. Explicitly written pointer types are handled, e.g., in `VisitTemplateSpecializationType` and `VisitTagType`.

Similar to #1079.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1653,27 +1653,25 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (type->getAs<EnumType>())
       return;
 
+    // Types in fwd-decl-context should be ignored here and reported from more
+    // specialized places, i.e. when they are explicitly written. But in fact,
+    // this check is redundant because TypeToDeclAsWritten returns nullptr for
+    // pointers and references.
+    if (IsPointerOrReferenceAsWritten(type))
+      return;
+
     // Map private types like __normal_iterator to their public counterpart.
     type = MapPrivateTypeToPublicType(type);
     // For the below, we want to be careful to call *our*
     // ReportDeclUse(), not any of the ones in subclasses.
-    if (IsPointerOrReferenceAsWritten(type)) {
-      type = RemovePointersAndReferencesAsWritten(type);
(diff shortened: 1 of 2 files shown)


iwyu.cc
```diff
@@ -1657,23 +1657,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     type = MapPrivateTypeToPublicType(type);
     // For the below, we want to be careful to call *our*
     // ReportDeclUse(), not any of the ones in subclasses.
-    if (IsPointerOrReferenceAsWritten(type)) {
```
bolshakov-a, line 1660: Maybe, leave here
```c++
if (IsPointerOrReferenceAsWritten(type))
  return;
```
just for clarity and robustness? Or redundant code should be avoided? @kimgr, what do you think?

kimgr: My preference is for an assertion instead, so we can learn when the assumption is broken: `CHECK_(!IsPointerOrReferenceAsWritten(type));  // this should already be handled in ...`

kimgr: Or wait, there is no separate handling?

kimgr: Are you saying we never get here for pointer types, or that the casting below somehow ignores them?

bolshakov-a: Pointers may get here, as in the cases of indexing of array of pointers, or taking `sizeof` variable of pointer type, but they should be ignored.

bolshakov-a: Yes, `TypeToDeclAsWritten` returns `nullptr` for them, at least at now.

kimgr: Yeah, I think it makes sense to short-circuit for pointer/references then, with the if-check you suggested.

But it would be nice to spell out in a comment why it's unnecessary, and who will cover the reporting instead (if anyone). It wouldn't surprise me if there's some construct we don't have test coverage for that will be missed now that this is removed.

kimgr:
Thanks! I think the changes are related and small enough that I'll squash the commits when merging.

PR #1179, opened 2023-01-07
Never report ignored types

kimgr:
@bolshakov-a Your `CanIgnoreType` call stuck in the back of my head, and I eventually got around to generalizing it. This seems to work well (as in: no tests fail).

If the approach makes sense to you, I'll try and cook up some test cases to lock this down.

Should fix: #1004 and #1110. For some reason all the other related problems no longer reproduce on master.

Let me know what you think.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1652,8 +1652,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // with the warning message that iwyu emits.
   virtual void ReportTypeUse(SourceLocation used_loc, const Type* type,
                              const char* comment = nullptr) {
-    // TODO(csilvers): figure out if/when calling CanIgnoreType() is correct.
-    if (!type)
+    if (CanIgnoreType(type))
       return;
 
     // Enum type uses can be ignored. Their size is known (either implicitly
@@ -2999,6 +2998,9 @@ class InstantiatedTemplateVisitor
     // clang desugars template types, so Foo<MyTypedef>() gets turned
     // into Foo<UnderlyingType>().  Try to convert back.
     type = ResugarType(type);
+    if (CanIgnoreType(type))
+      return;
+
     for (CacheStoringScope* storer : cache_storers_)
       storer->NoteReportedType(type);
     Base::ReportTypeUse(caller_loc(), type, comment);
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2900,6 +2900,9 @@ class InstantiatedTemplateVisitor
 
   bool CanIgnoreType(const Type* type, IgnoreKind ignore_kind =
                                            IgnoreKind::ForUse) const override {
+    if (!type)
+      return true;
+
     if (!IsTypeInteresting(type))
       return true;
 
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2104,9 +2104,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     //    typedef Foo* FooPtr; ... static_cast<FooPtr>(...) ...
     for (const Type* type : required_full_types) {
       const Type* deref_type = RemovePointersAndReferences(type);
-      if (CanIgnoreType(deref_type))
-        continue;
-
       ReportTypeUse(CurrentLoc(), deref_type);
     }
 
@@ -2153,10 +2150,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreCurrentASTNode())
       return true;
 
-    const Type* element_type = GetTypeOf(expr);
-    if (CanIgnoreType(element_type))
-      return true;
-    ReportTypeUse(CurrentLoc(), element_type);
+    ReportTypeUse(CurrentLoc(), GetTypeOf(expr));
     return true;
   }
 
@@ -2178,8 +2172,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         // It's a pointer-typed expression. Get the pointed-to type (which may
         // itself be a pointer) and report it.
         const Type* deref_type = pointer_type->getPointeeType().getTypePtr();
-        if (!CanIgnoreType(deref_type))
-          ReportTypeUse(CurrentLoc(), deref_type);
+        ReportTypeUse(CurrentLoc(), deref_type);
       }
     }
 
@@ -2200,10 +2193,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       clang::QualType qual_type = arg->getType();
       const Type* type = qual_type.getTypePtr();
       const Type* deref_type = RemovePointersAndReferencesAsWritten(type);
-
-      if (!CanIgnoreType(deref_type)) {
-        ReportTypeUse(CurrentLoc(), deref_type);
-      }
+      ReportTypeUse(CurrentLoc(), deref_type);
     }
 
     return true;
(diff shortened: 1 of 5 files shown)


bolshakov-a:
I also thought about placing `CanIgnoreType` into `ReportTypeUse`, maybe in the superclass implementation. But at now, the problem with this approach is that at least this case breaks:
```c++
#include "Direct.h"

template <typename T>
struct Tpl {
  void fn() {
    (void)t->i;
  }
  T& t;
};

typedef Indirect* IndirectPtr;

int main() 
  Tpl<IndirectPtr>* t;
  t->fn();
}
```
Type to be reported don't always occur to be in `resugar_map`, and hence checks like `CanIgnoreType(param_type)` _and_ `CanIgnoreType(deref_param_type)` facilitate not to erroneously filter it out. It seems to me that `resugar_map` should be rethinked at first...

bolshakov-a:
I think it should be sufficient just to add components of complex sugared type into `resugar_map`, not only template arguments, as it was done in #1136. But correct handling of typedefs requires adding only the types not "provided" by typedef, AFAIU, and I intend to do this in the future.

bolshakov-a:
It LGTM now. But, maybe, add `CanIgnoreType` check into `IwyuBaseAstVisitor::ReportTypeUse` and remove it from all the `Visit*` methods?

kimgr:
@bolshakov-a Remind me, all the test cases you provided to falsify my earlier attempts are on mainline now, right?

I think I'll spend some time writing up some new testcases for these "template-spill" scenarios based on the examples in associated bugs.

bolshakov-a:
Yes, if you rebase this branch onto f9f4fd2, they should become failed, but passed after rebase onto the current master.

kimgr:
@bolshakov-a This branch rebased on master fails two of your newly added testcases:

```diff --git a/tests/cxx/template_args.cc b/tests/cxx/template_args.cc
index 8a160a7..eb38c28 100644
--- a/tests/cxx/template_args.cc
+++ b/tests/cxx/template_args.cc
@@ -139,7 +139,6 @@ void NestedTemplateArguments() {
   // IWYU: TplInI1 is...*-i1.h
   // IWYU: IndirectClass is...*indirect.h
   Outer<decltype(StaticTemplateFieldStruct::aliasedNonProviding)> oanpi;
-  // IWYU: TplInI1 is...*-i1.h
   // IWYU: IndirectClass is...*indirect.h
   (void)oanpi.t;
 
@@ -147,7 +146,6 @@ void NestedTemplateArguments() {
   (void)oanpip.t;
 
   Outer<decltype(StaticTemplateFieldStruct::aliasedNonProviding)>* opanpi;
-  // IWYU: TplInI1 is...*-i1.h
   // IWYU: IndirectClass is...*indirect.h
   (void)opanpi->t;
 }
```

I think that's consistent with the change I had to make to `explicit_instantiation2.cc` in this patch, but I can't really convince myself that it's right.

Adding `CanIgnoreType` in the base visitor causes a series of test failures, so I don't think the world is ready for that yet.

bolshakov-a:
Debugging shows an inconsistency between types stored in the template instantiation cache and in `resugar_map_`. Seems like `Desugar` calls in `IsKnownTemplateParam` and `ResugarType` functions should be replaced by `GetCanonicalType` calls.
> Adding CanIgnoreType in the base visitor causes a series of test failures

That's strange. `CanIgnoreTypeType` for `IwyuAstConsumer` is just `nullptr` checking.

kimgr:
> >   Adding CanIgnoreType in the base visitor causes a series of test failures
>
> That's strange. CanIgnoreTypeType for IwyuAstConsumer is just nullptr checking.

Ah, I see what's going on. `InstantiatedTemplateVisitor::ReportTypeUse` first checks `CanIgnoreType` for the incoming type, then resugars it and calls `Base::ReportTypeUse`. Adding a `CanIgnoreType` call in `IwyuBaseAstVisitor::ReportTypeUse` will just have it call back down into the derived visitor for the sugared type, which apparently ignores more than it should.

bolshakov-a:
It could be avoided with moving `IwyuAstConsumer` implementation of `CanIgnoreType` into `IwyuBaseAstVisitor` and calling exactly it in `IwyuBaseAstVisitor::ReportTypeUse`...

kimgr:
@bolshakov-a That would boil down to `if (!type) return;`, which `ReportTypeUse` already does :). But yeah, it would be nice to be able to rely on `CanIgnoreType` to be the one point of extension, maybe.

I'm trying to get my head around the other thing you mentioned about instantiation cache vs resugar map, I think that may help unlock some more peculiarities.

On the other hand, this PR as-is fixes all known `template-spill` bugs (https://github.com/include-what-you-use/include-what-you-use/labels/template-spill), so I'm tempted to just take the slight loss in diagnostics and look into that separately.

bolshakov-a:
> Seems like Desugar calls in IsKnownTemplateParam and ResugarType functions should be replaced by GetCanonicalType calls.

But beware that the current `GetCanonicalType` implementation doesn't handle `nullptr`, in contrast to `Desugar`.

And two more little changes should be made: place a "do-nothing" implementation of `ReportDeclForwardDeclareUse` into `InstantiatedTemplateVisitor` (instantiated template traversal should not emit fwd-decls) and remove the first branch of the condition in `InstantiatedTemplateVisitor::ReportDeclUse` ([I already did it](https://github.com/bolshakov-a/include-what-you-use/commit/ef27dfa1f4fe49e9ef7f7d7f5eae0f4b08a83692), and the tests had passed on that commit). And 3 tests remain still failed...

bolshakov-a:
> I'm tempted to just take the slight loss in diagnostics and look into that separately.

The problem with re-reporting cached template specialization types looks like regression. I would prefer to spend some more time to obtain correct solution...

bolshakov-a:
Ah, `IsDefaultTemplateParameter` also needs replacement of `Desugar` to `GetCanonicalType`. And now, 1 test (`explicit_instantiation2`) remains failed...

bolshakov-a:
Take [one more commit](https://github.com/bolshakov-a/include-what-you-use/commit/f4f17c93ff3d0b831a915f174be3eb2359d1b18f) (which fixes `CanForwardDeclareType` for instantiated template traversal) and you should get all the tests passed.

kimgr:
Heh, I'm entirely out of my depth now -- maybe you would like to take this on instead? It seems to me this branch is mostly pointless compared to the lead-up work you've suggested. Would you mind creating PRs for the individual improvements? Or are they interconnected?

kimgr:
@bolshakov-a This is the best I could come up with: https://github.com/include-what-you-use/include-what-you-use/commit/25ecd4f2935b9ab9b04d349dad7b1ccc24d02b18. Many of the bugs have already gone silent, probably after your recent resugaring fixes!

bolshakov-a:
Please rebase.

kimgr:
Threw in the new testcase here too, so we can see if it gets fixed.

kimgr:
@bolshakov-a Very nice! I worked through this now, and:

* reverted all test changes (they were no longer necessary/correct after your recent fixes)
* added `CanIgnoreType` to base `ReportTypeUse` (worked without issue, as you predicted)
* removed all now-redundant `CanIgnoreType` calls
* cleaned up the branch for merge

I've run this on test inputs provided in all bugs marked https://github.com/include-what-you-use/include-what-you-use/labels/template-spill (including the counter-examples you've provided in various places) and they all run as expected.

Please take a look and let me know what you think.

bolshakov-a:
Double calling of `CanIgnoreType` in `InstantiatedTemplateVisitor::ReportTypeUse` and then in `IwyuBaseAstVisitor::ReportTypeUse` seems confusing a little, after all. But I hope it doesn't lead to any problems. So, looks overall good!

kimgr:
Ah, thanks, that raises an interesting point: the ignore-check should probably happen on the resugared type. I tried that locally, and all tests still pass, so I'm going to go with that.

It also makes sense to me not to store ignored types in the cache, otherwise it could be removed entirely in `InstantiatedTemplateVisitor::ReportTypeUse`.

PR #1180, opened 2023-01-08
Report only explicit typedefs

bolshakov-a:
Reporting typedef declarations doesn't make sense when typedef type isn't explicitly written (similar to #1079 and #1173).

Previously, `ReportDeclUse` was called for `TypedefDecl` when `TypedefType` is passed to `ReportTypeUse`. This change short-circuits that path. But an underlying type should still be reported if the typedef doesn't "provide" it. Hence, corresponding logic is moved into `ReportTypeUse`. So `ReportDeclUse` reports typedef declarations and `ReportTypeUse` reports typedef underlying types now.


iwyu.cc
```diff
@@ -4123,25 +4125,26 @@ class IwyuAstConsumer
...
+    if (!CanForwardDeclareType(current_ast_node()))
+      ReportTypeUse(CurrentLoc(), type);
     return Base::VisitTypedefType(type);
   }
 
   bool VisitUsingType(clang::UsingType* type) {
     if (CanIgnoreCurrentASTNode())
       return true;
 
-    // UsingType is similar to TypedefType, so treat it the same.
```
kimgr, line 4138: This makes me wonder why using-declarations do not have the same semantics as typedefs with regards to author intent. E.g.

```
// vended_aliases.h
class Foo;
typedef Foo foo_t;
using foo_u = Foo;

// provided_aliases.h
#include "bar.h"
typedef Bar bar_t;
using bar_u = Bar;

// main.cc
#include "vended_aliases.h"
#include "provided_aliases.h"

void test() {
    foo_t a;  // requires complete Foobar
    foo_u b;  // does not?   

    bar_t c; // provided by header, right?
    bar_u d; // not sure what the story is here
}
```

iwyu.cc
```diff
@@ -4123,25 +4125,26 @@ class IwyuAstConsumer
     if (CanIgnoreCurrentASTNode())
       return true;
     // TypedefType::getDecl() returns the place where the typedef is defined.
-    if (CanForwardDeclareType(current_ast_node())) {
-      ReportDeclForwardDeclareUse(CurrentLoc(), type->getDecl());
-    } else {
-      ReportDeclUse(CurrentLoc(), type->getDecl());
-    }
+    ReportDeclUse(CurrentLoc(), type->getDecl());
```
kimgr, line 4128: The old forward-declaration policy looks a little confused. We would only get here when a `typedef` type was used, right? And if we were in a forward-declare context, we would allow the `typedef` _declaration_ to be forward-declared (which essentially means duplicating it)?

On the other hand, I don't really understand why it's OK to omit `ReportTypeUse` if it's a forward-decl use? Ah, because if it's not a full use, this is just some implicit use of the type that we can ignore? Such as:
```
// blah.h
#include "some_typedef.h"
some_typedef_t* foo();

// main.cc
void bar() {
   (void)sizeof(foo());
}
```
?

iwyu.cc
```diff
@@ -4123,25 +4125,26 @@ class IwyuAstConsumer
...
 
-    // UsingType is similar to TypedefType, so treat it the same.
     if (CanForwardDeclareType(current_ast_node())) {
       ReportDeclForwardDeclareUse(CurrentLoc(), type->getFoundDecl());
     } else {
       ReportDeclUse(CurrentLoc(), type->getFoundDecl());
     }
 
+    const Type* underlying_type = type->getUnderlyingType().getTypePtr();
+    if (isa<TypedefType>(underlying_type))
```
kimgr, line 4145: The special handling of `TypedefType` here is a little surprising. Don't you want to take `CanForwardDeclareType` into account here as well?

kimgr (commented):
Very nice! I'm a little frustrated by the new asymmetry between `TypedefType` and `UsingType`, but maybe that's because they're already slightly different?

iwyu.cc
kimgr: (trying this out on master now)

kimgr: Surprisingly, all of the declarations in `test` are visited as `TypedefType` now and no `UsingType`s are seen. It's as if Clang has reverted to treating `UsingType` as discardable sugar.

bolshakov-a: It's just a naming mess in C++. `clang::UsingType` is about `using ns::Type;`, not about `using Typedef = Type;`.

iwyu.cc
bolshakov-a: You're almost correct, except that in your case there should not be `TypedefType` in `bar()`'s AST at all, and `ReportTypeUse` is called by `sizeof` node handler. Forward-decl use in `VisitTypedefType` — it's about
`some_typedef_t* p = nullptr;`,
i.e. when typedef type is explicitly written.

iwyu.cc
bolshakov-a: This seems like a bug, thanks! Firstly, I wanted rather to call `VisitTypedefType` from here, but `UsingDecl` reporting should be changed in that case...

iwyu.cc
kimgr: :exploding_head:. Thanks. That makes sense, though the comparison to a typedef seems a little unfair now.

iwyu.cc
kimgr: It's probably dangerous to cross-call between `Visit` functions, you'd presumably have to extract some kind of helper to share code.

iwyu.cc
bolshakov-a: > And if we were in a forward-declare context, we would allow the typedef declaration to be forward-declared (which essentially means duplicating it)?

No, `TypedefDecl` fwd-decl uses are recategorized into full uses (see [`DeclCanBeForwardDeclared` logic](https://github.com/include-what-you-use/include-what-you-use/blob/0.19/iwyu_output.cc#L741)), so there is no difference between `ReportDeclUse` and `ReportDeclForwardDeclareUse` calls for `TypedefDecl`.

tests/cxx/no_implicit_typedef_reporting.cc
```diff
@@ -0,0 +1,34 @@
...
+
+// Tests that IWYU doesn't suggest to include typedef-containing header due to
+// implicit-only use of the typedef, i.e. when it is not explicitly written
+// in a source.
+
+#include "tests/cxx/no_implicit_typedef_reporting-d1.h"
+#include "tests/cxx/no_implicit_typedef_reporting-d2.h"
+
+void Fn() {
+  (void)sizeof(Struct::typedefed);
```
kimgr, line 21: I played with some variations:
```
template<class T>
struct Template {
  T t;
  Struct s;
};

void Fn() {
  // Compile-time expression.
  (void)sizeof(Struct::typedefed);

  // Member expression.
  Struct s;
  int i = s.typedefed + 10;

  // Pointer arithmetic.
  void* p = &s.typedefed + 10;

  // Template, both argument and explicitly used.
  Template<Struct> temp;
  (void)temp.t.typedefed;
  (void)temp.s.typedefed;
}
```

At least the pointer arithmetic example reports a use of `Int` on master, so that could at least be worth covering.

The others seem like good standard examples even if no uses are triggered on master either.

iwyu.cc
kimgr: Ah, OK, so that happens very late in the process, when doing the IWYU analysis. OK, fair enough. I think I see what's going on now.

kimgr (commented):
This looks good to me. Added a little suggestion for more test coverage, but given how generic this is, I can't think of any particular "missed" cases.

tests/cxx/no_implicit_typedef_reporting.cc
bolshakov-a: I've applied your suggestion with some changes. Firstly, it should be noted that a use of a member variable after the dot in a member expression doesn't constitute the full use of its type. I could convert `typedefed` underlying type into a `struct` and name its field (`s.typedefed.field`) to test a use in the member expression, if you insist.
Secondly, binary operators are either builtin operators, as in this case, or user-defined operators which may be declared somewhere else, not necessarily in the class or nearby it. So, a use of such operators doesn't generally constitute the full type use.
In my variant, all the three lines are triggered on master.

kimgr (approved):
Nice, thanks!

tests/cxx/no_implicit_typedef_reporting.cc
kimgr: No, I think these additional examples are perfectly fine, thanks!

PR #1182, opened 2023-01-15
Add sugared type components into resugar_map

bolshakov-a:
Prior to this, if e.g. pointer type was sugared, the pointee type was absent in resugar_map. Sometimes this situation was correctly handled by means of excessive `CanIgnoreType` checks, sometimes was not, as in the added test case.

This should make #1179 able to be merged.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -165,6 +165,22 @@ TemplateSpecializationKind GetTemplateSpecializationKind(const Decl* decl) {
   return clang::TSK_Undeclared;
 }
 
+// Desugars to a non-alias template specialization type, if present in the sugar
+// chain, or to canonical type.
+static const Type* DesugarAliasTypes(const Type* type) {
+  type = Desugar(type);
+
+  if (const auto* typedef_type = dyn_cast<TypedefType>(type)) {
+    const TypedefNameDecl* decl = typedef_type->getDecl();
+    return DesugarAliasTypes(decl->getUnderlyingType().getTypePtr());
+  }
+  if (const auto* tpl_spec_type = dyn_cast<TemplateSpecializationType>(type)) {
+    if (tpl_spec_type->isTypeAlias())
+      return DesugarAliasTypes(tpl_spec_type->getAliasedType().getTypePtr());
+  }
+  return type;
+}
+
 }  // anonymous namespace
 
 //------------------------------------------------------------
@@ -546,32 +562,33 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
 
   // --- Methods on RecursiveASTVisitor
   bool TraverseType(QualType type) {
-    TraverseArgumentsOfSugaredTemplates(type);
-    return Base::TraverseType(type);
+    if (type.isNull())
+      return Base::TraverseType(type);
+    return TraverseTypeHelper(type);
   }
 
   bool TraverseTypeLoc(TypeLoc type_loc) {
-    TraverseArgumentsOfSugaredTemplates(type_loc.getType());
-    return Base::TraverseTypeLoc(type_loc);
-  }
-
-  bool VisitType(Type* type) {
-    seen_types_.insert(type);
-    return true;
+    if (!type_loc)
+      return Base::TraverseTypeLoc(type_loc);
+    return TraverseTypeHelper(type_loc.getType());
   }
 
  private:
-  // Clang doesn't traverse underlying type for most of the sugar types.
-  // If a TemplateSpecializationType occurs in a sugaring chain, traverse its
-  // arguments explicitly, otherwise they aren't put into resugar_map.
-  void TraverseArgumentsOfSugaredTemplates(QualType type) {
-    if (type.isNull())
-      return;
+  bool TraverseTypeHelper(QualType qual_type) {
+    CHECK_(!qual_type.isNull());
 
-    if (const auto* template_spec = type->getAs<TemplateSpecializationType>()) {
-      for (const TemplateArgument& arg : template_spec->template_arguments())
-        TraverseTemplateArgument(arg);
-    }
+    const Type* desugared_until_alias_or_tpl = Desugar(qual_type.getTypePtr());
+    seen_types_.insert(desugared_until_alias_or_tpl);
+
+    const Type* desugared = DesugarAliasTypes(desugared_until_alias_or_tpl);
+
+    // Add desugared type components, if any. Don't add two types with the same
+    // canonical type.
+    // TODO(bolshakov): it doesn't always work properly. For example, both A and
+    // TypedefForA would be inserted in seen_types_ for Tpl<A, TypedefForA>,
+    // leading to indeterminate resugar_map construction.
+    // TODO(bolshakov): check whether typedef components are "provided".
+    return Base::TraverseType(QualType(desugared, 0));
   }
 
   set<const Type*> seen_types_;
--- a/tests/cxx/member_expr.cc
+++ b/tests/cxx/member_expr.cc
@@ -128,6 +128,15 @@ void MemberExprInside() {
   (*t)->Method();
 }
 
+template <typename T>
+void MemberExprWithDerefInside() {
+  T t;
+  auto& r = *t;
+  // A type after dereferencing t should be in the resugaring map in order
+  // to be reported.
+  r.Method();
+}
+
 class IndirectClass;  // IndirectClassPtr doesn't provide IndirectClass.
 typedef IndirectClass* IndirectClassPtr;
 
@@ -139,6 +148,8 @@ void Fn() {
   // IndirectClass is hidden by a pointer and (at least) two levels of sugar.
   // IWYU: IndirectClass is...*indirect.h
   MemberExprInside<ns::IndirectClassPtr>();
+  // IWYU: IndirectClass is...*indirect.h
+  MemberExprWithDerefInside<IndirectClassPtr>();
 }
 
 /**** IWYU_SUMMARY
--- a/tests/cxx/template_args.cc
+++ b/tests/cxx/template_args.cc
@@ -13,6 +13,8 @@
 // function-proto template arguments, for both classes and functions.
 
 #include "tests/cxx/direct.h"
+#include "tests/cxx/template_args-d1.h"
+#include "tests/cxx/template_args-d2.h"
 
 // ---------------------------------------------------------------
 
@@ -83,6 +85,8 @@ template<typename T> struct Inner { T t; };
 struct StaticTemplateFieldStruct {
   // IWYU: IndirectClass needs a declaration
   static Inner<IndirectClass> tpl;
+  static ProvidingAlias<5> aliasedProviding;
+  static NonProvidingAlias<5> aliasedNonProviding;
 };
 
 void NestedTemplateArguments() {
@@ -115,6 +119,37 @@ void NestedTemplateArguments() {
   Outer<decltype(StaticTemplateFieldStruct::tpl)>* opsi;
   // IWYU: IndirectClass is...*indirect.h
   (void)opsi->t;
+
+  // Test obtaining template argument through alias template, typedef and
+  // several other layers of sugar. Underlying type provision status of aliases
+  // should be accounted for.
+
+  // IWYU: IndirectClass is...*indirect.h
+  Outer<decltype(StaticTemplateFieldStruct::aliasedProviding)> oapi;
+  // IWYU: IndirectClass is...*indirect.h
+  (void)oapi.t;
+
+  Outer<decltype(StaticTemplateFieldStruct::aliasedProviding)*> oapip;
+  (void)oapip.t;
+
+  Outer<decltype(StaticTemplateFieldStruct::aliasedProviding)>* opapi;
+  // IWYU: IndirectClass is...*indirect.h
+  (void)opapi->t;
+
+  // IWYU: TplInI1 is...*-i1.h
+  // IWYU: IndirectClass is...*indirect.h
+  Outer<decltype(StaticTemplateFieldStruct::aliasedNonProviding)> oanpi;
+  // IWYU: TplInI1 is...*-i1.h
+  // IWYU: IndirectClass is...*indirect.h
+  (void)oanpi.t;
+
+  Outer<decltype(StaticTemplateFieldStruct::aliasedNonProviding)*> oanpip;
+  (void)oanpip.t;
+
+  Outer<decltype(StaticTemplateFieldStruct::aliasedNonProviding)>* opanpi;
+  // IWYU: TplInI1 is...*-i1.h
+  // IWYU: IndirectClass is...*indirect.h
+  (void)opanpi->t;
 }
 
 // ---------------------------------------------------------------
@@ -175,12 +210,16 @@ void TestResugaringOfTypedefs() {
 
 tests/cxx/template_args.cc should add these lines:
 #include "tests/cxx/indirect.h"
+#include "tests/cxx/template_args-i1.h"
 
 tests/cxx/template_args.cc should remove these lines:
 - #include "tests/cxx/direct.h"  // lines XX-XX
 
 The full include-list for tests/cxx/template_args.cc:
 #include "tests/cxx/indirect.h"  // for IndirectClass
+#include "tests/cxx/template_args-d1.h"  // for ProvidingAlias
+#include "tests/cxx/template_args-d2.h"  // for NonProvidingAlias
+#include "tests/cxx/template_args-i1.h"  // for TplInI1
 template <typename F> struct FunctionStruct;  // lines XX-XX
 
 ***** IWYU_SUMMARY */
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1858,7 +1858,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         continue;
       const Type* deref_param_type =
           RemovePointersAndReferencesAsWritten(param_type);
-      if (CanIgnoreType(param_type) && CanIgnoreType(deref_param_type))
+      if (CanIgnoreType(deref_param_type))
         continue;
 
       // TODO(csilvers): remove this 'if' check when we've resolved the
@@ -2127,7 +2127,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     CHECK_(base_type && "Member's base does not have a type?");
     const Type* deref_base_type  // For myvar->a, base-type will have a *
         = expr->isArrow() ? RemovePointerFromType(base_type) : base_type;
-    if (CanIgnoreType(base_type) && CanIgnoreType(deref_base_type))
+    if (CanIgnoreType(deref_base_type))
       return true;
     if (const TypedefType* typedef_type = DynCastFrom(deref_base_type)) {
       deref_base_type = DesugarDependentTypedef(typedef_type);
@@ -2232,7 +2232,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       const TypeLoc& arg_tl = expr->getArgumentTypeInfo()->getTypeLoc();
       if (const auto* reftype = arg_tl.getTypePtr()->getAs<ReferenceType>()) {
         const Type* dereftype = reftype->getPointeeTypeAsWritten().getTypePtr();
-        if (!CanIgnoreType(reftype) || !CanIgnoreType(dereftype))
+        if (!CanIgnoreType(dereftype))
           ReportTypeUse(GetLocation(&arg_tl), dereftype);
       } else {
         // No need to report on non-ref types, RecursiveASTVisitor will get 'em.
@@ -2281,7 +2281,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // actual type being deleted.
     const Type* delete_ptr_type = GetTypeOf(delete_arg);
     const Type* delete_type = RemovePointerFromType(delete_ptr_type);
-    if (CanIgnoreType(delete_ptr_type) && CanIgnoreType(delete_type))
+    if (CanIgnoreType(delete_type))
       return true;
 
     if (delete_type && !IsPointerOrReferenceAsWritten(delete_type))
--- a/tests/cxx/template_args-d1.h
+++ b/tests/cxx/template_args-d1.h
@@ -1,0 +1,28 @@
+//===--- template_args-d1.h - test input file for iwyu --------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Aliases in this file provide TplInI1 declaration because it is directly
+// included, but don't provide IndirectClass.
+
+#include "tests/cxx/template_args-i1.h"
+
+class IndirectClass;
+
+struct StructInD1 {
+  static TplInI1<IndirectClass> t;
+};
+
+typedef decltype(StructInD1::t) ProvidingTypedef;
+
+namespace ns_in_d1 {
+using ::ProvidingTypedef;
+}
+
+template <int>
+using ProvidingAlias = ns_in_d1::ProvidingTypedef;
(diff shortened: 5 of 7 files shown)


bolshakov-a:
Maybe, additional test cases on previously working (due to removed `CanIgnoreType` calls) functional should be added? They are obviously absent, as the green state of #1179 shows.

bolshakov-a:
This branch also has a bug with `typedef`s (`Desugar` doesn't "see through" them) and should not be merged at now. But during investigation, I discovered that there is a UB _on the master branch_ after clang starts to insert a lot of empty `ElaboratedType` nodes. E.g., for a typedef, there is such a tree:
```
ElaboratedType 0x2585d9d0a50 'IndirectTypedef' sugar
`-TypedefType 0x2585d9d0a20 'IndirectTypedef' sugar
  |-Typedef 0x2585d9d0838 'IndirectTypedef'
  `-ElaboratedType 0x2585d9d0750 'Indirect' sugar
    `-RecordType 0x2585d9bbf10 'struct Indirect'
      `-CXXRecord 0x2585d9bbe60 'Indirect'
```
and `GetComponentsOfType` returns _two_ types (`ElaboratedType` 0x2585d9d0a50 and `TypedefType` 0x2585d9d0a20) referring to a single desugared type `RecordType`. It can be said that it works only by accidence, because `TypedefType` object address occurs to be less than `ElaboratedType`'s one, and the `TypedefType` occurs to be the one that is inserted into `resugar_map`.

bolshakov-a:
It seems like I've obtained a correct solution. But the name `resugar_map` becomes unfair now, because `TypedefType` and `TemplateSpecializationType` are the only sugar that can appear as the map values (in fact, IWYU doesn't need any other).
Maybe, `resugar_map` can and should be replaced by a `set` of types allowed for reporting from instantiated template traversal, not yet sure...

kimgr:
> I discovered that there is a UB on the master branch after clang starts to insert a lot of empty ElaboratedType nodes.

Yikes, good catch! That can probably be fixed quite easily by `Desugar`-ing types before putting the in the `seen_types_` set or resugar map. `Desugar` is intentionally IWYU-friendly and stops at typedefs. Where that's not desired, we can use Clang's more aggressive desugaring (`Type::getUnqualifiedDesugaredType`).

kimgr:
Also, possibly good call on `Desugar` vs. `UsingType` -- I had confused that for `using X = Y;` as you noted earlier. But I can't quite see clearly what happens if we strip it out as part of desugaring. OTOH, if no tests fail, I'm inclined to get rid of it.

iwyu_ast_util.cc
```diff
@@ -165,6 +165,18 @@ TemplateSpecializationKind GetTemplateSpecializationKind(const Decl* decl) {
   return clang::TSK_Undeclared;
 }
 
+const Type* GetNonAliasType(const Type* type) {
```
kimgr, line 168: Isn't this the equivalent of `QualType::getCanonicalType` or some other full desugaring?

bolshakov-a: I don't want to lose `TemplateSpecializationType` (which is a sugar type with just a `RecordType` as the canonical type) along with its arguments.

kimgr: Ah, I missed the special case. Maybe `DesugarAliasTypes` to indicate that it's a form of desugaring after all?

A bit too early for cosmetic review, but I think that could help clarify what's going on.

iwyu_ast_util.cc
```diff
@@ -546,32 +558,32 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
...
-    if (const auto* template_spec = type->getAs<TemplateSpecializationType>()) {
-      for (const TemplateArgument& arg : template_spec->template_arguments())
-        TraverseTemplateArgument(arg);
-    }
+    const Type* const type = qual_type.getTypePtr();
+    const Type* const desugared_until_alias = Desugar(type);
+    seen_types_.insert(desugared_until_alias);
+
+    const Type* const desugared =
+        Desugar(GetNonAliasType(desugared_until_alias));
```
kimgr, line 580: Also a bit curious about this `Desugar` -- once the alias types are gone, do you expect any other sugar to remain? Could be `ElaboratedType` I suppose...

kimgr: And doing a full desugar here using `Type::getUnqualifiedDesugaredType` would strip too much (`TemplateSpecializationType`)?

bolshakov-a: > Also a bit curious about this Desugar -- once the alias types are gone, do you expect any other sugar to remain?

Yes, there may still be a lot of sugar! E.g., for such a code:
```cpp
namespace ns {
  using ::Indirect;
}
ns::Indirect i;
typedef decltype(i) Typedef;
```
clang constructs such a tree:
```
TypedefType 0x18e19b10400 'Typedef' sugar
|-Typedef 0x18e19b10260 'Typedef'
`-DecltypeType 0x18e19b10220 'decltype(i)' sugar
  |-DeclRefExpr 0x18e19b101f8 'ns::Indirect':'struct Indirect' lvalue Var 0x18e19b0c688 'i' 'ns::Indirect':'struct Indirect' non_odr_use_unevaluated
  `-ElaboratedType 0x18e19b0c610 'ns::Indirect' sugar
    `-UsingType 0x18e19b0c5e0 'Indirect' sugar
      |-UsingShadow 0x18e19b0c538 'Indirect'
      `-RecordType 0x18e19afbf10 'struct Indirect'
        `-CXXRecord 0x18e19afbe60 'Indirect'
```
> And doing a full desugar here using Type::getUnqualifiedDesugaredType would strip too much (TemplateSpecializationType)?

Yes, I think so...

bolshakov-a: > Also a bit curious about this Desugar -- once the alias types are gone, do you expect any other sugar to remain?

Of course, this `Desugar` may be moved into `GetNonAliasType`, but I think it would be excessive responsibility for `GetNonAliasType` (and three places of `Desugar` call instead of one).

Btw, `desugared_until_alias` should better be renamed into something like `desugared_until_alias_or_tpl_spec`.

kimgr:
> But the name resugar_map becomes unfair now, because TypedefType and TemplateSpecializationType
> are the only sugar that can appear as the map values (in fact, IWYU doesn't need any other).

I haven't looked into the patch in detail, but I wonder if we're just getting lucky when doing lookups in the resugar map then? Or do we always `Desugar` before lookups?

One alternative might be to build a fully expanded resugar map from every level of desugared type to the resugared type, so that lookups are cheaper. And it might make more sense than the current tiptoeing around types everywhere.

> Maybe, resugar_map can and should be replaced by a set of types allowed for reporting from
> instantiated template traversal, not yet sure...

There's the handling of default-template arguments where the type has a null value, but that should probably be expressed some better way anyway.

iwyu_ast_util.cc
bolshakov-a: I've realized that this version of `GetNonAliasType`, of course, doesn't work as expected for recursive typedefs, like these:
```cpp
typedef Indirect Typedef1;
typedef Typedef1 Typedef2;
```
at least due to `ElaboratedType` between them in the tree. `Desugar` should really be placed into `GetNonAliasType`.

bolshakov-a:
> Or do we always Desugar before lookups?

Yes, you've placed it in `IsKnownTemplateParam` and in `ResugarType`.

bolshakov-a:
I've decided to add tests for the working cases as a separate PR #1185.

bolshakov-a:
I see three aims of `resugar_map`:
1) determine which types could be reported from instantiated template traversal because they are template argument types, and the template instantiator is responsible for providing them;
2) report aliased arguments as aliases, so that `IwyuBaseAstVisitor::ReportTypeUse` could determine whether to report the underlying type or not;
3) keep track of defaulted template arguments for ignoring them if they are "provided by template".

It can be seen that all these aims are for determination whether a type from instantiated template body should be reported. But it could be done with a list of allowed types which could be formed outside `InstantiatedTemplateVisitor`, as currently `resugar_map` is.

The problem with this approach is that a dependent type could be totally different from template arguments. E.g.:
```cpp
template <typename>
struct TplStruct {
};

template <typename T>
void TplFn() {
  (void)sizeof(TplStruct<T>::field);
};

template <>
struct TplStruct<int> {
  static IndirectClass field;
};

void Fn() {
  TplFn<int>();  // IndirectClass full type is needed here, but how to determine this?
}
```
Even the fact that the type under `sizeof` is a dependent type is lost in the specialized template declaration AST!

A solution to this may be to traverse an original template declaration with "reporting" types in a list, and then use this list as a list of types blocked for reporting on template specialization decl traversal. Reports of all other types should then be allowed. Types "provided by" a typedef and "provided" default argument types could be excluded from the list.

But, in my opinion, the more important thing with `resugar_map` is that it potentially allows for separate handling of different template specializations. E.g., for `Tpl1<Tpl2<AliasProvidingA>, A, B>`, it may be undesirable to report `A`, or `B` if it is the default argument of `Tpl2`, during traversal of `Tpl2` used inside of `Tpl1`. It currently doesn't work because all the components fall into a single `resugar_map`, but the analysis can be improved in the future. So, I'm biased to retain `resugar_map`. Maybe, it should not contain components of arguments at all, but they could serve as a temporary solution.

kimgr:
> A solution to this may be to traverse an original template declaration with "reporting" types in a list, and then use this list as a list of types blocked for reporting on template specialization decl traversal. Reports of all other types should then be allowed. Types "provided by" a typedef and "provided" default argument types could be excluded from the list.

I was under the impression that something like this already happens :sweat_smile:. Let me see if I can find what I was thinking of. Right, I was thinking of this comment: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L3402. This is slightly different from what you're describing, I guess?

Or could some typedef/provision logic go into the flattener?

bolshakov-a:
Yes, `nodeset_getter` looks like it was introduced for this purpose, but, actually, it's nearly useless, I think. Because, in general, nodes of an original template declaration aren't reused in a specialized template declaration. I suspect it can be removed.

bolshakov-a:
I tried to remove `nodes_to_ignore_` filling and had a couple of tests failed, though. But, e.g., for such a code:
```cpp
struct S {
  char ch;
};

template<typename T> void TplFn() {
  S s;
  s.ch;
}

void fn() {
  TplFn<int>();
}
```
clang produces such a tree for `TplFn()` original declaration and specialization:
```
FunctionTemplateDecl 0x225fcdfd6d0 <line:5:1, line:8:1> line:5:27 TplFn
|-TemplateTypeParmDecl 0x225fcdfd518 <col:10, col:19> col:19 typename depth 0 index 0 T
|-FunctionDecl 0x225fcdfd628 <col:22, line:8:1> line:5:27 TplFn 'void ()'
| `-CompoundStmt 0x225fce17dc0 <col:35, line:8:1>
|   |-DeclStmt 0x225fce17d58 <line:6:3, col:6>
|   | `-VarDecl 0x225fcdfdc00 <col:3, col:5> col:5 referenced s 'S':'S' callinit
|   |   `-CXXConstructExpr 0x225fce17d30 <col:5> 'S':'S' 'void () noexcept'
|   `-MemberExpr 0x225fce17d90 <line:7:3, col:5> 'char' lvalue .ch 0x225fcdfd4a8
|     `-DeclRefExpr 0x225fce17d70 <col:3> 'S':'S' lvalue Var 0x225fcdfdc00 's' 'S':'S'
`-FunctionDecl 0x225fcdfd9e0 <line:5:22, line:8:1> line:5:27 used TplFn 'void ()'
  |-TemplateArgument type 'int'
  | `-BuiltinType 0x225fcdc7780 'int'
  `-CompoundStmt 0x225fce17ed8 <col:35, line:8:1>
    |-DeclStmt 0x225fce17e70 <line:6:3, col:6>
    | `-VarDecl 0x225fce17de0 <col:3, col:5> col:5 used s 'S':'S' callinit
    |   `-CXXConstructExpr 0x225fce17e48 <col:5> 'S':'S' 'void () noexcept'
    `-MemberExpr 0x225fce17ea8 <line:7:3, col:5> 'char' lvalue .ch 0x225fcdfd4a8
      `-DeclRefExpr 0x225fce17e88 <col:3> 'S':'S' lvalue Var 0x225fce17de0 's' 'S':'S'
```
You can see that, despite the fact that the type of `s` variable isn't a dependent type, AST nodes have different addresses, and the `nodes_to_ignore_` set wouldn't prevent the `S` type from reporting. And yes, it contains a list of types, but only such types which were explicitly written in the source, even without any anlysis if they are in fwd-decl context or not.

iwyu_ast_util.cc
```diff
@@ -546,32 +560,32 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
...
-        TraverseTemplateArgument(arg);
-    }
+    const Type* const type = qual_type.getTypePtr();
+    const Type* const desugared_until_alias_or_tpl = Desugar(type);
+    seen_types_.insert(desugared_until_alias_or_tpl);
+
+    const Type* const desugared =
+        DesugarAliasTypes(desugared_until_alias_or_tpl);
+    // A desugared type shouldn't be inserted into resugar_map along with its
+    // sugared version as a value instead of a key, hence it is important to
```
kimgr, line 584: What does `along with its sugared version as a value instead of a key` mean? I suspect there's a small typo in this comment that makes it hard to understand...?

iwyu_ast_util.cc
```diff
@@ -546,32 +560,32 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
...
+    // call the base class implementation of TraverseType and not the own.
+    // TODO(bolshakov): check whether typedef components are "provided".
```
kimgr, line 586: It would be nice if we didn't have to do that in here -- most of that logic lives in iwyu.cc where we have more of a high-level view.

iwyu_ast_util.cc
```diff
@@ -546,32 +560,32 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
...
+    if (type != desugared)
+      Base::TraverseType(QualType{desugared, 0});
```
kimgr, line 588: Please avoid initialization syntax, I find it's just a distraction -- `QualType(desugared, 0)` means the same in this context.

iwyu_ast_util.cc
bolshakov-a: `resugar_map` maps desugared types to their original sugared versions, obviously. So, desugared (canonical) types are the keys, and (probably) sugared types are the values. `TypeEnumerator` prepares _values_ for the map, hence it should return only single (sugared) version for every canonical type.

iwyu_ast_util.cc
bolshakov-a: Ok, but the "C++ Core Guidelines" [suggest the opposite](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es64-use-the-tenotation-for-construction).

kimgr: This two-step recursion makes my head spin a little. Does this alter the traversal so that for every type we also traverse its fully desugared type? I'm trying to visualize what that looks like, but coming up short. Is there a reasonably compact C++ example to demonstrate what is picked up, that you could put in a doc comment for this function?

I'm a little thrown by the `Base::TraverseType` here, and then another `Base::TraverseType` in the calling `TraverseType` function, but I guess they traverse the desugared and sugared types respectively?

Could this be called `TraverseDesugared` instead?

tests/cxx/template_args.cc
```diff
@@ -115,6 +119,37 @@ void NestedTemplateArguments() {
   Outer<decltype(StaticTemplateFieldStruct::tpl)>* opsi;
   // IWYU: IndirectClass is...*indirect.h
   (void)opsi->t;
+
+  // Test obtaining template argument through alias template, typedef and
```
kimgr, line 123: This test sequence is pretty hard to read as there's so many different mechanisms going on (templates, `decltype` for sugar, providing vs. non-providing template definitions, pointers vs. not), _BUT_ they're ultimately very clear in contrasting the impact of the different mechanisms. Nicely done!

iwyu_ast_util.cc
kimgr: And the type is not collected into `seen_types_` in `VisitType` because... you would have to desugar and only report it if it's different there too? It feels like the type collection would be more suitable for `Visit`, whereas traversal adjustments belong in `Traverse`.

But these are mostly questions, not suggestions.

bolshakov-a: I suspect that `if (type != desugared)` could be removed along with `Base::TraverseType*` calls inside `Base::TraverseType` and `Base::TraverseTypeLoc` methods. I had just faced with infinite recursion during the work, but I hope that situation won't repeat anymore in the last version...

bolshakov-a: If `VisitType` filled in `seen_types_`, then `desugared` would be inserted into `seen_types_` due to `VisitType` call from inside `Base::TraverseType(QualType{desugared, 0});` call, which should be avoided, as I mentioned before.

iwyu_ast_util.cc
bolshakov-a: I don't see at now any convenient way to do this except moving `TypeEnumerator` in iwyu.cc along with all the dependent functions...

iwyu_ast_util.cc
```diff
@@ -546,32 +560,29 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
...
-    if (type.isNull())
-      return;
-
-    if (const auto* template_spec = type->getAs<TemplateSpecializationType>()) {
-      for (const TemplateArgument& arg : template_spec->template_arguments())
-        TraverseTemplateArgument(arg);
-    }
+  bool TraverseTypeHelper(QualType qual_type) {
+    if (qual_type.isNull())
+      return false;
```
bolshakov-a, line 573: I'm not sure about `false`. Maybe, empty type node should not stop the traversal? Or let the base class implementation decide?

iwyu_ast_util.cc
```diff
@@ -165,6 +165,20 @@ TemplateSpecializationKind GetTemplateSpecializationKind(const Decl* decl) {
   return clang::TSK_Undeclared;
 }
 
+const Type* DesugarAliasTypes(const Type* type) {
```
kimgr, line 168: This function can be `static`.

It would be helpful with a comment to describe what the subtle difference is between this and a full clang desugar vs. a normal IWYU Desugar. This basically desugars everything except non-alias template specializations, right?

bolshakov-a: > This function can be static.

What for? It is inside an anonymous namespace, so already has internal linkage.

bolshakov-a: > This basically desugars everything except non-alias template specializations, right?

Yes, it is. I can leave a comment.

iwyu_ast_util.cc
kimgr: I suppose we can move the mechanics for checking back in here when the time comes. Hopefully.

iwyu_ast_util.cc
```diff
@@ -546,32 +560,29 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
...
-
-    if (const auto* template_spec = type->getAs<TemplateSpecializationType>()) {
-      for (const TemplateArgument& arg : template_spec->template_arguments())
-        TraverseTemplateArgument(arg);
-    }
+  bool TraverseTypeHelper(QualType qual_type) {
+    if (qual_type.isNull())
+      return false;
+
+    const Type* const type = qual_type.getTypePtr();
```
kimgr, line 575: No need for the second `const` in `const Type* const`. I prefer to only use that when forced by the standard library :). Shorter lines with fewer tokens are easier to read.

iwyu_ast_util.cc
kimgr: Please see if you can clarify this comment a little. The mention of `resugar_map` here is also a little confusing, since it's implied that it's the only user of `TypeEnumerator`.

Also the mention of base class `TraverseType` below seems a little off now, because if you called `this->TraverseType(...)`, it would basically go into infinite recursion, right?

Let me know if you want help phrasing something.

iwyu_ast_util.cc
kimgr: That's what `iwyu.cc` `TraverseType` does:
```
  bool TraverseType(QualType qualtype) {
    if (qualtype.isNull())
      return Base::TraverseType(qualtype);
```

Whether it's correct or not, I think it would make sense to be consistent.

iwyu_ast_util.cc
kimgr: Ah, I see. The problem with anonymous namespaces is exactly this: you can't see on the declaration whether it's external/internal, you need to find the (usually very distant) enclosing scope. See also https://llvm.org/docs/CodingStandards.html#anonymous-namespaces.

But not a lot of functions in this file are explicitly marked static, so as long as it's internal linkage, I'm happy with it as is.

iwyu_ast_util.cc
bolshakov-a: > Also the mention of base class TraverseType below seems a little off now, because if you called this->TraverseType(...), it would basically go into infinite recursion, right?

Yes, it would do, but avoiding of inserting multiple types with the same canonical type seems also important, as for me (please recall [the UB that I've mentioned](https://github.com/include-what-you-use/include-what-you-use/pull/1182#issuecomment-1384322467)).

I can think about some better wording for a while...

iwyu_ast_util.cc
bolshakov-a: Thanks!

iwyu_ast_util.cc
```diff
@@ -546,32 +562,30 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
...
-      for (const TemplateArgument& arg : template_spec->template_arguments())
-        TraverseTemplateArgument(arg);
-    }
+    const Type* desugared_until_alias_or_tpl = Desugar(qual_type.getTypePtr());
+    seen_types_.insert(desugared_until_alias_or_tpl);
+
+    const Type* desugared = DesugarAliasTypes(desugared_until_alias_or_tpl);
+
+    // Add desugared type components, if any. Don't add two types with the same
+    // canonical type.
```
bolshakov-a, line 586: In fact, it doesn't always occur in this manner. Imagine `Template<A, TypedefForA>`. Maybe, it could be addressed in some future PRs...

iwyu_ast_util.cc
kimgr: Up to you, there's no strong consistency anyway :)

iwyu_ast_util.cc
kimgr: Ah, good catch. Please add a little TODO  so we remember it's a known weakness. It seems to me it should be solvable either by double-storing all seen canonical types, or by brute-force looping over `seen_types_` and comparing.

That small example would be nice to have as part of the TODO comment.

tests/cxx/member_expr.cc
```diff
@@ -128,6 +128,15 @@ void MemberExprInside() {
   (*t)->Method();
 }
 
+template <typename T>
+void MemberExprWithDotInside() {
```
kimgr, line 132: Not critical, but maybe call this `MemberExprWithDerefInside`?

iwyu_ast_util.cc
kimgr: I think the comment as stands now is good, but you have a much deeper understanding of this than I do. Does the UB still exist? If so, should we open a separate issue for it, so it's not forgotten?

kimgr:
Very nice, thanks so much for polishing this so I can understand it better. A few micro-suggestions, but then I think we should merge this. As I understand it, it would make #1179 a viable solution to template misattribution? There's so many bugs that would be fixed with that in place, so I'm anxious to try it out.

iwyu_ast_util.cc
bolshakov-a: Makes sense, of course. Done.

iwyu_ast_util.cc
bolshakov-a: The described UB with more than one type from a single sugar chain inserted into `seen_types_` has been fixed here, I hope. But the issue with sibling types like in `Tpl<A, TypedefForA>` still remains. But I'm not sure that it will be solved in general in some day, because it is unknown in the common case to which template argument should be attributed some `A` usage inside the specialization.

bolshakov-a:
> As I understand it, it would make https://github.com/include-what-you-use/include-what-you-use/pull/1179 a viable solution to template misattribution?

Yes, I think so.

PR #1184, opened 2023-01-22
Desugar UsingType among others

bolshakov-a:
`clang::UsingType` denotes a sugar type which is introduced by a using-declaration (e.g., `using ns::SomeType;`). It has not such a special meaning in IWYU as `TypedefType` or `TemplateSpecializationType`.

No functional change is expected.


kimgr:
Thanks!

PR #1185, opened 2023-01-25
Test reporting types hidden in template arguments

bolshakov-a:
The full type use reporting in the three covered cases is ensured with excessive `CanIgnoreType` calls which check not only the type to be reported but also the type before dereferencing (which occurs to be in the resugaring map). This approach should be refactored.

### DIFF
--- a/tests/cxx/operator_new.cc
+++ b/tests/cxx/operator_new.cc
@@ -43,6 +43,12 @@ void ExpressionsBuiltinTypes() {
 }
 
 // New- and delete-expressions with user-defined types.
+
+template <typename T>
+void TplFnWithDelete(T p) {
+  delete p;
+}
+
 void ExpressionsUserTypes() {
   // IWYU: IndirectClass needs a declaration
   // IWYU: IndirectClass is...*indirect.h
@@ -55,6 +61,11 @@ void ExpressionsUserTypes() {
   IndirectClass* arr = new IndirectClass[4];
   // IWYU: IndirectClass is...*indirect.h
   delete[] arr;
+
+  // Hide the pointer type behind "decltype" sugar.
+  decltype(elem) elem2 = nullptr;
+  // IWYU: IndirectClass is...*indirect.h
+  TplFnWithDelete(elem2);
 }
 
 // Aligned allocation uses operator new(size_t, std::align_val_t) under the
--- a/tests/cxx/sizeof_reference.cc
+++ b/tests/cxx/sizeof_reference.cc
@@ -103,6 +103,22 @@ SizeofTakingStructTplRef2<IndirectClass> sizeof_taking_struct5;
 // IWYU: IndirectClass needs a declaration
 SizeofTakingStructTplRef2<IndirectClass&> sizeof_taking_struct6;
 
+// The same with some sugar.
+
+// IWYU: IndirectClass is...*indirect.h
+SizeofTakingStruct<decltype(ref)> sizeof_taking_struct7;
+
+// IWYU: IndirectClass is...*indirect.h
+SizeofTakingStructRef<decltype(dummy)> sizeof_taking_struct8;
+
+SizeofTakingStructTpl<decltype(ref)> sizeof_taking_struct9;
+
+SizeofTakingStructTplRef<decltype(dummy)> sizeof_taking_struct10;
+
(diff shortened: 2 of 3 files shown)


tests/cxx/sizeof_reference.cc
```diff
@@ -76,19 +76,23 @@ SizeofTakingStruct<IndirectClass&> sizeof_taking_struct1;
 // IWYU: IndirectClass needs a declaration
 SizeofTakingStructRef<IndirectClass> sizeof_taking_struct2;
 
-// Not sure why, but C++ doesn't require full type of IndirectClass here.
```
kimgr, line 79: It's true that its size does not depend on the the type definition, but that's equally true in the naive case:
```
$ cat t.cc
struct Foo;

int main() {
  (void)sizeof(Foo*);
  (void)sizeof(Foo&);
  return 0;
}

$ clang t.cc
t.cc:5:9: error: invalid application of 'sizeof' to an incomplete type 'Foo'
  (void)sizeof(Foo&);
        ^     ~~~~~~
t.cc:1:8: note: forward declaration of 'Foo'
struct Foo;
       ^
1 error generated.
```

So in the general case, the compiler requires the full type to do sizeof(reference), but something happens with templates:
```
$ cat tt.cc
struct Foo;

template<class T>
struct Tpl {
  T v;
};

int main() {
  (void)sizeof(Tpl<Foo&>);
  return 0;
}

$ clang tt.cc
$
```

I think this inconsistency is what the comment is getting at, and I'm not entirely convinced by your explanation :)

tests/cxx/operator_new.cc
```diff
@@ -55,6 +61,10 @@ void ExpressionsUserTypes() {
   IndirectClass* arr = new IndirectClass[4];
   // IWYU: IndirectClass is...*indirect.h
   delete[] arr;
+
+  decltype(elem) elem2 = nullptr;
```
kimgr, line 65: I would add a comment here to explain that the `decltype` is only used to inject sugar

tests/cxx/sizeof_reference.cc
bolshakov-a: It seems for me quite reasonable that `sizeof(Type&)` means the size of `Type` and not the size of a pointer because, e.g., when someone writes:
```cpp
Type obj;
sizeof(obj)
```
he/she wants obviously to obtain the size of a `Type` instance, despite that the value category of `obj` expression is lvalue and its type is `Type&` and not `Type`, AFAIU. I can try to dig into the standard...

bolshakov-a: > When applied to a reference type, the result is the size of the referenced type[.](https://eel.is/c++draft/expr.sizeof#2.sentence-1) When applied to a class, the result is the number of bytes in an object of that class including any padding required for placing objects of that type in an array[.](https://eel.is/c++draft/expr.sizeof#2.sentence-2)

https://eel.is/c++draft/expr.sizeof#ntref:type-id

And, obviously, the number of bytes in an object of `class Class { Type& t; }` depends on the pointer size in the system and not on the `Type` object size.

kimgr: > When applied to a reference type, the result is the size of the [referenced type.](https://eel.is/c++draft/expr.sizeof#2.sentence-1)

Ah, I see the subtle difference now. With my two examples above, the equivalence is between:
```
sizeof(Foo&)
sizeof(Tpl<Foo&>::v);  // note the direct access of the reference member
```
These two cause the same incomplete type error.

Whereas, as your standard ref says, we take the size of the template specialization as a whole, it falls under the "When applied to a class..." clause:
```
sizeof(Tpl<Foo&>);  // the template is fully defined and knows the size of a reference == size of a pointer
```
It would be nice to steal some of the wording from the standard text there to clarify the difference between `sizeof(T&)` and `sizeof(TypeContainingARef)`. I think the original author's instinct was that `sizeof` would apply recursively to record types and therefore a struct containing a references to another struct would be the same as a struct containing a value struct, which as you noted is not the case.

I think the standard text makes that clearer than the reasoning about references as pointers (even if that's how I think about them too).

bolshakov-a: Ok, done.

kimgr (approved):
Nice, thanks!

PR #1186, opened 2023-02-01
[clang compat] Remove unused llvm::Optional

BredPet:
Clang [125f4457a54a550846732763ee36b1447ec8d66e](https://github.com/llvm/llvm-project/commit/125f4457a54a550846732763ee36b1447ec8d66e) changed.
The LLVM project is migrating from llvm::Optional to std::optional.

We would previously see llvm::Optional via an indirect include:
```
    iwyu_preprocessor.h
      iwyu_output.h
        clang/AST/Decl.h
          clang/AST/APValue.h
            clang/Basic/LLVM.h
```

Since we don't use it, remove the using-declaration.

### DIFF
--- a/iwyu_preprocessor.cc
+++ b/iwyu_preprocessor.cc
@@ -44,7 +44,6 @@ using clang::SourceLocation;
 using clang::SourceRange;
 using clang::Token;
 using llvm::errs;
-using llvm::Optional;
 using llvm::StringRef;
 using std::make_pair;
 using std::string;


kimgr:
Thanks! Can I ask for an updated commit message:
```
[clang compat] Remove unused llvm::Optional

The LLVM project is cleaning up its use of llvm::Optional in favor of
std::optional.

We would previously see llvm::Optional via an indirect include, but it
has now been removed. Since we don't use it, remove the
using-declaration.
```
(ideally we'd include the Clang commit SHA that broke the build, but there's a long series of llvm::Optional cleanups in the llvm commit log, and I don't know which one hit us first.)

kimgr:
Beautiful, thanks!

PR #1188, opened 2023-02-07
Never forward-declare expressions used in typeid()

### LINKED ISSUE #1168: iwyu suggests forward declaration when using typeid
Thanks for all your work on iwyu. I can't imagine sorting out my dependency messes without it :)

I've recently seen an issue where iwyu will suggest a forward-declaration instead of a `#include` for `Dependency` here:

```c++
// user/user.cpp
#include <dependency.h>
#include <typeindex>

auto useTypeIndexOfDependency(const Dependency &dependency) -> void {
  std::type_index type = typeid(dependency);
}
```

```
$ iwyu_tool -p build user/user.cpp

/home/matthew/Documents/iwyu/minimal-example/user/user.cpp should add these lines:
class Dependency;

/home/matthew/Documents/iwyu/minimal-example/user/user.cpp should remove these lines:
- #include <dependency.h>  // lines 1-1

The full include-list for /home/matthew/Documents/iwyu/minimal-example/user/user.cpp:
#include <typeindex>  // for type_index
class Dependency;
---
```
If I follow its suggestion, I get 
```
[ 50%] Building CXX object user/CMakeFiles/user.dir/user.cpp.o
/home/matthew/Documents/iwyu/minimal-example/user/user.cpp: In function ‘void useTypeIndexOfDependency(const Dependency&)’:
/home/matthew/Documents/iwyu/minimal-example/user/user.cpp:5:33: error: invalid use of incomplete type ‘class Dependency’
    5 |   std::type_index type = typeid(dependency);
      |                                 ^~~~~~~~~~
/home/matthew/Documents/iwyu/minimal-example/user/user.cpp:2:7: note: forward declaration of ‘class Dependency’
    2 | class Dependency;
      |       ^~~~~~~~~~
make[2]: *** [user/CMakeFiles/user.dir/build.make:82: user/CMakeFiles/user.dir/user.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:134: user/CMakeFiles/user.dir/all] Error 2
make: *** [Makefile:103: all] Error 2
```
I believe this is because you cannot apply `typeid` to an incomplete type.

I tested this with
```
include-what-you-use 0.15 based on Debian clang version 11.0.1-2
```
which is the current version on Debian Stable.

I've uploaded a full minimal example [here](https://github.com/cottonxray/iwyu-typeid-potential-issue).

I would be happy to spend some time to contribute a fix for this, if that would help. Equally, if there's any extra information that I can provide, then let me know.

But I wanted to check:
* do you also think it's an issue, or have I missed something?
* if you'd like me to work on a fix - where would be the best place to make this change in the code?

Thanks again for all your work on this tool.

cottonxray:
The C++ spec requires expressions found in typeid() calls to be fully-formed. Therefore, any necessary classes should not be forward-declared.
This commit adds a new visitor that checks for typeid() calls, and reports the necessary types to be included.

I'm hoping the tests will pass. The changes passed the tests for me on the `clang_11` branch, but I had issues compiling clang on my under-powered laptop to run them against master!

Let me know if there's anything you think can be improved, it's a small patch but I wouldn't be surprised if I've made some rookie errors :)

Hopefully closes #1168

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2549,6 +2549,23 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  // The C++ spec does not allow us to apply typeid to an incomplete type.
+  // Therefore, we must report the type to be included.
+  bool VisitCXXTypeidExpr(clang::CXXTypeidExpr* expr) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+
+    QualType type;
+    if (expr->isTypeOperand()) {
+      type = expr->getTypeOperandSourceInfo()->getType();
+    } else {
+      type = expr->getExprOperand()->getType();
+    }
+
+    ReportTypeUse(CurrentLoc(), type.getTypePtr());
+    return true;
+  }
+
   //------------------------------------------------------------
   // Visitors of types derived from clang::Type.
 
--- a/tests/cxx/typeid.cc
+++ b/tests/cxx/typeid.cc
@@ -1,0 +1,48 @@
+//===--- typeid.cc - test input file for iwyu -----------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+// Tests that IWYU suggests an include instead of a forward declaration for
+// expressions used as arguments to typeid(). This is necessary as calling
+// typeid() on an incomplete type causes the program to be ill-formed.
+// We need both typeid(expr) and typeid(type) calls to cause the necessary files
+// to be included, rather than forward declared.
+
+#include "tests/cxx/direct.h"
+#include <typeinfo>
+
+// Tests that typeid(expr) triggers an include, rather than a
+// forward declaration.
+// IWYU: IndirectClass needs a declaration
+void GetTypeInfoOfIndirectExpr(const IndirectClass &indirect) {
+  // IWYU: IndirectClass is...*indirect.h
+  const std::type_info &typeInfo = typeid(indirect);
+}
+
+// Tests that typeid(type) triggers an include, rather than a
+// forward declaration.
+void GetTypeInfoOfIndirectType() {
+  // IWYU: IndirectClass is...*indirect.h
+  const std::type_info &typeInfo = typeid(IndirectClass);
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/typeid.cc should add these lines:
+#include "tests/cxx/indirect.h"
+
+tests/cxx/typeid.cc should remove these lines:
+- #include "tests/cxx/direct.h"  // lines XX-XX
+
+The full include-list for tests/cxx/typeid.cc:
+#include <typeinfo>  // for type_info
+#include "tests/cxx/indirect.h"  // for IndirectClass
+
+***** IWYU_SUMMARY */


iwyu.cc
```diff
@@ -2549,6 +2549,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  // The C++ spec does not allow us to apply typeid to an incomplete type.
+  // Therefore, we must report the type to be included.
+  bool VisitCXXTypeidExpr(clang::CXXTypeidExpr* expr) {
+    if (CanIgnoreCurrentASTNode()) return true;
```
kimgr, line 2555: Would be nice if you could clang-format the new code, the `return true;` should be on a new line. I did a mass cleanup earlier this year: https://github.com/include-what-you-use/include-what-you-use/commit/890a3150d74f834a31e57c649195d6d68da682b1.

iwyu.cc
```diff
@@ -2549,6 +2549,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    if (expr->isTypeOperand()) {
```
kimgr, line 2556: The method looks a little unbalanced. I wonder if it would be nice to do:
```
QualType type;
if (expr->isTypeOperand()) {
  type = expr->getTypeOperandSourceInfo()->getType();
} else {
  type = expr->getExprOperand()->getType();
}

ReportTypeUse(CurrentLoc(), type.getTypePtr());
return true;
```

tests/cxx/typeid.cc
```diff
@@ -0,0 +1,37 @@
...
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+// Tests that IWYU suggests an include instead of a forward declaration for
+// expressions used as arguments to typeid(). This is necessary as calling
+// typeid() on an incomplete type causes the program to be ill-formed.
+
+#include "tests/cxx/direct.h"
+#include <typeinfo>
```
kimgr, line 17: It would be nice not to have to pull this standard library header in. Is it possible to do `(void)typeid(indirect);` and still see the full use of `IndirectClass`?

tests/cxx/typeid.cc
```diff
@@ -0,0 +1,37 @@
...
+
+// IWYU: IndirectClass needs a declaration
+void GetTypeInfoOfIndirectClass(const IndirectClass &indirect) {
+  // IWYU: IndirectClass is...*indirect.h
+  const std::type_info &typeInfo = typeid(indirect);
```
kimgr, line 22: I'm guessing this only covers the type-operand path. Would be nice to see one for the expr-operand as well!

kimgr:
Very nice, thanks! If you rebase on latest master the build should run to completion too, and you can see how the test suite fares.

tests/cxx/typeid.cc
bolshakov-a: > The type std​::​type_­info ([[type.info]](https://eel.is/c++draft/type.info)) is not predefined; if a standard library declaration ([[typeinfo.syn]](https://eel.is/c++draft/typeinfo.syn), [[std.modules]](https://eel.is/c++draft/std.modules)) of std​::​type_­info does not precede ([[basic.lookup.general]](https://eel.is/c++draft/basic.lookup.general)) a typeid expression, the program is ill-formed[.](https://eel.is/c++draft/expr.typeid#7.sentence-1)

https://eel.is/c++draft/expr.typeid

kimgr: :eyes: Oh, C++, you leave me no comfort. @bolshakov-a Thanks for the heads-up!

cottonxray:
Thank you very much for your comments! I think I addressed everything, but let me know if I missed something. Now I can cross my fingers for the build to pass :)

iwyu.cc
```diff
@@ -2549,6 +2549,20 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  // The C++ spec does not allow us to apply typeid to an incomplete type.
+  // Therefore, we must report the type to be included.
+  bool VisitCXXTypeidExpr(clang::CXXTypeidExpr* expr) {
+    QualType type;
```
kimgr, line 2555: Sorry I tricked you, we still need the `CanIgnoreCurrentASTNode` check here.

kimgr:
```
      return true;
```

tests/cxx/typeid_on_type.cc
```diff
@@ -0,0 +1,36 @@
+//===--- typeid_on_type.cc - test input file for iwyu ---------------------===//
```
kimgr, line 1: Ah, you can put both `typedid(type)` and `typeid(expr)` into the same file, makes it a little easier to reason about. I suggest `typeid.cc` like you had before, maybe with two functions and a little comment to describe what the difference is between the two forms.

kimgr (commented):
Thanks! Some details got lost in translation, nits inline :)

iwyu.cc
cottonxray: Sorry about that - if I was paying attention, I should have spotted that myself :)

tests/cxx/typeid_on_type.cc
cottonxray: Gotcha, thanks!

iwyu.cc
kimgr: OK, in the interest of nit-picky consistency, can you also lose the braces and move it first in the method?
```
bool VisitCXXTypeidExpr(clang::CXXTypeidExpr* expr) {
    return true;

  QualType type;
  ...
```

I'm too old to learn better tools (;-)), so I regularly use crazy pattern matching to find and rewrite code. Much easier if everything looks the same.

cottonxray: Sure thing!

kimgr:
Thanks, much appreciated!

cottonxray:
> Thanks, much appreciated!

Thank you for all of your help and time spent reviewing this - it's been a pleasure contributing to this project!

PR #1189, opened 2023-02-08
Fix the incorrect include path of Triple.h

lxsameer:
The `Triple.h` header is moved from `llvm/ADT/Triple.h` to `llvm/TargetParser/Triple.h` in LLVM's master (commit: f09cf34d00625e57dea5317a3ac0412c07292148). This patch will fix the pass and unblock the CI.


kimgr:
Thanks for fixing this up, and refering to the offending commit!

We have a strict message pattern for changes for Clang compatibility, so I will update the commit message as part of the merge.

PR #1193, opened 2023-02-24
Don't report into template definition file

bolshakov-a:
Reporting of symbol usage in one file depending on another file analysis seems strange. It could make some sense in order to note that template definition file provides some internal stuff, but this form of analysis depends on from which file it is performed and could not work when performing analysis of the sole template definition file.

This is a preparatory work for #1179.

Not sure if some test is worth to be written.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2971,12 +2971,9 @@ class InstantiatedTemplateVisitor
                      const char* comment = nullptr,
                      UseFlags extra_use_flags = 0) override {
     const SourceLocation actual_used_loc = GetLocOfTemplateThatProvides(decl);
-    if (actual_used_loc.isValid()) {
-      // If a template is responsible for this decl, then we don't add
-      // it to the cache; the cache is only for decls that the
-      // original caller is responsible for.
-      Base::ReportDeclUse(actual_used_loc, decl, comment, extra_use_flags);
-    } else {
+    // Report use only if template doesn't itself provide the declaration.
+    if (!actual_used_loc.isValid() ||
+        GetFileEntry(actual_used_loc) == GetFileEntry(caller_loc())) {
       // Let all the currently active types and decls know about this
       // report, so they can update their cache entries.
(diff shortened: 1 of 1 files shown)


iwyu.cc
```diff
@@ -2971,12 +2971,9 @@ class InstantiatedTemplateVisitor
...
     const SourceLocation actual_used_loc = GetLocOfTemplateThatProvides(decl);
-    if (actual_used_loc.isValid()) {
-      // If a template is responsible for this decl, then we don't add
-      // it to the cache; the cache is only for decls that the
-      // original caller is responsible for.
-      Base::ReportDeclUse(actual_used_loc, decl, comment, extra_use_flags);
-    } else {
+    // Report use only if template doesn't itself provide the declaration.
+    if (!actual_used_loc.isValid() ||
+        GetFileEntry(actual_used_loc) == GetFileEntry(caller_loc())) {
```
kimgr, line 2976: The only wrinkle I can think of is `.tcc` or `.inl` files, sometimes used to split up a header into chunks, e.g.
```
// foobar.hpp
template<class T> struct Something {
};

#include "foobar.tcc"

// foobar.tcc
template class Something<int> {
  // do something for int
}

template class Something<char> {
  // do something for char
}
// ... etc
```

I suppose that could end us up in a situations where the files are different even though they're both responsible for declaration of the template...? But this code only looks at the instantiation side of things... so maybe that's different?

bolshakov-a: > But this code only looks at the instantiation side of things... so maybe that's different?

I think so. Moreover, even if you check .tcc file along with .hpp file, you should place .tcc-required`#include`s into .tcc and not into .hpp, as IWYU principle states, in my opinion.

More importantly, the analysis is dependent on from that file it is performed, and it seems not a good thing. This PR avoids it.

kimgr: Makes sense! LGTM.

PR #1194, opened 2023-02-24
Fix CanForwardDeclareType for tpl instantiations

bolshakov-a:
`InstantiatedTemplateVisitor` traverses `SubstTemplateTypeParmType` underlying type retaining `SubstTemplateTypeParmType` node in the node stack. Its presence broke analysis in some cases.

This is one more lead-up PR for #1179.

### DIFF
--- a/tests/cxx/explicit_instantiation2.cc
+++ b/tests/cxx/explicit_instantiation2.cc
@@ -19,13 +19,13 @@
 //        definition (2), as it affects the semantics.
 // 2: Explicit instantiation definition.
 // 3: Full use in a template, provided as an explicit parameter.
-// 4: Fwd-decl use in a template, provided as an explicit parameter.
 // 5: Full use in a template, provided as an default parameter.
 // 7: Full use in a template, provided as a template template parameter.
 // 8: Fwd-decl use in a template, provided as a template template parameter.
 //
 // Negative scenarios, where the dependent template specialization is not
 // required, or it does not provide an explicit instantiation:
+// 4: Fwd-decl use in a template, provided as an explicit parameter.
 // 6: Fwd-decl use in a template, provided as a default parameter.
 // 9: Implicit instantiation of Template<int>
 // 10: Specialization of Template<T>
@@ -53,7 +53,6 @@ FullUseArg<
     // IWYU: Template is...*explicit_instantiation-template.h
     t3; // 3
 
-// IWYU: Template is...*template_short.h.*for explicit instantiation
 FwdDeclUseArg<
     // IWYU: Template needs a declaration...*
     Template<short>>
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3072,34 +3072,6 @@ class InstantiatedTemplateVisitor
     return TraverseTemplateSpecializationTypeHelper(typeloc.getTypePtr());
   }
 
-  bool TraverseSubstTemplateTypeParmTypeHelper(
-      const clang::SubstTemplateTypeParmType* type) {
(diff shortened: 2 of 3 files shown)


tests/cxx/explicit_instantiation2.cc
```diff
@@ -53,7 +53,6 @@ FullUseArg<
     // IWYU: Template is...*explicit_instantiation-template.h
     t3; // 3
 
-// IWYU: Template is...*template_short.h.*for explicit instantiation
```
kimgr, line 56: This one makes me a little nervous, since there's a an explicit specialization declaration for `Template<short>` in `...template_short.h`. But there's also an explicit specialization definition of that same specialization in this file, so I'm not sure why the remote decl would be crucial.

kimgr:
Looks good! If you have any intuition for the removed specialization dependency in `explicit_instantiations2.cc`, I'd love to hear it, but it seems like a safe step to me.

bolshakov-a:
But wait... I think there is a better way to fix the problem with `SubstTemplateTypeParmType` node in the stack: don't traverse its underlying type at all.

tests/cxx/explicit_instantiation2.cc
bolshakov-a: I don't think that reporting of explicit instantiation of `Template<short>` is actually needed for a fwd-decl use inside `FwdDeclUseArg`.
> But there's also an explicit specialization definition of that same specialization in this file, so I'm not sure why the remote decl would be crucial.

Yes, that seems strange, and all this test (maybe after line 42) could become broken after some explicit instantiation analysis improval.

kimgr: I think [this section](https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L3268) is what does it.

If `CanForwardDeclareType` works better now, the explicit instantiation won't be reporteed, I guess.

Moreover, that `ReportExplicitInstantiation` implementation looks naive -- I think we should do something like:
* collect all redecls that are explicit instantiations and not provided by template
* if one exists in same file, use it
* if one exists in another file, and is a declaration, use it (on the assumption that an `extern template ...` declaration is cheaper to process than a full template instantiation)
* otherwise, use an arbitrary one

kimgr:
I noticed the rationale for the test regressions in the commit message, thanks!

> but it is due to the fact that typedef complex underlying types still aren't handled properly.

Out of curiosity, would the "typedef complex underlying types" be covered by this TODO: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_ast_util.cc#L590, or is there more going on?

This patch LGTM, are you comfortable merging it too?

bolshakov-a:
I meant primarily [this TODO](https://github.com/include-what-you-use/include-what-you-use/blob/0.19/iwyu.cc#L1440), but they are related.

I have yet nothing against getting this PR merged, hoping that the regression isn't very severe.

kimgr:
Ah, thanks. `It will require pretty substantial iwyu surgery` is ominous, but looking at the particular regression it seems like a very contrived scenario. I think we can live with it, if it moves us forward with regards to template spills in general.

I'll merge, thanks!

bolshakov-a:
> It will require pretty substantial iwyu surgery is ominous

Don't worry! It is almost done already. You can take a look at a couple of draft commits:
https://github.com/bolshakov-a/include-what-you-use/commit/6cc256b9570b80fb5bc9517dd6ec2eb564a33651
https://github.com/bolshakov-a/include-what-you-use/commit/087eedd89ba038030244ee352f61a7662847e9ec

PR #1195, opened 2023-02-25
Use the correct decl location for use logging

kimgr:
Back in 27da44e009660d8efc9726b1a4d80d8f9f063539 (Oct, 2018), the location of the decl was precomputed and stored separately in the data structure for uses.

Unfortunately it was only used for analysis then, not logging. Update logging to use OneUse::decl_loc() consistently.

This clarifies log output particularly for uses with custom flags for function definitions and explicit template instantiations, which are the only ones where decl_loc() is different from GetLocation(decl()).

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -618,7 +618,7 @@ static void LogSymbolUse(const string& prefix, const OneUse& use) {
   string decl_loc;
   string printable_ptr;
   if (use.decl()) {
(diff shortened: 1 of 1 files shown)


kimgr:
@bolshakov-a I ran into this while trying to understand the explicit instantiations analysis flow. Please take a look.

bolshakov-a:
I cannot say that I fully understand explicit instantiation handling mechanics, but this change LGTM.

PR #1196, opened 2023-02-26
More consistent work with resugaring map

bolshakov-a:
Keys of `resugar_map` are canonical types, but `Desugar` doesn't give canonical type if template specialization or alias is present in the sugar chain. It may become important if `ReportTypeUse` starts to call `CanIgnoreType` check for types stored in the cache because the cache stores sugared types, e.g. `TemplateSpecializationType`s.

After this change, `IwyuBaseAstVisitor::VisitTemplateSpecializationType` starts to work from `InstantiatedTemplateVisitor` in some cases (at least, on traversal original template specialization type, e.g. `Tpl1<Tpl2<int>>`). Hence, an empty implementation of `ReportDeclForwardDeclareUse` is needed in `InstantiatedTemplateVisitor`, because it should not report forward declarations.

No functional change is expected.

This is one more preparatory PR for #1179.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2942,7 +2942,7 @@ class InstantiatedTemplateVisitor
     // Among all subst-type params, we only want those in the resugar-map. If
     // we're not in the resugar-map at all, we're not a type corresponding to
     // the template being instantiated, so we can be ignored.
-    type = Desugar(type);
+    type = GetCanonicalType(type);
     return ContainsKey(resugar_map_, type);
   }
 
@@ -2980,6 +2980,15 @@ class InstantiatedTemplateVisitor
         storer->NoteReportedDecl(decl);
       Base::ReportDeclUse(caller_loc(), decl, comment, extra_use_flags);
     }
+  }
+
+  void ReportDeclForwardDeclareUse(SourceLocation, const NamedDecl*,
+                                   const char* /*comment*/ = nullptr) override {
+    // Forward declarations make sense only when a type is explicitly written.
+    // But 'InstantiatedTemplateVisitor' is to traverse implicit template
+    // specializations, actually. (Template definitions and explicit
+    // specializations are handled by 'IwyuAstConsumer'). Because it traverses
+    // implicit stuff, it should not suggest forward declarations.
   }
 
   void ReportTypeUse(SourceLocation used_loc, const Type* type,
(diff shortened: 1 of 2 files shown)


iwyu.cc
```diff
@@ -2982,6 +2982,10 @@ class InstantiatedTemplateVisitor
     }
   }
 
+  void ReportDeclForwardDeclareUse(SourceLocation, const NamedDecl*,
+                                   const char* /*comment*/ = nullptr) override {
+  }
```
kimgr, line 2987: Add a comment here to explain why we never want forward-declare uses of decls in template instantiation.

iwyu.cc
```diff
@@ -3358,7 +3362,7 @@ class InstantiatedTemplateVisitor
   // If we're not in the resugar-map, then we weren't canonicalized,
   // so we can just use the input type unchanged.
   const Type* ResugarType(const Type* type) const {
-    type = Desugar(type);
```
kimgr, line 3361: Before my change to `Desugar` this only did `RemoveSubstTemplateTypeParm`, which I suppose is a form of resugaring. Now that we're going all the way to the canonical type, it's very much the reverse of resugaring.

What would happen if we did something like
```
const Type* canonical_type = GetCanonicalType(type);
if (ContainsKeyValue(resugar_map_, canonical_type, static_cast<const Type*>(nullptr)))
  return type;  // keep the original type if no match
return GetOrDefault(resugar_map_, canonical_type, type);  // keep the original type if no match
```

That seems to me it would return a more sugared type. But on the other hand I suspect it becomes much less deterministic what the result of `ResugarType` is. Thoughts?

bolshakov-a: In fact, the change in this line isn't needed. I've added it only for consistency. I can apply your suggestion or revert it back at all. Not sure.

kimgr: I also appreciate the consistency, so I'm inclined to keep it like it is in this patch.

But are you saying that `ResugarType` is already always called with canonical types? I.e. the following would hold:
```
CHECK_(type == GetCanonicalType(type));
```
? Might be nice to that instead.
```

bolshakov-a: No, I meant that it is called either with canonical type, or with sugared type (from the cache). In the latter case, it just returns back that sugared type.

kimgr: Ah, thanks. As I touched on earlier, I think it makes sense for `ResugarType` to return something deterministic if it's called with a type that is not in the resugar map. And I think using the canonical type is the best way to highlight missing resugaring (as opposed to returning back an accidentally-/half-sugared type). So let's keep this as-is. LGTM.

bolshakov-a: I've recalled that I intend to make `InstantiatedTemplateVisitor` reporting dependent types which are unrelated to original template arguments (see [that comment](https://github.com/include-what-you-use/include-what-you-use/pull/1182#issuecomment-1406681708)). So, returning original type here makes sense. But it could be done along with aforementioned change (and along with writing corresponding test case).

Issue #558, opened 2018-05-31
Template explicit instantiation declarations aren't handled properly


scramsby:
# Code

## some_class.h

```C++
#pragma once

template<typename T>
class some_class
{
};

extern template class some_class<int>;
```

## some_class.cpp

```C++
#include "some_class.h"

// template instantiations
template class some_class<int>;
```

# Output

```
C:/temp/some_class.h should add these lines:

C:/temp/some_class.h should remove these lines:
- class some_class;  // lines 4-4

The full include-list for C:/temp/some_class.h:
---
```

# Issues

1. The line suggested to be removed is quoted from some_class.h, but the actual line in question with that line number is actually from some_class.cpp.
2. The actual code text doesn't quite match what's actually in the source file.
3. The tool should not be suggesting to remove explicit template instantiations in the first place. Just because they aren't used by the declaring class doesn't mean they won't be used by dependent code.

See [cppreference.com](http://en.cppreference.com/w/cpp/language/class_template#Explicit_instantiation) for description of explicit template instantiation declarations and definitions.

scramsby:
Note: I confirmed that the suggestion goes away if you add an "// IWYU pragma: keep" in the .cpp, despite the instruction being to remove it from the .h.

jru:
Hey there, thanks for the great repro! 

I just opened a PR to fix this. To be exact: it addresses (3), the handling of explicit template instantiations, which is the main issue. 

(1) and (2) have to do with forward declarations only and how they are created by the tool, in *AddForwardDeclare*. In some cases it will select a declaration from a different file and mix the original filename with the source ranges of where it thinks the decl lives, messing up the output big time :) 

Anyway, I think this should probably go into its own issue.  @kimgr Are you good with that?

kimgr:
@jru Yeah, sounds like a good idea, thanks!

PR #1202, opened 2023-03-02
Report explicit instantions with more discretion

kimgr:
Whenever we ran into an explicit instantiation declaration/definition,
or the use of one, we would report all previously seen declarations of
it.

If the same explicit instantiation declaration showed up in multiple
headers, we might unnecessarily keep otherwise-unused #includes.

Use more discretion when reporting these uses:

* If there is a decl/defn in the same file, use that.
* If not, but one of the prior redecl is a declaration, not a
  definition, prefer that, because explicit instantiation declarations
  should be cheaper than their definitions
* If there are no declarations, use whatever definition is first in the
  redecl chain

This should minimize the includes necessary for explicit instantiation
declarations/definitions.

Add a testcase to explicitly capture selection of explicit
instantiations, and update explicit_instantiation2 testcase for new
policy.

### DIFF
--- a/tests/cxx/expl_inst_select.cc
+++ b/tests/cxx/expl_inst_select.cc
@@ -1,0 +1,62 @@
+//===--- expl_inst_select.cc - test input file for iwyu -------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// This checks deduplication/selection of explicit template instantiations.
+
+// IWYU_ARGS: -I .
+
+#include "tests/cxx/expl_inst_select-d1.h"
+
+// An explicit instantiation definition anchors a prior declaration.
+// IWYU: Template is...*expl_inst_select-i1.h
+// IWYU: Template is...*expl_inst_select-i2.h.*for explicit instantiation
+template class Template<char>;
+
+// An explicit instantiation declaration for later use.
+// IWYU: Template needs a declaration...*
+// IWYU: Template is...*expl_inst_select-i1.h
+extern template class Template<int>;
+
+// Use of an explicit instantiation for which there is both a declaration and
+// definition in the include closure should prefer a declaration.
+// IWYU: Template is...*expl_inst_select-i1.h
+// IWYU: Template is...*expl_inst_select-i2.h.*for explicit instantiation
+Template<short> ts;
+
+// Use of an explicit instantiation for which there is only a definition.
+// IWYU: Template is...*expl_inst_select-i1.h
+// IWYU: Template is...*expl_inst_select-i3.h.*for explicit instantiation
+Template<double> td;
+
+// No 'for explicit instantiation' diagnostic for use of an instantiation
+// definition available in the same file.
+// IWYU: Template is...*expl_inst_select-i1.h
+Template<char> tc;
+
+// No 'for explicit instantiation' diagnostic for use of an instantiation
+// declaration available in the same file.
+// IWYU: Template is...*expl_inst_select-i1.h
+Template<int> ti;
+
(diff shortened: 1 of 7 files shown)


tests/cxx/expl_inst_select.cc
```diff
@@ -0,0 +1,62 @@
...
+
+// This checks deduplication/selection of explicit template instantiations.
+
+// IWYU_ARGS: -I .
+
+#include "tests/cxx/expl_inst_select-d1.h"
+
+// An explicit instantiation definition anchors a prior declaration.
+// IWYU: Template is...*expl_inst_select-i1.h
+// IWYU: Template is...*expl_inst_select-i2.h.*for explicit instantiation
```
bolshakov-a, line 18: @kimgr, do you know any reason to include an explicit instantiation declaration here?

kimgr: It seems to be covering the reporting driven from here:
https://github.com/include-what-you-use/include-what-you-use/blob/53ce70ecc5ee84899ca067f26ddce5cd77dcb064/iwyu.cc#L3263

... which as far as I can tell was added to solve:
https://github.com/include-what-you-use/include-what-you-use/issues/558

via https://github.com/include-what-you-use/include-what-you-use/pull/578.

The bug report in 558 stated that the explicit instantiation declaration is required for the definition, so presumably that's why? There may be some misunderstanding along the way there, of course.

bolshakov-a: > The bug report in 558 stated that the explicit instantiation declaration is required for the definition

I don't see that. The original bug as-reported is just that IWYU produced some garbage on explicit instantiations.

> 3. The tool should not be suggesting to remove explicit template instantiations in the first place.

Agreed, but it should not do it regardless of the presence of the instantiation definition somewhere in the translation unit.

From [your comment](https://github.com/include-what-you-use/include-what-you-use/pull/578#issuecomment-421732597):
> Explicit template instantiation definition of `Template<short>`, right? That should require `template-decl-short.h`, otherwise it will create a reusable instantiation of the wrong template.

I've failed to understand, could you explain please? As far as I know, instantiation declarations are to suppress instantiations in a translation unit, and instantiation definitions are to the opposite purpose: to instantiate regardless of being used or not. Thus, they are mutually exclusive in essence, and definitions "beat" declarations, in fact. Consider:
```cpp
// tpl.h

template <typename T>
struct Tpl {
    static int i;
    int local_i;
    static int getStatic() { return i; }
    int getLocal() const { return local_i; }
};

// tpl.cpp

#include "tpl.h"

template <typename T>
int Tpl<T>::i = 2;

template class Tpl<int>;

// main.cpp

#include "tpl.h"

extern template class Tpl<int>;

int main() {
    Tpl<int> t{5};
    return t.getLocal() + t.getStatic();
}

```
This compiles, links and runs fine. Moreover, [clang LLVM IR output](https://godbolt.org/z/ensnfb744) and [GCC assembly output](https://godbolt.org/z/YjbzEPnK3) are the same with and without the instantiation declaration in the `tpl.cpp` translation unit.

An explicit instantiation definition requires the class template definition to be available (likewise an explicit instantiation declaration) but not the declaration, as seemed to me.

kimgr: > > Explicit template instantiation definition of Template<short>, right? That should require template-decl-short.h,
> > otherwise it will create a reusable instantiation of the wrong template.
>
> I've failed to understand, could you explain please?

I think I confused myself as mentioned in the next comment: https://github.com/include-what-you-use/include-what-you-use/pull/578#issuecomment-421734997.

If `Template<short>` was a _specialization_, I think that sentence would make a bit more sense. Let's say:

* There is an explicit specialization for `Template<short>` in `template-decl-short.h`
* Explicit instantiation definition `Template<short>` would instantiate the base template or the specialization, depending on what was visible at the time of definition

But that does seem like a separate constraint altogether.

Issue #1181, opened 2023-01-13
IWYU suggests forward declaring enum from .cpp in header


ts826848:
This is on macOS 10.15.7 with IWYU 0.19 based on Homebrew Clang 15.0.6.

I have a scoped enum declared/defined in a .cpp file I'm using to test a template class. IWYU suggests forward declaring this enum in the header file, which doesn't seem like a good suggestion. In my original code, this enum is in an anonymous namespace, but it appears this is not necessary to reproduce the issue.

Reduced example:

`reduced.hpp`:

```cpp
template <typename> struct enum_data {
  static const int staticVar;
  virtual void var() const;
};
template <typename E> const int enum_data<E>::staticVar = 0;
template <typename E> void enum_data<E>::var() const { staticVar; }
```

`reduced.cpp`:

```cpp
#include "reduced.hpp"
enum class color { RED = 2 };
enum_data<color> d;
```

IWYU output:

```
/tmp % include-what-you-use -w reduced.cpp

reduced.hpp should add these lines:
enum class color;

reduced.hpp should remove these lines:

The full include-list for reduced.hpp:
enum class color;
---

(reduced.cpp has correct #includes/fwd-decls)
```

Verbose output, with "Adding mapping from..." and defines removed:

```
/tmp % include-what-you-use -w -Xiwyu --verbose=12 reduced.cpp
Setting verbose-level to 12
<snip: Adding mapping from...>
Search path: /usr/local/Cellar/include-what-you-use/0.19/libexec/lib/clang/15.0.6/include/ (system)
Search path: /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/ (system)
Search path: /usr/local/Cellar/include-what-you-use/0.19/libexec/include/c++/v1/ (system)
[ #include    ]  (reduced.cpp)
Found include: <built-in>:-1 -> reduced.cpp
Added an #include: <built-in> ->
Added to main compilation unit: reduced.cpp
[ #include    ]  (<built-in>)
[ Renaming to ] <built-in>:1:1
<snip: [ #define     ]...>
[ Exiting to  ] <built-in>:1:1
[ Exiting to  ] reduced.cpp:1:1
[ #include    ] "reduced.hpp" (reduced.hpp)
Found include: reduced.cpp:1 -> reduced.hpp
Adding reduced.hpp as associated header for reduced.cpp
Marked reduced.hpp as associated header of reduced.cpp.
[...]

PR #1203, opened 2023-03-05
Avoid visiting substituted template parameter types

kimgr:
For a template where definition and instantiation are in different files, but the same component (associated header/source files), IWYU could get confused about attribution of types visited via SubstTemplateTypeParmType.

Template instantiation and the policy around which types are reported where is handled by InstantiatedTemplateVisitor with a lot of precision.

Stub out the traversal of SubstTemplateTypeParmType in IwyuAstConsumer (the top-level visitor for source files) on the assumption that instantiation/substitution should be handled exclusively by InstantiatedTemplateVisitor.

Fixes issue #1181.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -197,6 +197,7 @@ using clang::Sema;
 using clang::SourceLocation;
 using clang::Stmt;
 using clang::SubstTemplateTypeParmType;
+using clang::SubstTemplateTypeParmTypeLoc;
 using clang::TagDecl;
 using clang::TagType;
 using clang::TemplateArgument;
@@ -4226,6 +4227,19 @@ class IwyuAstConsumer
     }
 
     return Base::VisitTemplateSpecializationType(type);
+  }
+
+  // RecursiveASTVisitor by default traverses the SubstTemplateTypeParmType
+  // replacement type, which means in some cases we'll see substituted types
+  // when instantiating out-of-line template members. Recording and attribution
+  // of template argument uses should be handled by InstantiatedTemplateVisitor,
+  // so short-circuit the traversal here.
+  bool TraverseSubstTemplateTypeParmType(SubstTemplateTypeParmType* type) {
+    return getDerived().WalkUpFromSubstTemplateTypeParmType(type);
+  }
+
+  bool TraverseSubstTemplateTypeParmTypeLoc(SubstTemplateTypeParmTypeLoc loc) {
+    return getDerived().WalkUpFromSubstTemplateTypeParmTypeLoc(loc);
   }
 
   // --- Visitors defined by BaseASTVisitor (not RecursiveASTVisitor).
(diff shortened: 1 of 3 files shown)


kimgr:
@bolshakov-a This may be a little heavy-handed, I'd appreciate your input.

bolshakov-a:
Looks a little strange, at a glance. Maybe, I'll take a closer look at the next weekend.

kimgr:
Thank you!

bolshakov-a:
The problem is due to presence of `VarDecl` node in the AST corresponding to the implicitly specialized template. That node has `TranslationUnitDecl` as its lexical scope and takes part in the AST traversal, but its location is attributed to the unspecialized varibale declaration, which is in the header. I think the more correct way to fix it would be to avoid traversal of such implicit `VarDecls`. Somehow...

kimgr:
Thank you! I wonder if that's a bug in the AST; the lexical scope of the `VarDecl` should still be the class template, right?

I'm having trouble coming up with examples where `SubstTemplateTypeParmType` are ever relevant/interesting to IWYU analysis in the top-level visitor... If you have any ideas here, I'd love to see examples.

bolshakov-a:
That occurs to be an interesting issue!

I don't think it is a clang bug. The lexical scope of static variable out-of-class definitions is a namespace or translation unit scope, and the semantic scope is the class definition, AFAIU. I think, all of it should be the same for templates. But I'm still not sure that static member fields of templates behave like C++17 inline-variables, and placing their definitions in a header could not lead to ODR violation...

I think your solution is acceptable, but not full, because when the static variable has the template parameter type instead of `int`, the use of `IndirectClass` (full-use in this case) is still attributed to the header. This report occurs from `InstantiatedTemplateVisitor` traversing nested name specifier (`Template<E>::`) in the variable definition. And this is another issue, in fact, because it is a report of static member field _declaration_, full type of which is not necessary. But if you fix it, you probably lose the `IndirectClass` full-type-use reporting at all. So, probably, the traversal of the implicit variable definition is actually needed, but should be attributed to the right place. Maybe from `InstantiatedTemplateVisitor`, maybe still from `IwyuAstConsumer`. I haven't yet a broad picture of the correct solution.

bolshakov-a:
> I'm still not sure that static member fields of templates behave like C++17 inline-variables, and placing their definitions in a header could not lead to ODR violation...

Yes, it is acceptable (at least, it was in C++17):
> There can be more than one definition of ... [static data member of a class template](https://timsong-cpp.github.io/cppwp/n4659/temp.static) ... in a program provided that each definition appears in a different translation unit ...

https://timsong-cpp.github.io/cppwp/n4659/basic.def.odr#6

kimgr:
@bolshakov-a Thanks for digging into this! I agree that it seems dramatic to skip over a whole subtree like this. I'm going to close this for now, and hopefully I can come back to it at some point. FYI, @ts826848 -- I saw you applied the patch and had it working. You can probably keep it for now, just so you know it's not landing on our master.

PR #1209, opened 2023-03-16
[clang-compat] Use new include path for Host.h

kilroyd:
The header was moved from llvm/Support/ to llvm/TargetParser/ as part of a breaking change in LLVM commit:
https://github.com/llvm/llvm-project/commit/f09cf34d00625e57dea5317a3ac0412c07292148

A warning was subsequently added in commit:
https://github.com/llvm/llvm-project/commit/a9329a96ada3

Use the new location. Also re-sort headers alphabetically.


kimgr:
Thanks! I tweaked the commit message in the interest of consistency, it seems we use `[clang compat]`, not `[clang-compat]`

PR #1212, opened 2023-03-17
Fix run_iwyu_test.py instructions in CONTRIBUTING

kilroyd:
Commit 8343f0ce8aa1 modified how tests are named. Update the CONTRIBUTING instructions to align.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,9 +22,9 @@ It runs one test for each `.cc` file in the `tests/` directory.  (We have additi
 
 The output can be a bit hard to read, but if a test fails, the reason why will be listed after the `ERROR:root:Test failed for xxx` line.
 
-You can select individual tests by listing their filename without extension as arguments
+You can select individual tests by listing them as arguments. Test names are derived from the file path and name, e.g. `tests/cxx/array.cc` will be named `cxx.test_array`. You can use `python run_iwyu_tests.py --list` to list all available test names.
 
-    python run_iwyu_tests.py array macro_location
+    python run_iwyu_tests.py cxx.test_array cxx.test_macro_location c.test_enum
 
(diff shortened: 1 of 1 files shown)


kimgr:
I've actually considered changing the generated method names -- maybe I should do that before we lock them down in documentation :). I'll see if I can get something done this weekend, otherwise this looks good to go.

kimgr:
Turns out the test naming is required by the runner, so I won't be able to do anything about that. `run_iwyu_tests.py` generates a set of Python `unittest` suites dynamically, and as far as I can tell, `unittest` can only discern test methods by prefix, by default `test`.

CONTRIBUTING.md
```diff
@@ -22,9 +22,9 @@ It runs one test for each `.cc` file in the `tests/` directory.  (We have additi
 
 The output can be a bit hard to read, but if a test fails, the reason why will be listed after the `ERROR:root:Test failed for xxx` line.
 
-You can select individual tests by listing their filename without extension as arguments
+You can select individual tests by listing them as arguments. Generally this is 'c.test_' or 'cxx.test_' (depending on whether the test in a C or C++ test) followed by the filename without any extension
```
kimgr, line 25: The more general pattern is `$suite.test_$filenamenoext`, where `suite` is one of the subdirectories of `tests/`, currently:

* `driver`
* `c`
* `cxx`

We probably don't need to go into that kind of detail, but the proposed wording feels a little _too_ non-committal :)

Maybe something like:
```
You can select individual tests by listing them as arguments. Test names are derived from the file path and name, e.g. `tests/cxx/array.cc` will be named `cxx.test_array`. You can use `python run_iwyu_tests.py --list` to list all available test names.

    ... your example here ...
```

kilroyd: Sounds much better, I'll post an update shortly

kimgr:
Nice, thanks!

Issue #828, opened 2020-09-17
Namespaces from using directives are attributed to header first declaring them, not necessarily relevant header.


roccomoretti:
For the following code:

```
//////////////// A.hpp
namespace nsp {
    class A {};
}

//////////////// B.hpp
namespace nsp {
    struct B {
	int func() { return 0; }
    };
}

//////////////// main.cpp
#include "A.hpp"
#include "B.hpp"

using namespace nsp;

int main() {
	B  b;
	return b.func();
}
```

Include-what-you-use says (incorrectly, I'd say) that all includes are correct. However, if you swap the order of A.hpp and B.hpp in main.cpp, then it recommends removing A.hpp (the expected behavior).

Behavior observed with include-what-you-use version 0.14, compiled from the Github release.

(I should clarify that it's the using namespace directive which is causing the issue, as removing that line and being explicit about nsp::B also will cause iwyu to suggest removing A.hpp.)

kimgr:
Interesting problem.

We're actually just leaning on Clang to do the resolution here via `UsingDirectiveDecl::getNominatedNamespaceAsWritten`: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L3891

And it makes sense for the compiler to do it this way -- assuming all namespace decls are interchangeable. But it would be nice if IWYU could do something smarter.

We already have custom handling for `using` _declarations_ (as opposed to _directives_), I wonder if a similar approach could work for directives too. I don't remember all the details, but I think we report `using` declarations when they're seen, then report uses of these decls if they are required for another use to work and finally report uses only for those that are. Something like that (see https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_output.cc#L580 and anything containing `UsingDecl` in that file if you're interested in more detail).

But it does sound like this is more difficult, as the compiler seems to just use whatever namespace-decl it sees first in the translation unit. The deferred analysis would have to somehow look through all redeclarations of a namespace and discard all coming from a header that is otherwise unused. Except if that's the only place it's declared. I'm not sure we have enough metadata available to do that kind of somersault.

mpusz:
I have a similar issue. `glide_computer.h` starts with:

```cpp
#include "units/isq/si/length.h"
#include "units/isq/si/speed.h"
#include "units/isq/si/time.h"

using namespace units;
using namespace units::isq;
```

and every header from the ones provided above defines those namespaces. However, IWYU raises the following:

```text
[...]

PR #1213, opened 2023-03-17
RFC: Improved suggestions for namespace use

kilroyd:
This series is looking to address #828, where IWYU just suggests the first header that declares a namespace to satisfy a using directive.

The first couple commits play with namespace aliases, which were easier to get a handle on. The first commit adds reporting for when namespace aliases use a namespace, and the second adds a warning when the alias is unused. Although useful, I suspect we want to drop the second commmit (because IWYU generally doesn't warn on unused declarations). I believe the first commit is technically correct, but I've not seen a codebase with aliases in headers so hopefully this won't be disruptive.

The third commit finally attempts to improve the suggestion for the inculde to use. The logic is mostly in a new file, iwyu_namespace.cc, as I was trying a bunch of different things. I ended up caching the ReportDeclUse() to call during ResolvePendingAnalysis(), so we can pick the prefered NamespaceDecl to report. This works for the simple test case.

Any feedback appreciated!

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -183,6 +183,7 @@ using clang::LValueReferenceType;
 using clang::LinkageSpecDecl;
 using clang::MemberExpr;
 using clang::NamedDecl;
+using clang::NamespaceDecl;
 using clang::NestedNameSpecifier;
 using clang::NestedNameSpecifierLoc;
 using clang::OverloadExpr;
@@ -1700,8 +1701,25 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       ReportTypeUse(used_loc, type);
   }
 
+  void ReportNamespaceDeclUse(SourceLocation used_loc,
+                              const NamespaceDecl* namespace_decl,
+                              const char* comment = nullptr,
+                              UseFlags extra_use_flags = 0) {
+    // Canonicalize the use location and report the use.
+    used_loc = GetCanonicalUseLocation(used_loc, namespace_decl);
+    const FileEntry* used_in = GetFileEntry(used_loc);
+    preprocessor_info().FileInfoFor(used_in)->ReportNamespaceDeclUse(
+        used_loc, namespace_decl, extra_use_flags, comment);
+  }
+
   //------------------------------------------------------------
   // Visitors of types derived from clang::Decl.
+  bool VisitNamedDecl(clang::NamedDecl* decl) {
+    preprocessor_info()
+        .FileInfoFor(CurrentFileEntry())
+        ->ReportDeclInNamespaces(decl);
+    return true;
+  }
 
   // Friend declarations only need their types forward-declared.
   bool VisitFriendDecl(clang::FriendDecl* decl) {
@@ -3889,7 +3907,8 @@ class IwyuAstConsumer
   bool VisitNamespaceAliasDecl(clang::NamespaceAliasDecl* decl) {
     if (CanIgnoreCurrentASTNode())
       return true;
-    ReportDeclUse(CurrentLoc(), decl->getNamespace());
+    ReportNamespaceDeclUse(CurrentLoc(), decl->getNamespace());
+
     return Base::VisitNamespaceAliasDecl(decl);
   }
 
@@ -4030,7 +4049,12 @@ class IwyuAstConsumer
   bool VisitUsingDirectiveDecl(clang::UsingDirectiveDecl *decl) {
     if (CanIgnoreCurrentASTNode())
       return true;
-    ReportDeclUse(CurrentLoc(), decl->getNominatedNamespaceAsWritten());
+
+    if (const auto* target_ns =
+            dyn_cast<NamespaceDecl>(decl->getNominatedNamespaceAsWritten())) {
+      ReportNamespaceDeclUse(CurrentLoc(), target_ns);
+    }
+
     return Base::VisitUsingDirectiveDecl(decl);
   }
 
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -36,6 +36,7 @@
 
 namespace include_what_you_use {
 
+using clang::BlockDecl;
 using clang::ClassTemplateDecl;
 using clang::ClassTemplateSpecializationDecl;
 using clang::CXXMethodDecl;
@@ -54,6 +55,7 @@ using clang::SourceRange;
 using clang::TagDecl;
 using clang::TemplateDecl;
 using clang::UsingDecl;
+using clang::UsingDirectiveDecl;
 using llvm::cast;
 using llvm::dyn_cast;
 using llvm::errs;
@@ -69,6 +71,36 @@ using std::vector;
 namespace internal {
 
 namespace {
+
+// Map containing the information we need to pick the best header
+// file.
+//
+// The namespace information is stored within the relevant
+// IwyuFileInfo in which the namespace is declared, in the nsinfo map.
+//
+// For this map, the main key is ns->getCanonicalDecl() - which
+// happens to be the first namespace declaration seen.
+//
+// For each namespace we have a secondary map pointing to the per file
+// information stored in IwyuFileInfo. These are the candidate header
+// files for the namespace. The namespace may be declared more than
+// once in each file. We just keep a reference to the first namespace
+// declaration in the file, and a count of all the full declarations
+// and using directives in the namespace.
+//
+// { canonical_namespace_decl :
+//     { fileentry : { first_ns_decl, decl_count, directive_count },
+//       fileentry : { first_ns_decl, decl_count, directive_count },
+//     },
+//   canonical_namespace_decl :
+//     { fileentry : { first_ns_decl, decl_count, directive_count },
+//       fileentry : { first_ns_decl, decl_count, directive_count },
+//     },
+//   etc
+// }
+static map<const clang::NamespaceDecl*,
+           map<const clang::FileEntry*, IwyuNamespaceInfo*>>
+    nsinfoPerFile;
 
 class OutputLine {
  public:
@@ -733,6 +765,17 @@ void IwyuFileInfo::ReportUsingDeclUse(SourceLocation use_loc,
   ReportFullSymbolUse(use_loc, using_decl, flags, comment);
 }
 
+void IwyuFileInfo::ReportNamespaceDeclUse(SourceLocation use_loc,
+                                          const NamespaceDecl* decl,
+                                          UseFlags flags, const char* comment) {
+  // In order to assign namespace usage to the best file, don't record the
+  // usage immediately. Instead keep the use in a vector and report it
+  // after we've identified the best file to satisfy it.
+  namespace_uses.push_back(
+      std::tuple{use_loc, decl, flags,
+                 comment ? optional(string(comment)) : std::nullopt});
+}
+
 // Given a collection of symbol-uses for symbols defined in various
 // files, figures out the minimal set of #includes needed to get those
 // definitions.  Typically this is a trivial task: if we need the full
@@ -749,6 +792,155 @@ static void LogIncludeMapping(const string& reason, const OneUse& use) {
   VERRS(6) << "Mapped " << use.decl_filepath() << " to "
            << use.suggested_header() << " for " << use.symbol_name()
            << " (" << reason << ")\n";
+}
+
+IwyuNamespaceInfo* IwyuFileInfo::getNamespaceInfo(const NamespaceDecl* ns,
+                                                  const FileEntry* file) {
+  const NamespaceDecl* canonical_ns = ns->getCanonicalDecl();
+
+  // Get the entry from nsinfo, creating it if necessary.
+  auto inserted = nsinfo.emplace(canonical_ns, ns);
+
+  IwyuNamespaceInfo* info = &(inserted.first->second);
+
+  if (inserted.second) {
+    // A new entry was created. Update nsinfoPerFile so we can lookup
+    // the namespace info by namespace.
+
+    // Get the intermediate map, creating it if necessary
+    map<const FileEntry*, IwyuNamespaceInfo*>& m =
+        internal::nsinfoPerFile[canonical_ns];
+
+    // Insert the new entry for this file
+    m[file] = info;
+  }
+
+  return info;
+}
+
+const NamespaceDecl* IwyuFileInfo::GetBestNamespaceDecl(
+    const NamespaceDecl* ns) {
+  string ns_ident = ns->getQualifiedNameAsString();
+  const NamespaceDecl* canonical_ns = ns->getCanonicalDecl();
+  auto m = internal::nsinfoPerFile[canonical_ns];
+
+  const NamespaceDecl* best_decl = nullptr;
+  const NamespaceDecl* decl = nullptr;
+  unsigned num_candidates = 0;
+  unsigned full_decl_count = 0;
+  unsigned best_using_directives = 0;
+
+  for (auto& [fileentry, nsinfo] : m) {
+    decl = nsinfo->GetNamespaceDecl();
+    unsigned num_decls = nsinfo->GetFullDeclCount();
+    unsigned num_using_directives = nsinfo->GetUsingDirectiveCount();
+
+    if (num_decls > 0 || num_using_directives > 0) {
+      if (num_candidates == 0) {
+        full_decl_count = num_decls;
+        best_decl = decl;
+        best_using_directives = num_using_directives;
+      } else {
+        full_decl_count += num_decls;
+        if ((best_using_directives == 0) && (num_using_directives > 0)) {
+          // Prefer a namespace with using directives over one
+          // without - forward declaration namespaces shouldn't have
+          // using directives
+          best_decl = decl;
+          best_using_directives = num_using_directives;
+        } else if ((best_using_directives > 0) && (num_using_directives == 0)) {
+          // stick with the one we have
+        } else {
+          // we have multiple namespaces with using directives
+          // or multiple with just declarations
+          best_decl = nullptr;
+        }
+      }
+      num_candidates++;
+    }
+  }
+
+  if (num_candidates == 1) {
+    int line = GetLineNumber(best_decl->getBeginLoc());
+    VERRS(7) << "Single namespace for " << ns_ident << " "
+             << GetFilePath(best_decl) << " line " << line << " with "
+             << full_decl_count << " full decls\n";
+    return best_decl;
+  } else if (num_candidates == 0) {
+    // No full declarations. Return the input NamespaceDecl
+    int line = GetLineNumber(ns->getBeginLoc());
+    VERRS(7) << "No namespace with a full decl for " << ns_ident << " "
+             << GetFilePath(ns) << " line " << line << "\n";
+    return ns;
+  }
+
+  if (best_using_directives > 0) {
+    int line = GetLineNumber(best_decl->getBeginLoc());
+    VERRS(7) << "Multiple namespace decls for " << ns_ident
+             << ", but only one with using directives "
+             << GetFilePath(best_decl) << " line " << line << "\n";
+    return best_decl;
+  }
+
+  VERRS(7) << "Multiple namespace decls for " << ns_ident << " with "
+           << full_decl_count << " full decls. Ignoring namespace use\n";
+  return nullptr;
+}
+
+void IwyuFileInfo::ReportDeclInNamespaces(const NamedDecl* decl) {
+  // Ignore declarations of sub-namespaces i.e. don't count namespace
+  // declarations themselves as a forward decl or full decl.
+  if (isa<NamespaceDecl>(decl))
+    return;
+
+  bool fwd_decl = IsForwardDecl(decl);
+  bool using_directive = isa<UsingDirectiveDecl>(decl);
+
+  if (fwd_decl && !using_directive)
+    return;
+
+  for (const DeclContext* decl_context = decl->getDeclContext(); decl_context;
+       decl_context = decl_context->getParent()) {
+    // Stop traversing up the context - function arguments, structure
+    // members and variables in block scope don't need to be recorded,
+    // as the function/structure will be.
+    //
+    // TODO: others?
+    if (isa<FunctionDecl, TagDecl, BlockDecl>(decl_context)) {
+      return;
+    }
+
+    if (const auto* ns = dyn_cast<NamespaceDecl>(decl_context)) {
+      // This is the namespace within which the used symbol is declared
+
+      // Get full namespace string...
+      string ns_ident = ns->getQualifiedNameAsString();
+      const FileEntry* declared_in = GetFileEntry(ns);
+
+      // Get or create file_info...
+      IwyuNamespaceInfo* nsinfo = getNamespaceInfo(ns, declared_in);
+
+      if (const auto* using_decl = dyn_cast<UsingDirectiveDecl>(decl)) {
+        VERRS(7) << "Recording using directive in namespace " << ns_ident
+                 << " for " << using_decl->getNominatedNamespace()->getName()
+                 << " in filepath " << GetFilePath(ns) << "\n";
+
+        nsinfo->AddUsingDirective();
+        // Only count using directives in their enclosing namespace.
+        // It shouldn't influence how parent namespaces are used.
+        break;
+      }
+
+      VERRS(7) << "Recording declaration in namespace " << ns_ident << " for "
+               << decl->getName() << " in filepath " << GetFilePath(ns) << "\n";
+
+      if (!fwd_decl) {
+        nsinfo->AddFullDecl();
+        // Count normal declarations in all parent namespaces - so don't
+        // break here.
+      }
+    }
+  }
 }
 
 namespace internal {
@@ -2214,6 +2406,16 @@ void IwyuFileInfo::ResolvePendingAnalysis() {
                                 /* flags */ UF_None,
                                 "(for un-referenced using)");
       }
+    }
+  }
+
+  // For each namespace use, report against the best NamespaceDecl.
+  for (const auto& use : namespace_uses) {
+    const NamespaceDecl* decl = GetBestNamespaceDecl(std::get<1>(use));
+    if (decl != nullptr) {
+      const optional<string>& comment = std::get<3>(use);
+      ReportFullSymbolUse(std::get<0>(use), decl, std::get<2>(use),
+                          comment ? comment->c_str() : nullptr);
     }
   }
 }
--- a/iwyu_output.h
+++ b/iwyu_output.h
@@ -17,6 +17,7 @@
 #define INCLUDE_WHAT_YOU_USE_IWYU_OUTPUT_H_
 
 #include <map>                          // for map
+#include <optional>                     // for optional
 #include <set>                          // for set
 #include <string>                       // for string, operator<
 #include <vector>                       // for vector
@@ -36,6 +37,7 @@ class ElaboratedTypeLoc;
 namespace include_what_you_use {
 
 using std::map;
+using std::optional;
 using std::set;
 using std::string;
 using std::vector;
@@ -222,6 +224,35 @@ class OneIncludeOrForwardDeclareLine {
   const clang::NamedDecl* fwd_decl_ = nullptr;
 };
 
+// Class to collect information about declarations within a namespace.
+class IwyuNamespaceInfo {
+  const clang::NamespaceDecl* decl;
+
+  unsigned full_decl_count = 0;
+  unsigned using_directive_count = 0;
+
+ public:
+  IwyuNamespaceInfo(const clang::NamespaceDecl* decl) : decl(decl) {
+  }
+
+  const clang::NamespaceDecl* GetNamespaceDecl() const {
+    return decl;
+  }
+  unsigned GetFullDeclCount() const {
+    return full_decl_count;
+  }
+  unsigned GetUsingDirectiveCount() const {
+    return using_directive_count;
+  }
+
+  void AddFullDecl() {
+    ++full_decl_count;
+  }
+  void AddUsingDirective() {
+    ++using_directive_count;
+  }
+};
+
 // This class holds IWYU information about a single file (FileEntry)
 // -- referred to, in the comments below, as "this file."  The keys to
 // most of these methods are all quoted header paths, which are the
@@ -304,6 +335,13 @@ class IwyuFileInfo {
                           const clang::UsingDecl* using_decl,
                           UseFlags flags, const char* comment);
 
+  // Called whenever a NamespaceDecl is used, generally via a using
+  // directive or an alias (a use via a symbol will force the right
+  // include via that symbol).
+  void ReportNamespaceDeclUse(clang::SourceLocation use_loc,
+                              const clang::NamespaceDecl* decl, UseFlags flags,
+                              const char* comment);
+
   // This is used when we see a // NOLINT comment, for instance.  It says
   // '#include this header file as-is, without any public-header mapping.'
   // Input is the include-line as desired: '<string.h>' or '"ads/foo.h"'.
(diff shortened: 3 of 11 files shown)


kimgr:
Wow, what an ambitious patch! It's going to take me a while to chew through, but it looks really nice and clean at first browse-through.

kilroyd:
Pushed an updated series:

* fixing issues reported by CI.
* improve test to use nested namespaces
* handle nested namespaces

bolshakov-a:
I haven't taken closer look at your changes, but I've just recalled that I also have a draft commit addressing the namespace issue, FYU:
https://github.com/bolshakov-a/include-what-you-use/commit/e1d6996ea187d1350ff1e5bf2079b0be0563f3f7

kimgr:
Sorry, I've been swamped with life in the past weeks, but I'm still hoping to get to reviewing this soon. My gut instinct says it shouldn't be it's own module (`iwyu_namespace.*`), but I see how the additional data structures would probably take up a lot of conceptual space in the existing modules. Let me see if I can get my head around it and think about alternatives. (@bolshakov-a's patch looks much smaller, so it would be interesting to hear if they solve different problems)

kilroyd:
I just had a look at @bolshakov-a's commit. I had to think a bit about what it's doing. If I understood correctly:

* if we have a namespace-use (i.e. namespace directive) ignore it if we find any other identifier use from a sub-namespace. We can do this on the basis that the other identifier use will necessarily require an appropriate header inclusion. My commit effectively resorts to this in the 'we have many namespace decls with declarations in them' case. This also works for the namespace alias support as added by my first commit in this PR.

My commit is _much_ heavier handed (tracking how many decls in each `NamespaceDecl` we see) - but helps in the case where there is no other identifier use ++. It can still suggest an appropriate header. One other affect of my commit is that it will still note that we get the namespace from a particular header in the include comments.

For the common case, I think Andrey's commit is good. We could commit both to get a good suggestion for the namespace when it's otherwise unused, though I'm not sure the extra logic is worth it.

> My gut instinct says it shouldn't be it's own module (iwyu_namespace.*), but I see how...

Agreed on both points. I think the separate file makes sense if we feel like the algorithm to select the best file needs to get more complicated.

I looks forward to hearing what you think.

---
++ With no other identifier use, you might think that the namespace is unused (and maybe we should report that). It's the case for a namespace alias, but not a namespace directive. Consider:

```
headerA.h:
namespace A {
    void functionA();
}

headerB.h:
namespace B {
   using namespace A;
   void functionB();
}

fileC.cc:
namespace C {
  // is this used?
  using namespace B;

  void myFunc() {
     // namespace B pulls in namespace A, so I can call functionA()!
     // So the using namespace B above is 'used', but just by considering
     // the decl's in use, you will only detect namespace A.
     functionA();
  }
}
```

I had been looking at whether I could detect this use of the namespace directive, but it involves detecting when the modified namespace search path has an affect on the symbol lookup.

There's further issues if we push that example further though. What if the `using namespace A` only occurs in a specific header, and that's not the one we suggest. My commit could be modified to further prioritise a `NamespaceDecl` containing using directives.

I think this example is degenerate, and you shouldn't be relying on the `using namespace B` to get a symbol from `A` - but I guess it's possible for a library to do this to re-export an internal namespace.

Apologies this comment ended up much longer than I planned...

bolshakov-a:
The first commit LGTM, at least conceptually.

Regarding to the second commit, I agree that it is annoying when IWYU suggests to include a header for the stuff (using-directive, e. g.) which you just forgot to remove. But, maybe, it should be concern of some other tool?

As for the fix of the original issue (the 3rd commit), honestly, it seems redundant to me to add some complex logic just for attribution to the most appropriated header in explanatory comments. But I like the idea that it should be in the pending analysis. Maybe, just make it more similar to `using_decl_referenced_` handling, like you've done for namespace aliases?
> What if the using namespace A only occurs in a specific header, and that's not the one we suggest.

Good catch! I suspect it may be a tricky issue which may be addressed some day later. But, probably, your tracking of declarations inside namespaces actually forms a good framework to this?

Concerning my commit, feel free to cherry-pick it or just to take inspiration from it, if you wish.

iwyu.cc
```diff
@@ -1699,6 +1700,17 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       ReportTypeUse(used_loc, type);
   }
 
+  void ReportNamespaceAliasUse(SourceLocation used_loc,
```
kimgr, line 1708: Can we skip the `Track unused namespace aliases` commit for now? I'm not sure if it's trying to do some additional cleanup of unused aliases or if it's a foundation for IWYU analysis. If the former, I don't think IWYU should do it. If the latter, it makes sense to tackle separately, I think.

iwyu.cc
```diff
@@ -2578,6 +2578,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // Visitors defined by BaseAstVisitor.
 
   bool VisitNestedNameSpecifier(NestedNameSpecifier* nns) {
+    // If this is a NamespaceAlias, then mark the alias as used,
```
kimgr, line 2609: Hmm, out of curiosity, why isn't the check for `NestedNameSpecifier::Namespace` necessary here too?

iwyu_namespace.cc
```diff
@@ -0,0 +1,249 @@
...
+
+  bool fwd_decl = IsForwardDecl(decl);
+
+  for (; decl_context; decl_context = decl_context->getParent()) {
+    // Stop traversing up the context - function arguments, structure
+    // members and variables in block scope don't need to be recorded,
+    // as the function/structure will be.
+    //
+    // TODO: others?
+    if (const clang::FunctionDecl* func = DynCastFrom(decl_context)) {
```
kimgr, line 196: Are these boundary conditions necessary? In the majority of cases, I think there might be a `RecordDecl` at most between these and the `TranslationUnitDecl`, so if it's an optimization to exit early, I don't think it's going to buy us very much.

Or does it have any functional impact?

iwyu_namespace.cc
```diff
@@ -0,0 +1,249 @@
...
+           << full_decl_count << " full decls. Ignoring namespace use\n";
+  return nullptr;
+}
+
+void RecordDeclNamespaces(const NamedDecl* decl) {
+  const NamespaceDecl* ns;
+
+  // Ignore declarations of sub-namespaces i.e. don't count them as a
+  // forward decl or full decl.
+  if ((ns = DynCastFrom(decl)))
```
kimgr, line 183: Prefer `isa<NamespaceDecl>(decl)` here

kimgr (commented):
I've turned this inside out a few times, and I think I roughly see what's going on. It looks like you're maintaining a structure like this:
```
{
    qualified-name: {
        filename: fileinfo(decl, full_count, fwd_count),
        filename: fileinfo(decl, full_count, fwd_count),
        filename: fileinfo(decl, full_count, fwd_count),
    },
    qualified-name: {
        filename: fileinfo(decl, full_count, fwd_count),
        filename: fileinfo(decl, full_count, fwd_count),
        filename: fileinfo(decl, full_count, fwd_count),
    }
}
```

I'm pretty sure you don't need the qualified-name in the mix, and I'm not entirely sure you need the keying by filename. I would suggest something like:
```
{
    canonical_decl: {decl: count},
    canonical_decl: {decl: count},
    canonical_decl: {decl: count},
}
```
instead.

* `NamespaceDecl::getCanonicalDecl()` should return a stable identity pointer for a namespace declaration chain. So for any `NamespaceDecl` you can call `getCanonicalDecl` and get something that identifies the namespace just like its qualified-name. I think.
* The forward-decl count is never used, so I suggest we get rid of it
* I don't think the filepath component of the index is necessary (happy to be proven wrong!), so you should be able to just use `GetFilePath(namespace_decl)` for diagnostics and whatnot.
* The `nsmap` appears to be a cache for `GetBestNamespaceDecl`. Let's skip that for now, unless performance suffers unbearably.
* I suspect those changes are going to collapse enough that the namespace use-count map could hang off `IwyuFileInfo` just like the `UsingDecl` used map. As far as I can tell it's not absolutely necessary that the namespace map is global. There might be some redundancy between different files, but maybe it also gives more granular results if broken down by use-file?
* Some style nits here and there, but we can discuss those once the bigger things have been ironed out.

Thanks!

iwyu.cc
kilroyd: Sure. This isn't a foundation for IWYU analysis.

Having played with the namespace reporting, I feel that as long as IWYU suggests a header to satisfy a reference to a namespace, then in order minimise unused headers IWYU also needs to identify unused aliases and using directives. I've got had some ideas on how we might be able to do this better.

iwyu.cc
kilroyd: I'm afraid you've lost me. This hunk is about detecting when an identifier is preceded by an explicit namespace alias, in which case we want to report the use of the alias.

As far as I understand it, we can't detect using-directive usage here (at least not in a way that tracks usage completely).

iwyu_namespace.cc
kilroyd: This has a functional impact on the count of declarations made within a namespace. `RecordNamespaceDecls()` is being called for at least function arguments and structure fields. I believe it's getting called for variable declarations within function scope too.

So these aborts ensure that a function within a namespace only results in an increment of 1 (on all parent namespaces), similarly a struct.

Things _may_ work ok with these removed, but it's easier to validate the counts in the error log if they conform to what you'd expect.

kilroyd (commented):
> It looks like you're maintaining a structure like this:

Correct.

> I'm pretty sure you don't need the qualified-name in the mix, and I'm not entirely sure you need the keying by filename. I would suggest something like ... instead

I'm not quite sure what you're proposing here... I'll need to have a think - it's been a couple weeks since I wrote it, but my initial reaction is that structure isn't sufficient.

> NamespaceDecl::getCanonicalDecl() should return a stable identity pointer for a namespace declaration chain. So for any NamespaceDecl you can call getCanonicalDecl and get something that identifies the namespace just like its qualified-name

Yes, `getCanonicalDecl()` can be used to identify the namespace like qualified-name. So we could swap the qualified-name index for canonical-decl. Other than that, the canonical-decl isn't useful.

We still need to record declaration counts separately for each file during ?ast travesal? so we can decide which is best in post-traversal analysis.

I'll see what I can do to address your comments...

iwyu.cc
kimgr: Right, so:
```
using alias = std;

void foo() {
    alias::vector<int> x;  // hits your VisitNestedNameSpecifier check
    std::vector<int> x;  // why don't we want to check this in VisitNestedNameSpecifier too?
}
```

iwyu_namespace.cc
kimgr: Side note: you can replace the three `DynCastFrom` with `isa<FunctionDecl, TagDecl, BlockDecl>(decl_context)`

kimgr:
> Yes, getCanonicalDecl() can be used to identify the namespace like qualified-name. So we could swap the qualified-name index > for canonical-decl. Other than that, the canonical-decl isn't useful.

Right; but it's much cheaper to carry and compare than a string.

> We still need to record declaration counts separately for each file during ?ast travesal? so we can decide which is best in
> post-traversal analysis.

You could record declaration counts for each _decl_ during AST traversal, and then aggregate for files post-traversal. So during traversal we could pick up:
```
canonical_decl: [(decl, count), (decl, count), (decl, count)]
```
all of which are register-sized and cheap to get. Then before analysis, you could transform the resulting set into a per-file breakdown, possibly precomputing the best decl for every canonical decl.

Lookup is then just `best_decls[ns->getCanonicalDecl()]`. Feels like that should be doable, but I haven't gotten around to trying it.

iwyu.cc
kilroyd: Note that this hunk will be removed once we drop commit 2 i.e. warning on unused aliases.

In VisitNestedNameSpecifier() we're able to identify alias use (and therefore spot unused aliases).

We could consider std::vector too, and try to identify potential using-directives that are being used. But this won't be able to catch all uses appropriately. std::vector specifically is generally not 'using' any using-directive. As I think I've mentioned previously, identifying where a using-directive is used requires hooking into the symbol lookup infrastructure of clang (Sema/SemaLookup.cpp).

iwyu_namespace.cc
kilroyd: Yes, that's right.

iwyu.cc
kilroyd: > Note that this hunk will be removed once we drop commit 2

I got confused between the commits - this was in the first commit so still being added.

kilroyd:
I've incorporated most of the changes from the kg-namespaces branch. Thanks for the heap cleanup - my background is C, so I fall back to using the heap when I can't figure out the C++ way to do things :)

Also made tweaks to the algorithm to track using directives, and prefer namespaces with them.

Not done:
* Still have filepath key in the map - I think we need to keep this, so we count the declarations within a file across multiple namespace blocks (with the same qualified name). I've expanded comments to hopefully explain better
* I dropped your commit with early out on forward declaration in RecordDeclNamespaces(). I had tried this independently too. I think we can do this, but at the moment I'm relying on the verbose output printing forward decl info. It has the effect of not creating an IwyuNsFileInfo which has a subtle effect on the reporting in getBestFile(). I expect to take another look. 

Still todo:
* Consider merging IwyuNsFileInfo into IwyuFileInfo. Global ns info vs per file.
* Double check behaviour vs Andrey's commit
* testcases for:
    * prefer namespace with using directives
    *  ??

iwyu.cc
kimgr: Yeah, I'm not saying we should do anything, I was just curious why alias uses were different from other nested name-specifiers. Maybe because a namespace nns ultimately refers to a symbol that is guaranteed to live in the same file as the namespace. The alias, on the other hand, could have been included from somewhere that has nothing to do with the aliased namespace/symbol. OK, that makes sense to me, whether it's correct or not :)

iwyu_namespace.h
```diff
@@ -0,0 +1,26 @@
...
+namespace clang {
+class NamedDecl;
+class NamespaceDecl;
+}  // namespace clang
+
+namespace include_what_you_use {
+void RecordDeclInNamespaces(const clang::NamedDecl *decl);
+
+const clang::NamespaceDecl *GetBestNamespaceDecl(
+    const clang::NamespaceDecl *ns);
```
kimgr, line 23: We bind the pointer asterisk to the type in IWYU (because Google conventions). Clang-format should have fixed that.

kimgr, line 23: I would remove the parameter name `ns` here, because that would make the whole declaration fit on one 80-col line.

iwyu_output.cc
```diff
@@ -721,6 +722,15 @@ void IwyuFileInfo::ReportUsingDeclUse(SourceLocation use_loc,
   ReportFullSymbolUse(use_loc, using_decl, flags, comment);
 }
 
+void IwyuFileInfo::ReportNamespaceDeclUse(SourceLocation use_loc,
+                                          const NamespaceDecl* decl,
+                                          UseFlags flags, const char* comment) {
+  // In order to assign Namespace usage to the best file, don't record the
```
kimgr, line 728: "Namespace" -> "namespace"

iwyu_output.cc
```diff
@@ -2204,6 +2214,14 @@ void IwyuFileInfo::ResolvePendingAnalysis() {
       }
     }
   }
+
+  for (NamespaceDeclUse& use : namespace_uses) {
+    const NamespaceDecl* decl = GetBestNamespaceDecl(use.decl);
+    if (decl != nullptr) {
+      // For each namespace use, report against the best NamespaceDecl
```
kimgr, line 2221: I'd move this comment to just before the loop. Also, add a full stop to end the sentence.

iwyu_output.cc
```diff
@@ -2204,6 +2214,14 @@ void IwyuFileInfo::ResolvePendingAnalysis() {
       }
     }
   }
+
+  for (NamespaceDeclUse& use : namespace_uses) {
```
kimgr, line 2218: Can be `const`

iwyu_output.h
```diff
@@ -216,6 +216,13 @@ class OneIncludeOrForwardDeclareLine {
   const clang::NamedDecl* fwd_decl_;
 };
 
+struct NamespaceDeclUse {
```
kimgr, line 219: Could you add a little doc comment to this type? Just something around how we delay uses of namespace decls and finalize them in ResolvePendingAnalysis.

iwyu_output.h
```diff
@@ -297,6 +304,13 @@ class IwyuFileInfo {
                           const clang::UsingDecl* using_decl,
                           UseFlags flags, const char* comment);
 
+  // Called whenever a NamespaceDecl is used, generally via a using
+  // directive or an alias (a use via a symbol will force the right
+  // include via that symbol)
```
kimgr, line 309: Full stop at end of comment

tests/cxx/multiple_decl_namespace-d1.h
```diff
@@ -0,0 +1,21 @@
...
+namespace test::ns3 {
+class Class3;
+}
+
+namespace test::ns1 {
+typedef unsigned Flags;
+const Flags F_None = 0;
+const Flags F_All = 1;
+
+void function1(const test::ns3::Class3 &c, Flags f);
```
kimgr, line 20: Let's use IWYU naming conventions in tests too; functions are `PascalCase`. Here and throughout the new tests

tests/cxx/multiple_decl_namespace.cc
```diff
@@ -0,0 +1,55 @@
...
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+// Tests that we handle multiply declared namespaces correctly.
+//
+// Where declarations of a namespace occur in multiple headers (either
+// because forward declarations were needed or otherwise), iwyu should
```
kimgr, line 15: I try to spell `iwyu` as `IWYU` in all new code/docs, so could you stick to that here and below?

tests/cxx/multiple_decl_namespace.cc
```diff
@@ -0,0 +1,55 @@
...
+  function2b();
+
+  // The first declaration of ns3 is seen in d1.h, but the
+  // declaration that we are using is in d3.h. We should _not_
+  // recommend including d1.h for ns3.
+}
+
+/**** IWYU_SUMMARY
+
+(tests/cxx/multiple_decl_namespace.cc has correct #includes/fwd-decls)
```
kimgr, line 53: I try to avoid succeeding testcases like this, because a lot of subtlety and information is lost. See https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#force-diagnostics.

See if you can move the symbols out of `-dx.h` headers and into `-ix.h` headers so IWYU is forced to complain about the indirect includes.

iwyu_namespace.h
kimgr: Turns out our `.clang-format` didn't configure this correctly, I pushed a fix to master, so if you format this again, it should align the stars correctly.

iwyu_output.h
```diff
@@ -216,6 +216,13 @@ class OneIncludeOrForwardDeclareLine {
   const clang::NamedDecl* fwd_decl_;
 };
 
+struct NamespaceDeclUse {
+  clang::SourceLocation used_loc;
+  const clang::NamespaceDecl* decl;
+  UseFlags flags;
+  const char* comment;
```
kimgr, line 223: I wonder if this needs to become a `std::string` to avoid lifetime issues. We mostly call `ReportXXXUse` with literal strings, but should someone add something like a concatenated string and use `c_str()`, this would refer to a dead value.

In the current flow, that doesn't matter, because the comment will be copied into `OneUse`, and materialized in a new `stdd:string`, but this intermediate storage is a little risky.

kilroyd: Good spot. I'll push an update fixing this and the majority of the issues you've raised. I still need to have a go at moving the logic to `IwyuFileInfo` which I'm struggling to find the time for. Maybe later this weekend🤞?

kimgr: Thanks! I've been a little occupied with other things as well, so definitely no rush from my side. Let's percolate on it until it feels stable and nice.

headhunter45:
I'm probably misunderstanding this, but if I have ```#include <string>
using namespace std;``` at the top of a file and I don't actually use std::string or anything else from the std namespace what will this do? I'd want it to remove both the include and the using statement, but I think having the using would keep the `#include <string>` in this case. and IWYU wouldn't remove the using because I guess that's out of scope for the tool?

kimgr:
@headhunter45 Yes, exactly. Otherwise it's hard to know when to stop - if you have an unused static function in your file that uses `std::string`, do we remove that? Etc.

IWYU's focus is the relationship between symbols and files, and I think it's nice to stay within those fences.

That said, it would be cool to see some clang-tidy checks to do e.g. unused `using` cleanup. Then IWYU would be free to remove the leftover headers.

iwyu_output.h
```diff
@@ -320,6 +368,10 @@ class IwyuFileInfo {
     return direct_includes_as_fileentries_;
   }
 
+  void RecordDeclInNamespaces(const clang::NamedDecl* decl);
```
kimgr, line 371: I understand that `RecordDecl` is a verb phrase here ("record the declaration"), but `RecordDecl` is also a type in Clang ("a record declaration"). Would be nice to find a different word than `Record` :-)

iwyu_output.h
```diff
@@ -417,6 +477,30 @@ class IwyuFileInfo {
...
+  //     { fileinfo : { first_ns_decl, decl_count, directive_count },
+  //       fileinfo : { first_ns_decl, decl_count, directive_count },
+  //     },
+  //   canonical_namespace_decl :
+  //     { fileinfo : { first_ns_decl, decl_count, directive_count },
+  //       fileinfo : { first_ns_decl, decl_count, directive_count },
+  //     },
+  //   etc
+  // }
+  map<const clang::NamespaceDecl*, map<const IwyuFileInfo*, IwyuNamespaceInfo>>
```
kimgr, line 502: This can be a `FileEntry*` for simpler implementation.

iwyu.cc
```diff
@@ -1700,8 +1701,25 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    const FileEntry* used_in = GetFileEntry(used_loc);
+    preprocessor_info().FileInfoFor(used_in)->ReportNamespaceDeclUse(
+        used_loc, namespace_decl, extra_use_flags, comment);
+  }
+
   //------------------------------------------------------------
   // Visitors of types derived from clang::Decl.
+  bool VisitNamedDecl(clang::NamedDecl* decl) {
+    preprocessor_info()
+        .FileInfoFor(preprocessor_info().main_file())
```
kimgr, line 1719: This use of `main_file()` was surprising to me. I tried to replace it with `CurrentFileEntry()`, but that caused test failues.

I think I understand now that this is why the namespace info was global before? Because it's statistics over the whole compilation unit, not really keyed by file?

kilroyd: To do this the statistics must be recorded per namespace. Under that namespace we subkey by file, allowing us to identify the best file for the namespace.

If we reverse the keys, recording stats on each namespace declared within a file, then we still need something to tell us which files a given namespace is in, or we have to iterate over all files for each namespace we want to find.

So this does look a bit odd, with the `nsinfoPerFile` map is in the `IwyuFileInfo`. If processing multiple translation units in one run, this should make the namespace processing independent for those translation units, which is a good thing.

kimgr: Sorry, I'm not following. You keep saying "this" about something that sounds interesting, but I'm not sure what it is :)

But yeah, the way you put all statistics about namespaces into the main file make me think that moving the state to `IwyuFileInfo` was not the right move.

> If processing multiple translation units in one run, this should make the namespace processing independent
> for those translation units, which is a good thing (not sure if IWYU actually does that).

IWYU only does one translation unit at a time.

But even if it did multiple, I think putting this state into the AST visitor might be a better fit. Let me play around with a refactor for the UsingDecl delayed reporting, I think there may be symmetries/synergies...

kilroyd: > Sorry, I'm not following. You keep saying "this" about something that sounds interesting, but I'm not sure what it is :)

My apologies, I've assumed we're on the same page. I guess this PR has been open a while.

> To do this the statistics...

Here I was refering to the PR, 'this' being "finding the best include file for a given namespace".

> So this does look a bit odd...

This instance refers to the current commit. :)

I'll wait for further feedback. I can back up to a version with all the changes except the move into `IwyuFileInfo`. Just let me know your preference.

kimgr: > Let me play around with a refactor for the UsingDecl delayed reporting, I think there may be symmetries/synergies...

Nope, the UsingDecl handling seems slightly different. It seems to keep track of uses within the same file. And it's not remembering uses to replay them later, it has enough information in the UsingDecl to create the uses from scratch. Or at least that's what it does :). Nevermind.

kilroyd: FYI, I'm reworking this commit so the `IwyuNamespaceInfo` is stored in `IwyuFileInfo` lists as you might expect. The `nsPerFile` map be static and point to these entries, allowing us to do the lookup during `GetBestNamespaceDecl` as we need to.

kimgr: Sorry for the prolonged silence, I've been alternating between busy with work and vacationing. I'm actually not sure anymore that the namespace statistics should go in `IwyuFileInfo`, given how they're global across files.

I'm a little cautious around static members, because they tend to cause trouble with reentrancy (and threading, which is not an immediate concern here), but let me take a look at your latest patch and see if I can get my head around it.

kimgr:
It hadn't occurred to me that the two commits are pretty much unrelated, right? Maybe it would make sense to split this PR up into two with one commit each. It's hard to review these large PRs, particularly with many new tests.

iwyu.cc
```diff
@@ -183,6 +183,7 @@ using clang::LValueReferenceType;
 using clang::LinkageSpecDecl;
```
kimgr, line 183: Commit message: please use "IWYU" over "iwyu" consistently.

iwyu_output.h
```diff
@@ -222,6 +224,45 @@ class OneIncludeOrForwardDeclareLine {
   const clang::NamedDecl* fwd_decl_ = nullptr;
 };
 
+// Namespace usage reporting is delayed until ResolvePendingAnalysis.
+// This structure stores the arguments to be passed to
+// RecordFullSymbolUse.
+struct NamespaceDeclUse {
```
kimgr, line 230: I think we can just use `OneUse` to buffer pending namespaces uses, instead of this custom type

kilroyd:
> It hadn't occurred to me that the two commits are pretty much unrelated, right?

TBH I'd forgotten there were 2 commits. They're both modifying reporting on namespace usage. If we commit the reporting on namespace aliases without improving the suggested filenames, then we end up with more inappropriate headers being suggested for a given project. If you want them separate, I'll drop the namespace alias commit.

kimgr:
@kilroyd I think that's fine; they solve independent problems that are both increments toward better behavior.

I have some review feedback on the first (namespace alias, right?) commit, but that's almost 100% ready to be merged, so it seems like a shame to block it on the the second (use-stats-based namespace attribution), which is more complex.

Could you pull the namespace alias commit into its own PR? I'd also be happy to adopt it and do the grunt work if you prefer.

kimgr:
I have some spare time for opensource work now, so if you don't mind, I'll play around with this patch set and experiment a little. I think that's more likely to make progress than bouncing review comments back and forth. Won't merge anything without running it by you first, we'll see how far I get :)

iwyu_output.h
kilroyd: You're right - It does hold much of the same data, but has quite a bit more too. I could see this being confusing. Maybe a tuple might be prefered here?

kimgr:
I've pulled on various strings and have something running now where:

* A new class `IwyuNamespaceStats` is added to `iwyu_output.{h,cc}`
* The class encapsulates the two-level map and exposes an `Update` method (previously `ReportDeclInNamespace`) and a `GetBestNamespaceDecl`, as before
* An `IwyuNamespaceStats` is held in the visitor state in `iwyu.cc`
* A reference to that stats object is passed to `ResolvePendingAnalysis`
* That way no globals are necessary, and `iwyu.cc` is mostly responsible for accounting namespace stats, whereas `iwyu_output` is responsible for reporting namespace uses based on those stats
* I can share the branch when I've cleaned it up a little

However, now I'm looking at the logic for `GetBestNamespaceDecl`, and I'm not sure I understand the rationale... For example;

* the decl count does not appear to be used
* I'm trying to get my head around this comment in the test:
```
// This namespace has a using-directive. This is a pretty clear sign
// that this namespace block is not a bunch of forward declarations,
// and need to be explicitly included by users.
```
-- why is that a clear sign?
* I was expecting `GetBest...` to do some kind of scoring, but it really seems to start arbitrarily with the first file, and then focusing primarily on the number of using-directives in the namespaces. Is it possible to clarify a little around the rules for forward-declarations, using-directives and multiple namespaces with usings and decls?

kilroyd:
>     * the decl count does not appear to be used
> 
>     * I'm trying to get my head around this comment in the test:
> 
> 
> ```
> // This namespace has a using-directive. This is a pretty clear sign
> // that this namespace block is not a bunch of forward declarations,
> // and need to be explicitly included by users.
> ```
> 
> -- why is that a clear sign?
> 
>     * I was expecting `GetBest...` to do some kind of scoring, but it really seems to start arbitrarily with the first file, and then focusing primarily on the number of using-directives in the namespaces. Is it possible to clarify a little around the rules for forward-declarations, using-directives and multiple namespaces with usings and decls?

I had to double check what we have in `GetBestNamespaceDecl`. The function encapsulates the decision logic, and we should be able to modify it to do whatever we want. I've tried to keep the logic as simple as possible, and from the look of it, avoiding incorrect suggestions.

Currently the logic is:
1. Only files declaring a namespace that has full declarations or using directives are candidates.
2. A file declaring a namespace containing a using-directive is chosen over a namespace without.
3. If more than one file declares a namespace with a using-directive, we don't try to pick one.
4. If more than one file declares a namespace with full declarations, we don't try to pick one.

A namespace with only forward declarations should be dropped at 1.

Next, consider two coding styles:
* single header containing all the interfaces within a given namespace
* multiple headers defining the content of a namespace, each defining a different aspect.

We can't handle the multiple header case without tracking which declarations we are actually using from each namespace declaration. This is equivalent to 4. (and probably 3.) - we avoid picking a header. This is why the full declaration count is not used in choosing a header. It would result in always choosing the file with the 'most declarations' or widest interface, but that may not be what is being used.

A using-directive within a namespace modifies the symbol lookup within that namespace.

```namespace a::b::c::d {
struct m;
}

namespace a::e {
  using namespace a::b::c::d;
  // symbol m is now available in the namespace a::e (as is everything else in a::b::c::d)
}

// Not limited to the scope
a::e::m x;

namespace a::e {
   // and still available in later scope
   m y;
}
```

https://godbolt.org/z/1aWsnv6W1

The persistence of the effect of the using-directive outside the namespace scope significantly changes how you would refer to `a::b::c::d` between picking up a namespace with the using-declaration vs without. This is essentially why I added the prioritisation of using-directives.

Thinking about this again, if this were an internal namespace that was being re-exposed then you would never have used `a::b::c::d` anyway... and it's possible that you only need to `a::b::c::d` with this partial interface. Maybe we should drop the using-directive check...

kimgr:
> We can't handle the multiple header case without tracking which declarations we are actually using from each namespace declaration.

So I guess I'm wondering: why can't we do this? Just intuitively, it seems more relevant to look at usage statistics than definition statistics, but I don't know if there's any concrete pros/cons...

kilroyd:
This has been dragging on a bit, and I haven't had time to address the issues. I propose dropping this RFC, and instead will push a commit that mostly ignores namespace declarations, along the lines of the @bolshakov-a 's commit. That should be sufficient for my interests, is probably good enough for 90% of cases, and is much simpler.

kimgr:
@kilroyd Alright, thanks for investing so much energy this far. I think there's some gold in this approach, but it's not 100% clear to me how to isolate the "best" namespace. Intuitively I don't think number of declarations is it, and I'm not sure if number of uses is much better. I'll see if I can muster time and energy to get back to this later.

kilroyd:
I'll leave the namespaces branch in my repo in case it's useful to resurrect in future

PR #1218, opened 2023-03-23
Macos run_iwyu_test.py improvements

kilroyd:
I've been working on MacOS for the namespace series that I raised. I've got 10 `run_iwyu_test.py` failures, so this series is an attempt to clean some of them up. The main problems stems from using `libstdcpp` mappings instead of `libcxx`.

The primary resolution is to add an internal mapping for `libcxx`, and to determine the selected C++ standard library at runtime. I've also updated the `libcxx.imp` mappings for what's in `libc++` 17, with some backwards compatibility for MacOS's `libc++` 13 (which is actually what I'm testing against - IWYU isn't happy when pointed at the standalone libc++ includes).

A secondary resolution is to allow the user to select canonical mappings using clang's newish `Tooling/Inclusion` library. See that commit for details. This seems to work pretty well, and is handy for double checking against the libcxx behaviour.

The final commit in the series adds a couple of tweaks to play nicer with MacOS `libc`, but we may want to drop this commit to do this 'properly' (I'm not sure how we'd want to do this)

After this series, the number of test failures is down to 5. I've also run the resulting IWYU on its source code, and have satisified myself that the suggestions in both `libcxx` and canonical mode are sensible. The one outlier is that IWYU suggests removing `<algorithm>` from `iwyu_stl_util.h` (as it doesn't get any full uses of `std::find`), which seems wrong but I'm not familiar enough with the template corner cases :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,6 +173,8 @@ else()
     # [1] https://llvm.org/viewvc/llvm-project?view=revision&revision=348907
     # [2] https://llvm.org/viewvc/llvm-project?view=revision&revision=348915
     clangSerialization
+
+    clangToolingInclusionsStdlib
   )
 endif()
 
--- a/iwyu_driver.cc
+++ b/iwyu_driver.cc
@@ -19,6 +19,7 @@
 #include <string>
 #include <utility>
 
+#include "iwyu_globals.h"
 #include "llvm/ADT/SmallString.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/Option/ArgList.h"
@@ -32,6 +33,7 @@
 #include "clang/Driver/Compilation.h"
 #include "clang/Driver/Driver.h"
 #include "clang/Driver/Tool.h"
+#include "clang/Driver/ToolChain.h"
 #include "clang/Frontend/CompilerInstance.h"
 #include "clang/Frontend/CompilerInvocation.h"
 #include "clang/Frontend/FrontendAction.h"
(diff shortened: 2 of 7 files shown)


kimgr:
Interesting, I hadn't seen the ClangToolingInclusion library. It would be nice if we could rely on that completely and get rid of our own symbol mappings. You mention some challenge around `std::less`, I wonder if we should try to address that case instead. E.g. suppress symbol mapping lookups for template specializations, if possible.

There's quite a lot of stuff going on here, I wonder if we could take the commits one by one in separate PRs, to make it easier to make progress on the individual directions?

There's a compat release for clang-16 coming up, and I'd like to avoid doing big changes just before that, but let's see if we can start landing improved macOS/libcxx support first thing after release.

Re `<algorithm>` and `std::find`, I wonder if you're hitting the UnresolvedLookupExpr issue: https://github.com/include-what-you-use/include-what-you-use/issues?q=is%3Aissue+is%3Aopen+label%3Aunresolvedlookupexpr.

kilroyd:
> Interesting, I hadn't seen the ClangToolingInclusion library.

I only spotted it digging around trying to find some symbol or another :) Looks related to [clang include fixer](https://clang.llvm.org/extra/clang-include-fixer.html).

> There's quite a lot of stuff going on here, I wonder if we could take the commits one by one in separate PRs, to make it easier to make progress on the individual directions?

Sure. I was thinking of doing that anyway, but didn't want to spend too much time disentangling stuff before you'd at least had a look (and maybe outright rejected one or other approach). I'll leave this PR up for reference, and look at creating ones to "add internal libcxx", and a separate one for "canonical stdlib". I'll leave the libcxx updates for a later PR (or reuse this).

Happy to wait for the compat clang release to be made first - I'd rather not break all the stdlib mappings with a last minute commit :)

kilroyd:
> Re <algorithm> and std::find, I wonder if you're hitting the UnresolvedLookupExpr

This sounds plausible. I recall seeing "UnresolvedLookupExpr" in the verbose output for that header with std::find.

kilroyd:
The majority of this has been merged via other pull requests. Closing.

PR #1222, opened 2023-03-25
Refactor to allow stdlib selection

kilroyd:
Allow independent selection of the internal standard library mappings in IncludePicker construction. For now only the GNU implementations are available, the alternative being no mapping (selected by --no_default_mappings).

In preparation for an alternative C++ library, move the current libstdcpp symbol mappings (which are kludges for how to treat particular symbols) to a generic table that can be shared, and add a separate table specifically for libstdcpp.

### DIFF
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -339,9 +339,9 @@ const IncludeMapEntry libc_symbol_map[] = {
   { "NULL", kPrivate, "<wchar.h>", kPublic },
 };
 
-// Symbol -> include mappings for GNU libstdc++
-const IncludeMapEntry libstdcpp_symbol_map[] = {
-  // Kludge time: almost all STL types take an allocator, but they
+// Common kludges for C++ standard libraries
+const IncludeMapEntry stdlib_cxx_symbol_map[] = {
+  // Almost all STL types take an allocator, but they
   // almost always use the default value.  Usually we detect that
   // and don't try to do IWYU, but sometimes it passes through.
   // For instance, when adding two strings, we end up calling
@@ -373,6 +373,9 @@ const IncludeMapEntry libstdcpp_symbol_map[] = {
   { "std::size_t", kPrivate, "<cuchar>", kPublic },
   { "std::size_t", kPrivate, "<cwchar>", kPublic },
 };
+
+// Symbol -> include mappings for GNU libstdc++
+const IncludeMapEntry libstdcpp_symbol_map[] = {};
 
 const IncludeMapEntry libc_include_map[] = {
   // Private -> public include mappings for GNU libc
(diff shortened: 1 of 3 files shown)


kilroyd:
This is an initial bit of refactoring from pull request #1218 to allow both internal libcxx and canonical stdlib support to be implemented independently, with minimal conflicts

iwyu_include_picker.h
```diff
@@ -67,6 +67,8 @@ struct IncludeMapEntry;
 
 enum class RegexDialect;
 enum IncludeVisibility { kUnusedVisibility, kPublic, kPrivate };
+enum LibCImpl { kNoC, kGlibc };
+enum LibCXXImpl { kNoCXX, kLibstdcpp };
```
kimgr, line 71: Clang calls this `CXXStdlibType`: https://github.com/llvm/llvm-project/blob/main/clang/include/clang/Driver/ToolChain.h#L95.

I think good names would be:
```
CStdlib
CXXStdlib
```
to indicate the relationship with the Clang construct (that said, ours are only used as "mapping keys", and we will probably need many more sets of mappings than Clang needs to care about individual standard libs, so they're not the same. Just closely related.)

I would also prefer to use `enum class` for any new enums, so how about:
```
enum class CStdLib {
  None,
  Glibc
};

enum class CXXStdLib {
  None,
  Libstdcxx
};
```

kimgr (commented):
Some naming suggestions, but otherwise makes sense.

It feels like there's a further generalization waiting to happen where `IncludePicker` could take a sequence of mapping sets in its ctor instead, but I don't have a clear idea for the best way to do it, so probably better not to get too crazy up-front.

iwyu_include_picker.cc
```diff
@@ -1231,28 +1234,43 @@ bool MappedInclude::HasAbsoluteQuotedInclude() const {
...
-  AddSymbolMappings(libc_symbol_map, IWYU_ARRAYSIZE(libc_symbol_map));
-  AddSymbolMappings(libstdcpp_symbol_map, IWYU_ARRAYSIZE(libstdcpp_symbol_map));
+void IncludePicker::AddDefaultMappings(CStdLib cstdlib,
+                                       CXXStdLib cxxstdlib) {
+  if (cstdlib == CStdLib::Glibc) {
+    AddSymbolMappings(libc_symbol_map, IWYU_ARRAYSIZE(libc_symbol_map));
+    AddIncludeMappings(libc_include_map, IWYU_ARRAYSIZE(libc_include_map));
+  }
+
+  if (cxxstdlib != CXXStdLib::None) {
```
kimgr, line 1252: Does the order matter, or could we join this with the `stdlib_cpp_public_headers` block below?

kimgr (approved):
Seems like it's possible to join a couple of `CXXStdLib::None` cases together, but even without that, LGTM.

This patch triggered some ideas:

* We don't need to add C++ mappings if the compilation is for C only
* While it's nice that `IncludePicker` is fully formed after construction, I wonder if it would make sense to do delayed initialization in `IwyuAction::CreateASTConsumer()` where the compiler instance is available (and thereby all compiler flags)
* ... that would make bullet #1 pretty trivial
* ... and also make it possible to derive which mappings to use based on the standard library actually in use by the compiler (`ToolChain::GetCXXStdLibType()`)
* I think the layering should be "IncludePicker knows how to set up default mappings based on `CStdLib` and `CXXStdLib`" and "IWYU frontend knows how to derive `CStdLib` and `CXXStdLib` from command-line switches (whether IWYU or clang)"

But yeah, that's good nutrition for future patches :)

iwyu_include_picker.cc
kilroyd: I wasn't sure about that, so I attempted to maintain the order of `AddIncludeMappings()` calls (and separately the `AddSymbolMappings()` calls). I'll do some further testing and post back.

kimgr: Yeah, good call from a pure refactoring standpoint. But if observable behavior is the same I'd love to get rid of the double take :-)

kilroyd: `run_iwyu_tests.py` doesn't notice a difference - but since I'm not using `libstdcxx`, that's not much. Thinking about it, we map `from` to `to`, and order will only matter for `to` symbols for the same `from`. We're swapping the order of:

* C header -> Associated C++ header mappings and C++ private headers -> C++ public headers (and maybe a few public to public. I don't think the C headers would normally be mentioned in the latter case, so swapping these should be fine.

* Kludges for `std::size_t`, `std:allocator`, `std::char_traits` and C++ symbol -> public header mappings. The stdlib mapping can mention these types, and as long as the first mention in the canonical location (which it must be), then there shouldn't be any difference. The kludge map just means that the standard library doesn't have to worry about handling this. Current `libstdcxx` is empty, and my proposed `libc++` don't mention these types. I would expect Clang's canonical mappings to cover this.

TLDR I think we can merge the blocks.

kimgr: Nice, thanks!

kimgr:
Thank you!

PR #1225, opened 2023-04-02
Libcxx.imp updates for clang 16

kilroyd:
Update the `libcxx` import mapping file with useful directory and symbol mappings. This is a step to improving `run_iwyu_tests.py` output on MacOS.

Sanity tested by running on IWYU source against `libcxx` 16, and MacOs's `libcxx` (13).

Added some test cases - I think it is useful to be able to verify that we aren't breaking standard library mappings, even if we can only do this on the current platform. Eventually, maybe CI can check a few different platforms. I don't expect that we'll have exhaustive tests - but we can at least add tests for issues that are identified.

I had a few further changes that I thought we'd need (spotted by browsing the source) - however I haven't actually seen IWYU complain about these.

I'm aware the IWYU branch for clang 16 is imminent - I've no expectation that these changes make that branch (although it might be nice).

### DIFF
--- a/libcxx.imp
+++ b/libcxx.imp
@@ -1,10 +1,45 @@
 # libc++ headers
 [
-  { include: ["<__functional_base>", private, "<functional>", public ] },
   { include: ["<__mutex_base>", private, "<mutex>", public ] },
-  { symbol: [ "std::declval", private, "<utility>", public ] },
-  { symbol: [ "std::forward", private, "<utility>", public ] },
-  { symbol: [ "std::move", private, "<utility>", public ] },
+
+  # For the following entries:
+  # cd llvm-project/libcxx/include ; find -type d -name "__*" | sort | sed -e "s#./__\(.*\)#  { include: [\"@<__\1/.*>\", private, \"<\1>\", public ] },#"
+  #
+  # tweak tuple_dir entry, and comment out debug_utils, fwd, support
+  { include: ["@<__algorithm/.*>", private, "<algorithm>", public ] },
+  { include: ["@<__atomic/.*>", private, "<atomic>", public] },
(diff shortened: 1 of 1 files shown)


kimgr:
I know I'm a stickler for granularity, but I still think there's a bit too much going on here.

* The libstdcxx `declval` change could easily go in by itself
* I'm not 100% sure about the tests. We don't have any tests for mappings at present (and we try to avoid using standard library headers in new tests) because it's so hard to get them working across all platforms. That said, the changes you're making here are making great strides toward more uniform mapping support across platforms, so maybe it's time?
* On the topic of tests; I think they would give stronger signal if you found a way to make them diagnose, as described in: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#force-diagnostics. I can see that that's harder with mapping tests, though, as they are one symbol per test. So the direct/indirect header machinery gets a little cumbersome. Maybe it's not a good idea after all.
* Another note on tests: IWYU is pretty slow, and running tests that include standard library headers might contribute to test runtime in a way that's detrimental to iteration time. Do you have any sense of how much additional time is spent on these new tests vs. the existing suites?

So I think I'd suggest the following series as separate PRs:

* `std::declval` mapping for GCC (should also fix the #1231 problem)
* Map/set improvements
* libcxx `__xxx` mapping improvements
* Tests (there's quite a bit of discussion and design to do here, I think, so let's save it for last to avoid blocking the other improvements)

kilroyd:
No problem. I'll reduce this PR to just the libcxx.imp change or libcxx 16, and push a PR to fix declval+MSVC ASAP. The others I'll follow up on shortly

PR #1226, opened 2023-04-02
Support Canonical C/C++ mappings

kilroyd:
Clang Tooling has recently gained knowledge about the canonical symbol mappings specified by the standards (sourced from cppreference). Add an option to use these instead of the mappings based on particular implementations.

The benefit of this is that the mapping is independent of any particular implementation, and should be portable.

The main downside is that the mapping is done per symbol, so if the user overloads a symbol that the standard library provides, then IWYU will suggest the standard library header instead of the user header. As an example check tests/cxx/precomputed_tpl_args.cc declaration of std::less

By observation, it looks like IWYU is not mapping operators() correctly (or they may not be getting defined) so implementation detail is leaking through.

### DIFF
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -77,6 +77,18 @@ Exit with error code
 (defaults to 1 if omitted) whether there are \(lqinclude-what-you-use\(rq
 violations or not (for use with \fBmake\fR(1)).
 .TP
+.BI \-\-experimental= flag[,flag...]
+Enable experimental feature. These feature may change or be removed
+without any notice. Current experimental features are:
+.RS
+.TP
+.B clang_mappings
+Use Clang libTooling standard library symbol mappings instead of
+built-in mappings. This potentially replaces the built-in mappings in
+the future, but is currently experimental while the issues are being
+evaluated.
+.RE
+.TP
 .BI \-\-keep= glob
 Always keep the includes matched by
 .IR glob .
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -118,6 +118,9 @@ static void PrintHelp(const char* extra_msg) {
          "   --regex=<dialect>: use specified regex dialect in IWYU:\n"
          "          llvm:       fast and simple (default)\n"
          "          ecmascript: slower, but more feature-complete\n"
+         "   --experimental=flag[,flag...]: enable experimental features\n"
+         "          clang_mappings: use Clang canonical standard library\n"
+         "                          mappings instead of built-in mappings\n"
          "\n"
          "In addition to IWYU-specific options you can specify the following\n"
          "options without -Xiwyu prefix:\n"
@@ -234,6 +237,7 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
     {"error_always", optional_argument, nullptr, 'a'},
     {"debug", required_argument, nullptr, 'd'},
     {"regex", required_argument, nullptr, 'r'},
+    {"experimental", required_argument, nullptr, 'p'},
     {nullptr, 0, nullptr, 0}
   };
   static const char shortopts[] = "v:c:m:d:nr";
@@ -314,6 +318,17 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
           exit(EXIT_FAILURE);
         }
         break;
+      case 'p': {
+        // Split argument on comma and save in global, ignoring empty elements.
+        vector<string> flags = Split(optarg, ",", 0);
+        exp_flags.insert(flags.begin(),
+                         std::remove(flags.begin(), flags.end(), string()));
+        // Print all effective flags for traceability.
+        for (const string& f : exp_flags) {
+          llvm::errs() << "Experimental flag enabled: '" << f << "'\n";
+        }
+        break;
+      }
       case -1:
         return optind;  // means 'no more input'
       default:
@@ -328,6 +343,10 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
 
 bool CommandlineFlags::HasDebugFlag(const char* flag) const {
   return dbg_flags.find(string(flag)) != dbg_flags.end();
+}
+
+bool CommandlineFlags::HasExperimentalFlag(const char* flag) const {
+  return exp_flags.find(string(flag)) != exp_flags.end();
 }
 
 // Though option -v prints version too, it isn't intercepted because it also
@@ -417,12 +436,16 @@ static vector<HeaderSearchPath> ComputeHeaderSearchPaths(
 static CStdLib DeriveCStdLib(clang::CompilerInstance&) {
   if (GlobalFlags().no_default_mappings)
     return CStdLib::None;
+  if (GlobalFlags().HasExperimentalFlag("clang_mappings"))
+    return CStdLib::ClangSymbols;
   return CStdLib::Glibc;
 }
 
 static CXXStdLib DeriveCXXStdLib(clang::CompilerInstance& compiler) {
   if (GlobalFlags().no_default_mappings)
     return CXXStdLib::None;
+  if (GlobalFlags().HasExperimentalFlag("clang_mappings"))
+    return CXXStdLib::ClangSymbols;
   if (cxx_stdlib_type == ToolChain::CXXStdlibType::CST_Libcxx)
     return CXXStdLib::Libcxx;
   return CXXStdLib::Libstdcxx;
@@ -545,6 +568,9 @@ void InitGlobalsAndFlagsForTesting() {
   if (GlobalFlags().no_default_mappings) {
     cstdlib = CStdLib::None;
     cxxstdlib = CXXStdLib::None;
+  } else if (GlobalFlags().HasExperimentalFlag("clang_mappings")) {
+    cstdlib = CStdLib::ClangSymbols;
+    cxxstdlib = CXXStdLib::ClangSymbols;
   }
 
   include_picker =
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -37,6 +37,7 @@
 #include "llvm/Support/SourceMgr.h"
 #include "llvm/Support/YAMLParser.h"
 #include "clang/Basic/FileManager.h"
+#include "clang/Tooling/Inclusions/StandardLibrary.h"
 
 using std::find;
 using std::make_pair;
@@ -1297,9 +1298,23 @@ IncludePicker::IncludePicker(RegexDialect regex_dialect,
 
 void IncludePicker::AddDefaultMappings(CStdLib cstdlib,
                                        CXXStdLib cxxstdlib) {
+  using clang::tooling::stdlib::Header;
+  using clang::tooling::stdlib::Lang;
+  using clang::tooling::stdlib::Symbol;
+
   if (cstdlib == CStdLib::Glibc) {
     AddSymbolMappings(libc_symbol_map, IWYU_ARRAYSIZE(libc_symbol_map));
     AddIncludeMappings(libc_include_map, IWYU_ARRAYSIZE(libc_include_map));
+  } else if (cstdlib == CStdLib::ClangSymbols) {
+    // Get canonical C standard library mappings from clang tooling
+    for (const Symbol& sym : Symbol::all(Lang::C)) {
+      string name = sym.name().str();
+
+      // the canonical header is returned first
+      for (const Header& header : sym.headers()) {
+        AddSymbolMapping(name, MappedInclude(header.name().str()), kPublic);
+      }
+    }
   }
 
   if (cxxstdlib == CXXStdLib::Libstdcxx) {
@@ -1310,6 +1325,16 @@ void IncludePicker::AddDefaultMappings(CStdLib cstdlib,
   } else if (cxxstdlib == CXXStdLib::Libcxx) {
     AddSymbolMappings(libcxx_symbol_map, IWYU_ARRAYSIZE(libcxx_symbol_map));
     AddIncludeMappings(libcxx_include_map, IWYU_ARRAYSIZE(libcxx_include_map));
+  } else if (cxxstdlib == CXXStdLib::ClangSymbols) {
+    // Get canonical C++ standard library mappings from clang tooling
+    for (const Symbol& sym : Symbol::all(Lang::CXX)) {
+      string name = sym.qualifiedName().str();
+
+      // the canonical header is returned first
+      for (const Header& header : sym.headers()) {
+        AddSymbolMapping(name, MappedInclude(header.name().str()), kPublic);
+      }
+    }
   }
 
   if (cxxstdlib != CXXStdLib::None) {
--- a/iwyu_include_picker.h
+++ b/iwyu_include_picker.h
@@ -67,8 +67,8 @@ struct IncludeMapEntry;
 
 enum class RegexDialect;
 enum IncludeVisibility { kUnusedVisibility, kPublic, kPrivate };
-enum class CStdLib { None, Glibc };
-enum class CXXStdLib { None, Libstdcxx, Libcxx };
+enum class CStdLib { None, ClangSymbols, Glibc };
+enum class CXXStdLib { None, ClangSymbols, Libstdcxx, Libcxx };
 
 // When a symbol or file is mapped to an include, that include is represented
 // by this struct.  It always has a quoted_include and may also have a path
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,6 +173,8 @@ else()
     # [1] https://llvm.org/viewvc/llvm-project?view=revision&revision=348907
     # [2] https://llvm.org/viewvc/llvm-project?view=revision&revision=348915
     clangSerialization
+
+    clangToolingInclusionsStdlib
   )
 endif()
 
--- a/iwyu_globals.h
+++ b/iwyu_globals.h
@@ -87,6 +87,7 @@ struct CommandlineFlags {
   CommandlineFlags();                     // sets flags to default values
   int ParseArgv(int argc, char** argv);   // parses flags from argv
   bool HasDebugFlag(const char* flag) const;
+  bool HasExperimentalFlag(const char* flag) const;
 
   set<string> check_also;  // -c: globs to report iwyu violations for
   set<string> keep;        // -k: globs to force-keep includes for
@@ -108,6 +109,7 @@ struct CommandlineFlags {
   int exit_code_error;   // Exit with this code for iwyu violations.
   int exit_code_always;  // Always exit with this exit code.
   set<string> dbg_flags; // Debug flags.
+  set<string> exp_flags;       // Experimental flags.
   RegexDialect regex_dialect;  // Dialect for regular expression processing.
 };
 


bolshakov-a:
Nice!
> The main downside is that the mapping is done per symbol

I think it's a defect of symbol name format in IWYU which doesn't allow to distinguish between main template declaration and template specializations, or between various function overloads.

kilroyd:
I've just pushed an update including a tweak to the man page adding the option, and explaining the status. I think I'd want more feedback on what does and doesn't work before making this considering making this the default behaviour - but I've definitely found it useful

include-what-you-use.1
```diff
@@ -38,6 +38,11 @@ All other options are interpreted as
 .BR clang (1)
 compiler options.
 .TP
+.B \-\-canonical_stdlib
```
kimgr, line 41: I'm a bit skeptical to the whole "canonical" nomenclature. Is it more like `--std_symbols`? Not great either.

iwyu_globals.cc
```diff
@@ -228,6 +231,7 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
     {"error_always", optional_argument, nullptr, 'a'},
     {"debug", required_argument, nullptr, 'd'},
     {"regex", required_argument, nullptr, 'r'},
+    {"canonical_stdlib", no_argument, nullptr, 's'},
```
kimgr, line 234: We should be careful with flags, because they need to stick around for backward compatibility.

I wonder if there should be something like `--experimental=<feature-name>` where we promise zero backward compat. Like `--debug`. Probably not worth the effort here, but interesting to consider.

iwyu_include_picker.cc
```diff
@@ -1244,16 +1245,42 @@ IncludePicker::IncludePicker(RegexDialect regex_dialect,
...
                                        CXXStdLib cxxstdlib) {
+  using clang::tooling::stdlib::Header;
+  using clang::tooling::stdlib::Lang;
+  using clang::tooling::stdlib::Symbol;
+
   if (cstdlib == CStdLib::Glibc) {
     AddSymbolMappings(libc_symbol_map, IWYU_ARRAYSIZE(libc_symbol_map));
     AddIncludeMappings(libc_include_map, IWYU_ARRAYSIZE(libc_include_map));
+  } else if (cstdlib == CStdLib::Canonical) {
+    // Get canonical C standard library mappings from clang tooling
```
kimgr, line 1256: I absolutely love this, compared to all the matchsticks, wire and old band-aids we do for standard library include mappings. Is `precomputed_tpl_args.cc` the only failure if you hack it on to every test invocation?

I wonder if we should look into fixing that first, and then just making this the default and only mapping set for the standard library.

kilroyd: With MacOS (and libcxx), I had the canonical mappings on by default for a run, and the only extra test failure was `precomputed_tpl_args.cc` (there were still 5 other failures that I was already seeing). Not sure if that's the case with libstdc++.

Running on normal code though, you also get a few things leaking from private libcxx headers - `operator<` for some type is one I recall seeing. I expect we need to check what (if anything) is being reported for that via cppreference, and if IWYU wants it in a slightly different style to 'match' properly.

The existing mappings mainly rely on file-mappings - this is nice because the full function prototypes are matched. I wouldn't be surprised if there's a reasonable amount of fallout from switching to mainly symbol-spelling matches.

As I mentioned previously, I'd be inclined to merge more or less as is (happy to put this under an exprimental arg) - so everyone can play with it, and see what problems this has in the wild.

kilroyd: urk, I misremembered. There are a few other failures in the tests when forcing the canonical mappings, at least badinc, check_also, no_char_traits also fail. Some of these also fail on macos anyway, but I think the output may have changed too.

kimgr: Here are the new failures on my Ubuntu system (normally all tests pass):
```
FAIL: test_badinc (__main__.cxx)
  File "/home/kimgr/code/iwyu/include-what-you-use/./run_iwyu_tests.py", line 88, in <lambda>
  File "/home/kimgr/code/iwyu/include-what-you-use/iwyu_test_util.py", line 478, in TestIwyuOnRelativeFile
AssertionError: 
tests/cxx/badinc.cc:1069: Unexpected diagnostic:
va_list is defined in <cstdarg>, which isn't directly #included.

tests/cxx/badinc.cc:1069: Unmatched regex:
va_list is...*<stdio.h>

+++ 
@@ -1,6 +1,7 @@
 tests/cxx/badinc.cc should add these lines:
 #include <ctype.h>
 #include <stddef.h>
+#include <cstdarg>
 #include <list>
 #include "tests/cxx/badinc-i1.h"
 class D2_Class;
@@ -31,7 +32,8 @@
 #include <setjmp.h>
 #include <stddef.h>  // for offsetof
 #include <algorithm>  // for find
-#include <fstream>  // for fstream
+#include <cstdarg>  // for va_list
+#include <fstream>  // for basic_fstream, fstream
 #include <list>  // for list
 #include <string>  // for basic_string, operator+, string
 #include <typeinfo>  // for type_info
---


FAIL: test_check_also (__main__.cxx)
  File "/home/kimgr/code/iwyu/include-what-you-use/./run_iwyu_tests.py", line 88, in <lambda>
  File "/home/kimgr/code/iwyu/include-what-you-use/iwyu_test_util.py", line 478, in TestIwyuOnRelativeFile
AssertionError: 
tests/cxx/check_also-d1.h:18: Unexpected diagnostic:
NULL is defined in <cstddef>, which isn't directly #included.

tests/cxx/check_also-d1.h:18: Unmatched regex:
NULL is...*<stddef.h>

Unexpected summary diffs for tests/cxx/check_also-d1.h:
+++ 
@@ -1,8 +1,8 @@
 tests/cxx/check_also-d1.h should add these lines:
-#include <stddef.h>
+#include <cstddef>
 
 tests/cxx/check_also-d1.h should remove these lines:
 - #include "check_also-i1.h"  // lines XX-XX
 
 The full include-list for tests/cxx/check_also-d1.h:
-#include <stddef.h>  // for NULL
+#include <cstddef>  // for NULL
---


FAIL: test_no_char_traits (__main__.cxx)
  File "/home/kimgr/code/iwyu/include-what-you-use/./run_iwyu_tests.py", line 88, in <lambda>
  File "/home/kimgr/code/iwyu/include-what-you-use/iwyu_test_util.py", line 478, in TestIwyuOnRelativeFile
AssertionError: 
tests/cxx/no_char_traits.cc:29: Unexpected diagnostic:
std::operator<< is defined in <ostream>, which isn't directly #included.

tests/cxx/no_char_traits.cc:31: Unexpected diagnostic:
std::char_traits is defined in <ostream>, which isn't directly #included.

tests/cxx/no_char_traits.cc:31: Unexpected diagnostic:
std::operator<< is defined in <ostream>, which isn't directly #included.

Unexpected summary diffs for tests/cxx/no_char_traits.cc:
+++ 
@@ -1 +1,8 @@
-(tests/cxx/no_char_traits.cc has correct #includes/fwd-decls)
+tests/cxx/no_char_traits.cc should add these lines:
+#include <ostream>
+
+tests/cxx/no_char_traits.cc should remove these lines:
+
+The full include-list for tests/cxx/no_char_traits.cc:
+#include <iostream>  // for basic_ostream, cout
+#include <ostream>  // for char_traits, operator<<
---


FAIL: test_precomputed_tpl_args (__main__.cxx)
  File "/home/kimgr/code/iwyu/include-what-you-use/./run_iwyu_tests.py", line 88, in <lambda>
  File "/home/kimgr/code/iwyu/include-what-you-use/iwyu_test_util.py", line 478, in TestIwyuOnRelativeFile
AssertionError: 
tests/cxx/precomputed_tpl_args.cc:43: Unexpected diagnostic:
std::less is defined in <functional>, which isn't directly #included.

tests/cxx/precomputed_tpl_args.cc:43: Unmatched regex:
std::less is...*precomputed_tpl_args-i1.h

tests/cxx/precomputed_tpl_args.cc:47: Unexpected diagnostic:
std::less is defined in <functional>, which isn't directly #included.

Unexpected summary diffs for tests/cxx/precomputed_tpl_args.cc:
+++ 
@@ -1,12 +1,14 @@
 tests/cxx/precomputed_tpl_args.cc should add these lines:
+#include <functional>
 #include "tests/cxx/precomputed_tpl_args-i1.h"
 
 tests/cxx/precomputed_tpl_args.cc should remove these lines:
 
 The full include-list for tests/cxx/precomputed_tpl_args.cc:
 #include <bitset>  // for bitset
+#include <functional>  // for less
 #include <map>  // for map
 #include <set>  // for set
 #include <vector>  // for vector
-#include "tests/cxx/precomputed_tpl_args-d1.h"  // for D1SpecializationClass, less
-#include "tests/cxx/precomputed_tpl_args-i1.h"  // for IndirectClass, SpecializationClass, less
+#include "tests/cxx/precomputed_tpl_args-d1.h"  // for D1SpecializationClass
+#include "tests/cxx/precomputed_tpl_args-i1.h"  // for IndirectClass, SpecializationClass
---


Ran 178 tests in 11.693s

FAILED (failures=4)
```

kimgr: I also see a regression in test runtime, from ~10.5 to ~11.5 seconds, so around 10%. Nothing here stood out as particularly slow, so I wonder where that time goes...

kilroyd: Those errors look similar, but not exaclty the same, as those that I'm seeing on MacOS.

> I also see a regression in test runtime

There are about 5200 symbols being mapped (934 C, 4221 C++). There might be time setting those up compared to the 100's we usually create? I hacked out the setup loop - for me that dropped runtime from 2.39s to 2.32s (for just processing badinc), so only accounting for about half of what you're seeing.

kimgr: Hmm. Maybe time to look into some mapping optimizations after this has landed. I don't think the regression is bad enough to do anything up-front.

iwyu_globals.cc
```diff
@@ -308,6 +313,13 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
           exit(EXIT_FAILURE);
         }
         break;
+      case 'p':
+        if (strcmp(optarg, "clang_mappings") == 0) {
```
kimgr, line 317: Could you model this after `--debug` instead, and store a set of flag strings? Seems nice to use the same design for both.

iwyu_include_picker.h
```diff
@@ -67,8 +67,8 @@ struct IncludeMapEntry;
 
 enum class RegexDialect;
 enum IncludeVisibility { kUnusedVisibility, kPublic, kPrivate };
-enum class CStdLib { None, Glibc };
-enum class CXXStdLib { None, Libstdcxx };
+enum class CStdLib { None, Canonical, Glibc };
```
kimgr, line 70: This enumerator might be better called `ClangSymbols`?

iwyu_include_picker.cc
```diff
@@ -1247,16 +1248,42 @@ IncludePicker::IncludePicker(RegexDialect regex_dialect,
...
   }
 
   if (cxxstdlib == CXXStdLib::Libstdcxx) {
     AddSymbolMappings(libstdcpp_symbol_map,
                       IWYU_ARRAYSIZE(libstdcpp_symbol_map));
     AddIncludeMappings(libstdcpp_include_map,
                        IWYU_ARRAYSIZE(libstdcpp_include_map));
+  } else if (cxxstdlib == CXXStdLib::Canonical) {
+    // Get canonical C++ standard library mappings from clang tooling
+    for (auto& sym : Symbol::all(Lang::CXX)) {
```
kimgr, line 1278: For all three `auto`-s here -- we only use `auto` when the type is clearly visible on the same line, so let's spell out the type.

iwyu_include_picker.cc
```diff
@@ -1247,16 +1248,42 @@ IncludePicker::IncludePicker(RegexDialect regex_dialect,
...
+      string name = sym.qualifiedName().str();
+      auto headers = sym.headers();
```
kimgr, line 1280: I'd use `sym.headers()` directly in the for loop below

iwyu_include_picker.cc
```diff
@@ -1247,16 +1248,42 @@ IncludePicker::IncludePicker(RegexDialect regex_dialect,
...
+  using clang::tooling::stdlib::Symbol;
+
   if (cstdlib == CStdLib::Glibc) {
     AddSymbolMappings(libc_symbol_map, IWYU_ARRAYSIZE(libc_symbol_map));
     AddIncludeMappings(libc_include_map, IWYU_ARRAYSIZE(libc_include_map));
+  } else if (cstdlib == CStdLib::Canonical) {
+    // Get canonical C standard library mappings from clang tooling
+    for (auto& sym : Symbol::all(Lang::C)) {
+      string name = sym.name().str();
+      auto headers = sym.headers();
```
kimgr, line 1262: See earlier comments about `auto` and `sym.headers()`.

iwyu_include_picker.cc
kimgr: There's actually a fun mix of problems there:

* `badinc.cc` has `va_list is defined in <cstdarg>, which isn't directly #included.` -- `va_list` is only mapped to `<cstdarg>` in the clang mappings, but we accept it from a few other headers, `<stdio.h>` most notably.
* `check_also-d1.h` wants `NULL` from `<stddef.h>`, but now `<cstddef>` is suggested (for some reason, not sure why; can't find any Clang mappings for `NULL`...?)
* `no_char_traits.cc` wants `<ostream>` for `operator<<` and `std::char_traits`. I think both of these were masked by public-to-public mappings from `<ostream>` to `<iostream>` on master
* `precomputed_tpl_args.cc:43` complains about `std::less` now. I believe this used to work because `std::less` was found in some private `<bits/...map...>` header, which was eventually mapped back to `<map>`. But not sure.

I don't see any obvious systematic fixes there... Some could probably be accepted as the "right" behavior, and fixed up in the tests, but many are simply a matter of practicality.

kilroyd: FWIW, macos+libcxx behaves similarly for the first 3 tests. So I suspect a few things are in the GCC mappings mappings that we don't have on libcxx / clang. I wanted to double check what each test was doing before proposing anything.

iwyu_globals.cc
```diff
@@ -114,6 +114,9 @@ static void PrintHelp(const char* extra_msg) {
          "   --regex=<dialect>: use specified regex dialect in IWYU:\n"
          "          llvm:       fast and simple (default)\n"
          "          ecmascript: slower, but more feature-complete\n"
+         "   --experimental=flag[,flag...]: enable experimental features\n"
+         "          clang_mappings: use Clang canonical standard library\n"
+         "                          mappings instead of in-built mappings\n"
```
kimgr, line 119: `in-built` -> `built-in`

iwyu_include_picker.cc
kimgr: Thanks! I think the patch looks great in shape and form. I wonder how we should think about moving forward with it. Do you think it's reasonable to augment Clang's symbol mappings with manual overlays to preserve IWYU behavior? Or should we just be less pragmatic in favor of simplicity? All of the examples brought to light by the test suite are pretty nice with the mainline built-in mappings, and pretty annoying with the Clang mappings.

kilroyd: In general I expect we should follow what the mainline behavior is doing, except where we can identify the new behavior as preferable. I'm hoping we can do that with mappings. Possibly, if we inspect the symbol namings that clang is outputting, there might be a few things we can massage to work better for IWYU (or change upstream)

kilroyd:
@kimgr - just double checking what you're expecting to do with this PR.

kimgr:
@kilroyd Thanks; I just lost steam and lost track of it. I think this feels like a 100% perfect direction, but I'm concerned about the regressions in our test suite (and subsequently for users).

Ah, I re-discovered now that it's all hidden behind a flag. So maybe that's a good way to approach it...

I think I still would like to sketch out a solution proposal for the issues surfaced in:
https://github.com/include-what-you-use/include-what-you-use/pull/1226#discussion_r1170519050

You said:
> In general I expect we should follow what the mainline behavior is doing, except where we can identify the new behavior as preferable. I'm hoping we can do that with mappings. 

Do you think it's it possible to do that with additional _header_ mappings on top the Clang symbol mappings?

> Possibly, if we inspect the symbol namings that clang is outputting, there might be a few things we can massage to work better for IWYU (or change upstream)

Are you saying there are quirks with symbol naming where Clang and IWYU disagree? If so I definitely think we should change IWYU to follow Clang.

kilroyd:
> Do you think it's it possible to do that with additional header mappings on top the Clang symbol mappings?

I'm hoping that as private => public header mappings would work, but I've not tried.

> Are you saying there are quirks with symbol naming where Clang and IWYU disagree? If so I definitely think we should change IWYU to follow Clang.

Possibly, yes. The check_also test definition for NULL is one case of this. IWYU currently says the canonical location is stddef.h (regardless of compiler language mode). I expect clang in C++ mode to say cstddef, and stddef.h only in C mode. This would be a small fix in IWYU just for some of these symbols (move the C++ headers from libc_symbol_map into a separate array, and load that array before libc_symbol_map when in C++ mode).

> I think I still would like to sketch out a solution proposal for the issues surfaced in ...

For check_also.cc see previous paragraph. I'd change the test to expect cstddef.

For precomputed_tpl_args.cc I don't think there's a fix. The issue stems from using _just_ symbol names to suggest a header. Maybe special case logic could be added to IWYU so it is aware that users may override/specialize std::less, but that doesn't feel like it scales well.

For no_char_traits.cc... I think this works in libstdcpp because of the private=>public header mappings of ostream=>iostream at the end of libstdcpp_include_map. So moving these to a common array probably fixes this.

For badinc.cc va_list, I think this is similar to check_also.cc - IWYU names stdarg.h as the canonical location (with stdio.h as an alternate), but in C++ mode clang is suggesting stdarg. I'd go with clang on this one. For the fstream header, I'm not sure why IWYU is listing an additional symbol...

Changing things in the symbol mappings feels fraught with potential problems, and I'm sure there will be different opinions too. I'd want to add testing as the changes are made, and to have that testing cover libstdcpp, libc++ and the canonical clang mappings, so we don't pick a solution specific to one of the implementations.

kimgr:
Hmm, yeah, the `<cname>` vs `name.h` headers is likely to be contentious.

My personal preference is to use the plain old `<name.h>` headers in C++ code too, since so many of the C headers expose macros which can't be wrapped in `namespace std`, and the additional header incurs a conceptual/visual (`std::` qualification) and  build-time (additional stat for the pointless wrapper) cost.

Some people are very diligent with only using `<cname>` headers to play well with standard recommendations. That's also fine.

Others have very large code bases with an unholy mix of `<cname>` and `<name.h>`, and would like for IWYU to make as few changes as possible.

It would be nice if IWYU could take a principled stance here, but I'm honestly not sure what it should be :)

kimgr:
> For no_char_traits.cc... I think this works in libstdcpp because of the private=>public header mappings of ostream=>iostream at the end of libstdcpp_include_map. So moving these to a common array probably fixes this.

By "common array", do you mean one canonical for all standard library implementations?

kilroyd:
> Hmm, yeah, the <cname> vs name.h headers is likely to be contentious.

I'm hoping it shouldn't be too contentious - I expect the existing behaviour accepting either <cname> or <name.h> should continue to be the case. We would just change the location IWYU suggests when neither header (or any of the alternates) is present. Yes, one group of people may be slightly inconvenienced (but that's already true). 

> By "common array", do you mean one canonical for all standard library implementations?

Yes. I hadn't previously noticed the private=> public header mappings that libstdcpp has. I think those can be used for all c++ libraries.

> I wonder if we could generate a test suite similar to how Clang generates symbol mappings?

It's a possibility. It gets harder if you actually want function calls with correct arguments, but maybe we can assign function pointers, and find other tricks to 'use' symbols easily.

kimgr:
@kilroyd Thanks! If there's some commitment from you to keep working on this so we can eventually remove the flag, I feel good about moving forward with the patch in this PR.

> > By "common array", do you mean one canonical for all standard library implementations?
>
> Yes. I hadn't previously noticed the private=> public header mappings that libstdcpp has. I think those can be used for all c++ libraries.

Cool, yeah, let's follow up with a split of those mappings in a later PR (into libstdc++-specific vs. shared convenience mappings).

> > I wonder if we could generate a test suite similar to how Clang generates symbol mappings?
>
> It's a possibility. It gets harder if you actually want function calls with correct arguments, but maybe we can assign function pointers, and find other tricks to 'use' symbols easily.

Ah, yeah. Maybe just `using X = std::foo;` for proper symbols and `#define X some_macro` for macros. Not sure how far that would get us, but it's an interesting rabbit hole to explore.

kilroyd:
> @kilroyd Thanks! If there's some commitment from you to keep working on this so we can eventually remove the flag, I feel good about moving forward with the patch in this PR.

Thanks! I definitely expect to work on making libc++ and libstdc++ output agree better. I'll see if I can make a start...

kimgr:
Thank you for doing this, and sorry about the extended wait.

PR #1231, opened 2023-04-08
Fix MSVC build

bolshakov-a:
MSVC doesn't accept arrays of zero size, so remove it.


kilroyd:
Apologies. That's rather restrictive of MSVC :(

My pull request #1225 adds an entry in this array - any chance of that being merged?

bolshakov-a:
Ok. Let's wait better for merging of your PR.

Issue #1065, opened 2022-06-18
typedef before struct


alejandro-colomar:
I got this bug from iwyu(1) while running it on some NGINX code.  Okay, the code is not ideal, but it's correct.  I simplified it to a minimial reproducer:
```c
$ cat typedef.c
typedef struct foo_s foo_t;

struct foo_s {
    int x;
};
```
```
$ iwyu -Xiwyu --no_fwd_decls typedef.c

typedef.c should add these lines:
struct foo_s;

typedef.c should remove these lines:

The full include-list for typedef.c:
struct foo_s;
---
```

I kind of get the reason iwyu(1) suggests the fwd decl _even_ if I said I don't want them: if it suggested me to include the file that provides the declaration, it would be asking me to do a self-include, and of course it doesn't. 

In the real code, I found that there's a cyclic dep between typedefs, and that's why the typedef was before the struct definition.  A struct declaration before the typedef seems to me useless, so iwyu(1) should avoid warning on this if possible.

kimgr:
Duplicates #793, but the fixing commit mentions this issue, so I'll close the other one.

kimgr:
Bugfix merged with #1077

PR #1232, opened 2023-04-08
Handle elaborated keywords as fwd-decls

bolshakov-a:
C++ doesn't require any forward-declaration after the declaration name occured in an elaborated type with corresponding tag (`class`/`struct`/`union`).

Initial support. Only those elaborated types are treated as fwd-declarations which occur before any other declaration of the underlying type ("owning" elaborated types).

"TypeLoc" is intercepted and handled instead of just the type because only explicitly written elaborated types matter. Moreover, location information could be used in the future for full support of treating elaborated types as fwd-declarations, even in the "non-owning" case.

Besides, it is the more correct fix of #1065.

Provision of underlying elaborated type in typedef or type alias declaration has been fixed. Otherwise, there was a bug when IWYU required a full type neither on alias declaration side nor on using side.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -42,6 +42,7 @@ using clang::CXXMethodDecl;
 using clang::CXXRecordDecl;
 using clang::Decl;
 using clang::DeclContext;
+using clang::ElaboratedTypeLoc;
 using clang::EnumDecl;
 using clang::FileEntry;
 using clang::FunctionDecl;
@@ -502,6 +503,17 @@ OneIncludeOrForwardDeclareLine::OneIncludeOrForwardDeclareLine(
 }
 
 OneIncludeOrForwardDeclareLine::OneIncludeOrForwardDeclareLine(
+    ElaboratedTypeLoc type_loc)
+    : is_desired_(true),
+      is_present_(true),
+      is_elaborated_type_(true),
+      fwd_decl_(type_loc.getTypePtr()->getOwnedTagDecl()) {
+  const SourceRange decl_lines = type_loc.getLocalSourceRange();
+  start_linenum_ = GetLineNumber(GetInstantiationLoc(decl_lines.getBegin()));
+  end_linenum_ = GetLineNumber(GetInstantiationLoc(decl_lines.getEnd()));
+}
+
+OneIncludeOrForwardDeclareLine::OneIncludeOrForwardDeclareLine(
     const FileEntry* included_file, const string& quoted_include, int linenum)
     : line_("#include " + quoted_include),
       start_linenum_(linenum),
@@ -590,6 +602,15 @@ void IwyuFileInfo::AddForwardDeclare(const clang::NamedDecl* fwd_decl,
            << GetFilePath(file_) << ":" << lines_.back().LineNumberString()
            << ": " << internal::PrintablePtr(fwd_decl)
            << internal::GetQualifiedNameAsString(fwd_decl) << "\n";
+}
+
+void IwyuFileInfo::AddElaboratedType(ElaboratedTypeLoc type_loc) {
+  if (!type_loc.getTypePtr()->getOwnedTagDecl())
+    return;
+  lines_.push_back(OneIncludeOrForwardDeclareLine(type_loc));
+  VERRS(6) << "Found owning elaborated type: " << GetFilePath(file_) << ":"
+           << lines_.back().LineNumberString() << ": "
+           << PrintableTypeLoc(type_loc) << "\n";
 }
 
 void IwyuFileInfo::AddUsingDecl(const UsingDecl* using_decl) {
@@ -2084,7 +2105,7 @@ size_t PrintableDiffs(const string& filename,
       OutputLine("\nThe full include-list for " + filename + ":"));
     for (const auto& key_line : sorted_lines) {
       const OneIncludeOrForwardDeclareLine* line = key_line.second;
-      if (line->is_desired()) {
+      if (line->is_desired() && !line->is_elaborated_type()) {
         output_lines.push_back(
           PrintableIncludeOrForwardDeclareLine(*line, aqi));
       }
--- a/iwyu_output.h
+++ b/iwyu_output.h
@@ -30,6 +30,7 @@
 namespace clang {
 class FileEntry;
 class UsingDecl;
+class ElaboratedTypeLoc;
 }  // namespace clang
 
 namespace include_what_you_use {
@@ -144,6 +145,7 @@ class OneUse {
 class OneIncludeOrForwardDeclareLine {
  public:
   explicit OneIncludeOrForwardDeclareLine(const clang::NamedDecl* fwd_decl);
+  explicit OneIncludeOrForwardDeclareLine(clang::ElaboratedTypeLoc);
   OneIncludeOrForwardDeclareLine(const clang::FileEntry* included_file,
                                  const string& quoted_include, int linenum);
 
@@ -157,6 +159,9 @@ class OneIncludeOrForwardDeclareLine {
   }
   bool is_present() const {
     return is_present_;
+  }
+  bool is_elaborated_type() const {
+    return is_elaborated_type_;
   }
   const map<string, int>& symbol_counts() const {
     return symbol_counts_;
@@ -206,9 +211,10 @@ class OneIncludeOrForwardDeclareLine {
   string line_;  // '#include XXX' or 'class YYY;'
   int start_linenum_ = -1;
   int end_linenum_ = -1;
-  bool is_desired_ = false;         // IWYU will recommend this line
-  bool is_present_ = false;         // line was present before the IWYU run
-  map<string, int> symbol_counts_;  // how many times we referenced each symbol
+  bool is_desired_ = false;          // IWYU will recommend this line
+  bool is_present_ = false;          // line was present before the IWYU run
+  bool is_elaborated_type_ = false;  // Fwd-decl introduced by elaborated type
+  map<string, int> symbol_counts_;   // how many times we referenced each symbol
   // Only either two following members are set for includes
   string quoted_include_;  // quoted file name we're including
   const clang::FileEntry* included_file_ = nullptr;  // the file we're including
@@ -258,6 +264,7 @@ class IwyuFileInfo {
   // the fwd-decl be removed, even if we don't see any uses of it.
   void AddForwardDeclare(const clang::NamedDecl* fwd_decl,
                          bool definitely_keep_fwd_decl);
+  void AddElaboratedType(clang::ElaboratedTypeLoc);
 
   void AddUsingDecl(const clang::UsingDecl* using_decl);
 
--- a/tests/cxx/elaborated_type.cc
+++ b/tests/cxx/elaborated_type.cc
@@ -60,6 +60,21 @@ void global_qualified_elab(class ::GlobalClass* c);
 // they must also be forward-declared.
 Elaboration::Template<int, float>* namespace_qualified_template;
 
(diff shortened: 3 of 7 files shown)


iwyu_output.cc
```diff
@@ -507,6 +508,21 @@ OneIncludeOrForwardDeclareLine::OneIncludeOrForwardDeclareLine(
   end_linenum_ = GetLineNumber(GetInstantiationLoc(decl_lines.getEnd()));
 }
 
+OneIncludeOrForwardDeclareLine::OneIncludeOrForwardDeclareLine(
```
kilroyd, line 511: Just checking - so this elaborated type line is treated the same as any other forward declaration for the type (it `matches()` the same), but if this one `is_desired()` we don't output anything, as it's not an explicit forward declaration.

From the perpective of what's needed in the code, this seems ok. From the perspective of the developer running IWYU, I feel like I'd expect to see some output for the type in question. i.e. 'ClassXX used - no forward declaration needed due to elaborated type on line XX-YY'

tests/cxx/elaborated_type.cc
```diff
@@ -60,6 +60,21 @@ void global_qualified_elab(class ::GlobalClass* c);
 // they must also be forward-declared.
 Elaboration::Template<int, float>* namespace_qualified_template;
 
+// Elaborated types can be treated as forward-declarations, hence no fwd-decl is
+// needed if a declaration is introduced by elaborated type.
+// A declaration "owned" by elaborated type is the one which doesn't appear
```
kilroyd, line 65: Any chance of expanding this explanation of 'owned' - is the thing being owned the declaration on line 69, or is it just the tag `OwningElaborated`? This explanation sounds to me like the former, but that would mean removing lines 69 should result in `getOwnedTagDecl()` returning null which doesn't feel right. From the code I feel like the first elaborated type declaration owns the tag if it occurs before a forward declaration for the tag.

kilroyd (commented):
I'm not particularly familiar with relying on elaborated types for a forward declaration, but I had time to have a look. The commit seems ok to me. Hopefully my comments make some sense...

iwyu_output.h
```diff
@@ -144,6 +145,8 @@ class OneUse {
 class OneIncludeOrForwardDeclareLine {
  public:
   explicit OneIncludeOrForwardDeclareLine(const clang::NamedDecl* fwd_decl);
+  explicit OneIncludeOrForwardDeclareLine(clang::ElaboratedTypeLoc,
```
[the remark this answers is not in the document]
bolshakov-a: Actually, it also changes behavior of some cases when initializer list is used, but yeah, it doesn't make much sense here. But there is another problem: the parameters aren't orthogonal. Just `ElaboratedTypeLoc` should be sufficient.

iwyu_output.cc
bolshakov-a: I think it would not fit very well into "should add"/"should remove"/"full list" sections. As for warnings, IWYU doesn't emit any warning for the true fwd-declaration case too, if the fwd-decl appears before any fwd-decl-use in the same file. Elaborated type occurence is itself reported under verbosity level 6.

tests/cxx/elaborated_type.cc
bolshakov-a: Yes, a declaration "owned" by elaborated type is actually an implicit declaration introduced by elaborated type (in line 67 in this case). It is different from any other forward declaration. (It is AFAIU; I haven't dug into clang code deeply.) Fwd-declaration in line 69 isn't very important for this test case and could be removed. The point here is that IWYU shouldn't suggest to place a new fwd-declaration if elaborated type presence is sufficient.

I understand that "owned" declaration concept is confusing, and should probably be considered as a temporary solution (if acceptable at all).

bolshakov-a:
@kilroyd, thanks for the review!

iwyu_output.cc
kimgr: There is some precedent for that with why-comments, e.g. `// for Type` which sometimes have a bit more context as to why a symbol is needed, e.g. `// for Type (for autocast)`. But these are set up at the point of reporting in iwyu.cc.

And I've actually considered removing them because they are a bit of hassle to maintain in traversal. Or potentially extending them to some kind of provenance (e.g. `std::vector<string> reasons` on `ASTNode`, so that traversal can push notes along the way and leave a breadcrumb trail up to where a use decision was made). But I'd really like not to add more noise in their current form.

bolshakov-a: > And I've actually considered removing them

I can say that they are useful for me. Because I usually replace headers suggested for autocast or fn return to fwd-declarations. But I actually want to add command line flags to supersede default IWYU policies.

kilroyd: > I think it would not fit very well into "should add"/"should remove"/"full list" sections.

Agreed. I was just thinking out loud what I'd hope to see.

> Or potentially extending them to some kind of provenance

Infrastructure to do this sounds interesting. Back to this commit, I've no issue with what you're currently doing here :)

tests/cxx/elaborated_type.cc
```diff
@@ -60,6 +60,21 @@ void global_qualified_elab(class ::GlobalClass* c);
 // they must also be forward-declared.
 Elaboration::Template<int, float>* namespace_qualified_template;
 
+// Elaborated types can be treated as forward-declarations, hence no fwd-decl is
+// needed if a declaration is introduced by elaborated type.
+// A declaration "owned" by elaborated type is an implicit declaration
+// introduced by elaborated type if no one was already present.
+void owning_elaborated(class OwningElaborated*);
```
kimgr, line 67: It's curious that this declaration "falls out" into the containing scope/decl context. Are the rules different between C and C++ here? https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L4171.

tests/cxx/elaborated_type.cc
```diff
@@ -60,6 +60,21 @@ void global_qualified_elab(class ::GlobalClass* c);
...
+
+class OwningElaborated;
+void previously_introduced(OwningElaborated*);
+
+// But if an elaborated type appears after the first fwd-decl use of the type, a
+// true forward declaration is still needed.
+class ForNonOwningElaborated;
```
kimgr, line 74: I wonder if we should skip the whole owning/non-owning terminology... Isn't a regular forward-declaration, such as `class ForNonOwningElaborated;` really also an elaborated type? cppreference explains it that way: https://en.cppreference.com/w/cpp/language/elaborated_type_specifier. But I suppose Clang does not model the elaborated type specifier here, because the declaration introduces the type anyway.

From a language user's point of view, the rule seems to be more like:

> If an elaborated type specifier does not have a previous declaration of the same kind + name, it _is_ a declaration (in global scope)

Maybe something like `FirstElaborated` and `FirstForwardDeclared` would get the point across with less compiler internals?

tests/cxx/elaborated_type.cc
```diff
@@ -60,6 +60,21 @@ void global_qualified_elab(class ::GlobalClass* c);
 // they must also be forward-declared.
 Elaboration::Template<int, float>* namespace_qualified_template;
 
+// Elaborated types can be treated as forward-declarations, hence no fwd-decl is
+// needed if a declaration is introduced by elaborated type.
+// A declaration "owned" by elaborated type is an implicit declaration
+// introduced by elaborated type if no one was already present.
+void owning_elaborated(class OwningElaborated*);
+
+class OwningElaborated;
```
kimgr, line 69: I'd also be interested in seeing a use of `OwningElaborated` without elaboration and before the first real forward-declare, e.g.
```
void f(class OwningElaborated*);
void g(OwningElaborated*);
class OwningElaborated;
void h(OwningElaborated*);
```

That would make it clearer that the first elaborated type specifier really pushes a decl out into the containing scope.

iwyu_output.h
```diff
@@ -208,6 +213,8 @@ class OneIncludeOrForwardDeclareLine {
   int end_linenum_;
   bool is_desired_;                 // IWYU will recommend this line
   bool is_present_;                 // line was present before the IWYU run
+  bool is_elaborated_type_ =  // Fwd-decl is introduced by elaborated type.
```
kimgr, line 216: I wonder if it would play out nicer if you did a commit first with the member initializers, and removed all the long initializer lists from the `OneIncludeOrForwardDeclareLine` constructors? That way it's fully consistent every step of the way.

kimgr, line 216: I think you could fit this one one line with a small rephrasing:
```
bool is_elaborated_type_ = false; // Fwd-decl introduced by elaborated type
```

iwyu_output.cc
kimgr: I'm actually a little surprised the whole `matches` machinery plays out correctly with pointer identity for `NamedDecl`'s. Do they all become the same decl in a scenario like this?
```
void f(class X*);  // decl=0x12345678?
void g(X*);        // decl=0x12345678?
class X;           // decl=0x12345678?
class X;           // decl=0x12345678?
```

Or maybe there's some canonicalization happening somewhere else.

kimgr (commented):
Very cool, thank you! Bunch of nits inline, but overall looks good.

iwyu_output.h
kimgr: Actually, let's do initializers and format in the same commit, and in a separate PR. It's a nice step improvement.

tests/cxx/elaborated_type.cc
bolshakov-a: That C language rule seems strange, as for me... But with the current implementation, specific visibility rules don't matter, I think. An elaborated type does or does not introduce a new "owned" declaration. Later usages refer to some declaration. IWYU matches usages to declarations, when all the lookup work should already be done by clang.

bolshakov-a: Btw, I thought whether I might break that case with C function, but it luckily seems that I haven't. The behavior seems to be the same as on the master branch.

tests/cxx/elaborated_type.cc
bolshakov-a: IWYU behavior probably should not depend on whether an elaborated type introduces an implicit declaration or not (e.g. because a declaration is already visible in some indirect header), despite it is not still the case. So I'm biased to remove all these subtle details from the test.

iwyu_output.cc
bolshakov-a: > Do they all become the same decl in a scenario like this?

The last two are actually redeclarations of the "owned" declaration from the first line, and they have their own addresses. Clang `TagType::getDecl()` function returns first declaration occured in a translation unit. `CalculateIwyuForForwardDeclareUse` can then remap a use to a forward declaration from the same file ([step D1](https://github.com/include-what-you-use/include-what-you-use/blob/0.20/iwyu_output.cc#L1034)).

kimgr: Ah, it was the D1 canonicalization I was missing, thanks!

tests/cxx/elaborated_type.cc
kimgr: Yeah, there are tests for that under `tests/c`, so we should be good :)

tests/cxx/elaborated_type.cc
kimgr: I think the "owning" terminology is internal to Clang, so I'd at least tone that down.

But it seems useful with tests to help answer the question;
```
Why does IWYU remove my forward declaration of Bar in this scenario???

// fwd-decls
void foo(class Bar*);
class Bar;

void baz() {
    return foo(nullptr);
}
```
The "elaborated-type-introduces-a-decl" rule is subtle enough that someone will run into it and be surprised.

kimgr:
Very nice, thanks!

PR #1236, opened 2023-04-13
libcxx: map <__tree> to <map> and <set>

kilroyd:
The <__tree> header defines red-black tree class used by both map and set implementations.

When using a const set iterator, IWYU sees this being defined in <__tree>. This doesn't happen for the map iterator (or other map functionality that I could spot).

For consistency, map the header to both <map> and <set> so including either header is accepted by IWYU.


kimgr:
Nice, thank you!

Issue #306, opened 2016-05-19
Deduce STL mappings from -stdlib


EugeneZelenko:
I may be wrong, but it looks like mapping could be specified only manually with --mapping_file.

However STL implementation and its mappings could be deduced automatically from -stdlib compile command line parameters: libc++ for LLVM and libstdc++ for GCC implementation.

vsapsai:
This will be useful when we have incompatibilities between libc++ and libstdc++ mappings. So far libstdc++-specific mappings like `{ include: ["<bits/stl_tree.h>", private, "<map>", public ] }` are safely ignored when they are not applicable.

How does Microsoft STL implementation fit in this scenario?

kimgr:
I think --driver-mode=cl implies the MS STL, so that might be a hook to detect it if none of the others are specified explicitly.

kimgr:
See also #377.

PR #1238, opened 2023-04-15
Add builtin libcxx mappings

kilroyd:
Automatically select the standard library that the clang toolchain identifies based on platform and command line arguments.

Fixes: #306

### DIFF
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -416,9 +416,11 @@ static CStdLib DeriveCStdLib(clang::CompilerInstance&) {
   return CStdLib::Glibc;
 }
 
-static CXXStdLib DeriveCXXStdLib(clang::CompilerInstance&) {
+static CXXStdLib DeriveCXXStdLib(clang::CompilerInstance& compiler) {
   if (GlobalFlags().no_default_mappings)
     return CXXStdLib::None;
+  if (compiler.getHeaderSearchOpts().UseLibcxx)
+    return CXXStdLib::Libcxx;
   return CXXStdLib::Libstdcxx;
 }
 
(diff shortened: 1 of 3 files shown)


kimgr:
I had an idea to split the logic for deciding cstdlib and cxxstdlib out of `InitGlobals`, while at the same time giving them access to the `CompilerInstance` to allow inspection of compiler flags.

Please take a look: https://github.com/include-what-you-use/include-what-you-use/pull/1241

Based on this, I think you could get rid of the `use_libcxx` flag, which is just used to smuggle information into `InitGlobals`, and inspect the `CompilerInstance` directly to see which stdlib was selected there.

And I think it should be possible to do some further smarts there later to e.g. select the right C standard library depending on target information, and remove C++ mappings from C-only compilations.

kimgr:
Moving this question here:

> It looks like we can identify that we want libcxx from compiler->getHeaderSearchOpts().UseLibcxx. That doesn't feel
> as clear to me as tool->getToolchain()->GetCXXStdLibType(), but it will do. If you know of a better way, do let me know

I think the benefit of `GetCXXStdLibType` is it returns an enum, so we can use it in a switch statement, which will warn if we don't handle all cases. That can be highly useful if Clang adds explicit support for a new standard library (e.g. the MSVC STL).

Oh. I blindly assumed `ToolChain` was available from `CompilerInstance`. That's apparently not the case. Yeah, in that case I think the header-search opts' `UseLibcxx` flag is the way to go. Hmm. Or you could explicitly parse the `OPT_stdlib_EQ` flag, like CompilerInvocation does when it populates `UseLibcxx`, but that doesn't really give any benefit over just using the bool.

Sorry about the confusion on my end.

kimgr:
Very slick, thanks!

kilroyd:
Looks like something's gone wrong here. `compiler.getHeaderSearchOpts().UseLibcxx` remains false on MacOS. I thought I'd actually checked this works, but I must have assumed it would work as advertised :(

I'll see what I can do to fix this up

PR #1239, opened 2023-04-16
Provide default values for class members

bolshakov-a:
Otherwise, one could get an undefined behavior when forgetting to set them in a constructor.

No functional change.


kimgr:
Nice, thank you!

PR #1241, opened 2023-04-16
Add functions to derive standard libs from compilation

kimgr:
Pass CompilerInstance directly to InitGlobals to give it access to the entire compilation state.

No functional change.


kilroyd (approved):
These changes looks good to me.

It looks like we can identify that we want libcxx from `compiler->getHeaderSearchOpts().UseLibcxx`. That doesn't feel as clear to me as `tool->getToolchain()->GetCXXStdLibType()`, but it will do. If you know of a better way, do let me know

Issue #1243, opened 2023-04-19
Regexes in mapping files not working


iven:
I'm using the [newest libcxx.imp](https://github.com/include-what-you-use/include-what-you-use/blob/14e9b208914a84fcdf49bf9f5d08897a4b3dc4b8/libcxx.imp) in this repo, and the mappings seem to be correctly imported:
![图片](https://user-images.githubusercontent.com/56165/233038074-ab0df786-45f2-420b-aadd-94cf0e8642e7.png)
But in the result include list, `<__algorithm/copy.h>` and other headers with `__` prefix is not replaced with `<algorithm>` as expected:
![图片](https://user-images.githubusercontent.com/56165/233037612-5030c5dc-2d0d-43f0-ab76-33dbd1b9aeac.png)
Tested both on `clang_15` and `clang_16` branch.

Is this intended? What should I do if I want to replace headers like `<__algorithm/copy.h>` with `<algorithm>`?

(I digged into the code a bit, and found that `<__algorithm/copy.h>` is not added by `AddDirectInclude()`, while `ExpandRegexes()` only works on headers that added by `AddDirectInclude()`. Not sure it's by design or not.)

kilroyd:
Odd... it works for me (on master), but I don't think anything's significantly changed wrt imp handling. Do you have the IWYU command line that you're using? Also, have you tried to set `-Xiwyu --no_default_mappings` to see if it changes behavior?

iven:
I'm using `include-what-you-use -Xiwyu --no_fwd_decls -Xiwyu --mapping_file=libcxx.imp`. Setting `-Xiwyu --no_default_mappings` doesn't help.

I did some debuging and found that `<__algorithm/copy.h>` is not seen by `quoted_includes_to_quoted_includers_` in `ExpandRegexes()`, but can be seen by `GetCandidateHeadersForFilepathIncludedFrom()`, not sure why.

BTW, I'm using precompiled headers, I don't know if it's related.

kilroyd:
If you have iwyu source, you can use one of its test files, `iterator.cc`, which uses `<algorithm>`. I've been using it as a test case here.

```
$ include_what_you_use -Xiwyu --no_default_mappings -Xiwyu --mapping_file=libcxx.imp -I. tests/cxx/iterator.cc

(tests/cxx/iterator.cc has correct #includes/fwd-decls)

$ include-what-you-use -Xiwyu --no_default_mappings -I. tests/cxx/iterator.cc

[...]

PR #1246, opened 2023-04-20
Fix for elaborated types in pch

kilroyd:
Found while investigating #1243. Based on the code in `VisitUsingDecl`

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -4204,9 +4204,21 @@ class IwyuAstConsumer
 
   bool VisitElaboratedTypeLoc(ElaboratedTypeLoc type_loc) {
     if (type_loc.getTypePtr()->getKeyword() != clang::ETK_None) {
-      preprocessor_info()
-          .FileInfoFor(CurrentFileEntry())
-          ->AddElaboratedType(type_loc);
+      IwyuFileInfo* fileinfo =
+          preprocessor_info().FileInfoFor(CurrentFileEntry());
+      if (fileinfo) {
+        fileinfo->AddElaboratedType(type_loc);
+      } else {
+        // For declarations in a PCH, the preprocessor won't have any
+        // location information. As far as we know, that's the only time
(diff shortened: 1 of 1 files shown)


kimgr:
The PCH behavior is leaving me no peace. I think the defensive handling in `VisitUsingDecl` is basically wrong; if that was a principled approach, we would need to do the same for every `FileInfoFor` lookup. The problem is we will continue running, but we will have a very fragmented idea of the full (uses x includes) image, which leads to a broken analysis.

So I suspect a more honest approach would be to exit early if `ImplicitPCHInclude` is enabled. There's no use doing IWYU analysis if arbitrary code is made available by a PCH in all translation units.

kilroyd:
I'm happy to drop this pull request if you (or @bolshakov-a) have something in mind. I had it in my tree so thought it worth sharing - as we're otherwise crashing with PCH's at the moment.

bolshakov-a:
@kimgr, I haven't understood fully. Do you suggest `CanIgnoreCurrentASTNode()` call? I suspect it won't help for traversal instantiations of templates from PCH. I think David's approach goes, because IWYU doesn't need stuff for analysis in implicitly imported files.

Thanks!

kimgr:
> I haven't understood fully. Do you suggest CanIgnoreCurrentASTNode() call? 

No, sorry I was unclear. I mean reject IWYU invocations with `ImplicitPCHInclude` with an error.

It's pretty harsh, but it makes it clear: IWYU cannot reason about programs where the physical/preprocessor structure has been hidden by precompiled headers.

kilroyd:
OK, I'll drop this commit then

bolshakov-a:
@kimgr, thanks for explanation! Reasonable.

kimgr:
It might also be interesting to drop the `ImplicitPCHInclude` option from the compilation, and blindly assume that the program compiles fine without precompiled header. All programs in good taste do :). Possibly hidden behind an `--ignore_pch` switch.

PR #1248, opened 2023-04-23
Handle alias components separately

bolshakov-a:
Separate determination of "provision" status for typedef or type alias underlying type components. It allows for e. g. some template arguments being provided and some not (alias user is responsible to `#include` them if they are required).

On alias declaration site, it is achieved with setting defaulted fwd-decl context flag and with reporting types if they are descendants of `TypedefNameDecl` and are "provided". On alias use site, underlying type components are enumerated and provided ones are blocked for report.

New test cases added, existing tests fixed (`basic_fstream` and `basic_string` in `badinc` test are reported due to their explicit instantiations).

I intend to make similar PRs for "autocast" and function return type handling later. (Or it should be incorporated into this one?)

Btw, it is the same PR which I mentioned in #1057, if I remember correctly...

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2803,10 +2803,12 @@ class InstantiatedTemplateVisitor
   void ScanInstantiatedFunction(
       const FunctionDecl* fn_decl, const Type* parent_type,
       const ASTNode* caller_ast_node,
-      const map<const Type*, const Type*>& resugar_map) {
+      const map<const Type*, const Type*>& resugar_map,
+      const set<const Type*>& blocked_types) {
     Clear();
     caller_ast_node_ = caller_ast_node;
     resugar_map_ = resugar_map;
+    blocked_types_ = blocked_types;
 
     // Make sure that the caller didn't already put the decl on the ast-stack.
     CHECK_(caller_ast_node->GetAs<Decl>() != fn_decl && "AST node already set");
@@ -2823,7 +2825,7 @@ class InstantiatedTemplateVisitor
   // MyClass<T>::size_type s;
   void ScanInstantiatedType(ASTNode* caller_ast_node,
                             const map<const Type*, const Type*>& resugar_map,
-                            const set<const Type*>& blocked_types = {}) {
+                            const set<const Type*>& blocked_types) {
     Clear();
     caller_ast_node_ = caller_ast_node;
     resugar_map_ = resugar_map;
@@ -4232,8 +4234,9 @@ class IwyuAstConsumer
       const map<const Type*, const Type*> resugar_map =
           GetTplTypeResugarMapForClass(type);
 
-      instantiated_template_visitor_.ScanInstantiatedType(current_ast_node(),
-                                                          resugar_map);
+      instantiated_template_visitor_.ScanInstantiatedType(
+          current_ast_node(), resugar_map,
+          ExtractProvidedTypeComponents(resugar_map));
     }
 
     return Base::VisitTemplateSpecializationType(type);
@@ -4309,8 +4312,8 @@ class IwyuAstConsumer
     }
 
     instantiated_template_visitor_.ScanInstantiatedFunction(
-        callee, parent_type,
-        current_ast_node(), resugar_map);
+        callee, parent_type, current_ast_node(), resugar_map,
+        ExtractProvidedTypeComponents(resugar_map));
     return true;
   }
 
@@ -4322,8 +4325,11 @@ class IwyuAstConsumer
         GetTplTypeResugarMapForClass(type);
     ASTNode node(type);
     node.SetParent(current_ast_node());
+    set<const Type*> merged_blocked =
+        ExtractProvidedTypeComponents(resugar_map);
+    merged_blocked.insert(blocked_types.begin(), blocked_types.end());
     instantiated_template_visitor_.ScanInstantiatedType(&node, resugar_map,
-                                                        blocked_types);
+                                                        merged_blocked);
   }
 
   const set<const Type*>& GetBlockedTypes() const {
@@ -4332,6 +4338,22 @@ class IwyuAstConsumer
   }
 
  private:
+  set<const Type*> ExtractProvidedTypeComponents(
+      const map<const Type*, const Type*>& resugar_map) const {
+    set<const Type*> result;
+    for (const auto& canonical_sugared_pair : resugar_map) {
+      const Type* desugared_until_typedef =
+          Desugar(canonical_sugared_pair.second);
+      if (const auto* typedef_type =
+              dyn_cast_or_null<TypedefType>(desugared_until_typedef)) {
+        set<const Type*> provided =
+            GetProvidedTypesForTypedef(typedef_type->getDecl());
+        result.insert(provided.begin(), provided.end());
+      }
+    }
+    return result;
+  }
+
   // Class we call to handle instantiated template functions and classes.
   InstantiatedTemplateVisitor instantiated_template_visitor_;
 };  // class IwyuAstConsumer
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1437,7 +1437,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return underlying_type;
   }
 
-  set<const Type*> GetCallerResponsibleTypesForTypedef(
+  set<const Type*> GetProvidedTypesForTypedef(
       const TypedefNameDecl* decl) const {
     set<const Type*> retval;
     const Type* underlying_type = decl->getUnderlyingType().getTypePtr();
@@ -1447,22 +1447,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       if (const auto* underlying_typedef_decl = dyn_cast<TypedefNameDecl>(TypeToDeclAsWritten(underlying_typedef))) {
         // TODO(csilvers): if one of the intermediate typedefs
         // #includes the necessary definition of the 'final'
-        // underlying type, do we want to return the empty set here?
-        return GetCallerResponsibleTypesForTypedef(underlying_typedef_decl);
+        // underlying type, do we want to return it here?
+        return GetProvidedTypesForTypedef(underlying_typedef_decl);
       }
     }
 
-    const Type* deref_type =
-        RemovePointersAndReferencesAsWritten(underlying_type);
-    if (isa<SubstTemplateTypeParmType>(underlying_type) ||
-        CodeAuthorWantsJustAForwardDeclare(deref_type, GetLocation(decl))) {
-      retval.insert(deref_type);
-      // TODO(csilvers): include template type-args if appropriate.
-      // This requires doing an iwyu visit of the instantiated
-      // underlying type and seeing which type-args we require full
-      // use for.  Also have to handle the case where the type-args
-      // are themselves templates.  It will require pretty substantial
-      // iwyu surgery.
+    for (const Type* type :
+         GetComponentsOfTypeWithoutSubstituted(underlying_type)) {
+      if (!CodeAuthorWantsJustAForwardDeclare(type, GetLocation(decl))) {
+        retval.insert(type);
+      }
     }
     return retval;
   }
@@ -1632,67 +1626,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // with the warning message that iwyu emits.
   virtual void ReportTypeUse(SourceLocation used_loc, const Type* type,
                              const char* comment = nullptr) {
-    if (CanIgnoreType(type))
-      return;
-
-    // Enum type uses can be ignored. Their size is known (either implicitly
-    // 'int' or from a mandatory transitive inclusion of a non-fixed enum full
-    // declaration, or explicitly using a C++ 11 enum base). Only if an enum
-    // type or its enumerators are explicitly mentioned will they be reported
-    // by IWYU from VisitTagType or VisitDeclRefExpr correspondingly.
-    if (type->getAs<EnumType>())
-      return;
-
-    // Types in fwd-decl-context should be ignored here and reported from more
-    // specialized places, i.e. when they are explicitly written. But in fact,
-    // this check is redundant because TypeToDeclAsWritten returns nullptr for
-    // pointers and references.
-    if (IsPointerOrReferenceAsWritten(type))
-      return;
-
-    // For typedefs, the user of the type is sometimes the one
-    // responsible for the underlying type.  We check if that is the
-    // case here, since we might be using a typedef type from
-    // anywhere.  ('autocast' is similar, but is handled in
-    // VisitCastExpr; 'fn-return-type' is also similar and is
-    // handled in HandleFunctionCall.)
-    if (const auto* typedef_type = type->getAs<TypedefType>()) {
-      // One exception: if this TypedefType is being used in another
-      // typedef (that is, 'typedef MyTypedef OtherTypdef'), then the
-      // user -- the other typedef -- is never responsible for the
-      // underlying type.  Instead, users of that typedef are.
-      const ASTNode* ast_node = MostElaboratedAncestor(current_ast_node());
-      if (!ast_node->ParentIsA<TypedefNameDecl>()) {
-        const TypedefNameDecl* typedef_decl = typedef_type->getDecl();
-        const set<const Type*>& underlying_types =
-            GetCallerResponsibleTypesForTypedef(typedef_decl);
-        if (!underlying_types.empty()) {
-          VERRS(6) << "User, not author, of typedef "
-                   << typedef_decl->getQualifiedNameAsString()
-                   << " owns the underlying type:\n";
-          // If any of the used types are themselves typedefs, this will
-          // result in a recursive expansion.  Note we are careful to
-          // recurse inside this class, and not go back to subclasses.
-          for (const Type* type : underlying_types)
-            IwyuBaseAstVisitor<Derived>::ReportTypeUse(used_loc, type);
-        }
-      }
-      return;
-    }
-
-    // Map private types like __normal_iterator to their public counterpart.
-    type = MapPrivateTypeToPublicType(type);
-    // For the below, we want to be careful to call *our*
-    // ReportDeclUse(), not any of the ones in subclasses.
-    if (const auto* template_spec_type =
-            dyn_cast<TemplateSpecializationType>(Desugar(type))) {
-      this->getDerived().ReportTplSpecComponentTypes(template_spec_type);
-    }
-    if (const NamedDecl* decl = TypeToDeclAsWritten(type)) {
-      decl = GetDefinitionAsWritten(decl);
-      VERRS(6) << "(For type " << PrintableType(type) << "):\n";
-      IwyuBaseAstVisitor<Derived>::ReportDeclUse(used_loc, decl, comment);
-    }
+    ReportTypeUseInternal(used_loc, type, comment,
+                          this->getDerived().GetBlockedTypes());
   }
 
   void ReportTypesUse(SourceLocation used_loc, const set<const Type*>& types) {
@@ -1751,22 +1686,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // itself can be forward-declared, just as in (2).  That is:
   //   template<class T> struct C { typedef pair<T,T> value_type; };
   // iwyu will demand the full type of pair, but not of its template
-  // arguments.  This is handled not here, but below, in
-  // VisitSubstTemplateTypeParmType.
+  // arguments.  This is handled in VisitSubstTemplateTypeParmType.
+  // Typedef underlying type is marked fwd-declarable by default, and the
+  // aforementioned analysis is performed with 'IsProvidedTypeComponent'
+  // function.
   bool VisitTypedefNameDecl(clang::TypedefNameDecl* decl) {
     if (CanIgnoreCurrentASTNode())
       return true;
-    const Type* underlying_type = decl->getUnderlyingType().getTypePtr();
-    const Type* deref_type =
-        RemovePointersAndReferencesAsWritten(underlying_type);
-
-    if (CodeAuthorWantsJustAForwardDeclare(deref_type, GetLocation(decl)) ||
-        isa<SubstTemplateTypeParmType>(underlying_type)) {
-      current_ast_node()->set_in_forward_declare_context(true);
-    } else {
-      current_ast_node()->set_in_forward_declare_context(false);
-    }
-
+    current_ast_node()->set_in_forward_declare_context(true);
     return true;
   }
 
@@ -2571,7 +2498,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     const NamedDecl* decl = TypeToDeclAsWritten(type);
 
     // If we are forward-declarable, so are our template arguments.
-    if (CanForwardDeclareType(current_ast_node())) {
+    if (CanForwardDeclareType(current_ast_node()) &&
+        !IsProvidedTypeComponent(current_ast_node())) {
       ReportDeclForwardDeclareUse(CurrentLoc(), decl);
       current_ast_node()->set_in_forward_declare_context(true);
     } else {
@@ -2699,6 +2627,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return visitor_state_->preprocessor_info;
   }
 
+  bool IsProvidedTypeComponent(const ASTNode* node) const {
+    return node->HasAncestorOfType<TypedefNameDecl>() &&
+           !CodeAuthorWantsJustAForwardDeclare(node->GetAs<Type>(),
+                                               node->GetLocation());
+  }
+
  private:
   template <typename T> friend class IwyuBaseAstVisitor;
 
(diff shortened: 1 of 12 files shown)


kimgr:
Could you expand a little on why `basic_fstream` and `basic_string` show up now? I wonder if we could address that in `ReportExplicitInstantiations` somehow.

iwyu_ast_util.cc
```diff
@@ -588,17 +594,37 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
...
+ public:
+  bool TraverseTypeHelper(QualType qual_type) {
+    CHECK_(!qual_type.isNull());
+    const Type* type = qual_type.getTypePtr();
+    if (type->getAs<SubstTemplateTypeParmType>())
+      return true;
+
+    seen_types_.insert(GetCanonicalType(type));
+
+    return TypeEnumerator<TypeEnumeratorWithoutSubstituted>::Base::TraverseType(
```
kimgr, line 613: Let's add a `typedef TypeEnumerator<TypeEnumeratorWithoutSubstituted> Base;` in this class, and use `retur Base::Base::TraverseType(...);`.

The double-Base looks a little funny, but it accurately highlights how far up the lineage we're jumping.

kimgr: And same in the other derived TypeEnumerator.

iwyu.cc
```diff
@@ -4272,22 +4296,41 @@ class IwyuAstConsumer
...
+    merged_blocked.insert(blocked_types.begin(), blocked_types.end());
+    instantiated_template_visitor_.ScanInstantiatedType(&node, resugar_map,
+                                                        merged_blocked);
   }
 
  private:
+  set<const Type*> ExtractProvidedComponents(
+      const map<const Type*, const Type*>& resugar_map) const {
+    set<const Type*> result;
+    for (const auto [_, sugar_type] : resugar_map) {
```
kimgr, line 4322: Very cool that destructuring assignments are becoming mainstream :).

I searched through the llvm codebase, though, and only found a single use besides tests. So I think we should stick with the old-school `for (const auto& item : resugar_map)` for now, so as not to alienate us old C++ folks.

iwyu.cc
```diff
@@ -2693,6 +2684,13 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return visitor_state_->preprocessor_info;
   }
 
+  bool IsProvidedComponent() const {
```
kimgr, line 2687: It feels slightly more explicit and nicer to me to pass in the `ASTNode*` here

iwyu.cc
```diff
@@ -1631,7 +1625,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // The comment, if not nullptr, is extra text that is included along
   // with the warning message that iwyu emits.
   virtual void ReportTypeUse(SourceLocation used_loc, const Type* type,
-                             const char* comment = nullptr) {
+                             const char* comment = nullptr,
+                             const set<const Type*>& blocked_types = {}) {
```
kimgr, line 1629: I'm not a huge fan of the `blocked_types` argument, because it smells of a control flag -- something the caller passes in only to modify the control flow of the function. My instinct is to either:

* move the checking of `type` vs. `blocked_types` out if this function and into the few callers where it matters, or
* resolve `blocked_types` (or "provided types") inside `ReportTypeUse`

But I realize that neither is probably very straightforward... Let me see if I can get my head around the interdependencies and figure something out.

kimgr: On that note, it's also a little suspicious that `ReportTypesUse` does not do anything with these provided types. I'm guessing it's because the filtering happens before types end up in the cache whose data is passed to `ReportTypesUse`. But that creates a strange implicit dependency between the cache and the type reporting. E.g. what if I wanted to use `ReportTypesUse` to report a set of types that _do_ need provision filtering? The API implies that `ReportTypesUse` is just a batch version of `ReportTypeUse`.

iwyu.cc
kimgr, line 2687: Might also be nice to name this `IsProvidedComponentType` (which seems to be how we're referring to these, though my instinct was that `IsProvidedTypeComponent` would be clearer...)

iwyu.cc
```diff
@@ -2951,7 +2953,11 @@ class InstantiatedTemplateVisitor
   }
 
   void ReportTypeUse(SourceLocation used_loc, const Type* type,
-                     const char* comment = nullptr) override {
+                     const char* comment = nullptr,
+                     const set<const Type*>& blocked_types = {}) override {
+    // 'blocked_types' argument can be filled only in the base class
+    // implementation.
+    CHECK_(blocked_types.empty());
```
kimgr, line 2962: This feels like another indication that `blocked_types` is not really an argument to `ReportTypeUse`.

iwyu_ast_util.cc
```diff
@@ -556,24 +557,29 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
 
   set<const Type*> Enumerate(const TemplateArgument& tpl_arg) {
     seen_types_.clear();
-    TraverseTemplateArgument(tpl_arg);
+    this->TraverseTemplateArgument(tpl_arg);
```
kimgr, line 560: I'm guessing `this->` qualification is not necessary here, since you don't have it for `TraverseType` above.

iwyu_ast_util.h
```diff
@@ -697,6 +697,13 @@ const clang::Type* Desugar(const clang::Type* type);
 // vector<pair<Foo, Bar>> has components pair<Foo,Bar>, Foo, and Bar.
 set<const clang::Type*> GetComponentsOfType(const clang::Type* type);
 
+// Returns types for determination of their "provision" status. They are
+// canonicalized because intermediate sugar should be always provided already
+// according to language rules. Substituted template parameter types (and their
+// underlying types) cannot be provided.
+std::set<const clang::Type*> GetComponentsOfTypeWithoutSubstituted(
```
kimgr, line 704: Could you add an example to contrast with the ones on lines 696-697? I find the "without substituted" naming a little opaque, but mostly because I don't understand all the AST details around template instantiation. So for the `Foo*` and `vector<Foo>` cases, will we still get `Foo`? And for `vector<pair<Foo,Bar>>` will we get the same as above (assuming `pair` is not a typedef? Might be interesting to expand that example with type aliases to show the difference.

tests/cxx/badinc.h
```diff
@@ -297,10 +296,12 @@ typedef std::vector<I2_Class> H_I2Class_Vector_Unused;
 // IWYU: I2_Enum is...*badinc-i2.h
 typedef I2_TemplateClass<I2_Enum> H_TemplateTypedef;
 
-// IWYU: I2_Struct needs a declaration
+// IWYU: I2_Struct is...*badinc-i2.h
```
kimgr, line 299: This one looks a little weird. Why not a forward declare?

tests/cxx/badinc.h
```diff
@@ -297,10 +296,12 @@ typedef std::vector<I2_Class> H_I2Class_Vector_Unused;
...
 typedef I2_Struct* H_StructPtr;
 
-// IWYU: I2_Class needs a declaration
+// forward-declaration suggestion may be better here,
```
kimgr, line 302: forward -> Forward

iwyu.cc
```diff
@@ -1751,22 +1752,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-        RemovePointersAndReferencesAsWritten(underlying_type);
-
-    if (CodeAuthorWantsJustAForwardDeclare(deref_type, GetLocation(decl)) ||
-        isa<SubstTemplateTypeParmType>(underlying_type)) {
-      current_ast_node()->set_in_forward_declare_context(true);
-    } else {
-      current_ast_node()->set_in_forward_declare_context(false);
-    }
-
+    if (CanIgnoreCurrentASTNode())  return true;
```
kimgr, line 1759: This looks copy-pasted from before 890a3150d74f834a31e57c649195d6d68da682b1. Let's clang-format the line.

kimgr:
Thanks! Most of the test case changes look like strict improvements, questions about the others inline. Overall I also find the `WithouthSubstituted` naming a little confusing, hoping some concrete examples can help clarify there.

bolshakov-a:
> Could you expand a little on why basic_fstream and basic_string show up now? I wonder if we could address that in ReportExplicitInstantiations somehow.

Prior to my changes, underlying type analysis of these typedefs didn't even start because the type was considered as fully provided. I think the behavior on my branch is more correct, because for such piece of code:
```cpp
// Tpl.h
template <typename>
struct Tpl{};

typedef Tpl<char> Alias;

// explicit-instantiation-header.h
#include "Tpl.h"

extern template struct Tpl<char>;

// main.cpp
#include "Tpl.h"
#include "explicit-instantiation-header.h"

Alias t;
```
IWYU from the master branch suggests to remove `#include "explicit-instantiation-header.h"` from `main.cpp` and to keep on my branch.

Determining whether or not a typedef provides explicit instantiation may be difficult because, specifically, `stringfwd.h` (where `std::string` typedef is defined) from libstdcxx doesn't include `basic_string.tcc` where explicit instantiation is. Some sort of `PublicHeaderIntendsToProvide` logic might help, but it seems like overengineering. Or, maybe, is it important?

iwyu.cc
bolshakov-a: As an alternative, I can suggest moving `blocked_types_` field from `InstantiatedTemplateVisitor` into `IwyuBaseAstVisitor` as a protected field. `IwyuBaseAstVisitor::ReportTypeUse` function would then fill in the set when it finds a typedef and then restore its state.

tests/cxx/badinc.h
bolshakov-a: I've decided simply to expand IWYU policy (stating that typedef author should provide underlying type by default) to typedef underlying type components. `I2_Struct` is just a component of `I2_Struct*`. It should be possible to override it by means of fwd-declaring and not including the header directly.

kimgr:
> Or, maybe, is it important?

It seems like a pretty painful regression if uses of `std::fstream` result in mentions of `basic_fstream`. Ah, but it's only in diagnostics and `why`-comments? I guess it doesn't affect the suggested includes?

bolshakov-a:
Yes, it should not affect suggested includes, because it seems like all of the stuff should be provided by `<string>` and `<fstream>` headers, correspondingly.

iwyu.cc
kimgr: That also feels a little funny. Let's think on this for a wile and see if a third alternative comes up.

tests/cxx/badinc.h
kimgr: Nice, I see. 

I keep going back and forth... At first I thought it felt a little blunt to consider `I2_Struct` a component of `I2_Struct*`, then I thought it makes a lot of sense for consistency.

But there seems to be tension between the definition loc and the usage loc now;

* typedef location is responsible for the full type even if it only uses a pointer
* can work around that by forward-declaring the type
* ... but that implies the usage loc is responsible

Or did I misunderstand that?

kimgr:
@bolshakov-a Yeah, that sounds like a reasonable tradeoff, then. If the net result is better, I think we can live with some noise in reporting. I can try and think something up for the explicit instantiations later.

tests/cxx/badinc.h
bolshakov-a: > ... but that implies the usage loc is responsible

Of course. If a typedef author doesn't provide a component which is needed at the use site (e.g. it is absent in `provided_types` / `blocked_types`), then a user should `#include` the corresponding header.

In fact, pointer typedefs are not always handled correctly at use site both on master and my branch. IWYU often suggests to `#include` an underlying type header when it is redundant. I want to address this with something like [such a code](https://github.com/bolshakov-a/include-what-you-use/commit/a043e0013fe0b8c2262eeb32f8e958f5368f2c7f).

iwyu_ast_util.cc
bolshakov-a: Neither MSVC nor GCC compiles this without `this->`. The difference with `TraverseType` is that the called overload of the latter is declared directly in `TypeEnumerator` and doesn't come from the templated base.

iwyu.cc
bolshakov-a: Maybe, it's because LLVM switched to C++17 relatively recently.

iwyu_ast_util.h
bolshakov-a: I've added a couple of examples. It's worth noting that these functions return not only "components" but also the given original type. It is exactly what is needed for the purpose of this PR. But, maybe, naming of the functions could be improved somehow...

iwyu.cc
bolshakov-a: > move the checking of `type` vs. `blocked_types` out if this function and into the few callers where it matters

The problem here is that `ReportTplSpecComponentTypes` analysis should be launched even when template specialization type is "blocked" (see [the comment](https://github.com/include-what-you-use/include-what-you-use/pull/1248/commits/7ac841682320b15b313edd65bca71ad655823f1c#diff-7ca107c8b836c1765e8b735b74f06833bec48f308a76017886cee1d5736002d9R1687)). Hence, `ReportTypeUse` should be called even when the type is blocked.

A virtual (or non-virtual using CRTP) function `GetBlockedTypes()` may be introduced which would return `blocked_types_` for `InstantiatedTemplateVisitor` and an empty set for `IwyuAstConsumer`. But how `IwyuBaseAstVisitor::ReportTypeUse` could transfer `provided_types` of a typedef between its recursive calls if not with an argument or class state? I can suggest moving its body into private function `IwyuBaseAstVisitor::ReportTypeUseInternal` which would take `blocked_type` parameter and its public (for subclasses) interface `ReportTypeUse` would not.

kimgr: > I can suggest moving its body into private function IwyuBaseAstVisitor::ReportTypeUseInternal which
> would take blocked_type parameter and its public (for subclasses) interface ReportTypeUse would not.

That actually sounds nicer to me.

`ReportTypeUse` would then get the type components and pass them as "blocked" as necessary? That would also nicely solve the `ReportTypesUse` asymmetry.

iwyu_ast_util.cc
kimgr: Aha, I see, thanks

iwyu.cc
kimgr: (please add it first as a separate commit on top, so we can change our minds if it somehow turns out worse)

tests/cxx/badinc.h
bolshakov-a: Seems like the original IWYU authors meant that pointed-to types may also be subject to "providing", as the example with function at the end of ["Author intent" section](https://github.com/include-what-you-use/include-what-you-use/blob/0.20/docs/WhatIsAUse.md#author-intent) shows. And later:
> As of May 2011, the rule does not apply when returning a pointer or reference
> ...
> This is considered a bug ...

Honestly, I don't think that this "bug" regarding function return type will be fixed some day because IWYU quickly loses information that a used type is from a function return. But for typedefs, it can be done because typedef type is tracked in many cases.

`RemovePointersAndReferencesAsWritten` calls in [`GetCallerResponsibleTypesForTypedef`](https://github.com/include-what-you-use/include-what-you-use/blob/0.20/iwyu.cc#L1454) and [`VisitTypedefNameDecl`](https://github.com/include-what-you-use/include-what-you-use/blob/0.20/iwyu.cc#L1759) versions from the master branch cause as well a thought that pointed-to types could be "provided" and should obey the default policy rules.

Btw, I want to introduce some day a command line flags to suppress that policies and require forward declarations of components of typedefs, function arguments and returns.

kimgr: I have an old branch to play with custom AST traversal to be able to analyze the constituent parts of function types: https://github.com/kimgr/include-what-you-use/tree/function-type-traversal. I think that might be a way to maintain the information about whether a type is found as a function return type, exception spec, or something else.

Now if only I could get my nose over water long enough to get it more finished... :-)

iwyu.cc
```diff
@@ -2712,7 +2646,83 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // void fn(const S<Class>& s) { // Forward declarations are sufficient here.
   //   (void)s.t; // Full 'Class' type is needed due to template instantiation.
   // }
-  void ReportTplSpecComponentTypes(const TemplateSpecializationType*) = delete;
+  void ReportTplSpecComponentTypes(const TemplateSpecializationType*,
+                                   const set<const Type*>& blocked_types) =
+      delete;
+
+  const std::set<const Type*>& GetBlockedTypes() const = delete;
```
kimgr, line 2653: Let's use `set` without namespace qualification for new code. I saw there was some explicit qualification in existing code, but I can clean that up after.

iwyu.cc
```diff
@@ -2712,7 +2646,83 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+
+  void ReportTypeUseInternal(SourceLocation used_loc, const Type* type,
+                             const char* comment,
+                             const set<const Type*>& blocked_types) {
```
kimgr, line 2657: I'm still not 100% on the "blocked" terminology.

I see that the callers use it to filter out already-provided types, and I appreciate the layering that we could use it for something else as well, i.e. we probably don't want it to be called `provided_types` here.

Or would that be clearer? Less general, for sure, but more to the point.

iwyu.cc
```diff
@@ -4034,6 +4057,16 @@ class IwyuAstConsumer
     return Base::VisitUsingDirectiveDecl(decl);
   }
 
+  bool AllTypesProvided(const Type* typedef_underlying_type) const {
```
kimgr, line 4060: I'd prefer it if this took the location as a parameter as well, to avoid the implicit dependency on `current_ast_node()`.

iwyu_ast_util.cc
```diff
@@ -556,24 +557,29 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
...
-    return TraverseTypeHelper(type_loc.getType());
+    return this->getDerived().TraverseTypeHelper(type_loc.getType());
   }
 
- private:
+ protected:
+  set<const Type*> seen_types_;
+};
+
+class TemplateArgEnumerator : public TypeEnumerator<TemplateArgEnumerator> {
```
kimgr, line 581: I was surprised by this name, I would have expected this to cover all sorts of types that aren't necessarily template arguments?

This looks more to me like `SugaredTypeEnumerator`?

iwyu.cc
```diff
@@ -4272,22 +4307,47 @@ class IwyuAstConsumer
...
+                                                        merged_blocked);
+  }
+
+  const std::set<const Type*>& GetBlockedTypes() const {
+    static const std::set<const Type*> empty_set;
+    return empty_set;
   }
 
  private:
+  set<const Type*> ExtractProvidedComponents(
```
kimgr, line 4335: Maybe `ExtractProvidedTypeComponents` to be fully explicit.

iwyu_ast_util.cc
```diff
@@ -587,18 +593,41 @@ class TypeEnumerator : public RecursiveASTVisitor<TypeEnumerator> {
...
-    return Base::TraverseType(QualType(desugared, 0));
+    return Base::Base::TraverseType(QualType(desugared, 0));
   }
 
-  set<const Type*> seen_types_;
+ private:
+  typedef TypeEnumerator<TemplateArgEnumerator> Base;
+};
+
+class TypeEnumeratorWithoutSubstituted
```
kimgr, line 603: These two enumerators share some abstract shape, but not much else. I wonder if it would be clearer/easier to name them if they were not associated by inheritance at all, but rather just side-by-side implementations? That is, just `TypeEnumerator` and `TypeEnumeratorWithoutSubstituted`. The `Enumerate` overload for `TemplateArgument` would not be necessary for the latter.

tests/cxx/badinc.h
kimgr: I think I agree with the more aggressive type requirement for pointers/references, but it's hard to see how it will affect real-world codebases.

bolshakov-a: > I have an old branch to play with custom AST traversal to be able to analyze the constituent parts of function types

Interesting, thanks! I think, though, that simple setting fwd-decl-context flag is insufficient for for determination of whether a type is a provided component of a fn-return type, so I've decided to use `ASTNode` stack for that purpose:
https://github.com/bolshakov-a/include-what-you-use/commit/d4666e32c74575c8b0f5197b109c5a471e3c17d1

Btw, I've just realized that the latest refactoring in this PR doesn't fit very well to the suggested way of reporting provided components of a fn-return:
https://github.com/bolshakov-a/include-what-you-use/commit/4a28fbc8c0371d41aed6341439e55a549dbafdfb#diff-7ca107c8b836c1765e8b735b74f06833bec48f308a76017886cee1d5736002d9R4212
Maybe, it can be replaced with some scoped setting of "provided types" set which would be returned from `IwyuAstConsumer::GetBlockedTypes()`.

> how it will affect real-world codebases

If an old header follows the old IWYU rules, then it already has the fwd-declaration and doesn't have `#include` in that case, generally. So the new rules shouldn't affect it. In a new headers, writers should pay more attention on "provision" policy which they want. But for value types, it is the case even with the old rules, of course. E.g., some style guide may insist that function parameter and return types should be fwd-declared regardless of whether they are pointers or not. So, code writers should check and correct IWYU output by itself.

iwyu.cc
bolshakov-a: This function reports `type` and its underlying types without types listed in `blocked_types`. So, this parameter is exactly for that purpose as it is named — it prevents some types from reporting — except the case with type aliases when a set of types "provided" with typedef supersedes `blocked_types` set.

Maybe, you have some suggestions about naming?

iwyu_ast_util.cc
bolshakov-a: This is just use case encoded in the name. But following that logic, `TypeEnumeratorWithoutSubstituted` should be named as `TypeEnumeratorToGetProvidedTypes` or something similar... On the one hand, it is not a good approach. On the other hand, these "enumerators" are inherently very specific.

iwyu_ast_util.cc
bolshakov-a: I might not fully understand your naming suggestion.

kimgr: Sorry, I see how that was unclear.

I mean that it could be nicer to separate these two AST visitors entirely, and not use inheritance to reuse code between them. I don't think there's any strict is-a relationship between them, and the reuse seems limited to a few lines.

But I could be underestimating the wins from sharing code, maybe it's worth it.

iwyu.cc
kimgr: The best I can come up with is `provided_types`, but I agree that that is not ideal.

It would be nice if it was possible to invert the logic entirely -- recursively collect all types we want to report starting from a `Type*`, filter out already-provided ("blocked") types, and then just call a dumber `ReportTypesUse` that simply does the reporting without any smarts for type components.

In general I like it when logic flows strictly forwards instead of like here, looking back into some set we prepared earlier to see if we should stop working. Hard to explain, but it feels like we're jumping back and forth here.

bolshakov-a: Unluckily, `ReportTypeUse` cannot simply report a type because it should itself launch `InstantiatedTemplateVisitor` analysis if the type occurs to be `TemplateSpecializationType` (see [`ReportTplSpecComponentTypes`](https://github.com/include-what-you-use/include-what-you-use/blob/0.20/iwyu.cc#L1688)). `blocked_types` parameter is mostly for propagating into it. In principle, it can be inverted into `allowed_types` along with the field member of `InstantiatedTemplateVisitor`, but I think that `blocked_types` suits better to [proposed implementation of arbitrary dependent type reporting](https://github.com/include-what-you-use/include-what-you-use/pull/1182#issuecomment-1406681708):
> A solution to this may be to traverse an original template declaration with "reporting" types in a list, and then use this list as a list of types blocked for reporting on template specialization decl traversal. Reports of all other types should then be allowed.

kimgr: > It would be nice if it was possible to invert the logic entirely -- recursively collect all types we want to report starting
> from a Type*, filter out already-provided ("blocked") types, and then just call a dumber ReportTypesUse that simply 
> does the reporting without any smarts for type components.

I realize it's not super useful in its current form, but I wanted to sketch out what I meant above:
https://github.com/kimgr/include-what-you-use/commit/kg-expand-instead-of-report

It feels like we could replace InstantiatedTemplateVisitor with something that just collects types, but maybe I'm just romanticizing :)

kimgr: Just to be clear: I don't think it's realistic to adopt any of my sketch above, particularly not for this PR, but I'm interested to hear what you think of its viability. Would it be possible to decouple `ReportTplSpecComponentTypes` from the AST visitation?

iwyu_ast_util.cc
kimgr: Aha, I didn't see it that way at all (use-case encoded in name). I was thinking it enumerated sugared types and returned a desugared set. But maybe `DesugaredTypeEnumerator` makes it clearer what it produces? Or I'm entirely confused.

iwyu.cc
bolshakov-a: I think it may be just a naming problem. I. e., `ReportTypeUse` is not actually `Report...` but `Handle...` or something like that?

Btw, "reporting" types to `std::set` should probably be done on implementing improvement of template dependent type handling.

> There are never any blocked types/provided types in IwyuAstConsumer

It will probably not be the case after implementing separate type component handling for fn-return and "autocast".

iwyu_ast_util.cc
bolshakov-a: No, just recall: it is used to produce _sugared_ side of `resugar_map`. But it keeps only "interesting" sugar (typedefs, template specializations, see `desugared_until_alias_or_tpl`). If that sugar hides itself another complex type, the visitor traverses its components too.

iwyu.cc
bolshakov-a: I understand your wish to diminish complexity, but `ReportTypeUse` is currently the central place to handle types that are needed fully but aren't explicitly written in the source. If that type occurs to be typedef, the "provision" logic should show whether its underlying type is needed or not. If that type occurs to be a template specialization, the handling should be done similar to as if it were explicitly written, and currently `InstantiatedTemplateVisitor` performs that handling.

bolshakov-a: > Would it be possible to decouple ReportTplSpecComponentTypes from the AST visitation?

I think recursive calls `Traverse... -> Report(Handle)TypeUse -> Traverse... -> Report(Handle)TypeUse -> ...` are unavoidable in general. Consider:
```cpp
struct A {};

template <typename T>
struct Tpl3 {
    int b;
    T t;
};

template <typename T>
struct Tpl2 {
    int a;
    Tpl3<T>* t3;
    decltype(t3->b) n;
};

template <typename T>
struct Tpl1 {
    int i;
    Tpl2<T>* t2;
    decltype(t2->a) k;
};

void Fn(Tpl1<A>* t1) {
    (void)sizeof(*t1);
}
```
`sizeof` expression in `Fn` constitutes full use of `Tpl1<A>`, which requires full type info for its member fields, whereas declaration of one of them (`Tpl1::k`) contains `MemberExpr` for `Tpl2<A>`, constituting its full use, which requires full type info for its fields, and so on. (Currently, IWYU doesn't handle this case correctly because `CanIgnoreType` with default argument `IgnoreKind::ForUse` returns `true` for `Tpl2<A>` during `Tpl1<A>` visitation).

But if you insist, I can try to implement "reporting" into `std::set` on an experimental branch. `Report(Expand)TplSpecComponentTypes` would then use its own stack-allocated `InstantiatedTemplateVisitor` object. This approach should allow using `set_difference` instead of passing `blocked_types` into `InstantiatedTemplateVisitor`, as you've suggested.

kimgr: > But if you insist, I can try to implement "reporting" into std::set on an experimental branch.
> Report(Expand)TplSpecComponentTypes would then use its own stack-allocated
> InstantiatedTemplateVisitor object. This approach should allow using set_difference instead
> of passing blocked_types into InstantiatedTemplateVisitor, as you've suggested.

I'm in no position to insist :), but it would be interesting to see. As I've mentioned several times, my understanding of the template traversal is fuzzy at best, so I don't feel comfortable trying to hack it up myself.

I don't think we should block this PR on that experiment, however. I would like to get this merged, but I still find the `blocked_types` terminology a little off. Will try to find time to come up with other suggestions.

kimgr:
@bolshakov-a I think this mostly looks good. Could you slice out `Extend comment for GetComponentsOfType after #1182` commit out into a separate PR, and then rewrite the history for this one to have the `GetBlockedTypes` refactor integrated into the original commits? I'll give the naming a bit more thought, but if nothing beautiful comes up, we should just get on with it.

iwyu.cc
```diff
@@ -2718,7 +2652,83 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-  void ReportTplSpecComponentTypes(const TemplateSpecializationType*) = delete;
+  void ReportTplSpecComponentTypes(const TemplateSpecializationType*,
+                                   const set<const Type*>& blocked_types) =
+      delete;
+
+  const set<const Type*>& GetBlockedTypes() const = delete;
+
+  void ReportTypeUseInternal(SourceLocation used_loc, const Type* type,
+                             const char* comment,
+                             const set<const Type*>& provided_types) {
```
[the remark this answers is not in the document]
bolshakov-a: I had decided to divide terminology into "provided" and "blocked" because types may be "blocked" not only because they are "provided by author". E.g., I intended to use `blocked_types_` inside `InstantiatedTemplateVisitor` also for blocking types on the instantiation side which are used on the declaration side, when (and if) implementing improved dependent type handling in the future. On the other hand, such types can also be considered as "provided by template declaration"... Not sure.

I like "author types" and "ignored types" terminology.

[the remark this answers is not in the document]
kimgr: I see. I was expecting the division to be the inverse, i.e. `ReportTypeUse` would use "blocked" and callers would use "provided", as there are more reasons to block than types being provided by a template defn. But yeah, this is all very fuzzy. :)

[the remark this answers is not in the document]
bolshakov-a: Reasonable, thanks. `ReportTypeUseInternal` parameter should better be called `blocked_types`.
> being provided by a **template** defn

Just in case, for clarity: the current PR treats only types in _typedef and type alias_ definitions.

"Definition-side types" term is also good, my only concern is that it looks too long.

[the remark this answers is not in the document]
kimgr: Agree that it's too long, I just wanted to throw it out there in case it would inspire something better yet.

tests/cxx/badinc.cc
```diff
@@ -213,7 +211,10 @@ Cc_tpl_typedef cc_tpl_typedef;
 // IWYU: I2_Class::InlFileTemplateFn is...*badinc-i2-inl.h
 // IWYU: I2_Class::InlFileStaticFn is...*badinc-i2-inl.h
 typedef I2_Class Cc_I2_Class_Typedef;
-// IWYU: I1_Struct needs a declaration
+// I1_Struct isn't really used by any possible operation with H_TemplateStruct,
```
kimgr, line 214: Sorry to come back to this at such a late stage, but I stumbled on the following bug report:
https://github.com/include-what-you-use/include-what-you-use/issues/1249#issuecomment-1541818734

Doesn't that sound kind of like the same behavior? In that case it's `shared_ptr<>` that triggers a requirement of its user-provided type for libstdc++ but not libc++. Probably their implementations are different in a way that triggers this. But it sounds like your change here would make that more the rule?

Would the workaround be to forward-declare the user type in the file instantiating `shared_ptr<>`?

kimgr:
@bolshakov-a Just to make sure we're not waiting for each other -- I was holding off merge until you'd renamed `ReportTypeUseInternal(..., provided_types)` to `blocked_types`, but maybe I misunderstood?

bolshakov-a:
Sorry, I wanted to clarify the naming issue, but in fact I have little problems with finding some spare time. If you agree with `blocked_type` term, I can try to complete the work on the nearest weekend.

Regarding the libc++ issue (#1249), it is probably due to the lack of alias template complete support in IWYU, but I haven't yet finished debugging.

kimgr:
No rush, I just wanted to make sure you weren't waiting for me 🙂

bolshakov-a:
However, indeed, renaming of a single parameter shouldn't take a lot of time. I can do it right now.

kimgr:
Thanks! Let get this merged.

bolshakov-a:
Thank you!

tests/cxx/badinc.cc
bolshakov-a: No, it is slightly different problem. `std::shared_ptr<T>::element_type` is defined through `std::remove_extent<T>::type` alias in libstdc++ and through `std::remove_extent_t<T>` alias template in libc++, and IWYU doesn't fully support alias templates at now. I performed [some work](https://github.com/bolshakov-a/include-what-you-use/commit/5e896ac6912bb735c43c10d1a202f81bc965605d) in that direction already, hoping it will arrive the main repository some day.

PR #1260, opened 2023-05-19
fixing bug where you can't use --error argument as it's ambigious with --error_always

RisingLiberty:


kimgr:
Hey @RisingLiberty, can you expand a little on what the problem is?

`iwyu_getopt` is only used for MSVC builds, otherwise we use the system `getopt.h`. Also, it is vendored from https://github.com/kimgr/getopt_port, so there's a chance the bug should be fixed there instead.

RisingLiberty:
hey, the problem is that when using -error (return an error code when any issues where spotted by iwyu). the parsing of "-error" would match both "-error" and "-error_always" meaning you can't use it
I just added a CHECK for future proofing, if we don't want that, the include and the CHECK itself can be removed to just have the bug fix

RisingLiberty:
yep

kimgr:
I'm working on fixing this in the upstream `getopt_port` project. But why are you passing both `--error` and `--error_always`? I don't see a usecase for that.

kimgr:
Superseded by https://github.com/include-what-you-use/include-what-you-use/pull/1268, but thanks for finding this bug; I had no idea!

PR #1266, opened 2023-06-24
Improve placement-new detection in templates

### LINKED ISSUE #1229: [Regression 0.18 -> 0.20] IWYU incorrectly wants to remove <new>
Hi!

We are looking into bumping IWYU from 0.18 to 0.20 (btw, amazing job with templates, lots of FPs fixed!) and encounter the following problem:

```cpp
// foo.h
#include <new>
#include <utility>

template <typename Ref, typename... Args>
void construct(Ref& ref, Args&&... args)
{
    new (std::addressof(ref)) Ref(std::forward<Args>(args)...);
}
```
```cpp
// foo.cpp
#include "foo.h"

int main()
{
  int x;
  construct(x, 123);
}
```

Running IWYU with default settings results in the following:
```
foo.h should add these lines:

foo.h should remove these lines:
- #include <new>  // lines 1-1

The full include-list for foo.h:
#include <utility>  // for addressof, forward
---
```

If I remove `new`, the code no longer compiles.

kimgr:
When placement-new is used in a template, and the buffer/target argument is provided by a call to a dependent name, the placement-new expression would not be recognized as such. This is commonplace now that std::addressof exists:

    template<class T, typename...Args>
    void construct(T& buf, Args&&... args) {
      // definitive placement-new below
      new(std::addressof(buf)) T(args)
    }

Prior to this fix, `<new>` would be suggested at the instantiation location rather than the definition location. If the template was defined in a file covered by IWYU analysis, IWYU would suggest to remove `<new>` there, breaking the template.

Add a few more testcases for placement syntax in templates.

Fixes #1229.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2417,17 +2417,23 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // Like in VisitOverloadExpr(), we update processed_overload_locs
     // regardless of the value of CanIgnoreCurrentASTNode().
 
-    // We say it's placement-new if the (lone) placment-arg is a
-    // pointer.  Unfortunately, often clang will just say it's a
-    // dependent type.  In that case, we can still say it's a pointer
-    // in the (common) case the placement arg looks like '&something'.
-    // (This is possibly wrong for classes that override operator&, but
-    // those classes deserve what they get.)
-    if (!expr->getOperatorNew() &&
-        expr->getNumPlacementArgs() == 1 &&
-        (GetTypeOf(expr->getPlacementArg(0))->isPointerType() ||
-         GetTypeOf(expr->getPlacementArg(0))->isArrayType() ||
-         IsAddressOf(expr->getPlacementArg(0)))) {
+    // We say it's placement-new if it has a single placement-arg, which _might_
+    // be a pointer. In templates clang will just say it's a dependent type (or
+    // dependent name, for function calls), so we use some heuristics to avoid
+    // matching things like 'new (std::nothrow)' and 'new
+    // (std::align_val_t(8))'. The pattern 'new (func())' can cause false
+    // positives if the eventually looked up 'func()' does not return a pointer
+    // type.
+    const Expr* placement_expr = nullptr;
+    if (!expr->getOperatorNew() && expr->getNumPlacementArgs() == 1) {
+      placement_expr = expr->getPlacementArg(0);
+    }
+
+    if (placement_expr &&
+        (GetTypeOf(placement_expr)->isPointerType() ||
+         GetTypeOf(placement_expr)->isArrayType() ||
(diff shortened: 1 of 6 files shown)


iwyu.cc
```diff
@@ -2423,11 +2423,18 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // in the (common) case the placement arg looks like '&something'.
     // (This is possibly wrong for classes that override operator&, but
     // those classes deserve what they get.)
-    if (!expr->getOperatorNew() &&
```
bolshakov-a, line 2426: Doesn't the comment above need updating?

tests/cxx/placement_new-d1.h
```diff
@@ -12,4 +12,7 @@
 
 #include "tests/cxx/placement_new-i1.h"
 
+// for std::addressof use in .cc
+#include <utility>
```
bolshakov-a, line 16: It is stated in the standard that `std::addressof` is in `<memory>`:
https://eel.is/c++draft/memory.syn
Hard-coded symbol mapping should be probably updated.

kimgr: Yeah, I saw that too. There's a header mapping from `bits/stl_move.h`, where `std::addressof` is defined, to `<utility>`, presumably for `std::move`. I'd rather not mess with this right now.

But maybe it would be nicer to get rid of the standard library dependency in the test; this function is easy enough to simulate.

iwyu.cc
kimgr: Hehe, I thought I did...? Are you missing something in particular? I'll look it over and try to edit it down a bit anyway, it's pretty wordy.

kimgr: Rewrote the comment, let me know what you think.

tests/cxx/placement_new-d1.h
kimgr: Added `placement_new-i2.h` with a custom `AddressOf` function.

iwyu.cc
bolshakov-a: It looks better, thanks! Btw, I'm not an expert in `operator new` overloading and cannot say whether the new logic is optimal tradeoff or not.

kimgr: I think it's good enough -- there is a loophole where:
```
template<class T>
T* construct(T& target) {
  return new(std::addressof(target)) T;
}
```
works as expected, but stuff like:
```
template<class T>
std_align_val_t& getAlign<T>() {
  return std_align_val_t(T::getPreferredAlignment());
}

template<class T>
T* construct() {
  return new(getAlign<T>()) T;
}
```
will falsely be marked as placement-new, needing `#include <new>` (on the other hand `std::align_val_t` is in `<new>`, so the result is still valid, just that diagnose the placement-new as also needing `<new>`).

PR #1269, opened 2023-06-26
Silently break cycles in mappings for v0.19

Jose5500:
Addresses the preprocessor metaprogramming techniques used in Boost, for v0.19.

Updated the transitive closure expansion to ignore cycles instead of failing on them. Added diagnostic logging at level 8 to look into its decision-making if necessary.

Identical implementation of - [link](https://github.com/include-what-you-use/include-what-you-use/pull/1152)
Fixes issue #424

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
 
     env:
-      LLVM_TAG:
+      LLVM_TAG: -15
 
     steps:
       - name: Install prerequisites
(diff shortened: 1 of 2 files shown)


kimgr:
@Jose5500 Thanks for the patch attempt! Unfortunately we can't take fixes for historical releases. They are what they are, and we don't have capacity to do service releases for old versions.

If you must keep on 0.19 but with the cycle break fix, I suggest you cherry-pick/patch and build locally, or try to petition your envinronment's package manager to take the patch.

Issue #1235, opened 2023-04-12
New IWYU pragma for extension point patterns


kadircet:
# IWYU always_keep pragma proposal

## Background

In clang-tooling space there is a new focus around getting more automated & safe include management, similar to IWYU with a new “use” policy focused around spelling of symbols (hence also called as include-what-you-spell, IWYS). See [clang-include-cleaner](https://github.com/llvm/llvm-project/tree/main/clang-tools-extra/include-cleaner/)

## Problem

It’s common in C++ to have extension points for types to enable their usage in certain generic algorithms, e.g. one common example could look like:

foo.h:
```cpp
struct Foo { … };
```
foo_ext.h:
```cpp
#include “foo.h”
bool isLess(Foo&, Foo&);
```
sort.h:
```cpp
template<typename Range> void sort(Range) { /* uses isLess(T&,T&) */ }
```
application.cc:
```cpp
#include “foo.h” // provides vector<Foo>& getStats();
#include “sort.h”
#include “foo_ext.h” // unused, there’s no "direct" usage of isLess(Foo&,Foo&)

void app() {
  auto &f = getStats(); // returns a vector<Foo>&.
  sort(f);
}
```
Here, removing `foo_ext.h` will prevent the code from compiling, so we need to avoid suggesting that. However no symbol from the file is "used" in our usual sense.

### Analyze template instantiations and mark the usage of such extension points

This is close to the current IWYU tool and powerful, but doesn’t align with IWYS’s definition of “use”. It complicates the implementation drastically and is impossible in certain tools (e.g. clangd, which doesn’t keep template bodies from headers around). Moreover, deducing intent fully is still error-prone. Most of these arguments and complexity is also mentioned in [https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYUIsDifficult.md#who-is-responsible-for-dependent-template-types](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYUIsDifficult.md#who-is-responsible-for-dependent-template-types). So having all this complexity and still getting certain cases wrong goes against our automation goals.

### Restrict such patterns

We can call such patterns IWYU violations (and they are), and forbid people from using them. But this is really hard to enforce in practice. Most of the common alternatives require:

* extensions to go with type, which is not always possible (e.g. type might be generated, provider of the extension point and type might be completely different sets of parties, etc.)
* introduce these symbols at linking time via registries or forward declaring them.

[...]

PR #1272, opened 2023-07-06
Add new IWYU pragma: always_keep

kimgr:
This was proposed by the clangd/clang-include-fixer developers, who
already implement many of IWYU's pragma comments.

The pragma makes sense for us too, so implement it and add documentation
to describe the semantics.

Semantics and rationale suggested by @kadircet, as well as a
documentation draft from which I've borrowed some words.

Fixes issue #1235.


kimgr:
@kadircet Please take a look and see if this meets your expectations. Thanks!

kadircet:
thanks, that LGTM!

Issue #1278, opened 2023-08-02
`struct timeval` mapped to `sys/time.h`


firewave:
```cpp
#include <sys/select.h>

void f()
{
    struct timeval tv;
    tv.tv_sec = 1;
    tv.tv_usec = 0;
    (void)tv;
}
```

The code compiles but the documentation attributes it to `sys/time.h`.

https://pubs.opengroup.org/onlinepubs/7908799/xsh/systime.h.html
https://man7.org/linux/man-pages/man2/settimeofday.2.html

If you provide both includes it correctly prefers the expected one:

```
/mnt/s/___temp/input.cpp should add these lines:

/mnt/s/___temp/input.cpp should remove these lines:
- #include <sys/select.h>  // lines 2-2

The full include-list for /mnt/s/___temp/input.cpp:
#include <sys/time.h>  // for timeval
---
```

kimgr-spire:
Yes, timeval is provided by both. See https://github.com/include-what-you-use/include-what-you-use/pull/930 for the patch that made it so.

Is this causing any problems in particular?

firewave:
> Yes, timeval is provided by both. See #930 for the patch that made it so.

Those changes seem (at least) partially kind of arbitrary. It somehow seems like introducing mappings to common files which also have it included (assuming to reduce the impact on existing code bases). I do get why this was done but in this case it is clearly documented which header provides it. This is also basically re-introducing reliance on transitive includes.

Also this type in particular is not even listed in the manpage which that other PR refers.

> Is this causing any problems in particular?

Not for me. I just came across this while filing https://github.com/llvm/llvm-project/issues/64336.

kimgr:
Maybe @alejandro-colomar can fill in on the status of the posix/Linux documentation.

In general iwyu has had mappings that limit unnecessary changes around sys includes. They're usually stable enough that it's OK to allow a little imprecision in the name of convenience.

alejandro-colomar:
Hi @kimgr !

I had a similar feeling to whar @firewave is suggesting.  Maybe I shouldn't have added back then so many mappings, and only keep the fundamental header for each type.  That way, we would be strict and force users to include what they use from libc headers, even if the standards guarantee more flexibility.

Back then I didn't do that to not produce a lot of churn reverting part of what I did, but @firewave 's complain gives more strength to those feelings I had.

I'll prepare a patch CCing both of you, and we can discuss it further there, if you want.

Cheers,
Alex

[...]

PR #1283, opened 2023-08-12
Reduce libc mappings to really include what you use

alejandro-colomar:
```
The standards (ISO C and POSIX.1) guarantee that many definitions are
provided by more than one header.  Programmers may #include <stdio.h>
for the definition of NULL, and be covered by the standards.  However,
that's not following the include-what-you-use principle, and we should
not promote that.

Cut the mappings to the minimum: the principal header that should define
a type or macro.  In most cases, POSIX.1 and ISO C agree on that header,
and we use that one.  In a few cases, ISO C prescribes <time.h> and
POSIX.1 prescribes <sys/types.h>; keep both, but give preference to the
one by POSIX.1, <sys/types.h>, which is smaller.

When the type or macro is not standardized, if there's a clearly
canonical header for it, keep that one; if there's not, let's keep all
of them.

Regarding C++'s <c...> headers and NULL, C++ uses nullptr.  C++ code
that uses NULL is probably C-compat code, and that code should be
including the C headers.

Also remove or update comments that are not up to date anymore.
```

Closes: <https://github.com/include-what-you-use/include-what-you-use/issues/1278>
Reported-by: Oliver Stöneberg <https://github.com/firewave>
Cc: Kim Gräsman <kim.grasman@gmail.com>


kimgr:
This seems like quite a dramatic and not necessarily helpful change. Principles are great, but I'm not sure we get any benefits from more fine-grained IWYU for system headers... Am I missing something?

alejandro-colomar:
Well, one could imagine a program doing something like:

```c
#include <monetary.h>

void
foo(size_t n, int arr[n])
{
    for (size_t i = 0; i < n; i++) {
        arr[i] = 42;
    }
}
```

I'd want iwyu(1) to complain about such aberration.  It's not so crazy to have such a program.  Let's say you had something from `<monetary.h>` and then removed that code, but forgot to remove the include.  iwyu(1) may keep it just for `size_t`, which would be bad.

kimgr:
Ah, I see now the `<sys/select.h>`/`<sys/time.h>` issue is also an example of using a superset header for a type that is more correctly/narrowly defined elsewhere. I had overlooked that detail when I returned to this issue.

I'm just worried this change will generate _a lot_ of seemingly-spurious suggestions that will only serve to annoy users.

The `offsetof` patch looks unrelated, let's do that in a separate PR, please?

alejandro-colomar:
@kimgr In the end, I decided to do the cut for part 1 at a somewhat logical point:

-  Mappings guaranteed by ISO C are respected.
-  Mappings for `<sys/types.h>` and `<unistd.h>` (POSIX.1) are respected.
-  Any other POSIX.1 mappings have been removed.

It seems this doesn't affect out test suite, so it seems to be a fair cut point, isn't it?

Please ping me after the next release for part 2.  :)

kimgr:
Thanks! Let's give this a shot. I created an issue to track part 2, whatever that might turn out to be :)

gcc.symbols.imp
```diff
@@ -75,154 +58,69 @@
...
-  { symbol: [ "suseconds_t", private, "<sys/time.h>", public ] },
-  { symbol: [ "time_t", private, "<time.h>", public ] },
-  { symbol: [ "time_t", private, "<sched.h>", public ] },
-  { symbol: [ "time_t", private, "<sys/msg.h>", public ] },
-  { symbol: [ "time_t", private, "<sys/select.h>", public ] },
-  { symbol: [ "time_t", private, "<sys/sem.h>", public ] },
-  { symbol: [ "time_t", private, "<sys/shm.h>", public ] },
-  { symbol: [ "time_t", private, "<sys/stat.h>", public ] },
-  { symbol: [ "time_t", private, "<sys/time.h>", public ] },
   { symbol: [ "time_t", private, "<sys/types.h>", public ] },
```
bolshakov-a, line 95: I'm worrying that IWYU may start to suggest POSIX headers on non-POSIX-compliant systems. Specifying `--no_default_mappings` could be impractical when e.g. some project uses both Windows CRT as the C library and libc++ as the C++ library.

alejandro-colomar: I opened a PR for fixing that.

<https://github.com/include-what-you-use/include-what-you-use/pull/1288>

PR #1287, opened 2023-08-12
Extend comment for GetComponentsOfType after #1182

bolshakov-a:
Extracted from #1248.


kimgr:
Thanks! I think you can skip the reference to #1182 in the commit subject, that's mostly a distraction. But otherwise lgtm, obviously.

PR #1288, opened 2023-08-12
Give preference to ISO C mappings

alejandro-colomar:
Otherwise, iwyu(1) might recommend a POSIX header in a non-POSIX system, which shouldn't happen.

While at it, remove a mapping for clockid_t that should have been removed in b9fde49306fa ("Reduce libc mappings to really include what you use (part 1)"), as it's not a type defined by ISO C.

Fixes: b9fde49306fa ("Reduce libc mappings to really include what you use (part 1)")
Reported-by: Andrey Ali Khan Bolshakov <https://github.com/bolshakov-a>


bolshakov-a:
Thanks!

kimgr:
Makes sense, thanks for noticing that!

Issue #1279, opened 2023-08-02
IWYU fails with "expected exactly one compiler job" error with multiple -arch flags


scztt:
When running iwyu with multiple -arch flags, the tools fails with an error:

```
> include-what-you-use test.cpp -arch arm64 -arch x86_64

[...]

PR #1295, opened 2023-08-17
Better support multi-arch command-lines

kimgr:
Compilation databases sometimes contain clang commands with multiple -arch switches; they would be rejected with the dreaded:

    expected exactly one compiler job

error.

Change the policy to pick the first available compile job from compilation and use the arguments from there.

This will select an arbitrary -arch, in the case of multi-arch builds, but it will at least allow IWYU to run.

Users interested in e.g. selecting a particular arch can edit their compile commands into separate sets.

Fix for #1279.


kimgr:
Closed in favor of #1368

PR #1298, opened 2023-09-01
Avoid redundant namespace reporting

kilroyd:
When referencing a symbol, `ns::sym`, the declaration of `sym` will also have a sufficient declaration for the associated namespace `ns`.

A using directive or namespace alias wants a declaration for the namespace. If there are uses of symbols in that namespace, then rely on IWYU for those symbols to satisfy the namespace use.

If there is no symbol use from the namespace, then the first header where the namespace is seen will still be recommended (existing behavior). This is expected to be uncommon.

This ameliorates the impact of #828, but does not fix it.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1025,6 +1025,9 @@ set<string> CalculateMinimalIncludes(
 //     a.h does not see b.h in its transitive #includes.  (Note: This
 //     happens before include-picker mapping, so it's still possible to
 //     see 'new' includes via a manual mapping.)
+// B8) Discard uses of NamespaceDecls where the file also uses symbols
+//     within that namespace. IWYU for the symbol will also satisfy IWYU
+//     for the containing namespace(s).
 // B1') Discard macro uses in the same file as the definition (B3 redux).
 // B2') Discard macro uses that form a 'backwards' #include (B6 redux).
 // B3') Discard macro uses from a 'new' #include (B7 redux).
@@ -1212,8 +1215,8 @@ static bool HasMapping(const string& symbol) {
   return !GlobalIncludePicker().GetCandidateHeadersForSymbol(symbol).empty();
 }
 
-void ProcessFullUse(OneUse* use,
-                    const IwyuPreprocessorInfo* preprocessor_info) {
+void ProcessFullUse(OneUse* use, const IwyuPreprocessorInfo* preprocessor_info,
+                    const vector<OneUse>& all_uses) {
   CHECK_(use->decl() && "Must call ProcessFullUse on a decl");
   CHECK_(use->is_full_use() && "Must not call ProcessFullUse on fwd-decl");
   if (use->ignore_use())   // we're already ignoring it
@@ -1396,6 +1399,20 @@ void ProcessFullUse(OneUse* use,
       VERRS(6) << "Ignoring use of " << use->symbol_name()
                << " (" << use->PrintableUseLoc() << "):"
                << " non-transitive #include\n";
+      use->set_ignore_use();
+      return;
+    }
+  }
+
+  // (B8) Discard uses of NamespaceDecls where the file also uses
+  // symbols within that namespace. IWYU for the symbol will also
+  // satisfy IWYU for the containing namespace(s).
+  if (const auto* ns_decl = dyn_cast<NamespaceDecl>(use->decl())) {
+    if (std::any_of(
+            all_uses.begin(), all_uses.end(), [ns_decl](const OneUse& use) {
+              const NamedDecl* decl = use.decl();
+              return decl && (ns_decl != decl) && IsInNamespace(decl, ns_decl);
+            })) {
       use->set_ignore_use();
       return;
     }
@@ -1641,7 +1658,7 @@ void IwyuFileInfo::CalculateIwyuViolations(vector<OneUse>* uses) {
   }
   for (OneUse& use : *uses) {
     if (use.is_full_use() && use.decl())
-      internal::ProcessFullUse(&use, preprocessor_info_);
+      internal::ProcessFullUse(&use, preprocessor_info_, *uses);
   }
   for (OneUse& use : *uses) {
     if (use.is_full_use() && !use.decl())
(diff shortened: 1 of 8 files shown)


iwyu_output.cc
```diff
@@ -1400,6 +1403,20 @@ void ProcessFullUse(OneUse* use,
...
   }
+
+  // (B8) Discard uses of NamespaceDecls where the file also uses
+  // symbols within that namespace. IWYU for the symbol will also
+  // satisfy IWYU for the containing namespace(s).
+  if (const auto* ns_decl = dyn_cast<NamespaceDecl>(use->decl())) {
+    if (std::any_of(all_uses.begin(), all_uses.end(),
+                    [ns_decl](const OneUse& use) {
+                      const NamedDecl* decl = use.decl();
+                      return decl && IsInNamespace(decl, ns_decl);
```
kimgr, line 1414: Feels like we should also make sure `decl != ns_decl` here...? I think it currently falls through a bit by accident, because `IsInNamespace` will start searching from the nearest declcontext of the passed-in decl, so it misses itself. Or maybe that's intentional and just subtle?

I'm thinking that it might be nice to make it more explicit.

tests/cxx/namespace_use-d1.h
```diff
@@ -0,0 +1,12 @@
...
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+namespace i1_ns {
+void action();
```
kimgr, line 11: Let's use Google style names for test symbols too. Functions are PascalCase, so this should be `Action`.

tests/cxx/namespace_use-d2.h
```diff
@@ -0,0 +1,12 @@
...
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+namespace i2_ns {
+int get_value();
```
kimgr, line 11: `GetValue`

tests/cxx/namespace_use-i1.h
```diff
@@ -0,0 +1,21 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#include "namespace_use-d1.h"
+#include "namespace_use-d2.h"
+
+namespace i1_ns {
+
+// Declare something extra in i1_ns to see if d1 or i1 is suggested
+enum Xtypes {
```
kimgr, line 16: This also looks a little off, let's go with just `X`.

tests/cxx/namespace_use.cc
```diff
@@ -0,0 +1,45 @@
...
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+// Tests that using-directives i.e. `using namespace`, get appropriate
+// include suggestions.
+
+#include "tests/cxx/namespace_use-i1.h"
```
kimgr, line 15: OK, so I think you've misunderstood the `-d` and `-i` convention (it's weird, so I don't blame you :)) --

* `*-dX.h` or `*-direct.h` indicate a _direct include_, a header the main file includes directly
* `*-iX.h` or `*-indirect.h` indicate an _indirect include_, a header that is indirectly included by a direct include

So the confusing thing is that direct includes are just necessary padding, they usually shouldn't contain anything -- their typical role is to be suggested for removal.

The indirect includes, on the other hand, are where the interesting symbols should go.

As far as I can tell, you've interpreted it exactly the other way around, so you should be able to "just" switch everything around.

tests/cxx/namespace_use.cc
```diff
@@ -0,0 +1,45 @@
...
+
+// Purely using the namespace, not any symbols.
+// This could be unused, or being used transitively in namespace lookup.
+// The include suggestion will be the first header that declares the namespace.
+// IWYU: i1_ns is defined in ...* which isn't directly #included.
```
kimgr, line 20: You don't need the ` which isn't directly #included.` tail, the `...*` pattern matches all the way to the end of the line.

Nice touch omitting the filename there, that's a nice way to check that there is a diagnostic even if we can't know for sure for which file.

tests/cxx/namespace_use.cc
```diff
@@ -0,0 +1,45 @@
...
+using namespace i1_ns;
+
+// This namespace gets used for symbols within it.
+// Suggestions for includes will come via the symbol usage rather than here.
+using namespace i2_ns;
+
+int main(int, const char**) {
+  // IWYU: i2_ns::get_value is defined in ...* which isn't directly #included.
```
kimgr, line 28: Here I think you _do_ know which filename, though? Such an assertion is usually spelled like this: https://github.com/include-what-you-use/include-what-you-use/blob/9e77b739be7014de0f8d17f86669cf8331e86337/tests/cxx/mapping_regex.cc#L27

tests/cxx/namespace_use.cc
```diff
@@ -0,0 +1,45 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+// Tests that using-directives i.e. `using namespace`, get appropriate
+// include suggestions.
```
kimgr, line 13: It looks like this can be rewrapped to fit `include` on the first line.

Also I tend to avoid backticks in code, but maybe only because it feels like an exotic character on my Swedish keyboard :)

kimgr (commented):
Thanks! Comments inline.

iwyu_output.cc
kilroyd: I think that's intentional - I'd expect `IsInNamespace(ns_decl, ns_decl)` to return false. Happy to put in the explicit check though.

tests/cxx/namespace_use.cc
kilroyd: Whoops - I didn't think too hard when naming these, I just remembered there were d's and i's. Fixing.

kimgr: Thanks, makes more sense now!

iwyu_output.cc
kimgr: I think it might even make sense as an optimization to `IsInNamespace`, but never mind, I saw you added an explicit check. I feel safe now.

kimgr:
Thank you, looks great!

PR #1299, opened 2023-09-04
Report overloaded function return type

bolshakov-a:
Function return type should be analyzed regardless from whether the overload set was reported or not, because overload set processing doesn't deal with return types. It is reasonable because different functions in an overload set may have different return types.


kimgr:
Beautiful, thanks!

PR #1300, opened 2023-09-04
Avoid redundant reports on function declaration references

bolshakov-a:
TraverseDeclRefExpr should work only with implicitly instantiated function template specialization references. Other types of function references don't require full return type info because they don't cause template instantiation.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -178,6 +178,7 @@ using clang::FriendTemplateDecl;
 using clang::FunctionDecl;
 using clang::FunctionProtoType;
 using clang::FunctionTemplateDecl;
+using clang::FunctionTemplateSpecializationInfo;
 using clang::FunctionType;
 using clang::LValueReferenceType;
 using clang::LinkageSpecDecl;
@@ -794,6 +795,13 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
       return true;
 
     if (FunctionDecl* fn_decl = DynCastFrom(expr->getDecl())) {
+      const FunctionTemplateSpecializationInfo* tpl_spec_info =
+          fn_decl->getTemplateSpecializationInfo();
+      const bool is_def = fn_decl->isThisDeclarationADefinition();
+      if (!tpl_spec_info ||  // I. e. not a template specialization, or...
+          tpl_spec_info->isExplicitInstantiationOrSpecialization() || !is_def) {
+        return true;
+      }
       // If fn_decl has a class-name before it -- 'MyClass::method' --
       // it's a method pointer.
       const Type* parent_type = nullptr;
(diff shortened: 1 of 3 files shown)


tests/cxx/funcptrs.cc
```diff
@@ -79,6 +80,14 @@ void FreeFunctions() {
   // IWYU: Retval is...*funcptrs-i1.h
   // IWYU: FunctionTemplate is...*funcptrs-i1.h
   &FunctionTemplate<Retval>;
+
+  // No need for full return type info for non-templated function...
+  // IWYU: FunctionReturningRecordType is...*funcptrs-i1.h
+  &FunctionReturningRecordType;
+  // ... and for function template specialization references.
```
kimgr, line 87: "or" instead of "and"?

iwyu.cc
```diff
@@ -794,6 +794,11 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
       return true;
 
     if (FunctionDecl* fn_decl = DynCastFrom(expr->getDecl())) {
+      if (!IsTemplatizedFunctionDecl(fn_decl) ||
```
kimgr, line 797: Ugh, the formatting here is pretty terrible... Could you trick it by introducing a temporary?

tests/cxx/funcptrs-i1.h
```diff
@@ -44,11 +44,21 @@ Enum Function(Class*) {
...
+
 template<typename T>
 int FunctionTemplate(Class*) {
   return T(10).Value();
 }
 
+template<typename T>
+T FunctionTemplate2() {}
+
+extern template IndirectClass FunctionTemplate2<IndirectClass>();
```
kimgr, line 60: Feels like it would be a stronger testcase if the template argument was different from the return type, e.g.
```
// here
template<typename T>
Retval FunctionTemplate2() {}

extern template Retval FunctionTemplate2<IndirectClass>();

// there
  // IWYU: FunctionTemplate2 is...*funcptrs-i1.h
  &FunctionTemplate2<IndirectClass>;
```

though I suspect that's not fully representative...? Might need something else than `Retval` that's forward-declarable for author intent, but even then I'm not sure if the template arg is somehow special?

kimgr (commented):
Looks good in general, some thoughts inline.

tests/cxx/funcptrs.cc
bolshakov-a: "Neither ... nor ..." is more clear to me.

tests/cxx/funcptrs-i1.h
bolshakov-a: You're right, the template argument is unrelated to the problem. Besides, I've found out that there is one more case when return type full info is not needed: for references to non-explicit specializations when the primary template has no definition in the translation unit. I've added some code and a test case.

tests/cxx/funcptrs.cc
kimgr: "Neither ... nand ..." maybe :)

kimgr:
Thanks, LGTM!

PR #1303, opened 2023-09-09
Require includes for namespace alias use

kilroyd:
Identify that the namespace alias is used from VisitNestedNameSpecifier().

This was part of the original submission for PR #1213.

### DIFF
--- a/tests/cxx/namespace_alias-d1.h
+++ b/tests/cxx/namespace_alias-d1.h
@@ -1,0 +1,19 @@
+//===--- namespace_alias-d1.h - test input file for IWYU ------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_NAMESPACE_ALIAS_D1_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_NAMESPACE_ALIAS_D1_H_
+
+// Nothing defined here, just including namespace_alias-i2.h and
+// namespace_alias-i3.h - IWYU should recommend dropping this header
+
+#include "tests/cxx/namespace_alias-i2.h"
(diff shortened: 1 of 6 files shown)


tests/cxx/namespace_alias.cc
```diff
@@ -0,0 +1,42 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+// Check that using a namespace alias requires an appropriate include
+
+#include "tests/cxx/namespace_alias-i1.h"
```
kimgr, line 14: Looks like the `-i*.h`/`-d*.h` naming is inside out here too.

tests/cxx/namespace_alias-d1.h
```diff
@@ -0,0 +1,20 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_NAMESPACE_ALIAS_D1_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_NAMESPACE_ALIAS_D1_H_
+
+// Just declare something in a namespace. Nested for good measure.
+namespace i1_ns1 {
```
kimgr, line 14: I actually forgot to ask for this for the prior PR, but I'm trying to move tests away from the filename-prefixed naming of symbols. I wrote up some test guidelines for this a while back: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#symbol-names

Could you rename the namespaces to reflect their role, e.g. `ns_top`, `ns_nested`, `ns_alias` or something?

I think that will both make the test clearer and make it easier to rename symbols/files.

kimgr: Or maybe even just `ns1` and `ns2` for the nested namespaces, actually.

kimgr:
Some test naming concerns, but otherwise looks great. Thanks for picking this up again!

kimgr:
Thanks, looks good!

PR #1304, opened 2023-09-10
Handle function return type components separately

bolshakov-a:
Separate determination of "provision" status for function return type components. Similar to #1248.

`blocked_types_` member variable is moved from `InstantiatedTemplateVisitor` into `IwyuBaseAstVisitor`. I think it is correct not to place it into `VisitorState` because `InstantiatedTemplateVisitor` has its own set of blocked types which is formed on launching instantiated type/function scan. Otherwise, the original blocked type set should be restored at the end of that scan.

A check for blocked type is performed now before adding to the caches in `InstantiatedTemplateVisitor`. It was not needed earlier because `blocked_types_` weren't changed during the scan, hence, despite such types might occur in the cache, they were filtered out inside `ReportTypeUseInternal`.

No explicit return type reporting for instantiated function templates because `InstantiatedTemplateVisitor` already does that taking into account template argument underlying type provision status.


iwyu.cc
```diff
@@ -1445,9 +1439,18 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return underlying_type;
   }
 
+  set<const Type*> GetProvidedTypes(const Type* type,
```
kimgr, line 1442: This refactor looks nice on its own, could you pull it out into a separate PR just to reduce the scope of this one? Thanks.

iwyu_ast_util.cc
```diff
@@ -1183,6 +1185,17 @@ bool IsBuiltinFunction(const clang::NamedDecl* decl) {
...
 }
 
+bool IsImplicitlyInstantiated(const clang::FunctionDecl* decl) {
+  const FunctionTemplateSpecializationInfo* tpl_spec_info =
+      decl->getTemplateSpecializationInfo();
+  if (!tpl_spec_info)
+    return false;  // Not a template specialization.
+  const bool is_implicit = tpl_spec_info->getTemplateSpecializationKind() ==
+                           clang::TSK_ImplicitInstantiation;
+  const bool is_def = decl->isThisDeclarationADefinition();
```
kimgr, line 1195: You could make this a guard clause up-front now, right?
```
if (!decl->isThisDeclarationADefinition())
  return false;
```
Not sure which is cheaper to test of `decl->getTemplateSpecializationInfo() == null` or `isThisDeclarationADefinition`, but you should be able to do them in that order.

tests/cxx/quoted_includes_first.h
```diff
@@ -18,16 +18,14 @@ std::unique_ptr<IndirectSubDirClass> CreateIndirectSubDirClass();
 /**** IWYU_SUMMARY
 
 tests/cxx/quoted_includes_first.h should add these lines:
```
kimgr, line 20: It's kind of annoying that this testcase has so much structure seeing how its only purpose is to check that includes are emitted in the right order.

I'll give it a think and see if I can reduce it to something even simpler so it's not affected by this change (e.g. a `unique_ptr<int>` to force use of `<memory>`).

Also, I'm a little bit surprised that `std::unique_ptr<IndirectSubDirClass> CreateIndirectSubDirClass();` in a header now requires the full type of `IndirectSubDirClass`... That seems like something that might upset users who want to optimize by leaving the responsibility of the pointer type to callers. But they can just add an explicit `class IndirectSubDirClass;` in the header, and it should be respected?

kimgr: With [this PR](https://github.com/include-what-you-use/include-what-you-use/pull/1305) I'm hoping you wouldn't have to touch `quoted_includes_first.h` for this patch?

bolshakov-a: Yes, it should. I just make the common policy for type components in all three cases of author-intent analysis (aliases, function return, "autocast"). Command line flags to suppress that policies can be later introduced.

Btw, in fact, returning `unique_ptr` requires pointed-to-type full info to be available at the call site, because it is needed for `~unique_ptr()`, in contrast to `shared_ptr`.

bolshakov-a: > With https://github.com/include-what-you-use/include-what-you-use/pull/1305 I'm hoping you wouldn't have to touch quoted_includes_first.h for this patch?

I think, yes. Thank you!

kimgr: > Btw, in fact, returning unique_ptr requires pointed-to-type full info to be available at the call site, because it is needed for ~unique_ptr(), in contrast to shared_ptr.

Yeah, my instinct was also that it would be nice if the defining header required it. I think for real code it can get annoying if you have wide interfaces where one function happens to return a type that brings in a lot of headers, and that function is rarely used. See also issue 1297 for a similar thing (didn't want to link it, because it doesn't really have anything to do with this PR)

bolshakov-a: There is currently a bug on the master branch. Consider:
```cpp
// --- GetClass.h

#include "Class.h"

#include <memory>

std::unique_ptr<Class> GetClass();


// --- User.cpp

#include "Class.h"
#include "GetClass.h"

void User() {
    GetClass();
}
```
IWYU suggests to remove `#include "Class.h"` from both `GetClass.h` and `User.cpp`. The code stops to compile after that, of course. This PR fixes the bug leaving `#include "Class.h"` in `GetClass.h`. Do you propose introducing of policy-suppression flags, which would push `#include` to the user sides from the decl sides, before merging this PR? Or, maybe, some different policy for return types?

iwyu_ast_util.cc
bolshakov-a: Maybe, just
```cpp
  return decl->isThisDeclarationADefinition() &&
         tpl_spec_info->getTemplateSpecializationKind() ==
             clang::TSK_ImplicitInstantiation;
```
?
Or the formatting is weird?

tests/cxx/quoted_includes_first.h
kimgr: No, I don't think we need to worry about it now, I was just thinking out loud. https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhatIsAUse.md#automatic-re-export-function-return-values clearly says headers declaring functions are responsible for their return type, unless they forward-declare to indicate user is responsible.

iwyu_ast_util.cc
kimgr: I think that looks pretty good, works for me!

iwyu_ast_util.cc
```diff
@@ -1183,6 +1185,17 @@ bool IsBuiltinFunction(const clang::NamedDecl* decl) {
   return false;
 }
 
+bool IsImplicitlyInstantiated(const clang::FunctionDecl* decl) {
```
kimgr, line 1188: Maybe `IsImplicitlyInstantiatedDefn`? Or maybe only definitions can be implicitly instantiated anyway?

bolshakov-a: No, clang marks as implicit instantiations also references to specializations of templates without definitions. Honestly, I'm still not sure whether it is valid C++ or not... But implementations support that.

iwyu.cc
```diff
@@ -1518,15 +1512,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-      const FunctionDecl* decl) const {
-    set<const Type*> retval;
-    const Type* return_type = Desugar(decl->getReturnType().getTypePtr());
-    if (CodeAuthorWantsJustAForwardDeclare(return_type, GetLocation(decl))) {
-      retval.insert(return_type);
-      // TODO(csilvers): include template type-args if appropriate.
-    }
-    return retval;
+  set<const Type*> GetProvidedTypesForFnReturn(const FunctionDecl* decl) const {
+    const Type* return_type = decl->getReturnType().getTypePtr();
```
kimgr, line 1516: This used to Desugar, but no longer does -- that's intentional to make sure we pick up as much sugar as possible, right?

iwyu.cc
```diff
@@ -2499,13 +2464,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    const char* comment = nullptr;
     // If we are forward-declarable, so are our template arguments.
     if (CanForwardDeclareType(current_ast_node()) &&
-        !IsProvidedTypeComponent(current_ast_node())) {
+        !IsProvidedTypeComponent(current_ast_node(), &comment)) {
       ReportDeclForwardDeclareUse(CurrentLoc(), decl);
       current_ast_node()->set_in_forward_declare_context(true);
     } else {
-      ReportDeclUse(CurrentLoc(), decl);
+      ReportDeclUse(CurrentLoc(), decl, comment);
```
kimgr, line 2474: This new API where the comment is returned in an out-param is pretty implicit.

* IsProvidedTypeComponent is called, and if it returns false...
* ... ReportDeclUse in the else leg uses the comment it returned

I wonder if that could be simplified. I'd even prefer to do the return-type detection again to the current state.

iwyu.cc
bolshakov-a: No, desugaring here is rather redundant now. `TypeEnumeratorWithoutSubstituted` collects canonical types. Maybe, avoiding of desugaring helps with template functions returning template parameter type, but I'm not sure.

iwyu.cc
bolshakov-a: It could probably be slightly better if `IsProvidedTypeComponent` would return `pair<bool, const char*>` or a `IsProvidedTypeComponentReturn` structure instead of out-parameter...

iwyu.cc
bolshakov-a: > Maybe, avoiding of desugaring helps with template functions returning template parameter type, but I'm not sure.

I forgot it:
> No explicit return type reporting for instantiated function templates
because `InstantiatedTemplateVisitor` already does that taking into
account template argument underlying type provision status.

iwyu.cc
bolshakov-a: Another variant can be to return
```cpp
enum ProvidedComponentType {
  None,
  ForTypeAlias,
  ForFnReturn,
//ForAutocast
}
```
and then analyze it in `GetTypeUseComment` function (also taking `CanForwardDeclareType` result for correct handling nested types).

kimgr: I think I prefer the `pair<bool, const char*>` retval, even if it gets a little inconvenient at the call sites.

iwyu.cc
kimgr: Makes sense, thanks!

iwyu.cc
bolshakov-a: Are you still against structured bindings?
```cpp
const auto [is_provided, comment] = IsProvidedTypeComponent();
```
looks better for me than `is_provided_comment_pair` or just `res`.

kimgr: Yeah, I agree. And it's nicer than `std::tie` because it doesn't need separate declarations.

Let's use structured bindings, I'll get used to it :)

iwyu.cc
```diff
@@ -2635,12 +2601,46 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      if (!IsInHeader(decl) || IsFriendDecl(decl))
+        return pair(false, nullptr);
+      const Type* return_type = decl->getReturnType().getTypePtr();
+      if (node->StackContainsContent(return_type)) {
+        const bool can_be_provided =
+            !decl->isThisDeclarationADefinition() &&
+            !IsPointerOrReferenceAsWritten(return_type);
+        return pair(can_be_provided,
+                    can_be_provided ? "(for fn return type)" : nullptr);
+      }
```
kimgr, line 2621: Isn't this just
```
if (!decl->isThisDeclarationADefinition() &&
    !IsPointerOrReferenceAsWritten(return_type)) {
  return pair(true, "(for fn return type)");
}
```
... and otherwise it falls through to the default return? Or does the line get too long so this formats better?

kimgr:
Looks good to me! Optional question/suggestion inline, but I'll merge either way tomorrow.

iwyu.cc
bolshakov-a: Correct, thanks!

kimgr:
Beautiful, thanks!

PR #1305, opened 2023-09-10
Simplify quoted_includes_first testcase

kimgr:
Instantiating `unique_ptr<T>` of a user-defined type T in a function return type is one of the more complex IWYU scenarios you can imagine and this testcase is really only intended to cover the sorting policy of include directives.

We still need both system angle-quoted headers and user quoted headers, so keep `unique_ptr<>`, but use a plain int instead. Add full uses of IndirectSubDirClass in both .h and .cc test files to see that includes are sorted properly.

### DIFF
--- a/tests/cxx/quoted_includes_first.cc
+++ b/tests/cxx/quoted_includes_first.cc
@@ -19,8 +19,12 @@
 #include "subdir/indirect_subdir.h"
 #include "quoted_includes_first.h"
 
-std::unique_ptr<IndirectSubDirClass> CreateIndirectSubDirClass() {
-  return std::unique_ptr<IndirectSubDirClass>(new IndirectSubDirClass);
+std::unique_ptr<int> CreateInt() {
+  return std::unique_ptr<int>(new int(20));
+}
+
+static IndirectSubDirClass GetIndirectSubDirClass() {
+  return global;
 }
 
 /**** IWYU_SUMMARY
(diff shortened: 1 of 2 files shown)


tests/cxx/quoted_includes_first.h
```diff
@@ -13,21 +13,21 @@
 #include <memory>
 #include "subdir/indirect_subdir.h"
 
-std::unique_ptr<IndirectSubDirClass> CreateIndirectSubDirClass();
+std::unique_ptr<int> CreateIndirectSubDirClass();
+
+IndirectSubDirClass global;
```
bolshakov-a, line 18: Nit: it should better be marked as `inline` to make the code closer to correct C++.

kimgr: Thanks, made it `inline const`, even, since the source-file function copies it.

tests/cxx/quoted_includes_first.cc
```diff
@@ -19,8 +19,12 @@
 #include "subdir/indirect_subdir.h"
 #include "quoted_includes_first.h"
 
-std::unique_ptr<IndirectSubDirClass> CreateIndirectSubDirClass() {
-  return std::unique_ptr<IndirectSubDirClass>(new IndirectSubDirClass);
+std::unique_ptr<int> CreateIndirectSubDirClass() {
```
kimgr, line 22: Oops, this name should also change

PR #1306, opened 2023-09-10
Extract 'GetProvidedTypes' function

bolshakov-a:
It can be reused for function return type or "autocast" analysis.


kimgr:
Ah, I misread the other PR, I thought the code was already duplicated across two functions.

Anyway, this looks good, thanks!

PR #1307, opened 2023-09-10
Further simplify and clarify quoted_includes_first

kimgr:
Represent the following types of headers:

* Precompiled header
* C system header
* C++ system header
* User header
* Associated header

and check that they are sorted correctly when IWYU diagnoses.

Uses headers selected because they're relative small and simple.

### DIFF
--- a/tests/cxx/quoted_includes_first.cc
+++ b/tests/cxx/quoted_includes_first.cc
@@ -11,35 +11,29 @@
 
 // Tests that IWYU will respect the --quoted_includes_first option.
 
-#include "tests/cxx/pch.h"  // this is the precompiled header
-#include <iostream>
-#include <list>
-#include <map>
-#include <memory>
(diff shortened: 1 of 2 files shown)


kimgr:
@bolshakov-a Maybe you want to take a look. I figured it was still unnecessarily complex for what it wants to check (and lacked C system header coverage). If you have a ideas for an even simpler C++ standard library header, I'd love to hear them.

bolshakov-a:
I suspect that `<initializer_list>` and `<exception>` should be the simplest ones because they don't depend on others. E.g., `<stack>` depends itself on `<initializer_list>`, `<compare>`, and `<deque>` at least. `<source_location>` is also a good one, but it is C++20.

kimgr:
@bolshakov-a Thanks -- `<exception>` is a good call. Updated.

PR #1309, opened 2023-10-03
Share public include mappings for C++ iostream

kilroyd:
Move the public->public include mappings specified for libstdcpp into a new array that we load whenever a C++ library is being used. This generally allows any of the iostream headers to stand in for each other.

This fixes the `cxx.test_no_char_traits` and `cxx.test_expl_inst_macro` tests on MacOS.

### DIFF
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -622,6 +622,31 @@ const char* stdlib_cpp_public_headers[] = {
   "<version>",
 };
 
+// Common public -> public include mappings for C++ standard library
+//
+// Note: make sure to sync this setting with stl.public.imp
+//
+const IncludeMapEntry stdlib_cpp_include_map[] = {
+  // The iostream .h files are confusing.  Lots of private headers,
+  // which are handled below, but we also have public headers
+  // #including each other (eg <iostream> #includes <istream>).  We
+  // are pretty forgiving: if a user specifies any public header, we
+  // generally don't require the others.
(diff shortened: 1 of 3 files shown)


kimgr:
Nice! Could you also split the corresponding part in https://github.com/include-what-you-use/include-what-you-use/blob/master/gcc.stl.headers.imp#L296 out into a separate file? We keep those "external" mappings around so people can run with `--no_mappings` and yet pick and choose which system mappings to include.

I feel like it should be called something like `stdlib.public.imp`.

iwyu_include_picker.cc
```diff
@@ -621,6 +621,28 @@ const char* stdlib_cpp_public_headers[] = {
   "<version>",
 };
 
+// Common public -> public include mappings for C++ standard library
```
kimgr, line 624: Once you've taken care of https://github.com/include-what-you-use/include-what-you-use/pull/1309#issuecomment-1753583267 -- please also add a note here to keep the external mapping in sync à la https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_include_picker.cc#L626.

And a corresponding note in the new .imp file.

kimgr:
Looks great! Just a few notes on keeping the external mapping file synced, otherwise ready to ship.

kilroyd:
> I feel like it should be called something like `stdlib.public.imp`.

I've used stl.public.imp, to align (a bit) with stl.c.headers.imp

kimgr:
Thanks!

Issue #1201, opened 2023-03-02
Assembly test will fail on architectures other than x86_64


bkmgit:
The test test_ms_inline_asm will fail on architectures such as Aarch64.  It may be good to either run it if architecture is detected to be x86_64 or make it an optional test.

kimgr:
What does `include-what-you-use --print-targets` say on your system?

Usually `clang` (and thereby IWYU) is built with support for all architectures, mine says:
```
kimgr@lardo:~/code/iwyu/include-what-you-use$ include-what-you-use --print-targets

  Registered Targets:
    aarch64    - AArch64 (little endian)
    aarch64_32 - AArch64 (little endian ILP32)
    aarch64_be - AArch64 (big endian)
    amdgcn     - AMD GCN GPUs
    arm        - ARM
    arm64      - ARM64 (little endian)
    arm64_32   - ARM64 (little endian ILP32)
    armeb      - ARM (big endian)
    avr        - Atmel AVR Microcontroller
    bpf        - BPF (host endian)
    bpfeb      - BPF (big endian)
    bpfel      - BPF (little endian)
    hexagon    - Hexagon
    lanai      - Lanai
    m68k       - Motorola 68000 family
    mips       - MIPS (32-bit big endian)
    mips64     - MIPS (64-bit big endian)
    mips64el   - MIPS (64-bit little endian)
    mipsel     - MIPS (32-bit little endian)
    msp430     - MSP430 [experimental]
    nvptx      - NVIDIA PTX 32-bit
    nvptx64    - NVIDIA PTX 64-bit
    ppc32      - PowerPC 32
    ppc32le    - PowerPC 32 LE
    ppc64      - PowerPC 64
    ppc64le    - PowerPC 64 LE
    r600       - AMD GPUs HD2XXX-HD6XXX
    riscv32    - 32-bit RISC-V
    riscv64    - 64-bit RISC-V
    sparc      - Sparc
    sparcel    - Sparc LE
    sparcv9    - Sparc V9
    systemz    - SystemZ
    thumb      - Thumb
    thumbeb    - Thumb (big endian)
    ve         - VE
    wasm32     - WebAssembly 32-bit
    wasm64     - WebAssembly 64-bit
    x86        - 32-bit X86: Pentium-Pro and above
    x86-64     - 64-bit X86: EM64T and AMD64
    xcore      - XCore
```
So it may be possible to just specify the architecture explicitly on the command-line in that test as a cheap first "solution".

[...]

PR #1312, opened 2023-10-14
Add infrastructure to allow skipping tests

kilroyd:
There may be reasons to skip a test. This infrastructure allows each test to define a condition under which it would skip. This is done by adding an '// IWYU_SKIP:' comment. The following text is evaluated as python code - if the result is true, the test is skipped.

I was looking at whether I could silence the `ms_inline_asm`, but saw the discussion on #1201. I'm not sure whether you'd prefer to set --target=x86_64, or mark the test for skipping with something like this. The infrastructure may come in handy to skip over tests that we can't get to work with libstdc++ or clang_mapping.

### DIFF
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -21,6 +21,7 @@ import re
 import shlex
 import subprocess
 import sys
+import unittest
 
 # These are the warning/error lines that iwyu.cc produces when --verbose >= 3
 _EXPECTED_DIAGNOSTICS_RE = re.compile(r'^\s*//\s*IWYU:\s*(.*)$')
@@ -44,6 +45,17 @@ _NODIFFS_RE = re.compile(r'^\((.*?) has correct #includes/fwd-decls\)$')
 # source file. Example:
 # // IWYU_ARGS: -Xiwyu --mapping_file=... -I .
 _IWYU_TEST_RUN_ARGS_RE = re.compile(r'^//\sIWYU_ARGS:\s(.*)$')
+
+# This is a line that specifies prerequisites for a test (test will be skipped
+# if they don't all hold). IWYU_REQUIRES specifies a condition that must be true
+# for the test to execute. IWYU_UNSUPPORTED specifies a condition in which the
+# test is invalid, and would be skipped. Multiple IWYU_REQUIRES and
+# IWYU_UNSUPPORTED can be specified.
+#
+# // IWYU_REQUIRES: feature(arg)
+# // IWYU_UNSUPPORTED: feature(arg)
+_IWYU_TEST_PREREQ_RE = re.compile(r'^// IWYU_(REQUIRES|UNSUPPORTED): '
+                                  r'([a-z][a-z0-9_-]*)\(([^)]+)\)$')
 
 
 def _Which(program, paths):
@@ -419,6 +431,68 @@ def _GetLaunchArguments(cc_file):
   return shlex.split(args)
 
 
+class Features:
+  """ Implements feature support functions for skip conditions """
+  def IwyuHasTargetSupportFor(target):
+    iwyu = _GetIwyuPath()
+    result = subprocess.run([iwyu, '-print-targets'],
+                            capture_output=True,
+                            text=True)
+    # TODO: parse output more rigorously
+    # TODO: maybe support a regular expression argument
+    return target in result.stdout
+
+
+# Maps a name to prerequisite feature functions.
+_KNOWN_PREREQ_FEATURES = {
+  "iwyu_has_target_support_for": Features.IwyuHasTargetSupportFor
+}
+
+
+def _ParsePrerequisites(cc_file):
+  """ Parse test prerequisites out of file. """
+  prerequisites = []
+  with open(cc_file) as it:
+    for lineno, line in enumerate(it):
+      m = _IWYU_TEST_PREREQ_RE.match(line)
+      if not m:
+          continue
+
+      directive = m.group(1)
+      feature = m.group(2)
+      arg = m.group(3)
+
+      feature_pred = _KNOWN_PREREQ_FEATURES.get(feature)
+      if not feature_pred:
+        raise SyntaxError("%s:%s syntax error in IWYU_%s, unsupported "
+                          "feature '%s'" % (ccfile, lineno, directive, feature))
+
+      if directive == "REQUIRES":
+        prerequisite = lambda arg=arg: feature_pred(arg)
+      elif directive == "UNSUPPORTED":
+        prerequisite = lambda arg=arg: not feature_pred(arg)
+      else:
+        assert False, ("_IWYU_TEST_PREREQ_RE should not have matched: %s" %
+                       directive)
+
+      desc = "%s: %s(%s)" % (directive, feature, arg)
+      prerequisites.append((prerequisite, desc))
+    return prerequisites
+
+
+def _CheckPrerequisites(cc_file):
+  """ Raises a SkipTest exception if any prerequisites in cc_file fail. """
+  prerequisites = _ParsePrerequisites(cc_file)
+
+  failed = []
+  for prerequisite, desc in prerequisites:
+    if not prerequisite():
+      failed.append(desc)
+
+  if failed:
+    raise unittest.SkipTest(", ".join(failed))
+
+
 def TestIwyuOnRelativeFile(cc_file, cpp_files_to_check, verbose=False):
   """Checks running IWYU on the given .cc file.
 
@@ -428,6 +502,9 @@ def TestIwyuOnRelativeFile(cc_file, cpp_files_to_check, verbose=False):
               to check the diagnostics on, relative to the current dir.
     verbose: Whether to display verbose output.
   """
+  # Parse and check IWYU_{REQUIRES,UNSUPPORTED}
+  _CheckPrerequisites(cc_file)
+
   cmd = [_GetIwyuPath()]
   # Require verbose level 3 so that we can verify the individual diagnostics.
   # We allow the level to be overriden by
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -42,6 +42,7 @@ def Partition(l, delimiter):
 
 def TestIwyuOnRelevantFiles(filename):
   logging.info('Testing iwyu on %s', filename)
+
   # Split full/path/to/foo.cc into full/path/to/foo and .cc.
   (all_but_extension, extension) = os.path.splitext(filename)
   (dirname, basename) = os.path.split(all_but_extension)
@@ -124,7 +125,16 @@ if __name__ == '__main__':
   (runner_args, _) = parser.parse_known_args(unittest_args)
 
   if runner_args.run_test_file:
-    exit(TestIwyuOnRelevantFiles(runner_args.run_test_file))
+    try:
+      r = TestIwyuOnRelevantFiles(runner_args.run_test_file)
+    except unittest.SkipTest as e:
+      # Catch the skip test exception that unittest would normally
+      # handle. Instead set a return code that informs Ctest that the
+      # test is being skipped
+      print(f"Skipped {runner_args.run_test_file}: {e}")
+      exit(99)
+    else:
+      exit(r)
 
   @GenerateTests(rootdir='tests/c', pattern='*.c')
   class c(unittest.TestCase): pass
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,6 +223,7 @@ if (Python3_Interpreter_FOUND)
       -- $<TARGET_FILE:include-what-you-use>
       WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     )
+    set_tests_properties(${name} PROPERTIES SKIP_RETURN_CODE 99)
   endfunction()
 
   execute_process(
--- a/docs/IWYUStyle.md
+++ b/docs/IWYUStyle.md
@@ -281,6 +281,23 @@ comment:
 The args are added to the `include-what-you-use` command as-written after line
 unwrapping, so IWYU flags must be explicitly prefixed with `-Xiwyu`.
 
+### Prerequisites ###
+
+If a test has prerequisites, annotate the `.cc` file itself using a
+directive comment:
+
+```
+// IWYU_REQUIRES: feature(value)
+// IWYU_UNSUPPORTED: feature(value)
+```
+
+`IWYU_REQUIRES` states that the following feature test must evaluate
+true for the test to be valid.
+
+`IWYU_UNSUPPORTED` states that the test is invalid when the feature
+test evaluates as false.
+
+If any prerequisite fails the test will be skipped.
 
 [1]: https://google.github.io/styleguide/cppguide.html
 [2]: https://llvm.org/docs/CodingStandards.html
--- a/tests/cxx/ms_inline_asm.cc
+++ b/tests/cxx/ms_inline_asm.cc
@@ -7,7 +7,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-// IWYU_ARGS: -fms-extensions
+// IWYU_ARGS: -fms-extensions --target=x86_64
+// IWYU_REQUIRES: iwyu_has_target_support_for(x86-64)
 
 // This file is not strictly an IWYU test, it just checks that the parser
 // doesn't choke on Microsoft inline assembly on any of our target platforms.


kimgr:
Thanks for taking a stab at this killer feature!

Unfortunately, I don't think we can take something based on `eval` in a public project, it's just too dangerous, for example, DO NOT try this on your local system:
```
// let's not think about this without /tmp prefix
// IWYU_SKIP_TEST: os.remove("/tmp/etc/passwd") == None
```
So I think we would need a more closed, configurable feature system, like `lit`: https://llvm.org/docs/TestingGuide.html#constraining-test-execution.

We can probably build out something simpler, but the general "tag-based" rather than "code-based" filtering feels a little less open-ended.

It's good to see this PR for end-to-end inspiration -- I think the general form is mostly nice and comfortable -- but it needs to be dumbed down a little :). I'll see if I can make time to hack on it if you're not interested. If you would like to explore alternatives, please open separate PRs so we can compare and contrast.

kilroyd:
I get the reluctance to use `eval()`. I used it here as a simple way to get the rest of the infrastructure up without having to invent a new syntax. My only push back is that we could also do bad things in `iwyu.cc`:

    system("rm -f /tmp/etc/passwd")

In either case I'd hope that system permissions stop anything catastrophic happening. The other difference is that we wildcard include the tests rather than explicitly name them in `CMakeLists.txt`, which make the `eval()` a bit more dangerous. That could be fixed if we wanted to.

Anyway, at the moment I can think of 5 things that we might want to test:
* Host OS (Linux/Windows/MacOS)
* Host arch (i386, x86_64, arm, arm64)? Can probably ignore.
* Compile target support (i.e. what has clang been compiled to support)
* C++ Standard library being used in compile
* (eventually) C standard library used in compile

I don't think we care about the target triple?

What about tags like:
* `system-windows`, `system-linux`, `system-darwin` (as LLVM is using)
* `target_supported={{regex for llvm target arch}}`
* `libc={{regex}}`
* `libcxx={{regex}}`

Do we want to follow LLVM with `UNSUPPORTED` and `XFAIL` rather than `IWYU_SKIP`? That wouldn't be too hard, though I'd do XFAIL in a separate PR.

kimgr:
I wonder if the host OS or CPU arch are actually necessary for IWYU? I think for Clang/LLVM these are used to test how the driver derives toolchain arguments from the environment, but I'm not sure.

It would be nice if we could interrogate clang (via `include-what-you-use`) for features, e.g. "do you support target triple/arch X?", "do you support standard library libc++?", etc. I think we want to be careful not to create too much divergence in the test suite based on the environment. Clang is a cross-compiler, so it's fine to parse/codegen Windows code on an Apple system (given Windows headers are available), for example. IWYU is only a front-end, so we should be able to be similarly flexible.

So it's nice if we can specify necessary explicit args in tests (e.g. `// IWYU_ARGS: -stdlib=libc++`), and only use skips to filter out these tests if libc++ is not installed (e.g. `IWYU_SKIP: !supports_stlib={{libc\+\+}}` or whatever the syntax might be).

Do you have any other use cases in mind besides `ms_inline_asm.cc`?

I don't think we should necessarily use the same `UNSUPPORTED` tag as lit, I prefer to have something `IWYU_`-prefixed to indicate that this is NOT lit.

XFAIL would be great, but I agree that it warrants its own PR.

kilroyd:
I've pushed a version that uses `IWYU_REQUIRES` and `IWYU_UNSUPPORTED`. I like how these avoid the need to support `not`. And I went with rather verbose (camel case) conditions, as I feel the verbosity reads better. Happy to rename as desired. For now I've included a condition in the `ms_inline_asm` test to see how it works.

kimgr:
Thanks! Please don't force-push overwrite existing versions, I think this is enough uncharted territory that I'm going to change my mind a few times along the way :)

For naming, I'd prefer `lowercase-with-dashes` to `camelCase`.

I was surprised to see you used `clang` to feature-test for stdlib, but I noticed that `include-what-you-use` does not support `-E -dM` for preprocessing. That's annoying. I wonder if we can somehow implement an introspection mode in IWYU (or maybe let it respect and forward flags to clang when it doesn't have anything to do).

iwyu_test_util.py
```diff
@@ -45,6 +45,23 @@
...
+# IWYU_UNSUPPORTED conditions must be false.
+#
+# See the function _EvalCondition for the supported condition syntax.
+#
+# // IWYU_REQUIRES: cond
+# // IWYU_UNSUPPORTED: cond
+_IWYU_TEST_SKIP_RE = re.compile(r'^//\sIWYU_(REQUIRES|UNSUPPORTED):\s(.*)$')
+
+# All conditions have the same format 'LABEL ARG'
+_IWYU_COND = re.compile(r'([A-z][A-z0-9_-]+)\s+([A-z][A-z0-9_-]+)')
```
kimgr, line 64: `_IWYU_SKIP_COND_RE`

kimgr, line 64: I wonder if it would be nicer to let them read as function calls, e.g. `LABEL(ARG)`? Maybe it just gets messy.

iwyu_test_util.py
```diff
@@ -419,6 +436,112 @@ def _GetLaunchArguments(cc_file):
   return shlex.split(args)
 
 
+def _CheckForCompileTimeMacro(macro):
+    """This assumes that the local clang++ and the include-what-you-use
```
kimgr, line 440: I don't think this is a valid assumption. I'd like for IWYU to build and the test suite to run even if the `clang` binary is not on the PATH. But I see why this is needed now, need to think up a way to do the same kind of feature testing on `include-what-you-use` itself.

iwyu_test_util.py
```diff
@@ -419,6 +436,112 @@ def _GetLaunchArguments(cc_file):
...
+    """
+    txt = '#include <cstddef>'
+
+    result = subprocess.run(['clang', '-dM', '-E', '-x', 'c++', '-'], capture_output=True,
+                            input=txt, text=True)
+
+    return " "+macro+" " in str(result.stdout)
+
+
+def _CheckExtraCommandLine(args):
```
kimgr, line 453: `IWYU_EXTRA_ARGS` only represents arguments that are passed by whoever runs the test suite. I don't think that's enough, in general, don't you also want to inspect `// IWYU_ARGS` and any builtin hard-coded args? I.e. we should probably extract the argument building from `TestIwyuOnRelativeFile` and evaluate against the full arglist.

iwyu_test_util.py
```diff
@@ -419,6 +436,112 @@ def _GetLaunchArguments(cc_file):
...
+  # Some helper functions to handle conditions
+  def isTestingLibCxx():
+    return _CheckExtraCommandLine('-stdlib=libc++') or _CheckForCompileTimeMacro('_LIBCPP_VERSION')
+  def isTestingLibstdcpp():
+    return _CheckExtraCommandLine('-stdlib=libstdc++') or _CheckForCompileTimeMacro('__GLIBCXX__')
+  def isTestingCanonicalClangMappings():
+    return _CheckExtraCommandLine('-Xiwyu --experimental=clang_mappings')
+  def iwyuHasTargetSupportFor(tgt):
+    iwyu = _GetIwyuPath()
+    result = subprocess.run([iwyu, '--print-targets'], capture_output=True)
```
kimgr, line 483: On my system, this prints:
```
kimgr@lardo:~/code/iwyu/include-what-you-use$ include-what-you-use -print-targets

  Registered Targets:
    aarch64     - AArch64 (little endian)
    aarch64_32  - AArch64 (little endian ILP32)
    aarch64_be  - AArch64 (big endian)
    amdgcn      - AMD GCN GPUs
    arm         - ARM
    arm64       - ARM64 (little endian)
    arm64_32    - ARM64 (little endian ILP32)
    armeb       - ARM (big endian)
    avr         - Atmel AVR Microcontroller
    bpf         - BPF (host endian)
    bpfeb       - BPF (big endian)
    bpfel       - BPF (little endian)
    hexagon     - Hexagon
    lanai       - Lanai
    loongarch32 - 32-bit LoongArch
    loongarch64 - 64-bit LoongArch
    m68k        - Motorola 68000 family
    mips        - MIPS (32-bit big endian)
    mips64      - MIPS (64-bit big endian)
    mips64el    - MIPS (64-bit little endian)
    mipsel      - MIPS (32-bit little endian)
    msp430      - MSP430 [experimental]
    nvptx       - NVIDIA PTX 32-bit
    nvptx64     - NVIDIA PTX 64-bit
    ppc32       - PowerPC 32
    ppc32le     - PowerPC 32 LE
    ppc64       - PowerPC 64
    ppc64le     - PowerPC 64 LE
    r600        - AMD GPUs HD2XXX-HD6XXX
    riscv32     - 32-bit RISC-V
    riscv64     - 64-bit RISC-V
    sparc       - Sparc
    sparcel     - Sparc LE
    sparcv9     - Sparc V9
    systemz     - SystemZ
    thumb       - Thumb
    thumbeb     - Thumb (big endian)
    ve          - VE
    wasm32      - WebAssembly 32-bit
    wasm64      - WebAssembly 64-bit
    x86         - 32-bit X86: Pentium-Pro and above
    x86-64      - 64-bit X86: EM64T and AMD64
    xcore       - XCore
    xtensa      - Xtensa 32
```

There's quite a bit of substring-ambiguity there, e.g. `x86`/`x86-64`, `mips`/`mips64`/`mips64el`/`mipsel`, etc.

It might be worth doing a little bit more rigorous parsing at some point (not a priority right now!)

iwyu_test_util.py
```diff
@@ -419,6 +436,112 @@ def _GetLaunchArguments(cc_file):
...
+    found_block = False
+    # Find the next '// IWYU_*: ' line.
+    for lineno, line in enumerate(it):
+      m = _IWYU_TEST_SKIP_RE.match(line)
+      if m:
+        found_block = True
+        typ = m.group(1)
+        cond = m.group(2)
+
+        try:
```
kimgr, line 520: I'd prefer to separate the parsing from the evaluation here. It looks like this function currently only evaluates the first skip-condition, but in theory there could be any number of them, right?

kimgr: Sorry, I conflated two concerns:

- Please split this up so there's a function that only parses and returns skip directives (maybe as a tuple `(skip_type, skip_cond)`)
- When evaluating the directives, we should probably evaluate all of them and skip if any hold

run_iwyu_tests.py
```diff
@@ -42,6 +42,11 @@ def Partition(l, delimiter):
 
 def TestIwyuOnRelevantFiles(filename):
   logging.info('Testing iwyu on %s', filename)
+
+  skip = iwyu_test_util.CheckSkipConditions(filename)
```
kimgr, line 46: Shouldn't the skip-check happen in `TestIwyuOnRelativeFile` instead? I don't think this function is executed in the plain `./run_iwyu_tests.py` run mode.

kimgr: My mistake; it is. Still feels like it's better to parse/skip in `TestIwyuOnRelativeFile`, but I might be overlooking some complication.

iwyu_test_util.py
```diff
@@ -419,6 +436,112 @@ def _GetLaunchArguments(cc_file):
...
 
+def _CheckForCompileTimeMacro(macro):
+    """This assumes that the local clang++ and the include-what-you-use
+    under test are using the same libclang, and therefore the same default libraries.
+
+    Look for the specified macro in the clang output.
+    """
+    txt = '#include <cstddef>'
+
+    result = subprocess.run(['clang', '-dM', '-E', '-x', 'c++', '-'], capture_output=True,
```
kimgr, line 447: I believe this should include any IWYU args; e.g. if a test specifies `-stdlib=libc++`, we want that to influence the probe here, right? That would also make a bit more sense if we could use IWYU instead of clang.

iwyu_test_util.py
```diff
@@ -419,6 +436,112 @@ def _GetLaunchArguments(cc_file):
...
+    return tgt in str(result.stdout)
+
+  m = _IWYU_COND.match(cond)
+  if m:
+    label = m.group(1)
+    arg = m.group(2)
+
+    if label == 'IwyuHasTargetSupportFor':
+      return iwyuHasTargetSupportFor(arg)
+    elif label == 'TestingCXXSTL':
```
kimgr, line 493: I don't really understand what this label is saying, particularly not with the `canonical` mapping mixed in.

Could you maybe split these features up into two:

- `stdlib {libcxx|libstdcxx}`
- `builtin-mappings {canonical|default?|...?}`

It's not 100% clear to me what the mappings tag would indicate, so maybe let's cut it for now and use the target feature as a motivating example?

run_iwyu_tests.py
kilroyd: I had it here originally, moved it to `TestIwyuOnRelativeFile` and moved it back because `iwyu_test_utils.py` doesn't currently import `unit_tests`... so we can't throw the exception. Adding the import is trivial, but it's kinda nice that only `run_iwyu_tests.py` imports `unit_tests`. Happy to move if you prefer it that way.

kilroyd:
> Thanks! Please don't force-push overwrite existing versions, I think this is enough uncharted territory that I'm going to change my mind a few times along the way :)

Apologies. I've pushed the original commit to the [test_skip0](/kilroyd/include-what-you-use/tree/test_skip0) branch in my repo in case you want to reference it.

iwyu_test_util.py
kilroyd: I just intended this to find the 'default' stdlib of libclang, so I didn't add the IWYU args. I explicitly check the IWYU args separately, which allows me to pick up canonical (or no_mappings). Calling clang, passing IWYU_EXTRA_ARGS means we would have to filter out the `-Iwyu` specific args so we don't confuse it.

If we can get IWYU to output the macros though, then yes, we can just pass the args in and remove the explicit arg check. I'll see whether I can get that working.

I'm thinking I should reduce this commit to handle just the `has_target_support_for`, and we can add more conditions later.

kimgr:
@kilroyd No worries, but thanks!

run_iwyu_tests.py
kimgr: Aha, I see. Yeah, I think it's fine to import `unittest` in `iwyu_test_util.py` for this purpose.

iwyu_test_util.py
kilroyd: FYI, this evaluated multiple conditions, but exited on the first condition indicating we should skip. Splitting as requested.

iwyu_test_util.py
```diff
@@ -45,6 +45,26 @@
 # // IWYU_ARGS: -Xiwyu --mapping_file=... -I .
 _IWYU_TEST_RUN_ARGS_RE = re.compile(r'^//\sIWYU_ARGS:\s(.*)$')
 
+# This is a line that specifies when a test should be skipped (because
+# the result would not be valid). IWYU_REQUIRES specifies a condition
+# that must be true for the test to execute. IWYU_UNSUPPORTED
+# specifies a condition in which the test is invalid, and would be
+# skipped. Multiple IWYU_REQUIRES and IWYU_UNSUPPORTED can be
+# specified, but they must be in a single comment block. For a test to
```
kimgr, line 53: > but they must be in a single comment block

This seems like an artificial constraint. Isn't parsing much more straightforward if you skip this rule?

iwyu_test_util.py
```diff
@@ -45,6 +45,26 @@
...
+#
+# // IWYU_REQUIRES: cond
+# // IWYU_UNSUPPORTED: cond
+_IWYU_TEST_SKIP_RE = re.compile(r'^//\sIWYU_(REQUIRES|UNSUPPORTED):\s(.*)$')
+
+# List of all supported condition labels
+_IWYU_COND_LABELS=['iwyu_has_target_support_for']
+
+# All conditions have the same format 'LABEL(ARG)'
+_IWYU_COND_RE = re.compile(r'('+'|'.join(_IWYU_COND_LABELS)+')\(([A-z][A-z0-9_-]+)\)')
```
kimgr, line 67: Can you put spaces around string concatenation `+`-es, so it's easier to tell them from the regular expression?

kimgr:
I got a little excited and maybe ran too far, but please take a look at this variant I hacked up: https://github.com/kimgr/include-what-you-use/commit/kg-test-skip -- I think it captures all the extensibility with less code.

kilroyd:
> I got a little excited and maybe ran too far, but please take a look at this variant I hacked up: [kimgr@kg-test-skip](https://github.com/kimgr/include-what-you-use/commit/kg-test-skip) -- I think it captures all the extensibility with less code.

That looks pretty neat. I've no objection if you want to merge that

kilroyd:
I've force pushed what is essentially your version. I've just fixed a bug with the lambda capture of `arg` when you have multiple prerequisites.

kimgr:
@kilroyd Oh, sorry, we were running in parallel there. I just opened #1315 with a bunch of fixup commits on top, mostly small things. I cherrypicked your doc commit and the bugfix (thanks for spotting that!).

The big change is `iwyu_has_target_support_for` -> `target`, which I think reads better with both `IWYU_REQUIRES:` and `IWYU_UNSUPPORTED`. Maybe let's continue discussion on #1315 and I can squash it once we're both happy?

kilroyd:
No worries :) I wasn't planning on doing anything further on this, I just wanted to test your change and add some documentation - you've picked those up, so that's great.

I'll close down this PR, and let you drive #1315

PR #1313, opened 2023-10-15
Handle "autocast" type components separately

bolshakov-a:
Similar to https://github.com/include-what-you-use/include-what-you-use/pull/1248 and https://github.com/include-what-you-use/include-what-you-use/pull/1304.

At the call site, not only argument type reporting is handled (in `VisitCXXConstructExpr`) but also constructor call in case of template (`HandleFunctionCall`). It is because traversal of implicit constructor call in "autocast" context should not produce provided type reportings as well.

`comment` parameter of `ReportTypeUse` function becomes unused and is removed.

One subtle thing: copy and move constructor usages are not reported as "autocast" ones. In fact, they are reported nowhere at all (along with some other implicit construction cases), and it is a bug. It could be fixed by removing that filtering in `IsAutocastExpr`, in principle, but, besides being conceptually incorrect, it results in redundant reporting for passed-by-value function templated parameter when template argument is "providing" type alias, which may be undesirable.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1600,11 +1600,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
   // Called when the given type is fully used at used_loc, regardless
   // of the type being explicitly written in the source code or not.
-  // The comment, if not nullptr, is extra text that is included along
-  // with the warning message that iwyu emits.
-  virtual void ReportTypeUse(SourceLocation used_loc, const Type* type,
-                             const char* comment = nullptr) {
-    ReportTypeUseInternal(used_loc, type, comment, blocked_types_);
+  virtual void ReportTypeUse(SourceLocation used_loc, const Type* type) {
+    ReportTypeUseInternal(used_loc, type, blocked_types_);
   }
 
   void ReportTypesUse(SourceLocation used_loc, const set<const Type*>& types) {
@@ -2610,7 +2607,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   }
 
   void ReportTypeUseInternal(SourceLocation used_loc, const Type* type,
-                             const char* comment,
                              const set<const Type*>& blocked_types) {
     if (CanIgnoreType(type))
       return;
@@ -2655,7 +2651,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         const Type* type = RemovePointersAndReferencesAsWritten(
             typedef_decl->getUnderlyingType().getTypePtr());
         IwyuBaseAstVisitor<Derived>::ReportTypeUseInternal(
-            used_loc, type, nullptr, provided_with_typedef);
+            used_loc, type, provided_with_typedef);
       }
       return;
     }
@@ -2677,7 +2673,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (const NamedDecl* decl = TypeToDeclAsWritten(type)) {
       decl = GetDefinitionAsWritten(decl);
       VERRS(6) << "(For type " << PrintableType(type) << "):\n";
-      IwyuBaseAstVisitor<Derived>::ReportDeclUse(used_loc, decl, comment);
+      IwyuBaseAstVisitor<Derived>::ReportDeclUse(used_loc, decl);
     }
   }
 
@@ -2957,8 +2953,7 @@ class InstantiatedTemplateVisitor
     // implicit stuff, it should not suggest forward declarations.
   }
 
-  void ReportTypeUse(SourceLocation used_loc, const Type* type,
-                     const char* comment = nullptr) override {
+  void ReportTypeUse(SourceLocation used_loc, const Type* type) override {
     // clang desugars template types, so Foo<MyTypedef>() gets turned
     // into Foo<UnderlyingType>().  Try to convert back.
     type = ResugarType(type);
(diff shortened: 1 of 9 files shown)


iwyu.cc
```diff
@@ -2611,6 +2541,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       // in source files, or for builtins, or for friend declarations.
       if (!IsInHeader(decl) || IsFriendDecl(decl))
         return pair(false, nullptr);
+      for (auto paramIt = decl->param_begin(); paramIt != decl->param_end();
```
kimgr, line 2541: `FunctionDecl` has a nice `ArrayRef<ParmVarDecl*> parameters()` wrapper to make this a little neater;
```
for (const ParmVarDecl* param : decl->parameters()) {
   // param instead of *paramIt
}
```

iwyu.cc
```diff
@@ -2753,10 +2699,23 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    // 'Autocast' -- calling a one-arg, non-explicit constructor
+    // -- is a special case when it's done for a function call.
+    // iwyu requires the function-writer to provide the #include
+    // for the casted-to type, just so we don't have to require it
+    // here.  *However*, the function-author can override this
+    // iwyu requirement, in which case we're responsible for the
+    // casted-to type.  See IwyuBaseASTVisitor::CanBeProvidedTypeComponent.
+    if (!IsAutocastExpr(current_ast_node()))
+      return;
+    const Type* type = Desugar(expr->getType().getTypePtr());
```
kimgr, line 2712: Can you add a little comment here as to why this `Desugar` is necessary? It feels counter-intuitive to me, but I know I have absolutely zero intuition for sugar :)

iwyu_ast_util.cc
```diff
@@ -660,6 +674,19 @@ bool IsTemplatizedFunctionDecl(const FunctionDecl* decl) {
 }
 
 bool HasImplicitConversionCtor(const CXXRecordDecl* cxx_class) {
+  // Member functions of class templates aren't instantiated unless used,
```
kimgr, line 677: Optional: This looks like a bug that just happened to surface now. Is it possible to commit it separately with an isolated testcase before this PR?

tests/cxx/iwyu_stricter_than_cpp.cc
```diff
@@ -263,6 +263,36 @@ void TestAutocast() {
       6, 7, 8, 9,
       // IWYU: TplIndirectStruct2 is...*iwyu_stricter_than_cpp-i2.h
       10);
+
+  TplFnValues(
+      // IWYU: TplIndirectStruct3 is...*iwyu_stricter_than_cpp-i5.h
+      1,  // onlyArgumentTypeProvided
```
kimgr, line 269: We use a tricky named-arguments style in IWYU (derived from LLVM, I think), like here: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_ast_util.cc#L433.

Would probably be nice to use that here as well, e.g. 
```
TplFnValues(
  // IWYU...
  /*onlyArgumentTypeProvided=*/1,
  // IWYU...
  /*allForwardDeclared=*/2,
  ...

File: tests/cxx/iwyu_stricter_than_cpp-autocast.h [hunk 5]
```diff
@@ -65,51 +63,111 @@ template <typename T>
...
+// --- With user-defined types as template parameters.
+
+// struct IndirectStruct2; (fwd-declared above)
+template <typename TFullTypeUsed, typename TForwardDeclarable>
+struct TplIndirectStruct3;
+
+void TplFnValues(
+    // IWYU: IndirectStruct1 is...*iwyu_stricter_than_cpp-i1.h.*for autocast
+    TplIndirectStruct3<IndirectStruct1, IndirectStruct2>
+        onlyArgumentTypeProvided,
```
kimgr, line 99: Variables/parameters use `hacker_case_with_underscore`, not `camelCase` (though I'm sure there are plenty of places this rule is broken).

tests/cxx/iwyu_stricter_than_cpp-autocast.h
```diff
@@ -65,51 +63,111 @@ template <typename T>
...
+struct NoAutocastForSpec {
+  NoAutocastForSpec(int) {
+  }
+  T t;
+};
+template <>
+// IWYU: IndirectStruct1 needs a declaration
+struct NoAutocastForSpec<IndirectStruct1> {};
+// IWYU: IndirectStruct1 needs a declaration
+void Fn(NoAutocastForSpec<IndirectStruct1>);
```
kimgr, line 162: Should there be a TODO here to match https://github.com/include-what-you-use/include-what-you-use/pull/1313/commits/d847df1e54a0299bc0d9d39ef9048925a5291e12#diff-7ec045b2b5739efe5504c9d31e7e935fe9405be09ae655d067bda89fdd9cf6f9R687?

kimgr:
Overall looks great! A few nits inline.

iwyu.cc
bolshakov-a: I think it is unnecessary because nearly all control flow paths in `ReportTypeUse` desugar the given type as well. The line just moved from [here](https://github.com/include-what-you-use/include-what-you-use/pull/1313/files/d847df1e54a0299bc0d9d39ef9048925a5291e12#diff-7ca107c8b836c1765e8b735b74f06833bec48f308a76017886cee1d5736002d9L2221).

iwyu_ast_util.cc
kimgr: #1314 merged, thanks!

kimgr:
Very nice, thanks!

PR #1315, opened 2023-10-25
Add infrastructure for test prerequisites

kimgr:
To be squashed...

### DIFF
--- a/docs/IWYUStyle.md
+++ b/docs/IWYUStyle.md
@@ -281,6 +281,23 @@ comment:
 The args are added to the `include-what-you-use` command as-written after line
 unwrapping, so IWYU flags must be explicitly prefixed with `-Xiwyu`.
 
+### Prerequisites ###
+
+If a test has prerequisites, annotate the `.cc` file itself using a
+directive comment:
+
+```
+// IWYU_REQUIRES: feature(value)
+// IWYU_UNSUPPORTED: feature(value)
+```
+
+`IWYU_REQUIRES` states that the following feature test must evaluate true for
+the test to be executed.
+
+`IWYU_UNSUPPORTED` states that the test is only executed when the feature test
+evaluates as false.
+
+If any prerequisite fails the test will be skipped.
 
 [1]: https://google.github.io/styleguide/cppguide.html
 [2]: https://llvm.org/docs/CodingStandards.html
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -21,6 +21,7 @@ import re
 import shlex
 import subprocess
 import sys
+import unittest
 
 # These are the warning/error lines that iwyu.cc produces when --verbose >= 3
 _EXPECTED_DIAGNOSTICS_RE = re.compile(r'^\s*//\s*IWYU:\s*(.*)$')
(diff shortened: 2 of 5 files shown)


docs/IWYUStyle.md
```diff
@@ -281,6 +281,23 @@ comment:
...
+```
+// IWYU_REQUIRES: feature(value)
+// IWYU_UNSUPPORTED: feature(value)
+```
+
+`IWYU_REQUIRES` states that the following feature test must evaluate
+true for the test to be valid.
+
+`IWYU_UNSUPPORTED` states that the test is invalid when the feature
+test evaluates as false.
```
kilroyd, line 298: I got this line wrong :(, the test is invalid when this feature test is true. Flipped too many things at once

iwyu_test_util.py
```diff
@@ -419,6 +430,78 @@ def _GetLaunchArguments(cc_file):
...
+class Features:
+  """ Implements feature support functions for prerequisites. """
+
+  # Parses the output of 'include-what-you-use -print-targets'.
+  SUPPORTED_TARGET_RE = re.compile(r'^\s+([a-z0-9_-]+)\s* - .*$')
+
+  def __init__(self):
+    # Maps all known features names to prerequisite predicates.
+    self.known = {
+      "target": Features.Target
```
kilroyd, line 442: My concern with just `target` is that it might be confused with the compiler target triple.

Also it's not the target that IWYU would compile by default, so I feel `supports-target` would be preferable. Maybe `supports-arch`? Not sure.

run_iwyu_tests.py
```diff
@@ -124,7 +125,16 @@ def PrintLoadedTestsAndFiles():
   (runner_args, _) = parser.parse_known_args(unittest_args)
 
   if runner_args.run_test_file:
-    exit(TestIwyuOnRelevantFiles(runner_args.run_test_file))
+    try:
+      r = TestIwyuOnRelevantFiles(runner_args.run_test_file)
+    except unittest.SkipTest as e:
+      # Catch the skip test exception that unittest would normally
+      # handle. Instead set a return code that informs Ctest that the
```
kilroyd, line 132: Maybe mention that we've aligned the return code with GNU?

kilroyd (commented):
Cool! Various observations inline

iwyu_test_util.py
kimgr: I found it reads really poorly with `IWYU_UNSUPPORTED: supports-target(arm64)`...

kilroyd: Good point. Some other options - `can-target(arm64)`, `valid-target(arm64)`, `targetable(arm64)`

kilroyd: Thinking about this overnight, `IWYU_UNSUPPORTED: supports-targets(arm64)` (or whatever we end up calling it) is not something anyone would ever use in a test.

If a test doesn't select a `-target` then the prerequisite would be saying that `arm64` support is somehow causing our test to fail even when it's not being used. Same if the test is running with a non-`arm64` target - support for the arm64 target would somehow be polluting the behavior of the target being used.

The final case to consider is when the test asks for `-target arm64`: the prerequisite would be saying the test is invalid if IWYU has support for target. That's clearly a contradiction.

Obviously it should be accepted syntax. If it reads wrong, I don't think that's not the worst thing - as it will make the reviewer think a bit more (hopefully). What must read well is the positive case `IWYU_REQUIRES: supports-target(arm64)`

kimgr: Yeah, good call, that makes sense.

Maybe `has-target`?

docs/IWYUStyle.md
kimgr: Good catch, fixed!

iwyu_test_util.py
kimgr: Pushed `has-target` to see what it looks like

run_iwyu_tests.py
kimgr: Done

kilroyd (approved):
What you've got looks good to me!

kimgr:
I found a few more cleanups, but no functional changes. Merging this as-is now. Thanks for the idea and for helping improve my prototype!

PR #1317, opened 2023-10-27
Remove EnumDecl visitation

bolshakov-a:
It was introduced in #742 to report underlying type aliases but it is seemingly not needed anymore, probably after #775 where `TypedefType` visitation has been added.

Actually, type aliases and typedefs cannot be forward-declared and thus should be reported if and only if they're explicitly written in the source (what `VisitTypedefType` already does). An enum underlying type itself [should be of integral type](https://eel.is/c++draft/dcl.enum#2) and hence cannot be defined in any of user's headers.

No functional change.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1627,17 +1627,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


kimgr:
Thanks!

PR #1318, opened 2023-10-27
Fix prerequisite syntax error reporting

kilroyd:


kimgr:
:cry: I should know better than developing without tests of the testsupport code. Thanks for this!

PR #1319, opened 2023-10-27
Add infrastructure for expected test fails

kilroyd:
Allow tests to indicate that they are expected to fail. Either unconditionally or in particular circumstances:

    IWYU_XFAIL
    IWYU_XFAIL: feature(value)

When the condition is true the test is expected to fail.

When the test fails it is reported as a pass. Unexpected passes are reported as a failure.

With `unittest`, one should normally use `@unittest.expectedFailure` as a decorator. Since we're generating the testcases dynamically we need to call `unittest.expectedFailure` with the test item lambda that we've just added.

Explicitly handle expected failure with `--run-test-file` so that CMake reports the same results as `unittest`. We don't try to use `WILL_FAIL` as that isn't conditional.

### DIFF
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -46,6 +46,10 @@ _NODIFFS_RE = re.compile(r'^\((.*?) has correct #includes/fwd-decls\)$')
 # // IWYU_ARGS: -Xiwyu --mapping_file=... -I .
 _IWYU_TEST_RUN_ARGS_RE = re.compile(r'^//\sIWYU_ARGS:\s(.*)$')
 
+# Text matching a condition, either as part of IWYU_REQUIRES,
+# IWYU_UNSUPPORTED or IWYU_XFAIL.
+_IWYU_CONDITION = r'([a-z][a-z0-9_-]*)\(([^)]+)\)'
+
 # This is a line that specifies prerequisites for a test (test will be skipped
 # if they don't all hold). IWYU_REQUIRES specifies a condition that must be true
 # for the test to execute. IWYU_UNSUPPORTED specifies a condition in which the
@@ -53,12 +57,18 @@ _IWYU_TEST_RUN_ARGS_RE = re.compile(r'^//\sIWYU_ARGS:\s(.*)$')
 # IWYU_UNSUPPORTED lines can be specified. Examples:
 # // IWYU_REQUIRES: feature(arg)
 # // IWYU_UNSUPPORTED: feature(arg)
-_IWYU_TEST_PREREQ_RE = re.compile(r'^// IWYU_(REQUIRES|UNSUPPORTED): '
-                                  r'([a-z][a-z0-9_-]*)\(([^)]+)\)$')
+_IWYU_TEST_PREREQ_RE = re.compile(r'^// IWYU_(REQUIRES|UNSUPPORTED): ' +
+                                  _IWYU_CONDITION + r'$')
 
+# This is a line that specifies conditions under which a test is expected to
+# fail.
+# // IWYU_XFAIL: feature(arg)
+_IWYU_TEST_XFAIL_RE = re.compile(r'^// IWYU_XFAIL(?:: ' +
+                                 _IWYU_CONDITION + r')?$')
 
 class Features:
-  """ Implements feature support functions for prerequisites. """
+  """ Implements feature support functions for prerequisites and
+      expected failures. """
 
   # Parses the output of 'include-what-you-use -print-targets'.
   SUPPORTED_TARGET_RE = re.compile(r'^\s+([a-z0-9_-]+)\s* - .*$')
@@ -501,6 +511,39 @@ def _CheckPrerequisites(cc_file):
 
   if failed:
     raise unittest.SkipTest(', '.join(failed))
+
+
+def _ParseExpectedFailures(cc_file):
+  """ Parses expected failure conditions out of cc_file. """
+  conditions = []
+  with open(cc_file) as it:
+    for lineno, line in enumerate(it):
+      m = _IWYU_TEST_XFAIL_RE.match(line)
+      if not m:
+        continue
+
+      feature = m.group(1)
+      value = m.group(2)
+
+      if feature is None:
+        # Bare IWYU_XFAIL, condition is always True
+        condition = lambda: True
+      else:
+        feature_cond = _FEATURES.known.get(feature)
+        if not feature_cond:
+          raise SyntaxError('%s:%s IWYU_XFAIL: unsupported feature: %s' %
+                            (cc_file, lineno, feature))
+
+        condition = lambda arg=value: feature_cond(arg)
+
+      conditions.append(condition)
+    return conditions
+
+
+def IsTestExpectedToFail(cc_file):
+  """ Checks if a test is expected to fail. """
+  xfail_conditions = _ParseExpectedFailures(cc_file)
+  return any(condition() for condition in xfail_conditions)
 
 
 def TestIwyuOnRelativeFile(cc_file, cpp_files_to_check, verbose=False):
--- a/run_iwyu_tests.py
+++ b/run_iwyu_tests.py
@@ -86,6 +86,11 @@ def GenerateTests(rootdir, pattern):
         test_name += '2'               # just append a suffix :-)
 
       setattr(cls, test_name, lambda x, f=filename: TestIwyuOnRelevantFiles(f))
+
+      if iwyu_test_util.IsTestExpectedToFail(filename):
+        test_item = getattr(cls, test_name)
+        unittest.expectedFailure(test_item)
+
       test_files[test_name] = filename
 
     setattr(cls, 'test_files', test_files)
@@ -117,6 +122,7 @@ def RunTestFile(cc_file):
   for failure, 77 if the test was skipped. If the test failed, an
   exception will have been raised (causing a non-zero exit code).
   """
+  xfail = iwyu_test_util.IsTestExpectedToFail(cc_file)
   try:
     TestIwyuOnRelevantFiles(cc_file)
   except unittest.SkipTest as e:
@@ -127,6 +133,15 @@ def RunTestFile(cc_file):
     # https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors
     print('Skipped %s: %s' % (cc_file, e))
     return 77
+  except AssertionError as e:
+    if xfail:
+      print('%s: Expected failure' % cc_file)
+      return 0
+    raise
+  else:
+    if xfail:
+      print('%s: Unexpected pass' % cc_file)
+      return 1
   return 0
 
 


iwyu_test_util.py
```diff
@@ -503,6 +508,42 @@ def _CheckPrerequisites(cc_file):
...
+      desc = '%s(%s)' % (feature, value)
+      conditions.append((condition, desc))
+    return conditions
+
+
+def IsTestExpectedToFail(cc_file):
+  """ Checks if a test is expected to fail """
+  xfail_conditions = _ParseExpectedFailures(cc_file)
+
+  xfails = []
```
kimgr, line 539: Unused

iwyu_test_util.py
```diff
@@ -56,6 +56,11 @@
 _IWYU_TEST_PREREQ_RE = re.compile(r'^// IWYU_(REQUIRES|UNSUPPORTED): '
                                   r'([a-z][a-z0-9_-]*)\(([^)]+)\)$')
 
+# This is a line that specifies conditions under which a test is expected to
+# fail.
+# // IWYU_XFAIL: feature(arg)
```
kimgr, line 61: I'm thinking it might be nice to add unconditional XFAILs too to be able to add testcases for known problems, e.g.:
```
// see https://github.com/include-what-you-use/include-what-you-use/issues/1181
// IWYU_XFAIL
// IWYU_ARGS: ...

...
```

But we can follow up with that later.

run_iwyu_tests.py
```diff
@@ -124,8 +129,9 @@ def PrintLoadedTestsAndFiles():
   (runner_args, _) = parser.parse_known_args(unittest_args)
 
   if runner_args.run_test_file:
+    xfail = iwyu_test_util.IsTestExpectedToFail(runner_args.run_test_file)
     try:
-      r = TestIwyuOnRelevantFiles(runner_args.run_test_file)
```
kimgr, line 128: Let's fix the exit code bug in a separate commit

run_iwyu_tests.py
```diff
@@ -134,8 +140,16 @@ def PrintLoadedTestsAndFiles():
       # https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors
       print('Skipped %s: %s' % (runner_args.run_test_file, e))
       exit(77)
+    except AssertionError as e:
+      if xfail:
+        print("%s: Expected failure" % runner_args.run_test_file)
+        exit(0)
+      raise e from None
```
kimgr, line 147: You can just say `raise` to re-throw the active exception, no need for `from None` (though thanks for demonstrating that, I hadn't seen the construct before!)

run_iwyu_tests.py
```diff
@@ -124,8 +129,9 @@ def PrintLoadedTestsAndFiles():
   (runner_args, _) = parser.parse_known_args(unittest_args)
 
   if runner_args.run_test_file:
```
kimgr, line 131: This whole block is getting complicated enough that I think it might make sense to extract it into a `RunTestFile` function, something like:
```
if runner_args.run_test_file:
  exit(RunTestFile(runner_args.run_test_file))
```

to mirror the existing `PrintLoadedTests...` functions

iwyu_test_util.py
```diff
@@ -503,6 +508,42 @@ def _CheckPrerequisites(cc_file):
...
+
+      condition = lambda arg=value: feature_cond(arg)
+
+      desc = '%s(%s)' % (feature, value)
+      conditions.append((condition, desc))
+    return conditions
+
+
+def IsTestExpectedToFail(cc_file):
+  """ Checks if a test is expected to fail """
```
kimgr, line 536: Google style is really strict about punctuation, so let's use a full stop after `fail`

run_iwyu_tests.py
kimgr: Thanks! I'm stealing this and retrofitting on current master; if you can rebase on that, it will be way clearer what's being added here. Sorry about the back and forth. I'll ping back when I've tested it.

kimgr: Now on master, please rebase.

iwyu_test_util.py
```diff
@@ -46,16 +46,25 @@
 # // IWYU_ARGS: -Xiwyu --mapping_file=... -I .
 _IWYU_TEST_RUN_ARGS_RE = re.compile(r'^//\sIWYU_ARGS:\s(.*)$')
 
+# Text matching a condition, either as part of IWYU_REQUIRED,
```
kimgr, line 49: IWYU_REQUIRED -> IWYU_REQUIRES

iwyu_test_util.py
```diff
@@ -503,6 +512,47 @@ def _CheckPrerequisites(cc_file):
...
+
+      conditions.append((condition, desc))
+    return conditions
+
+
+def IsTestExpectedToFail(cc_file):
+  """ Checks if a test is expected to fail. """
+  xfail_conditions = _ParseExpectedFailures(cc_file)
+
+  for condition, _ in xfail_conditions:
```
kimgr, line 549: You could consider `return any(condition() for condition, _ in xfail_conditions)`; then this would be a two-line function.

iwyu_test_util.py
```diff
@@ -503,6 +512,47 @@ def _CheckPrerequisites(cc_file):
...
+        continue
+
+      feature = m.group(1)
+      value = m.group(2)
+
+      if feature is None:
+        # Bare IWYU_XFAIL, condition is always True
+        condition = lambda: True
+        desc = 'always'
+
```
kimgr, line 531: Kill blank line before else

iwyu_test_util.py
kimgr: But if it looks too opaque to you (I'm on the fence), please remove the empty lines in this functions

run_iwyu_tests.py
```diff
@@ -111,6 +116,39 @@ def PrintLoadedTestsAndFiles():
...
+    # set an exit code that informs the test runner that test is being
+    # skipped. It's an almost-convention to use exit code 77, and some build
+    # systems natively report such failures as skipped, see e.g.
+    # https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors
+    print('Skipped %s: %s' % (runner_args.run_test_file, e))
+    rc = 77
+  except AssertionError as e:
+    if xfail:
+      print("%s: Expected failure" % runner_args.run_test_file)
+      rc = 0
```
kimgr, line 141: I'd skip `rc` and return in place here, then you can also drop the `else` on the next line.

run_iwyu_tests.py
```diff
@@ -111,6 +116,39 @@ def PrintLoadedTestsAndFiles():
...
+    else:
+      raise e
```
kimgr, line 143: Remove `e` to make it a re-throw and keep the context

run_iwyu_tests.py
```diff
@@ -111,6 +116,39 @@ def PrintLoadedTestsAndFiles():
...
+  xfail = iwyu_test_util.IsTestExpectedToFail(cc_file)
+  try:
+    TestIwyuOnRelevantFiles(cc_file)
+  except unittest.SkipTest as e:
+    # Catch the skip test exception that unittest would normally handle and
+    # set an exit code that informs the test runner that test is being
+    # skipped. It's an almost-convention to use exit code 77, and some build
+    # systems natively report such failures as skipped, see e.g.
+    # https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors
+    print('Skipped %s: %s' % (runner_args.run_test_file, e))
```
kimgr, line 136: runner_args.run_test_file -> cc_file here and below

run_iwyu_tests.py
kimgr: and same throughout this function

iwyu_test_util.py
kimgr: Cool, thanks!

run_iwyu_tests.py
kilroyd: Doh! I didn't retest --run-test-file after that rebase!

kilroyd:
Thanks for taking a look! Rebased now. I've left all the fixes for your comments as separate commits for now. Expect to squash the lot. I'll revise the description on the PR to match the overall functionality

kimgr:
Oops, pushed a fixup for the `RunTestFile` move to master. Please rebase again.

iwyu_test_util.py
```diff
@@ -46,16 +46,25 @@
...
+                                  _IWYU_CONDITION + r'$')
 
+# This is a line that specifies conditions under which a test is expected to
+# fail.
+# // IWYU_XFAIL: feature(arg)
+_IWYU_TEST_XFAIL_RE = re.compile(r'^// IWYU_XFAIL(?:: ' +
+                                 _IWYU_CONDITION + r')?$')
 
 class Features:
   """ Implements feature support functions for prerequisites. """
```
kimgr, line 70: ... and expected failures

iwyu_test_util.py
```diff
@@ -503,6 +512,41 @@ def _CheckPrerequisites(cc_file):
...
+      if not m:
+        continue
+
+      feature = m.group(1)
+      value = m.group(2)
+
+      if feature is None:
+        # Bare IWYU_XFAIL, condition is always True
+        condition = lambda: True
+        desc = 'always'
```
kimgr, line 530: Since we don't have a way to communicate the reason for XFAIL, maybe we should just not produce the desc here? Then you can just return a list of conditions and don't have to ignore the desc when evaluating?

run_iwyu_tests.py
kimgr: Fun thing is it actually works, see my commit message :) -- `runner_args` is global, so everything is hunky-dory.

iwyu_test_util.py
kilroyd: Yeah, happy to do that. I wasn't sure if we might want to share the code handling prerequiesites and expected failures so left in the descriptions.

kimgr: Yeah, I think trying to merge them will just make them more complex, but there may be opportunities. Let's remove the desc for now, and get this merged, and we can refactor if we see opportunities later.

run_iwyu_tests.py
```diff
@@ -127,6 +133,15 @@ def RunTestFile(cc_file):
     # https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors
     print('Skipped %s: %s' % (cc_file, e))
     return 77
+  except AssertionError as e:
+    if xfail:
+      print("%s: Expected failure" % cc_file)
```
kimgr, line 138: Ah, sorry, one more thing: historically, our Python has preferred single-quotes for strings unless the string value contains single quotes. So let's use singles here and below.

Otherwise everything LGTM, and I've tested it locally to great effect, so if you would also squash the branch to a single commit, that'd be great.

kilroyd: Done. Squashed and force pushed.

kimgr:
Thanks, beautiful!

PR #1320, opened 2023-10-28
Avoid using TestIwyuOnRelavantFiles return code

kilroyd:
The function doesn't return a value. It just raises an exception upon any errors. In --run-test-file handling just explicitly set a successful exit code (0).


kimgr:
Thanks!

PR #1321, opened 2023-10-28
Remove DesugarDependentTypedef

bolshakov-a:
It had been added in #214 but seemingly not needed at now, probably after blocking of reporting implicit typedefs in #1180. Type alias underlying types are extracted in the `ReportTypeUse` function, with respect to "provision" status (and substituted template parameter types are not considered as provided).

`CanIgnoreType` call is removed from `VisitMemberExpr`. `ReportTypeUse` overloads check it after #1179 already.

Btw, it seems like `CanIgnoreType` and `ResugarType` calls could be removed from `InstantiatedTemplateVisitor::VisitCXXConstructExpr` as well. A confusing point is that `ReportTypeUse` calls them in reverse order, but all tests passed at least.

No functional change.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1903,8 +1903,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     CHECK_(base_type && "Member's base does not have a type?");
     const Type* deref_base_type  // For myvar->a, base-type will have a *
         = expr->isArrow() ? RemovePointerFromType(base_type) : base_type;
-    if (CanIgnoreType(deref_base_type))
-      return true;
     // Technically, we should say the type is being used at the
     // location of base_expr.  That may be a different file than us in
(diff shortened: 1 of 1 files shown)


kimgr:
Thanks!

kimgr:
> Btw, it seems like CanIgnoreType and ResugarType calls could be removed from
> InstantiatedTemplateVisitor::VisitCXXConstructExpr as well. A confusing point is
> that ReportTypeUse calls them in reverse order, but all tests passed at least.

Shouldn't we check both the desugared and resugared type for being ignored?

bolshakov-a:
It would be strange because most of the functions called from `InstantiatedTemplateVisitor::CanIgnoreType` canonicalize the given argument, so sugared and desugared versions of the same canonical type should be treated the same. Maybe, something related with `nodes_to_ignore_`, but I think it's rather happened accidentally.

PR #1322, opened 2023-10-28
Support checking standard library provenance

kilroyd:
Update `Features` to support `using-cxxstdlib` and `using-cstdlib` to check the provenance of the C++ and C standard libraries respectively.

Add support to `include-what-you-use` to dump defined macros. Use the `-dM` argument to trigger this (but skip the requirement for `-E`). I couldn't see a nice way to get clang to execute the frontend action by itself, but it's pretty trivial for us to do it.

Let me know what you think.


iwyu_test_util.py
```diff
@@ -92,6 +134,32 @@ def HasTarget(target):
...
+      assert False, ('Unknown STL: %s' % lib)
+
+    return Features._CheckForCXXPreprocessorMacro(id_macro[lib])
+
+  @staticmethod
+  def UsingCstdlib(lib):
+    id_macro = {
+      'glibc' : '__GLIBC__',
+      'macos' : '__APPLE_CC__', # This identifies the Apple compiler, is this good enough?
+      'msvc' : '_MSVC_VER', # This identifies the MS compiler, is this good enough?
```
kimgr, line 154: The compiler-identifying symbol for MSVC actually `_MSC_VER`. And yeah, I think it's good enough!

iwyu_test_util.py
```diff
@@ -73,12 +73,54 @@ class Features:
...
+
+  @staticmethod
+  def _CheckForCXXPreprocessorMacro(macro):
+    """ Look for the specified macro in the C++ preprocessor output. """
+    return macro in Features._GetCXXPreprocessorMacros()
+
+  @staticmethod
+  def _CheckForCPreprocessorMacro(macro):
+    """ Look for the specified macro in the C preprocessor output. """
+    return macro in Features._GetCPreprocessorMacros()
```
kimgr, line 122: I don't think these two methods are necessary, let's just inline the `macro in` into the `Using.*Lib` methods?

iwyu_test_util.py
```diff
@@ -92,6 +134,32 @@ def HasTarget(target):
...
 
+  @staticmethod
+  def UsingCXXstdlib(lib):
+    id_macro = {
+      'libcxx' : '_LIBCPP_VERSION',
+      'libstdcxx' : '__GLIBCXX__',
+      'msvc' : '_MSVC_STL_UPDATE', # https://github.com/microsoft/STL/wiki/Macro-_MSVC_STL_UPDATE
+    }
+    if lib not in id_macro:
+      assert False, ('Unknown STL: %s' % lib)
```
kimgr, line 145: I don't think we should assert on user input.  How about `raise SyntaxError(...)` instead, even if we don't have file/line information?

Same for `UsingCstdlib` below.

iwyu_test_util.py
```diff
@@ -73,12 +73,54 @@ class Features:
...
+      'has-target': Features.HasTarget,
+      'using-cstdlib': Features.UsingCstdlib,
+      'using-cxxstdlib': Features.UsingCXXstdlib,
     }
 
+  def _GetPreprocessorMacros(lang, header):
+    iwyu = _GetIwyuPath()
+    txt = f'#include {header}'
+
+    result = subprocess.run([iwyu, '-dM', '-x', lang, '-'], capture_output=True,
```
kimgr, line 91: For one, I was thinking that we should probably use the same command-line as would be used for the test (the one built in `TestIwyuOnRelativeFile`), but I don't really see how to get that into here nicely.

But second, shouldn't the feature be more like "is <library> installed?", in which case the test can explicitly specify `-stdlib=libc++`, or whatever it needs. Now it seems the test suite runs or not depending on what happens to be the compiler default for the platform.

Or maybe both are necessary...?

iwyu_globals.cc
```diff
@@ -177,6 +177,8 @@ OptionsParser::OptionsParser(int argc, char** argv) {
       intercepted_argv[intercepted_argc++] = argv[i];     // intercept --help
     else if (strcmp(argv[i], "--version") == 0)
       intercepted_argv[intercepted_argc++] = argv[i];     // intercept --version
+    else if (strcmp(argv[i], "-dM") == 0)
```
kimgr, line 180: While it's nice that this matches what you would do for the compiler, I wonder if it shouldn't be an IWYU option instead, i.e. something like `-Xiwyu --dump_macros`.

I had just started experimenting with getting IWYU to delegate preprocessing to Clang for `-dM -E`, but you beat me to it :)

iwyu_test_util.py
kimgr: I did a little experiment with Clang in a clean docker environment:
```bash
$ apt install clang

$ head /usr/include/c++/11/cstddef 
// -*- C++ -*- forwarding header.

// Copyright (C) 1997-2021 Free Software Foundation, Inc.
//
// This file is part of the GNU ISO C++ Library.  This library is free
// software; you can redistribute it and/or modify it under the
// terms of the GNU General Public License as published by the
// Free Software Foundation; either version 3, or (at your option)
// any later version.

# shows libstdc++ ^

$ echo "#include <cstddef>" | clang -stdlib=libstdc++ -xc++ -fsyntax-only -
# no error -- libstdc++ is installed ^

$ echo "#include <cstddef>" | clang -stdlib=libc++ -xc++ -fsyntax-only -
<stdin>:1:10: fatal error: 'cstddef' file not found
    1 | #include <cstddef>
      |          ^~~~~~~~~
1 error generated.
# error: libc++ is not installed ^
```

It seems we should be able to do the same using include-what-you-use?

kimgr: Note, only for C++ standard library, libc is hairier.

iwyu_globals.cc
kimgr: I think I have a plan, let me see if I can hack something up and post a branch.

kimgr: Ugh, I got stuck on weird API boundaries, and need to start thinking about dinner.

What I had in mind was roughly...

In iwyu_driver.cc
* only append `-fsyntax-only` if `-E` is NOT on the command-line
* search `compiler->getJobs()` for a job with action `PreprocessJobAction`; if it exists, go into preprocessing-only mode and don't run IWYU
* search `compiler->getJobs()` for a job with action `CompileJobAction`; if it exists, "replace" it by using `IwyuAction` as frontend action
* If neither exist, fail with "expected exactly one compiler job" or similar

Unfortunately I didn't find a nice way for iwyu_driver to communicate back to iwyu.cc's main which frontend action should be executed, etc.

I'll keep poking at it -- the goal is to allow `include-what-you-use -dM -E myfile.c` to follow clang's without us having to override their switches. And I think there may be other preprocessing tasks that would be useful if IWYU supported as normal.

iwyu_test_util.py
kilroyd: Where I'm expecting to use this is in tests for standard library symbols. In the specific case of libc++ and libstdc++ a check using `-stdlib=*` might work. I don't think it extends to the libc's and not familiar enough with whether it would handle other STL's.

I suspect it also depends on clang having been compiled with appropriate support (i.e. knowing the paths to the different headers).

For IWYU though, if the right `-isystem` was passed in I think it would support being pointed at any libc/libc++ headers.

You're right though - I should be passing in all the IWYU test arguments... and that breaks my plan to try and cache the macros :(

Will have to have a think.

iwyu_globals.cc
kilroyd: We could have IWYU `--dump_macros`... we just need to propagate `-dM` into `clang-args` to appropriately configure the preprocessing action. That would be an extra complication... hopefully not too bad.

iwyu_test_util.py
kimgr: I have some systematic ideas for caching, so let's forget about that for now

iwyu_globals.cc
kimgr: I think I've found the way:
https://github.com/kimgr/include-what-you-use/commit/detect-preprocess

This should make it possible to patch up the arguments, job list and driver options all in one place.

iwyu_test_util.py
kimgr: I still have a feeling this is checking the wrong thing. Your C++ stdlib feature checks "does include-what-you-use use this standard library for this test run?".

I would have expected it to check is "is this standard library available on the system?" -- which will let you do stuff like:
```
// IWYU_REQUIRES: has-cxx-stdlb(libc++)
// IWYU_ARGS: -stdlib=libc++ -I .
```

I guess you could use the current feature for stuff like:
```
// IWYU_XFAIL: using-cxxstdlib(libc++)
// IWYU_ARGS: -I .
```

iwyu_globals.cc
kimgr: Updated the branch with some semi-unrelated stuff; I've now smuggled the toolchain out into iwyu.cc so it can be used in `InitGlobals`, as I wanted to do earlier this year :)

Anyway -- based on this branch, I think I can "easily" let the driver decide what it wants to do, and then either replace compiler jobs with IWYU _or_ see that there's only preprocessing to be done, and let Clang handle that transparently.

kilroyd: I kinda get what you're trying to do with the branch. Happy to rebase this over it when it's ready (dropping the first commit)

iwyu_test_util.py
kilroyd: "is this standard library available on the system" is a test that we can only apply for `libc++` and `libstdc++`, as those are the only things that clang is explicitly aware of.

I'm more interested in the test suite supporting the general case where someone has other libc or C++ standard library headers available. The IWYU test suite should be able to run against those headers.

For the standard library tests, I expect the same tests to apply to all libraries. And we'd just mark a few with `IWYU_XFAIL: using-cstdlib(macos)`. I don't think we want tests selecting a library using `IWYU_ARGS`. While it allows you to execute `libc++` and `libstdc++` tests in one run, it means duplicating tests for each standard library we want to support. Using the same files to test all libraries forces us to align IWYU behaviour on all implementations.

So someone could try and run:

    ./run_iwyu_tests.py   // run test suite with default libs
    IWYU_EXTRA_ARGS="-stdlib=libc++" ./run_iwyu_tests.py  // test clangs libc++
    IWYU_EXTRA_ARGS="-stdlib=libstdc++" ./run_iwyu_tests.py  // test clangs libstdc++
    IWYU_EXTRA_ARGS="--nostdlib -I /work/hpx -I /work/musl" ./run_iwyu_tests.py // test local HP C++ and musl
    IWYU_EXTRA_ARGS="--nostdlib -I /work/libcu++ -I /work/uclibc" ./run_iwyu_tests.py // test local Nvidia C++ and uclibc

Not that these libraries necessarily work together - that's up to the user to supply appropriate include paths and #defines in `IWYU_EXTRA_ARGS`.

iwyu_globals.cc
kimgr: Thanks! I'll try to finish it up, but I haven't had much focus time for IWYU lately.

iwyu_test_util.py
kimgr: Thanks for the extended rationale, that makes a lot of sense to me.

I think there might be a case for `IWYU_REQUIRES: <some standard library>` + `-stdlib=that library` + `IWYU_XFAIL` for stuff like repro cases for bugs that only surface with particular standard libraries. It's always nicer to reduce to the language use that triggers the bug, but that usually takes quite a bit of effort.

But I see what you want to do here, and that also seems perfectly valid.

kimgr:
I'm preparing to release IWYU 0.21; I think this PR is a better fit for post-release, so I'll let it sit until that's done.

iwyu_test_util.py
kilroyd: Yes, we might want `IWYU_REQUIRES` + forcing the library for certain tests - those wouldn't behave in the above scenario, but I think that's fine.

FYI, I've started a [branch](/kilroyd/include-what-you-use/tree/cppref_tests) trying to generate some basic tests from cppreference. I think I want to allow manually adding `IWYU_XFAIL: using-libc()` in those and have the generator preserve any `IWYU_XFAIL` on updates. It already feels fiddly with macros. Functions feel doable. Templates... we'll have to see. Anyway more work required.

kilroyd:
> I'm preparing to release IWYU 0.21; I think this PR is a better fit for post-release, so I'll let it sit until that's done.

Yep, no rush :)

iwyu_globals.cc
kimgr: I think `include-what-you-use -dM -E` should work on master now

iwyu_test_util.py
```diff
@@ -73,12 +74,56 @@ class Features:
...
+    # the user may be trying to test with different includes, standard
+    # libraries or whatever.
+    #
+    # For feature tests we _don't_ take into account test-specific
+    # IWYU_ARGS.
+    #  * the argument could be the thing that breaks things
+    #  * the feature result should not be test specific
+    env_iwyu_extra_args = os.getenv('IWYU_EXTRA_ARGS')
+    if env_iwyu_extra_args:
+      cmd += shlex.split(env_iwyu_extra_args)
```
kimgr, line 104: Seems worth it to extract this into a function that we can reuse between here and `TestIwyuOnRelativeFile`

iwyu_test_util.py
```diff
@@ -73,12 +74,56 @@ class Features:
...
-      'has-target': Features.HasTarget
+      'has-target': Features.HasTarget,
+      'using-cstdlib': Features.UsingCstdlib,
+      'using-cxxstdlib': Features.UsingCXXstdlib,
     }
 
+  @staticmethod
+  def _GetPreprocessorMacros(lang, header):
+    iwyu = _GetIwyuPath()
+    txt = f'#include {header}'
```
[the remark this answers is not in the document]
kimgr: Also `src` seems like a better name than `txt`?

iwyu_test_util.py
```diff
@@ -92,6 +137,32 @@ def HasTarget(target):
         supported.append(m.group(1))
     return target in supported
 
+  @staticmethod
+  def UsingCXXstdlib(lib):
+    id_macro = {
+      'libcxx' : '_LIBCPP_VERSION',
```
kimgr, line 143: No space before colon, here and throughout

iwyu_test_util.py
```diff
@@ -92,6 +137,32 @@ def HasTarget(target):
...
+      'libstdcxx' : '__GLIBCXX__',
+      'msvc' : '_MSVC_STL_UPDATE', # https://github.com/microsoft/STL/wiki/Macro-_MSVC_STL_UPDATE
```
kimgr, line 145: I think it would look nicer with this comment on a separate line above `'msvc': ...`

iwyu_test_util.py
```diff
@@ -92,6 +137,32 @@ def HasTarget(target):
...
+    }
+    if lib not in id_macro:
```
kimgr, line 147: Try:
```
macro = id_macro.get(lib)
if not macro:
   ...
return macro in ...
```
So you don't have to look up twice.

iwyu_test_util.py
```diff
@@ -92,6 +137,32 @@ def HasTarget(target):
...
+  @staticmethod
+  def UsingCstdlib(lib):
+    id_macro = {
+      'glibc' : '__GLIBC__',
+      'macos' : '__APPLE_CC__', # This identifies the Apple compiler rather than libc.
+      'msvc' : '_MSC_VER', # This identifies the MS compiler rather than libc.
+      'newlib' : '_NEWLIB_VERSION',
+      'uclibc' : '__UCLIBC__',
+    }
+    if lib not in id_macro:
```
kimgr, line 161: Same here with double lookup

iwyu_test_util.py
```diff
@@ -92,6 +137,32 @@ def HasTarget(target):
...
+      raise SyntaxError('Unknown C++ standard library: %s' % lib)
+
+    return id_macro[lib] in Features._GetCXXPreprocessorMacros()
+
+  @staticmethod
+  def UsingCstdlib(lib):
+    id_macro = {
+      'glibc' : '__GLIBC__',
+      'macos' : '__APPLE_CC__', # This identifies the Apple compiler rather than libc.
```
kimgr, line 156: I guess this will only happen if we've somehow built `include-what-you-use` against the libraries that form the Apple clang compiler. Are those libraries ever shipped? I.e. can this ever happen?

iwyu_test_util.py
```diff
@@ -92,6 +137,32 @@ def HasTarget(target):
...
+      'msvc' : '_MSC_VER', # This identifies the MS compiler rather than libc.
```
kimgr, line 157: It's been a while since I was on Windows, but I think the MS libc is most often refered to as "msvcrt", because that's what the dynamic library housing it is called. The docs refer to it as "Microsoft CRT": https://learn.microsoft.com/en-us/cpp/c-runtime-library/c-run-time-library-reference?view=msvc-170, but I've never seen "mscrt" being used as an acronym for it.

There's also the Universal CRT, "UCRT", which seems to be the successor to msvcrt.

Anyway, both these libraries can be and are used by other compilers than MSVC (e.g. MinGW gcc), so I don't think it's 100% fair to tag them "msvc".

iwyu_test_util.py
```diff
@@ -73,12 +74,56 @@ class Features:
...
+  # Parses the output of 'include-what-you-use -dM -E'.
+  PREPROCESSOR_MACRO_RE = re.compile(r'^#define ([A-z_][A-z0-9_]*)')
+
   def __init__(self):
     # Maps all known features names to prerequisite predicates.
     self.known = {
-      'has-target': Features.HasTarget
+      'has-target': Features.HasTarget,
+      'using-cstdlib': Features.UsingCstdlib,
+      'using-cxxstdlib': Features.UsingCXXstdlib,
```
kimgr, line 85: I'm still not convinced about the `using-` naming here. I would like to try and come up with some alternatives and compare and contrast. Just throwing some out here, I haven't thought about them much:

* `with-...stdlib`
* `active-...stdlib`
* `...stdlib`

More welcome :)

kimgr: Actually, it occurred to that maybe we should make it even more primitive: `has-predefined-macro()`?

But hmm, yeah, that doesn't work with the symbols in headers (provided by `<cstddef>` and `<stddef.h>` respectively). Worth considering something more basic at least, if there's some other way of achieving that.

iwyu_test_util.py
kilroyd: Sorry, the comment is incorrect - I clearly wasn't thinking too hard. Given it's name and the documentation I found online, that's what I'd expect it to indicate.
We do actually see this defined on MacOS when either `stddef.h` or `cstddef` are included. I get:
    #define __APPLE_CC__ 6000
    #define __APPLE__ 1
Actually, if I pipe "" into IWYU, I still get these defines, so I think this must be comming from clang's Darwin platform definitions?
Happy to take suggestions for what to comment here. For now I think I'll remove...

iwyu_test_util.py
kilroyd: I think `msvcrt` would probably be most understood even if it's actually the UCRT. I'll go with that for now

For the C++ library MS seem to just use "STL" or on their Github "MSVC implementation of C++ STL". Are you ok to stick with `msvc` there?

iwyu_test_util.py
kilroyd: I'm happy to expose some preprocessor checks. Maybe something like
* `file-defines("<header>", "BLAH")` - 2 args though
* `has-predefined-macro("BLAH")`
Just thinking having a think how they read with IWYU_UNSUPPORTED and IWYU_REQUIRED...

    IWYU_REQUIRED: file-defines("abc.h", "FOO")
    IWYU_REQUIRED: has-predefined-macro("BAR")
    IWYU_UNSUPPORTED: file-defines("abd.c", "FOO")
    IWYU_UNSUPPORTED: has-predefined-macro("BAR")
    IWYU_XFAIL: file-defines("abc.h", "FOO")
    IWYU_XFAIL: has-predefined-macro("BAR")

I think we still want to encapsulate the standard library prvenance checks. `active-...stdlib` sounds reasonable.

iwyu_test_util.py
kimgr: Yeah, it seems like if the target is any Darwin platform, this will be defined: https://github.com/llvm/llvm-project/blob/main/clang/lib/Basic/Targets/OSTargets.cpp#L25 and https://github.com/search?q=repo%3Allvm%2Fllvm-project%20getDarwinDefines&type=code.

I guess the comment could say: `# Assume Apple libc for any Darwin platform`... or is that too obtuse?

kimgr: Hmm, it looks like `DARWIN_API_VERSION` might be a good detector for Apple libc: https://github.com/apple-open-source-mirror/Libc/blob/5e566be7a7047360adfb35ffc44c6a019a854bea/os/api.h#L74.

Haven't had a chance to test it on an Apple system, but it looks promising.

iwyu_test_util.py
kimgr: Yes, that sounds good for the C++ standard library.

iwyu_test_util.py
kilroyd: Just checked and I'm not seeing `*DARWIN*` being defined. Looking at the header files on my machine I thought I should be getting `__DARWIN_C_LEVEL` via `sys/cdefs.h`, but I'm not seeing that either (with IWYU or Xcode clang). Not quite sure what's going on.

iwyu_test_util.py
```diff
@@ -73,12 +74,69 @@ class Features:
...
+  # Parses the output of 'include-what-you-use -dM -E'.
+  PREPROCESSOR_MACRO_RE = re.compile(r'^#define ([A-z_][A-z0-9_]*)')
+
   def __init__(self):
     # Maps all known features names to prerequisite predicates.
     self.known = {
-      'has-target': Features.HasTarget
+      'has-target': Features.HasTarget,
+      'has-c-predefined-macro': Features.HasCPredefinedMacro,
+      'has-cxx-predefined-macro': Features.HasCXXPredefinedMacro,
```
kimgr, line 85: Let's not add these `has-...predefined-macro` features in this PR, and focus on one feature at a time.

iwyu_test_util.py
```diff
@@ -92,6 +150,43 @@ def HasTarget(target):
...
+    }
+    macro = id_macro.get(lib)
+    if macro is None:
+      raise SyntaxError('Unknown C++ standard library: %s' % lib)
+
+    return macro in Features._GetCXXPreprocessorMacros()
+
+  @staticmethod
+  def ActiveCstdlib(lib):
+    id_macro = {
```
kimgr, line 177: Do you actually need the libc detection for anything upcoming? Otherwise I'm inclined to either not do it at all for now, or at least just support the particular libraries that we need (uclibc and newlib looks a little esoteric at this point).

iwyu_test_util.py
kimgr: I'm on the fence about `active-`, I mostly suggested it to stimulate other ideas :).

I think `selected-` might look nice. Or to rhyme better with `has-target`, how about `uses-`?

iwyu_test_util.py
kilroyd: I was thinking the same myself.

With the test generation I'm looking at, I'm starting with C headers (to keep things simple), so probably want macos+glibc working. Newlib was actually simple to double check. If you could verify that we're correctly identifying glibc on a linux host that'd be great

iwyu_test_util.py
kilroyd: Haven't been able to look at this for a while. I've updated the latest commits with what I've found running on MacOS. Using `__APPLE__` and `__MACH__` appears to be the best option. Would be good to verify that this code identifies `glibc` on a Linux host

kilroyd: Hmmm. Looks like switching to a header like `<stdio.h>` allows us to see the system definitions in more cases

kimgr (approved):
Sorry, I lost track of this. Looks great, thanks!

Can you squash the branch and write up a proper commit message?

PR #1323, opened 2023-10-29
Fix handling aliased indirect types

bolshakov-a:
In some cases, IWYU produced redundant usage reports of type alias underlying types when the direct type under alias is a pointer or a reference. The reason is that the type alias uses were handled in `ReportTypeUse` uniformly regardless of context, and context-dependent dereferencing was in specific `Visit...` functions, but that dereferencing was not applied to type aliases (`typedef`s or C++11 type aliases). Indeed, it should not be applied to them in those functions because `ReportTypeUse` is the central place to analyze type aliases (with respect to underlying type "provision" status). But
the handling should be different in various contexts. E. g., when an alias hides a pointer:
```cpp
typedef Class* ClassPtr;
```
a use of such an alias in a variable declaration:
```cpp
ClassPtr object_ptr;
```
doesn't constitute full type use, whereas a use in a `catch(ClassPtr)` statement does.

Three cases have been found:
* direct use (doesn't require complete pointed-to or referenced type; e. g. in variable declarations; `UseKind::Direct`),
* referenced type use (e.g. in `sizeof` expressions; references are stripped before reporting; `UseKind::RemoveRefs`), and
* pointed-to type use (e.g. in array subscript expressions; at most one pointer and, possibly, references are stripped before reporting; `UseKind::RemoveRefsAndPtr`).

Hence, `UseKind` parameter is introduced into `ReportTypeUse` interface. Pointers and references are stripped from type alias underlying type during recursive calls of `ReportTypeUseInternal` with respect to the value of that parameter. Explicit pointer or refererence removals are removed (sorry for pun) from specific `Visit...` functions.

`VisitCXXDeleteExpr` function checked previously that the pointed-to type isn't a pointer itself. It isn't needed anymore as well, because `ReportTypeUse` performs at most one pointer removal.

In `InstantiatedTemplateVisitor::ReportTypeUse`, a slightly simplified `UseKind` handling logic is used because there is seemingly no need to handle type aliases (`resugar_map_` and `blocked_types_` define already which types can be reported and which should not due to their "provision" status). But that code (despite it duplicates base class logic somehow) is still needed: without it, the `type` may not be found in `resugar_map_`.

The changes in the `binary_type_trait` test look reasonable: full type is not needed to claim that pointers to pointers to different types aren't convertible to each other, even when the types are related by inheritance.

Determination of expression type is improved (`GetTypeOf` and related changes). But there is one suspicious thing: range-based `for` initializer expression type was checked for `nullptr` previously. I don't know whether that `nullptr` happens in real life or not.

Because C++11 type aliases and old C-style `typedef`s are handled by the same code, this PR is a step towards an important milestone: the full C89 support!

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -251,6 +251,8 @@ bool CanIgnoreLocation(SourceLocation loc) {
 
 }  // anonymous namespace
 
+enum class DerefKind { None, RemoveRefs, RemoveRefsAndPtr };
+
 // ----------------------------------------------------------------------
 // --- BaseAstVisitor
 // ----------------------------------------------------------------------
@@ -1570,13 +1572,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
   // Called when the given type is fully used at used_loc, regardless
   // of the type being explicitly written in the source code or not.
-  virtual void ReportTypeUse(SourceLocation used_loc, const Type* type) {
-    ReportTypeUseInternal(used_loc, type, blocked_types_);
+  virtual void ReportTypeUse(SourceLocation used_loc, const Type* type,
+                             DerefKind deref_kind) {
+    ReportTypeUseInternal(used_loc, type, blocked_types_, deref_kind);
   }
 
   void ReportTypesUse(SourceLocation used_loc, const set<const Type*>& types) {
     for (const Type* type : types)
-      ReportTypeUse(used_loc, type);
+      ReportTypeUse(used_loc, type, DerefKind::None);
   }
 
   //------------------------------------------------------------
@@ -1677,15 +1680,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       return true;
 
     if (HasCovariantReturnType(method_decl)) {
-      const Type* return_type = RemovePointersAndReferencesAsWritten(
-          method_decl->getReturnType().getTypePtr());
-
       VERRS(6) << "Found covariant return type in "
                << method_decl->getQualifiedNameAsString()
-               << ", needs complete type of " << PrintableType(return_type)
-               << "\n";
+               << ", needs complete type\n";
 
-      ReportTypeUse(CurrentLoc(), return_type);
+      ReportTypeUse(CurrentLoc(), method_decl->getReturnType().getTypePtr(),
+                    DerefKind::RemoveRefsAndPtr);
     }
 
     return true;
@@ -1707,8 +1707,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       const Type* caught_type = typeloc.getType().getTypePtr();
 
       // Strip off pointers/references to get to the pointee type.
-      caught_type = RemovePointersAndReferencesAsWritten(caught_type);
-      ReportTypeUse(GetLocation(&typeloc), caught_type);
+      ReportTypeUse(GetLocation(&typeloc), caught_type,
+                    DerefKind::RemoveRefsAndPtr);
     } else {
       // catch(...): no type to act on here.
     }
@@ -1726,7 +1726,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       return true;
 
     if (const Type* type = stmt->getRangeInit()->getType().getTypePtrOrNull()) {
-      ReportTypeUse(CurrentLoc(), RemovePointersAndReferencesAsWritten(type));
+      ReportTypeUse(CurrentLoc(), type, DerefKind::RemoveRefs);
 
       // TODO: We should probably find a way to require inclusion of any
       // argument-dependent begin/end declarations.
@@ -1881,12 +1881,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         break;
     }
 
-    // TODO(csilvers): test if we correctly say we use FooPtr for
-    //    typedef Foo* FooPtr; ... static_cast<FooPtr>(...) ...
-    for (const Type* type : required_full_types) {
-      const Type* deref_type = RemovePointersAndReferences(type);
-      ReportTypeUse(CurrentLoc(), deref_type);
-    }
+    for (const Type* type : required_full_types)
+      ReportTypeUse(CurrentLoc(), type, DerefKind::RemoveRefsAndPtr);
 
     return true;
   }
@@ -1902,8 +1898,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     const Expr* base_expr = expr->getBase()->IgnoreParenImpCasts();
     const Type* base_type = GetTypeOf(base_expr);
     CHECK_(base_type && "Member's base does not have a type?");
-    const Type* deref_base_type  // For myvar->a, base-type will have a *
-        = expr->isArrow() ? RemovePointerFromType(base_type) : base_type;
     // Technically, we should say the type is being used at the
     // location of base_expr.  That may be a different file than us in
     // cases like MACRO.b().  However, while one can imagine
@@ -1914,9 +1908,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // that's responsible for its type or not, we just assume it is.
     // TODO(csilvers): fix when we can determine what the macro-text
     // is responsible for and what we're responsible for.
-    // TODO(csilvers): we should be reporting a fwd-decl use for
-    // GetTypeOf(expr), not on deref_base_type.
-    ReportTypeUse(CurrentLoc(), deref_base_type);
+    ReportTypeUse(CurrentLoc(), base_type, DerefKind::RemoveRefsAndPtr);
     return true;
   }
 
@@ -1926,7 +1918,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreCurrentASTNode())
       return true;
 
-    ReportTypeUse(CurrentLoc(), GetTypeOf(expr));
+    ReportTypeUse(CurrentLoc(), GetTypeOf(expr->getLHS()),
+                  DerefKind::RemoveRefsAndPtr);
     return true;
   }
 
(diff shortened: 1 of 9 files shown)


iwyu.cc
```diff
@@ -250,6 +250,8 @@ bool CanIgnoreLocation(SourceLocation loc) {
 
 }  // anonymous namespace
 
+enum class UseKind { Direct, RemoveRefs, RemoveRefsAndPtr };
```
kimgr, line 253: `UseKind` is a little open-ended, but I like it because it's short. Maybe `DerefKind` could work?

iwyu.cc
```diff
@@ -1676,15 +1679,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      const Type* return_type = method_decl->getReturnType().getTypePtr();
+      const Type* deref_return_type =
+          RemovePointersAndReferencesAsWritten(return_type);
 
       VERRS(6) << "Found covariant return type in "
                << method_decl->getQualifiedNameAsString()
-               << ", needs complete type of " << PrintableType(return_type)
-               << "\n";
+               << ", needs complete type of "
+               << PrintableType(deref_return_type) << "\n";
```
kimgr, line 1689: We could also remove the type name, since it will be reported by `ReportTypeUse` at loglevel 3, right?

Something like `Found covariant return type in ..., needs complete type\n"`

iwyu.cc
```diff
@@ -2041,9 +2036,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // We always delete a pointer, so do one dereference to get the
     // actual type being deleted.
     const Type* delete_ptr_type = GetTypeOf(delete_arg);
-    const Type* delete_type = RemovePointerFromType(delete_ptr_type);
-    if (delete_type && !IsPointerOrReferenceAsWritten(delete_type))
-      ReportTypeUse(CurrentLoc(), delete_type);
+    ReportTypeUse(CurrentLoc(), delete_ptr_type, UseKind::RemoveRefsAndPtr);
```
kimgr, line 2038: Curious: why `RemoveRefsAndPtr` here? Is to cover scenarios like:
```
T* p = new T;
T& r = *p;
delete &r;
```
?

iwyu.cc
```diff
@@ -2560,7 +2553,24 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   }
 
   void ReportTypeUseInternal(SourceLocation used_loc, const Type* type,
-                             const set<const Type*>& blocked_types) {
+                             const set<const Type*>& blocked_types,
+                             UseKind use_kind) {
+    type = Desugar(type);
+    if (use_kind == UseKind::RemoveRefs ||
+        use_kind == UseKind::RemoveRefsAndPtr) {
+      if (const auto* ref_type = dyn_cast_or_null<ReferenceType>(type))
```
kimgr, line 2560: I'm guessing this cast is necessary (rather than `type->isReferenceType()`) because you want to `getPointeeTypeAsWritten()`?

....

iwyu.cc
```diff
@@ -2560,7 +2553,24 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        type = ref_type->getPointeeTypeAsWritten().getTypePtr();
+      // Keep use_kind to remove possible additional references in type alias
+      // chain.
+      // TODO(bolshakov): shouldn't Desugar be called here and below as well?
+    }
+    if (use_kind == UseKind::RemoveRefsAndPtr) {
+      if (const auto* ptr_type = dyn_cast_or_null<PointerType>(type)) {
```
kimgr, line 2567: ... but here you should be able to get away with `type->isPointerType()`. But maybe you want to keep them both as casts for symmetry?

Ah, also `dyn_cast_or_null`. Yeah, nevermind.

iwyu.cc
```diff
@@ -2906,7 +2915,18 @@ class InstantiatedTemplateVisitor
     // implicit stuff, it should not suggest forward declarations.
   }
 
-  void ReportTypeUse(SourceLocation used_loc, const Type* type) override {
+  void ReportTypeUse(SourceLocation used_loc, const Type* type,
+                     UseKind use_kind) override {
+    if (use_kind == UseKind::RemoveRefs ||
```
kimgr, line 2919: `type` can never be nullptr here, I guess? I think `Resugar` and `CanIgnoreType` are both null-friendly.

tests/cxx/ptr_ref_aliases.cc
```diff
@@ -0,0 +1,182 @@
...
+
+void Fn() {
+  (void)sizeof(NonProvidingPtrAlias);
+  (void)sizeof(ptr_nonprovided);
+  // IWYU: Indirect is...*ptr_ref_aliases-i2.h
+  (void)sizeof(NonProvidingRefAlias);
+  // IWYU: Indirect is...*ptr_ref_aliases-i2.h
+  (void)sizeof(ref_nonprovided);
+  (void)sizeof(ProvidingRefAlias);
+  // TODO: avoid redundant reporting.
```
kimgr: Here and below -- what do you mean by "redundant" here?

It seems more like a proper bug; "redundant" makes me think it's an inconvenience and that the resulting IWYU suggestions are still correct. Could you maybe clarify these comments a little?

iwyu.cc
bolshakov-a: `Direct` should probably be replaced with `DerefKind::None` in that case?

iwyu.cc
bolshakov-a: It's rather for:
```cpp
T* p = new T;
T*& r = p;
delete r;
```
And there may be additional references through type aliases.

iwyu.cc
bolshakov-a: Both `isReferenceType()` and `isPointerType()` check canonical type directly, through all the `TypedefType`s that may occur in the sugar chain. It should be avoided here because `TypedefType`s should be "caught" and handled here (in contrast to `InstantiatedTemplateVisitor` implementation of `ReportTypeUse`).

tests/cxx/ptr_ref_aliases.cc
bolshakov-a: "Unwanted" would be better? Or just "false-positive"? But these comments are removed in the second commit.

kimgr: Aha, I missed that 🙂. In that case it doesn't matter so much, but "unwanted" definitely seems better!

iwyu.cc
kimgr: Sounds good!

iwyu.cc
kimgr: Aah, subtle difference. Could you add an educational comment about that here?

iwyu.cc
kimgr: ... so you may want to avoid `type->` before you know it's non-null.

iwyu.cc
bolshakov-a: > since it will be reported by ReportTypeUse at loglevel 3, right

Only when the return type is not defined in the same file and not properly `#include`d, but I agree that such a verbose diagnostic doesn't seem very useful.

iwyu.cc
bolshakov-a: Let `ReportTypeUse` be null-friendly.

iwyu.cc
```diff
@@ -1570,13 +1572,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-    ReportTypeUseInternal(used_loc, type, blocked_types_);
+  virtual void ReportTypeUse(SourceLocation used_loc, const Type* type,
+                             DerefKind deref_kind) {
+    ReportTypeUseInternal(used_loc, type, blocked_types_, deref_kind);
   }
 
   void ReportTypesUse(SourceLocation used_loc, const set<const Type*>& types) {
     for (const Type* type : types)
-      ReportTypeUse(used_loc, type);
+      ReportTypeUse(used_loc, type, DerefKind::None);
```
kimgr, line 1582: Just out of curiosity, could this ever need per-type deref kind? Or are batch-reported types always non-ptr/ref by some implicit rule?

kimgr: (definitely not saying we should do anything about it now, just curious)

tests/cxx/ptr_ref_aliases.cc
```diff
@@ -0,0 +1,170 @@
...
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --check_also=tests/cxx/ptr_ref_aliases-d1.h \
+//            -Xiwyu --check_also=tests/cxx/ptr_ref_aliases-d2.h \
+//            -I .
+
+// Tests various cases of type alias usages when the underlying type is
+// a pointer or a reference. Particularly, tests that `UseKind` parameter value
```
kimgr, line 15: `DerefKind` here now

iwyu.cc
bolshakov-a: Yes, an implicit rule that `ReportTypesUse` is called for reporting instantiated template cached types which are stored already dereferenced according to original `DerefKind`s.

But it seems for me quite natural that `DerefKind` is `None` when is not explicitly specified. If ever needed, `ReportTypesUse` interface should be changed to accept `DerefKind` for all the types or for every type.

kimgr: Makes sense, thanks!

tests/cxx/ptr_ref_aliases.cc
kimgr: Oh, :cry: . The commit message of https://github.com/include-what-you-use/include-what-you-use/pull/1323/commits/180f18da7e1d4b3db03a0aa31cee34bdd1a93b4f also refers to `UseKind` throughout. I'm so sorry I didn't notice that earlier. Could you rewrite it with `DerefKind`, please?

bolshakov-a: Of course, no problem at all! Thanks for attentiveness!

kimgr: Thank you!

PR #1324, opened 2023-11-03
Fix build after upstream change

bolshakov-a:
Breaking commit:
https://github.com/llvm/llvm-project/commit/4ad2ada5216ee2bb3c334a3233a9

No functional change.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -166,6 +166,7 @@ using clang::Decl;
 using clang::DeclContext;
 using clang::DeclRefExpr;
 using clang::DeducedTemplateSpecializationType;
+using clang::ElaboratedTypeKeyword;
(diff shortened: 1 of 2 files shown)


kimgr:
Thanks! I use a fairly particular format for these kinds of clang-compat commits, so rather than bounce wording back and forth, I reworded and merged: https://github.com/include-what-you-use/include-what-you-use/commit/d459ef4582981500bf75d34df751409ed6668262.

kimgr:
This is also the first Clang change that breaks compatibility with Clang 17 from IWYU's point of view. So now the iwyu-0.21 release needs some priority :)

bolshakov-a:
Ok, got it.

PR #1325, opened 2023-11-05
Analyze class template bases

bolshakov-a:
Previously, class template bases were handled on definitions of template specialization type variables due to the traversal of member initializer list of (probably implicit) constructor. It didn't work for other contexts, such as taking `sizeof` template specialization type.


kimgr:
Nice, thanks!

PR #1326, opened 2023-11-05
Update precomputed STL type list

bolshakov-a:
C++17 doesn't require full contained type info for `std::vector`, `std::list`, and `std::forward_list` instantiations (see [N4510](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4510.html)). The full info is needed only when any of container class members is referenced.

Some more STL containers have been added to precomputed list. Unordered associative containers need contained type to be complete at instantiation point. `std::forward_list` - only before C++17.

### DIFF
--- a/iwyu_cache.cc
+++ b/iwyu_cache.cc
@@ -17,8 +17,10 @@
 #include "clang/AST/DeclTemplate.h"
 #include "clang/AST/TemplateBase.h"
 #include "clang/AST/Type.h"
+#include "clang/Basic/LangOptions.h"
 
 using clang::ClassTemplateSpecializationDecl;
+using clang::LangOptions;
 using clang::NamedDecl;
 using clang::Type;
 using clang::TemplateArgument;
@@ -41,13 +43,11 @@ static const char* const kFullUseTypes[] = {
     "__gnu_cxx::hash_multiset",
     "__gnu_cxx::hash_set",
     "std::deque",
-    "std::list",
     "std::map",
     "std::multimap",
     "std::multiset",
     "std::set",
     "std::slist",
-    "std::vector",
 };
 
 // If the passed-in tpl_decl is one of the classes we have hard-coded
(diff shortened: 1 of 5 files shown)


tests/cxx/precomputed_tpl_args.cc
```diff
@@ -11,6 +11,8 @@
 
 // Tests the precomputed template-arg-use list in iwyu_cache.cc.
 
+#include <forward_list>
```
kimgr, line 14: `<forward_list>` seems a little confused about which commit it should be part of... C++17 special handling for it is introduced in the next commit, but the testcase is added here.

iwyu_cache.cc
```diff
@@ -64,7 +68,7 @@ map<const Type*, const Type*> FullUseCache::GetPrecomputedResugarMap(
       (sizeof(kFullUseTypes) / sizeof(*kFullUseTypes));
   set<string> fulluse_types(kFullUseTypes, kFullUseTypes + fulluse_size);
   if (!lang_opts.CPlusPlus17)
-    fulluse_types.insert({"std::list", "std::vector"});
+    fulluse_types.insert({"std::list", "std::forward_list", "std::vector"});
```
kimgr, line 71: Alphabetical order here would be nice.

iwyu_cache.cc
```diff
@@ -47,6 +47,10 @@ static const char* const kFullUseTypes[] = {
     "std::multimap",
     "std::multiset",
     "std::set",
+    "std::unordered_map",
```
kimgr, line 50: Oh, and alphaorder here too, please

tests/cxx/precomputed_tpl_args.cc
bolshakov-a: It was not in the precomputed list at all, hence forward-declaration suggestions for contained type already worked for it. The second commit adds it to the list for pre-C++17 modes, and this test works in the default (C++17 currently) language standard.

kimgr: I see. It would look more symmetrical and balanced if you moved the `<forward_list>` testing together with the `<forward_list>` special-casing for pre-C++17, though, could you please transfer it to the [Add some more STL containers to precomputed list](https://github.com/include-what-you-use/include-what-you-use/pull/1326/commits/60a4ea0a3c3be1915519f17018551e3eaa7006cc) commit?

bolshakov-a: Ok, sure.

kimgr:
Oops, I now have failures on master:
```
FAIL: test_precomputed_tpl_args (__main__.cxx)
  File "/home/kimgr/code/iwyu/include-what-you-use/./run_iwyu_tests.py", line 88, in <lambda>
  File "/home/kimgr/code/iwyu/include-what-you-use/iwyu_test_util.py", line 600, in TestIwyuOnRelativeFile
AssertionError: 
tests/cxx/precomputed_tpl_args.cc:37: Unexpected diagnostic:
IndirectClass is defined in "tests/cxx/precomputed_tpl_args-i1.h", which isn't directly #included.

tests/cxx/precomputed_tpl_args.cc:91: Unexpected diagnostic:
IndirectClass is defined in "tests/cxx/precomputed_tpl_args-i1.h", which isn't directly #included.

tests/cxx/precomputed_tpl_args.cc:93: Unexpected diagnostic:
IndirectClass is defined in "tests/cxx/precomputed_tpl_args-i1.h", which isn't directly #included.

tests/cxx/precomputed_tpl_args.cc:95: Unexpected diagnostic:
IndirectClass is defined in "tests/cxx/precomputed_tpl_args-i1.h", which isn't directly #included.
```

It seems like some kind of interplay between your recent PRs. I'm looking for the offending commit.

PR #1328, opened 2023-11-07
Avoid unwanted primary template decl report due to explicit instantiation

bolshakov-a:
Instantiated template use is already reported from `VisitTemplateSpecializationType`. The removed code just didn't work as expected because clang attributes explicit instantiation `ClassTemplateSpecializationDecl`s to the template (primary or partial spec, whatever is instantiated) location (which is usually in a header) and not to the location where the explicit instantiation is requested. I.e., `ClassTemplateSpecializationDecl` node source location is not what was expected, hence when an explicit instantiation instantiated the primary template, the primary template declaration was reported to its own location (i.e. IWYU did nothing, essentially).

But even worse, when an explicit instantiation instantiated a partial specialization, the primary template was reported to the partial specialization location, where its fwd-declaration is sufficient. Due to this, an explicit instantiation in one file could cause an erroneous suggestion to `#include` another file to a third file!

In fact, it is questionable whether a type use should be reported on visitation of type node or declaration node. But even if `VisitTemplateSpecializationType` stops reporting full type uses one day, `VisitClassTemplateSpecializationDecl` should better call `ReportTypeUse` and not `ReportDeclUse` in order to point to the correct specialization (if any) and to report all the types needed for template type instantiation.

The less code - the less bugs!


kimgr:
Looks good to me!

PR #1332, opened 2023-11-09
Correct traversal of class template explicit instantiations

bolshakov-a:
Clang attributes `ClassTemplateSpecializationDecl`s of explicit instantiations to the instantiated template definition location. Traversing them may cause reporting of template arguments incorrectly to that location (which is usually in a header). Hence, the standard traversal of explicitly instantiated specializations is avoided.

`TraverseClassTemplateSpecializationDecl` handler was added in #774 so as to analyze explicit specializations. It was seemingly an upstream issue fixed in [that commit](https://github.com/llvm/llvm-project/commit/c119a17e7fd6fb528d8c1e1bbb0a). At now, default `TraverseClassTemplateSpecializationDecl` implementation works well: traverses explicit specialization definition, but only the name in case of an implicit specialization.

`TraverseCXXRecordDecl` is directly called on `ClassTemplateSpecializationDecl` in `AnalyzeTemplateTypeParmUse` to handle implicit specialization which may be needed on instantiation of some other template (when the former doesn't occur to be written as `TemplateSpecializationType` in the initial template arguments, or in the template body). `alias_template.cc` test showed relating regression.

`Inner` class template isn't very much needed for the test purpose (a field of type `T` may be sufficient in `Template`), but it is added to assure that the correct instantiated template visitation logic is used.

Explicit instantiation definitions cause instantiation of class template member functions. This PR adds scanning them in the usual way, taking into account type template argument provision status.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3976,16 +3976,36 @@ class IwyuAstConsumer
   // we don't want iwyu to recommend removing the 'forward declare' of Foo.
   //
   // Additionally, this type of decl is also used to represent explicit template
-  // instantiations, in which case we want the full type, not only a forward
-  // declaration. But it is reported from VisitTemplateSpecializationType, hence
-  // may be ignored here.
+  // instantiations. Only instantiation definitions causing instantiation of
+  // member functions are handled here. Other instantiated parts of the template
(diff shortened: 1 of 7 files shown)


kimgr:
Thanks!

Issue #1147, opened 2022-11-23
IWYU wants to remove includes for types that are needed in template default values


honkstar1:
Here's an example:

```
#include "BaseClass.h"

template<typename TheSizeType>
struct TSizeLargerThan4
{
	static inline constexpr bool TheValue = sizeof(TheSizeType) > 4;
};

template <typename TheType, bool bIsSizeLarger = TSizeLargerThan4<TheType>::TheValue>
struct TGetCastType2
{
	static inline constexpr bool SizeValue = bIsSizeLarger;
};

void myfunc()
{
	auto v = TGetCastType2<BaseClass>::SizeValue;
	(void)v;
}

```

In the above code IWYU wants to turn BaseClass into a forward declaration (but it is needed for the sizeof())

Trying to figure out where this code lives but again, happy for pointers.

honkstar1:
seems like the SubstNonTypeTemplateParmExpr is not implemented at all? I know nothing about llvm and clang implementation but I added a visitor for this and log it out and I see the sizeof(TheType) in there.. trying to figure out how the non-logging path is handling but could need some help :-)

kimgr:
Templates are pretty tricky in IWYU, partly, I think, because its AST visitor predates the `shouldVisitTemplateInstantiations` in Clang. I don't have a clear idea either exactly how it works :).

But my high-level understanding is that when IWYU finds a template instantiation, it starts a new AST traversal (`InstantiatedTemplateASTVisitor`), which visits all types in the instantiation and tries to map them back to the actually-written ("sugared") types provided as template arguments (I'm probably slaughtering this completely).

AFAIK, the handling of default template arguments is in the `GetTplTypeResugarMapForClass` family of functions, so you might want to start there.

honkstar1:
I think I kind of know where the problem is.. but don't know how to solve it.
The InstantiatedTemplateASTVisitor need to also go in to the default expression of the NonTypeTemplateParmDecl found in SubstNonTypeTemplateParmExpr.

So I've added TraverseSubstNonTypeTemplateParmExpr and get the call.. trying to figure out how to have InstantiatedTemplateASTVisitor get past the stmt and in to the UnaryExprOrTypeTraitExpr that is underneath..

kimgr:
I can see that that would make sense as a working theory. I can't say whether it's the right one :)

[...]

PR #1333, opened 2023-11-09
Quick fix of crash on scanning noninstantiated implicit specialization

### LINKED ISSUE #1331: iwyu 0.21 crashes with assertion failure
IWYU version: `include-what-you-use 0.21 (git:dbecade) based on Ubuntu clang version 17.0.4 (++20231031083134+309d55140c46-1~exp1~20231031083240.62)`

IWYU crashes with an assertion when using [foonathan/lexy](https://github.com/foonathan/lexy) library.

Assertion:
>Error running '/usr/local/bin/include-what-you-use': include-what-you-use: /usr/lib/llvm-17/include/clang/AST/DeclCXX.h:452: clang::CXXRecordDecl::DefinitionData& clang::CXXRecordDecl::data() const: Assertion `DD && "queried property of class with no definition"' failed.
>Subprocess aborted

main.cpp:
```
#include <lexy/dsl/literal.hpp>
#include <lexy/dsl/choice.hpp>

constexpr auto separator = lexy::dsl::lit_c<'a'> | lexy::dsl::lit_c<'b'>;

int main()
{
}
```

CMakeLists.txt
```
cmake_minimum_required(VERSION 3.24)

project(lexy-example)

include(FetchContent)
FetchContent_Declare(
    lexy
    GIT_REPOSITORY https://github.com/foonathan/lexy.git
    GIT_TAG f68737b725116d00e5582602e22604a14fc26547
)
FetchContent_MakeAvailable(lexy)

find_program(IWYU_EXE NAMES include-what-you-use iwyu REQUIRED)

set(CMAKE_CXX_STANDARD 20)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)

add_executable(lexy_example main.cpp)
target_link_libraries(lexy_example PRIVATE foonathan::lexy)

set_property(
    TARGET
        lexy_example
    PROPERTY 
        CXX_INCLUDE_WHAT_YOU_USE ${IWYU_EXE})

```

Compiled with `CC=clang-17 CXX=clang++-17 cmake -S . -B build && cmake --build build`

So far I've been able to narrow down the issue that it's coming using `operator|` from `lexy/dsl/choice.hpp`. I ran into this when switching IWYU from version 0.20 to 0.21, so I believe this was introduced between those versions.

bolshakov-a:
This fixes #1331.

The crash had been introduced in #1325, because there is an assert in clang that the `bases()` method should be called only on class declarations having definitions. But clang doesn't instantiate definition of class template implicit specialization until that specialization is fully used.

Maybe, the root of the problem is that IWYU tries to provide the full type info when it is not actually needed, specifically, in default type template arguments. Not sure what the correct behavior should be. Currently, this goal isn't fully achieved because template internals aren't scanned when the specialization is not instantiated. Even worse, IWYU may give different suggestions to the header containing the template definition depending on the translation unit as part of which the header is analyzed.

For example, given:
```cpp
// Header.h
#include "Class.h"
template <typename T>
struct Tpl {
  T t;
};

template <typename T = Tpl<Class>>
struct Tpl2 {
  T t;
};

// Source.cpp
#include "Header.h"
Tpl2<> tpl;
```
IWYU suggests to keep `#include "Class.h"` in `Header.h` when analyzed along with `Source.cpp` but suggests to remove it when the `Header.h` is analyzed alone.

Switching to forward-declarations of template arguments can be easily achieved with adding the following lines into `IwyuBaseAstVisitor`:
```cpp
bool VisitTemplateTypeParmDecl(clang::TemplateTypeParmDecl* decl) {
  current_ast_node()->set_in_forward_declare_context(true);
  return true;
}
```
but there are two test cases having contradictory comments. [The first one](https://github.com/include-what-you-use/include-what-you-use/blob/0.21/tests/cxx/badinc.cc#L297) is a TODO stating that template arguments are fwd-declarable. [The second one](https://github.com/include-what-you-use/include-what-you-use/blob/0.21/tests/cxx/badinc.cc#L1003) claims that requiring full types for them is by-design. And, of course, there may be more cases when IWYU tries to scan uninstantiated specializations, when the full type is not really needed. @kimgr, wdyt?

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3476,6 +3476,12 @@ class InstantiatedTemplateVisitor
     if (ReplayClassMemberUsesFromPrecomputedList(type))
       return true;
 
+    // Sometimes, an implicit specialization occurs to be not instantiated.
+    // TODO(bolshakov): don't report them at all as full uses or figure out
+    // how to scan them.
+    if (!class_decl->hasDefinition())
+      return true;
+
     // Make sure all the types we report in the recursive TraverseDecl
     // calls, below, end up in the cache for class_decl.
     CacheStoringScope css(&cache_storers_, ClassMembersFullUseCache(),
--- a/tests/cxx/default_tpl_arg-d1.h
+++ b/tests/cxx/default_tpl_arg-d1.h
@@ -1,0 +1,10 @@
+//===--- default_tpl_arg-d1.h - test input file for iwyu ------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/default_tpl_arg-i1.h"
--- a/tests/cxx/default_tpl_arg-i1.h
+++ b/tests/cxx/default_tpl_arg-i1.h
@@ -1,0 +1,13 @@
+//===--- default_tpl_arg-i1.h - test input file for iwyu ------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+template <typename T>
+struct UninstantiatedTpl {
+  T t;
+};
--- a/tests/cxx/default_tpl_arg.cc
+++ b/tests/cxx/default_tpl_arg.cc
@@ -1,0 +1,41 @@
+//===--- default_tpl_arg.cc - test input file for iwyu --------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+// Tests handling default type template arguments. In particular, that IWYU
+// doesn't crash when they refer to uninstantiated template specializations.
+
+#include "tests/cxx/default_tpl_arg-d1.h"
+
+// IWYU: UninstantiatedTpl is...*default_tpl_arg-i1.h
+template <typename = UninstantiatedTpl<int>>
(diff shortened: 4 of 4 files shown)


bolshakov-a:
Probably, default type template arguments should be a customization point likewise typedef/fnreturn/autocast triad (i.e., the template author should be able to decide whether or not to "provide" the default argument type). If so, we should find a way to scan uninstantiated template specialization types in default template arguments.

It is worth noting that, nevertheless, there is no point to scan them during the process of scanning an outer template implicit specialization during traversal of a template user's code.

kimgr:
The comments in badinc.cc don't look very convinced/convincing. Does forward-declarability really hold for all `TemplateTypeParmDecl`s, though? Ah, those declarations only show up in template declarations themselves, OK, I think I see what the goal is.

I guess in some cases it will become really annoying, e.g.
```
template<class T, typename SequenceType=std::vector<T>>
struct Foo {
const SequenceType* blah;
};
```
I'm guessing the template declaration will not require `<vector>` with the forward-declare-always rule in place, but what will happen for people who instantiate `Foo<int>`? The compiler will need `std::vector<>` visible there, right? Will IWYU agree?

kimgr:
> Probably, default type template arguments should be a customization point likewise typedef/fnreturn/autocast triad 

Sounds right.

I just remembered this discussion from ~ a year ago: https://github.com/include-what-you-use/include-what-you-use/issues/1147. @honkstar1's examples all seem to be instantiation-side misses, though.

bolshakov-a:
> The compiler will need `std::vector<>` visible there, right? Will IWYU agree?

At the template instantiation side, there already is a check of whether default template argument is provided or not: https://github.com/include-what-you-use/include-what-you-use/blob/0.21/iwyu.cc#L2847

bolshakov-a:
Heh, it turns out that me [fixed it already](https://github.com/bolshakov-a/include-what-you-use/commit/1856a81e6f6f10b84b01f0d60a8a04f0943e2520) but forgot about it completely before making #1325. My bad.

PR #1338, opened 2023-11-11
Don't require full type for '__is_same' arguments

bolshakov-a:
Full type definitions are not needed to decide whether two types are the same or not.

This fixes the 'precomputed_tpl_args' test that failed on the master branch due to interplay between #1323 and #1326.


kimgr:
Super-fast, thanks!

bolshakov-a:
> Super-fast, thanks!

In fact, this commit is more than one-year old:
https://github.com/bolshakov-a/include-what-you-use/commit/29ecc272c91fbbcc5c6d9f8ad46f8d13b5d435fb

PR #1339, opened 2023-11-11
Move frontend execution into iwyu_driver

kimgr:
This keeps all the setup code in a single place, so we can make smarter choices for globals, frontend actions, etc.

No functional change.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -4388,7 +4388,7 @@ class IwyuAction : public ASTFrontendAction {
 
 using include_what_you_use::OptionsParser;
 using include_what_you_use::IwyuAction;
-using include_what_you_use::CreateCompilerInstance;
+using include_what_you_use::ExecuteAction;
 
 int main(int argc, char **argv) {
   llvm::llvm_shutdown_obj scoped_shutdown;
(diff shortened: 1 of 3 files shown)


kimgr:
@kilroyd First step toward `include-what-you-use` respecting `-E` (and other goodies!)

kimgr:
Added another commit to inline `CreateCompilerInstance`, since I had it finalized anyway.

iwyu.cc
```diff
@@ -4403,16 +4403,9 @@ int main(int argc, char **argv) {
   //   path/to/iwyu -Xiwyu --verbose=4 [-Xiwyu --other_iwyu_flag]... \
   //       CLANG_FLAGS... foo.cc
   OptionsParser options_parser(argc, argv);
-
-  std::unique_ptr<clang::CompilerInstance> compiler(CreateCompilerInstance(
-      options_parser.clang_argc(), options_parser.clang_argv()));
-  if (!compiler) {
+  if (!ExecuteAction(options_parser.clang_argc(), options_parser.clang_argv(),
+                     []() { return std::make_unique<IwyuAction>(); })) {
```
kilroyd, line 4407: Is the lambda function/factory here necessary? To me it seems easier and more obvious if we just pass `std::make_unique<IwyuAction>()` into `ExecuteAction`.

kilroyd (commented):
Looks fine to me. Just one observation

iwyu.cc
kimgr: I have a few PRs buffered where it will be necessary to delay construction:

* ... to be able to pass in ToolChain into IwyuAction constructor
* ... to sometimes avoid IwyuAction entirely in favor of a real preprocessing job

so yeah, this looks a little overengineered currently, but it will make the coming PRs easier to understand.

kilroyd: Ack.

kilroyd (approved):

PR #1340, opened 2023-11-12
Remove redundant type scan on ctor instantiation

bolshakov-a:
Implicitly instantiated template specialization scanning should be already performed in every context when the construction can take place. Eventually, every template member function call requires template specialization to be instantiated, and a constructor is just one of them.

No functional change.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3405,21 +3405,7 @@ class InstantiatedTemplateVisitor
     }
 
     // We need to iterate over the function.
-    if (!TraverseDecl(const_cast<FunctionDecl*>(fn_decl)))
-      return false;
-
-    // If we're a constructor, we also need to construct the entire class,
-    // even typedefs that aren't used at construct time. Try compiling
-    //    template<class T> struct C { typedef typename T::a t; };
-    //    class S; int main() { C<S> c; }
-    if (isa<CXXConstructorDecl>(fn_decl)) {
-      CHECK_(parent_type && "How can a constructor have no parent?");
-      parent_type = Desugar(parent_type);
-      if (!TraverseDataAndTypeMembersOfClassHelper(
-              dyn_cast<TemplateSpecializationType>(parent_type)))
-        return false;
-    }
-    return true;
+    return TraverseDecl(const_cast<FunctionDecl*>(fn_decl));
   }
 
   // Does the actual recursing over data members and type members of


kimgr:
I'm utterly lost. Removal without test failures always looks good, obviously, but how can this code now be redundant? Can you walk me through a small example?

kimgr:
Is this somehow obviated by improvements in RecursiveASTVisitor? Or in our own traversal?

bolshakov-a:
When a variable of a template specialization type is constructed, the internals of the specialization are usually handled in `IwyuAstConsumer::VisitTemplateSpecializationType`. But I've just found a case when it doesn't work:
```cpp
struct WithStatic {
  static Tpl<Indirect> s;
};

decltype(WithStatic::s) tpl;
```
The problem here is that `DecltypeType` is not currently handled by IWYU. It erroneously suggests to fwd-declare `Tpl` but at least to `#include` `Indirect` type defining header. This branch causes a regression in this case, suggesting fwd-decl even for `Indirect`. However, it is not a problem of the proposed changes, but lacking of handling all the types of sugar which may occur in variable declarations.

Probably, the PR should be postponed until the problem with `decltype` is fixed.

bolshakov-a:
How to fix the problem with `decltype` is another interesting question. I could add `VisitDecltypeType` next to `VisitTagType`, `VisitTypedefType`, `VisitUsingType` methods, but this doesn't seem a correct way. Instead of catching all the sugar and figuring out the context inside `CanForwardDeclareType` and `AnalyzeTemplateTypeParmUse`, maybe, just report the type from `VisitVarDecl` and other handlers of declarations and expressions requiring full type? `VisitTagType` would just report fwd-decls in that case, and `CanForwardDeclareType` would be probably removed. It looks like quite a substantial overhaul and would cause a lot of `needs a declaration` noise in the tests, but seems like it is worth.

kimgr:
Thanks! Might be nice to mention in the commit message that `IwyuAstConsumer::VisitTemplateSpecializationType` is what covers this. And do you think it's always done so, or have template improvements over the years made it redundant?

kimgr:
As for `decltype` handling, I couldn't say what the best way is. Your suggested overhaul sounds interesting, because it sounds like it's turning a lot of introspective checks ("walk up the tree to understand the context of this type use") into straightforward ones ("in the context of [e.g. var decls], we know that full type is required under these conditions").

But it's hard for me to visualize how it will work out.

> It looks like quite a substantial overhaul and would cause a lot of `needs a declaration` noise in the tests, but seems like it is worth.

Just for clarity, do you expect `needs a declaration` needed to be added or removed? I think of `needs a declaration` as "forward-declare required", and everywhere I see both `X needs a declaration` and `X is defined in.*` on the same line, I think it's a quality-of-implementation issue in our reporting. But maybe it's not so clear-cut?

bolshakov-a:
> And do you think it's always done so, or have template improvements over the years made it redundant?

All off this is very old code. Maybe folks just forgot to remove the redundant part, maybe reassurance, maybe I'm missing some cases.

bolshakov-a:
> Just for clarity, do you expect `needs a declaration` needed to be added or removed?

To be added. `VisitTagType` and `VisitTemplateSpecializationType` would call `ReportDeclForwardDeclareUse` almost unconditionally (leaving `CanForwardDeclareType` would cause duplication of logic). However, some people (like me) may launch IWYU with `--verbose=3` and inspect the reasons to `#include` or fwd-declare something. That change would have negative impact on their experience. I think `needs a declaration` lines near `is defined in` lines could be filtered out before the output, however.

kimgr:
I'm not too worried about `decltype`, but I never use it myself. Do you think we should merge this PR now, or wait until `decltype` works better?

bolshakov-a:
I'm not sure. I agree that my example looks extravagant, but `decltype` may be heavily used inside some library code. I just work on reporting types unrelated to template arguments from template instantiation, and the removed code looks hindering.

kimgr:
Yeah, modern C++ libraries appear to be pretty shameless :)

Let's wait with this, then. Can we add a little testcase instead, to prove to what extent `decltype` currently works?

bolshakov-a:
Added in #1346.

kimgr:
Love it, thanks!

kimgr:
> I could add VisitDecltypeType next to VisitTagType, VisitTypedefType, VisitUsingType methods, but this doesn't seem a correct way. Instead of catching all the sugar and figuring out the context inside CanForwardDeclareType and AnalyzeTemplateTypeParmUse, maybe, just report the type from VisitVarDecl and other handlers of declarations and expressions requiring full type?

In light of our recent discussions, would it be possible to mark types as handled on the `ASTNode`, something like `in_fwd_declare_context`, but more `ok_this_is_part_of_decl_and_we_dont_need_to_report_types_again_below`? Or just `std::set<const Type*> parent_reported_types`, which flows downwards from a node?

bolshakov-a:
Good idea, but on the other hand, it may introduce additional complexity instead of the one which I want to remove. Especially, in case like base classes:
```cpp
struct A : B {
  B* m_b;
};
```
Full use of `B` as the base class can be reported from `VisitCXXRecordDecl`, but setting it there as type with fwd-decls suppressed would cause that the fwd-decl warning would not be reported even for `m_b`. Graceful handling would require some change of `CXXRecordDecl` traversal.

kimgr:
I was thinking that could be handled via `VisitFieldDecl` in this case, on the assumption that all forward-decl reporting would come via decls (which is how I read your original proposal).

bolshakov-a:
I suggested that `VisitVarDecl` and similar would just call `ReportTypeUse(loc, type, DerefKind::Direct)`, and then `ReportTypeUse` short-circuits if the type is actually a pointer or a reference (what it already does). Fwd-decls would be reported from `VisitTagType` and similar, but without complex `CanForwardDeclareType` analysis.
`TagType`s needed fwd-decl can occur not only in declarations, but even in expressions like `sizeof(Class*)`. And these expressions may also occur under `CXXRecordDecl` node.
To sum up, my idea was to _separate_ full use reports from fwd-decl reports, because some declarations and expressions require complete type info but don't require even fwd-decl if that type occurs to be sugared pointer or reference

bolshakov-a:
I have a trouble with working with GitHub from phone...

bolshakov-a:
... and there are type nodes that require at least fwd-decl but not necessarily full type info.

bolshakov-a:
#1427 fixed reporting of a type when its constructor has been called. I hope this can be merged now.

kimgr:
Nice, LGTM!

PR #1341, opened 2023-11-12
Remove unused parameters in Derive{C,CXX}StdLib

kimgr:
The CompilerInstance params were added because we thought we would be able to pull currently-selected C/C++ standard library out the CompilerInstance.

Unfortunately that information is not available, and the parameters are unused, so remove them.


kilroyd (approved):
LGTM!

Issue #1336, opened 2023-11-10
include-what-you-use 0.21 behaves strangely with anonymous unions and structs.


Jhuighuy:
Hi!

The latest version of include-what-you-use (0.21, from brew) generates weird output for the following piece of code:
```cpp
#pragma once

template<class T>
class Class {};

template<>
class Class<int> {
  union { // or `struct`, same output
    float x;
    int y;
  };
};
```

Output:
```sh
$ include-what-you-use ./test.hpp

test.hpp should add these lines:

test.hpp should remove these lines:
-   // lines 8-8

The full include-list for test.hpp:
---
```

Everything works fine with version 0.20.

bolshakov-a:
IWYU erroneously handles anonymous `union`s and `struct`s (as well as unnamed ones like `struct { int x; } y;`) like forward declarations after #1232. But the direct reason of the reporting is that a class template explicit specialization is traversed twice. After the second pass, the unnamed/anonymous struct/union declaration is treated as the duplicating forward declaration, and IWYU suggests to remove the duplicate.

The double traversal is removed in the PR #1332. The problem is not reproduced at that branch. Hence, I'm not sure if this issue needs a specific fixing.

kimgr:
PR #1332 is now merged, @Jhuighuy let us know if this still reproduces on master, please.

Jhuighuy:
@kimgr Everything works fine, thanks!

PR #1342, opened 2023-11-12
Add testcase for issue #1336

kimgr:
This bug report came in roughly in parallel with PR #1332, which happens to fix it.

The anonymous type aspect of the bug never made it into the test suite, so add it as an isolated test case here.


kimgr:
@bolshakov-a I thought we'd get coverage for this quirk too

bolshakov-a:
Good idea, thanks! LGTM, but probably, unnamed entity (like `union { int x; } y;`) test case is worth adding, besides anonymous ones?

kimgr:
Like this?

bolshakov-a:
Yes, thanks!

kimgr:
Thank you!

PR #1343, opened 2023-11-12
Bounce ToolChain around so we can use it in InitGlobals

kimgr:
The Clang ToolChain is part of the Compilation, which is a transient structure set up by the Driver to eventually get a fully initialized CompilerInstance.

The ToolChain also seems to be the only place where we can query for which C++ standard library is in effect for the compilation unit.

Rather than initializing globals both partway through iwyu_driver and inside IwyuAction, funnel the ToolChain through to IwyuAction's constructor and pass it to InitGlobals in IwyuAction::CreateASTConsumer.

This resolves some unfortunate temporal coupling in iwyu_globals where we had to assume that ParseToolChain had been called before InitGlobals.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -140,6 +140,10 @@
 namespace clang {
 class FileEntry;
 class PPCallbacks;
+
+namespace driver {
+class ToolChain;
+}
 }  // namespace clang
 
 namespace include_what_you_use {
@@ -223,6 +227,7 @@ using clang::UsingDirectiveDecl;
 using clang::UsingShadowDecl;
 using clang::ValueDecl;
 using clang::VarDecl;
+using clang::driver::ToolChain;
 using llvm::cast;
 using llvm::dyn_cast;
 using llvm::dyn_cast_or_null;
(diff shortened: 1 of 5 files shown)


kimgr:
@kilroyd Please take a look. This is what I had in mind when I was rambling back in your old PR: https://github.com/include-what-you-use/include-what-you-use/pull/1244#issuecomment-1523936270 :slightly_smiling_face:

iwyu.cc
```diff
@@ -4362,12 +4367,17 @@ class IwyuAstConsumer
 
 // We use an ASTFrontendAction to hook up IWYU with Clang.
 class IwyuAction : public ASTFrontendAction {
+ public:
+  IwyuAction(const ToolChain& toolchain) : toolchain(toolchain) {
+  }
+
```
kilroyd, line 4373: Explicitly delete the default constructor?

kilroyd (commented):
Looks good. Just checking if you want to explicitly delete the `IwyuAction` default costructor.

iwyu.cc
kimgr: Good idea! Also made the toolchain ctor explicit.

kilroyd (approved):

Issue #1286, opened 2023-08-12
Reduce symbol mappings to be even more fine-grained


kimgr:
Created from #1283, by @alejandro-colomar:

In the end, I decided to do the cut for part 1 at a somewhat logical point:

-  Mappings guaranteed by ISO C are respected.
-  Mappings for `<sys/types.h>` and `<unistd.h>` (POSIX.1) are respected.
-  Any other POSIX.1 (or nonstandard) mappings have been removed.

It seems this doesn't affect out test suite, so it seems to be a fair cut point, isn't it?

Please ping me after the next release for part 2.  :)

_Originally posted by @alejandro-colomar in https://github.com/include-what-you-use/include-what-you-use/issues/1283#issuecomment-1675989244_

kimgr:
@alejandro-colomar IWYU 0.21 is just released, so now would be a good time to look into this. Do you remember what part 2 was? :slightly_smiling_face:

alejandro-colomar:
Probably I didn't decide, because I didn't keep a branch with a v2 patch.  Or I forgot.  :)

I propose chopping the remaining POSIX headers: `<sys/types.h>` and `<unistd.h>`.  Those alone should already add many warnings in users.

kimgr:
Remind me: what's the goal? To prefer recommending ISO C headers before POSIX headers?

alejandro-colomar:
Nope.  Being pedantic about IWYU guidelines, even when ISO C and POSIX are designed in a way that multiple headers are guaranteed to provide the same identifier in several places (by inclusion of other headers).

This was the best example of why we want this:
<https://github.com/include-what-you-use/include-what-you-use/pull/1283#issuecomment-1675924844>

```c
#include <monetary.h>

void
foo(size_t n, int arr[n])
{
    for (size_t i = 0; i < n; i++) {
        arr[i] = 42;
    }
}
```

This is perfect POSIX code, but we'll agree to disagree with POSIX.  ;)

alejandro-colomar:
So, we remove secondary headers that ISO C and POSIX say are guaranteed to provide an identifier, keeping only the canonical one, be it from ISO C or POSIX.

kimgr:
Ah, yes. That example is convincing.

So we're looking at a mass update to the symbol mappings, that will probably cause a lot of noise for users, but be more stringent? Oh, yeah, I see the issue title says "symbol mappings". I mean, I love pedantry, please move ahead with it when you have a chance.

[...]

PR #1344, opened 2023-11-12
Redoce libc mappings to really include what you use (part 2)

alejandro-colomar:
Continues: b9fde49306fa ("Reduce libc mappings to really include what you use (part 1)")

=== part 2: ===

Remove all secondary mappings that are in POSIX but not in ISO C, except when those mappings are actually better than the ISO C ones (e.g., clock_t from <sys/types.h> is lighter than from <time.h>, so keep the mapping).

This change is small, but may cause a lot of noise to users.  That's why we still kept some unnecessary mappings, which will go away in a part 3, in a future release of iwyu(1).

Reported-by: @firewave
Cc: @kimgr


kimgr:
> Remove all secondary mappings that are in POSIX but not in ISO C,

Should this say "[...] but **also** in ISO C"? I don't fully understand what this sentence is saying.

Could you also include some kind of motivation for why we're doing this? Is it to generate more portable suggestions by default?

And finally, would it be better to just leave POSIX includes alone? The `<monetary.h>` example is compelling, but I don't understand how it happens -- we don't have a mapping from `size_t` to `<monetary.h>`. Is it picked up via a definition of `size_t` in `bits/monetary-ldbl.h`?

Please add some more context in the commit message to answer these questions.

alejandro-colomar:
> > Remove all secondary mappings that are in POSIX but not in ISO C,
> 
> Should this say "[...] but **also** in ISO C"? I don't fully understand what this sentence is saying.

I'll completely rewrite that sentence.  How about this?

```
    === part 2: ===

    Since this change is quite radical, and may be annoying to users, we
    agreed to do it gradually.  In this part 2, we respect any mappings
    guaranteed by ISO C.

    In future releases of iwyu(1), those unnecessary mappings will probably
    also go away.
```

> 
> Could you also include some kind of motivation for why we're doing this? Is it to generate more portable suggestions by default?

I didn't include because I wrote:
```
Continues: b9fde49306fa ("Reduce libc mappings to really include what you use (part 1)")
```

So I expected one to read that commit message for the motivation, instead of repeating it.  But yeah, maybe I should repeat it here.

> 
> And finally, would it be better to just leave POSIX includes alone? The `<monetary.h>` example is compelling, but I don't understand how it happens -- we don't have a mapping from `size_t` to `<monetary.h>`. Is it picked up via a definition of `size_t` in `bits/monetary-ldbl.h`?

We don't have it because it was removed in part 1.

```sh
$ git log -1 --oneline b9fde49
b9fde49 Reduce libc mappings to really include what you use (part 1)
$ git log -1 --oneline b9fde49 -p | grep monetary
-  { symbol: [ "ssize_t", private, "<monetary.h>", public ] },
-  { symbol: [ "size_t", private, "<monetary.h>", public ] },
-  { "ssize_t", kPrivate, "<monetary.h>", kPublic },
-  { "size_t", kPrivate, "<monetary.h>", kPublic },
```

The most egregious mappings were already removed in that commit.

In that commit we decided to exclude from the removal the following mappings (the commit message says so):

```
    Since this change is quite radical, and may be annoying to users, we
    agreed to do it gradually.  In this part 1, we respect any mappings
    guaranteed by ISO C.  We also respected any mappings for a couple of
    POSIX.1 header files, <sys/types.h> and <unistd.h>, as they are very
    commonly used.
```

It seems I also missed the following mapping, which should have been removed in part1:
```
-  { symbol: [ "ino64_t", private, "<dirent.h>", public ] },
```

This part 2 does the removal of mappings to `<sys/types.h>` and `<unistd.h>`, when those headers aren't the canonical header for an identifier.

> 
> Please add some more context in the commit message to answer these questions.

Okay, proposal for a full commit message:

```
    Reduce libc mappings to really include what you use (part 2)
    
    Continues: b9fde49306fa ("Reduce libc mappings to really include what you use (part 1)")
    
    The standards (ISO C and POSIX.1) guarantee that many definitions are
    provided by more than one header.  Programmers may #include <stdio.h>
    for the definition of NULL, and be covered by the standards.  However,
    that's not following the include-what-you-use principle, and we should
    not promote that.
    
    Cut the mappings to the minimum: the principal header that should define
    an identifier.  In most cases, POSIX.1 and ISO C agree on that header,
    and we use that one.  In a few cases, ISO C prescribes <time.h> and
    POSIX.1 prescribes <sys/types.h>; keep both, but give preference to the
    one by POSIX.1, <sys/types.h>, which is smaller.
    
    When the type or macro is not standardized, if there's a clearly
    canonical header for it, keep that one; if there's not, let's keep all
    of them.
    
    Regarding C++'s <cstddef> and NULL, C++ uses nullptr.  C++ code that
    uses NULL is probably C-compat code, and that code should be including
    the C header.
    
    Also remove or update comments that are not up to date anymore.
    
    === part 1: ===
    
    Since this change is quite radical, and may be annoying to users, we
    agreed to do it gradually.  In this part 1, we respect any mappings
    guaranteed by ISO C.  We also respected any mappings for a couple of
    POSIX.1 header files, <sys/types.h> and <unistd.h>, as they are very
    commonly used.
    
    In future releases of iwyu(1), those unnecessary mappings will probably
    also go away.
    
    Reported-by: Oliver Stöneberg <https://github.com/firewave>
    Cc: Kim Gräsman <kim.grasman@gmail.com>
    Signed-off-by: Alejandro Colomar <alx@kernel.org>
```

kimgr:
Hehe, I read all of that and I still don't feel I understand what the patch is doing.

I don't think we need to repeat all of b9fde49306fa's message, but maybe summarize it if there's anything relevant there.

It sounds from your current description like there is a high-level goal across these parts. Please summarize that in a sentence or two.

Then there must be some part-way goal for this part 2. Please summarize that. Or is it just a few random changes, so as not to cause too much impact on user diagnostics?

Some examples of questions a reader will have are:

* What is the motivation for the change?
* If we didn't make this change, how would the world be worse?
* What is the tradeoff, that is, how will users suffer?

alejandro-colomar:
> Hehe, I read all of that and I still don't feel I understand what the patch is doing.

Heh!

> Then there must be some part-way goal for this part 2. Please summarize that. Or is it just a few random changes, so as not to cause too much impact on user diagnostics?

It is the latter.  It's just a random cut point to not cause too much impact on users.  If it were me, I'd remove all remaining non-IWYU mappings.

> 
> Some examples of questions a reader will have are:
> 
>     * What is the motivation for the change?

For the whole series:  Remove mappings that are not the canonical standard header.

For this part 2:  Don't make it all in one step, to allow users to adapt or complain.

> 
>     * If we didn't make this change, how would the world be worse?

Some programs could be including `<unistd.h>` just for `NULL`.
It's not as bad as `<monetary.h>` for `size_t`, which is why part1 was more urgent than part2, but it's still bad.

> 
>     * What is the tradeoff, that is, how will users suffer?

They'll be asked to replace some cases of, say, `#include <unistd.h>` by `#include <stddef.h>`, if they were only using `size_t` (just an example).

kimgr:
Thanks! I think I would write a commit message something like this:
```
Reduce attribution to POSIX headers for standard symbols

This is part of an effort to remove mappings that encourage implicit dependencies,
i.e. where the target is not the canonical standard header; the header that actually
has the definition.

We do this incrementally in small steps to allow users to adapt or complain.

This time, move all symbols defined in `<sys/types.h>` to that header, because it's
typically cheaper to include than many of the other POSIX headers that reexport its
types.
```

Looking at the diff, it looks like that's mostly what's going on? But now the diff also tells another story -- it looks like you're removing `<unistd.h>` wherever possible, in favor of `<sys/types.h>`? So I wonder if you should make the patch focus on actually removing `<unistd.h>` where there are better candidates for it? That way you have a clear strategy to describe in the commit message, and people who are surprised that `<unistd.h>` disappeared can `git grep 'unistd\.h` and learn why.

kimgr:
... and once `<unistd.h>` is cleaned up, I think we can accept another similar cleanup commit for the IWYU 0.22 timeline.

My take is that focused changes are always better than random changes, even if the focus is randomly selected.

kimgr:
Thanks! Still a little weird with three commits with almost identical subject lines, but I'll take it.

alejandro-colomar:
Heh!  I was going to ask if you liked it after getting on a train, and when I open the laptop I see you've merged already.  Thanks!

alejandro-colomar:
I see you've released 0.22.  I'll open a new issue for part 3.

PR #1345, opened 2023-11-12
Accept -E to preprocess-only

kimgr:
Until now IWYU, like other Clang tools, have only accepted compile commands, rewritten them and replaced them with its own frontend action for IWYU analysis.

But it can be useful to ask IWYU about its preprocessor state to answer questions like:

* which macros were defined for this execution?
* what does the input look like after macro expansion?

For example, we've been discussing letting the test framework use '-dM -E' to dump macros and detect the active C and C++ standard library based on what definitions are available, and then use the reverse-engineered library names to enable/disable tests based on environment.

Thus, try to detect preprocess-only commands (both for MSVC and GCC-style spellings) and let Clang do whatever it sees fit for preprocessing jobs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,6 +162,7 @@ else()
     clangAST
     clangSema
     clangFrontend
+    clangFrontendTool
     clangDriver
 
     # Revision [1] in clang moved PCHContainerOperations from Frontend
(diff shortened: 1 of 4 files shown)


kimgr:
@kilroyd Pretty happy with this! I have some ideas for cleanup on top of this, but I wanted to get the minimal change for `-E` in place first.

kilroyd (commented):
Looks good to me!

PR #1346, opened 2023-11-12
Test 'DecltypeType' handling

bolshakov-a:
Currently, it doesn't work almost completely, except reporting template argument types due to scanning template specialization declaration during templated object constructor handling.


bolshakov-a:
Perhaps too early. The test is passed on the branch from #1340 because `IndiectClass` reports from constructor scanning itself (it is a field of template specialization which is initialized in the tpl spec constructor). I should check it before making PR...

kimgr:
Aha, oops! Oh well, any coverage is better than perfect coverage. But would be nice if you could tweak something to cover the case you noted earlier.

PR #1347, opened 2023-11-12
Improve 'decltype' test

bolshakov-a:
... so that it doesn't pass on the branch from #1340. `Tpl` needs complete type of the template parameter, but not for construction.


kimgr:
Yup. I cherrypicked both this and #1340 onto a branch, and the test does fail. Looks good to me.

PR #1348, opened 2023-11-19
Avoid fwd-decl warnings along with full use ones

bolshakov-a:
The output with 3rd verbosity level is filtered from fwd-decl warnings when there is a full use of the same entity in the same line. This makes the output cleaner.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1736,9 +1736,35 @@ static string GetWarningMsg(const OneUse& use) {
 
 int IwyuFileInfo::EmitWarningMessages(const vector<OneUse>& uses) {
   set<pair<int, string>> iwyu_warnings;   // line-number, warning-msg.
+
+  // Report all full uses, and build an index of their (line,decl).
+  set<pair<int, const NamedDecl*>> full_use_index;
   for (const OneUse& use : uses) {
-    if (use.is_iwyu_violation())
-      iwyu_warnings.insert(make_pair(use.UseLinenum(), GetWarningMsg(use)));
+    if (!use.is_iwyu_violation() || !use.is_full_use())
+      continue;
+
+    int line = use.UseLinenum();
+    iwyu_warnings.insert(std::make_pair(line, GetWarningMsg(use)));
+    full_use_index.insert(std::make_pair(line, use.decl()));
+  }
+
+  // Report all forward-declaration, except for (line,decl) already covered by a
+  // full use.
+  for (const OneUse& use : uses) {
+    if (!use.is_iwyu_violation() || use.is_full_use())
+      continue;
+
+    // For class template implicit instantiations, fwd-decl uses refer to
+    // ClassTemplateDecl whereas full uses refer to their definition nodes.
+    // Partial specs aren't considered for simplicity.
+    const NamedDecl* decl = use.decl();
+    if (isa<ClassTemplateDecl>(decl))
+      decl = GetTagDefinition(decl);
+
+    int line = use.UseLinenum();
+    if (!full_use_index.count(std::make_pair(line, decl))) {
+      iwyu_warnings.insert(std::make_pair(line, GetWarningMsg(use)));
+    }
   }
   // Nice that set<> automatically sorts things for us!
   for (const pair<int, string>& warning : iwyu_warnings) {
--- a/tests/cxx/alias_template.cc
+++ b/tests/cxx/alias_template.cc
@@ -20,14 +20,12 @@ template<class T> struct FullUseTemplateArgInSizeof {
 // Test that we go through alias template and handle aliased template
 // specialization.
 template<class T> using Alias = FullUseTemplateArgInSizeof<T>;
-// IWYU: IndirectClass needs a declaration
 // IWYU: IndirectClass is...*indirect.h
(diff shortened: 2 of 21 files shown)


iwyu_output.cc
```diff
@@ -1737,8 +1737,29 @@ static string GetWarningMsg(const OneUse& use) {
...
+            return false;
+          if (other_use.decl() == decl)
+            return true;
+          // For class template implicit instantiations, fwd-decl uses refer to
+          // ClassTemplateDecl whereas full uses refer to their definition
+          // nodes. Partial specs aren't considered for simplicity.
+          return isa<ClassTemplateDecl>(decl) &&
+                 GetTagDefinition(decl) == other_use.decl();
+        };
+        if (std::any_of(uses.begin(), uses.end(), full_use_in_line))
```
kimgr, line 1757: The nested loop here is a little annoying, I guess a file with many uses (like badinc.cc) might suffer.

Would it be possible to do something like this instead:
```diff
diff --git a/iwyu_output.cc b/iwyu_output.cc
index 99958ef..dd15ade 100644
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1736,29 +1736,34 @@ static string GetWarningMsg(const OneUse& use) {
 
 int IwyuFileInfo::EmitWarningMessages(const vector<OneUse>& uses) {
   set<pair<int, string>> iwyu_warnings;   // line-number, warning-msg.
+
+  // Report all full uses, and build an index of their (line,decl).
+  set<pair<int, const NamedDecl*>> full_use_index;
   for (const OneUse& use : uses) {
-    if (use.is_iwyu_violation()) {
-      const int linenum = use.UseLinenum();
-      if (!use.is_full_use()) {
-        // Avoid fwd-decl warnings nearby complete-use warnings to make output
-        // cleaner.
-        auto full_use_in_line = [linenum,
-                                 decl = use.decl()](const OneUse& other_use) {
-          if (!other_use.is_full_use() || other_use.UseLinenum() != linenum)
-            return false;
-          if (other_use.decl() == decl)
-            return true;
-          // For class template implicit instantiations, fwd-decl uses refer to
-          // ClassTemplateDecl whereas full uses refer to their definition
-          // nodes. Partial specs aren't considered for simplicity.
-          return isa<ClassTemplateDecl>(decl) &&
-                 GetTagDefinition(decl) == other_use.decl();
-        };
-        if (std::any_of(uses.begin(), uses.end(), full_use_in_line))
-          continue;
-      }
+    if (!use.is_iwyu_violation() || !use.is_full_use())
+      continue;
+
+    int line = use.UseLinenum();
+    iwyu_warnings.insert(std::make_pair(line, GetWarningMsg(use)));
+    full_use_index.insert(std::make_pair(line, use.decl()));
+  }
+
+  // Report all forward-declaration, except for (line,decl) already covered by a
+  // full use.
+  for (const OneUse& use : uses) {
+    if (!use.is_iwyu_violation() || use.is_full_use())
+      continue;
+
+    // For class template implicit instantiations, fwd-decl uses refer to
+    // ClassTemplateDecl whereas full uses refer to their definition nodes.
+    // Partial specs aren't considered for simplicity.
+    const NamedDecl* decl = use.decl();
+    if (isa<ClassTemplateDecl>(decl))
+      decl = GetTagDefinition(decl);
 
-      iwyu_warnings.insert(make_pair(linenum, GetWarningMsg(use)));
+    int line = use.UseLinenum();
+    if (!full_use_index.count(std::make_pair(line, decl))) {
+      iwyu_warnings.insert(std::make_pair(line, GetWarningMsg(use)));
     }
   }
   // Nice that set<> automatically sorts things for us!
```

I played with it a little, and it appears to save ~1 second for the test suite on average.

bolshakov-a: Nice optimization, thank you!

kimgr:
Hmm, I was just about to merge this, but then hesitated... Is this really what we want to do?

The line warnings give some visibility into what IWYU is seeing in the code and the `... needs a declaration` warnings indicate that IWYU found and registered a forward-declare use. Just throwing it away seems like we lose a lot of information about (annoying) decisions IWYU has made earlier.

The full uses trump forward-decl uses later in the add/remove analysis anyway.

bolshakov-a:
I just want to try to remove `CanForwardDeclareType`, report full type uses in `VisitVarDecl`, `VisitNestedNameSpecifier` and similar handlers, and report fwd-decl uses in `VisitTagType`, `VisitTemplateSpecializationType` (i.e. when a type is explicitly written) unconditionally, as proposed [here](https://github.com/include-what-you-use/include-what-you-use/pull/1340#issuecomment-1807097820). It could lead to lots of new `needs a declaration` warnings.

kimgr:
Yeah, I remember the suggestion, and I thought it made sense initially. But now it seems to me we're just papering over IWYU behavior, making it harder to test and reason about.

I think we can take the additional diagnostics.

bolshakov-a:
Btw, if you prefer to keep the current mechanism which chooses between fwd-decl and full use report on a written type, I could just add `VisitDecltypeType` handler instead of refactoring.

kimgr:
Let's try that and see what it looks like.

bolshakov-a:
Ok, dropping this.

PR #1354, opened 2023-12-03
Remove 'IsTemplateTypeParmUseFullUse' function

bolshakov-a:
More general `CanForwardDeclareType` works already well even for `SubstTemplateTypeParmType`s.

The case of using a reference in a `sizeof` expression is already handled in `IwyuBaseAstVisitor::VisitUnaryExprOrTypeTraitExpr`.

Probably, the commits should be squashed (they are separated for clarity). Or even, the test case should be extracted as the first commit, and all the code changes as the second one.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3022,67 +3022,6 @@ class InstantiatedTemplateVisitor
     return TraverseTemplateSpecializationTypeHelper(typeloc.getTypePtr());
   }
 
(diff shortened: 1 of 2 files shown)


kimgr:
Looks good to me! I don't have a strong preference for commit partitioning here, feel free to do what you prefer and let me know when it's ready to merge.

PR #1356, opened 2023-12-03
Simplify 'CanForwardDeclareType'

bolshakov-a:
`TypedefNameDecl` node is marked as being "in forward-declare context", so the control flow doesn't even reach the check in `CanForwardDeclareType` because it is short-circuited by `in_forward_declare_context()` check above.

The typedef or type alias underlying types which a user wants to provide are found out in `IsProvidedTypeComponent` function.

No functional change.


kimgr:
Beautiful, thanks!

PR #1357, opened 2023-12-03
Improve handling template parameter aliases

bolshakov-a:
Full uses of aliased template parameters are now handled in `IwyuBaseAstVisitor::VisitTypedefType`, in the common manner with handling type aliases outside template instantiations. This handler only reports type alias underlying types. The uses of the type alias declarations themselves are reported from `IwyuAstConsumer` implementation, as before, and not during implicitly instantiated template traversal, because they should be already reported at the traversal of the template as written.
    
`ReportTypeUse` function reports not only direct aliases of template parameters but also when the aliased template parameter was sugared. A corresponding test case is added.

The main goal of this PR is to simplify `AnalyzeTemplateTypeParmUse` function.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2348,6 +2348,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  bool VisitTypedefType(clang::TypedefType* type) {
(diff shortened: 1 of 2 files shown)


kimgr:
Makes sense, thanks

PR #1359, opened 2023-12-03
Add support for driver test assertions without location

kimgr:
Add '// IWYU~:' to complement '// IWYU:'.

This new form checks for diagnostics anywhere in the output, e.g. driver
diagnostics without `<file>:<line>:` decoration. Designed to catch errors
like:

    error: unable to handle compilation, expected exactly one compiler job
    error: unknown argument: '-fdump-rtl-expand'
    warning: argument unused during compilation: '-fobjc-exceptions'

Add a motivating testcase for multiple inputs.

### DIFF
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -29,6 +29,11 @@ _EXPECTED_DIAGNOSTICS_RE = re.compile(r'^\s*// IWYU:\s*(.*)$')
 _ACTUAL_DIAGNOSTICS_RE = re.compile(r'^(.*?):(\d+):\d+:\s*'
                                     r'(?:warning|error|fatal error):\s*(.*)$')
 
+# Diagnostics without line/column number (e.g. from driver), e.g.:
+# // IWYU~: argument unused during compilation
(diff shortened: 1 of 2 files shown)


kimgr:
@kilroyd I came up with an alternative which looks kind-of nice. This associates more clearly with the `// IWYU` diagnostic assertion, but is a little... eclectic. I think I still prefer that plain ol' `// IWYU_DRIVER`. Thoughts?

kilroyd:
I think I agree. `IWYU(DRIVER)` makes sense compared to `IWYU`, but the `()` feels odd.

What if you just used `IWYU` but with a different separator character to indicate no location specification? Something like `IWYU!` or `IWYU:~`?

Either way, if you can write it up somewhere so it's clear that should be fine.

kimgr:
My kneejerk reaction to `// IWYU:~` was "ew!", but when I played around with it a little I actually grew to like it. I switched it up so the tilde comes before the colon, because that felt a little better. I think it's reasonable to read `// IWYU~: expected exactly one` as `match diagnostic 'expected exactly one' anywhere in the output`.

kilroyd:
Looks good to me!

Issue #1349, opened 2023-11-30
IWYU wants to include header for re-exported inherited member functions


carlosgalvezp:
Hi,

Consider the following example where we create our own custom `Vector` class, that inherits most of the functionality from `std::vector`, and where we might customize some member functions.

```
$ cat vector.h 
#include <vector>

template <typename T>
class Vector : private std::vector<T>
{
 public:
    using Base = std::vector<T>;

    using Base::Base;
    using typename Base::allocator_type;
    using typename Base::const_pointer;
    using typename Base::const_reference;
    using typename Base::pointer;
    using typename Base::reference;
    using typename Base::iterator;
    using typename Base::const_iterator;
    using typename Base::reverse_iterator;
    using typename Base::const_reverse_iterator;
    using typename Base::value_type;
    using typename Base::size_type;
    using typename Base::difference_type;

    using Base::begin;
    using Base::end;
    using Base::cbegin;
    using Base::cend;
};
```

```
$ cat main.cpp 
#include <iostream>

#include "vector.h"

int main()
{
  Vector<int> v{1,2,3};

  for(int& x : v)
  {
    std::cout << x << std::endl;
  }  
}
```

IWYU however requires that `<vector>` be included here, which I believe shouldn't be - if one day we decide that `Vector` should be implemented in a different way that doesn't need `<vector>` at all, then the include required by IWYU would have to be removed as well:

```
$ include-what-you-use main.cpp 

main.cpp should add these lines:
#include <vector>    // for vector

main.cpp should remove these lines:

The full include-list for main.cpp:
#include <iostream>  // for char_traits, basic_ostream, basic_ostream::opera...
#include <vector>    // for vector
#include "vector.h"  // for Vector
---
```

PR #1363, opened 2023-12-08
Don't "provide" types on tpl instantiation side

bolshakov-a:
The problem was in analyzing in `IwyuBaseAstVisitor` (and hence in `InstantiatedTemplateVisitor`) whether a written template specialzation type is provided. Only template author can decide to provide a type inside a template, not template user for implicitly generated specialization code. To fix it, the analysis is moved into `IwyuAstConsumer` which doesn't traverse implicit template instantiation code.

`IsProvidedTypeComponent` and `CanBeProvidedTypeComponent` are moved into the private section of `IwyuAstConsumer` because they are now (and should be) needed only in that visitor class.

The test case uses two template arguments: `T1` to avoid putting `TemplateSpecializationType` into `nodes_to_ignore_` set and `T2` to allow the type to be reported (IWYU can currently report only template argument types from instantiated template traversal). `typedef` could in principle be used instead of `using`.

Noisy fwd-decl suggestions are appeared in the tests because they aren't now filtered out inside `IwyuBaseAstVisitor::VisitTemplateSpecializationType`.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2335,14 +2335,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
     const NamedDecl* decl = TypeToDeclAsWritten(type);
 
-    const auto [is_provided, comment] =
-        IsProvidedTypeComponent(current_ast_node());
     // If we are forward-declarable, so are our template arguments.
-    if (CanForwardDeclareType(current_ast_node()) && !is_provided) {
+    if (CanForwardDeclareType(current_ast_node())) {
       ReportDeclForwardDeclareUse(CurrentLoc(), decl);
       current_ast_node()->set_in_forward_declare_context(true);
     } else {
-      ReportDeclUse(CurrentLoc(), decl, comment);
+      ReportDeclUse(CurrentLoc(), decl);
     }
 
     return true;
@@ -4137,6 +4135,14 @@ class IwyuAstConsumer
           current_ast_node(), resugar_map,
           ExtractProvidedTypeComponents(resugar_map));
     }
+
+    const auto [is_provided, comment] =
+        IsProvidedTypeComponent(current_ast_node());
+    // Don't call ReportTypeUse here, because scanning of template instantiation
+    // internals should be avoided: it is allowed not to provide some of
+    // template args.
+    if (is_provided)
+      ReportDeclUse(CurrentLoc(), TypeToDeclAsWritten(type), comment);
 
     return Base::VisitTemplateSpecializationType(type);
   }
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2469,49 +2469,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return visitor_state_->preprocessor_info;
   }
 
-  pair<bool, const char*> CanBeProvidedTypeComponent(
-      const ASTNode* node) const {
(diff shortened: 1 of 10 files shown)


iwyu.cc
```diff
@@ -4138,6 +4093,14 @@ class IwyuAstConsumer
           ExtractProvidedTypeComponents(resugar_map));
     }
 
+    const auto [is_provided, comment] =
+        IsProvidedTypeComponent(current_ast_node());
+    // Don't call ReportTypeUse here, because scanning of template instantiation
+    // internals should be avoided: it is allowed not to provide some of
+    // template args.
+    if (is_provided)
```
kimgr, line 4101: The testcase looks right, but I don't understand why this isn't `!is_provided`...?

Is it because a template specialization is "definition-side" and any type that's provided by the base template also needs to be reported here?

bolshakov-a: `IsProvidedTypeComponent` function is used to determine types provided by the type alias or function author on the alias or function definition side. So, this code handles occurrences of (written) template spec types in that definitions and not on any use site of the alias or the function. For example:
```cpp
typedef Tpl<Class> Alias;
```
If the `Alias` author decided to provide the underlying type, IWYU should report `Tpl` as well as `Class` when traversing `Alias` definition. `VisitTemplateSpecializationType` should do that for `Tpl` and `VisitTagType` should do the same for `Class`.

bolshakov-a: The problem was in that such an analysis was performed not only on type alias definitions (and probably fn declarations) _as-written_ but also on implicitly generated code (inside implicit template instantiations) which is handled by `InstantiatedTemplateVisitor`.

kimgr: Thanks, I think that makes sense -- so `VisitTemplateSpecializationType` hits the template instantiation/expansion/specialization and because it's part of a typedef/alias/function decl, the custom provision rule hits? I wonder if this affects any of the label:template-spill issues we had before. I happen to have a stack of testcases, I'll run them over this patch just to see if there's any coverage missing.

kimgr: Also, I feel like this bug might be related: https://github.com/include-what-you-use/include-what-you-use/issues/1349 -- but I can imagine there's more subtleties there.

kimgr: > I wonder if this affects any of the label:template-spill issues we had before. I happen to have a stack of testcases, I'll run them over this patch just to see if there's any coverage missing.

No surprises here, everything still works as before (which I think/hope means that my 10 reduced examples are covered by the test suite :crossed_fingers:).

bolshakov-a: > Thanks, I think that makes sense -- so VisitTemplateSpecializationType hits the template instantiation/expansion/specialization and because it's part of a typedef/alias/function decl, the custom provision rule hits?

The problem appeared when `VisitTemplateSpecializationType` hit a template specialization type which is a part of a typedef/alias/function decl _which in turn is a part of an implicit instantiation of another template_.

I think this case is very specific, and users hardly faced it. The change is rather conceptual and is aimed to enable possibilities of further refactorings and improvements.

bolshakov-a: > * `WithAlias<int, TplIndirectStruct2<int>> wa;` is the implicit instantiation of another template
> * `using Type = TplIndirectStruct2<T1>;` is the alias declaration
> * `TplIndirectStruct2<T1>` is the template specialization type

Yes, correct.

Issue #1296, opened 2023-08-24
IWYU macro bug


liuluheng:
Say we have the following code `iwyu_rm_macro.cc`

``` cpp
#include <stdio.h>

#define MY_MACRO \
  namespace A {  \
  struct a;      \
  }              \
  int b = 0;

MY_MACRO

int main() { printf("%d\n", b); }
```
Then we run below
```
$ bin/include-what-you-use iwyu_rm_macro.cc

wyu_rm_macro.cc should add these lines:

wyu_rm_macro.cc should remove these lines:
- namespace A { struct a; }  // lines 9-9

The full include-list for iwyu_rm_macro.cc:
#include <stdio.h>  // for printf
---
```
line 9 is `MY_MACRO`, if we run `fix_includes.py`, `MY_MACRO` will be removed. Then the code **CANNOT COMPLIE**.

`IWYU pragma: keep/begin_keep/end_keep` dose **NOT** work on this case.

IWYU version 0.20

kimgr:
Yes, I can reproduce. I think it should be possible to detect that the forward-declaration is inside a macro expansion and just skip the removal.

kimgr:
Huh. There's actually a test for exactly this scenario: https://github.com/include-what-you-use/include-what-you-use/blob/master/tests/cxx/forward_declare_in_macro.cc, except it never imagined that a macro could do two things.

I think it's questionable to remove macros containing forward-declarations, so I think I'll try to change this.

kimgr:
@liuluheng Please take a look at https://github.com/include-what-you-use/include-what-you-use/pull/1365

liuluheng:
@kimgr Thank you for the fix. It appears to be working. I will test the patch as soon as possible.

liuluheng:
@kimgr I had tested it in a real complex macro like [POINT_CLOUD_REGISTER_POINT_STRUCT](https://github.com/PointCloudLibrary/pcl/blob/master/common/include/pcl/register_point_struct.h#L64). It WORKs.

kimgr:
Thank you! Available on master now.

PR #1365, opened 2023-12-09
Keep forward-declare lines containing macro expansions

kimgr:
IWYU used to look through macros to be able to remove unused forward declaration lines even if they were the result of a macro expansion.

Macros are tricky, though. As demonstrated in #1296, a macro can do more than one thing, e.g. declare two symbols or even generate code with side effects. Consider for example:

    #define PROLOG_PIN(x)  class Pin##x; TogglePin(x);

    void foo() {
        PROLOG_PIN(2)
        // Pin2 class is unused, macro would be removed and hardware
        // would be unhappy.
    }

Work around this by force-keeping any forward declaration whose location is even partially inside a macro.

Fixes issue #1296.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1654,4 +1654,9 @@ bool IsDeclaredInsideFunction(const Decl* decl) {
   return isa<FunctionDecl>(decl_ctx);
 }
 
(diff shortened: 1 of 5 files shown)


liuluheng:
LGTM

PR #1382, opened 2023-12-29
Rewrite codebase to use OptionalFileEntryRef over FileEntry*

kimgr:
This is the result of:

    sed -i -e 's/using clang::FileEntry;/using clang::OptionalFileEntryRef;/g' *.cc *.h
    sed -i -e 's/const FileEntry\*/OptionalFileEntryRef/g' *.cc *.h
    sed -i -e 's/const clang::FileEntry\*/clang::OptionalFileEntryRef/g' *.cc *.h
    sed -i -e 's/GetFileEntry(/GetFileEntryRef(/g' *.cc *.h
    sed -i -e 's/GetLocFileEntry(/GetLocFileEntryRef(/g' *.cc *.h

    git restore iwyu_location_util.*

followed by fixing up a small number of trivial type errors manually, and formatting.

No functional change.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -178,7 +178,6 @@ using clang::EnumConstantDecl;
 using clang::EnumDecl;
 using clang::EnumType;
 using clang::Expr;
-using clang::FileEntry;
 using clang::FriendDecl;
 using clang::FriendTemplateDecl;
 using clang::FunctionDecl;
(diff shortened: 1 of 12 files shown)


kimgr:
I would have prefered to do this in smaller steps, but after 50+ commits I came up with this big-bang approach instead.

I did enough preparatory fixes that the machine rewrite caught 98% of the necessary changes, and the rest is covered by "fixing up a small number of trivial type errors manually".

PR #1387, opened 2023-12-31
Allow not to provide default type tpl arguments

bolshakov-a:
Prior to this, IWYU always required complete default template argument type info at template definition side. However, as discussed in #1333, a template author should be able to decide whether to provide it or not, because they are forward-declarable in principle. Hence, they are now handled similarly to typedef-fnreturn-autocast triad, but only at the template definition side. At the template instantiation side, there already is some logic to detect their provision status (see `InstantiatedTemplateVisitor::CanIgnoreType`). Hence, there is some redundancy when IWYU requires complete type info both at the template definition and instantiation sides at the first pass, but this behavior was already present on the master branch, and this PR should not make it worse.

Besides, this changeset fixes a subtle bug similar to the one fixed in #1363, when IWYU suggested complete type info for an inner template default type argument at the outer template instantiation site. The corresponding test case added to `default_tpl_arg` test.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -213,6 +213,7 @@ using clang::TemplateDecl;
 using clang::TemplateName;
 using clang::TemplateSpecializationType;
 using clang::TemplateSpecializationTypeLoc;
+using clang::TemplateTypeParmDecl;
 using clang::TranslationUnitDecl;
 using clang::Type;
 using clang::TypeDecl;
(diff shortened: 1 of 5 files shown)


kimgr:
Nice! I finally understand what `CanBeProvidedTypeComponent` is different from `IsProvidedTypeComponent` :-)

bolshakov-a:
`CanBeProvidedTypeComponent` is just a utility for `IsProvidedTypeComponent`, and could be incorporated into the latter, in principle.

kimgr:
Yeah, I just didn't appreciate why it was relevant before, but now it was nice to see that `TemplateTypeParmDecl` _can_ be provided, but `IsProvidedTypeComponent` actually decides whether it _is_ provided based on the `CodeAuthor...` predicate. I like it.

PR #1389, opened 2024-01-02
Improve provided deduced type determination

bolshakov-a:
It turns out that extracting provided types from `resugar_map` doesn't always work when type template arguments are deduced (e. g. with partial class template specializations and function template argument deduction). Consider this code:
```cpp
template <typename T> class Tpl;
template <typename T> class Tpl<T*> {};

using ProvidingPtrAlias = Class*;

Tpl<ProvidingPtrAlias> t;
```
The resulting `resugar_map` would contain only `Class -> Class` mapping and not `Class* -> ProvidingPtrAlias`. Because it wouldn't contain `ProvidingPtrAlias`, there is no way to determine from it that `Class` is provided actually. To solve this problem, provided type set determination is moved inside `resugar_map`-getting algorithms, where full sets of type template or function argument type components are available.

The solution may look ugly, but I hope to rewrite the code soon because I want to remove `resugar_map` completely as proposed in [that discussion](https://github.com/include-what-you-use/include-what-you-use/pull/1182#issuecomment-1385818682). Nevertheless, better design suggestions are welcome! An alternative could be to insert components interesting for provided type determination into `resugar_map`. Another one could be to separate `resugar_map`- and `provided_types`-getting functions.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -837,7 +837,6 @@ static map<const Type*, const Type*> GetTplTypeResugarMapForFunctionNoCallExpr(
 
 static map<const Type*, const Type*>
 GetTplTypeResugarMapForFunctionExplicitTplArgs(
-    const FunctionDecl* decl,
     const TemplateArgumentListInfo& explicit_tpl_list) {
   map<const Type*, const Type*> retval;
   for (const TemplateArgumentLoc& loc : explicit_tpl_list.arguments()) {
@@ -922,8 +921,8 @@ map<const Type*, const Type*> GetTplTypeResugarMapForFunction(
     const TemplateArgumentListInfo& explicit_tpl_args =
         GetExplicitTplArgs(callee_expr);
     if (explicit_tpl_args.size() > 0) {
-      retval = GetTplTypeResugarMapForFunctionExplicitTplArgs(
-          decl, explicit_tpl_args);
+      retval =
+          GetTplTypeResugarMapForFunctionExplicitTplArgs(explicit_tpl_args);
       start_of_implicit_args = explicit_tpl_args.size();
     }
   } else {
@@ -931,8 +930,8 @@ map<const Type*, const Type*> GetTplTypeResugarMapForFunction(
     const TemplateArgumentListInfo& explicit_tpl_args =
         GetExplicitTplArgs(calling_expr);
     if (explicit_tpl_args.size() > 0) {
-      retval = GetTplTypeResugarMapForFunctionExplicitTplArgs(
-          decl, explicit_tpl_args);
+      retval =
+          GetTplTypeResugarMapForFunctionExplicitTplArgs(explicit_tpl_args);
       retval = ResugarTypeComponents(retval);
     }
     return retval;
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -889,7 +889,7 @@ static const Type* GetSugaredTypeOf(const Expr* expr) {
 
 map<const Type*, const Type*> GetTplTypeResugarMapForFunction(
     const FunctionDecl* decl, const Expr* calling_expr) {
-  map<const Type*, const Type*> retval;
+  map<const Type*, const Type*> resugar_map;
 
   // If calling_expr is nullptr, then we can't find any explicit template
   // arguments, if they were specified (e.g. 'Fn<int>()'), and we
@@ -897,9 +897,10 @@ map<const Type*, const Type*> GetTplTypeResugarMapForFunction(
   // can't resugar at all.  We just have to hope that the types happen
   // to be already sugared, because the actual-type is already canonical.
   if (calling_expr == nullptr) {
-    retval = GetTplTypeResugarMapForFunctionNoCallExpr(decl, 0);
-    retval = ResugarTypeComponents(retval);  // add in retval's decomposition
-    return retval;
+    resugar_map = GetTplTypeResugarMapForFunctionNoCallExpr(decl, 0);
+    resugar_map = ResugarTypeComponents(
+        resugar_map);  // add in resugar_map's decomposition
+    return resugar_map;
   }
 
   // If calling_expr is a CXXConstructExpr of CXXNewExpr, then it's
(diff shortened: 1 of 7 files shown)


iwyu_ast_util.h
```diff
@@ -548,6 +549,11 @@ clang::SourceRange GetSourceRangeOfClassDecl(const clang::Decl* decl);
 set<clang::FunctionDecl*> GetLateParsedFunctionDecls(
     clang::TranslationUnitDecl* decl);
 
+struct TemplateInstantiationData {
```
kimgr, line 552: Could you add two constructors to this type; one for the resugar map and one for resugar map + provided types? I find the initialization syntax currently used to construct it is a little too implicit.

kimgr:
I think it looks like a pretty good approach; nice to see that it's possible to expose the type provision knowledge from `iwyu.cc` to `iwyu_ast_util.cc` like that.

iwyu_ast_util.h
bolshakov-a: Wouldn't the constructor calling syntax look the same? Eventually, we can turn on `--std=c++20` and use [designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers).

bolshakov-a: > Wouldn't the constructor calling syntax look the same?

I mean that constructors can equally be called with curly braced syntax as well as with parentheses. But if you prefer parentheses, I can add constructors, of course.

kimgr: I want to make sure the type name is spelled out when constructing it; `TemplateInstantiationData(resugar_map)` instead of `{resugar_map}`

kimgr: Hmm, maybe it's possible to do `TemplateInstantiationData{resugar_map}` even without explicit constructors?

bolshakov-a: Yes, it is.

kimgr: Ah, cool. Yeah, my main priority is that the type is visible where used. I still rely 100% on `grep` for navigating code, so I'm a big fan of type names :)

bolshakov-a: Btw, I've noticed recently that writing
```cpp
for (const std::pair<KeyT, ValueT>& item : map)
```
leads to silent data copying because `std::map::value_type` is std::pair<**const** KeyT, ValueT>. So, I prefer automatic type deduction.

iwyu_ast_util.cc
```diff
@@ -963,28 +968,38 @@ map<const Type*, const Type*> GetTplTypeResugarMapForFunction(
...
+      resugar_map[desugared_type] = type;
       if (desugared_type != type) {
         VERRS(6) << "Remapping template arg of interest: "
                  << PrintableType(desugared_type) << " -> "
                  << PrintableType(type) << "\n";
       }
     }
+    // TODO(bolshakov): more fine-grained determination of provided types could
+    // e.g. consider only arguments that actually take part in type deduction.
+    // Simple moving the line below into the if-statement body above doesn't
```
kimgr, line 984: Typo: Simple -> Simply

iwyu_ast_util.h
```diff
@@ -821,17 +830,21 @@ bool HasImplicitConversionConstructor(const clang::Type* type);
 // We also include mappings for component types: if we have an entry
 // 'vector<TypedefType>' -> 'vector<Foo>', we also add an entry
 // 'TypedefType' -> 'Foo'.
+// Besides, the returned data contains a set of "provided" types (e.g. from
+// a typedef being type template argument).
 // For ease of calling, this accept any type, but will return an empty
-// map for any input that's not a template specialization type.
-map<const clang::Type*, const clang::Type*> GetTplTypeResugarMapForClass(
-    const clang::Type* type);
+// data for any input that's not a template specialization type.
```
kimgr, line 836: Maybe "an empty result" instead of "an empty data" here?

iwyu_ast_util.cc
bolshakov-a: I used "simple" as an adjective to "moving", not an adverb, but I can be wrong, of course.

kimgr: Then I would have expected "Simple moving **_of_** the line...". But my Swedish English is not exactly authoritative :).

I think the "simply" wording is simpler in either case.

iwyu_ast_util.cc
```diff
@@ -888,19 +887,23 @@ static const Type* GetSugaredTypeOf(const Expr* expr) {
   return v.sugared.getTypePtr();
 }
 
-map<const Type*, const Type*> GetTplTypeResugarMapForFunction(
-    const FunctionDecl* decl, const Expr* calling_expr) {
-  map<const Type*, const Type*> retval;
+TemplateInstantiationData GetTplInstDataForFunction(
+    const FunctionDecl* decl, const Expr* calling_expr,
+    std::function<set<const clang::Type*>(const clang::Type*)>
```
kimgr, line 892: `clang::Type` already seems to be imported into the current namespace, so you should be able to simplify this signature just using `Type`

iwyu_ast_util.cc
```diff
@@ -1441,12 +1456,14 @@ bool HasImplicitConversionConstructor(const Type* type) {
   return HasImplicitConversionCtor(cxx_class);
 }
 
-map<const clang::Type*, const clang::Type*>
-GetTplTypeResugarMapForClassNoComponentTypes(const clang::Type* type) {
+TemplateInstantiationData GetTplInstDataForClassNoComponentTypes(
```
kimgr, line 1459: Same here with `clang::Type` -> `Type`. I think it's a nice change for readability with the additional `function<>` even if it's not strictly necessary.

iwyu_ast_util.cc
```diff
@@ -1503,13 +1524,19 @@ GetTplTypeResugarMapForClassNoComponentTypes(const clang::Type* type) {
...
 
-  return retval;
+  return TemplateInstantiationData{retval, provided_types};
 }
 
-map<const clang::Type*, const clang::Type*> GetTplTypeResugarMapForClass(
-    const clang::Type* type) {
-  return ResugarTypeComponents(  // add in the decomposition of retval
-      GetTplTypeResugarMapForClassNoComponentTypes(type));
+TemplateInstantiationData GetTplInstDataForClass(
```
kimgr, line 1530: ... and `clang::Type` -> `Type` here too

iwyu_ast_util.h
```diff
@@ -15,6 +15,7 @@
 #include <map>                          // for map
 #include <set>                          // for set
 #include <string>                       // for string
+#include <functional>                   // for function
```
kimgr, line 18: Please move to the top; includes should be sorted

kimgr:
:+1:  Thanks!

Issue #1121, opened 2022-10-06
Invalid suggestion for templated factories, defaulted arguments and indirect includes


graveljp:
IWYU produces invalid recommendations when using `make_unique`-style templates to create objects whose constructor has parameters with default values. See a minimalist example at: https://github.com/graveljp/iwyu-bugs/tree/main/default_param

You can reproduce by doing:
```shell
git clone https://github.com/graveljp/iwyu-bugs.git
cd iwyu-bugs/default_param
include-what-you-use main.cc
```

This outputs:
```
main.cc should add these lines:
#include "bar.h"   // for MakeBar

main.cc should remove these lines:

The full include-list for main.cc:
#include "bar.h"   // for MakeBar
#include "foo.h"   // for Foo
#include "make.h"  // for make
---
```

`main.cc` shouldn't be including `bar.h`. This seems to be caused by a missing header in `foo.h`, adding `#include "bar.h"` there fixes the problem. While this is easily worked around, finding where the missing include is can be tricky and doesn't scale to very large code base that never used IWYU (i.e. with missing `includes` all over the code base).

Tests at on latest master (IWYU at https://github.com/include-what-you-use/include-what-you-use/commit/c1d8dd310a3b2e0601d44ece73fba9bc037e82d1 and LLVM at https://github.com/llvm/llvm-project/commit/bc4bcbcfc820b324f680e8f260691c38052eedc9).

graveljp:
It's worth pointing out that clangd (in [UnusedIncludes: Strict](https://clangd.llvm.org/guides/include-cleaner#configuration) mode) also agrees that `main.cc` shouldn't be including `bar.h`.

kimgr:
Thanks for the excellent repro case! I know IWYU currently has some confusion around attribution for template instantiations, but I'm not 100% sure of the details. I'll try to look into this, but there's some other stuff in the pipeline (e.g. getting a release out compatible with clang-15), so I can't promise when.

graveljp:
I found a simpler case similar to this one, this time involving a defaulted template parameter. See here: https://github.com/graveljp/iwyu-bugs/tree/main/default_template_param. Steps to reproduce:
```
git clone https://github.com/graveljp/iwyu-bugs.git
cd iwyu-bugs/default_template_param
include-what-you-use main.cc
```

This outputs:
```
main.cc should add these lines:
#include "functors.h"    // for Identity
```

[...]

PR #1397, opened 2024-01-02
Handle function template default type arguments

bolshakov-a:
Besides adding default type arguments of function templates into `resugar_map`, this PR introduces short-circuiting of handling a function reference when such a reference occur due to calling that function. Previously, that handling was just redundant. But after turning default argument analysis on, it started causing a bug with unwanted default type template argument reporting when it is actually overridden by the deduced one.

The problem arises because there is a lack of the full context for `resugar_map` and provided type set determination: actual function call arguments are not considered on function reference analyzing. Hence, a default type argument of a function template may be placed into `resugar_map` as defaulted despite being overridden by some of function call argument types.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -122,6 +122,7 @@
 #include "clang/AST/DeclBase.h"
 #include "clang/AST/DeclCXX.h"
 #include "clang/AST/DeclTemplate.h"
+#include "clang/AST/Expr.h"
 #include "clang/AST/ExprConcepts.h"
 #include "clang/AST/NestedNameSpecifier.h"
 #include "clang/AST/OperationKinds.h"
@@ -798,6 +799,11 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
     if (FunctionDecl* fn_decl = DynCastFrom(expr->getDecl())) {
       if (!IsImplicitlyInstantiatedDfn(fn_decl))
         return true;
+      if (const auto* call_expr =  // Skip intermediate ImplicitCastExpr node.
+          current_ast_node_->GetAncestorAs<CallExpr>(2)) {
+        if (call_expr->getDirectCallee() == fn_decl)
+          return true;
+      }
       // If fn_decl has a class-name before it -- 'MyClass::method' --
       // it's a method pointer.
       const Type* parent_type = nullptr;
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -76,6 +76,7 @@ using clang::Expr;
 using clang::ExprWithCleanups;
 using clang::FullSourceLoc;
 using clang::FunctionDecl;
+using clang::FunctionTemplateDecl;
 using clang::FunctionTemplateSpecializationInfo;
 using clang::FunctionType;
 using clang::ImplicitCastExpr;
@@ -109,6 +110,7 @@ using clang::TemplateName;
 using clang::TemplateParameterList;
 using clang::TemplateSpecializationKind;
 using clang::TemplateSpecializationType;
+using clang::TemplateTypeParmDecl;
 using clang::TranslationUnitDecl;
 using clang::Type;
 using clang::TypeAliasTemplateDecl;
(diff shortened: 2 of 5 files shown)


iwyu_ast_util.cc
```diff
@@ -888,6 +890,29 @@ static const Type* GetSugaredTypeOf(const Expr* expr) {
   return v.sugared.getTypePtr();
 }
 
+static map<const Type*, const Type*> GetDefaultedArgResugarMap(
+    const FunctionDecl* decl) {
+  map<const Type*, const Type*> res;
+  const FunctionTemplateDecl* template_decl = decl->getPrimaryTemplate();
+  if (!template_decl)
+    return map<const Type*, const Type*>{};
```
kimgr, line 898: `return res;`

kimgr (commented):
This looks like it might be related and might be fixed after these changes: https://github.com/include-what-you-use/include-what-you-use/issues/1121

kimgr:
> This looks like it might be related and might be fixed after these changes: https://github.com/include-what-you-use/include-what-you-use/issues/1121

I meant to ask: is there a testcase there you could salvage? The tests you have look good, but there might be corner cases in @graveljp's suite.

bolshakov-a:
The problems described in #1121 occur to be different from this one, despite the second one is related.

tests/cxx/default_tpl_arg.cc
```diff
@@ -43,6 +43,25 @@ struct Outer2 {
 // IWYU: IndirectTemplate needs a declaration
 Outer2<int, IndirectTemplate<int>> o2;
 
+void Fn() {
+  // IWYU: FnWithDefaultTplArg is...*default_tpl_arg-i1.h
+  // IWYU: IndirectClass is...*indirect.h
```
kimgr, line 48: So are these attributed here because of the reexport hint (`class IndirectClass;`) before the definition? I think it would be nice to show the "default" behavior as well, where `IndirectClass` should be attributed to the template file... right?

I was playing with graveljp's example here: https://github.com/graveljp/iwyu-bugs/tree/main/default_template_param, but it still reports the default template type at point of instantiation.

bolshakov-a: > I was playing with graveljp's example here: https://github.com/graveljp/iwyu-bugs/tree/main/default_template_param, but it still reports the default template type at point of instantiation.

Fixed in #1403. Probably, the test case there is sufficient? Or there should be some without normal default arguments, just with template default arguments?

kimgr: Could you add a comment in the test explaining why `IndirectClass` is required at point of instantiation? It looks wrong to me, unless it's due to the forward-decl near definition.

Once I understand that I'll be in a better position to think about more tests :)

iwyu.cc
```diff
@@ -798,6 +799,11 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
     if (FunctionDecl* fn_decl = DynCastFrom(expr->getDecl())) {
       if (!IsImplicitlyInstantiatedDfn(fn_decl))
         return true;
+      if (const auto* call_expr = current_ast_node_->GetAncestorAs<CallExpr>(
```
kimgr, line 802: The formatting on this block is pretty horrible. Seems hard to shorten it, but can you trick `clang-format` by making it longer? E.g. use `current_ast_node()` instead of the raw member?

You could also consider moving the comment line down into the body of the if, but I don't think that alone helps with the wrapping.

tests/cxx/default_tpl_arg.cc
```diff
@@ -43,6 +44,41 @@ struct Outer2 {
...
+  using ProvidingAlias = IndirectClass;
+  ProvidingAlias* p = 0;
+  // IWYU: NonProvidingAlias is...*default_tpl_arg-i1.h
+  NonProvidingAlias* n = 0;
+
+  // IWYU: FnWithNonProvidedDefaultTplArgAndDefaultCallArg is...*default_tpl_arg-i1.h
+  // IWYU: IndirectClass is...*indirect.h
+  FnWithNonProvidedDefaultTplArgAndDefaultCallArg();
+  // TODO: should not report here.
```
bolshakov-a, line 67: Fixed in #1403. The separate test cases there become probably redundant.

tests/cxx/default_tpl_arg.cc
bolshakov-a: In fact, instantiation-side analysis doesn't consider forward declarations currently. That's the reason for redundancy on the first IWYU pass when template definition header neither `#include`s nor forward-declares default template argument type. It is suggested for full type info inclusion on the both sides.

kimgr:
@bolshakov-a Yeah, this LGTM. Are you ready to merge?

bolshakov-a:
I think so.

PR #1398, opened 2024-01-02
Add pragmas to suppress IWYU warnings in IWYU itself

kimgr:
To run IWYU on IWYU, we can simply do:

* mkdir build && cd build
* cmake -DCMAKE_GENERATE_EXPORT_COMPILE_COMMANDS ...
* cd ..
* iwyu_tool.py -p build/ *.cc | ./fix_includes.py ...

I ran these steps and worked through the results. I found some recurring IWYU mistakes:

* Some headers are requested though they don't appear to be used. I think this is mostly due to "template spill", i.e. IWYU fails to identify whether a type should be provided by the template or covered by the instantiation (clang/AST/Redeclarable.h, clang/AST/StmtIterator.h, clang/AST/UnresolvedSet.h, clang/Basic/CustomizableOptional.h, llvm/ADT/iterator.h)
* PPCallbacks is thrown around in a std::unique_ptr, and IWYU fails to understand that a forward-declaration will suffice for that.
* std::pair appears to somehow confuse IWYU; it's attributed to both `<utility>` and `<iterator>`
* The getopt family is attributed to `<unistd.h>`. While that is correct according to the Linux manpage, it is not portable. We have a portable getopt in tree.
* Clang exposes a set of aliases for LLVM's Support library, and that appears to confuse IWYU so it suggests clang/Basic/LLVM.h even if better LLVM headers are already available.
* In iwyu_stl_util.h, we include <algorithm> for std::find. Unfortunately std::find is only used in a single, uninstantiated template, so IWYU can't see it (search issues for "UnresolvedLookupExpr" for more details).
* Similar, but different, in iwyu_verrs.h, the only use of raw_ostream is in an unexpanded macro.

Use IWYU pragmas to deal with these weaknesses:

* "pragma: no_include" to avoid pulling in implementation detail headers misdiagnosed by IWYU
* "pragma: keep" to force IWYU to retain misdiagnosed forward declarations and #includes
* "pragma: export" to let iwyu_verrs.h provide llvm::errs()

Once this is in place, we should be able to run a single IWYU pass over all IWYU's source files without any glaring mistakes.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -140,9 +140,20 @@
 #include "llvm/Support/TargetSelect.h"
 #include "llvm/Support/raw_ostream.h"
 
+// TODO: Clean out pragmas as IWYU improves.
+// IWYU pragma: no_include "clang/AST/Redeclarable.h"
+// IWYU pragma: no_include "clang/AST/StmtIterator.h"
+// IWYU pragma: no_include "clang/AST/UnresolvedSet.h"
+// IWYU pragma: no_include "clang/Basic/CustomizableOptional.h"
+// IWYU pragma: no_include "clang/Lex/PPCallbacks.h"
+// IWYU pragma: no_include "llvm/ADT/iterator.h"
+// IWYU pragma: begin_keep
 namespace clang {
 class PPCallbacks;
+}  // namespace clang
+// IWYU pragma: end_keep
 
+namespace clang {
 namespace driver {
 class ToolChain;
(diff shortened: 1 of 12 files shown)


firewave:
I think it would make sense to add this to the CI.

kimgr:
@firewave That's the plan -- I'm just building up slowly to make sure the first run will have no complaints.

PR #1400, opened 2024-01-03
Run IWYU on IWYU (with why-comments)

kimgr:
This is the unadorned output of:

    $ ./iwyu_tool.py -p .../out/main+deb *.cc -- \
        -Xiwyu --check_also="*/include-what-you-use/iwyu_*.h" | \
        ./fix_includes.py --nosafe_headers --reorder --comments
    (skipping .../include-what-you-use/iwyu_port.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_use_flags.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_string_util.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_stl_util.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_verrs.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_driver.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_getopt.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_getopt.cc: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_regex.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_version.h: iwyu reports no contentful changes)
    >>> Fixing #includes in '.../include-what-you-use/iwyu_ast_util.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_globals.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_location_util.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_path_util.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_ast_util.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_cache.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_cache.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_lexer_utils.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_output.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_preprocessor.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_driver.cc'
[...]

PR #1401, opened 2024-01-03
Run IWYU on IWYU

kimgr:
This is the unadorned output of:

    $ ./iwyu_tool.py -p .../out/main+deb *.cc -- \
      -Xiwyu --check_also="*/include-what-you-use/iwyu_*.h" | \
      ./fix_includes.py --nosafe_headers --reorder
    (skipping .../include-what-you-use/iwyu_port.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_use_flags.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_string_util.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_stl_util.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_verrs.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_driver.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_getopt.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_getopt.cc: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_regex.h: iwyu reports no contentful changes)
    (skipping .../include-what-you-use/iwyu_version.h: iwyu reports no contentful changes)
    >>> Fixing #includes in '.../include-what-you-use/iwyu_ast_util.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_globals.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_location_util.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_path_util.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_ast_util.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_cache.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_cache.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_lexer_utils.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_output.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_preprocessor.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_driver.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_include_picker.h'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_globals.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_include_picker.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_lexer_utils.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_location_util.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_output.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_path_util.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_preprocessor.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_regex.cc'
    >>> Fixing #includes in '.../include-what-you-use/iwyu_verrs.cc'
    IWYU edited 22 files on your behalf.

Some added #includes are suboptimal (`<std*.h>` instead of `<cstd*>`), but
I'll clean those up manually after.

No functional change.


kimgr:
@bolshakov-a @kilroyd @firewave Please take a look and let me know if you prefer this PR (without `// for xyz` comments), or #1400, which has them.

I think there's merit to both: the diff is much easier to read here, but the end result is somewhat more informative in the other one. Thoughts?

The longer-term goal is to self-IWYU on PRs and post an informative comment with the diff or something (I don't want to break the build on IWYU changes).

kimgr:
One practical benefit I'm hoping to get out of this is that we'll be running IWYU on real code, so we have a better chance of catching crashes or big regressions than with the test suite (which might inspire better test coverage for new changes).

firewave:
> Please take a look and let me know if you prefer this PR (without // for xyz comments)

I don't think the comments are necessary in the code. Depending on the code changes that would also generate quite some noise in the comments and would require constant modifications. If there is a workflow which runs this on the code you could simply keep the output with that information.

> One practical benefit I'm hoping to get out of this is that we'll be running IWYU on real code, so we have a better chance of catching crashes or big regressions than with the test suite (which might inspire better test coverage for new changes).

You just always dogfood/bootstrap your code.

As a further step it might also make sense to run the code against `clang-include-cleaner`.

> The longer-term goal is to self-IWYU on PRs and post an informative comment with the diff or something (I don't want to break the build on IWYU changes).

Workflows give you the possibility to add comments to a PR (have never done that before but should be easy) so you could add the diff (in a `<details>` tag) in that way without failing the job.

bolshakov-a:
> I don't think the comments are necessary in the code. Depending on the code changes that would also generate quite some noise in the comments and would require constant modifications. If there is a workflow which runs this on the code you could simply keep the output with that information.

Agreed.

> As a further step it might also make sense to run the code against `clang-include-cleaner`.

It hardly gives the same output as IWYU.

LGTM.

kilroyd:
My personal preference would be to include the comments - otherwise it's hard to double check whether the addition/removal of any include makes sense without re-running IWYU locally (or knowing the codebase really well).

Still, if IWYU has recommended these changes its worth committing this as is, and maybe getting the comment formatting into better shape so people are happier to include them. (e.g. by aligning differently per block to maybe get more in, always making the first identifier a full symbol, ...)

Re noise from comments changing constantly. If we can make things so the comments could be 'stable' that would be good. Maybe something like output most common identifiers first, upto the line limit. But on running fix_includes, list all the identifiers, and only change the comment if an identifier is no longer used.

Anyway, LGTM

kimgr:
> Re noise from comments changing constantly. If we can make things so the comments could be 'stable' that would be good.
 > Maybe something like output most common identifiers first, upto the line limit. But on running fix_includes, list all the identifiers, and only change the comment if an identifier is no longer used.

I think all these things are already in place, actually :)

To me the biggest friction is when adding a new include, there's not enough incentive to get the comment exactly right.

I think what I'll try to do is include two parts to the output when running on a PR:
* the raw IWYU output (with comments)
* the diff when that output is applied (without comments)

That way we'll have enough information to dig around without necessarily having to re-run IWYU locally.

firewave:
> I think what I'll try to do is include two parts to the output when running on a PR:
> 
>     * the raw IWYU output (with comments)
> 
>     * the diff when that output is applied (without comments)

That was what I was thinking of. Store the previous as an artifact and post the latter as a comment.

firewave:
> > As a further step it might also make sense to run the code against `clang-include-cleaner`.
> 
> It hardly gives the same output as IWYU.

Still it would be good to be able to spot potential issues - on either side.

PR #1402, opened 2024-01-03
Short-circuit default argument expression traversal

bolshakov-a:
There is no point in default argument analysis at function call site because all the needed types and declarations should be provided by the function definition header. IWYU reports them during `ParmVarDecl` traversal.

This fixes the original problem reported in #1121.

### DIFF
--- a/tests/cxx/no_def_arg_call_site-d1.h
+++ b/tests/cxx/no_def_arg_call_site-d1.h
@@ -1,0 +1,28 @@
+//===--- no_def_arg_call_site-d1.h - test input file for iwyu -------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/no_def_arg_call_site-i1.h"
+
+struct Struct {
+  // IWYU: GetInt is...*no_def_arg_call_site-i2.h
+  static void FnWithDefArg(int = GetInt());
+};
+
+/**** IWYU_SUMMARY
+
+tests/cxx/no_def_arg_call_site-d1.h should add these lines:
+#include "tests/cxx/no_def_arg_call_site-i2.h"
+
+tests/cxx/no_def_arg_call_site-d1.h should remove these lines:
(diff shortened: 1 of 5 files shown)


tests/cxx/no_def_arg_call_site.cc
```diff
@@ -0,0 +1,30 @@
...
+#include "tests/cxx/no_def_arg_call_site-d1.h"
+
+template <typename T>
+void TplFn() {
+  T::FnWithDefArg();
+}
+
+void Fn() {
+  TplFn<Struct>();
+}
```
kimgr, line 24: Would it be prudent to test `Struct::FnWithDefArg();` directly here? Just to have coverage outside of template too.

bolshakov-a: Without template context, symbols used inside default argument expression are correctly reported to the function declaration location, so there was no observable bug, just redundancy in traversal. `InstantiatedTemplateVisitor` reports them to the instantiation location.

tests/cxx/no_def_arg_call_site.cc
```diff
@@ -0,0 +1,30 @@
...
+
+/**** IWYU_SUMMARY
+
+(tests/cxx/no_def_arg_call_site.cc has correct #includes/fwd-decls)
```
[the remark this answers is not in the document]
bolshakov-a: I thought about it and decided that the test should indeed test that there is no unwanted diagnostics, and no unwanted header suggestions.

But I've not found any decl-side test even in `badinc`. It is good idea to test it as well.

tests/cxx/no_def_arg_call_site.cc
kimgr: Yeah, I mean it would be nice to show that here, for completeness.

tests/cxx/no_def_arg_call_site-d1.h
```diff
@@ -0,0 +1,28 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/no_def_arg_call_site-i1.h"
+
+struct Struct {
+  // IWYU: GetInt is...*no_def_arg_call_site-i2.h
```
kimgr, line 13: Nice to get this covered here!

PR #1403, opened 2024-01-03
Ignore default function call arguments on "type deduction"

bolshakov-a:
Function call default arguments don't take part in deducing type template parameters, and should be ignored on `resugar_map` building.

This fixes the second problem reported in #1121.

### DIFF
--- a/tests/cxx/default_tpl_arg-d2.h
+++ b/tests/cxx/default_tpl_arg-d2.h
@@ -22,6 +22,16 @@ void FnWithProvidedDefaultTplArg() {
 // 'IndirectClass' type because this header '#include's it directly and hence
 // provides.
 template <typename T = IndirectClass>
-void FnWithProvidedDefaultTplArgAndDefaultCallArg(T* = nullptr) {
+void FnWithProvidedDefaultTplArgAndDefaultCallArg1(T* = nullptr) {
+  T t;
+}
+
+template <typename T = IndirectClass>
+void FnWithProvidedDefaultTplArgAndDefaultCallArg2(T = T{}) {
+  T t;
+}
+
+template <typename T = IndirectClass>
+void FnWithProvidedDefaultTplArgAndDefaultCallArg3(T = {}) {
   T t;
 }
--- a/tests/cxx/default_tpl_arg.cc
+++ b/tests/cxx/default_tpl_arg.cc
@@ -64,19 +64,22 @@ void Fn() {
   // IWYU: FnWithNonProvidedDefaultTplArgAndDefaultCallArg is...*default_tpl_arg-i1.h
   // IWYU: IndirectClass is...*indirect.h
   FnWithNonProvidedDefaultTplArgAndDefaultCallArg();
-  // TODO: should not report here.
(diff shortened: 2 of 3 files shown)


tests/cxx/default_tpl_arg.cc
```diff
@@ -43,6 +43,18 @@ struct Outer2 {
 // IWYU: IndirectTemplate needs a declaration
 Outer2<int, IndirectTemplate<int>> o2;
 
+// IWYU: IndirectClass is...*indirect.h
+template <typename T = IndirectClass>
+// IWYU: IndirectClass is...*indirect.h
+void FnWithDefArgs(T = IndirectClass{}) {
```
kimgr, line 49: I think it would make more sense with `T = T()` (or `T = T{}` if you insist on the curlies :smile:) -- for two reasons:

* No need to `// IWYU:` capture the use
* Probably more closely mirrors what a real-world usecase would look like

Oh, by the way, the graveljp example also had the equivalent of `T = {}`, which might be nice to test separately.

bolshakov-a:
Oh, wait, there is the difference between default template argument handling between template definition and instantiation sites which [I've mentioned earlier](https://github.com/include-what-you-use/include-what-you-use/pull/1387). Due to that, the test case fails after rebasing #1397 onto this. I'd move the template better to a header including `indirect.h` directly.

tests/cxx/default_tpl_arg.cc
```diff
@@ -43,17 +43,23 @@ struct Outer2 {
...
+}
+
 /**** IWYU_SUMMARY
 
 tests/cxx/default_tpl_arg.cc should add these lines:
 #include "tests/cxx/default_tpl_arg-i1.h"
 #include "tests/cxx/indirect.h"
 
 tests/cxx/default_tpl_arg.cc should remove these lines:
-- #include "tests/cxx/default_tpl_arg-d1.h"  // lines XX-XX
```
kimgr, line 53: Hmm, it's a little sad that this is no longer removed in favor of `-i1.h`. Could you instead add a new `-d2.h` header for the new test?

kimgr:
Yep, LGTM! Can you rebase on latest master after #1397?

tests/cxx/default_tpl_arg-d2.h
```diff
@@ -25,3 +25,13 @@ template <typename T = IndirectClass>
...
   T t;
 }
+
+template <typename T = IndirectClass>
+void FnWithDefArgs1(T = T{}) {
+  T t;
+}
+
+template <typename T = IndirectClass>
+void FnWithDefArgs2(T = {}) {
```
bolshakov-a, line 35: Maybe, they should be renamed, if still needed.

kimgr: Aha, you mean they are redundant now because the previous PR already has all combinations of default template args x default function args? Do whatever you feel is best, I can't really visualize the alternatives.

kimgr: Nice :+1:

PR #1405, opened 2024-01-03
Use C++ C compatibility headers consistently

kimgr:
Use `<cxxx>` instead of `<xxx.h>` in the IWYU code base.

The only exception is iwyu_getopt.cc, which is C code vendored from https://github.com/kimgr/getopt_port, so we want to stay as close to it as possible.


kimgr:
Once this lands, running IWYU on itself is stable and suggests no changes.

PR #1406, opened 2024-01-04
WIP: IWYU dogfood in CI

kimgr:
Not sure this will go anywhere, but worth a try.


kimgr:
@firewave Phew! That was a lot worse than I expected :smile:. I don't think there is a way to grant PR write access to the `GITHUB_TOKEN` when executing in the context of a PR.

I'm going to take a stab at setting up a personal access token, and see if I can use that from the GH action instead.

kimgr:
Closed in favor of #1408

PR #1409, opened 2024-01-04
include-what-you-use/iwyu-dogfood-ci

kimgr:
Let's see if the workflow_run triggers if the PR is on a branch in the same repo.


kimgr:
Managed to get the out-of-repo PR to work, so this is no longer interesting.

Issue #1301, opened 2023-09-06
False positives with `std::map` and alias declaration


carlosgalvezp:
Hi!

Consider the following example:

```
// foo.h

#include <map>
#include <string>

using MyMap = std::map<std::string, int>;
```

```
// main.cpp
#include "foo.h"

#include <iostream>

void foo(MyMap const& map)
{
  for (auto const& x : map)
  {
    std::cout << x.first << "," << x.second << std::endl;
  }
}
```
Running IWYU gives:

```
main.cpp:7:22: warning: std::_Rb_tree_const_iterator is defined in <map>, which isn't directly #included.
main.cpp:7:22: warning: std::operator!= is defined in <map>, which isn't directly #included.
main.cpp:9:15: warning: std::operator<< is defined in <string>, which isn't directly #included.
main.cpp:9:20: warning: std::pair is defined in <utility>, which isn't directly #included.
main.cpp:9:38: warning: std::pair is defined in <utility>, which isn't directly #included.

main.cpp should add these lines:
#include <map>       // for operator!=, _Rb_tree_const_iterator
#include <string>    // for operator<<
#include <utility>   // for pair

main.cpp should remove these lines:

The full include-list for main.cpp:
#include <iostream>  // for basic_ostream, char_traits, basic_ostream::operator<<, operator<<, endl, cout
#include <map>       // for operator!=, _Rb_tree_const_iterator
#include <string>    // for operator<<
#include <utility>   // for pair
#include "foo.h"     // for MyMap
---
```

While I would agree that `string` and `utility` might technically be needed I believe re-including `map` in `main.cpp` is unreasonable, since it leaks in implementation details from libstdc++.

This test is done on Ubuntu 20 + GCC 9 + IWYU trunk.

Thanks!

kimgr:
As far as I can tell, this is fixed on main. The results aren't perfectly consistent between libstdc++ and libc++, but close enough, IMHO:
```
$ include-what-you-use -stdlib=libstdc++ tests/bugs/1301/1301.cc 

tests/bugs/1301/1301.cc should add these lines:
#include <string>    // for operator<<, basic_string
#include <utility>   // for pair

tests/bugs/1301/1301.cc should remove these lines:

The full include-list for tests/bugs/1301/1301.cc:
#include <iostream>  // for char_traits, basic_ostream, basic_ostream::opera...
#include <string>    // for operator<<, basic_string
#include <utility>   // for pair
[...]

PR #1411, opened 2024-01-04
Fix dereferencing of forwarding references

bolshakov-a:
The code in `ReportTypeUseInternal` was designed with assumption that clang collapses subsequent reference types into a single one. But it turns out that `getPointeeTypeAsWritten()` may return another reference type when applied to a reference type produced with forwarding reference syntax. The subsequent `IsPointerOrReferenceAsWritten` check short-circuited referenced type reporting despite it should be reported. This problem is solved here with recursive `ReportTypeUseInternal` calls from dereferencing branches.

The explicit traversal of `CXXForRangeStmt` is introduced to avoid regression with code like this:

```cpp
using ProvidingRefAlias = Container&;
ProvidingRefAlias ref_provided = ...;
for (int x : ref_provided) ...
```

During initialization of implicit range variable, the specific providing type occurs to be lost. That variable is initialized like being declared with forwarding reference, and didn't show the bug prior to this fix of forwarding references. I've decided to write explicit traversal without handling for-loop implicit variables to avoid unwanted reporting of the provided type. The code does essentially the same that default implementation does when `shouldVisitImplicitCode()` returns `false`. In principle, it could allow more finely-grained control over reporting use of iterator type and related operators in the future.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1761,6 +1761,34 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  // The explicit traversal of 'CXXForRangeStmt' is introduced to avoid
+  // a problem with code like this:
+  //
+  // using ProvidingRefAlias = Container&;
+  // ProvidingRefAlias ref_provided = ...;
+  // for (int x : ref_provided) ...
+  //
+  // After initialization of implicit range variable, the specific providing
+  // type is desugared. Due to this, it is reported e.g. on handling
+  // the implicit call of 'begin()' method despite being provided by alias.
+  // Explicit traversal without handling for-loop implicit variables allows
+  // to avoid such an unwanted reporting of the provided type. The code does
+  // essentially the same that the default implementation does when
+  // 'shouldVisitImplicitCode()' returns 'false'.
+  bool TraverseCXXForRangeStmt(CXXForRangeStmt* stmt) {
+    if (!this->getDerived().WalkUpFromCXXForRangeStmt(stmt))
+      return false;
+    if (Stmt* init_stmt = stmt->getInit()) {
+      if (!this->getDerived().TraverseStmt(init_stmt))
+        return false;
+    }
+    if (!this->getDerived().TraverseStmt(stmt->getLoopVarStmt()))
+      return false;
+    if (!this->getDerived().TraverseStmt(stmt->getRangeInit()))
+      return false;
+    return this->getDerived().TraverseStmt(stmt->getBody());
+  }
+
   // The type of the for-range-init expression is fully required, because the
   // compiler generates method calls to it, e.g. 'for (auto t : ts)' translates
   // roughly into 'for (auto i = std::begin(ts); i != std::end(ts); ++i)'.
@@ -2563,16 +2591,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     type = Desugar(type);
     if (deref_kind == DerefKind::RemoveRefs ||
         deref_kind == DerefKind::RemoveRefsAndPtr) {
-      if (const auto* ref_type = dyn_cast_or_null<ReferenceType>(type))
-        type = ref_type->getPointeeTypeAsWritten().getTypePtr();
-      // Keep deref_kind to remove possible additional references in type alias
-      // chain.
-      // TODO(bolshakov): shouldn't Desugar be called here and below as well?
+      if (const auto* ref_type = dyn_cast_or_null<ReferenceType>(type)) {
+        // Keep deref_kind to remove possible additional references in type
+        // alias chain.
+        return ReportTypeUseInternal(
+            used_loc, ref_type->getPointeeTypeAsWritten().getTypePtr(),
+            blocked_types, deref_kind);
+      }
     }
     if (deref_kind == DerefKind::RemoveRefsAndPtr) {
       if (const auto* ptr_type = dyn_cast_or_null<PointerType>(type)) {
-        type = ptr_type->getPointeeType().getTypePtr();
-        deref_kind = DerefKind::None;
+        return ReportTypeUseInternal(used_loc,
+                                     ptr_type->getPointeeType().getTypePtr(),
+                                     blocked_types, DerefKind::None);
       }
     }
 
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2646,11 +2646,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
                  << typedef_decl->getQualifiedNameAsString()
                  << " owns the underlying type:\n";
         // If any of the used types are themselves typedefs, this will
-        // result in a recursive expansion.  Note we are careful to
-        // recurse inside this class, and not go back to subclasses.
+        // result in a recursive expansion.
         const Type* type = typedef_decl->getUnderlyingType().getTypePtr();
-        IwyuBaseAstVisitor<Derived>::ReportTypeUseInternal(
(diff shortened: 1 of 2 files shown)


iwyu.cc
```diff
@@ -1761,6 +1761,20 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  bool TraverseCXXForRangeStmt(CXXForRangeStmt* stmt) {
```
kimgr, line 1764: It might be nice to add a short description as to why this is necessary (from the commit message) as a code comment here.

And number two; I wonder if there's a way to detect at reporting time that the use is in implicit code and discard it there instead? I guess I don't 100% understand what the AST looks like in this case.

kimgr:
Commit message typo, btw:
> During initialization of implicit range variable, the specific providing type occurs to be lost. 

Should probably be `... appears to be lost`

kimgr:
I just merged the `IWYU dogfood` job to master. If you rebase on master, you should be able to see if this PR has any effects.

iwyu.cc
bolshakov-a: I'm not sure that inspecting AST node stack in order to search `CXXForRangeStmt` node from reporting visitors directly would be simpler (besides that it could affect performance). But the following works:
```cpp
bool TraverseDeclStmt(clang::DeclStmt* stmt) {
  if (const auto* range_for =
          current_ast_node()->template GetParentAs<CXXForRangeStmt>()) {
    if (stmt->getSingleDecl()->isImplicit() &&
        range_for->getRangeStmt() != stmt) {
      return true;
    }
  }
  return Base::TraverseDeclStmt(stmt);
}
```

bolshakov-a: > I guess I don't 100% understand what the AST looks like in this case.

The code
```cpp
for (char ch; int x : container)
  float f;
```
is equivalent to
```cpp
char ch;
auto && __range1 = container;
auto __begin1 = __range1.begin();
auto __end1 = __range1.end();
for ( ; __begin1 != __end1; ++__begin1) {
  int x = *__begin1;
  float f;
}
```
hence produces such an AST:
изображение

The problem arises on reporting type of base expression of `CXXMemberCallExpr`.

Note that "InitExpr" still needs to be traversed because it is explicitly written in the source despite it initializes the `__range1` implicit variable.

bolshakov-a: In fact, the iterator type info and iterator operator declarations are also needed, of course, but it is rare situation when they are not provided by the container class header, and reporting them may just produce unwanted noise in the output.

bolshakov-a: Yes, I thought about it as well, but it breaks at least this case:
```cpp
void TakeByRef(const Indirect&);

void TestImplicit(Indirect& a){
  int array[] = {1,2,3};
  for (int x : array) {
    [=] {
        TakeByRef(a);
    }();
  }
}
```
The complete `Indirect` type info is needed due to the _implicit_ capture by value.

iwyu.cc
```diff
@@ -1761,6 +1761,32 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  // The explicit traversal of 'CXXForRangeStmt' is introduced to avoid problem
```
kimgr, line 1764: Some notes on this comment:

* `avoid problem` -> 'avoid problems' or 'avoid a problem'
* `the specific providing type is lost` -- is this because it's been desugared by Clang? If so, maybe say "desugared" instead of "lost"
* It's not clear to me from the text what the higher-level "problem" is we're avoiding here? What is the bad thing that happens if we remove this custom traversal?

I checked out the branch and tried removing this code, and could see that:
```
  for (auto i : ref_provided)
    ;
```
started complaining about a full use of `Indirect`.

So I guess the problem is:

* Types in `CXXForRangeStmt` are implicitly desugared
* The default traversal sees these desugared types and reports them
* The actual type in the written code is a sugared alias type
* The underlying desugared type is provided from the alias definition loc (because it's an alias, and there are no author intent hints to the contrary), so it should never be reported in use locations

I guess this leaves me with the question: how does the non-providing ref [here](https://github.com/include-what-you-use/include-what-you-use/blob/master/tests/cxx/ptr_ref_aliases.cc#L76) get covered if we never visit it?

Or is it that the implicit part of the `CXXForRangeStmt` exposes desugared types and the explicit part the sugared ones?

bolshakov-a: > `the specific providing type is lost` -- is this because it's been desugared by Clang? If so, maybe say "desugared" instead of "lost"

I wanted to say: the type is desugared hence the information that it is provided and should not be reported is lost.

> how does the non-providing ref [here](https://github.com/include-what-you-use/include-what-you-use/blob/master/tests/cxx/ptr_ref_aliases.cc#L76) get covered if we never visit it?

It is reported from `VisitCXXForRangeStmt` handler ([you have written it](https://github.com/include-what-you-use/include-what-you-use/commit/2a474258599b22a3022b503975747658d55234da) especially for that). After fixing the issue with forwarding references, one can remove both `Visit...` and `Traverse...` handlers of `CXXForRangeStmt` and have the same output as with the sole `VisitCXXForRangeStmt`. (But it doesn't mean that `VisitCXXForRangeStmt` is not needed anymore, due to the abovementioned reason.)

kimgr:
I ran into https://github.com/include-what-you-use/include-what-you-use/issues/1301, and there seems to be a bit of overlap with this patch. I ran the test case with `include-what-you-use` built from this PR branch, but it still complains about missing `<map>`:
```c++
$ include-what-you-use tests/1301/main.cc

tests/1301/main.cc should add these lines:
#include <map>       // for _Rb_tree_const_iterator
#include <string>    // for operator<<
#include <utility>   // for pair

tests/1301/main.cc should remove these lines:

The full include-list for tests/1301/main.cc:
#include <iostream>  // for char_traits, basic_ostream, basic_ostream::opera...
#include <map>       // for _Rb_tree_const_iterator
#include <string>    // for operator<<
#include <utility>   // for pair
#include "foo.h"     // for MyMap
---
```

It's interesting that the trigger is `_Rb_tree_const_iterator` -- I think there's a similar issue if the iterator type is an alias, then we sometimes only see the desugared name.

On a whim, I added `std::_Rb_tree_const_iterator` to the hard-coded mappings here: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L1197 -- and true enough, that removed the requirement for `<map>`.

iwyu.cc
kimgr: Another random thought I had was that it would be cool if we could build an incremental resugar-map as we traverse the normal tree too, not just for templates. So e.g. `TraverseCXXForRangeStmt` could extract all types from all its immediate children and map desugared to whatever sugared is available. Then when we visit the children we could simply try-resugar from that map.

bolshakov-a:
Iterator type use is still reported during "LoopVarStmt" traversal.

iwyu.cc
bolshakov-a: That's interesting idea...

However, imagine that `begin()` and `end()` methods of some container class of some library return custom `Iterator` type. That type may even not be aliased, but the container-defining header almost definitely would "provide" it by including the corresponding header. The custom traversal of `CXXForRangeStmt` node could allow to decide whether report it or not, based on its "provision status".

kimgr: Makes sense, thanks!

kimgr:
> Iterator type use is still reported during "LoopVarStmt" traversal. To fix it, it is sufficient to replace that traversal with just
> reporting of the loop variable type. I think it would be better to do in a separate PR. 

Definitely separate, if at all, yes.

PR #1415, opened 2024-01-05
Fix -Wmissing-field-initializers warning

kimgr:
056396f424995a850b6ca27ba0f6251cdc43a013 added TemplateInstantiationData struct to bundle resugar map with provided types.

Clang complains if not all fields of a struct are initialized as part of an aggregate initializer; GCC doesn't.

Explicitly initialize both fields always.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -947,7 +947,7 @@ TemplateInstantiationData GetTplInstDataForFunction(
     resugar_map = GetTplTypeResugarMapForFunctionNoCallExpr(decl, 0);
     resugar_map = ResugarTypeComponents(
         resugar_map);  // add in resugar_map's decomposition
-    return TemplateInstantiationData{resugar_map};
+    return TemplateInstantiationData{resugar_map, {}};
   }
 
   // If calling_expr is a CXXConstructExpr of CXXNewExpr, then it's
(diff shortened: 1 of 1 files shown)


kimgr:
I stumbled on this in the CI build logs when I was working on dogfooding: https://github.com/include-what-you-use/include-what-you-use/actions/runs/7386223434/job/20092396917#step:7:389

kimgr:
Oops:
> include-what-you-use: /usr/lib/llvm-18/include/clang/AST/TemplateBase.h:299: QualType
> clang::TemplateArgument::getAsType() const: Assertion `getKind() == Type && "Unexpected kind"' failed.
>
> include-what-you-use: /usr/lib/llvm-18/include/clang/AST/TemplateBase.h:299: QualType
> clang::TemplateArgument::getAsType() const: Assertion `getKind() == Type && "Unexpected kind"' failed.

@bolshakov-a Using nested aggregate initializers like this appears to trigger an assertion :fearful:. I'll see if I can isolate it.

kimgr:
It's the `getAsType()` call here that triggers it: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_ast_util.cc#L923. I'm guessing the `ArgKind` is not `Type`, but one of the alternatives: https://clang.llvm.org/doxygen/classclang_1_1TemplateArgument.html#ad5158465bd96aebe0d9dc44207821220.

kimgr:
> Using nested aggregate initializers like this appears to trigger an assertion 😨. I'll see if I can isolate it.

Oops, no. The assertion was there before too, I just overlooked it in the logs. It's triggered by `std::make_unique<>`, I'll post an XFAIL testcase shortly.

kimgr:
@bolshakov-a https://github.com/include-what-you-use/include-what-you-use/pull/1416

bolshakov-a:
It's pretty good that "dogfooding" has found the crash earlier than users. @kimgr, nice job, thanks!

kimgr:
@bolshakov-a Yeah, that's exactly the kind of thing I was hoping for -- I just didn't expect it to happen so soon :)

PR #1416, opened 2024-01-05
Add XFAIL testcase for make_unique crasher

kimgr:
This can probably be reduced not to depend on <memory> and rewritten once the bug is fixed.

Tested to behave consistently with both libstdc++ and libc++.


bolshakov-a:
Interesting... Let me see what happens...

bolshakov-a:
Let's better merge #1417.

kimgr:
Yup, closed in favor of fix + really reduced testcase in #1417

PR #1421, opened 2024-01-06
Analyze template-nested typedef usage context

bolshakov-a:
Figure out if `typedef` or type alias referred type is actually provided by some of template specialization arguments when used outside of template instantiation (inside it `InstantiatedTemplateVisitor` should already do that job in `CanIgnoreType` method).

`IsProvidedByTplArg` routine can be further extended to work with alias templates as well (currently, even simplest cases like `template <typename T> using Identity = T;` aren't handled).

TODO: take into account provided default type template arguments.


kimgr:
Very nice!

PR #1425, opened 2024-01-06
[ci] Generalize PR comment handling

kimgr:
Turn things around so we always run the IWYU dogfood step, even for scheduled builds. Put the step summary back.

For PRs, bundle dogfood results into pr-comments artifact to be picked up by pr-comments-bot workflow.

Generalize wording and fix a typo in pr-comments-bot.yml.

This makes it possible to add a new informational step by simply producing a Markdown doc in ci.yml, bundle it, and push it as a comment in pr-comments-bot.yml.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,20 +85,26 @@ jobs:
               xargs ./iwyu-check-license-header.py
 
       - name: IWYU dogfood
-        if: github.event_name == 'pull_request'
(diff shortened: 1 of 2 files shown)


kimgr:
Slightly bigger PR than I'd prefer, but the value of splitting up is fairly low as everything needs to be merged at the same time (`workflow_run` only runs from `master`).

Issue #1230, opened 2023-04-07
Segfault when trying to analyze file with precompiled header added via command line


ts826848:
I ran into a segfault when trying to set up precompiled headers in a CMake-based project.

The immediate cause is [this bit in iwyu.cc](https://github.com/kimgr/include-what-you-use/blob/c30cee2ee6eca80ab341bfb5a56af03d9888aac7/iwyu.cc#L4020):

```cpp
  bool VisitTagDecl(clang::TagDecl* decl) {
    // snip
    if (IsForwardDecl(decl)) {
      // snip
      preprocessor_info().FileInfoFor(CurrentFileEntry())->AddForwardDeclare(
          decl_to_fwd_declare, definitely_keep_fwd_decl);
    }
    return Base::VisitTagDecl(decl);
  }
```

`preprocessor_info().FileInfoFor(CurrentFileEntry())` returns `nullptr`, which causes a `this` dereference further down the stack to segfault.

I was able to get the test case reduced to:

`main.cpp`:
```cpp
#include "main.hpp"
```

`main.hpp`:
```cpp
#ifndef MAIN_HPP_INCLUDED
#define MAIN_HPP_INCLUDED
#include "core.hpp"
ADD()
#endif
```

`core.hpp`:
```cpp
#define ADD(...) namespace { struct data; }
```

`driver.sh`, derived from the CMake-generated command line:
```sh
#!/bin/sh

/usr/local/opt/llvm/bin/clang++ \
	-I. \
	-std=c++20 \
	-fpch-instantiate-templates \
	-Xclang -emit-pch \
	-Xclang -include \
	-Xclang main.hpp \
	-x c++-header \
	-o cmake_pch.hxx.pch \
	-c empty.cpp \
	&& \
/usr/local/opt/include-what-you-use/bin/include-what-you-use \
	-std=c++20 \
	-Xclang -include-pch \
	-Xclang cmake_pch.hxx.pch \
	-c main.cpp
```

Trimmed verbose log:
```
Setting verbose-level to 20
<Snip: Adding mappings and search paths>
[ #include    ]  (main.cpp)
Found include: <built-in>:-1 -> main.cpp
Added an #include: <built-in> ->
Added to main compilation unit: main.cpp
[ #include    ]  (<built-in>)
[ Exiting to  ] main.cpp:1:1
[ (#include)  ] "main.hpp" (main.hpp)
Found include: main.cpp:1 -> main.hpp
[...]

PR #1426, opened 2024-01-06
Reject attempts at using PCH

kimgr:
IWYU as-is cannot support PCH, because we care deeply about the include graph, and it does not appear to be preserved when Clang interprets the precompiled header format.

We used to fail mysteriously with segmentation faults due to missing files; now we fail with an explicit message on startup.

Remove null-check for this scenario in iwyu:VisitUsingDecl, that code should now be unreachable.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -123,10 +123,8 @@
 #include "clang/Basic/Specifiers.h"
 #include "clang/Basic/TypeTraits.h"
 #include "clang/Frontend/CompilerInstance.h"
-#include "clang/Frontend/CompilerInvocation.h"
 #include "clang/Frontend/FrontendAction.h"
 #include "clang/Lex/Preprocessor.h"
-#include "clang/Lex/PreprocessorOptions.h"
 #include "clang/Sema/Sema.h"
 #include "iwyu_ast_util.h"
 #include "iwyu_cache.h"
@@ -3869,23 +3867,10 @@ class IwyuAstConsumer
     // though, because that will drag in every overload even if we're
     // only using one.  Instead, we keep track of the using decl and
     // mark it as touched when something actually uses it.
-    IwyuFileInfo* file_info =
-        preprocessor_info().FileInfoFor(CurrentFileEntry());
-    if (file_info) {
-      file_info->AddUsingDecl(decl);
-    } else {
-      // For using declarations in a PCH, the preprocessor won't have any
-      // location information. As far as we know, that's the only time the
-      // file-info will be null, so assert that we have a PCH on the
-      // command-line.
-      const string& pch_include =
-           compiler()->getInvocation().getPreprocessorOpts().ImplicitPCHInclude;
-      CHECK_(!pch_include.empty());
-    }
+    preprocessor_info().FileInfoFor(CurrentFileEntry())->AddUsingDecl(decl);
 
     if (CanIgnoreCurrentASTNode())
       return true;
-
     return Base::VisitUsingDecl(decl);
   }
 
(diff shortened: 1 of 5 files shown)


kimgr:
@kilroyd This is what I had in mind for PCH handling earlier.

@iven @ts826848 A workaround/clarification for #1230 and #1243.

kimgr:
> ### Informational: IWYU dogfood results
> include-what-you-use (exit: 0)
> fix_includes.py (exit: 0)
> diff
> ```diff
> diff --git a/iwyu.cc b/iwyu.cc
> index 1474b78..c32dd33 100644
> --- a/iwyu.cc
> +++ b/iwyu.cc
> @@ -123,7 +123,6 @@
>  #include "clang/Basic/Specifiers.h"
>  #include "clang/Basic/TypeTraits.h"
>  #include "clang/Frontend/CompilerInstance.h"
> -#include "clang/Frontend/CompilerInvocation.h"
>  #include "clang/Frontend/FrontendAction.h"
>  #include "clang/Lex/Preprocessor.h"
>  #include "clang/Sema/Sema.h"
> ```

How cool is this??? :magic_wand:

kilroyd:
IWYU has the option `--pch-in-code`. Does IWYU still support including a PCH through an include directive? Or should that option be removed?

If keeping `--pch-in-code` should we add some documentation about what PCH usage is supported? Also clarify the error message to avoid saying all PCH is not supported.

kimgr:
@kilroyd `--pch_in_code` is a poor way of saying "hey, consider the first `#include` in every .cpp file to be an MSVC-style precompiled header, which can be considered equivalent to a command-line -include".

MSVC used to have this terrible default setup for PCH that the precompiled header was called `"stdafx.h"`, there was a `"stdafx.cpp"`, whose only purpose was to generate the PCH, and then every file explicitly including `"stdafx.h"` would instead get the PCH prepended to the compilation unit.  The nice thing is, if you compile such a .cpp file without special PCH flags, it will just work like a normal include.

(I'm not sure if that's still the case, I haven't used MSVC for many years)

So older MSVC codebases will typically have `#include "stdafx.h"` as the first tokens in every .cpp file, which makes it necessary for IWYU to "see past" it. So `--pch_in_code` essentially makes IWYU treat the first `#include` in every .cpp file as if it was a prefix header (provided on command-line using `-include`), and never removes it. The `--prefix_header_includes` switch can be used to control the policy in the presence of such "God" headers.

I wish there was a better name for that switch, but there it is. The help text tries to explain it tersely.

Anyway, long story short: yes!, some more docs would be nice! :smile:

kimgr:
OK, I started writing up an `IWYUFAQ.md`, but it turned into a bigger thing that I'd rather not publish as part of this PR. I'll follow up with a separate thing explaining why it doesn't work, possible ways to tweak the invocation to make it work, and how `--pch_in_code` relates to all this.

kilroyd:
No problem, sounds good to me.

The commit itself seems fine.

firewave:
> > ### Informational: IWYU dogfood results
> > include-what-you-use (exit: 0)
> > fix_includes.py (exit: 0)
> > diff
> 
> How cool is this??? 🪄

🥳

As the exitcodes are `0` if there is a diff it would be helpful to add an indicator to the title so you do not have to check by hand if there is an actual diff.

kimgr:
@firewave Yeah, maybe that would be nice. I think maybe I've spent enough time on this stuff now, though. Should be a small patch to `iwyu-dogfood.bash` if you're feeling up for it :)

Issue #89, opened 2015-06-15
Test template_args fails on MSVC


vsapsai:
Originally reported on Google Code with ID 89

```
What steps will reproduce the problem?
1. $ python run_iwyu_tests template_args

What is the expected output? What do you see instead?

 Expected: success.

 Actual: AssertionError: tests/template_args.cc:18: Unexpected diagnostic:
  IndirectClass is defined in "tests/indirect.h", which isn't directly
  #included (for autocast).

This is because all classes are erroneously classified as having a conversion constructor.

For some reason an implicit move ctor is generated, which tripped up our conversion
ctor detection. I thought IWYU ran in C++03 mode?

Attached patch do not treat implicit move ctors as conversion ctors. Two more tests
pass for MSVC, I haven't been able to try other platforms.
```

Reported by `kim.grasman` on 2013-01-01 22:29:15

<hr>
- _Attachment: [conversionctor.patch](https://storage.googleapis.com/google-code-attachments/include-what-you-use/issue-89/comment-0/conversionctor.patch)_

PR #1427, opened 2024-01-07
Handle non-autocast cases of implicit construction

bolshakov-a:
Efforts have been made to avoid unwanted reporting of provided types.

Surprisingly, some `decltype` cases became fixed. This is because `CXXConstructExpr` appears also in explicit construction cases, e.g. on variable definitions. It implies that there is currently some redundancy in reporting types of variables: they are reported on `CXXConstructExpr` visiting as well as on written type visiting. It doesn't mean that `decltype` is fully supported now: there remain non-construction cases where its underlying type should be reported but isn't, like using as a base class in some class declaration, or using as a nested name specifier.

Union type suggestions are appeared in the `badinc` test because the constructors take them by value, therefore their complete types are actually needed for copy-initialization.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -178,6 +178,7 @@ using clang::ArraySubscriptExpr;
 using clang::Attr;
 using clang::BinaryOperator;
 using clang::CXXBaseSpecifier;
+using clang::CXXBindTemporaryExpr;
 using clang::CXXCatchStmt;
 using clang::CXXConstructExpr;
 using clang::CXXConstructorDecl;
@@ -2170,7 +2171,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     ReportIfReferenceVararg(expr->getArgs(), expr->getNumArgs(),
                             expr->getConstructor());
 
-    HandleAutocastOnCallSite(expr);
+    if (IsAutocastExpr(current_ast_node()))
+      HandleAutocastOnCallSite(expr);
+    else
+      HandleNonAutocastConstruction(expr);
 
     return true;
   }
@@ -2589,6 +2593,13 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // 'Tpl<Providing>::Type'.
   bool IsProvidedByTplArg(const Type*) = delete;
 
+  // Returns 'true' if either argument or "destination" (currently, function
+  // argument type if it is an argument construction) of 'CXXConstructExpr'
+  // provides the constructed type. In that case, the type should not
+  // be reported.
+  bool SourceOrTargetTypeIsProvided(const ASTNode* construct_expr_node) const =
+      delete;
+
   set<const Type*> GetProvidedTypes(const Type* type,
                                     SourceLocation loc) const {
     set<const Type*> retval;
@@ -2722,11 +2733,20 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // here.  *However*, the function-author can override this
     // iwyu requirement, in which case we're responsible for the
     // casted-to type.  See IwyuBaseASTVisitor::CanBeProvidedTypeComponent.
-    if (!IsAutocastExpr(current_ast_node()))
-      return;
     const Type* type = expr->getType().getTypePtr();
     ReportWithAdditionalBlockedTypes(
         type, GetProvidedTypesForAutocast(current_ast_node()));
+  }
+
+  void HandleNonAutocastConstruction(clang::CXXConstructExpr* expr) {
+    // Avoid reporting of class members from implicit constructor member
+    // initializers: it is redundant and may be attributed to incorrect location
+    // on macro expansion.
+    if (IsCXXConstructExprInInitializer(current_ast_node()))
+      return;
+    if (this->getDerived().SourceOrTargetTypeIsProvided(current_ast_node()))
+      return;
+    ReportTypeUse(CurrentLoc(), expr->getType().getTypePtr(), DerefKind::None);
   }
 
   // Do not add any variables here!  If you do, they will not be shared
@@ -3265,6 +3285,12 @@ class InstantiatedTemplateVisitor
   bool IsProvidedByTplArg(const Type*) {
     // Already inside template instantiation analysis. Types provided
     // by template arguments should be in 'blocked_types_' set.
+    return false;
+  }
+
+  bool SourceOrTargetTypeIsProvided(const ASTNode* construct_expr_node) const {
+    // During instantiated template analysis, provided types are extracted
+    // from template arguments and are in 'blocked_types_' set.
     return false;
   }
 
@@ -4309,6 +4335,35 @@ class IwyuAstConsumer
           return data.provided_types.count(GetCanonicalType(type));
         }
       }
+    }
+    return false;
+  }
+
+  bool SourceOrTargetTypeIsProvided(const ASTNode* construct_expr_node) const {
+    const auto* expr = construct_expr_node->GetAs<CXXConstructExpr>();
+    CHECK_(expr);
+    const Type* type = GetCanonicalType(GetTypeOf(expr));
+    if (expr->getConstructor()->isCopyOrMoveConstructor()) {
+      const Type* source_type =
+          GetTypeOf(expr->getArg(0)->IgnoreParenImpCasts());
+      if (GetProvidedTypeComponents(source_type).count(type))
+        return true;
+    }
+
+    auto get_call_expr = [](const ASTNode* node) -> const CallExpr* {
+      if (const auto* call_expr = node->template GetParentAs<CallExpr>())
+        return call_expr;
+      if (node->template ParentIsA<CXXBindTemporaryExpr>())
+        return node->template GetAncestorAs<CallExpr>(2);
+      return nullptr;
+    };
+
+    if (const CallExpr* call_expr = get_call_expr(construct_expr_node)) {
+      const FunctionDecl* fn_decl = call_expr->getDirectCallee();
+      // For nontemplated functions, 'CXXConstructExpr' type appears to be
+      // sugared and hence handled correctly inside 'ReportTypeUse'.
+      if (GetTplInstData(fn_decl, call_expr).provided_types.count(type))
+        return true;
     }
     return false;
   }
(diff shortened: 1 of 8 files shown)


tests/cxx/implicit_ctor.cc
```diff
@@ -58,6 +62,55 @@ IndirectWithImplicitCtor indirect
...
+template <typename T>
+void TplFn(T t);
+
+// IWYU: NoAutocastCtor needs a declaration
+void TestNonAutocastConstruction(const NoAutocastCtor& par) {
+  // IWYU: NoAutocastCtor is...*implicit_ctor-i2.h
+  FnTakingDirectly(par);
+  FnTakingByProvidingAlias(par);
+  // IWYU: NoAutocastCtor is...*implicit_ctor-i2.h
+  FnTakingByNonProvidingAlias(par);
```
bolshakov-a, line 84: In fact, I'm not sure that cases of function argument initialization should not be considered as "autocast".

kimgr: Can you elaborate? I don't fully understand what you're saying above.

kimgr (commented):
Looks good overall, some questions and a little suggestion.

tests/cxx/badinc.cc
```diff
@@ -704,6 +704,7 @@ I1_TemplateClass<I1_Class> i1_templateclass_object;
 // IWYU: I1_TemplateClass is...*badinc-i1.h
 // IWYU: I1_Class needs a declaration
 // IWYU: I1_Class is...*badinc-i1.h
+// IWYU: I1_Union is...*badinc-i1.h
```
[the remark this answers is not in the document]
bolshakov-a: That's what I've asked about above. Currently, the "author intent" logic works only for those function parameters which have a non-explicit one-arg constructor _except copy- or move-constructors_. I thought that copy-ctors may be just forgotten, but it seems like it is by design, because this logic exists even [in the very first commit](https://github.com/include-what-you-use/include-what-you-use/commit/dee92b5e0acf7188d4e5bb3db6dec833a6d1ff2b#diff-4d5366b53d869b87c2b6cad93f1b0acc2fc8c5d20e28a6fbd058d5a2642b970fR712). Moreover, [you have extended it](https://github.com/include-what-you-use/include-what-you-use/commit/7b58488e95dcc25918ac7f2373ccc8b9c9d39f19) for C++11 move-constructors (issue #89).

> Or is this something that happened to work by accident and now needs explicit work to get the reexports right?

I'd say that it didn't work at all. If you have such files:
```cpp
// A.h

#pragma once

class A {};

// Fn.h

#include "A.h"

void Fn(A);

// UserFn.cpp

#include "Fn.h"
#include "A.h"

void UserFn(const A& a) {
    Fn(a);
}
```
IWYU version from the master branch suggests replacing `#include "A.h"` with fwd-declaration both in `Fn.h` and in `UserFn.cpp`. `A` is not considered for reexport. The code doesn't compile after that, of course.

Another case is C++11 braced initialization syntax, which may be used for implicit constructors without mentioning class name, like in such a function call: `Fn({1, 'a', 5.0});`. Shouldn't all implicit ctors be considered for "autocast"? Should IWYU require their full type info in function-declaring headers by default?

My personal opinion is: no. IWYU should maximize fwd-declarations in headers and `#include`s in `.cpp`-files, because it makes compilation faster and supports refactorings. And when a user writes `Fn({arguments, of, the, constructor})`, he probably knows what he want to construct and can add the corresponding `#include`.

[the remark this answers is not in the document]
bolshakov-a: Btw, this PR doesn't handle list-initialization of [aggregates](https://en.cppreference.com/w/cpp/language/aggregate_initialization), only braced syntax of constructor calls.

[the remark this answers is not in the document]
kimgr: I agree with your sentiment; it does seem strange that implicit construction is different from implicit conversion. I don't have a clear idea what the larger-scale effects would be if we change this policy, but it does seem worth an experiment.

Anyway, that seems unrelated to the changes in this PR, and I think it looks good as a whole.

But I wonder if we could preserve the (broken) policy for copy-initialization in this PR? It's not entirely clear to me what changed it, and it seems orthogonal to the core of the PR.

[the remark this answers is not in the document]
bolshakov-a: > But I wonder if we could preserve the (broken) policy for copy-initialization in this PR?

I'm biased that it is rather fixed than broken here. It is in all senses better to have an `#include` at the caller site than have no one at all.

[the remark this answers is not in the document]
kimgr: Fair enough, that makes sense to me too.

I somehow read the PR as being all about getting provision right, but now I see what you mean, thanks.

> A doubtful case is when the function is inline, or defined inside a class definition (which is essentially the same case).
> In that cases, the passed-by-value argument type should be #included in the function header. However, again, the
> requirement to #include the argument type header at the caller site supports refactoring when someone makes the
> function non-inline and places argument type fwd-declaration in the function header.

Yes, and same for return type. For _this_ PR I think it's fine to require them on both sides, but the ideal behavior should be to reexport them, right? It's a tradeoff between optimizing for stability or for build speed, I guess.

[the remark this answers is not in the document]
bolshakov-a: > the ideal behavior should be to reexport them, right?

I don't think so now. Do you really want an `#include` for every passing-by-value function parameter in the function-defining header (because nearly all classes have copy- and/or move-ctor) instead of the forward-declarations? I think it is contrary to the IWYU goals, and the current "autocast" policy is sufficient.

It should be noted that IWYU cannot have policy to suggest fwd-declaring something on the author/header side and simultaneously to analyze the "provision" status of that thing on the user's side, because it would lead to uncompilable code, as in the example above.

iwyu.cc
```diff
@@ -4313,6 +4338,25 @@ class IwyuAstConsumer
     return false;
   }
 
+  bool SourceOrTargetTypeIsProvided(const ASTNode* construct_expr_node) const {
+    const auto* expr = construct_expr_node->GetAs<CXXConstructExpr>();
```
kimgr, line 4342: Can you add a `CHECK_` here that this return non-null, and so is actually a `CXXConstructExpr`?

iwyu_ast_util.cc
```diff
@@ -433,6 +438,14 @@ const DeclContext* GetDeclContext(const ASTNode* ast_node) {
   return nullptr;
 }
 
+const CallExpr* GetCallExpr(const ASTNode* node) {
```
kimgr, line 441: The function name is a little too generic here, since we can't lean on the type system to know what kind of ASTNode we can pass in. Maybe `GetConstructCallExpr` to narrow it down a little?

tests/cxx/badinc.cc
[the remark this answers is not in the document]
kimgr: Just a few more nits, then let's get this merged.

iwyu_ast_util.cc
bolshakov-a: `GetConstructCallExpr` name may erroneously suggest that it returns a constructor call expression, but it actually returns function call expression which involves its argument construction. I hope that this function may be used for getting `CallExpr` for any direct argument node, not only `CXXConstructExpr`. However, what is "direct" is questionable, in the presence of `CXXBindTemporaryExpr` special handling...

bolshakov-a: Maybe, `GetCallExprOfArgument`?

iwyu.cc
bolshakov-a: Btw, I remember you had the opposite opinion: https://github.com/include-what-you-use/include-what-you-use/pull/1059#discussion_r884479721

iwyu_ast_util.cc
bolshakov-a: But, yeah, `CXXBindTemporaryExpr` handling is present here just because this function handles argument construction. So, `GetCallExprForConstructedArgument` or something shorter?

bolshakov-a: Or, maybe, it should be just a lambda inside `SourceOrTargetTypeIsProvided`?

tests/cxx/badinc.cc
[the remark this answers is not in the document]
kimgr: Oops, our messages crossed.

I think what I'm aiming for is that _if_ the function-declaring header includes the full type, the user should not have to. But my thinking on this is obviously not clear.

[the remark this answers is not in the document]
bolshakov-a: If so, you have a problem like there is on the master branch: IWYU suggests to remove `#include "A.h"` both from the header `Fn.h` and from the `UserFn.h` (according to the policy in that case, because `#include` is still present in `Fn.h` at the moment of analysis), and the code is not compilable after that.

iwyu.cc
kimgr: Yeah, I had a feeling, thanks for looking it up. It's somewhat arbitrary, but here's my reasoning:

* `const EnumDecl* decl = type->getDecl();` should normally never fail; we have strong types and a reasonable assumption about how things work
* in this method, we have zero support from the type system, the only hint is the name of the parameter. Adding a `CHECK_` here is executable documentation that `construct_expr_node` MUST be a `CXXConstructExpr`. It might be nicer with `CHECK_(node->IsA<CXXConstructExpr>());` up-front, but then we have to cast twice.

iwyu_ast_util.cc
kimgr: Yeah, it's probably better to inline it there if it's hard to name it.

`SourceOrTargetTypeIsProvided` is also kind-of dangerously named, but it's a difficult enough name that you're not tempted to call it on any old `ASTNode*` :-)

bolshakov-a:
Thanks!

bolshakov-a:
FWIW, this breaks "autocast" logic for pre-C++17 modes. Clang inserts an additional `CXXConstructExpr` node corresponding to the implicit move constructor call (in the absence of the guaranteed copy elision C++17 feature, I think) under `CallExpr` of a function call. Hence, the type is reported at the call site even if it is provided for "autocast". Personally, I don't think it is very bad regression.

kimgr:
@bolshakov-a Thanks for the heads-up! I think I'll actually open a bug from your comment so we have the behavior recorded.

Issue #1308, opened 2023-09-24
Cannot find system headers on macOS


mohd-akram:
I have a simple C file:

```c
#include <stdio.h>
int main(void)
{
	puts("hello world");
	return 0;
}
```

Running `include-what-you-use main.c` nets:

```c
main.c:1:10: fatal error: 'stdio.h' file not found
#include <stdio.h>
         ^~~~~~~~~
```

I'm running version 0.20 with MacPorts Clang 16 on macOS Ventura. I can confirm that Clang can build the file via `clang-mp-16 main.c`. How does it look for headers?

EDIT: It was working in clang because `clang-mp-16` is a shell script that runs the actual clang via `xcrun`. See my comment below on ways to resolve this for `include-what-you-use`.

kimgr:
I recently added a little section on platform tweaks to the readme: https://github.com/include-what-you-use/include-what-you-use#platform-tweaks.

But after publishing that, I realized that the easiest way to use IWYU on Mac is to use the `-isysroot` switch:
```
include-what-you-use -isysroot $(xcrun --sdk macosx --show-sdk-path) main.c
```
(here using Bash substitutions, not sure if that works with zsh, for example).

Setting the sysroot takes care of both the C headers and the C++ headers.

mohd-akram:
I investigated the problem more deeply. There are two issues:

- Clang needs to find system directories. It does this via the `SDKROOT` environment variable. Running via `xcrun` sets `SDKROOT` (try eg. `xcrun env`), so one can run `include-what-you-use` like so:

   ```shell
   xcrun include-what-you-use main.c
   ```

   Alternatively, add this to the shell profile:

   ```shell
   alias include-what-you-use="xcrun include-what-you-use"
   ```

   or:

   ```shell
   export SDKROOT=$(xcrun --show-sdk-path)
   ```

  or `clang` must be built with the `-DDEFAULT_SYSROOT` option.

- Clang needs to find its own include directories. This is solved by installing `include-what-you-use` next to the clang binary so that it can find the `include` directory in the level above. This can potentially be solved in the tool itself by changing the path passed to the `Driver` constructor to be the same as `clang`'s path. I don't know if the clang library exposes that information. I was also looking into using `dladdr` to get the path of the linked clang library and finding it that way.

kimgr:
Thanks, those are nice techniques. See also issue #100, which covers the challenges with the builtin headers.

[...]

PR #1432, opened 2024-01-09
Migrate platform tweaks README section to FAQ

kimgr:
Touch up the content with some things I learned from issue #1308.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,36 +204,6 @@ If you don't like the way `fix_includes.py` munges your `#include` lines, you ca
 
 Current IWYU pragmas are described in [IWYUPragmas](docs/IWYUPragmas.md).
 
-### Platform tweaks ###
-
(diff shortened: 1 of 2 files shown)


kimgr:
@kilroyd @mohd-akram I don't use macOS as my primary system, so these notes are a little underbaked. If you have nice tips to share, feel welcome to contribute.

mohd-akram:
Looks good. My only comment is that we do handle the `libc++` include paths out of the box in MacPorts, so it's only Homebrew with this issue :)

Also, unrelated and I'm not sure if intentional, but the last bit about IWYU mappings seems unrelated to macOS.

kimgr:
@mohd-akram Thanks -- updated directly on master in bb873de86cccfdb3eb839bb66b38157720803ede

PR #1436, opened 2024-01-12
Handle 'decltype'

bolshakov-a:
I wanted to fix it by refactoring IWYU as discussed in #1340, but the `else` branch inside `InstantiatedTemplateVisitor::AnalyzeTemplateTypeParmUse` hinders. Let's proceed with the current approach.


kimgr:
Looks good, thanks!

PR #1439, opened 2024-01-19
Handle variables marked as 'extern'

bolshakov-a:
They are forward-declarable when not initialized.

An inline variable added to `quoted_includes_first.h` so as to keep the quoted `#include` in the IWYU suggestion for the header.

Hoping that C89 is eventually supported...


kimgr:
Nice, thanks!

PR #1440, opened 2024-01-20
Improve handling nested typedefs

bolshakov-a:
This fixes two problems:

- The aliased type was analyzed as a whole despite it may be a template specialization type, and the outer template specialization type may provide some of its arguments but not the aliased template spec type as a whole.

- In the case of a nested name specifier chain, e. g.
`Tpl1<Alias1>::Class::Tpl2<Alias2>::Struct::Tpl3<Alias3>::Typedef`,
only the rightmost prefix was analyzed. Now, provided types from all the chain are harvested.


kimgr:
Looks good, thanks!

PR #1441, opened 2024-01-21
Improve alias template support

### LINKED ISSUE #1249: iwyu removes some headers incorrectly
I have these three files:

FontMgr.h:
```c++
class IFontMgr {
 public:
  void getFont();
};
```

main.h:
```c++
#include <memory>

class IFontMgr;

class FontRenderer {
  void render() const;

 private:
  std::shared_ptr<IFontMgr> fontMgr_;
};
```

main.cpp:
```c++
#include "main.h"
#include "FontMgr.h"

void FontRenderer::render() const {
  fontMgr_->getFont();
}
```

Running iwyu(main branch & clang_16 branch):
```bash
$ include-what-you-use -stdlib=libc++ main.cpp
```

I get:
```
main.h should add these lines:
#include <__memory/shared_ptr.h>  // for shared_ptr

main.h should remove these lines:
- #include <memory>  // lines 1-1

The full include-list for main.h:
#include <__memory/shared_ptr.h>  // for shared_ptr
class IFontMgr;  // lines 3-3
---

- #include "FontMgr.h"  // lines 2-2

#include "main.h"
---
```

`FontMgr.h` is removed by iwyu, and `main.cpp` cannot be compiled by clang, I get errors like:
```
error: member access into incomplete type 'element_type' (aka 'IFontMgr')
```

This happens not only when calling `shared_ptr<IFontMgr>->method()`, but also `optional<IFontMgr>->method()`.

bolshakov-a:
This fixes #1249.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3090,6 +3090,9 @@ class InstantiatedTemplateVisitor
     if (CanIgnoreCurrentASTNode())
       return true;
 
+    if (type->isTypeAlias())
+      return TraverseType(type->getAliasedType());
+
     // Skip the template traversal if this occurrence of the template name is
     // just a class qualifier for an out of line method, as opposed to an object
     // instantiation, where the templated code would need to be inspected.
@@ -4219,7 +4222,7 @@ class IwyuAstConsumer
 
     // If we're not in a forward-declare context, use of a template
     // specialization requires having the full type information.
-    if (!CanForwardDeclareType(current_ast_node())) {
+    if (!CanForwardDeclareType(current_ast_node()) || type->isTypeAlias()) {
       const TemplateInstantiationData data = GetTplInstData(type);
       instantiated_template_visitor_.ScanInstantiatedType(
           current_ast_node(), data.resugar_map, data.provided_types);
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1447,8 +1447,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
-  set<const Type*> GetProvidedTypesForTypedef(
-      const TypedefNameDecl* decl) const {
+  set<const Type*> GetProvidedTypesForAlias(const TypedefNameDecl* decl) const {
     const Type* underlying_type = decl->getUnderlyingType().getTypePtr();
     // If the underlying type is itself a typedef, we recurse.
     if (const auto* underlying_typedef =
@@ -1457,7 +1456,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         // TODO(csilvers): if one of the intermediate typedefs
         // #includes the necessary definition of the 'final'
         // underlying type, do we want to return it here?
-        return GetProvidedTypesForTypedef(underlying_typedef_decl);
+        return GetProvidedTypesForAlias(underlying_typedef_decl);
       }
     }
     // TODO(bolshakov): handle underlying alias templates.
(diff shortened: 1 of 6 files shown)


tests/cxx/typedef_in_template.cc
```diff
@@ -122,6 +122,12 @@ struct Identity {
   struct Inner {
     using Type = T;
   };
+
+  static T t;
+  using SugaredType = decltype(t);
+
+  template <int>
+  using AliasTemplate = T;
```
bolshakov-a, line 130: I'm not sure that the test named "Typedef in template" is the best place for alias template testing, but very convenient.

tests/cxx/typedef_in_template.cc
```diff
@@ -167,6 +173,19 @@ void ArgumentTypeProvision() {
   Outer<Providing>::Inner<IndirectClass*>::AliasedTpl p_ptr;
   // IWYU: IndirectClass needs a declaration
   Outer<IndirectClass*>::Inner<Providing>::AliasedTpl ptr_p;
+
+  // IWYU: IndirectClass needs a declaration
+  // IWYU: IndirectClass is...*indirect.h
+  (void)sizeof(Identity<IndirectClass>::SugaredType);
```
bolshakov-a, line 179: This line should probably be moved out of this function, because the function tests slightly different kind of the argument type provision.

kimgr: Yeah, that makes sense to me too.

tests/cxx/typedef_in_template.cc
kimgr: I think they're sufficiently related that it makes sense.

iwyu.cc
```diff
@@ -1447,21 +1447,23 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
     // If the underlying type is itself a typedef, we recurse.
     if (const auto* underlying_typedef =
             underlying_type->getAs<TypedefType>()) {
       if (const auto* underlying_typedef_decl = dyn_cast<TypedefNameDecl>(TypeToDeclAsWritten(underlying_typedef))) {
         // TODO(csilvers): if one of the intermediate typedefs
         // #includes the necessary definition of the 'final'
         // underlying type, do we want to return it here?
-        return GetProvidedTypesForAlias(underlying_typedef_decl);
+        return GetProvidedTypesForAlias(
+            underlying_typedef_decl->getUnderlyingType().getTypePtr(),
```
kimgr, line 1460: Wait, are we now going from type to decl and then back to type? Is there no way to get from `underlying_typedef` to the underlying type directly? OTOH, it seems like we need the decl location.

kimgr (commented):
Generally looks good! Some feedback inline.

iwyu.cc
bolshakov-a: `underlying_typedef->desugar().getTypePtr()` would probably work. However, the approach under discussion was already present on the master branch [in the calling code](https://github.com/include-what-you-use/include-what-you-use/blob/0.21/iwyu.cc#L2607) (inside `ReportTypeUseInternal`). I think, if it should be changed here, it should also be changed there. But it is not obvious to me why one approach is better than the other, if the `TypedefNameDecl` is still needed for other purposes.

kimgr: It looked very roundabout, but as you say, it makes sense if the decl is needed anyway (and I suppose it is, both to determine if we should report, and for the location).

bolshakov-a: To be more correct, its location is taken to determine if a particular type component should be reported.

iwyu.cc
```diff
@@ -2676,19 +2672,33 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
                  << typedef_decl->getQualifiedNameAsString()
                  << " owns the underlying type:\n";
         // If any of the used types are themselves typedefs, this will
         // result in a recursive expansion.
-        const Type* type = typedef_decl->getUnderlyingType().getTypePtr();
         ReportTypeUseInternal(used_loc, type, blocked_types, deref_kind);
       }
       return;
     }
+    if (const auto* template_spec_type =
```
kimgr, line 2688: I see the typedef case on line 2671/2667 is using `getAs<>` and we're using `dyn_cast<>` here.

I can't fully understand which is the best here (`getAs<>` desugars and `dyn_cast<>` does not, as I understand it)... But maybe they should use the same mechanism? Or is it important that they don't?

bolshakov-a: If it is `TemplateSpecializationType`, it should be exactly `TemplateSpecializationType` after desugaring at the beginning of `ReportTypeUseInternal`. But, yeah, let the code be similar.

bolshakov-a:
Thank you!

PR #1442, opened 2024-01-21
Add and use IsNamedFile and IsInNamedFile

kimgr:
The IsBuiltinFile and IsBuiltinOrCommandLineFile predicates were used in a few places to determine if includers/includees were actual files.

This would fail to match some non-files like "<stdin>", but was also not very ergonomic, as callers would have to resolve file entry in order to check for filey-ness.

Add IsNamedFile as the logical opposite of IsBuiltinOrCommandLineFile, but that covers even more special files.

Add a templated IsInNamedFile, which maps any supported type (tokens, decls, locations, etc) to a file, and then checks if said file is named.

I don't see this solving any concrete bugs, but it seems like a more consistent model, and the building blocks are more reusable.

### DIFF
--- a/iwyu_location_util.h
+++ b/iwyu_location_util.h
@@ -75,23 +75,6 @@ inline const clang::FileEntry* RawFileEntry(clang::OptionalFileEntryRef file) {
   return &file->getFileEntry();
 }
 
-// Some symbols are directly defined by the compiler.  For them, the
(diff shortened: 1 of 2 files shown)


kimgr:
I'm glad I didn't merge this. I've had some time to think, and I don't think this is the right move. I'll see if I can come up with a better design.

PR #1444, opened 2024-01-26
Add a test on 'IgnoreKind::ForExpansion'

bolshakov-a:
Otherwise, all the tests are passed when removing it and the `else` branch inside `AnalyzeTemplateTypeParmUse` function.


kimgr:
If you rebase on latest master, the build should be able to get past the configuration error.

PR #1448, opened 2024-01-27
Fix crash on 'type_info' in MS compatibility mode

bolshakov-a:
When running in MS-compatibility mode, one of the `std::type_info` redeclarations is implicitly added by clang [here](https://github.com/llvm/llvm-project/blob/llvmorg-17.0.6/clang/lib/Sema/Sema.cpp#L333). It has no valid source location information.


kimgr:
Nice, thanks!

Issue #1271, opened 2023-06-27
Figure out how to reconcile with libc++ libcxx.imp


kimgr:
It turns out libc++ ships with a mapping file for itself: https://github.com/llvm/llvm-project/blob/main/libcxx/include/libcxx.imp.

It makes no sense for us to maintain a separate version of this, so it would be nice to rely on this canonical mapping (closer to the source).

A few open questions:

* Can/should we use the libc++ mappings automatically if they happens to exist?
* Should we drop our version?
* Are they equivalent? If not, which is better?
* Should we look for mappings on the include path?

Inspired by this comment from one of the libc++ maintainers: https://github.com/include-what-you-use/include-what-you-use/commit/5540ff09d853605abd328121381a9b720e4269cd#commitcomment-119816408.

kimgr:
@kilroyd @ldionne So I've let this bake for a while, and I'd like to add a proposal:

* Have IWYU search the include path for `*.imp` mapping files and load them implicitly
* We need to soften at least one assertion to make that work: https://github.com/include-what-you-use/include-what-you-use/blob/26b48e7/iwyu_include_picker.cc#L1272 -- otherwise there's a good chance picking up mappings automatically will lead to crashes the user can't get themselves out of
* On that note, I'd like to add a `-nostdinc`-like switch, e.g. `-Xiwyu -no_mapping_search`, to disable this behavior

Once that's in place we should be able to get rid of both the built-in mappings for libcxx and `libcxx.imp` in the IWYU tree.

I do think it might be prudent to do have IWYU provide mappings for libcxx for a while, since upstream `libcxx.imp` was recently released. I have it on both my Linux and Mac systems, but I've probably installed recent LLVM distros on both of them. @ldionne Do you know if libcxx.imp will make it into the XCode SDK (which I assume is what bundles libc++ on the Mac)?

There's [a lot more use of `.imp` files than I expected](https://github.com/search?q=path%3A*.imp&type=code), so there's a good chance that picking them up automatically from the include path will encourage libraries to ship them which should in turn make IWYU's life much easier.

I'm going to start hacking out small changes to replace assertions with warning logs in the coming days, and hopefully have somethink functional for implicit search pretty soon.

Thoughts? Thank you!

kilroyd:
Conceptually I think this is reasonable.

[...]

PR #1449, opened 2024-01-28
Add mapping for static_assert(3)

alejandro-colomar:
It is a keyword in C++, but in C it is a macro from <assert.h>.

### DIFF
--- a/gcc.symbols.imp
+++ b/gcc.symbols.imp
@@ -105,6 +105,7 @@
   { symbol: [ "size_t", private, "<time.h>", public ] },
   { symbol: [ "size_t", private, "<uchar.h>", public ] },
   { symbol: [ "size_t", private, "<wchar.h>", public ] },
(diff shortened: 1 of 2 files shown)


kimgr:
This might be a net loss for C++ users -- won't `static_assert` in C++ now require `<assert.h>`?

alejandro-colomar:
Yeah, I was worried about that.  Is there any way to avoid it?

kimgr:
Not that I can think of off-hand.

I've wanted to make mappings a bit more flexible for a while (particularly to make progress with #1271), but I haven't got around to it. Maybe by-language mappings is something that would also be interesting to explore.

I'm not sure we can use the current translation unit's language options (`compiler()->getLangOpts()`) at startup to decide, because I think there may be cases where a C++ program pulls in a header with `extern "C"` where we might want to apply C mappings.

alejandro-colomar:
Hmm, I'll silence it with a pragma in my code for now.  Thanks!

PR #1456, opened 2024-02-04
Improve recursive handling of type aliases

bolshakov-a:
It turns out that the presence of `SubstTemplateTypeParmType` inside an alias type sugar chain doesn't necessarily mean that it is a user-provided template parameter. It may occur e. g. when an alias template is used inside the `typedef` declaration, and its template parameter is specified inside that declaration. The regression had been introduced in 2d26dd1b33 and found on MS STL `vector` implementation when IWYU started to require `_Simple_types` type info for such a code:
```cpp
std::vector v{1, 2, 3};
(void)v.begin();
```
due to the returned `iterator` type alias.

Besides that, this PR brings several other improvements:

- avoiding erroneous consideration of aliased component types as provided just because their forward declaration is in the other (internal alias declaring) file;

- correct handling of aliased types in contexts other than direct aliased type usage (e. g. when handling function return type);

- recursive handling of nested alias templates similar to `typedef`s;

- avoiding of losing nested `typedef` internals during recursive collecting their provided types, like `Class` inside such an elaborated alias: `using AliasedClass = Identity<Class>::Alias;`.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1450,20 +1450,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
   set<const Type*> GetProvidedTypesForAlias(const Type* underlying_type,
                                             SourceLocation decl_loc) const {
-    // If the underlying type is itself a typedef, we recurse.
-    if (const auto* underlying_typedef =
-            underlying_type->getAs<TypedefType>()) {
-      if (const auto* underlying_typedef_decl = dyn_cast<TypedefNameDecl>(TypeToDeclAsWritten(underlying_typedef))) {
-        // TODO(csilvers): if one of the intermediate typedefs
-        // #includes the necessary definition of the 'final'
-        // underlying type, do we want to return it here?
-        return GetProvidedTypesForAlias(
-            underlying_typedef_decl->getUnderlyingType().getTypePtr(),
-            GetLocation(underlying_typedef_decl));
-      }
-    }
-    // TODO(bolshakov): handle underlying alias templates.
-
     return GetProvidedTypes(underlying_type, decl_loc);
   }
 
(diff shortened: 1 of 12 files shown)


iwyu.cc
```diff
@@ -1450,20 +1450,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
   set<const Type*> GetProvidedTypesForAlias(const Type* underlying_type,
```
kimgr, line 1451: I wonder if it would make sense to remove this now that it's just an empty wrapper for `GetProvidedTypes`. Or would it make the patch more confusing?

tests/cxx/iwyu_stricter_than_cpp-i4.h
```diff
@@ -7,10 +7,15 @@
 //
 //===----------------------------------------------------------------------===//
 
+#ifndef IWYU_STRICTER_THAN_CPP_I4_H_
```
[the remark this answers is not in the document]
kimgr: (I'll follow up with a refactor later to make them all follow convention)

iwyu.cc
bolshakov-a: I'm not sure. I thought that leaving them keeps the code more consistent with the other cases (`GetProvidedTypesForFnReturn`, `GetProvidedTypesForAutocast`), and a good compiler should throw away redundant indirection...

kimgr: `FnReturn` and `Autocast` are all about the context, whereas `Alias` is just a quality of the type itself. I find it confusing now that `GetProvidedTypes` recurses via `GetProvidedTypesForAlias`.

But if there's a chance that we may want to tweak processing for alias types in the future, then I guess it makes sense to leave it.

bolshakov-a: Done as a separate commit.

kimgr:
Thanks, looks great!

PR #1469, opened 2024-02-26
[ci] Build with static libraries

kimgr:
There does not appear to be a valid libclang-cpp.so shipped with the llvm-19 apt packages. Explicitly ask IWYU to link to the static libraries.

This will be slower to link, but hopefully more stable and faster to run tests.


kimgr:
No, this didn't pan out. See #1470 for the eventual fix.

PR #1487, opened 2024-03-15
Isolate platform specific headers to a translation unit

### LINKED ISSUE #1485: build error on master under Windows
```
D:\legalize\llvm\llvm-project\llvm\include\llvm\Support\DXILABI.h(27,3): error C2144: syntax error: 'void' should be preceded by '}' [D:\legalize\llvm\build-iwyu\tools\iwyu\include-what-you-use.vcxproj]
  (compiling source file '../../../include-what-you-use/iwyu_ast_util.cc')
```

Line 27 of `DXILABI.h` contains the following:
```
enum class ParameterKind : uint8_t {
  INVALID = 0,
  VOID,  // line 27
  HALF,
  FLOAT,
  DOUBLE,
  I1,
  I8,
  I16,
  I32,
  I64,
  OVERLOAD,
  CBUFFER_RET,
  RESOURCE_RET,
  DXIL_HANDLE,
};
```

A Windows header files define `VOID` as a macro that expands to `void`.  The offending header is the inclusion of `Shlwapi.h` on line 78 of `iwyu_port.h`.

Hacking in a simple `#undef VOID` after the inlusion of that header results in a clean build.

Branch `clang_18` of include-what-you-use with branch `release/18.x` of llvm/clang builds without errors.

[iwyu-build-log.txt](https://github.com/include-what-you-use/include-what-you-use/files/14608595/iwyu-build-log.txt)

LegalizeAdulthood:
Including Windows related headers in iwyu_port.h pollutes other translation units with Windows-specific definitions.  Move the function GlobMatchesPath to have external linkage instead of being inline and move the platform-specific code to a new file iwyu_port.cc

Fixes #1485

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,7 @@ add_llvm_executable(include-what-you-use
   iwyu_location_util.cc
   iwyu_output.cc
   iwyu_path_util.cc
+  iwyu_port.cc
   iwyu_preprocessor.cc
   iwyu_regex.cc
   iwyu_verrs.cc
--- a/iwyu_port.cc
+++ b/iwyu_port.cc
@@ -1,0 +1,26 @@
+//===--- iwyu_port.cc - portability shims ---------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#if defined(_WIN32)
+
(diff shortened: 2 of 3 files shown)


kimgr:
Bah, sorry. The header line pattern requires three leading dashes; `//###--- `, but you only had room for two. Could you trim the description down to just `portability shims`? Or let me know if you don't want to spend time on this, and I can amend and merge it.

Looks great otherwise, it hadn't occurred to me to isolate that thing. Should help build times too.

LegalizeAdulthood:
Fixed.

kimgr:
Thank you!

PR #1488, opened 2024-03-17
Add bits/memory_resource.h (memory_resource) and bits/uses_allocators.h (memory) - close #1486

### LINKED ISSUE #1486: bits/memory_resource.h is missing in gcc.stl.headers.imp
I think `gcc.stl.headers.imp` should include the following:
```cpp
{ include: ["<bits/memory_resource.h>", private, "<memory_resource>", public ] },
```
This is what we get otherwise from IWYU:

> <filename> should add these lines:
> #include <bits/memory_resource.h>              // for polymorphic_allocator

See [here](https://github.com/skypjack/entt/actions/runs/8295051177/job/22701286876#step:5:447) for more details. This is a project of mine where I use IWYU. It has a few false positives, including `bits/memory_resource.h`.
I can also prepare a PR if needed. I would just like confirmation before proceeding because I'm not 100% sure. 🙂

skypjack:
As mentioned in #1486 this suppresses a warning due to ie uses of `std::polymorphic_allocator` among the others.
I've also added what's needed for `bits/uses_allocator.h` (see [here](https://github.com/skypjack/entt/actions/runs/8314254276/job/22751212347#step:5:677) for false positive reported by IWYU).

### DIFF
--- a/gcc.stl.headers.imp
+++ b/gcc.stl.headers.imp
@@ -129,6 +129,7 @@
   { include: ["<bits/unique_ptr.h>", private, "<memory>", public ] },
   { include: ["<bits/unordered_map.h>", private, "<unordered_map>", public ] },
   { include: ["<bits/unordered_set.h>", private, "<unordered_set>", public ] },
+  { include: ["<bits/uses_allocator.h>", private, "<memory>", public ] },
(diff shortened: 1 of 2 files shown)


kimgr:
Thanks for taking care of that!

Issue #1490, opened 2024-03-18
[0.22] `iwyu_include_picker.cc:1359`: Assertion failed: (*map)[key] == visibility Same file seen with two different visibilities: <debug/safe_iterator.h> Old vis: 2 New vis: 1


Jacobfaib:
To reproduce:
```
$ include-what-you-use -Xiwyu --mapping_file=/path/to/include-what-you-use/0.22/libexec/share/include-what-you-use/gcc.stl.headers.imp
```

If I had to guess, these lines are out of sync:
- https://github.com/include-what-you-use/include-what-you-use/blob/master/gcc.stl.headers.imp#L311 (`{ include: ["<debug/safe_iterator.tcc>", private, "<debug/safe_iterator.h>", public ] },`)
- https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_include_picker.cc#L806 (`{ "<debug/safe_iterator.tcc>", kPrivate, "<debug/safe_iterator.h>", kPrivate },`)

kimgr:
Oops! I think it's actually https://github.com/include-what-you-use/include-what-you-use/blob/78472a725b1dd414e0c6ccdef464a8058593d6eb/gcc.stl.headers.imp#L165 vs. https://github.com/include-what-you-use/include-what-you-use/blob/78472a725b1dd414e0c6ccdef464a8058593d6eb/gcc.stl.headers.imp#L313

I had no idea that safe_iterator.tcc mapping existed since 2012.

Just FYI, many of the `.imp` files shipped with IWYU are also hard-coded inside IWYU, so you typically can't pass these mapping files like this without conflicts. They're intended for use together with `-Xiwyu --no_default_mappings`, which suppresses the hard-coded ones (and lets you pick and choose among the external mappings).

That said, it's reproducible with 
```
include-what-you-use -Xiwyu --no_default_mappings -Xiwyu --mapping_file=iwyu.gcc.imp t.cc
```
too, and works when the extra mapping is removed.

Jacobfaib:
> Just FYI, many of the .imp files shipped with IWYU are also hard-coded inside IWYU, so you typically can't pass these mapping files like this without conflicts. 

Ah, I did not know this! I found them when I had to start rolling my own mapping file for Apple clang's STL (I'd be happy to share that too if it helps).

BTW, by "many" do you mean all of them? Or a subset? And if a subset, is it clearly defined or just whatever you happened to include?

kimgr:
[...]

PR #1491, opened 2024-03-18
Remove redundant debug/safe_iterator.tcc mapping

kimgr:
I added complete mappings for debug/ GCC headers in abc3aeefa0af810e4ec00a2c14bae538568dd8c8, but I failed to notice that debug/safe_local_iterator.tcc was already mapped.

That leads to an assertion failure for conflicting visibilities for:

    include-what-you-use \
        -Xiwyu --no_default_mappings \
        -Xiwyu --mapping_file=gcc.stl.headers.imp \
        ...

Remove the old mapping, it's superseded by the new.


kimgr:
Fixes #1490

PR #1492, opened 2024-03-21
Fix handling function references

bolshakov-a:
Additional filtering of function references has been removed. This fixes reporting of template argument type when the template has no definition in the translation unit but the full type info is needed to build the instantiated function prototype, as well as handling instantiations of regular member functions of class templates. It was a regression introduced in #1300.

To make it possible to do, function reference handling has been decoupled from return type handling. Function references differ from calls in that the return type should not be reported in the former case. It was achieved by means of (not fully) complemetary `if`-conditions inside `TraverseDeclRefExpr` and `IwyuBaseAstVisitor::HandleFunctionCall` functions stating that fn refs should be analyzed only for instantiated template functions with definitions, meanwhile return types should be reported in the opposite case. But it is not fully correct. In fact, the full return type info is required on any function call. It is just convenient to report it on instantiated template visitation because the type may be provided by some `typedef`-ed template argument. But it stops working e.g. when the function template has no definition in the translation unit, just a declaration. This change enables separate improvement of function reference and function call handling.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -842,7 +842,8 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
       const Type* parent_type = nullptr;
       if (expr->getQualifier() && expr->getQualifier()->getAsType())
         parent_type = expr->getQualifier()->getAsType();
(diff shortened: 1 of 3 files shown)


kimgr:
Thanks!

Issue #1250, opened 2023-04-25
Incorrect include with function template


carlosgalvezp:
Hi,

Consider the following case:

```cpp
// bar.h
#pragma once
#include <iterator>
```

```cpp
// foo.h
#pragma once

#include "bar.h"

template <typename T>
T get_begin(T const(&x)[10])
{
  return *std::begin(x);
}
```

```cpp
// foo.cpp
#include "foo.h"

int main()
{
  int x[10]{};
  return get_begin(x);
}
```

If I run IWYU, on `foo.cpp`, I get:

```
$ include-what-you-use foo.cpp 

foo.h should add these lines:

foo.h should remove these lines:
- #include "bar.h"  // lines 3-3

The full include-list for foo.h:
---

foo.cpp should add these lines:
#include <iterator>  // for begin

foo.cpp should remove these lines:

The full include-list for foo.cpp:
#include "foo.h"
#include <iterator>  // for begin
---
```

This is incorrect - `foo.h` should be the one including `iterator`. Currently, if N clients include `foo.h`, they are asked to include `iterator` just because they are calling a function that internally uses something from it. Implementation details are leaked to clients - if `foo.h` no longer needs `iterator`, suddenly all clients need to remove it. This creates a lot of code churn.

This behavior does not happen if the function `get_begin` is a regular (non-template) function.

bolshakov-a:
@carlosgalvezp, does the buggy behaviour reproduce if you place `#include <iterator>` directly in `foo.h`?

carlosgalvezp:
@bolshakov-a If I do that and remove `#include bar.h` it actually asks me to remove it:

```
foo.h should add these lines:

foo.h should remove these lines:
- #include <iterator>  // lines 3-3

The full include-list for foo.h:
---

(foo.cpp has correct #includes/fwd-decls)
```

Which is even worse since the code no longer compiles. I now remember the sequence of events:

- I had code with the correct includes on IWYU 0.18.
- I bump to IWYU 0.20, which asked to remove `iterator` from `foo.h`.
- When `iterator` was removed from `foo.h`, IWYU started asking to include it in all the clients of `foo.h`, which led me to this opening this ticket.

bolshakov-a:
[...]

PR #1503, opened 2024-03-30
Fix provision analysis with public headers

bolshakov-a:
This fixes determination of types provided with type aliases and function declarations (fn return, "autocast") in the presence of mapping files. The problem was in that the `PublicHeaderIntendsToProvide` function considers as provided all the headers transitively included from a public header, despite they are may be themselves public and not have the explicit public-to-public mapping specified. It is convenient to avoid noisy suggestions of the library internals when handling instantiations of some external library templates, but may hinder expressing author intent when writing e.g. some own (user's) `typedef` (non-)providing a type from such a library.

To achieve the more correct behavior, the logic from `OneUse::SetPublicHeaders()` function has been reused which takes into account only explicitly specified mappings. The drawback is that the use site vs. the decl site difference for default template arguments becomes worse (see `TestDefaultTplArgs()`). The use site handling still uses `PublicHeaderIntendsToProvide` (which is common way for template instantiation handling, as stated before). It may end up in that IWYU may suggest to remove an indirectly providing header from the template defining file (`-d1.h` from `-def_tpl_arg.h` in the test case) not suggesting to `#include` the default argument header at the use site because at the time of analysis `-d1.h` is present in `-def_tpl_arg.h` and "provides" all its directly and indirectly included stuff. OTOH, it may be fixed easily by means of including the appropriate header in the template instantiating file manually, and the IWYU behavior should become consistent.

The main goal of these changes is to introduce `IsDirectlyIncluded` function which may be reused when fixing `UnresolvedLookupExpr`-related bugs (`PublicHeaderIntendsToProvide` doesn't work well in real-world cases for it).

@kimgr, wdyt? Is it worth merging? Or it needs more work to make the default template argument handling consistent? Or it's better to keep the old `typedef`/"autocast"/fnreturn behavior and introduce `IsDirectlyIncluded` just for `VisitOverloadExpr` in a separate PR? To understand the problem, you may take `iwyu_stricter_than_cpp.imp` mapping file and observe how drastically it affects the analysis on the current master.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1436,17 +1436,18 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       return false;
 
     // Check if the the author is not #including the file with the
-    // definition.  PublicHeaderIntendsToProvide has exactly the
-    // semantics we want.  Note if there's no definition anywhere, we
+    // definition. Note if there's no definition anywhere, we
     // say the author does not want the full type (which is a good
     // thing, since there isn't one!)
     if (const NamedDecl* dfn = GetTagDefinition(decl)) {
       if (IsBeforeInSameFile(dfn, use_loc))
         return false;
-      if (preprocessor_info().PublicHeaderIntendsToProvide(
-              GetFileEntry(use_loc), GetFileEntry(dfn))) {
+      const IwyuFileInfo* use_file =
+          preprocessor_info().FileInfoFor(GetFileEntry(use_loc));
+      if (!use_file)  // Not sure whether this is possible.
+        return true;
+      if (IsDirectlyIncluded(dfn, *use_file))
         return false;
-      }
     }
 
     // OK, looks like the author has stated they don't want the fulll type.
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -35,6 +35,7 @@
 #include "iwyu_stl_util.h"
 #include "iwyu_string_util.h"
 #include "iwyu_verrs.h"
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/Casting.h"
 
@@ -63,6 +64,7 @@ using clang::UsingDecl;
 using llvm::cast;
 using llvm::dyn_cast;
 using llvm::errs;
+using llvm::find_if;
 using llvm::isa;
 using llvm::raw_string_ostream;
 using std::map;
@@ -335,14 +337,8 @@ void OneUse::SetPublicHeaders() {
   // We should never need to deal with public headers if we already know
   // who we map to.
   CHECK_(suggested_header_.empty() && "Should not need a public header here");
-  const IncludePicker& picker = GlobalIncludePicker();   // short alias
-  const string use_path = GetFilePath(use_loc_);
-  // If the symbol has a special mapping, use it, otherwise map its file.
-  public_headers_ = picker.GetCandidateHeadersForSymbolUsedFrom(
-      symbol_name_, use_path);
-  if (public_headers_.empty())
-    public_headers_ = picker.GetCandidateHeadersForFilepathIncludedFrom(
-        decl_filepath(), use_path);
+  public_headers_ = GlobalIncludePicker().GetMappedPublicHeaders(
+      symbol_name_, GetFilePath(use_loc_), decl_filepath());
   if (public_headers_.empty())
     public_headers_.push_back(ConvertToQuotedInclude(decl_filepath()));
 }
@@ -2296,6 +2292,24 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
   errs() << diff_output;
 
   return num_edits;
+}
+
+bool IsDirectlyIncluded(const NamedDecl* decl, const IwyuFileInfo& includer) {
+  // If the decl or its header is private then a user should '#include' any of
+  // the specified public headers for it, so check their presence.
+  const vector<string> mapped_headers =
+      GlobalIncludePicker().GetMappedPublicHeaders(
+          GetWrittenQualifiedNameAsString(decl),
+          GetFilePath(includer.file_entry()), GetFilePath(decl));
+  if (!mapped_headers.empty()) {
+    auto pred = [&direct_includes =
+                     includer.direct_includes()](const string& header) {
+      return direct_includes.count(header);
+    };
+    return find_if(mapped_headers, pred) != mapped_headers.end();
+  }
+  // Otherwise, check the decl header directly.
+  return includer.direct_includes_as_fileentries().count(GetFileEntry(decl));
 }
 
 }  // namespace include_what_you_use
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -1760,6 +1760,18 @@ void IncludePicker::AddMappingsFromFile(const string& filename) {
   return AddMappingsFromFile(filename, default_search_path);
 }
 
+vector<string> IncludePicker::GetMappedPublicHeaders(
+    const string& symbol_name,
+    const string& use_path,
+    const string& decl_filepath) const {
+  // If the symbol has a special mapping, use it, otherwise map its file.
+  vector<string> symbol_headers =
+      GetCandidateHeadersForSymbolUsedFrom(symbol_name, use_path);
+  if (!symbol_headers.empty())
+    return symbol_headers;
+  return GetCandidateHeadersForFilepathIncludedFrom(decl_filepath, use_path);
+}
+
 // Parses a YAML/JSON file containing mapping directives of various types:
 //  symbol   - symbol name -> quoted include
 //  include  - private quoted include -> public quoted include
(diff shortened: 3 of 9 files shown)


iwyu.cc
```diff
@@ -1436,17 +1436,18 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
     // say the author does not want the full type (which is a good
     // thing, since there isn't one!)
     if (const NamedDecl* dfn = GetTagDefinition(decl)) {
       if (IsBeforeInSameFile(dfn, use_loc))
         return false;
-      if (preprocessor_info().PublicHeaderIntendsToProvide(
-              GetFileEntry(use_loc), GetFileEntry(dfn))) {
+      const IwyuFileInfo* use_file =
+          preprocessor_info().FileInfoFor(GetFileEntry(use_loc));
+      if (!use_file)  // Not sure whether this is possible.
```
kimgr, line 1447: It's tempting to leave this out. I think it can only happen in two cases:

* built-in types (should be covered by https://github.com/include-what-you-use/include-what-you-use/pull/1503/files#diff-7ca107c8b836c1765e8b735b74f06833bec48f308a76017886cee1d5736002d9R1412)
* precompiled headers (should be impossible as of https://github.com/include-what-you-use/include-what-you-use/commit/4f4bc53060670f095bd8192e32ffc4e0ed9272d5)

but it seems better to exit silently than crash since we don't run IWYU on a large corpus of real C++ code yet.

Should it really be `return true;`, though? Assuming it's a builtin-like thing, `false` seems potentially better.

iwyu_output.cc
```diff
@@ -2298,4 +2306,19 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
...
+bool IsDirectlyIncluded(const NamedDecl* decl, const IwyuFileInfo& includer) {
+  const vector<string> mapped_headers = internal::GetMappedPublicHeaders(
+      internal::GetQualifiedNameAsString(decl),
+      GetFilePath(includer.file_entry()), GetFilePath(decl));
+  if (!mapped_headers.empty()) {
+    auto pred = [&direct_includes =
+                     includer.direct_includes()](const string& header) {
+      return direct_includes.count(header);
+    };
+    return find_if(mapped_headers.begin(), mapped_headers.end(), pred) !=
```
kimgr, line 2318: You could consider `llvm::find_if` from `llvm/STLExtras.h` instead; it takes the container/range instead of individual iterators. Not sure if it makes the expression fit on one line, but it might.

iwyu_output.cc
```diff
@@ -2298,4 +2306,19 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
...
+           mapped_headers.end();
+  }
+  return includer.direct_includes_as_fileentries().count(GetFileEntry(decl));
```
kimgr, line 2321: This line speaks for itself, but I think a comment is needed for the case above where mappings exist to flesh out why mappings affect the decision making.

iwyu_output.cc
```diff
@@ -2298,4 +2306,19 @@ size_t IwyuFileInfo::CalculateAndReportIwyuViolations() {
   return num_edits;
 }
 
+bool IsDirectlyIncluded(const NamedDecl* decl, const IwyuFileInfo& includer) {
```
kimgr, line 2309: This feels like it belongs more with `IwyuPreprocessorInfo` or `IncludePicker` than the output module (it seems suspicious for `iwyu.cc` to ask `output` about core decisions). But this is really just a feeling, there may be good reasons to keep it in iwyu_output that I'm not seeing.

kimgr:
Some scattered thoughts:

* Can you cook up a small example of before/after? My imagination is slow these days, so I can't fully see how this relates to overload resolution/unresolved lookups.
* `IsDirectlyIncluded` seems like a more straightforward predicate than `PublicHeaderIntendsToProvide`
* It feels a little bit backwards to let the shape of the include graph affect IWYU analysis, but we have similar things already and I suspect it's unavoidable for more advanced analysis
* Moreover, I think `IsDirectlyIncluded` seems more generally useful (being able to ask "is this decl visible because of a direct include?" feels like a power tool)
* It would be nice to think about splitting `iwyu_stricter_than_cpp*` -- that test is growing a bit too large and multi-faceted for my taste. Does it attract changes because it covers most of the provision logic?

But overall, if we can find the right home for `IsDirectlyIncluded`, I don't mind a small regression for template default arguments; there seems to be enough problems with those that need to be solved either way.

bolshakov-a:
Thanks for the review!

> I can't fully see how this relates to overload resolution/unresolved lookups.

I tried to fix `UnresolvedLookupExpr`-related bugs as proposed in [my comment](https://github.com/include-what-you-use/include-what-you-use/issues/1250#issuecomment-2016487945). I used `PublicHeaderIntendsToProvide` to report directly included overloads (just to keep corresponding `#include`s). However, for the example from #1250, I had got the output like this:
```
#include <pstl/glue_algorithm_defs.h>  // for copy
#include <initializer_list>            // for begin, end

#include <pstl/glue_algorithm_defs.h>  // for copy
#include <algorithm>                   // for copy
#include <array>                       // for array
#include <cstddef>                     // for size_t
#include <initializer_list>            // for begin, end
#include <iterator>                    // for begin, end, copy
```
Transitive inclusions from public STL headers cause that `PublicHeaderIntendsToProvide` returns `true` for overloads not `#include`d directly, like from `<pstl/glue_algorithm_defs.h>` and `<initializer_list>`. I started thinking why the logic for `VisitOverloadExpr` should be different from the one for `CodeAuthorWantsJustAForwardDeclare` and realized that it should not. Actually, adding that simple mapping file to iwyu_stricter_than_cpp test case changes the output drastically because all the stuff contained in `-i1.h` and `-i2.h` becomes considered as provided by `-d1.h` even if it is not mapped explicitly. And IWYU started suggesting to `#include "...-i1.h"` for `IndirectStruct2` in e.g. `-typedefs.h` even when the both fwd-decl requirements are met.

I've decided to fix `CodeAuthorWantsJustAForwardDeclare` at first in this PR, and then to open a separate PR changing `VisitOverloadExpr`.

> Does it attract changes because it covers most of the provision logic?

I think it is its primary goal, at least for `typedef`/"autocast"/fnreturn triple. Default template arguments are slightly different because they use another logic at the use (i.e. template instantiation) site. Nonetheless, it doesn't look bad for me that such a test case coupling helps revealing problems.

iwyu.cc
bolshakov-a: I just wanted to keep the old behavior when `PublicHeaderIntendsToProvide` should return `false` on nullopt, which would cause `CodeAuthorWantsJustAForwardDeclare` to return `true`.

iwyu_output.cc
bolshakov-a: Nice, thanks! Probably, it would be possible to upgrade C++ standard version used up to C++20 in order to use the standard range algorithms?

iwyu_output.cc
bolshakov-a: I agree that it is strange place for it. The reason to insert it here was in that `IsDirectlyIncluded` takes `IwyuFileInfo` and could be considered as part of its interface and hence be placed next to it.

I've moved `GetMappedPublicHeaders` to `iwyu_include_picker` and `IsDirectlyIncluded` to `iwyu_preprocessor` (placing `IsDirectlyIncluded` into `iwyu_include_picker` would require additional dependency for the latter from `iwyu_output.h`). Still not sure that it is the best decision.

iwyu_output.cc
kimgr: I've just followed LLVM's lead to avoid surprises (and put less pressure on user toolchains), their lowest supported version is C++17

iwyu_output.cc
kimgr: Thanks for rationalizing, now I see it more clearly. I think the best would be:

* Make `GetMappedPublicHeaders` a member function of `IncludePicker` (would obviate the need to reach out to `GlobalIncludePicker` for itself :))
* Move `IsDirectlyIncluded` back into `iwyu_output`

I think that would make most sense.

kimgr:
Thanks, much clearer to me now! Finally made up my mind on where I think `IsDirectlyIncluded` and friends should live, we'll probably agree.

iwyu_output.cc
bolshakov-a: Authorities on C++ claim that [if a function can be nonmember nonfriend then it should be](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c4-make-a-function-a-member-only-if-it-needs-direct-access-to-the-representation-of-a-class). There is also [Herb Sutter's article](http://www.gotw.ca/gotw/084.htm) about it.

kimgr: Yeah, that makes sense for value types in general. But the dependency on `GlobalIncludePicker` in `GetMappedPublicHeaders` makes it almost absurd. We want clients to be able to use whatever `IncludePicker` they have available (which in most cases will be `GlobalIncludePicker()`, but that could change). So either we'd pass in `IncludePicker&` to `GetMappedPublicHeaders`, or we make it a member. The most obvious, IMO, is a member.

bolshakov-a: If `IsDirectlyIncluded` will be returned into `iwyu_output`, then `GetMappedPublicHeaders()` can be made `static` function of `iwyu_output.cc` again, in principle...

kimgr: Yeah, but I think it makes more sense as a member of `IncludePicker`, since it owns all the mappings

kimgr:
Thank you!

PR #1511, opened 2024-04-05
Keep directly included overloads

### LINKED ISSUE #636: std::move does not require <utility> with template parameters

bolshakov-a:
I've decided to report only that overloaded function decls which are already directly `#include`d, as opposed to the suggestion in the removed TODO, so as to avoid lots of unwanted `#include`s in real-world STL cases (things like `std::begin` may be defined in many headers). When no overload is included directly but the name is qualified (i.e. can not be found via ADL), an arbitrary (the first one) overloaded declaration is suggested to keep the code compilable.

Filtering of suitable overloads has not been done.

This fixes #636, #823, #1111, #1293, and, partially, #1250.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -139,6 +139,7 @@
 #include "iwyu_verrs.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/StringRef.h"
+#include "llvm/ADT/iterator_range.h"
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/TargetSelect.h"
@@ -2213,18 +2214,15 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     }
     const NamedDecl* first_decl = *expr->decls_begin();
     OptionalFileEntryRef first_decl_file_entry = GetFileEntry(first_decl);
-    for (OverloadExpr::decls_iterator it = expr->decls_begin();
-         it != expr->decls_end(); ++it) {
-      if (GetFileEntry(*it) != first_decl_file_entry)
+    for (const NamedDecl* decl : expr->decls()) {
+      if (GetFileEntry(decl) != first_decl_file_entry)
         return true;
     }
 
     // For now, we're only worried about function calls.
     // TODO(csilvers): are there other kinds of overloads we need to check?
     const FunctionDecl* arbitrary_fn_decl = nullptr;
-    for (OverloadExpr::decls_iterator it = expr->decls_begin();
-         it != expr->decls_end(); ++it) {
-      const NamedDecl* decl = *it;
+    for (const NamedDecl* decl : expr->decls()) {
       // Sometimes a UsingShadowDecl comes between us and the 'real' decl.
       if (const UsingShadowDecl* using_shadow_decl = DynCastFrom(decl))
         decl = using_shadow_decl->getTargetDecl();
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -223,7 +223,6 @@ using clang::NamespaceAliasDecl;
 using clang::NestedNameSpecifier;
 using clang::NestedNameSpecifierLoc;
 using clang::OptionalFileEntryRef;
-using clang::OverloadExpr;
 using clang::PPCallbacks;
 using clang::ParmVarDecl;
 using clang::PointerType;
@@ -260,6 +259,7 @@ using clang::TypedefDecl;
 using clang::TypedefNameDecl;
 using clang::TypedefType;
 using clang::UnaryExprOrTypeTraitExpr;
+using clang::UnresolvedLookupExpr;
 using clang::UsingDecl;
 using clang::UsingDirectiveDecl;
 using clang::UsingShadowDecl;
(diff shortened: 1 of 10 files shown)


iwyu.cc
```diff
@@ -2201,24 +2201,55 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
 
-    // Make sure all overloads are in the same file.
     if (expr->decls_begin() == expr->decls_end()) {
       // This can occur when there are no overloads before the template
       // definition, and a callee may be found only via ADL.
       return true;
     }
     const NamedDecl* first_decl = *expr->decls_begin();
     OptionalFileEntryRef first_decl_file_entry = GetFileEntry(first_decl);
+    OptionalFileEntryRef current_file_entry = GetFileEntry(CurrentLoc());
```
kimgr, line 2216: It's a little tempting to capture the information and pass it to `iwyu_output` to do the full analysis, e.g. something like `ReportOverloadUse`. That way iwyu.cc can focus more on traversing the AST and collecting interesting nodes, and less on the policy for those nodes (i.e. if they're directly included or not).

But there's enough tangling with current file entry and current AST node that that will probably be tricky. So just throwing it out there as a presumably-bad idea :)

iwyu.cc
```diff
@@ -2201,24 +2201,55 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // an iwyu warning now, even without knowing the exact overload.
   // In that case, we store the fact we warned, so we won't warn again
   // when the template is instantiated.
-  // TODO(csilvers): to be really correct, we should report *every*
-  // overload that callers couldn't match via ADL.
+  // Otherwise, all directly '#include'd overloads are reported just to
```
kimgr, line 2204: Could you just say `included` instead of `'#include'd` -- easier to read, and doesn't interfere with `#include` grepping.

tests/cxx/overload_expr-d2.h
```diff
@@ -0,0 +1,39 @@
...
+
+template <typename T>
+T TplFn2() {
+  // IWYU: ns::add is...*using_overload-int.h
+  return ns::add(T{}, T{});
+}
+
+template <typename T>
+void TplFn3() {
+  // It is allowed not to '#include' any definition of 'OverloadedFn' at this
```
kimgr, line 23: `include`

tests/cxx/using_overload-float.h
```diff
@@ -7,7 +7,11 @@
 //
 //===----------------------------------------------------------------------===//
 
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_USING_OVERLOAD_FLOAT_H_
```
kimgr, line 10: I'd prefer not to introduce coupling between tests by reusing this header (it appears to be part of `using_overload_{multiple,single}.cc`). But I see that it's annoying to add new headers for these nonsense decls. I also can't see any idiomatic way to add reusable two-part headers (we have `direct.h`/`indirect.h` that are sanctioned to reuse between tests, but there's no precedent for numbering there, I think).

bolshakov-a: There is no problem with adding specific headers. I was just concerned about `tests/cxx` folder bloating.

kimgr: Yeah, it's getting a bit crowded. We should probably think about further subdivision at some point, but let's keep them separated for now.

iwyu.cc
kimgr: Yeah, that makes sense. I think for me, looking at whether the decl is directly included is already a non-local concern. But I don't think we should do anything right now, the code is fine where it is, but I wish I had a clearer intuition for what kind of functionality should go where. It would be nice if there were clearer passes through IWYU's logic.

kimgr:
Thank you so much for working on this!

kimgr:
I've tested and closed all mentioned issues that no longer reproduce, thanks!

PR #1512, opened 2024-04-06
mappings: Map sig_atomic_t to signal.h

cdown:
Otherwise on glibc this maps to bits/types/sig_atomic_t.h, which is glibc internal.

### DIFF
--- a/gcc.symbols.imp
+++ b/gcc.symbols.imp
@@ -97,6 +97,7 @@
   { "symbol": [ "useconds_t", "private", "<sys/types.h>", "public"] },
   { "symbol": [ "wchar_t", "private", "<stddef.h>", "public"] },
   { "symbol": [ "wchar_t", "private", "<stdlib.h>", "public"] },
+  { "symbol": [ "sig_atomic_t", "private", "<signal.h>", "public"] },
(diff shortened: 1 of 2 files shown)


cdown:
I see that some of this is from a grep -- my run of that doesn't generate the same list, but this is one of the cases where __foo_defined can go wrong:

```
% grep -R 'sig_atomic_t' /usr/include
/usr/include/bits/types/sig_atomic_t.h:#ifndef __sig_atomic_t_defined
/usr/include/bits/types/sig_atomic_t.h:#define __sig_atomic_t_defined 1
/usr/include/bits/types/sig_atomic_t.h:typedef __sig_atomic_t sig_atomic_t;
[...]
/usr/include/signal.h:#include <bits/types/sig_atomic_t.h>
```

Maybe there's another way you'd suggest to go about this?

kimgr:
@cdown I think particularly this `grep` pipeline may have rotted, we've had many patches to manually tweak the symbol mappings over the years, and I didn't put any effort into seamless automation. The pipeline doesn't really work on my system either.

So I think it's fine to hand-edit this for now, and if we can think of better automation later, we can always replace them.

Thanks!

Issue #1525, opened 2024-04-17
Processing overloads from a sole file


bolshakov-a:
IWYU has [a special logic](https://github.com/include-what-you-use/include-what-you-use/blob/93d8793081e5aa457114ef99d1e887390b370e4a/iwyu.cc#L2256-L2289) for handling overloaded functions when they are all in a single file: it suggests to include it in the file defining a template referring to such a set, and suggests not to include in files instantiating such a template. The problem is that whether all the overloads are in a single file or not may depend on the translation unit currently processed. Consider:
```cpp
// overloaded-indirect.h

namespace ns {
void OverloadedFn(int);
void OverloadedFn(char);
}


// overloaded-direct.h

#include "overloaded-indirect.h"


// overloaded-double.h

namespace ns {
void OverloadedFn(double);
}


// tpl.h

#include "overloaded-direct.h"

template <typename T>
void TplFn(T t) {
  ns::OverloadedFn(t);
}


// tpl.cpp

#include "overloaded-double.h"
#include "tpl.h"


// user.cpp

#include "tpl.h"

void User() {
  TplFn(1);
}

```
Note that, in `tpl.cpp`, `#include "overloaded-double.h"` is before including the associated header `tpl.h`. This may occur e.g. because `#include "overloaded-double.h"` is placed in `stdafx.h`, while `user.cpp` may be from another project not using PCH.

When running IWYU on `user.cpp`, clang finds only two overloads from `overloaded-indirect.h`, and IWYU considers hence that `overloaded-indirect.h` should be included directly in `tpl.h`, and doesn't suggest to add anything to `user.cpp`. However, when running on `tpl.cpp`, the third overload from the different file is seen, and IWYU suggests to replace `#include "overloaded-direct.h"` with `#include "overloaded-double.h"` in `tpl.h` (because `OverloadedFn(double)` appears to be the first decl; or not to include anything at all in `tpl.h` if `OverloadedFn` were in the global namespace). It may result in that the code would not be compilable after IWYU run. Even worse, in my example, the code would still be compiled, but with the wrong `OverloadedFn` overload called (because `int` is convertible to `double`)!

Is `processed_overload_locs`-related logic really useful for `UnresolvedLookupExpr`, or may be removed? Or only `AddProcessedOverloadLoc` call should be removed from `VisitUnresolvedLookupExpr`, keeping the logic only at the template definition site? @kimgr, wdyt?

kimgr:
[...]

PR #1528, opened 2024-04-23
Remove same-file-overload-set logic at inst side

bolshakov-a:
See discussion in #1525.

Besides, this fixes reporting ADL-found function declarations when all the non-ADL declarations are in a single file. The corresponding test case has been added to `overload_expr`.

### DIFF
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -1121,16 +1121,21 @@ int main() {
 
   // Calling an overloaded function.  In the first two cases,
   // CallOverloadedFunctionSameFile() is responsible for the call,
-  // since it's just a single file.  In the second two cases, we
-  // can't know the file required until now (when the templated
-  // function is instantiated).
+  // since it's just a single file. But in general, we cannot know it for sure.
+  // Hence it is reported because the template-defining file doesn't include
+  // the function-defining file directly. In the second two cases, we can't know
+  // the file required until now (when the templated function is instantiated).
+  // IWYU: I1_OverloadedFunction is...*badinc-i1.h
   CallOverloadedFunctionSameFile(5);
+  // IWYU: I1_OverloadedFunction is...*badinc-i1.h
   CallOverloadedFunctionSameFile(5.0f);
   // IWYU: I1_And_I2_OverloadedFunction is...*badinc-i1.h
   CallOverloadedFunctionDifferentFiles(5);
   // IWYU: I1_And_I2_OverloadedFunction is...*badinc-i2.h
   CallOverloadedFunctionDifferentFiles(5.0f);
-  // This should not be an IWYU violation either: the iwyu use is in the fn.
+  // Again, it is reported because the template defining file doesn't include
+  // the function defining file directly.
+  // IWYU: i1_ns::I1_NamespaceTemplateFn is...*badinc-i1.h
   CallOverloadWithUsingShadowDecl(5);
 
   // Calling operator<< when the first argument is a macro.  We should
--- a/tests/cxx/overload_expr-d2.h
+++ b/tests/cxx/overload_expr-d2.h
@@ -25,15 +25,25 @@ void TplFn3() {
   return OverloadedFn(T{});
 }
 
+template <typename T>
+void TplFn4() {
+  // All the overloads which are visible without ADL are in a single file, hence
+  // report it here even if not included directly.
+  // IWYU: OverloadedFn2 is...*overload_expr-i4.h
+  OverloadedFn2(T{});
+}
+
 /**** IWYU_SUMMARY
 
 tests/cxx/overload_expr-d2.h should add these lines:
+#include "tests/cxx/overload_expr-i4.h"
 #include "tests/cxx/overload_expr-int.h"
 
 tests/cxx/overload_expr-d2.h should remove these lines:
 - #include "tests/cxx/overload_expr-i1.h"  // lines XX-XX
 
 The full include-list for tests/cxx/overload_expr-d2.h:
+#include "tests/cxx/overload_expr-i4.h"  // for OverloadedFn2
 #include "tests/cxx/overload_expr-int.h"  // for add
 
 ***** IWYU_SUMMARY */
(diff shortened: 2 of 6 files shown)


tests/cxx/badinc.cc
```diff
@@ -1119,18 +1119,16 @@ int main() {
   // IWYU: I1_PtrDereferenceClass is...*badinc-i1.h
   local_i1_ptrdereference_class->a();
 
-  // Calling an overloaded function.  In the first two cases,
```
kimgr, line 1122: This looks like a regression to me; would it be warranted to add a TODO here to explain that this is not ideal, and what's blocking us from improving?

Or am I misunderstanding, and this is actually better?

tests/cxx/overload_expr-d2.h
```diff
@@ -25,15 +25,23 @@ void TplFn3() {
   return OverloadedFn(T{});
 }
 
+template <typename T>
+void TplFn4() {
+  // IWYU: OverloadedFn2 is...*overload_expr-i4.h
```
kimgr, line 30: Can you add a comment here to tell the story of why this depends on `OverloadedFn2(int)` and/or `OverloadedFn2(char)`?

kimgr: It could be argued that this expression should not depend on any known `OverloadedFn2` overload, because the type used for overload resolution is a template argument, unbound at this point.

It would be nice to cover the whys and whats of why that isn't happening here too.

tests/cxx/badinc.cc
bolshakov-a: The primary reason of the changes was to fix reporting ADL-found overloads at the use side. The problem is because when all the non-ADL overloads are in the same file, the calling expression appears in `processed_overload_locs` and is ignored. Fixing the bug while keeping the current logic would introduce additional complexity, but then I realized that the current logic cannot work well in all cases: in general, while scanning a user of `badinc.h` IWYU cannot in principle say that it will actually suggest to include `badinc-i1.h` in `badinc.h` while fixing `badinc.h`, as far as it doesn't analyze all the translation units simultaneously (see the example in #1525).

tests/cxx/overload_expr-d2.h
bolshakov-a: > It could be argued that this expression should not depend on any known OverloadedFn2 overload, because the type used for overload resolution is a template argument, unbound at this point.

Reporting of the possible overloads of an unbound (dependent) function call was exactly the original purpose of `VisitOverloadExpr`. The suggestion should be the same as on the master branch and as on the old IWYU releases, I suspect (AFK now). I suggested to remove the old logic in #1529.

tests/cxx/badinc.cc
kimgr: Thanks for the summary. I suggest adding something like:
```
  // TODO: CallOverloadedFunction should really be responsible for its
  // implementation details, in this case which function overload is called. We
  // don't yet handle the combination of overload resolution and template
  // instantiation with full fidelity.
```
just to make it clear that the tested behavior is not really the desired behavior.

bolshakov-a: My point is that it's not achievable for IWYU. IWYU _should_ suggest `#include` in such cases in general, because it doesn't know whether it will actually suggest it in tpl-defining header or not.

kimgr: I see. So the old behavior was a heuristic that masked that fact in some cases, but also caused other issues... Right?

I know better than to argue with you about template analysis in IWYU :slightly_smiling_face:, but these attributions look obviously wrong to me, and I'm concerned that they will make IWYU unusable. So the improvements come for instantiation-side ADL resolution -- I don't have any intuition for how much more common that is than non-ADL lookup on definition side. It's essentially these two constructs played out against each other, right?

To cover both cases, it seems like we need some kind of author-intent hint or pragma, then, eventually?

kimgr: I had some time away from the keyboard, and just wanted to clarify: it's not super important to me what the eventual behavior is, but my priority is to make sure there's enough information available in the tests and code that someone sitting down with this problem later will have some context for why IWYU behaves this way.

So I don't want to block this patch on functional concerns, I just want to make sure there's some rationale for _why_ `CallOverloadedFunction.*` don't claim ownership of the implementation-side functions.

bolshakov-a: > So the old behavior was a heuristic that masked that fact in some cases, but also caused other issues... Right?

It was a heuristic, but rather the issues were just not fully understood.

> I'm concerned that they will make IWYU unusable

That would be a strong statement! IWYU with the similar patch was successfully used at my previous work.

> I don't have any intuition for how much more common that is than non-ADL lookup on definition side

Anyway, IWYU should produce compilable code, hence some redundancy is better (assuming the presence of double-inclusion guards in headers, of course).

> To cover both cases, it seems like we need some kind of author-intent hint or pragma, then, eventually?

It works already: just include the function-defining header in the template-defining header directly so that IWYU doesn't suggest `#include` at the instantiation side. I even stated that in [my original commit](https://github.com/bolshakov-a/include-what-you-use/commit/45ae521e301aaf29c073d32a2803bb24fa671ec4#diff-b4ba3b28c8dd51625839eed7fa55c371893205bb3ed9142b7884a913dc68d129R1141). Probably, the comments should be copied from there?

kimgr: > Anyway, IWYU should produce compilable code, hence some redundancy is better (assuming the presence of double-inclusion guards in headers, of course).

This is a very fair point! 

> It works already: just include the function-defining header in the template-defining header directly so that IWYU
> doesn't suggest #include at the instantiation side. I even stated that in
> [my original commit](https://github.com/bolshakov-a/include-what-you-use/commit/45ae521e301aaf29c073d32a2803bb24fa671ec4#diff-b4ba3b28c8dd51625839eed7fa55c371893205bb3ed9142b7884a913dc68d129R1141). Probably, the comments should be copied from there?

Yeah, that would make sense to me -- those comments help clarify a lot.

kimgr:
Thank you!

PR #1536, opened 2024-05-08
Reduce libc mappings to really include what you use (part 3)

### LINKED ISSUE #1531: Reduce symbol mappings to be even more fine-grained (part 3)
Link: <https://github.com/include-what-you-use/include-what-you-use/pull/1344>

I'll complete this bug report when I find some time.  :-)

alejandro-colomar:
The mappings removed here are guaranteed by the relevant standards, but are not the canonical includes for the identifiers.  To be strict with IWYU guidelines, remove those mappings, keeping only the canonical ones.

Closes: <https://github.com/include-what-you-use/include-what-you-use/issues/1531>


alejandro-colomar:
Cc: @firewave, @kimgr

alejandro-colomar:
This is the last step.

kimgr:
Please be careful with those tests, they're our greatest asset :)

I recommend changing one symbol at a time to get more focused failures. Make the smallest change possible that's still an improvement, and then it's easier to make tradeoffs with tests.

Unfortunately the current master is broken vs. upstream Clang, so it might be a while until we're back on our feet.

alejandro-colomar:
> Please be careful with those tests, they're our greatest asset :)
> 
> I recommend changing one symbol at a time to get more focused failures. Make the smallest change possible that's still an improvement, and then it's easier to make tradeoffs with tests.
> 
> Unfortunately the current master is broken vs. upstream Clang, so it might be a while until we're back on our feet.

Hmmm; okay.  I'll prepare a patch set, changing one per patch.

kimgr:
@alejandro-colomar Master is working again, if you feel like trying a smaller patch. Thanks!

alejandro-colomar:
> @alejandro-colomar Master is working again, if you feel like trying a smaller patch. Thanks!

Thanks!  And sorry for the delay.  I've opened the first one here: <https://github.com/include-what-you-use/include-what-you-use/actions/runs/9856379004/job/27213243818?pr=1566>.

I'll use one PR per identifier, to go step by step.

firewave:
The `size_t` changes go a bit too far according to https://en.cppreference.com/w/c/types/size_t.

Same for `NULL`: https://en.cppreference.com/w/c/types/NULL.

`va_list` changes look fine: https://en.cppreference.com/w/c/variadic/va_list

alejandro-colomar:
> The `size_t` changes go a bit too far according to https://en.cppreference.com/w/c/types/size_t.

As said in another PR, don't trust cppreference too much.

If you want to go to the source, read the ISO C standard.  It's not too difficult to read.  You can just grep for a specific type, and you'll find it (plus some noise, of course).  Clause 7 is for libc.

If you don't want to do that, the second best reference is the manual page, which I wrote comparing to the ISO C and POSIX.1 standards:

<https://www.man7.org/linux/man-pages/man3/ssize_t.3type.html>

> 
> Same for `NULL`: https://en.cppreference.com/w/c/types/NULL.

<https://www.man7.org/linux/man-pages/man3/NULL.3const.html>

> 
> `va_list` changes look fine: https://en.cppreference.com/w/c/variadic/va_list

<https://www.man7.org/linux/man-pages/man3/va_list.3type.html>

cppreference seems to be documenting which headers provide a type according to ISO C (ignoring POSIX.1), but it doesn't say which is the one that actually defines the type.

firewave:
Thanks. That are the types of references which would be good to provide in the PR descriptions.

> As said in another PR, don't trust cppreference too much.

LLVM puts a lot of trust in it so if it known that it is has incorrect information we should probably try to address those. It is after all community maintained.

alejandro-colomar:
> Thanks. That are the types of references which would be good to provide in the PR descriptions.
> 
> > As said in another PR, don't trust cppreference too much.
> 
> LLVM puts a lot of trust in it so if it known that it is has incorrect information we should probably try to address those.

From what I see, it only informs about ISO C.  It ignores POSIX.  I guess they only care about ISO C, which is fine, but for IWYU (and LLVM), POSIX.1 is also relevant.

It also doesn't note which header is the canonical provider of an identifier.

For example, the canonical header for `WCHAR_MIN` is `<stdint.h>`.  `<wchar.h>` is another provider of the type, but it is described in 7.20.3 in the standard, which makes the canonical header be `<stdint.h>` (7.20).

In <https://en.cppreference.com/w/c/string/wide>, for `WCHAR_MIN`, `<wchar.h>` is shown above `<stdint.h>`.  That is probably a mistake, and they could fix it.

> It is after all community maintained.

Ahh, I didn't know that.  Nice.  :)

Thanks!

firewave:
Even though it touches the tests I think the `va_list` change could still be applied before the next release as its scope is rather limited.

firewave:
> Even though it touches the tests I think the `va_list` change could still be applied before the next release as its scope is rather limited.

I still think we should try to get that into the upcoming release.

alejandro-colomar:
Your wishes are available for review.  :)

<https://github.com/include-what-you-use/include-what-you-use/pull/1583>

alejandro-colomar:
I forgot about this.  I'll have a look soon.

firewave:
Looks like this is superseded by #1990.

firewave:
CC @bolshakov-a

kimgr:
@alejandro-colomar #1990 has landed on master. I think we can close this?

alejandro-colomar:
Yup; thanks!

PR #1542, opened 2024-05-22
Replace LLVM_LIBRARY_DIR with clang -print-resource-dir

tuliom:
The variable LLVM_LIBRARY_DIR points to the directory where LLVM libraries are installed. This may not be the same directory where clang's include directory is available.

Using `clang -print-resource-dir` will guarantee the correct path is always taken.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,9 +53,13 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
 # Synthesize clang-resource-headers target if necessary.
 if (NOT TARGET clang-resource-headers)
+  find_program(CLANG_EXECUTABLE NAMES clang${LLVM_VERSION_MAJOR} clang-${LLVM_VERSION_MAJOR} clang REQUIRED)
+  execute_process(COMMAND ${CLANG_EXECUTABLE} -print-resource-dir
+    OUTPUT_VARIABLE clang_resource_dir)
+  string(STRIP "${clang_resource_dir}" clang_resource_dir)
+  set(clang_headers_src ${clang_resource_dir}/include)
   # Use only LLVM_VERSION_MAJOR for the resource directory structure; some
   # platforms include suffix in LLVM_VERSION.
-  set(clang_headers_src "${LLVM_LIBRARY_DIR}/clang/${LLVM_VERSION_MAJOR}/include")
   set(clang_headers_dst "${CMAKE_BINARY_DIR}/lib/clang/${LLVM_VERSION_MAJOR}/include")
 
   file(GLOB_RECURSE in_files RELATIVE "${clang_headers_src}"


CMakeLists.txt
```diff
@@ -53,9 +53,12 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
 # Synthesize clang-resource-headers target if necessary.
 if (NOT TARGET clang-resource-headers)
+  execute_process (COMMAND clang -print-resource-dir
```
xvitaly, line 56: `find_program(CLANG_EXECUTABLE NAMES clang16 clang17 clang REQUIRED)` should be used here for searching for supported Clang versions.

tuliom: @xvitaly using `find_program()` is a great suggestion! But I have some concerns with the parameters being used:

1. This list would have to be reviewed and updated frequently in order to update the supported versions.
2. I think we need to support both formats `clang17` and `clang-17`.
3. I don't think we want to prioritize clang16 over other versions.
4. We should use the same version found previously with `find_package()`.

Please let me know what you think about this implementation I'm proposing.

xvitaly: Looks great now. Thank you.

CMakeLists.txt
```diff
@@ -53,9 +53,13 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
 # Synthesize clang-resource-headers target if necessary.
 if (NOT TARGET clang-resource-headers)
+  find_program(CLANG_EXECUTABLE NAMES clang${LLVM_VERSION_MAJOR} clang-${LLVM_VERSION_MAJOR} clang REQUIRED)
```
kimgr, line 56: I think it's a little risky to include `clang` here -- it might refer to a version of Clang that is not compatible with the version we're linking with.

Maybe for this to work it would be safer to try and check that `clang --version` matches `LLVM_VERSION` somehow?

kimgr: It's a little uncomfortable that we're jumping outside the CMake-published config entirely trying to find headers that might match. I wonder if:

* LLVM/Clang's CMake modules contain a variable for the resource dir already
* ... and if not, whether it would be possible to get a patch in there to add it?

tuliom: > LLVM/Clang's CMake modules contain a variable for the resource dir already

LLVM-internal projects already have access to [get_clang_resource_dir](https://github.com/llvm/llvm-project/blob/main/cmake/Modules/GetClangResourceDir.cmake#L9). But other projects are expected to get this information from clang.

More info: https://github.com/llvm/llvm-project/tree/main/cmake#what-goes-here

kimgr: Yes, that page describes the current state. But I wonder if a case could be made to move `get_clang_resource_dir` into the "installed modules" in Clang/LLVM's CMake system.

I still feel like using `clang*` to probe is likely to fail, e.g. with:

* `clang` command is not installed, only libraries
* we accidentally pick the wrong `clang` and get a resource dir that's not compatible with libraries
* we're missing some weird `clang` decoration and don't find any binary

So it would be nice to push the challenge upstream :)

kimgr:
What distribution are you on where this breaks, btw?

CMakeLists.txt
kimgr: I opened https://github.com/llvm/llvm-project/issues/95747 to see if it's possible to do anything on the Clang side (I'm not too good with CMake, so I don't have a clear idea how hard this is).

kimgr: Also raised this on the LLVM forums to see if someone bites: https://discourse.llvm.org/t/export-clang-resource-dir-from-clangconfig-cmake/80000

kimgr:
@tuliom I pushed a very WIP branch at https://github.com/kimgr/include-what-you-use/commits/cmake-builtin-headers based on input from the LLVM crowd -- could you see how well that works in your environment? Thanks!

tuliom:
@kimgr I've just provided some feedback to your PR.
I also started the discussion upstream about the dependency on `CLANG_RESOURCE_DIR`.
Let's see what the community thinks about it.

PR #1543, opened 2024-05-23
Fix README.md on use python scripts

amelonpie:


README.md
```diff
@@ -81,7 +81,7 @@ This build mode assumes you already have compiled LLVM and Clang libraries on yo
       iwyu/build$ cmake -G "Unix Makefiles" -DIWYU_LLVM_ROOT_PATH=/usr/lib/llvm-6.0 ../include-what-you-use
 
       # For IWYU 0.11/Clang 7 and later
-      iwyu/build$ cmake -G "Unix Makefiles" -DCMAKE_PREFIX_PATH=/usr/lib/llvm-7 ../include-what-you-use
+      iwyu/build$ cmake .. -G "Unix Makefiles" -DCMAKE_PREFIX_PATH=/usr/lib/llvm-7
```
kimgr, line 84: This won't work if you've followed the checkout instructions before. And it's also inconsistent with the `# For IWYU 0.10...` section above.

README.md
```diff
@@ -118,6 +118,13 @@ So for IWYU to function correctly, you need to copy the Clang `include` director
 
 This weirdness is tracked in [issue 100](https://github.com/include-what-you-use/include-what-you-use/issues/100), hopefully we can make this more transparent over time.
 
+Another way to insatall is
```
kimgr, line 121: typo: insatall -> install

kimgr, line 121: Hmm, actually, this new paragraph seems to imply you can use `make install` to work around all the Clang dependencies described just above. It definitely doesn't.

If anything, we could add a little paragraph at the top that says "To install, use the `install` target", but that's pretty much implied.

kimgr:
Thanks for the suggestions, but I think these changes would ultimately make the README worse. That doesn't mean I think the current README is great, just that your suggested changes make it self-contradictory and self-inconsistent. What problems did you run into when you followed it?

PR #1545, opened 2024-05-26
[clang compat] Use StringRef::operator==

kimgr:
llvm/llvm-project@de483ad513895c0adf7f21c7001c30f031998ea3 marked StringRef::equals deprecated.

Update our only use to overloaded operator==.


kimgr:
There are multiple breakages on master right now, so I'll merge this against a red build. It gets rid of a large number of deprecation warnings.

PR #1549, opened 2024-06-01
Not in alpha anymore

ShalokShalom:
I am assuming, this is not alpha anymore after 5 years of additional development since 2018

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This puts us in a state where every file includes the headers it needs to declar
 
 ### CAVEAT ###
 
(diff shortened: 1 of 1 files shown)


kimgr:
I'm not sure what the definition of "alpha" is, but we're still in very experimental development, so I think it makes sense to set correct expectations. Why are you anxious to remove it?

ShalokShalom:
I am fine with leaving it. I did expect, after 5 years of additional development, it would be stable by now.

ShalokShalom:
I would eventually adjust the date. If its still alpha, consider changing the 2018 to 2024.
As its still relevant. It looks outdated.

Issue #1538, opened 2024-05-15
No member named 'getTypeAsWritten' in 'clang::ClassTemplateSpecializationDecl' in upstream latest clang


gamesh411:
When compiling with the latest llvm-main/main, the following compilation error occurs:
```
.../include-what-you-use/iwyu.cc:4080:38: error: no member named 'getTypeAsWritten' in 'clang::ClassTemplateSpecializationDecl'
4080 |       const TypeLoc type_loc = decl->getTypeAsWritten()->getTypeLoc();
```
I suspect this is caused by a recently landed PR: https://github.com/llvm/llvm-project/pull/91393
This is a re-application of this change: https://github.com/llvm/llvm-project/pull/81642
The interface of template specializations is changed.

kimgr:
Thanks, I'm trying to wrinkle my head around this.

bolshakov-a:
It really appears to be slightly tricky and requires some refactoring. I'm working on this too.

kimgr:
@bolshakov-a Glad to hear it, I had just about given up 🥲.

bolshakov-a:
Hey @kimgr, this can be closed now.

kimgr:
Fixed on master as of dc9fd2c258a9f29881d73330c9b6fdb13b41442c.

PR #1554, opened 2024-06-06
[clang compat] Update for TemplateTypeParmDecl API change

kimgr:
llvm/llvm-project@e42b799bb28815431f2c5a95f7e13fde3f1b36a1 redesigned the way the type is exposed from function template default type template arguments.

Update our use, and shuffle things around a little for readability.


kimgr:
We still have one more API compat issue pending (#1538), so this doesn't fully build + test. But the necessary refactor looked straightforward enough.

Issue #1552, opened 2024-06-06
False possive with GCC lib with <chrono>


correaa:
I am using g++ 13 and iwyu 14

```shell
g++ (Ubuntu 13.2.0-4ubuntu3) 13.2.0

include-what-you-use 0.18 based on Ubuntu clang version 14.0.6
```

And my code that uses `std::chrono::high_resolution_clock` makes IWYU to ask for 

```cpp
/home/user/boost-multi/include/boost/multi/adaptors/fftw/test/transpose_square.cpp should add these lines:
#include <bits/chrono.h>                  // for operator-, duration, system...

/home/user/boost-multi/include/boost/multi/adaptors/fftw/test/transpose_square.cpp should remove these lines:
- #include <chrono>  // lines 10-10
```

I think this is a false positive that might suggest a needed modification in `/usr/share/include-what-you-use/gcc.stl.headers.imp` consisting of adding:

```
    { "include": ["@<bits/chrono.h>", "private", "<chrono>", "public"] },
```

Is this a reasonable request?

Thanks!

PS: this seems a near duplicate of https://github.com/include-what-you-use/include-what-you-use/issues/1277 , close if necessary

kimgr:
Yeah, I think that would make sense. Don't forget the built-in mapping table at https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_include_picker.cc#L655 if you want to make this change.

correaa:
do you mean that it is better to add this as a built-in mapping? modifying the .cc file only?

not sure what is the criterion for one or the other.

kimgr:
The built-in mapping in .cc will always be used, the external mapping in `*.imp` will only be used if explicitly provided on command-line with `-Xiwyu --mapping_file=...`.

correaa:
Ok, are you trying to say that it makes sense to have mappings for the standard library in the .cc file? 

Seems reasonable, but I am trying to understand the logic since there is also a `stl.imp` in the iwyu distribution.

kimgr:
Yeah, it's a little confusing:

* At first there were only built-in mappings in the .cc file
* I contributed external mappings, with the intent of removing the built-in mappings (as they are not strictly portable)
* I think we failed to figure out how to select external mappings, or something like that (~10 years ago, so not sure), so the built-in mappings stayed on
[...]

PR #1555, opened 2024-06-10
Map bits/chrono.h to chrono

correaa:
As discussed in https://github.com/include-what-you-use/include-what-you-use/issues/1552

Not sure if `@` is needed


kimgr:
@correaa Thanks! `@` is not necessary; it's a marker to say that the mapping contains a regex pattern rather than an exact match.

kimgr:
This looks good to me (I would prefer it in a single commit, but I can squash it on merge). I'll just postpone merging until we've managed to fix the build error on master, so we can see that it runs clean.

kimgr:
@correaa Master is back to green; could I ask you to rebase and repush to run the build steps with this patch? Thanks.

correaa:
I didn't know how to rebase from the Github GUI, so accidentally opened a different PR #1560

PR #1556, opened 2024-06-15
[clang compat] Fix explicit instantiation handling

bolshakov-a:
llvm/llvm-project@12028373020739b388eb2b8141742509f1764e3c removed `TemplateSpecializationType` child node from
`ClassTemplateSpecializationDecl`, replacing it just with template arguments.

To maintain instantiation scanning, the new `InstantiatedTemplateVisitor` interface and other functions have been introduced which don't require `TemplateSpecializationType`. They may be further reused for handling the cases then a specific `TemplateSpecializationType` is lost.

The comment in `VisitClassTemplateSpecializationDecl` was wrong: it is not clang but IWYU `GetLocation` function which attributes `ClassTemplateSpecializationDecl` to the original template location. An exception for explicit instantiations has been added to it. However, method declarations should still be attributed to the class template definition, hence the change for them to maintain the old behavior.

Because `GetLocation` doesn't "canonicalize" explicit instantiation locations now, the change in `VisitTemplateSpecializationType` is needed to handle the cases when the instantiation is explicit (as opposed to implicit ones). Besides, it looks more straightforward to report template declarations directly, not through
`ClassTemplateSpecializationDecl`.

A couple of new test cases added to make sure that the work has been done properly.

One functional change: explicit instantiation definitions don't require explicit instantiation declarations. [I think it was just useless](https://github.com/include-what-you-use/include-what-you-use/pull/1202#discussion_r1639933032).

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2457,7 +2457,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreCurrentASTNode() || CanIgnoreType(type))
       return true;
 
-    const NamedDecl* decl = TypeToDeclAsWritten(type);
+    // If not a full explicit specialization, report the original template
+    // or a partial specialization from which the specialization is instantiated
+    // (and don't report explicit instantiations here).
+    const NamedDecl* decl = GetInstantiatedFromDecl(TypeToDeclAsWritten(type));
 
     // If we are forward-declarable, so are our template arguments.
     if (CanForwardDeclareType(current_ast_node())) {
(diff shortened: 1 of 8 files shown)


iwyu.cc
```diff
@@ -4063,35 +4093,44 @@ class IwyuAstConsumer
...
-      const ASTNode type_loc_node(&type_loc);
-      const TemplateInstantiationData data =
-          GetTplInstData(type_loc.getTypePtr());
-      // Clang instantiates methods in the first ("canonical") spec decl context
-      // (which may correspond to instantiation declaration, not to definition).
-      for (const CXXMethodDecl* member : decl->getCanonicalDecl()->methods()) {
-        instantiated_template_visitor_.ScanInstantiatedFunction(
-            member, &type_loc_node, data.resugar_map, data.provided_types);
+      // Report the original template definition, or a partial or full
+      // specialization, which appropriate. GetTagDefinition returns the right
```
kimgr, line 4108: which -> whichever is

kimgr:
My gratitude knows no bounds; this looks great! One little comment nit, otherwise I'm glad to see this get mainline back to green.

bolshakov-a:
No need to thank me!

PR #1559, opened 2024-06-20
[iwyu_tool] Default to -j=0 and infer the core count in that case

pvutov:
Make the behavior of the -j argument more similar to other tools which offer it. Instead of hanging forever when "zero jobs" are requested, attempt to automatically pick a job count suitable for the current system. Default to this behavior if the user has not explicitly specified some other job count.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -497,7 +497,10 @@ def _bootstrap(sys_argv):
                         choices=FORMATTERS.keys(), default=DEFAULT_FORMAT,
                         help='Output format (default: %s)' % DEFAULT_FORMAT)
     parser.add_argument('-j', '--jobs', type=int, default=1,
-                        help='Number of concurrent subprocesses')
+                        nargs='?', const=0,
+                        help=('Number of concurrent subprocesses. If zero, '
+                              'will try to match the logical cores of the '
+                              'system.'))
     parser.add_argument('-l', '--load', type=float, default=0,
                         help='Do not start new jobs if the 1min load average is greater than the provided value')
     parser.add_argument('-p', metavar='<build-path>', required=True,
@@ -517,8 +520,12 @@ def _bootstrap(sys_argv):
     argv, extra_args = partition_args(sys_argv[1:])
     args = parser.parse_args(argv)
 
+    jobs = args.jobs
+    if jobs == 0:
+        jobs = os.cpu_count() or 1
+
     return main(args.dbpath, args.source, args.verbose,
-                FORMATTERS[args.output_format], args.jobs, args.load, extra_args)
+                FORMATTERS[args.output_format], jobs, args.load, extra_args)
 
 
 if __name__ == '__main__':
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -502,7 +502,8 @@ def _bootstrap(sys_argv):
                               'will try to match the logical cores of the '
                               'system.'))
     parser.add_argument('-l', '--load', type=float, default=0,
-                        help='Do not start new jobs if the 1min load average is greater than the provided value')
+                        help=('Do not start new jobs if the 1min load average '
+                              'is greater than the provided value'))
     parser.add_argument('-p', metavar='<build-path>', required=True,
                         help='Compilation database path', dest='dbpath')
     parser.add_argument('source', nargs='*',
--- a/iwyu_tool_test.py
+++ b/iwyu_tool_test.py
@@ -363,6 +363,35 @@ class BootstrapTests(unittest.TestCase):
         self.assertEqual(['arg1', '--', 'another_arg1'],
                          self.main.call_args['extra_args'])
 
+    def test_jobs_normal(self):
+        """ Nonzero values for -j are forwarded as they are. """
+        argv = ['iwyu_tool.py', '-p', '.', '-j', '4']
+        iwyu_tool._bootstrap(argv)
+        self.assertEqual(4, self.main.call_args['jobs'])
+
+
+    def test_jobs_zero(self):
+        """ -j=0 is translated to a nonzero job count. """
+        argv = ['iwyu_tool.py', '-p', '.', '-j', '0']
+        iwyu_tool._bootstrap(argv)
+        self.assertIsNotNone(self.main.call_args['jobs'])
+        self.assertLess(0, self.main.call_args['jobs'],
+                        'The automatically chosen job count should be >0.')
+
+    def test_jobs_novalue(self):
+        """ It is allowed to use -j without a value. """
+        argv = ['iwyu_tool.py', '-j', '-p', '.']
+        iwyu_tool._bootstrap(argv)
+        self.assertIsNotNone(self.main.call_args['jobs'])
+        self.assertLess(0, self.main.call_args['jobs'],
+                        'The automatically chosen job count should be >0.')
+
+    def test_jobs_default(self):
+        """ By default, we run singlethreaded. """
+        argv = ['iwyu_tool.py', '-p', '.']
+        iwyu_tool._bootstrap(argv)
+        self.assertEqual(1, self.main.call_args['jobs'])
+
 
 class CompilationDBTests(unittest.TestCase):
     def setUp(self):


kimgr:
Nice, thanks! I didn't know about the -j0 convention.

pvutov:
On second thought, I may have overstated the "convention".

- GNU make offers `make -j`. If no value is provided, the job count is automatically determined. If 0 is provided, they actually complain about wanting a positive integer. If `-j` is not provided at all, runs single-threaded.

- ninja runs multithreaded with or without `-j`, and also supports `-j=0`

- `run-clang-tidy` accepts both `-j` and `-j=0`.

- `cppcheck` does not accept `-j` and hangs on `-j=0` (same as current iwyu).

tldr I will update the patch tonight to also support `-j` with no value. I assume it's fine to run multithreaded if the user has not provided `-j` at all?

kimgr:
National holiday yesterday, so I didn't get a chance to look into this.

> I assume it's fine to run multithreaded if the user has not provided -j at all?

I would actually prefer to be single threaded unless -j is added. I'll take a closer look at the patch later today.

iwyu_tool_test.py
```diff
@@ -363,6 +363,28 @@ def test_extra_args_with_sep(self):
...
+        self.assertNotEqual(None, self.main.call_args['jobs'])
+
+    def test_jobs_novalue(self):
+        """ It is allowed to use -j without a value """
+        argv = ['iwyu_tool.py', '-j', '-p', '.']
+        iwyu_tool._bootstrap(argv)
+        self.assertNotEqual(0, self.main.call_args['jobs'])
+        self.assertNotEqual(None, self.main.call_args['jobs'])
+
+
```
kimgr, line 387: I'd like to see a testcase here for a command-line without `-j` -- I think that will currently evaluate to `jobs == 0`.

iwyu_tool.py
```diff
@@ -496,8 +496,11 @@ def custom_help():
     parser.add_argument('-o', '--output-format', type=str,
                         choices=FORMATTERS.keys(), default=DEFAULT_FORMAT,
                         help='Output format (default: %s)' % DEFAULT_FORMAT)
-    parser.add_argument('-j', '--jobs', type=int, default=1,
-                        help='Number of concurrent subprocesses')
+    parser.add_argument('-j', '--jobs', type=int, default=0,
```
kimgr, line 499: It's a bit tricky, but I'd like to see the absence of `-j` evaluate to jobs == 1, and `-j` without a value evaluate to jobs == 0 (which is further refined to `os.cpu_count()`).

Is `nargs='?'` required to make the integer value optional? I'm not sure what effect `const=0` has here.

I can't think of an argparse flag configuration to allow `() => 1`, `-j => 0`, `-j4 => 4`, so I wonder if it's easier to just remove all the argparse settings and check `args.jobs` using `hasattr`/`getattr`

iwyu_tool_test.py
```diff
@@ -363,6 +363,28 @@ def test_extra_args_with_sep(self):
...
                          self.main.call_args['extra_args'])
 
+    def test_jobs_normal(self):
+        """ Nonzero values for -j are forwarded as they are. """
+        argv = ['iwyu_tool.py', '-p', '.', '-j', '4']
+        iwyu_tool._bootstrap(argv)
+        self.assertEqual(4, self.main.call_args['jobs'])
+
+
+    def test_jobs_auto(self):
```
kimgr, line 373: `test_jobs_zero` seems more apt for this one

iwyu_tool.py
pvutov: `const=0` determines the value if the argument is used as a flag (`-j` case, as opposed to `-jX`). To get what you want, I just need to change to `default=1, const=0`.

kimgr: Ah, nice. I learn something new about argparse every time I use it :upside_down_face:

iwyu_tool_test.py
```diff
@@ -363,6 +363,34 @@ def test_extra_args_with_sep(self):
...
 
+    def test_jobs_normal(self):
+        """ Nonzero values for -j are forwarded as they are. """
+        argv = ['iwyu_tool.py', '-p', '.', '-j', '4']
+        iwyu_tool._bootstrap(argv)
+        self.assertEqual(4, self.main.call_args['jobs'])
+
+
+    def test_jobs_zero(self):
+        """ -j=0 will not hang the program. """
```
kimgr, line 374: Hmm, I don't think this testcase demonstrates `-j=0` doesn't hang, just that zero translates to some arbitrary non-zero?

iwyu_tool_test.py
```diff
@@ -363,6 +363,34 @@ def test_extra_args_with_sep(self):
...
+        argv = ['iwyu_tool.py', '-p', '.', '-j', '0']
+        iwyu_tool._bootstrap(argv)
+        self.assertNotEqual(0, self.main.call_args['jobs'])
+        self.assertNotEqual(None, self.main.call_args['jobs'])
```
kimgr, line 378: Prefer `assertIsNotNone` for None-checking: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsNotNone

iwyu_tool_test.py
```diff
@@ -363,6 +363,34 @@ def test_extra_args_with_sep(self):
...
+
+    def test_jobs_novalue(self):
+        """ It is allowed to use -j without a value """
+        argv = ['iwyu_tool.py', '-j', '-p', '.']
+        iwyu_tool._bootstrap(argv)
+        self.assertNotEqual(None, self.main.call_args['jobs'])
```
kimgr, line 384: `assertIsNotNone`

iwyu_tool_test.py
```diff
@@ -363,6 +363,34 @@ def test_extra_args_with_sep(self):
...
+        argv = ['iwyu_tool.py', '-p', '.', '-j', '4']
+        iwyu_tool._bootstrap(argv)
+        self.assertEqual(4, self.main.call_args['jobs'])
+
+
+    def test_jobs_zero(self):
+        """ -j=0 will not hang the program. """
+        argv = ['iwyu_tool.py', '-p', '.', '-j', '0']
+        iwyu_tool._bootstrap(argv)
+        self.assertNotEqual(0, self.main.call_args['jobs'])
```
kimgr, line 377: Should this be `assertLess` too, like below?

iwyu_tool_test.py
pvutov: I guess I just change the comment?

iwyu_tool.py
```diff
@@ -497,7 +497,10 @@ def custom_help():
                         choices=FORMATTERS.keys(), default=DEFAULT_FORMAT,
                         help='Output format (default: %s)' % DEFAULT_FORMAT)
     parser.add_argument('-j', '--jobs', type=int, default=1,
-                        help='Number of concurrent subprocesses')
+                        nargs='?', const=0,
+                        help='Number of concurrent subprocesses. '
```
kimgr, line 501: This fits on two lines with 80-column width:
```
                        help='Number of concurrent subprocesses. If zero, will '
                             'try to match the logical cores of the system.')

```

iwyu_tool_test.py
kimgr: Yeah, that makes sense. I was trying to come up with a way of testing not hanging, but these tests are heavily mocked, so I don't think that's straightforward.

iwyu_tool.py
kimgr: Oops, lost the leading parenthesis, let's do it as:
```
                        help=('Number of concurrent subprocesses. If zero, '
                              'will try to match the logical cores of the '
                              'system.')
```
to help editors that understand ()-wrapped multiline strings.

kimgr: See the `source` argument for conventional wrapping. The others seem to have snuck through review before.

iwyu_tool.py
```diff
@@ -497,9 +497,13 @@ def custom_help():
                         choices=FORMATTERS.keys(), default=DEFAULT_FORMAT,
                         help='Output format (default: %s)' % DEFAULT_FORMAT)
     parser.add_argument('-j', '--jobs', type=int, default=1,
-                        help='Number of concurrent subprocesses')
+                        nargs='?', const=0,
+                        help=('Number of concurrent subprocesses. If zero, '
+                              'will try to match the logical cores of the '
+                              'system.'))
     parser.add_argument('-l', '--load', type=float, default=0,
-                        help='Do not start new jobs if the 1min load average is greater than the provided value')
```
kimgr, line 502: Ah, sorry, please revert this change, it has nothing to do with the jobs patch.

I'll push a refactor to clean these up after.

kimgr: Oh, I just saw this was in a separate commit. That's fine, let's leave it in.

kimgr:
Thanks, and sorry about all the back and forth!

PR #1560, opened 2024-06-23
Map bits/chrono.h to chrono

correaa:
As discussed in https://github.com/include-what-you-use/include-what-you-use/issues/1552

Not sure if `@` is needed


kimgr:
Squashed and merged, thanks!

Issue #1558, opened 2024-06-19
False positive: IWYU wants to remove `<type_traits>` despite using `std::is_arithmetic_v`


carlosgalvezp:
```cpp
$ cat /tmp/foo.cpp 
#include <type_traits>

template <typename T>
struct Foo
{
  static_assert(std::is_arithmetic_v<T>, "");
};

int main()
{
  Foo<int> f;
}

```

```
$ ./bin/include-what-you-use  /tmp/foo.cpp 

/tmp/foo.cpp should add these lines:

/tmp/foo.cpp should remove these lines:
- #include <type_traits>  // lines 1-1

The full include-list for /tmp/foo.cpp:
---
```

Note: the problem does not happen if I use `std::is_arithmetic<T>::value`. Seems like some adaptations are needed for the C++17 `_v` variables?

bolshakov-a:
I think it is one more `UnresolvedLookupExpr` issue.

kimgr:
Also seems same as #1518?

kimgr:
Fixed on master as of https://github.com/include-what-you-use/include-what-you-use/commit/a82c71187d3e3a8a0d0caac1034138a02090669d

Issue #1516, opened 2024-04-07
std::is_unsigned_v not seen in uninstantiated template


kimgr:
(Broken out from #823)

              I think I've noticed something similar.

For example, given this simple template:

```cpp
// test.cpp
#include <type_traits>
template<typename T>
struct wrapper {
	constexpr static bool b = std::is_unsigned_v<T>;
};
```

IWYU says that `type_traits` should be removed:

```
bc@r2 /tmp % include-what-you-use -std=c++17 test.cpp

test.cpp should add these lines:

test.cpp should remove these lines:
- #include <type_traits>  // lines 2-2

The full include-list for test.cpp:
---
```

Making the suggested change results in a compilation error, though:

```
bc@r2 /tmp % include-what-you-use -std=c++17 test.cpp
test.cpp:4:28: error: use of undeclared identifier 'std'
        constexpr static bool b = std::is_unsigned_v<T>;
                                  ^
test.cpp:4:47: error: 'T' does not refer to a value
        constexpr static bool b = std::is_unsigned_v<T>;
                                                     ^
test.cpp:2:19: note: declared here
template<typename T>
                  ^
test.cpp:4:49: error: expected expression
        constexpr static bool b = std::is_unsigned_v<T>;
```

However, if the template is explicitly instantiated, IWYU no longer complains:

```cpp
// test.cpp
#include <type_traits>
template<typename T>
struct wrapper {
	constexpr static bool b = std::is_unsigned_v<T>;
};
template class wrapper<int>;
```

```
bc@r2 /tmp % include-what-you-use -std=c++17 test.cpp

(test.cpp has correct #includes/fwd-decls)
```

Verbose invocation of the explicit instantiation-less version:

```
bc@r2 /tmp % include-what-you-use -Xiwyu -v7 -std=c++17 test.cpp
Setting verbose-level to 7
Search path: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/ (system)
Search path: /usr/local/Cellar/include-what-you-use/0.15_2/libexec/lib/clang/11.1.0/include/ (system)
Search path: /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/ (system)
Search path: /usr/local/Cellar/include-what-you-use/0.15_2/libexec/include/c++/v1/ (system)
Search path: /Library/Developer/CommandLineTools/usr/include/c++/v1/ (system)
[...]

Issue #1335, opened 2023-11-10
include-what-you-use does not recognize used names in concepts.


Jhuighuy:
Hi!

include-what-use-use does not "see" what was used inside of `concept` declaration.

Example:
```cpp
#pragma once

#include <type_traits>

template<class T>
concept object = std::is_object_v<T>;
```

Output:
```sh
$ include-what-you-use -std=c++20 ./test.hpp

test.hpp should add these lines:

test.hpp should remove these lines:
- #include <type_traits>  // lines 3-3

The full include-list for test.hpp:
---
```

Example (2):
```cpp
#pragma once

#include <type_traits>

template<class T>
  requires std::is_object_v<T>
void f();
```

Output:
```sh
$include-what-you-use -std=c++20 ./test.hpp

test.hpp should add these lines:

test.hpp should remove these lines:
- #include <type_traits>  // lines 3-3

The full include-list for test.hpp:
---
```

kimgr:
I think this TODO indicates that it's not implemented yet: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu.cc#L4145

bolshakov-a:
I think it is due to another reason: variable templates are not supported. `std::is_object<T>::value` works.

kimgr:
Fixed on master as of https://github.com/include-what-you-use/include-what-you-use/commit/a82c71187d3e3a8a0d0caac1034138a02090669d

Issue #966, opened 2021-11-04
iwyu removes include required by variable template specialization 


tberghammer:
```
// a.h
template <typename A, typename B>
struct is_base_of {
    constexpr static bool value = __is_base_of(A, B);
};
template <class Base, class Derived>
inline constexpr bool is_base_of_v = is_base_of<Base, Derived>::value;

// b.cc
#include "a.h"

class Foo {};
template <typename T>
struct Bar {
    static const bool y = is_base_of_v<T, Foo>;
};
bool foo() { return Bar<Foo>::y; }
```

```
$ ./include-what-you-use b.cc -std=c++20
b.cc should add these lines:

b.cc should remove these lines:
- #include "a.h"  // lines 1-1

The full include-list for b.cc:
---
```

iwyu suggests that I remove the include for a.h even though is_base_of_v is only defined in a.h and it is clearly used in b.cc

kimgr:
`-Xiwyu -v7` shows the presence of `UnresolvedLookupExpr`s, which makes me think this is related to #636 and #823

kimgr:
Fixed on master as of https://github.com/include-what-you-use/include-what-you-use/commit/a82c71187d3e3a8a0d0caac1034138a02090669d

PR #1562, opened 2024-07-02
Report unresolved references to variable templates

### LINKED ISSUE #792: IWYU does not recognize function uses inside enable-if's
The tool insists that I remove a type-trait based header because it does not see any use in the code, but in really the header is heavily used in the template guards. My current workaround for this issue is to add a `constexpr` constant after the header inclusion that is left unused. 

`include-what-you-use 0.14 based on clang version 10.0.0`

bolshakov-a:
The change in `badinc.cc` is because `I1_NamespaceTemplateFn` shadow decl is now reported on `CallOverloadWithUsingShadowDecl` scanning, hence using-declaration is considered as being used and may not require `#include` by itself (as for the other using-decls). If its reference is removed from `CallOverloadWithUsingShadowDecl`, the suggestion comes back, as it should be.

This fixes #792, #966, #1335, #1516, and #1558.

### DIFF
--- a/tests/cxx/overload_expr-i4.h
+++ b/tests/cxx/overload_expr-i4.h
@@ -7,7 +7,16 @@
 //
 //===----------------------------------------------------------------------===//
 
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_OVERLOAD_EXPR_I4_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_OVERLOAD_EXPR_I4_H_
+
 // All the OverloadedFn2 overloads visible without ADL are in the same file.
 
 void OverloadedFn2(int);
 void OverloadedFn2(char);
+
+// Variable template.
+template <typename>
+int var_tpl;
+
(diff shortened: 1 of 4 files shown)


tests/cxx/overload_expr-i4.h
```diff
@@ -7,7 +7,16 @@
...
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_OVERLOAD_EXPR_I4_H_
+
 // All the OverloadedFn2 overloads visible without ADL are in the same file.
 
 void OverloadedFn2(int);
 void OverloadedFn2(char);
+
+// Variable template.
+template <typename>
+int var_tpl;
```
kimgr, line 20: I was a bit surprised not to see:
```
template <typename T>
T var_tpl;
```
here, but I guess the result is the same?

kimgr: Yep, I tried that locally, and no change. OK, then I prefer the `int` version, because it's the simplest.

PR #1566, opened 2024-07-09
Reduce 'FILE' mappings to just the canonical header

alejandro-colomar:
<wchar.h> is guaranteed by the relevant standards, but is not the canonical include for FILE.

Link: <https://github.com/include-what-you-use/include-what-you-use/issues/1531>
Link: <https://github.com/include-what-you-use/include-what-you-use/pull/1344>
Link: <https://github.com/include-what-you-use/include-what-you-use/pull/1283>
Reported-by: @firewave
Cc: @kimgr

### DIFF
--- a/gcc.symbols.imp
+++ b/gcc.symbols.imp
@@ -30,7 +30,6 @@
   { "symbol": [ "fenv_t", "private", "<fenv.h>", "public"] },
   { "symbol": [ "fexcept_t", "private", "<fenv.h>", "public"] },
   { "symbol": [ "FILE", "private", "<stdio.h>", "public"] },
-  { "symbol": [ "FILE", "private", "<wchar.h>", "public"] },
   { "symbol": [ "float_t", "private", "<math.h>", "public"] },
   { "symbol": [ "fsblkcnt_t", "private", "<sys/types.h>", "public"] },
   { "symbol": [ "fsfilcnt_t", "private", "<sys/types.h>", "public"] },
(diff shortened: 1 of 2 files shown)


kimgr:
Thanks! I'm going to remove all the links and trailers from the commit message, they don't really add any pertinent information. But otherwise looks great!

alejandro-colomar:
> Thanks! I'm going to remove all the links and trailers from the commit message, they don't really add any pertinent information. But otherwise looks great!

I thought that the links might help point readers to the relevant discussions, because this patch alone might seem weird.  But as you wish.  Maybe in my head it sounded better, and it's just a small detail in the end.  :)

kimgr:
I thought the first paragraph in the commit message made it perfectly clear :-)

alejandro-colomar:
> I thought the first paragraph in the commit message made it perfectly clear :-)

Thanks; I tried!  However, I usually fail at that, which is why I tend to also link to discussions, for full context.  :-)

firewave:
Change looks correct.

References:
https://en.cppreference.com/w/c/io/FILE
https://en.cppreference.com/w/cpp/io/c/FILE

alejandro-colomar:
> Change looks correct.
> 
> References: https://en.cppreference.com/w/c/io/FILE https://en.cppreference.com/w/cpp/io/c/FILE

I wouldn't trust too much that reference.

I recommend having a look at ISO/IEC 9899 (a.k.a., ISO C).  You can find an HTML version here: <http://port70.net/~nsz/c/c11/>.  You'll want to look at clause 7, which is libc, and search for the type definition.

PR #1567, opened 2024-07-09
Reduce 'timespec' mappings to just the canonical header

alejandro-colomar:
<signal.h> is guaranteed by the relevant standards, but is not the canonical include for timespec.


alejandro-colomar:
-  <https://github.com/include-what-you-use/include-what-you-use/pull/1536>

kimgr:
Nice, thanks!

firewave:
Change looks corrected.

References:
https://en.cppreference.com/w/c/chrono/timespec
https://en.cppreference.com/w/cpp/chrono/c/timespec

Issue #1540, opened 2024-05-21
includes for `std::*stream*` are treated as interchangeable


firewave:
```
#include <istream>      // for std::ostream
```

```
#include <sstream>         // for std::basic_ostringstream, std::basic_istringstream, std::istringstream, std::ostringstream, std::ostream
```

```
#include <fstream>                // for std::basic_ostream, std::operator<<, std::basic_ofstream, std::basic_istream, std::basic_ifstream, std::endl, std::ofstream, std::basic_ios, std::ios, std::ostream, std::ifstream, std::operator|, std::ios_base
```

https://en.cppreference.com/w/cpp/io/basic_ostream

Occurs with 0.21 and 0.22.

kimgr:
Just to follow up, it seems like this is intentional:
https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_include_picker.cc#L630

The mappings there are all public-to-public, which I believe means if the symbol is defined in `$from`, but you already have `$to` included, no action will be suggested.

Are you suggesting IWYU should be more rigid here and suggest the individual `<*stream>` headers even if their siblings are already included? I'm inclined to agree, but I'm not sure what practical problems the original mappings were added for.

firewave:
> Are you suggesting IWYU should be more rigid here and suggest the individual `<*stream>` headers even if their siblings are already included?

Yes.

> I'm inclined to agree, but I'm not sure what practical problems the original mappings were added for.

As these mappings already existed over 12 years ago (that's already the age of the code just on the second blame) I assume those were just added because that's how they work in actual code. But I would consider this implementation-specific as it does not reflect what is documented as requirement for these symbols.

This looks quite similar to the public-to-public C mappings which are in the midst of being cleaned up right now.

I also assume this was partially done for convenience to reduce the amount of changes required to appease IWYU. But IMO that is not the goal of this tool. It is supposed to be verbose and granular about these things.

The opposite case would be transitive includes which technically also might not require actions because they "just work". But as highlighted by cleanups in the standard libraries in recent years that will start to come back at you at some point.

I also have the feeling that down the road we might end up at cases where `iosfwd` would be sufficient to use in a header. And these cases would be masked by the mappings in question. But let's take it one step at a time.

firewave:
Even some confusion within the same file:
```
#include <iostream>          // for std::basic_ostream, std::cout, std::endl, std::istringstream
[...]
#include <sstream>           // for std::basic_istringstream
```

[...]

Issue #1438, opened 2024-01-18
ICYU wants regex.h include when not needed


Mango0x45:
I have some code which does regular expression matching using `regex.h`, however you may optionally compile the code with PCRE support via the PCRE2 library. As a result, my includes look like so:

```c
/* libc includes omitted */

#if USE_PCRE
#   include <pcre2posix.h>
#else
#   include <regex.h>
#endif
```

This works fine, my code compiles fine, I have no issues. When I scan my code with IWYU however (passing `-DUSE_PCRE=1`), I get the following:

```
src/grab.c should add these lines:
#include <regex.h>       // for regmatch_t, regex_t
```

This makes little sense to me, as `pcre2posix.h` includes the following:

```c
/* The structure representing a compiled regular expression. It is also used
for passing the pattern end pointer when REG_PEND is set. */

typedef struct {
  void *re_pcre2_code;
  void *re_match_data;
  const char *re_endp;
  size_t re_nsub;
  size_t re_erroffset;
  int re_cflags;
} regex_t;

/* The structure in which a captured offset is returned. */

typedef int regoff_t;

typedef struct {
  regoff_t rm_so;
  regoff_t rm_eo;
} regmatch_t;
```

You can see my copy of the header [here][1].

Any ideas what is causing this or how I can solve it?

[1]: https://paste.thomasvoss.com/421

kimgr:
@Mango0x45 IWYU has built-in symbol mappings for a bunch of known libc types, among many others the regex types: https://github.com/include-what-you-use/include-what-you-use/blob/abc3aee/iwyu_include_picker.cc#L159.

In this case that's clearly problematic... We don't really have any fine-grained control over mappings, the only knob is probably the `-Xiwyu --no_default_mappings` switch, which turns off the mappings entirely. But that will presumably break other necessary mappings. You can add them back group-by-group using `-Xiwyu --mapping_file=gcc.libc.imp` for each of the necessary sets, but I suspect the libc mappings are necessary for most code.

You could probably work around it using some kind of nesting/indirection, e.g.:
```
// regex_wrap.h
#ifdef USE_PCRE
#include <...>  // IWYU pragma: keep
typedef regex_t my_regex_t;
typedef regmatch_t my_regmatch_t;
#else
// same for libc <regex.h>
#endif
```

... but that's clearly not very ergonomic.

alejandro-colomar:
Hmmm, why didn't PCRE define their own type with a different name?

[...]

PR #1568, opened 2024-07-09
Reduce 'wchar_t' mappings to just the canonical header

alejandro-colomar:
<stdlib.h> is guaranteed by the relevant standards, but is not the canonical include for wchar_t.

### DIFF
--- a/gcc.symbols.imp
+++ b/gcc.symbols.imp
@@ -94,7 +94,6 @@
   { "symbol": [ "uid_t", "private", "<sys/types.h>", "public"] },
   { "symbol": [ "useconds_t", "private", "<sys/types.h>", "public"] },
   { "symbol": [ "wchar_t", "private", "<stddef.h>", "public"] },
-  { "symbol": [ "wchar_t", "private", "<stdlib.h>", "public"] },
   { "symbol": [ "wint_t", "private", "<wchar.h>", "public"] },
   { "symbol": [ "sig_atomic_t", "private", "<signal.h>", "public"] },
   { "symbol": [ "size_t", "private", "<stddef.h>", "public"] },
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -178,7 +178,6 @@ const IncludeMapEntry libc_symbol_map[] = {
   { "uid_t", kPrivate, "<sys/types.h>", kPublic },
   { "useconds_t", kPrivate, "<sys/types.h>", kPublic },
   { "wchar_t", kPrivate, "<stddef.h>", kPublic },
-  { "wchar_t", kPrivate, "<stdlib.h>", kPublic },
   { "wint_t", kPrivate, "<wchar.h>", kPublic },
   // It is unspecified if the cname headers provide ::size_t.
   // <locale.h> is the one header which defines NULL but not size_t.


firewave:
It would be good to add references to these changes.

I think this change might not be correct. According to `Types` https://en.cppreference.com/w/c/string/wide this type is provided by three different headers:
```
Defined in header <stddef.h>
Defined in header <stdlib.h>
Defined in header <wchar.h>
wchar_t
	integer type that can hold any valid wide character
(typedef)
```

alejandro-colomar:
> It would be good to add references to these changes.
> 
> I think this change might not be correct. According to `Types` on https://en.cppreference.com/w/c/string/wide this type is provided by three different headers:
> 
> ```
> Defined in header <stddef.h>
> Defined in header <stdlib.h>
> Defined in header <wchar.h>
> wchar_t
> 	integer type that can hold any valid wide character
> (typedef)
> ```

ISO C describes `wchar_t` in <http://port70.net/~nsz/c/c11/n1570.html#7.19> (`<stddef.h>`), and while it says the type is also provided by other headers, those refer to `<stddef.h>` for the description of the type.  Just grep for `wchar_t` in the ISO C standard, and you'll find plenty of references to `<stddef.h>` as the canonical header that defines the type.

The fact that other headers also provide the type (by inclusion), doesn't matter much.  It's like `size_t` being also defined in `<monetary.h>`; we don't care too much about it.

alejandro-colomar:
> I think this change might not be correct. According to `Types` on https://en.cppreference.com/w/c/string/wide this type is provided by three different headers:
> 
> ```
> Defined in header <stddef.h>
> Defined in header <stdlib.h>
> Defined in header <wchar.h>
> wchar_t
> 	integer type that can hold any valid wide character
> (typedef)
> ```

BTW, they forgot about inttypes.h, which (at least according to POSIX; maybe ISO C doesn't) also defines the type.

And ironically enough, glibc seems to provide a definition in `<inttypes.h>` , and not on the others (except `<stddef.h>`, which is a compiler-magic header).

firewave:
https://man7.org/linux/man-pages/man3/wchar_t.3type.html only shows `<stddef.h>`.

So I guess this is right.

alejandro-colomar:
> https://man7.org/linux/man-pages/man3/wchar_t.3type.html only shows `<stddef.h>`.
> 
> So I guess this is right.

It actually shows more headers, in NOTES.

See: <https://man7.org/linux/man-pages/man3/wchar_t.3type.html#NOTES>

firewave:
> It actually shows more headers, in NOTES.
> 
> See: https://man7.org/linux/man-pages/man3/wchar_t.3type.html#NOTES

Ugh. In this case they are explicit about it so I am not sure if not treating these as providers is actually okay. In other cases they used "may provide".

alejandro-colomar:
> > It actually shows more headers, in NOTES.
> > See: https://man7.org/linux/man-pages/man3/wchar_t.3type.html#NOTES
> 
> Ugh. In this case they are explicit about it so I am not sure if not treating these as providers is actually okay. In other cases they used "may provide".

In the man pages, all of them are "provides".  POSIX.1-2008 requires that the types are provided.

I didn't find any case in the man pages where it says "may provide" (and I don't remember having written that in any page).  I've only documented the headers that are required to provide a type by POSIX.1-2008 (or ISO C11), so any documented headers are required providers.

However, as the counterargument for `"so I am not sure if not treating these as providers is actually okay"`, I'd say that POSIX.1-2008 **requires** that `<monetary.h>` provides `size_t`.  Yet, we don't want to allow users to include `<monetary.h>` just because it's a valid provider of the type (usually, that'll be via inclusion of `<stddef.h>`).

alejandro-colomar:
> > However, as the counterargument for `"so I am not sure if not treating these as providers is actually okay"`, I'd say that POSIX.1-2008 **requires** that `<monetary.h>` provides `size_t`. Yet, we don't want to allow users to include `<monetary.h>` just because it's a valid provider of the type (usually, that'll be via inclusion of `<stddef.h>`).
> 
> There is a difference between "commonly accepted" providers

We should then draw a line of what's good taste.  I'm not sure I could draw a good line.  To me, good taste is being super pedantic.  I understand that's not everybody's taste, though.

> because that's how it works in real code (which is what you are trying to address with these changes) and documented providers.

I think real code depends on the authors of such code.  If authors are using IWYU, they are probably more pedantic than average, and don't mind us being a little bit more pedantic.  They probably don't want us to be suddenly pedantic if we weren't before, but if we do it gradually, it can make sense, I think.

> Users will baffled if they need to change their code although it adheres to what is correct according to the documentation.

You might be baffled to learn that `<monetary.h>` is documented as a provider of `size_t`.  Check the manual page: <https://www.man7.org/linux/man-pages/man3/ssize_t.3type.html#NOTES>.

> 
> I am all in favor of being pedantic but often such warnings usually cover things you might not be aware of or is not trivially to look up.

That's why I first wrote manual pages for all of these types (I hope I didn't forget any).  If a type is in an IWYU mapping, there's probably a manual page for it.  Just `man wchar_t` (or more explicitly, `man 3type wchar_t`) or whatever type you want, and you'll find the canonical header in the SYNOPSIS, plus all other headers that ISO C and POSIX.1 require to be providers below, usually in NOTES.

> So I am a bit split on these as they are not there for convenience to lower the required changes but because they _seem_ canonical. If that should not be canon it might be worth to push changes upstream because essentially IWYU and clang-include-cleaner are the de-facto standard tools to push the canon and might imply that they have more knowledge about the topic and are allowed to make such judgements.

I wrote those mappings, and I claim that I'm not good for that kind of judgements.  Or if you want a judgement from me, it'd be the super pedantic one.  :-)

> 
> I hope I have not missed my point (did I even have one?) here...

Understood.  :)

firewave:
> I think real code depends on the authors of such code. If authors are using IWYU, they are probably more pedantic than average, and don't mind us being a little bit more pedantic. They probably don't want us to be suddenly pedantic if we weren't before, but if we do it gradually, it can make sense, I think.

With "real code" I refer to the "problem" of transitive includes. If you have a standards implementation where e.g. `<string>` contains *every* other header, even if it doesn't require it, you might get away with only a single system include in your whole code base. I do not consider that a taste and the line you need to draw is quite obvious. That's why adding public-to-public mappings from `<string>` to everything makes no sense even if it makes things convenient for the user (see also #1540).

The two big implementations of the STL have been cleaning up includes recently so relying on transitive includes is no longer feasible and you need to be "pedantic" to make sure your code is future-proof.

Also weighting some community sourced documentation (which is possibly just based on a more reasonable observation of the "real code") against the actual canon of a specification is also not a taste. So that is also a line you should draw. But these things have to be aligned in the common documentation and implementations so they do not contradict each other. That is probably the main issue.

> I wrote those mappings, and I claim that I'm not good for that kind of judgements. Or if you want a judgement from me, it'd be the super pedantic one. :-)

Well, someone has to start and if it makes sense and the community isn't too toxic towards it, it should prevail and grow.

alejandro-colomar:
> > I think real code depends on the authors of such code. If authors are using IWYU, they are probably more pedantic than average, and don't mind us being a little bit more pedantic. They probably don't want us to be suddenly pedantic if we weren't before, but if we do it gradually, it can make sense, I think.
> 
> With "real code" I refer to the "problem" of transitive includes. If you have a standards implementation where e.g. `<string>` contains _every_ other header, even if it doesn't require it, you might get away with only a single system include in your whole code base. I do not consider that a taste and the line you need to draw is quite obvious. That's why adding public-to-public mappings from `<string>` to everything makes no sense even if it makes things convenient for the user (see also #1540).

Yup, we agree that would fall outside of the line, whatever the line is.

> The two big implementations of the STL have been cleaning up includes recently so relying on transitive includes is no longer feasible and you need to be "pedantic" to make sure your code is future-proof.

Hmm, interesting.

> Also weighting some community sourced documentation (which is possibly just based on a more reasonable observation of the "real code") against the actual canon of a specification is also not a taste. So that is also a line you should draw. But these things have to be aligned in the common documentation and implementations so they do not contradict each other. That is probably the main issue.

Then, since the main documentation for C is the man pages (at least on Unix-like systems), and the man pages now describe the canonical header for a type, I guess we can require users to follow that.

The line in the manual pages is already drawn, and is quite clear: Include the canonical header, which is specified in the SYNOPSIS, whenever you want to declare a variable of that type.  Let's use that line.  :)

Back before I wrote those pages, people had the excuse that it wasn't clearly documented (<https://stackoverflow.com/questions/55190317/where-is-ssize-t-defined-in-linux>).  Now there's no excuse anymore.

firewave:
> > The two big implementations of the STL have been cleaning up includes recently so relying on transitive includes is no longer feasible and you need to be "pedantic" to make sure your code is future-proof.
> 
> Hmm, interesting.

`libc++` even has infrastructure in place to track/control the transitivity. They also use more granularized versions of the headers internally.

Some of the more recent changes are these: https://github.com/llvm/llvm-project/pulls?q=is%3Apr+author%3Aphilnik777.

> Back before I wrote those pages, people had the excuse that it wasn't clearly documented (https://stackoverflow.com/questions/55190317/where-is-ssize-t-defined-in-linux). Now there's no excuse anymore.

There is still the excuse that the common (probably mostly observational and not canonical) documentation says so.

alejandro-colomar:
> > Back before I wrote those pages, people had the excuse that it wasn't clearly documented (https://stackoverflow.com/questions/55190317/where-is-ssize-t-defined-in-linux). Now there's no excuse anymore.
> 
> There is still the excuse that the common (probably mostly observational and not canonical) documentation says so.

Let's spread the word that manual pages document types in section `3type`, and shape a better world.  :-)

firewave:
Maybe we should document the goals of using canonical includes and avoiding the reliance on transitive includes somewhere.

Could be either in an initial "mission statement or/and in the FAQ. Questions could be
- "My code compiles just fine. Why does IWYU suggest to add so many additional headers?" - outlining the transitive include case
- "According to x symbol y is provided by z. Why is IWYU attributing it to a different include?" - outlining the canonical case with a side note that it might be a false positive in cases where it is very off

alejandro-colomar:
> Maybe we should document the goals of using canonical includes and avoiding the reliance on transitive includes somewhere.

I think that's the entire purpose of this project.  I kind of assumed it was obvious, and implied by the existing docs.  :-)

> 
> Could be either in an initial "mission statement or/and in the FAQ. Questions could be
> 
>     * "My code compiles just fine. Why does IWYU suggest to add so many additional headers?" - outlining the transitive include case
> 
>     * "According to x symbol y is provided by z. Why is IWYU attributing it to a different include?" - outlining the canonical case with a side note that it might be a false positive in cases where it is very off

I think I would just wait for complaints, and then point to the manual page if someone asks.

To minimize complaints, we need to do it very gradually, though.  I think I'll skip from this release those changes that require important changes to the tests.

kimgr:
This discussion got me thinking about a special case that made me a little annoyed with symbol mappings: https://github.com/include-what-you-use/include-what-you-use/issues/1438.

For highly portable codebases, it's not unusual with conditional tricks like this (though I think the canonical way might be a library prefix, not shadowing standard library types), so it's annoying that IWYU just decides in its final analysis steps that "oh, this thing named X must come from header x.h, no matter what".

This doesn't have anything to do with this `wchar_t` patch specifically, I just wanted to highlight it as one more thing to worry about when messing about with symbol mappings.

alejandro-colomar:
Let's say PCRE2 is doomed to not be supported.  It has broken the sacred rules.  Thou shalt not name two different things with the same name.  Even more if the original one is a POSIX.1 standard API.

firewave:
I think this can be merged, right?

kimgr:
Yeah, it looks good to me, I just wanted to wait until you both reached agreement :)

Issue #1142, opened 2022-11-20
inconsistent source of `size_t`


firewave:
`test.cpp`
```cpp
#include "test.h"

#include <cstring>

static void f()
{
    size_t pos;
    std::strchr((const char*)NULL, 'i');
}

```

`test.h`
```cpp
#ifndef testH
#define testH

#include <cstddef>
#include <cstring>

void flags() {
    std::size_t pos;
    std::strchr((const char*)NULL, 'i');
}

#endif
```

```
test.h should add these lines:

test.h should remove these lines:
- #include <cstddef>  // lines 4-4

The full include-list for test.h:
#include <cstring>  // for strchr, NULL, size_t
---

test.cpp should add these lines:
#include <cstddef>  // for NULL, size_t

test.cpp should remove these lines:

The full include-list for test.cpp:
#include "test.h"
#include <cstddef>  // for NULL, size_t
#include <cstring>  // for strchr
---
```

kimgr:
Curious. See if you can get more information about the decision process with more verbose logging.

firewave:
FYI I am using `include-what-you-use 0.18 based on Debian clang version 14.0.6-2`.

kimgr:
I get the same results on current mainline (0.19+)

firewave:
I forgot a `std::` in the .cpp but that doesn't make a difference.

alejandro-colomar:
I've tried some small changes to remove some diagnostics (unused variable and nonnull misuse).  I've also removed some more `std::` prefixes to see if it changed anything.

```c++
alx@debian:~/tmp/iwyu$ cat test.cpp 
#include "test.h"

#include <cstring>

static size_t
f(char *s)
{
    return strchr(s, 'i') - s;
}
```
```c++
alx@debian:~/tmp/iwyu$ cat test.h 
#ifndef testH
#define testH

#include <cstddef>
#include <cstring>

size_t
flags(char *s)
{
    return std::strchr(s, 'i') - s;
}

#endif
```
```c++
alx@debian:~/tmp/iwyu$ iwyu test.cpp 

test.h should add these lines:

test.h should remove these lines:
- #include <cstddef>  // lines 4-4

The full include-list for test.h:
#include <cstring>  // for strchr, size_t
---

test.cpp should add these lines:
#include <cstddef>  // for size_t

test.cpp should remove these lines:

The full include-list for test.cpp:
#include "test.h"
#include <cstddef>  // for size_t
[...]

Issue #227, opened 2015-10-20
Needless boost header changes recommended


pnorman:
This **might** be a case of #226 exhibiting a different way

```
node-ram-cache.cpp should add these lines:
#include "boost/format/format_class.hpp"  // for basic_format
#include "boost/format/format_fwd.hpp"  // for format
#include "boost/format/format_implementation.hpp"

node-ram-cache.cpp should remove these lines:
- #include <boost/format.hpp>  // lines 14-14
```

See [boost/format.hpp](http://www.boost.org/doc/libs/1_55_0/boost/format.hpp)

vsapsai:
Are you using [boost-all.imp](https://github.com/include-what-you-use/include-what-you-use/blob/master/boost-all.imp)? Because it has a mapping
```
{ include: ["@<boost/format/.*>", private, "<boost/format.hpp>", public ] },
```

firewave:
It has been a few years since I regularly used Boost but the following should be apply.

Although the Boost project tries to streamline things the way libraries are structed might differ. Some might be header-only and others might (optionally) allow for parts to be compiled. That will also impact the way includes are being handled.

Some of the libraries just provide very basic functionality and thus there are less dependencies also leading to smaller amount of includes. With increased functionality the amount of dependencies al increases. As it tries to re-use code that means pulling in more dependencies.

Most of the libraries provide an umbrella header which pulls in all the functionality. 
In some cases that is done for convenience - e.g. Boost.Format.
In other cases it is obviously because it is likely impossible for a person to figure out what includes to actually provide (albeit a tool like IWYU might) - e.g. BEAST or Boost.Test.
And there are also cases where these include contain header from a different internal (private) structure which should not be used by the user code so the code will still compile in the future - e.g. Boost.Regex .
Overall this also allows you to target different versions of Boost without requiring you to adjust headers.

The umbrella headers should be fine for most people.
But if you are passing Boost objects across your code base you probably do not want to spill all the Boost headers via your own header so that would obviously wreck havoc and you might possibly end up with only a single Boost include and nothing else in parts of your code as everything else is being pulled in by it.

So IMO there are the following approaches:
- map private-to-public headers (should always be done)
- map public-to-public headers for the commonly suggested umbrella headers (optionally - should be the default though)
- allow for more granular includes instead of using umbrella headers (basically the current default if you have no mappings but should be optionally)
  - limit the granular approach to headers only

PR #1572, opened 2024-07-21
Reduce 'std::size_t' mappings to just the canonical header

alejandro-colomar:
Let's see if this passes the tests.


alejandro-colomar:
It only required minor changes to the tests.  :)

tests/cxx/std_size_t.cc
```diff
@@ -7,33 +7,22 @@
 //
 //===----------------------------------------------------------------------===//
 
-// This tests a little quirk with the mappings for the new <cstd...> headers.
```
kimgr, line 10: I think it's actually better to remove this testcase.

It was a little confusing; I don't think its purpose is to test anything around `size_t`, but rather to demonstrate that mappings depend on include order and use order, with `size_t` as a common example.

Now that `size_t` is more unambiguous, let's get rid of it (I tried to rephrase it in terms of `NULL`/`printf`, but it's not really adding any relevant coverage, IMO)

gcc.symbols.imp
```diff
@@ -176,15 +176,6 @@
   { "symbol": [ "std::char_traits", "private", "<ostream>", "public"] },
   { "symbol": [ "std::char_traits", "private", "<istream>", "public"] },
 
-  # std::ptrdiff_t is often an architecture specific definition, force the
```
kimgr, line 179: Please leave this comment, here and in `iwyu_include_picker.cc`

kimgr:
Please squash into a single commit, so tests are passing at every point through commit history. Thanks!

tests/cxx/std_size_t.cc
alejandro-colomar: I originally removed it, but then I thought that the file name of the test makes it appropriate for testing that std::size_t  is mapped to <cstddef>.  We don't have any tests for that, do we?

gcc.symbols.imp
alejandro-colomar: Ok.

tests/cxx/std_size_t.cc
kimgr: We typically don't test mappings, because they're inherently platform specific.

alejandro-colomar: > We typically don't test mappings, because they're inherently platform specific.

Makes sense.

BTW, I hadn't found the wording to respond to your request that I keep the ptrdiff comment, but now I have.

> # std::ptrdiff_t is often an architecture specific definition, force the

Typedefs are inherently platform-specific definitions.  If they weren't, they'd use plain `long` (or `int`).  What does that comment add?

kimgr: In the context of IWYU, it implies that `ptrdiff_t` is defined in a private header, possibly multiple different headers. But yeah, I guess that's true for most mapped symbols.

I just don't want to mix that comment removal up with the `size_t` changes, so please just leave it alone for now.

alejandro-colomar: I've moved the comment removal to a separate commit.

alejandro-colomar: And also removed the test case.

gcc.symbols.imp
alejandro-colomar: I've moved it to a separate commit.  Is that ok?

alejandro-colomar:
Squashed.

bolshakov-a:
I agree with @firewave that such changes are doubtful. The C++ standard clearly states that `std::size_t` is provided by more than one header. The situation is the same as if someone writes in its own custom library:
```cpp
// my_monetary.h
#include "my_stddef.h" // IWYU pragma: export
```
If some file using the library doesn't use any other stuff from `my_monetary.h`, it should be better replaced with `my_stddef.h`, I agree. But IWYU doesn't currently do so. I think that some code should be written in `iwyu_output.cpp` to prefer canonical headers in such cases instead.

alejandro-colomar:
> I agree with @firewave that such changes are doubtful. The C++ standard clearly states that `std::size_t` is provided by more than one header. The situation is the same as if someone writes in its own custom library:
> 
> ```c++
> // my_monetary.h
> #include "my_stddef.h" // IWYU pragma: export
> ```
> 
> If some file using the library doesn't use any other stuff from `my_monetary.h`, it should be better replaced with `my_stddef.h`, I agree. But IWYU doesn't currently do so. I think that some code should be written in `iwyu_output.cpp` to prefer canonical headers in such cases instead.

That would make the code in IWYU unnecessarily complex.

And today there's no reason to not include <stddef.h>.  It's a compiler-magic header that is actually injected by the compiler, and need not read any actual file.  Performance is good.  And one or two more lines of includes shouldn't hurt too much.



Are there any good reasons to not include

bolshakov-a:
I suspect that there will be lots of complaints from users. I faced with them on my previous work even when `<cstddef>` was actually needed according to the IWYU principle.

alejandro-colomar:
> I suspect that there will be lots of complaints from users. I faced with them on my previous work even when `<cstddef>` was actually needed according to the IWYU principle.

TBH, I think this change also follows the IWYU principle.  Those IWYU pragmas that you showed are more for things like using internal `<bits/...>` headers, but not for providing the same type in more than one header.  That usage is for silencing non-IWYU code warnings.

But yeah, I've split the change across several IWYU versions to make it gradual, so that users can adapt, and complain less.

bolshakov-a:
> Those IWYU pragmas that you showed are more for things like using internal <bits/...> headers ...

A useful case of the public-to-public mapping is e.g. headers with only forward declarations, when someone writes in a full-declaration header `helpers.h`:
```cpp
#include "helpers-fwd.h" // IWYU pragma: export
```
so that IWYU cannot suggest to include `helpers.h` and `helpers-fwd.h` simultaneously, which would be rather pointless. Currently, it can result in a problem similar to that you attempt to solve: after some refactoring, the full definitions may not be needed anymore, but IWYU will still accept `#include "helpers.h"` instead of replacing it with `"helpers-fwd.h"`. I suspect that if you fix this problem, the issue with canonical standard library headers would be solved automatically.

Maybe, it should be noted that including `"...-fwd.h"` header in the implementation header may be inevitable at least for class templates with default template arguments.

kimgr:
@bolshakov-a

> The C++ standard clearly states that std::size_t is provided by more than one header.

Do you have a link/reference handy? I went looking and couldn't find the right paragraph.

kimgr:
@alejandro-colomar 
> And today there's no reason to not include <stddef.h>. It's a compiler-magic header that is actually injected by the compiler, and need not read any actual file.

Are you sure? Can you show some evidence? It sounds too good to be true to me :)

kimgr:
I'm on the fence about this after hearing the feedback.

I'm thinking our job is not to be dogmatic, but to make IWYU (which is a pretty difficult problem) do good work without getting in our users' way. I personally like these "canonical header" changes, because it makes IWYU less ambiguous and its behavior easier to reason about, but I don't want to sacrifice user experience.

I guess the question is how much user experience will suffer. Can we produce some kind of metrics? I don't work on real-life C++ code anymore, so all I have to try on is IWYU itself, but maybe somebody else has some significant code base they could compare before/after on?

alejandro-colomar:
> @alejandro-colomar
> 
> > And today there's no reason to not include <stddef.h>. It's a compiler-magic header that is actually injected by the compiler, and need not read any actual file.
> 
> Are you sure? Can you show some evidence? It sounds too good to be true to me :)

```sh
alx@debian:~$ find /usr/include/ 2>/dev/null | grep stddef.h
/usr/include/linux/stddef.h
```

Hmmm, actually the header is here:

```
alx@debian:~$ find /usr/lib/gcc/ 2>/dev/null | grep stddef.h
/usr/lib/gcc/x86_64-linux-gnu/14/include/stddef.h
/usr/lib/gcc/x86_64-linux-gnu/13/include/stddef.h
```

I had heard that a long time ago and just repeated it. :(
But it seems there's an actual header file hidden in a secret gcc location.

alejandro-colomar:
> I'm on the fence about this after hearing the feedback.
> 
> I'm thinking our job is not to be dogmatic, but to make IWYU (which is a pretty difficult problem) do good work without getting in our users' way. I personally like these "canonical header" changes, because it makes IWYU less ambiguous and its behavior easier to reason about, but I don't want to sacrifice user experience.
> 
> I guess the question is how much user experience will suffer. Can we produce some kind of metrics? I don't work on real-life C++ code anymore, so all I have to try on is IWYU itself, but maybe somebody else has some significant code base they could compare before/after on?

I updated neomutt(1) recently, and the change was:

```sh
$ git show 3d865eb7d7e57203cb8f8a803084259b6d465683 --stat | tail -n1
 75 files changed, 77 insertions(+), 29 deletions(-)
```

That change was easy to do.

I could try building neomutt(1) with the size_t changes, and see how much I need to change.  I expect the std::size_t changes to be similar.

alejandro-colomar:
> @bolshakov-a
> 
> > The C++ standard clearly states that std::size_t is provided by more than one header.

I don't know the ISO C++ standard, but the ISO C one _declares_ the size_t type in several headers, as **described in <stddef.h>**:

<http://port70.net/~nsz/c/c11/n1570.html#7.21.1p2>

> 
> Do you have a link/reference handy? I went looking and couldn't find the right paragraph.

bolshakov-a:
> Do you have a link/reference handy? I went looking and couldn't find the right paragraph.

http://eel.is/c++draft/cstdlib.syn
http://eel.is/c++draft/cstring.syn
http://eel.is/c++draft/cwchar.syn

The remainder can be easily found by opening [the PDF version](https://timsong-cpp.github.io/cppwp/n4950/draft.pdf) and searching for `using size_t =`

It may be worth noting that type aliases can not be forward-declared, hence their declarations should be definitions as well.

alejandro-colomar:
> > I'm on the fence about this after hearing the feedback.
> > I'm thinking our job is not to be dogmatic, but to make IWYU (which is a pretty difficult problem) do good work without getting in our users' way. I personally like these "canonical header" changes, because it makes IWYU less ambiguous and its behavior easier to reason about, but I don't want to sacrifice user experience.
> > I guess the question is how much user experience will suffer. Can we produce some kind of metrics? I don't work on real-life C++ code anymore, so all I have to try on is IWYU itself, but maybe somebody else has some significant code base they could compare before/after on?
> 
> I updated neomutt(1) recently, and the change was:
> 
> ```shell
> $ git show 3d865eb7d7e57203cb8f8a803084259b6d465683 --stat | tail -n1
>  75 files changed, 77 insertions(+), 29 deletions(-)
> ```
> 
> That change was easy to do.
> 
> I could try building neomutt(1) with the size_t changes, and see how much I need to change. I expect the std::size_t changes to be similar.

I can't install all deps; I'll try again when libclang-19-dev is fixed in Debian

```
$ sudo apt-get install libclang-19-dev
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 libclang-19-dev : Depends: libclang1-19 (= 1:19.1.0~++20240724103243+7af27be6633a-1~exp1) but 1:19~++20240125092523+41fe98a6e7e5-1~exp1 is to be installed
                   Depends: libclang-common-19-dev (= 1:19.1.0~++20240724103243+7af27be6633a-1~exp1) but 1:19~++20240125092523+41fe98a6e7e5-1~exp1 is to be installed
E: Unable to correct problems, you have held broken packages.
```

firewave:
The apt.llvm.org issues were fixed two days ago. So it should be working again.

alejandro-colomar:
> The apt.llvm.org issues were fixed two days ago. So it should be working again.

The Debian package isn't fixed yet, it seems.

firewave:
> The Debian package isn't fixed yet, it seems.

You might need to re-run the script.

firewave:
> I guess the packagers are busy with something else.

Ubuntu was fixed two days ago. 

Please file an issue at https://github.com/llvm/llvm-project/issues. That is used to track issues with the packages.

alejandro-colomar:
> > I guess the packagers are busy with something else.
> 
> Ubuntu was fixed two days ago.
> 
> Please file an issue at https://github.com/llvm/llvm-project/issues. That is used to track issues with the packages.

Hmmm, I've been able to install it from experimental; I guess the package wasn't in unstable yet.

```
$ sudo apt-get install -t experimental libclang-19-dev
[sudo] password for alx: 
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following packages were automatically installed and are no longer required:
  libavfilter9 libplacebo338
Use 'sudo apt autoremove' to remove them.
The following additional packages will be installed:
  clang-19 libclang-common-19-dev libclang-cpp19 libclang-rt-19-dev
  libclang1-19 libllvm19 llvm-19 llvm-19-dev llvm-19-linker-tools
  llvm-19-runtime llvm-19-tools
The following NEW packages will be installed:
  libclang-19-dev
The following packages will be upgraded:
  clang-19 libclang-common-19-dev libclang-cpp19 libclang-rt-19-dev
  libclang1-19 libllvm19 llvm-19 llvm-19-dev llvm-19-linker-tools
  llvm-19-runtime llvm-19-tools
11 upgraded, 1 newly installed, 0 to remove and 254 not upgraded.
Need to get 141 MB of archives.
After this operation, 295 MB of additional disk space will be used.
Do you want to continue? [Y/n] y
```

I'll try to build IWYU now.

firewave:
> Hmmm, I've been able to install it from experimental; I guess the fix wasn't in unstable yet.

Oh sorry. You are not using `apt.llvm.org` so nothing I said applies to you...

alejandro-colomar:
After **all** of my planned changes, that is, after applying <https://github.com/include-what-you-use/include-what-you-use/pull/1536>, I get the following changes in neomutt(1):

```
$ git diff -U0 | grep '^[-+]#' | sort | uniq -c
    199 +#include <stddef.h>
      1 +#include <stdio.h>
      1 +#include <time.h>
      1 -#include <signal.h>
     44 -#include <stdio.h>
      9 -#include <stdlib.h>
      8 -#include <string.h>
      1 -#include <wchar.h>
```

This includes most notably, the mappings for `size_t` and `NULL`.  It requires adding a couple of hundred includes of `<stddef.h>`, which is a lot, but doable.  On the other side, it allowed discovering many includes of `<stdio.h>`, and some of `<stdlib.h>`, and `<string.h>` that were included only for size_t or NULL.

I'd say this change is good, and should be acceptable, by projects which believe in IWYU guidelines.

firewave:
> (Still, splitting those changes over several iwyu(1) versions will probably be smoother.)

If there are multiple includes which need to be phased out for a symbol maybe we could limit the amount of mappings which are removed for a single symbol at a time. That would stretch things out longer but lessen the changes you need to do at once.

Or make the complete removal of all non-canonical mapping opt-in with the default being the above suggested.

alejandro-colomar:
> > (Still, splitting those changes over several iwyu(1) versions will probably be smoother.)
> 
> If there are multiple includes which need to be phased out for a symbol maybe we could limit the amount of mappings which are removed for a single symbol at a time. That would stretch things out longer but lessen the changes you need to do at once.

Hmmm, removing some headers at a time of all mappings instead of removing all headers of a single mapping could be another way.  @kimgr , what do you think?

> 
> Or make the complete removal of all non-canonical mapping opt-in with the default being the above suggested.

I'll let @kimgr decide.

kimgr:
@alejandro-colomar It's hard to weigh the neomutt numbers without knowing how big that code base is. I feel like it needs to be some kind of ratio across several projects, i.e. before and after number of changes/lines of code or something. But even then I'm not convinced that it's a very useful metric.

The diffstats you posted are interesting, though. 199 "redundant" `<stddef.h>` includes added is quite a lot. But 44 removed `<stdio.h>` includes (presumably replaced by `<stddef.h>`) seems like a significant win. So it seems like the more unambiguous mapping(s) makes it easier for IWYU to make decisions at the cost of some verbosity.

kimgr:
I really don't have a clear view of the impact. If @bolshakov-a has heard explicit complaints about extra `stddef.h` includes, then maybe that's the only signal we have. I'd be happy to move forward with this in the interest of reducing the mappings to a minimum, but not at any cost.

bolshakov-a:
I'd say that @alejandro-colomar questions the public-to-public mapping concept which IWYU supports currently. A similar issue is that many libraries provide an "umbrella" header (like `boost/variant.hpp`, e.g.). Including it instead of some specific public header is redundant in many cases, but IWYU allows (and should allow, I think) to do so.

alejandro-colomar:
> I'd say that @alejandro-colomar questions the public-to-public mapping concept which IWYU supports currently.

Yup; I question that.

> A similar issue is that many libraries provide an "umbrella" header (like `boost/variant.hpp`, e.g.). Including it instead of some specific public header is redundant in many cases, but IWYU allows (and should allow, I think) to do so.

It allows (and should allow) that, but IMO that's for compatibility with non-IWYU-friendly projects.

A library is free to claim that some header is private and should not be included, such as glibc's <bits/...> headers.  But if a header is public, I doubt the value of providing a catch-all (or even worse, a catch-many) header.  It's precisely the opposite of IWYU guidelines.

alejandro-colomar:
> @alejandro-colomar It's hard to weigh the neomutt numbers without knowing how big that code base is. I feel like it needs to be some kind of ratio across several projects, i.e. before and after number of changes/lines of code or something. But even then I'm not convinced that it's a very useful metric.
> 
> The diffstats you posted are interesting, though. 199 "redundant" `<stddef.h>` includes added is quite a lot. But 44 removed `<stdio.h>` includes (presumably replaced by `<stddef.h>`) seems like a significant win. So it seems like the more unambiguous mapping(s) makes it easier for IWYU to make decisions at the cost of some verbosity.

According to sloccount(1), there are around 190k lines of C code.

But yeah, I think the ratio of added stddef vs removed stdio is more important.  It tells us that in a large amount (~ 1/4) of cases, a bogus header was being used, and this helped fix that.

firewave:
I am taking quite a big swing here. So sorry if I go too of the path but maybe some of these points could be split into separate tickets.

> A similar issue is that many libraries provide an "umbrella" header (like `boost/variant.hpp`, e.g.). Including it instead of some specific public header is redundant in many cases, but IWYU allows (and should allow, I think) to do so.

The umbrella concept is a bit different though and Boost has reasons for it but is not consistent on that - see https://github.com/include-what-you-use/include-what-you-use/issues/227#issuecomment-2213243993 for a write-up on that.

I think the stream headers case (see #1540) is closer to what we are dealing with here.

> According to sloccount(1), there are around 190k lines of C code in neomutt(1).

LOC might not be the metric to look at but rather files in the project. And that probably needs to be differentiated between source files, private headers and public headers.

If the changes are just in source files than they could be categorized to "just" appease the IWYU tool (there's still parse/compilation time but lets leave that out for now). If they are in private headers they might actually clean up things. Public headers have an unknown impact as it will affect the API which will ripple effects on user/implementation code. These might be changes you could only make in a certain versions to avoid build breakage.

Maybe there would be even a different metric in how many files are affected by each change.

Transitive includes is another case to be made. Current or common implementations might provide certain symbols through non-canonical header but that might not apply to any of them. Or it might change in the future (but possibly might not because of peer pressure against the ripple effect). I mentioned that LLVM tries to minimize such includes in their STL/C implementation.

There was also a really eye-opening issue in the LLVM project by a game dev giving an interesting look into how they operate and the most shocking thing was that they are *not* using any existing STL since the overhead is way to big. They should a custom light layer to compile against and that apparently is standard in the games industry.

bolshakov-a:
> Current or common implementations might provide certain symbols through non-canonical header but that might not apply to any of them. Or it might change in the future (but possibly might not because of peer pressure against the ripple effect). I mentioned that LLVM tries to minimize such includes in their STL/C implementation.

I want to pay attention on the difference between implementations and the standards (C, C++, or POSIX). The standards may change as well, but it is way more difficult than to change some implementation. IWYU mappings can and should be based on the standards.

bolshakov-a:
I've just revealed that the `<cstdlib>` header doesn't include `<cstddef>` at least in libstdc++ implementation. Hence, by including `<cstddef>` along with `<cstdlib>` due to just `size_t`, you can actually make (slightly) worse.

alejandro-colomar:
> I've just revealed that the `<cstdlib>` header doesn't include `<cstddef>` at least in libstdc++ implementation. Hence, by including `<cstddef>` along with `<cstdlib>` due to just `size_t`, you can actually make (slightly) worse.

While stdlib is not a huge header (compared to stdio, or compared the template-heavy C++-only ones), it's not a light one either.  I don't think anyone including stdlib would care about the negligible build-time cost of including stddef.


```sh
alx@debian:~$ cat /usr/include/c++/14/cstddef | wc -l
192
alx@debian:~$ cat /usr/include/c++/14/cstdlib | wc -l
279
```
```sh
alx@debian:~$ cat /usr/include/c++/14/cstddef | grep ^.include
#include <bits/c++config.h>
#include <stddef.h>
#include <bits/version.h>
alx@debian:~$ cat /usr/include/c++/14/cstdlib | grep ^.include
#include <bits/c++config.h>
#include_next <stdlib.h>
#include <bits/std_abs.h>
```
```sh
alx@debian:~$ cat /usr/lib/gcc/x86_64-linux-gnu/14/include/stddef.h | wc -l
463
alx@debian:~$ cat /usr/include/stdlib.h | wc -l
1169
```
```sh
alx@debian:~$ cat /usr/lib/gcc/x86_64-linux-gnu/14/include/stddef.h | grep ^.include
#include <machine/ansi.h>
#include <sys/_types.h>
alx@debian:~$ cat /usr/include/stdlib.h | grep ^.include
#include <bits/libc-header-start.h>
#include <stddef.h>
#include <bits/floatn.h>
#include <bits/stdlib-float.h>
#include <bits/floatn.h>
```

bolshakov-a:
> And you're only skipping <cstddef>, because in the end you need <stddef.h>, which has the actual meat.

It turns out that the actual `stdlib.h` file included in `cstdlib` is not the glibc implementation (`/usr/include/stdlib.h`) but a compatibility header from the same folder as `cstdlib`. It doesn't `#include <stddef.h>`. [This](https://godbolt.org/z/8nGsr7xMo) doesn't compile unless including `cstddef` or `stddef.h`.

So if some file references `size_t` along with some other symbol defined in `cstdlib`, including `cstddef` would pull additional lines of code into the translation unit, albeit not too many.

kimgr:
I've lost track of what we're talking about here. If the standard and cppreference.com agree that `size_t` is part of a well-defined set of headers, I feel like it's best to keep those headers listed as possible providers of `size_t`. So I think I vote for this PR to be closed, unless something new comes up.

alejandro-colomar:
> I've lost track of what we're talking about here. If the standard and cppreference.com agree that `size_t` is part of a well-defined set of headers, I feel like it's best to keep those headers listed as possible providers of `size_t`. So I think I vote for this PR to be closed, unless something new comes up.

The original problem was that these mappings allow a program to include `<monetary.h>`, or `<stdio.h>`, or `<unistd.h>` or any of the huge libc headers just for basic stuff like `size_t` or `NULL`.

The ISO C standard and POSIX both centralize the definition of identifiers in a single header.  For example, `size_t` is in `<stddef.h>`, a very slim header.

See what ISO C says:

<https://port70.net/~nsz/c/c11/n1570.html#6.5.3.4p5>
> The value of the result of both operators is implementation-defined,
> and its type (an unsigned integer type) is size_t,
> defined in [<stddef.h>](https://port70.net/~nsz/c/c11/n1570.html#7.19) (and other headers).

While the identifiers are defined in a single header, other headers are also providers of the type.  See the `and other headers` note in ISO C above.  There are other similar lines in the standard, such as:

<https://port70.net/~nsz/c/c11/n1570.html#7.19p1>
> The header [<stddef.h>](https://port70.net/~nsz/c/c11/n1570.html#7.19) defines the following macros and declares the following types.
> Some are also defined in other headers, as noted in their respective subclauses.

Some of those extra headers that provide an identifier are mandated by ISO C or POSIX as providers of the type (such as `<stdio.h>` and `<monetary.h>` for size_t):

<https://port70.net/~nsz/c/c11/n1570.html#7.21.1p2>
> The types declared are size_t (described in [7.19](https://port70.net/~nsz/c/c11/n1570.html#7.19));

<https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/monetary.h.html>
> The <monetary.h> header shall define the size_t type as described in [<stddef.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stddef.h.html).

And other identifiers are allowed (but not mandated) to be provided by headers:

<https://pubs.opengroup.org/onlinepubs/9699919799/>
> Inclusion of the <stdio.h> header may also make visible all symbols from [<stddef.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stddef.h.html).

In ISO C and POSIX, there are 3 distinctions, as we can see:

-  The canonical header for an identifier, which is the one that actually defines the identifier.
-  Other headers which SHALL provide an identifier (it usually happens through inclusion).
-  Other headers which MAY provide an identifier through inclusion.

Previously, we had mappings for the first 2 items in that list, and I'm reducing it to just the first one.

cppreference.com considers the first 2 items on that list, and makes a small (undocumented) distinction between them.  It lists the canonical header first in their list, but I've seen at least one case where they got it wrong: they consider <wchar.h> as the canonical header for wchar_t, while the correct one is stddef.h> (it's a reasonable error, admittedly).

ISO C++ does not have wording for distinguishing between the first two items of that list, but that's a defect in ISO C++.  Since it defers to ISO C for the C-compatible stuff, we don't need to worry about that C++ defect.

kimgr:
> Previously, we had mappings for the first 2 items in that list, and I'm reducing it to just the first one.

Unless there are issues with that in practice (e.g. if we suggest esoteric headers like `<monetary.h>` when they are not otherwise used), I vote to keep the current behavior for `size_t`.

@bolshakov-a and @firewave You had opinions on this before, can you weigh in?

bolshakov-a:
I also prefer to keep the current mappings.

alejandro-colomar:
> > Previously, we had mappings for the first 2 items in that list, and I'm reducing it to just the first one.
> 
> Unless there are issues with that in practice (e.g. if we suggest esoteric headers like `<monetary.h>` when they are not otherwise used), I vote to keep the current behavior for `size_t`.

When suggesting additions, we always suggest a good header, because we have the first mapping in the list be the canonical mapping.

It's when removing dead code, when IWYU doesn't recommend removing esoteric headers, while it should.  Imagine some code is using <monetary.h> functions, and stops using them at some point.  If we got size_t from monetary.h, IWYU will not recommend replacing it by stddef.h.

On the other hand, if we keep size_t and NULL as special cases, I wouldn't mind too much.  We could say that, at least for now, all identifiers will only have one mapping, for the canonical header, and for these two special identifiers we could keep a few more.

Maybe we can revisit that policy in the future.

bolshakov-a:
I don't think about `size_t` as being special case, and I repeat that the problem described is present for the whole public-to-public (or one-to-many) mapping concept and should be solved by some IWYU logic change, IMHO.

alejandro-colomar:
An alternative way to reduce the mappings of these headers in the most gradual way:

Remove one secondary mapping per release.

For the next release we could remove the `<uchar.h>` / `<cuchar>` mapping from these identifiers.  Next version `<wchar.h>` / `<cwchar>`, and so on.  It would take around a dozen releases to remove them all, which would make the changes in existing software very small.

We could also start by the least used headers and leave the most used ones for the last.

Does that sound good to you?

bolshakov-a:
I still think that IWYU should not extend the list of `#include`s in a source code file if the current list already covers all the used symbols when taking into account pragmas and mappings. But it may suggest to replace more common headers to more specific (canonical) headers if the extra declarations from a common header are not used in the file. Some code should be written inside IWYU to achieve this goal.

alejandro-colomar:
> I still think that IWYU should not extend the list of `#include`s in a source code file if the current list already covers all the used symbols when taking into account pragmas and mappings.

This statement would hold.  If we remove a mapping, then it's not a mapping anymore, and IWYU will suggest adding an include so that the includes are covered by mappings again.

bolshakov-a:
I mean of course that a mapping reflects all the possibilities provided by some standard or by library author intention. To avoid a round-trip in the discussion, I want to note that I've already mentioned a couple of examples including the case when a public-to-public mapping may be inevitable even in a newly written code even without any external library.

firewave:
Just to shortly sum up my previous posts.

Out-of-the box it should be consist with (common) documentation. I wouldn't mind opt-in to only having the canonical headers though.

alejandro-colomar:
> Out-of-the box it should be consist with (common) documentation.

```
$ MANWIDTH=64 man size_t | head -n20
size_t(3type)                                     size_t(3type)

NAME
     size_t, ssize_t - count of bytes

LIBRARY
     Standard C library (libc)

SYNOPSIS
     #include <stddef.h>

     typedef /* ... */  size_t;

     #include <sys/types.h>

     typedef /* ... */  ssize_t;

DESCRIPTION
     size_t
            Used for a count of bytes.  It is the result of the
```

and

```
$ MANWIDTH=64 man NULL | head -n20
NULL(3const)                                       NULL(3const)

NAME
     NULL - null pointer constant

LIBRARY
     Standard C library (libc)

SYNOPSIS
     #include <stddef.h>

     #define NULL  ((void *) 0)

DESCRIPTION
     NULL  represents  a  null  pointer  constant,  that  is, a
     pointer that does not point to anything.

STANDARDS
     C11, POSIX.1‐2008.

```

:)

firewave:
> > Out-of-the box it should be consist with (common) documentation.
> 
> https://www.man7.org/linux/man-pages/man3/size_t.3type.html
> https://www.man7.org/linux/man-pages/man3/NULL.3const.html

https://en.cppreference.com/w/cpp/types/size_t (also `<cuchar>` starting with C++17)
https://cplusplus.com/reference/cstring/size_t/ 

```
<cstddef>
<cstdio>
<cstdlib>
<cstring>
<ctime>
<cwchar>
``` 

---

https://en.cppreference.com/w/c/types/size_t

```
<stddef.h>
<stdio.h>
<stdlib.h>
<string.h>
<time.h>
<uchar.h> (C11)
<wchar.h> (C95)
```

---

https://learn.microsoft.com/en-us/cpp/c-runtime-library/standard-types

`CRTDEFS.H and other include files`

kimgr:
@alejandro-colomar Here's my take on this: IWYU is not primarily for `#include` optimization, but for making modules express their dependencies correctly, so they are easier to change manually. One such change might be replacing `<monetary.h>` with `<stddef.h>`, but it's up to users to make that call (until @bolshakov-a improves the IWYU capabilities :slightly_smiling_face:).

So I prefer to keep the `size_t` mappings as reflecting all public headers. If users want to optimize their build times by using smaller headers, IWYU will put them in a better position to do so, but will not yet be able to do the work for them.

Let's not bounce this any more, I think we've heard all arguments. Thanks for all the valuable information and viewpoints!

firewave:
Since a lot of looking into happened here it might be worth to also have a look at #1142.

alejandro-colomar:
I've tried some small changes to remove some diagnostics (unused variable and nonnull misuse).  I've also removed some more `std::` prefixes to see if it changed anything.

```c++
alx@debian:~/tmp/iwyu$ cat test.cpp 
#include "test.h"

#include <cstring>

static size_t
f(char *s)
{
    return strchr(s, 'i') - s;
}
```
```c++
alx@debian:~/tmp/iwyu$ cat test.h 
#ifndef testH
#define testH

#include <cstddef>
#include <cstring>

size_t
flags(char *s)
{
    return std::strchr(s, 'i') - s;
}

#endif
```
```c++
alx@debian:~/tmp/iwyu$ iwyu test.cpp 

- #include <cstddef>  // lines 4-4

#include <cstring>  // for strchr, size_t
---

#include <cstddef>  // for size_t

#include "test.h"
#include <cstddef>  // for size_t
#include <cstring>  // for strchr
---
```

If I remove the remaining `std::`, then iwyu(1) reports everything is fine.  :|

PR #1575, opened 2024-07-25
Use Clang builtin headers in-place

kimgr:
In standalone mode we would copy the Clang builtin headers by replicating one of Clang's targets (clang-resource-headers), poorly.

Instead, build a -resource-dir argument implicitly on IWYU startup to point directly to the Clang headers. The path to the headers are computed by the Clang Driver based on the Clang executable path, resolved by CMake.

This does not need to happen for the external-project build mode (building as part of the LLVM tree); there the headers are already available in the build tree, and 'include-what-you-use' is installed in the same place as 'clang' so the default Clang Driver policy for resource dir lookup will work.

### DIFF
--- a/iwyu_driver.cc
+++ b/iwyu_driver.cc
@@ -274,6 +274,21 @@ bool ExecuteAction(int argc, const char** argv,
     args.push_back("-Qunused-arguments");
   }
 
+#if defined(IWYU_CLANG_EXECUTABLE_PATH)
+  // Use the full Clang executable path provided by CMake to compute Clang's
+  // resource dir, which is the base for the compiler builtin headers (stddef.h
+  // and friends). Add an explicit -resource-dir argument to tell IWYU to look
+  // there.
+  std::string resource_dir =
+      Driver::GetResourcesPath(IWYU_CLANG_EXECUTABLE_PATH);
+  args.push_back("-resource-dir");
+  args.push_back(resource_dir.c_str());
+#else
+  // The default is to look relative to the IWYU executable, which is fine if
+  // IWYU and the builtin headers are installed into the same prefix, or if IWYU
+  // is installed into the same prefix as Clang.
+#endif
+
   // Build a compilation, get the job list and filter out irrelevant jobs.
   unique_ptr<Compilation> compilation(driver.BuildCompilation(args));
   if (!compilation)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,33 +51,6 @@ option(IWYU_LINK_CLANG_DYLIB
 
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
-# Synthesize clang-resource-headers target if necessary.
-if (NOT TARGET clang-resource-headers)
-  # Use only LLVM_VERSION_MAJOR for the resource directory structure; some
-  # platforms include suffix in LLVM_VERSION.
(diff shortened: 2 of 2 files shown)


kimgr:
@tuliom Should solve the same problem as https://github.com/include-what-you-use/include-what-you-use/pull/1542

iwyu_driver.cc
```diff
@@ -274,6 +274,21 @@ bool ExecuteAction(int argc, const char** argv,
     args.push_back("-Qunused-arguments");
   }
 
+#if defined(IWYU_CLANG_EXECUTABLE_PATH)
+  // Use the full Clang executable path provided by CMake to compute Clang's
+  // resource dir, which is the base for the compiler builtin headers (stddef.h
+  // and friends). Add an explicit -resource-dir argument to tell IWYU to look
+  // there.
+  std::string resource_dir =
+      Driver::GetResourcesPath(IWYU_CLANG_EXECUTABLE_PATH);
```
tuliom, line 283: On Fedora, this function is returning `/usr/lib64/clang/18`.
The correct path is `/usr/lib/clang/18`.

I can't tell if this is a bug upstream or downstream.
So, I reported https://bugzilla.redhat.com/show_bug.cgi?id=2299949 for further investigation.

kimgr: Thank you!

tuliom: I've just found more information about this.
I'm afraid the second argument of `Driver::GetResourcesPath()` is important, otherwise it doesn't provide the correct path all the time.
I believe this should be written as `Driver::GetResourcesPath(IWYU_CLANG_EXECUTABLE_PATH, CLANG_RESOURCE_DIR)`.

See an example here:
https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/Driver.cpp#L242

And I'm afraid that moves us back to the original issue: does `include-what-you-use` get the value of `CLANG_RESOURCE_DIR`.

kimgr: @tuliom Thanks! I saw that `CLANG_RESOURCE_DIR` argument in the constructor, but didn't quite think about the ramifications.Hmm. I don't know enough about packaging on various platforms to suggest what a better behavior might be, we'll see if something falls out in the upstream discussions.

bolshakov-a:
Nice, thanks! Just yesterday, I've spent significant amount of time figuring out why those headers aren't found by IWYU...

tuliom (approved):
This now works when used together with https://github.com/llvm/llvm-project/pull/103388.

Thank you!

kimgr:
After ruminating on this for a while, I think it would be nice if it was possible to configure the build with a user-defined base binary dir and/or relative resource dir, They should default to `clang`'s output directory + nothing, which delegates to `Driver::GetResourcesDir` to figure out where Clang thinks they are.

I believe that should offer full flexibility.

@tuliom Hmm, actually, maybe that's the patch we should start with -- to let IWYU users configure resource dir base + relative resource dir, so you can configure more explicitly on Fedora? That way you have more control until https://github.com/llvm/llvm-project/pull/103388 lands (which is IWYU 0.24 at the earliest).

tuliom:
> @tuliom Hmm, actually, maybe that's the patch we should start with -- to let IWYU users configure resource dir base + relative resource dir, so you can configure more explicitly on Fedora? That way you have more control until [llvm/llvm-project#103388](https://github.com/llvm/llvm-project/pull/103388) lands (which is IWYU 0.24 at the earliest).

LGTM.
@xvitaly, considering that you're the Fedora package maintainer, what do you think?

xvitaly:
> @xvitaly, what do you think?

LGTM.

kimgr:
Superseded by #1637

Issue #1499, opened 2024-03-25
IWYU suggests to remove an include needed at compilation


gdebast:
Hello,

I'm currently investigating the usage of IWYU for our company which develop a C++ based software. I built IWYU.exe with LLVM version 16 and IWYU branch "clang_16". I'm testing it manually on our code base and I have detected what looks like a bug.

Please tell me if you need other information
Thanks already for reading
Gatien

**How to reproduce:**

Command:
`include-what-you-use.exe iwyu_objecttracker.cpp`

iwyu_objecttracker.cpp:
```
#include "iwyu_objecttracker.h"

#include "iwyu_iobject.h"

using namespace IWYU;

ObjectTracker::ObjectTracker() = default;
ObjectTracker::~ObjectTracker() = default;
```

iwyu_objecttracker.h:
```
#pragma once

#include <memory>
#include <unordered_map>

namespace IWYU
{
   class IObject;
}

namespace IWYU
{
   class ObjectTracker
   {
   public:
      ObjectTracker();

      ~ObjectTracker();

   private:
      std::unordered_map<int, std::unordered_map<int, std::unique_ptr<IObject>>> m_trackingObjects;
   };
}
```

iwyu_iobject.h:
```
#pragma once

namespace IWYU
{
   class IObject
   {
   public:
      virtual ~IObject() = default;
   };
}
```

report:
```
(iwyu_objecttracker.h has correct #includes/fwd-decls)

iwyu_objecttracker.cpp should add these lines:

iwyu_objecttracker.cpp should remove these lines:
- #include "iwyu_iobject.h"  // lines 3-3

The full include-list for iwyu_objecttracker.cpp:
#include "iwyu_objecttracker.h"
---
```

Issue:
The suggestion to remove the line `#include "iwyu_iobject.h"` in `iwyu_objecttracker.cpp` is wrong as removing the header will make compilation to fail as the destructor of IWYU:IObject is needed by the std::unique_ptr contained in the double std::unordered_map member of IWYU::ObjectTracker.

bolshakov-a:
On libc++ and libstdc++ and the latest master of IWYU, I have another output:
```
iwyu_objecttracker.h should add these lines:
#include "iwyu_iobject.h"  // for IObject

iwyu_objecttracker.h should remove these lines:
- namespace IWYU { class IObject; }  // lines 8-8

The full include-list for iwyu_objecttracker.h:
#include <memory>          // for unique_ptr
[...]

PR #1581, opened 2024-08-03
Collect types provided by class template alias

bolshakov-a:
Prior to this, provided types were harvested only from template specialization type components. Types provided by an alias of the whole template specialization type were not taken into account.

Currently, it works with nested aliases (through `GetProvidedByTplArg` -> `GetTplInstData` -> `GetTplInstDataForClass` call chain), but may further be used for class template member function calls (analysis of which doesn't currently work when the class type is aliased).

A test case added. (`NonProvidingNested` case worked already and has been added only for symmetry.)

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1594,6 +1594,7 @@ TemplateInstantiationData GetTplInstDataForClass(
     const Type* type, function<set<const Type*>(const Type*)> provided_getter) {
   TemplateInstantiationData result =
       GetTplInstDataForClassNoComponentTypes(type, provided_getter);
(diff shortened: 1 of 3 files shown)


kimgr:
LGTM. Is there a real-world code pattern that this captures? Looks like it would help with nested iterator types, for example?

bolshakov-a:
I'm not sure. I've found it investigating #1499, but it is rather unrelated, it would just help to write a smaller test case.

bolshakov-a:
Btw, what do you mean by nested iterator type problem?

kimgr:
I didn't have any specific problem in mind. The test cases just reminded me vaguely of iterators, so I was curious what the inspiration was.

PR #1582, opened 2024-08-04
Scan aliased class template member functions

bolshakov-a:
The main change is in `iwyu_ast_util.h`. It allows not to short-circuit aliased template specialization types inside
`TraverseFunctionIfInstantiatedTpl` method.

The change in `iwyu.cc` is to avoid regression with cases like
```cpp
Outer<Providing>::Inner<Providing>::AliasedTpl pp;
```
(see `tests/cxx/typedef_in_template.cc`). It facilitates collecting provided types from member function (ctor or dtor in that case) base type qualifiers.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -4412,7 +4412,7 @@ class IwyuAstConsumer
                                                         data.provided_types);
   }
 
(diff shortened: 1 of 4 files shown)


kimgr:
Nice, thanks!

PR #1583, opened 2024-08-04
Reduce 'va_list' mappings to just the canonical header

alejandro-colomar:
Reported-by: @firewave
Cc: @kimgr 
Cc: @bolshakov-a


kimgr:
I did a quick pass over the C++ draft and cppreference.com, and I only see `va_list` attributed to `stdarg`, so this seems non-controversial. Thanks!

PR #1584, opened 2024-08-05
Scan destructors of aliased type class members

bolshakov-a:
`TypeToDeclAsWritten` returns `TypedefNameDecl` in that case instead of the original `CXXRecordDecl`. But `Type::getAsCXXRecordDecl` does the right thing.

Found while investigating #1499. The type of the `_Hashtable` member of `std::unordered_map` is a `typedef` at least in the libstdc++ implementation.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -610,9 +610,8 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
       member_types.insert(it->getType().getTypePtr());
     }
     for (const Type* type : member_types) {
-      const NamedDecl* member_decl = TypeToDeclAsWritten(type);
       // We only want those fields that are c++ classes.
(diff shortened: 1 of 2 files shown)


bolshakov-a:
https://github.com/llvm/llvm-project/pull/101858 breaks the tests currently.

kimgr:
So implicit destructor handling has been effectively disabled for all fields that are typedefs/aliases instead of plain record decls? That seems like it should make a huge difference, good find!

bolshakov-a:
For heavily templated libraries, like STL, it should, probably.

bolshakov-a:
Seems like the suggested changes in the dogfood diff are from [that bug](https://github.com/include-what-you-use/include-what-you-use/issues/1499#issuecomment-2028095783). It's probably better to fix it before merging this PR.

kimgr:
Looks like this fixes #1499 now that #1588 is merged.

PR #1586, opened 2024-08-08
[clang compat] Fix alias template handling

bolshakov-a:
llvm/llvm-project@7f78f99fe5af82361d37adcbd2daa4d04afba13d removed `SubstTemplateTypeParmType` nodes from an instantiated alias template type. `GetProvidedTypes` function relied on them to filter out substituted argument types. Now, it is done "manually" inside the added `GetAliasTemplateProvidedTypes` function, as discussed in https://github.com/llvm/llvm-project/pull/101858.

When traversing instantiated type alias internals, bare `RecordType` nodes appear now instead of ones wrapped in `SubstTemplateTypeParmType`, hence the new visiting method added into `InstantiatedTemplateVisitor`. (Enum types are not expected to need handling there, hence `VisitRecordType` instead of `VisitTagType`.)

A test case added to cover the change inside `GetProvidedTypes` function (i.e. correct handling of intermediate alias templates in a chain).

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -230,6 +230,7 @@ using clang::Preprocessor;
 using clang::QualType;
 using clang::QualifiedTypeLoc;
 using clang::RecordDecl;
+using clang::RecordType;
 using clang::RecursiveASTVisitor;
 using clang::ReferenceType;
 using clang::Sema;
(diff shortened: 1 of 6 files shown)


bolshakov-a:
Something strange in the dogfood result. Maybe, I'll take a look sometime later.

bolshakov-a:
```diff
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3273,7 +3273,7 @@ class InstantiatedTemplateVisitor
   }
 
   bool VisitRecordType(RecordType* type) {
-    if (CanIgnoreCurrentASTNode())
+    if (CanIgnoreCurrentASTNode() || CanIgnoreType(type))
```
returns the old behavior, at least for `iwyu.cc`.

bolshakov-a:
Added `CanIgnoreType`, otherwise IWYU traversed the whole `RecursiveASTVisitor` due to a reference in nested name specifier on [that line](https://github.com/llvm/llvm-project/blob/llvmorg-20-init/clang/include/clang/AST/RecursiveASTVisitor.h#L631). It was at least redundant, but also caused an unwanted scanning of methods as well, due to the bug described in [that comment](https://github.com/include-what-you-use/include-what-you-use/issues/1499#issuecomment-2028095783).

kimgr:
Nice, thanks!

kimgr:
Oh, by the way, is this backward compatible with Clang 19, or is it a breaking change only compatible with Clang 20? It looks like it might work for both?

bolshakov-a:
The IWYU tests on the previous LLVM commit after this change pass, so it should be compatible with Clang 19.

PR #1588, opened 2024-08-12
Don't redundantly traverse template methods

bolshakov-a:
Otherwise, IWYU suggests the full type for `Class` for `std::map<int, std::unique_ptr<Class>>` even when none of its member functions, including the destructor, is actually used. C++ requires template member function instantiation only when used.

The bug was mentioned in https://github.com/include-what-you-use/include-what-you-use/issues/1499#issuecomment-2028095783.

### DIFF
--- a/tests/cxx/template_member_functions.cc
+++ b/tests/cxx/template_member_functions.cc
@@ -23,9 +23,24 @@ struct Tpl {
 class IndirectClass;
 using NonProviding = Tpl<IndirectClass>;
 
+template <typename T>
+struct Outer {
+  T t;
+};
+
+template <typename T>
+void FnUsingNestedTpl() {
+  Outer<Tpl<T>> o;
+}
+
 void Fn() {
   // IWYU: IndirectClass is...*indirect.h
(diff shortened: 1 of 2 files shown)


tests/cxx/template_member_functions.cc
```diff
@@ -23,9 +23,23 @@ struct Tpl {
...
+void FnUsingNestedTpl() {
+  Outer<Tpl<T>> o;
+}
+
 void Fn() {
   // IWYU: IndirectClass is...*indirect.h
   NonProviding::StaticFn();
+
+  // IndirectClass is not fully used here. For this test case, it is important
+  // that StaticFn is used in the translation unit to be instantiated.
```
kimgr, line 41: Can you just add a little clause to also describe _why_ `StaticFn` is important for this (seemingly-unrelated) construct?

kimgr: I'm looking for a `, so that ...` or `, because ...` -- hopefully there's a short rationale hiding.

Issue #1000, opened 2022-01-30
additional unnecessary platform include suggested for std::abs()


firewave:
```cpp
#include <cstdlib>

static void f()
{
	std::abs(-12);
}
```

```
/mnt/s/clion/example_lite_2/test.cpp should add these lines:
#include <bits/std_abs.h>  // for abs

/mnt/s/clion/example_lite_2/test.cpp should remove these lines:

The full include-list for /mnt/s/clion/example_lite_2/test.cpp:
#include <bits/std_abs.h>  // for abs
#include <cstdlib>         // for abs
---
```

kimgr:
Thanks! Just to make sure, which iwyu version is this? There was a big mapping cleanup for lib headers recently, which could affect this behavior.

firewave:
I am using `include-what-you-use 0.17 based on Debian clang version 13.0.1-+rc1-1~exp4`.

firewave:
I just tested it with `<cmath>` and that gets a bit weird:

```cpp
#include <cmath>

static void f()
{
	std::abs(-12);
}
```

```
/mnt/s/clion/example_lite_2/test.cpp should add these lines:
#include <bits/std_abs.h>  // for abs
#include <stdlib.h>        // for abs

/mnt/s/clion/example_lite_2/test.cpp should remove these lines:
- #include <cmath>  // lines 1-1

The full include-list for /mnt/s/clion/example_lite_2/test.cpp:
#include <bits/std_abs.h>  // for abs
#include <stdlib.h>        // for abs
---
```

That probably needs to be tracked in another ticket.

Bidski:
I am also seeing this issue with version 0.18. I am using `std::abs` and including `<cmath>` (I also tried `<cstdlib>`) but IWYU says 
```
Warning: include-what-you-use reported diagnostics:

/home/abiddulph/vo/utility/find_roots.hpp should add these lines:
#include <bits/std_abs.h>  // for abs

/home/abiddulph/vo/utility/find_roots.hpp should remove these lines:
- #include <cmath>  // lines 6-6

The full include-list for /home/abiddulph/vo/utility/find_roots.hpp:
#include <bits/std_abs.h>  // for abs
#include <algorithm>       // for transform, swap_ranges, generate
#include <array>           // for swap, array
#include <complex>         // for operator-, operator/, operator*, operator+
#include <cstdlib>         // for size_t
#include <iterator>        // for next
#include <numeric>         // for accumulate
[...]

PR #1591, opened 2024-08-14
Mapped `<bits/std_abs.h>` to `<stdlib.h>`

firewave:
Refs #1000.

### DIFF
--- a/gcc.libc.imp
+++ b/gcc.libc.imp
@@ -79,6 +79,7 @@
   { "include": [ "<bits/stat.h>", "private", "<sys/stat.h>", "public"] },
   { "include": [ "<bits/statfs.h>", "private", "<sys/statfs.h>", "public"] },
   { "include": [ "<bits/statvfs.h>", "private", "<sys/statvfs.h>", "public"] },
+  { "include": [ "<bits/std_abs.h>", "private", "<stdlib.h>", "public"] },
   { "include": [ "<bits/stdio-ldbl.h>", "private", "<stdio.h>", "public"] },
   { "include": [ "<bits/stdio-lock.h>", "private", "<libio.h>", "public"] },
   { "include": [ "<bits/stdio.h>", "private", "<stdio.h>", "public"] },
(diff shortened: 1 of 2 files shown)


firewave:
This fixes the suggestion of the implementation header but afterwards the actual suggestion will flip-flop between `<math.h>` and `<stdlib.h>`.

firewave:
> This fixes the suggestion of the implementation header but afterwards the actual suggestion will flip-flop between `<math.h>` and `<stdlib.h>`.

Maybe this is related to the symbol being attributed to `<stdlib.h>` only in C and `<cstdlib>`/`<cmath>` in C++.

Mapping it to `<stdlib.h>` fixes the flip-flop but will cause correct C++ code using `<cmath>` suggest to use `<stdlib>` instead.

kimgr:
Looks like `<bits/std_abs.h>` is a C++-only header, it's only included from `<cmath>` and `<cstdlib>` on my system:
```
kimgr@lardo:/usr/include$ grep -r "std_abs\.h"
c++/11/cmath:#include <bits/std_abs.h>
c++/11/cstdlib:#include <bits/std_abs.h>
c++/11/bits/std_abs.h:/** @file include/bits/std_abs.h
```

So it can't be right to map the private header to the public C headers (since they never include it).

kimgr:
Hmm, actually, I just stumbled on a bug -- IWYU parses `@file` and `@headername` directives in libstdc++ headers. A bunch of headers are incorrectly marked with `@file include/...actual name...`, which causes IWYU to generate incorrect mappings. Turns out `std_abs.h` is one of them: https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/std_abs.h#L25

kimgr:
I thought I'd fix that bug, and then this PR should not be necessary

kimgr:
This has been fixed by #1598 and #1593 so that mapping for `bits/std_abs.h` happens automatically.

PR #1593, opened 2024-08-17
Remove explicit @headername mappings

kimgr:
IWYU detects libstdc++ @headername directives and dynamically generates mappings from them, so the explicit mappings based on grep-ing the libstdc++ headers for @headername were entirely redundant.

These explicit mappings were added in
1e407719011dad4b317a3b29f25f22c755e176b3, presumably without any observable effect. Possibly masked by the fact that another set of mapping changes happened at the same time.

No functional change expected, all tests pass unmodified.


kimgr:
Dogfood results highlighted some weaknesses in automatic headername mappings. Addressed in #1594.

PR #1594, opened 2024-08-17
Handle broken @file directives

kimgr:
Looking closer at libstdc++'s @file directives (used for @headername auto-mapping), I noticed some irregularities:

* bits/shared_ptr.h has @file without a name
* bits/std_abs.h (and many others) has @file with a leading 'include/', which doesn't match the file's actual include-name

These are strictly speaking bugs in libstdc++, but we can't fix them and wait for users to upgrade. Instead, be helpful and use some heuristics to normalize the names:

* If name is empty, synthesize a best-effort name from full file path
* If name starts with 'include/', simply strip the prefix

Fixes issue #1000, and should improve the automatic @headername mappings in general.

### DIFF
--- a/iwyu_path_util.cc
+++ b/iwyu_path_util.cc
@@ -213,6 +213,12 @@ bool IsQuotedInclude(const string& s) {
           (StartsWith(s, "\"") && EndsWith(s, "\"")));
 }
 
+// Removes <> or "" from a quoted include.
+string StripQuotes(const string& quoted_include) {
+  CHECK_(IsQuotedInclude(quoted_include));
+  return quoted_include.substr(1, quoted_include.size() - 2);
+}
+
 // Returns whether this is a system (as opposed to user) include file,
 // based on where it lives.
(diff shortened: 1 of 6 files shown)


kimgr:
@firewave Please take a look. This should replace #1591.

kimgr:
There was some good feedback from a libstdc++ maintainer here: https://gcc.gnu.org/pipermail/libstdc++/2024-August/059385.html.

I had completely missed this, but `@file` is a Doxygen construct, and `@headername` is a libstdc++-specific alias. So most of them are not broken, they just follow Doxygen's loose semantics: https://www.doxygen.nl/manual/commands.html#cmdfile

Looks like the `@file` directive is not a good source of information for the include-name, since its path elements are optional.

kimgr:
I don't think this PR is the way to go. I'm investigating trying to generate the relevant relative path from the actual filename instead.

kimgr:
I have a better solution coming shortly. Closing this.

PR #1596, opened 2024-08-22
Take most basic method for author-intent analysis

bolshakov-a:
Because overriding method prototypes need generally neither the full info nor the forward-declaration for the return type and for the parameter types, the author intent analysis should be performed only for the original declaration from the upmost class in the hierarchy. This change does that at the call site.

This is a step towards fixing #105.

The test case has been partially taken from #675.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1458,6 +1458,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (!fn_decl)     // TODO(csilvers): what to do for fn ptrs and the like?
       return set<const Type*>();
 
+    if (const auto* method = dyn_cast<CXXMethodDecl>(fn_decl))
+      fn_decl = GetFromLeastDerived(method);
+
     // Collect the non-explicit, one-arg constructor ('autocast') types.
     set<const Type*> autocast_types;
     for (FunctionDecl::param_const_iterator param = fn_decl->param_begin();
@@ -1494,8 +1497,11 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   }
 
   set<const Type*> GetProvidedTypesForFnReturn(const FunctionDecl* decl) const {
-    const Type* return_type = decl->getReturnType().getTypePtr();
-    return GetProvidedTypes(return_type, GetLocation(decl));
+    const FunctionDecl* providing_decl = decl;
+    if (const auto* method = dyn_cast<CXXMethodDecl>(decl))
+      providing_decl = GetFromLeastDerived(method);
+    const Type* return_type = providing_decl->getReturnType().getTypePtr();
+    return GetProvidedTypes(return_type, GetLocation(providing_decl));
   }
 
   //------------------------------------------------------------
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1310,6 +1310,12 @@ bool IsImplicitlyInstantiatedDfn(const FunctionDecl* decl) {
              clang::TSK_ImplicitInstantiation;
 }
 
+const CXXMethodDecl* GetFromLeastDerived(const CXXMethodDecl* decl) {
+  while (decl->size_overridden_methods())
+    decl = *decl->begin_overridden_methods();
+  return decl;
+}
+
 // --- Utilities for Type.
 
 const Type* GetTypeOf(const Expr* expr) {
--- a/iwyu_ast_util.h
+++ b/iwyu_ast_util.h
@@ -717,6 +717,11 @@ bool IsBuiltinFunction(const clang::NamedDecl* decl);
 // (in particular, not just a declaration).
 bool IsImplicitlyInstantiatedDfn(const clang::FunctionDecl*);
 
+// If the given method overrides base class methods, returns the overridden
+// method declaration from the least derived class, otherwise returns the given
+// argument.
+const clang::CXXMethodDecl* GetFromLeastDerived(const clang::CXXMethodDecl*);
+
 // --- Utilities for Type.
 
 const clang::Type* GetTypeOf(const clang::Expr* expr);
--- a/tests/cxx/reexport_overridden-d1.h
+++ b/tests/cxx/reexport_overridden-d1.h
@@ -1,0 +1,29 @@
+//===--- reexport_overridden-d1.h - test input file for iwyu --------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/reexport_overridden-i1.h"
+
+class Derived1 : public Base {
+ public:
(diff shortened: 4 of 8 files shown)


kimgr:
Super cool, thanks for taking this on!

iwyu_ast_util.cc
```diff
@@ -1310,6 +1310,12 @@ bool IsImplicitlyInstantiatedDfn(const FunctionDecl* decl) {
              clang::TSK_ImplicitInstantiation;
 }
 
+const CXXMethodDecl* GetUpmostInHierarchy(const CXXMethodDecl* decl) {
```
kimgr, line 1313: I don't love this name, but it's not easy to come up with a better one... How about `GetLeastDerived`?

iwyu_ast_util.h
```diff
@@ -717,6 +717,11 @@ bool IsBuiltinFunction(const clang::NamedDecl* decl);
 // (in particular, not just a declaration).
 bool IsImplicitlyInstantiatedDfn(const clang::FunctionDecl*);
 
+// If the given method overrides base class methods, returns the overridden
+// method declaration from the most basic class, otherwise returns the given
```
kimgr, line 721: Here I think `least derived` is definitely better than `most basic`.

iwyu.cc
```diff
@@ -1494,8 +1497,11 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   }
 
   set<const Type*> GetProvidedTypesForFnReturn(const FunctionDecl* decl) const {
-    const Type* return_type = decl->getReturnType().getTypePtr();
-    return GetProvidedTypes(return_type, GetLocation(decl));
+    const FunctionDecl* providing_decl = decl;
+    if (const auto* method = dyn_cast<CXXMethodDecl>(decl))
+      providing_decl = GetUpmostInHierarchy(method);
```
kimgr, line 1502: So in a scenario like:
```
class Base {};
class Derived0 : public Base {};
class Derived1 : public Derived0 {};
class Derived2 : public Derived1 {};
class Derived3 : public Derived2 {};
```
... we run through the whole chain up to the least derived/most base and use its return type/parameter types.

Do you think maybe we'd need to check each intermediary step too? I.e. if a type is provided at any level in the inheritance chain, it would be provided below, right? I feel like that would make this a lot more complicated, though.

iwyu_ast_util.cc
```diff
@@ -1310,6 +1310,12 @@ bool IsImplicitlyInstantiatedDfn(const FunctionDecl* decl) {
...
+  while (decl->size_overridden_methods())
```
kimgr, line 1314: I'm not sure it's cleaner, but isn't the least derived/most basic method `std::prev(decl->end_overridden_methods())`? Maybe that iterator is not backward-friendly, though.

iwyu.cc
bolshakov-a: I just thought that IWYU should generally remove all `#include`s and fwd-decls for return and parameter types at any level except `Base`, hence a user cannot provide a type at the other levels. Moreover, according to the default policy, the absence of both `#include` and fwd-decl means for IWYU that the type is provided, which would be obviously wrong in many cases.

iwyu_ast_util.cc
bolshakov-a: I tried that and it doesn't work. I suspect that `overridden_method_range` includes only method declarations from one level, i.e. there is more than one item only in the case of a diamond hierarchy with virtual inheritance.

bolshakov-a:
```cpp
      // A method can only override one function from each base class. We
      // don't track indirectly overridden methods from bases of bases.
```
https://github.com/llvm/llvm-project/blob/llvmorg-20-init/clang/lib/Sema/SemaDecl.cpp#L8847

iwyu.cc
kimgr: I see, that makes sense.

> Moreover, according to the default policy, the absence of both #include and fwd-decl means for IWYU that the type is provided, which would be obviously wrong in many cases.

I thought the policy was:

* `#include`: author is responsible (i.e. provided)
* fwd decl: user is responsible (i.e. not provided)
* neither: can't compile?

so I'm not really getting that... Or wait, you mean that "is provided?" only returns true if there is a forward-decl, whether there's an `#include` directive in place or not (on the assumption that code can't compile without a declaration)?

bolshakov-a: > neither: can't compile?

`#include` or fwd-decl can come via some other header, even via `stdafx.h`, so the code can actually compile. `CodeAuthorWantsJustAForwardDeclare` function considers only direct `#include`s and fwd-decls from the file defining a type alias or a function prototype.

Indeed, the proposed changes are to avoid intentionally the collecting of provided types from intermediate method declarations, otherwise `FwdRetType` and `FwdAutoconvParam` are considered as provided by `Derived1` and by `Derived2` just because they are not forward-declared (neither included) in `reexport_overridden-d1.h` (see [the comment](https://github.com/include-what-you-use/include-what-you-use/pull/1596/files#diff-4476e0e75321badf44d0ba46b72836581736bf2628b62859baf7d0774f9d8f28R20)). You can compare with the behavior on the current master.

iwyu_ast_util.cc
bolshakov-a: `GetFromLeastDerived`, maybe?

iwyu_ast_util.cc
kimgr: Got it, thanks

iwyu_ast_util.cc
kimgr: Sounds good to me!

iwyu.cc
kimgr: > considers only *direct* #includes and fwd-decls from the file defining a type alias or a function prototype.

Ah, that's what I was missing, thanks.

I feel like there might be a use-case for breaking the reexport chain halfway through a hierarchy, but I can't really put it into words. And the suggested behavior is so nice and simple that it feels better to find out through testing in the wild.

kimgr:
I also wanted to ask about covariant return types. As I understand this, a covariant return type could never be reexported from base, because it's a different type? And `VisitCXXMethodDecl` is free to report the type as required?

bolshakov-a:
Yeah, and it would be a drawback of the proposed approach (because a covariant return type can be redundantly required for including at call sites of the derived class method) if the provision of returned pointers and references worked actually. But it doesn't work (it is even [stated in the docs](https://github.com/include-what-you-use/include-what-you-use/blob/0.22/docs/WhatIsAUse.md?plain=1#L113)) because the information that the referenced type is returned from a function is lost at the point of reporting a full use.

kimgr:
> But it doesn't work (it is even [stated in the docs](https://github.com/include-what-you-use/include-what-you-use/blob/0.22/docs/WhatIsAUse.md?plain=1#L113)) because the information that the referenced type is returned from a function is lost at the point of reporting a full use.

Ah, yeah. I was trying to fix that with custom traversal that adds `Traverse/VisitFunctionReturnType`, but I never got it ready.

PR #1598, opened 2024-08-25
Use HeaderSearch for include name lookup instead of @file

kimgr:
The @headername directive is a Doxygen alias used in libstdc++, but other projects have also picked up the convention. When generating implicit mappings for @headername, we used to use the Doxygen @file directive to get the from-name and @headername to get the preferred to-names.

There were a couple of problems with our interpretation of these pragmas:

* @file is manually maintained, and may be empty, incorrect or ambiguous
* Doxygen requires that @file contains enough of the file path in the build tree to be unique, so the name spelled out there is not necessarily the include-name we should use (e.g. many files are prefixed with "include/" for disambiguation)
* We used to assume all @headername names were system headers. This is true for libstdc++, but not necessarily for other projects.

Change this to use the private header FileEntry instead, and use the Clang preprocessor to get the actual include-name that brought the file into the source manager in the first place.

Move the include-name handling to after a @headername has been found to avoid the work if there's no mapping information anyway.

Change quoting for the to-name so it's not hard-coded to angle quotes, but rather use the same quote style as the private header (figuring out the quote style for an arbitrary public header string is non-trivial). Update a test to use the right quoting.

This fixes a bunch of mis-mappings for libstdc++.

### DIFF
--- a/iwyu_preprocessor.cc
+++ b/iwyu_preprocessor.cc
@@ -17,6 +17,7 @@
 #include "clang/AST/Decl.h"
 #include "clang/Basic/IdentifierTable.h"
 #include "clang/Basic/TokenKinds.h"
+#include "clang/Lex/HeaderSearch.h"
 #include "clang/Lex/MacroInfo.h"
 #include "clang/Lex/Token.h"
 #include "iwyu_ast_util.h"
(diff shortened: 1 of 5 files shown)


kimgr:
@firewave After some discussions with a libstdc++ maintainer, I came to the conclusion that we can't really trust `@file` at all, so I found a way to resolve the private include-name without it. This should fix the bits/std_abs.h issue you're seeing in #1591 (except for the `<cmath>` thing, which has a deeper root cause).

firewave:
Sorry for the late reply. I have been out sick and am still not fully recovered.

I followed the discussion and was actually aware of the @file  quirks as I had to deal with them in the past so I could have chimed in on that. Have not dealt with @headername though.

> This should fix the bits/std_abs.h issue you're seeing in https://github.com/include-what-you-use/include-what-you-use/pull/1591 (except for the <cmath> thing, which has a deeper root cause).

Should there be a test added for such a case? The `<functional>` case exposed in dogfooding would be a candidate.

kimgr:
Hoping for a quick recovery!

There are tests for the mechanism in `comment_pragmas*`. I try to avoid bringing in standard library headers in tests, because they're not portable, and it's hard to make the test succeed/fail for the right reasons across implementations.

I actually might revert part of this patch, so it only catches `@headername` after `@file`, because I think that's a reasonable constraint for Doxygen-enabled projects. I will still ignore the filename after `@file`.

firewave:
> Hoping for a quick recovery!

Five years and counting so far ... but I digress.
 
> There are tests for the mechanism in `comment_pragmas*`. I try to avoid bringing in standard library headers in tests, because they're not portable, and it's hard to make the test succeed/fail for the right reasons across implementations.

Makes sense. The dogfooding is a good integration test in that sense. I wonder if that should be extended with libc++ though.
 
> I actually might revert part of this patch, so it only catches `@headername` after `@file`, because I think that's a reasonable constraint for Doxygen-enabled projects. I will still ignore the filename after `@file`.

Oh, I missed the fact that this is a general thing. Might make sense to look into some existing projects how they are using doxygen. Or just wait for feedback ...

kimgr:
> Oh, I missed the fact that this is a general thing. Might make sense to look into some existing projects how they are using
> doxygen. Or just wait for feedback ...

Yeah, `@file` is plain Doxygen, but `@headername` is a libstdc++ alias that generates a documentation fragment: https://github.com/gcc-mirror/gcc/blob/898f013e195fa828bb30ae6ba4ad50abbd804fbd/libstdc%2B%2B-v3/doc/doxygen/user.cfg.in#L255.

But to make matters more interesting, other projects have also picked up the `@headername` convention, like: https://github.com/jkovacic/math/blob/4d5ee0e68b2db0e3bb4b9bd23d0df2ded2b8586c/lib/matrix/MatrixGeneric.hpp#L23 (doesn't appear to use Doxygen, or at least has no in-tree config).

firewave:
> Yeah, `@file` is plain Doxygen, but `@headername` is a libstdc++ alias that generates a documentation fragment:

`@headername` is an official doxygen command: https://www.doxygen.nl/manual/commands.html#cmdheaderfile

firewave:
> `@headername` is an official doxygen command: https://www.doxygen.nl/manual/commands.html#cmdheaderfile

Oh shoot, `@headername` vs. `@headerfile`. Nevermind.

kimgr:
Oof, those two are close enough as to be indistinguishable.

PR #1602, opened 2024-08-29
Don't report overriding method signature types

### LINKED ISSUE #105: types used in functions marked "override" need not have the corresponding header included

bolshakov-a:
There is no need to include or fwd-declare parameter and return (except covariance) types for overriding method declarations because they should be available through the file declaring the primary (least derived) method. This has been achieved with two steps:

1) avoid suggestions to include the full type definitions to provide them (the change inside `CanBeProvidedTypeComponent`), and

2) block forward-declaration suggestions for the return type and all the parameter types and their components.

The latter has been done not only for signatures but also for bodies so as to avoid suggestions to add a fwd-decl for `Class` in cases like:
```cpp
void Derived::Method(std::vector<Class>& v) {
  for (Class& obj : v)
    TakeByRef(obj);
}
```
Fwd-declaration suggestions inside source files are annoying, and I don't want to increase their number. Fwd-decls are blocked even for components of complex aliased types, but I'm not sure if it is worth introducing `CanonicalTypeEnumerator`...

Any sugaring declarations (type aliases, using-declarations, and other) should still be reported because `Base` and `Derived` definitions may use them independently from each other.

There remains a question whether types provided by `Base` declarations should be reported for inclusion when fully used at `Derived` method implementation side or not. This change doesn't block them, hence it effectively pushes them out from `derived.h` to `derived.cpp` in many cases. But indeed, the fact that `Base` provides them may be considered as a `Base` "implementation" detail, likewise with sugaring. So it allows to change the author-intent without changing `derived.cpp`. If it will be desired, this behavior may be changed in a separate PR.

This fixes #105.

The test cases have been partially taken from #675.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1579,8 +1579,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // Canonicalize the use location and report the use.
     used_loc = GetCanonicalUseLocation(used_loc, target_decl);
     OptionalFileEntryRef used_in = GetFileEntry(used_loc);
-    preprocessor_info().FileInfoFor(used_in)->ReportForwardDeclareUse(
-        used_loc, target_decl, use_flags, comment);
+    if (!blocked_for_fwd_decl_.count(target_decl->getCanonicalDecl())) {
+      preprocessor_info().FileInfoFor(used_in)->ReportForwardDeclareUse(
+          used_loc, target_decl, use_flags, comment);
+    }
 
     // If we're a use that depends on a using declaration, make sure
     // we #include the file with the using declaration.
@@ -2622,6 +2624,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   }
 
   set<const Type*> blocked_types_;
+  set<const Decl*> blocked_for_fwd_decl_;
 
  private:
   template <typename T> friend class IwyuBaseAstVisitor;
@@ -4223,6 +4226,36 @@ class IwyuAstConsumer
     return HandleAliasedClassMethods(decl);
   }
 
+  // Avoid forward-declaration warnings for types which _should_ be already
+  // provided by the primary method declaration. For example,
+  // for 'struct Derived : public Base { T1 method(const T2&); };', T1 and T2
+  // are provided by Base, and should not be forward-decl reported in Derived.
+  bool TraverseCXXMethodDecl(CXXMethodDecl* method_decl) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+
+    if (!method_decl->size_overridden_methods())
+      return Base::TraverseCXXMethodDecl(method_decl);
+
+    set<const Decl*> fwd_blocked = blocked_for_fwd_decl_;
+    auto add_canonical_components = [&fwd_blocked](QualType part) {
+      for (const Type* type : GetCanonicalComponentsOfType(part.getTypePtr())) {
+        // For templates, use GetInstantiatedFromDecl so as to match
+        // what VisitTemplateSpecializationType does.
+        const Decl* decl = GetInstantiatedFromDecl(TypeToDeclAsWritten(type));
+        if (decl)
+          fwd_blocked.insert(decl->getCanonicalDecl());
+      }
+    };
+    add_canonical_components(method_decl->getReturnType());
+    for (const ParmVarDecl* param : method_decl->parameters())
+      add_canonical_components(param->getType());
+    ValueSaver<set<const Decl*>> s{&blocked_for_fwd_decl_, fwd_blocked};
+    // Full type info suggestions are blocked in CanBeProvidedTypeComponent.
+
+    return Base::TraverseCXXMethodDecl(method_decl);
+  }
+
   // --- Visitors of types derived from Stmt.
 
   // Called whenever a variable, function, enum, etc is used.
@@ -4540,6 +4573,11 @@ class IwyuAstConsumer
       // in source files, or for builtins, or for friend declarations.
       if (!IsInHeader(decl) || IsFriendDecl(decl))
         return pair(false, nullptr);
+      if (const auto* method = dyn_cast<CXXMethodDecl>(decl)) {
+        // Only the primary (least derived) declaration can provide types.
+        if (method->size_overridden_methods())
+          return pair(false, nullptr);
+      }
       for (const ParmVarDecl* param : decl->parameters()) {
         if (node->StackContainsContent(param)) {
           if (HasImplicitConversionConstructor(GetTypeOf(param)))
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -683,6 +683,48 @@ class TypeEnumeratorWithoutSubstituted
   set<const Type*> seen_types_;
 };
 
+class CanonicalTypeEnumerator
+    : public RecursiveASTVisitor<CanonicalTypeEnumerator> {
+ public:
+  // --- Public interface
+  set<const Type*> Enumerate(const Type* type) {
+    seen_types_.clear();
+    if (!type)
+      return seen_types_;
+    TraverseType(QualType(type, 0));
+    return seen_types_;
+  }
+
+  // --- Methods on RecursiveASTVisitor
+  bool TraverseType(QualType type) {
+    if (type.isNull())
+      return Base::TraverseType(type);
+    return TraverseTypeHelper(type);
+  }
+
+  bool TraverseTypeLoc(TypeLoc type_loc) {
+    if (!type_loc)
+      return Base::TraverseTypeLoc(type_loc);
+    return TraverseTypeHelper(type_loc.getType());
+  }
+
+ private:
+  typedef RecursiveASTVisitor<CanonicalTypeEnumerator> Base;
+
+  bool TraverseTypeHelper(QualType qual_type) {
+    CHECK_(!qual_type.isNull());
+
+    const Type* type = qual_type.getTypePtr();
+    seen_types_.insert(GetCanonicalType(type));
+
+    const Type* desugared = DesugarAliasTypes(type);
+    // Add desugared type components.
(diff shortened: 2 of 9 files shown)


kimgr:
Beautiful!

While I like the rationale for requiring re-declaration/-inclusion of aliased types in derived classes, is it practical? Won't that disable the "optimization" for basically all STL types, i.e. `std::string` is typically just a `std::basic_string`, etc.

> There remains a question whether types provided by Base declarations should be reported for inclusion when fully used
> at Derived method implementation side or not. 

When I was experimenting with this, I aimed for them to be ignored in the function prototype (return type and parameter types), but not in the body. That way if a method was forced by its based to mention a type, the base should take responsibility. But if the method chose to use a type in its body, it would be responsible.

On the other hand it's hard to define a function with a return type without also mentioning it in the body. So maybe it makes sense to rely on base to provide them throughout declaration, prototype definition and body.

kimgr:
For example with:
```
// base.h
#include <string>
#include <variant>

class Base {
  virtual std::string GetString() const;
  virtual void Assign(std::variant<int, std::string>& v, int i);
};
```
and its derived declaration:
```
#include "base.h"

class Derived : public Base {
public:
  std::string GetString() const override;
  void Assign(std::variant<int, std::string>& v, int i) override;
};
```
and definition:
```
#include "derived.h"

std::string Derived::GetString() const {
  return "foo";
}

void Derived::Assign(std::variant<int, std::string>& v, int i) {
  v = i;
}
```

IWYU says:
```
tests/105/derived.h should add these lines:
#include <string>  // for string

tests/105/derived.h should remove these lines:

The full include-list for tests/105/derived.h:
#include <string>  // for string
#include "base.h"  // for Base
---

tests/105/derived.cc should add these lines:
#include <variant>  // for variant

tests/105/derived.cc should remove these lines:

The full include-list for tests/105/derived.cc:
#include "derived.h"
#include <variant>  // for variant
```

This follows from your explanation, but intuitively I would expect IWYU to leave `<string>` and `<variant>` in `base.h` only.

kimgr:
... and I replace `std::string` and `std::variant` with plain struct types `X` and `Y` the behavior changes. Also expected, per your explanation, but I think somewhat questionable.

bolshakov-a:
Yes, despite my approach obeys IWYU principle, it is rather impractical, because overriding method signatures are usually just copy-pasted from the primary declaration, and requirements to `#include` type alias declarations into `Derived`-defining files are annoying. Moreover, declarations from `std` cannot be forward-declared, hence they _should_ be available through `Base`-defining file.

But honestly, blocking of full type infos and sugar declarations seems to be trickier, and some more questions need to be answered. E.g. what about default arguments and `noexcept`-specifiers? Should declarations involved in them be considered as provided? For example:
```cpp
struct A {};

struct B {
    B(const A&);
};

void TakeB(const B&);

struct Base {
  virtual void Fn() noexcept(noexcept(TakeB(A{})));
};

struct Derived : Base {
  void Fn() noexcept(noexcept(TakeB(A{}))) override;
};
```
The `Base`-defining file should `#include` the `B`-defining file to allow implicit conversion inside the `TakeB` call. However, `B` is not explicitly mentioned inside `Base::Fn` declaration, hence it cannot be found by just enumerating types or `DeclRefExpr`s. Seems like a child `IwyuAstConsumer` should be launched to collect all the provided declarations. In fact, I want to do something similar someday (so as to allow `InstantiatedTemplateVisitor` to report types unrelated to template arguments), but it looks like a serious refactoring.

kimgr:
I see (to some extent). I thought the current behavior was an explicit choice, and it would be even easier to re-export all of Base (haven't really looked at the patch yet). But yeah, in that case I guess this is a good step which re-exports a controlled subset of types from base to derived (ie all non-aliases)... Right?

bolshakov-a:
> I thought the current behavior was an explicit choice, and it would be even easier to re-export all of Base

Do you mean considering `base.h` as a `derived.cpp` associated header, along with `derived.h`? I don't think it is a good idea, because it would lead to re-export of many implementation details of inherited classes from frameworks like Qt. If e.g. `QDialog` uses e.g. `QList` in its `private` section, `class YourDialog : public QDialog` should not rely on that so as not to be broken if the Qt people decide to refactor the `QDialog` internals.

Regarding the author-intent analysis, I've realized that it may follow the similar logic: if some overridden function from `QDialog` returns e.g. `QVariant`, it may be better to `#include <QVariant>` in `YourDialog.cpp`, so as not to be broken if the Qt people decide to replace the `#include` with a `QVariant` forward-declaration.

kimgr:
> > I thought the current behavior was an explicit choice, and it would be even easier to re-export all of Base
> 
> Do you mean considering `base.h` as a `derived.cpp` associated header, along with `derived.h`? I don't think it is a good idea, because it would lead to re-export of many implementation details of inherited classes from frameworks like Qt. If e.g. `QDialog` uses e.g. `QList` in its `private` section, `class YourDialog : public QDialog` should not rely on that so as not to be broken if the Qt people decide to refactor the `QDialog` internals.

Yeah, agreed. I meant blindly exporting virtual method types to overrides, not everything 

> Regarding the author-intent analysis, I've realized that it may follow the similar logic: if some overridden function from `QDialog` returns e.g. `QVariant`, it may be better to `#include <QVariant>` in `YourDialog.cpp`, so as not to be broken if the Qt people decide to replace the `#include` with a `QVariant` forward-declaration in `qdialog.h`.

Hmm, yeah, it's a tricky trade-off between isolation/modularization and obvious convenience. I'm thinking inheritance is such a strong coupling that such a change to the base is expected to have repercussions in the derived subtree. Isn't the change from provided to forward-declared a new hint to the derived classes that they're now responsible? A "breaking change" of sorts.

bolshakov-a:
> Isn't the change from provided to forward-declared a new hint to the derived classes that they're now responsible?

It may be. Furthermore, callers of the method would still become broken, because the author-intent analysis at their side still works.

bolshakov-a:
> I meant blindly exporting virtual method types to overrides, not everything

However, again, not all (for the full type info), but at most only those, for which IWYU should keep `#include` in `base.h` (due to being provided by the author-intent, which doesn't work for parameter types without an implicit one-arg ctor, or due to being fully used e.g. inside `noexcept`-specifier).

bolshakov-a:
> I thought the current behavior was an explicit choice, and it would be even easier to re-export all of Base

I've tried to block provided types at first, but I've faced with questions about implementation (`VisitTagType` doesn't check `blocked_types_`, how to do it correctly? How to collect sugaring types, when `GetProvidedTypes` returns only canonical? What to do with `UsingType`s?) and about the desired behavior (which types should be considered as provided actually?) And then I've realized that this feature, besides involving difficulties, _would contradict with the IWYU principle_, strictly speaking. And I've decided to stop with the strict policy proposed in this PR. I think that any relaxation should be made as a separate commit. Maybe, the behavior is excessively strict regarding to the types from `std` namespace, but technically, you generally can forward declare them despite the standard forbids that.

I can try to do some relaxation, if the desired behavior will be determined.

Previously, I used an IWYU version [which blocks return and parameters type completely](https://github.com/bolshakov-a/include-what-you-use/commit/65941893156e9ebc1120e51e403998c7c2a62d23), but only at the declarations and not at the definitions. But it is incorrect, because an overriding method declaration may use custom sugaring in general. There were suggestions of types at the definition side, and especially annoying for `typedef`s, because I thought that `typedef`s should be provided by the primary declaration. But now, I understand that it is a reasonable strict policy.

kimgr:
> [...] I've realized that this feature, besides involving difficulties, would contradict with the IWYU principle,[...]

Yeah, it tries to force the absolute opposite: exploit transitive includes to the max.

Just to make sure I understand the high-level, though: this PR does not make IWYU _more_ strict in any cases, only less strict in some cases, right? And by "strict" i mean "requires more local `#include`s". So it seems to me either way it's an improvement on the status quo, as it begins to re-export from Base methods to Derived methods for a few safe cases, where before we did none.

iwyu.cc
```diff
@@ -4223,6 +4226,34 @@ class IwyuAstConsumer
     return HandleAliasedClassMethods(decl);
   }
 
+  // Avoid warnings for types which _should_ be already provided by the primary
+  // method declaration.
```
kimgr, line 4230: Might be worth being more specific with `Avoid forward-declare warnings...`
 
Also, a micro-example here would make it easier to understand this, something like;
```
  // For example, for 'struct Derived : public Base { T1 method(const T2&); };',
  // T1 and T2 are provided by Base, and should not be forward-decl reported in
  // Derived.
```

kimgr (commented):
LGTM, let's just polish up the comment for fwd-decl handling

bolshakov-a:
> Just to make sure I understand the high-level, though: this PR does not make IWYU _more_ strict in any cases, only less strict in some cases, right? And by "strict" i mean "requires more local `#include`s".

Yes, because in all cases when it suggests an additional `#include` in a source file compared to the master branch, IWYU from the master branch reports the need of that `#include` in the associated header file, I think.

kimgr:
I just merged some ugly fixes for the broken packaging, please rebase so we can see the build pass.

kimgr:
Thanks!

PR #1606, opened 2024-08-30
Add mapping generator for libstdc++

kimgr:
If this seems to work out, the procedure for updating mappings should be:
```
# python3 ./mapgen/iwyu-mapgen-libstdcxx.py /usr/include/c++/11/ /usr/include/x86_64-linux-gnu/c++/11/ --lang=imp > libstdcxx.imp
# python3 ./mapgen/iwyu-mapgen-libstdcxx.py /usr/include/c++/11/ /usr/include/x86_64-linux-gnu/c++/11/ --lang=c++ > libstdcxx.cc
```
followed by some cutting and pasting to get the right lines into the right place (this can obviously be improved over time).

Poor mappings can now be improved in the python code instead of manually in the mapping tables, which makes tracking much easier.

And above all, users running a very new or very old GCC can simply regenerate their mappings in situ and use a generated `.imp` table or patch their IWYU source and rebuild with matching builtin mappings.

@firewave @alejandro-colomar @correaa Let me know what you think.


kimgr:
also cc @justanotheranonymoususer if you're interested in an example of generated mappings

alejandro-colomar:
I don't know python enough for reviewing this.  The changes to the mappings seem reasonable.

correaa:
I don't know python either, but more fundamental I don't understand how the library detects automatically the mappings from the libc++ code alone.

kimgr:
> more fundamental I don't understand how the library detects automatically the mappings from the lib std++ code alone.

It's possible to see some patterns in the library code, which lets you tell the difference between public and private headers (this is described in detail in the script doc comments).

Armed with that knowledge, you can say that an include from a public header to a private header implies a mapping from public to private. There are exceptions and special cases, and they are also encoded in the script. 

> What is the advantage of maintaining the python script vs the mappings themselves?
> (avoid redundancy for different versions?)

See PR description, but yes, pretty much. The mappings were originally generated by shell pipelines for GCC 4.6, then manually edited, then manually patched as issues were found. At this point the set of mappings on master is basically unmaintainable when faced with a new GCC version.

correaa:
> It's possible to see some patterns in the library code, which lets you tell the difference between public and private headers (this is described in detail in the script doc comments).
> 
> Armed with that knowledge, you can say that an include from a public header to a private header implies a mapping from public to private. There are exceptions and special cases, and they are also encoded in the script.
> 

Interesting, are those patterns general or only applicable to libstdc++?
In the unlikely case that they are general, they could be applied to any library and be part of the iwyu main program heuristics.
I suspect it is not that it may not be that simple.

Also, I found that concrete mappings for libstd++ (GNU) and libc++ (LLVM) are very different, for that matter.

> > What is the advantage of maintaining the python script vs the mappings themselves?
> > (avoid redundancy for different versions?)
> 
> See PR description, but yes, pretty much. The mappings were originally generated by shell pipelines for GCC 4.6, then manually edited, then manually patched as issues were found. At this point the set of mappings on master is basically unmaintainable when faced with a new GCC version.

I see.
Can each major version of GCC have its own mapping? GCC is somewhat unique and should be active by default, and as much careful work as needed should be put in each of the standard library major versions.

Also, would this cover the cases for libc++?

kimgr:
> Interesting, are those patterns general or only applicable to libstdc++?
> In the unlikely case that they are general, they could be applied to any library and be part of the iwyu main program heuristics.
> I suspect it is not that it may not be that simple.

No, those exact patterns are only relevant for libstdc++. Other libraries have other patterns, as encoded in the other mapgen scripts.

> Also, I found that concrete mappings for libstd++ (GNU) and libc++ (LLVM) are very different, for that matter.

Yeah, libc++ is a different library with difference conventions.

> Can each major version of GCC have its own mapping?

I was just thinking about that the other day. Assuming the patterns hold across GCC versions, it should be possible to install and generate mappings for every major GCC version, then keep version-keyed maps in `include_picker.cc`, and dynamically select mapping depending on which GCC/libstdc++ version is found on startup. All interesting followup work.

firewave:
The build failure looks like this: https://github.com/llvm/llvm-project/issues/99819

kimgr:
@firewave Thanks, yeah, I started trying to work around it, but there were many different files missing. It would be nice if CMake could be convinced to ignore missing files (or convert the error to a warning); very few of these errors are actually relevant to us.

firewave:
> @firewave Thanks, yeah, I started trying to work around it, but there were many different files missing. It would be nice if CMake could be convinced to ignore missing files (or convert the error to a warning); very few of these errors are actually relevant to us.

There is unfortunately a lack of documentation for the CMake integration. But I had a quick look at the file and it doesn't seem like any of this might be configurable. It also references CMake 2.8 so it seems like this stuff hasn't been touched in ages.

kimgr:
OK, I'm going to say this is ready for action now. Yolo, etc!

PR #1607, opened 2024-08-31
Updated public C++ header list for C++23

firewave:


firewave:
These commands are rather problematic. The one where I didn't update the list actually produces less output since several headers had been removed in C++20. So we would actually need to run it across all available standards and merge those results to have a completely list of all possible files.

kimgr:
Thanks!

PR #1608, opened 2024-09-02
More broken packaging workarounds

kimgr:


kimgr:
The missing dylibs are actually used by the compiler (and necessary for linking `include-what-you-use`), so this can't work. We'll have to wait for the package to repair.

PR #1610, opened 2024-09-04
Consider associated header desired includes

### LINKED ISSUE #1142: inconsistent source of `size_t`

bolshakov-a:
The statement that IWYU doesn't change associated headers is clearly wrong. When handling a main file, it should take into account what it suggests to do with its associated headers, not what those headers look like at the moment of the analysis.

It is assumed that the desired `#include`s of the associated headers have been calculated before handling the main file, as it was already assumed in the code.

This fixes #1142.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -828,13 +828,13 @@ bool DeclIsVisibleToUseInSameFile(const Decl* decl, const OneUse& use) {
 // might be "calculate minimal-ish includes". :-)  It populates
 // each OneUse in uses with the best #include for that use.
 // direct_includes: this file's direct includes only.
-// associated_direct_includes: direct includes for 'associated'
+// associated_desired_includes: desired includes for 'associated'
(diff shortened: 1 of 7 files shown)


kimgr:
I'm surprised this hasn't caused more trouble over the years. Thanks for the fix!

bolshakov-a:
Thanks to @firewave for pointing out.

PR #1630, opened 2024-09-25
Fix up LLVM APT packages in CI

kimgr:


kimgr:
@sylvestre I've been struggling with the Ubuntu Jammy snapshot packages for a while. This workaround is kind-of nuclear, but it seems to work. Does this give you any detail that might indicate a bug on the llvm-toolchain side? Is there a more obvious workaround? I think I convinced myself that I could also install a bunch of extra packages, and it would solve everything except the soname symlink issue. Grateful for any hints.

kimgr:
Replaced by #1632

sylvestre:
yeah, sorry, I will have a look, sorry

kimgr:
@sylvestre no worries, and thank you! I wish I understood the debian packaging enough to help with patches 🙂

sylvestre:
@kimgr I think it is now fixed :)

kimgr:
@sylvestre oh, cool! I'll try disabling these workarounds tomorrow and we'll see how it goes. Thanks!

PR #1633, opened 2024-09-29
Add mappings for macOS system headers

petere:
I have accumulated a bunch of mappings for macOS system headers. I wonder if there is interest in keeping this in IWYU upstream.  These are similar in spirit to for example `gcc.libc.imp`. Right now, these are just manually maintained, I haven't done any scripted analysis like the comments in some of the other mapping files suggest. If there is general interest, I can do that too. Let me know, please.


kimgr:
@petere Thanks! Yes, this would help improve the macOS experience, and I think it's a good idea to start with the libc as you've done. There's been a bunch of discussions recently on manual mappings, and I'm pretty firmly convinced that shipping generators for mappings is the way to go (so that people can easily refresh mappings for their local install of IWYU using the same tools we use upstream).

Mapping generators so far are in https://github.com/include-what-you-use/include-what-you-use/tree/master/mapgen, and I'd prefer to keep them all python for portability reasons. See also [this PR](https://github.com/include-what-you-use/include-what-you-use/pull/1606) for a recent attempt to do the same for GNU libstdc++.

I'm reluctant to bring in the macOS mappings as listed in this PR, because I suspect they only cover a practical subset, and will see a lot of churn before they reach more general usefulness. But maybe I'm overthinking?

Let me know what you think.

firewave:
> I'm reluctant to bring in the macOS mappings as listed in this PR, because I suspect they only cover a practical subset, and will see a lot of churn before they reach more general usefulness. But maybe I'm overthinking?

Such concerns are warranted.

But in this case it appears to be just a small amount of entries needed so I think a manual list might be okay for now.

I added macOS to the Cppcheck IWYU workflow and the entries added here seem to get rid of all the libc/system related suggestions. So I think this change is fine.

The next step would be to improve some of the libc++ mappings.

What is more concerning is that I am apparently seeing wildly different results compared to a Linux run on some files. And not having access to a local macOS system will make reducing these quite problematic. But that is obviously not in the scope of this PR.

firewave:
BTW the approval here was made by a spam account (according to the Wayback Machine it might have only existed for 4 days).

firewave:
See https://github.com/danmar/cppcheck/actions/runs/13135442079  for a job which includes macOS results.

> What is more concerning is that I am apparently seeing wildly different results compared to a Linux run on some files.

Disregard this. This applied to headers which had no source before and thus were never scanned.

kimgr:
A mapping generator for Apple Libc was merged in https://github.com/include-what-you-use/include-what-you-use/pull/1704, closing this.

PR #1644, opened 2024-11-01
Use expansion loc when the use macro defined in command line

BlackWaters:
This PR uses expansion loc as the use_loc to report for the declare when the spelling loc is in `<comman line>` (macro defined through command line -DXXX, which is not a reasonable location) to avoid some wrong removal.

### DIFF
--- a/tests/cxx/namespace_macro-direct.h
+++ b/tests/cxx/namespace_macro-direct.h
@@ -1,0 +1,16 @@
+
+//===--- namespace_macro-direct.h - test input file for IWYU --------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_NAMESPACE_MACRO_DIRECT_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_NAMESPACE_MACRO_DIRECT_H_
+
+#include "tests/cxx/namespace_macro-indirect.h"
+
+#endif
--- a/tests/cxx/namespace_macro-indirect.h
+++ b/tests/cxx/namespace_macro-indirect.h
@@ -1,0 +1,25 @@
+//===--- namespace_macro-indirect.h - test input file for IWYU ------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_NAMESPACE_MACRO_INDIRECT_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_NAMESPACE_MACRO_INDIRECT_H_
+
+#ifndef MYNS
+#define MYNS myns
+#endif
+
+namespace MYNS {
+
+void foo() {
+  // do something
+}
+
+}  // namespace MYNS
+
+#endif
--- a/tests/cxx/namespace_macro.cc
+++ b/tests/cxx/namespace_macro.cc
@@ -1,0 +1,32 @@
+//===--- namespace_macro.cc - test input file for IWYU --------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/namespace_macro-direct.h"
+
+// IWYU_ARGS: -I . -DMYNS=myns
+
+int main() {
+  // IWYU: myns::foo is...*namespace_macro-indirect.h
+  MYNS::foo();
+
+  return 0;
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/namespace_macro.cc should add these lines:
+#include "tests/cxx/namespace_macro-indirect.h"
+
+tests/cxx/namespace_macro.cc should remove these lines:
+- #include "tests/cxx/namespace_macro-direct.h"  // lines XX-XX
+
+The full include-list for tests/cxx/namespace_macro.cc:
+#include "tests/cxx/namespace_macro-indirect.h"  // for foo
+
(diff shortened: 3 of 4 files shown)


BlackWaters:
Fixed the previous build failure, thanks.

kimgr:
Thanks for the patch! I wonder if this is the right fix, though.

When reporting a use, IWYU uses `GetCanonicalUseLocation` to try and figure out for symbols in a macro if the file where the symbol is defined (spelling location) or the file where the macro is expanded (expansion location) is responsible for the particular macro. This is typically used to say that macro arguments belong to the "caller" of a macro, and symbols from inside the macro body belong to the "definer" of a macro. The file we decide on is the one suggested for inclusion.

I suspect this patch will attribute responsibility for `MYNS` to the .cc file itself (that's where it expands), and then `myns::foo` resolves to the header, which keeps it in place. You could run the test with env var `IWYU_VERBOSE` set to 5 or higher to see what is resolved where.

I'm thinking that with `-DMACRO=1`, the definer is not a file at all, so no include is necessary. I think we should probably add an explicit check earlier, that fully ignores uses of macros that are satisfied from the command line. OTOH, I'm not sure what the exact sequence is. Maybe the report is of `myns`, which is hidden behind `MYNS`, in which case your fix does the right thing.

Overall, for tests, please see if you can design them so they complain rather than say "has correct ...", because that will also produce localized diagnostics (in this case there would be a diagnostic on the line that uses `MYNS::foo`, containing the symbol name and location as IWYU sees them). There's some more tips in https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#force-diagnostics.

BlackWaters:
> I suspect this patch will attribute responsibility for MYNS to the .cc file itself (that's where it expands), and then myns::foo resolves to the header, which keeps it in place. You could run the test with env var IWYU_VERBOSE set to 5 or higher to see what is resolved where.

Yes, that's right. For my test case, the declare is in the right header but `GetCanonicalUseLocation` resolves the `use_loc` to `<command line>`, which doesn't make sense.

> I'm thinking that with -DMACRO=1, the definer is not a file at all, so no include is necessary. I think we should probably add an explicit check earlier, that fully ignores uses of macros that are satisfied from the command line.

Thanks for the advice! I will have a try later, or try to extract a better use_loc for the function use.

kimgr:
Ah, wait, I may have confused myself. I got hung up on the macro reporting, but that's handled here, just as I said -- macros defined in special files are ignored: https://github.com/include-what-you-use/include-what-you-use/blob/d2d092919f2774b5463e236e1ee9d56fb46ceb60/iwyu_preprocessor.cc#L923.

`GetCanonicalUseLocation` is only used for decl reporting, and does its dance to try and guess where the decl should be reported. So I think what's happening is that the `MYNS` -> `myns` expansion will be attributed to `<command line>` on master, and to `namespace_macro.cc` on your branch. But the true answer is really the header that declares `myns`.

So I'm curious who marks the header as needed. You can find out if you force your test to produce diagnostics using an indirect/direct header pair as is done in other tests.

kimgr:
Actually I'm still confused. `GetCanonicalUseLocation` computes the _use_ location, not the _decl_ location. And the use does happen in `namespace_macro.cc`. So I'm pretty sure your patch does the right thing. Sorry for all the back and forth.

Can you fix up the tests to have license headers and trailing newlines and indirect/direct headers to force diagnostics, and then I think we're good!

BlackWaters:
Thanks!  All issues have been fixed.

tests/cxx/namespace_macro-direct.h
```diff
@@ -0,0 +1,12 @@
+//===--- namespace_macro-direct.h -----------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
```
kimgr, line 10: Oops, sorry I didn't catch this before. We only use `#pragma once` for testing `#pragma once`, please add a real include guard (with the very annoying prefix `INCLUDE_WHAT_YOU_USE_TESTS_CXX_` + filename on symbol form). See other headers for examples.

tests/cxx/namespace_macro-indirect.h
```diff
@@ -0,0 +1,22 @@
+//===--- namespace_macro-indirect.h ---------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
```
kimgr, line 10: Include guard instead of pragma once

tests/cxx/namespace_macro.cc
```diff
@@ -0,0 +1,32 @@
...
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/namespace_macro-direct.h"
+
+// IWYU_ARGS: -I . -DMYNS=myns
+
+int main() {
+  // IWYU: myns::foo is defined in "tests/cxx/namespace_macro-indirect.h", which isn't directly #included.
```
kimgr, line 15: We usually use an abbreviated form: `// IWYU: myns::foo is...*namespace-macro-indirect.h`.

See the rest of the test suite for examples.

kimgr:
A few more nits, but overall looks great. See the license header checker too, the error message isn't super clear, but I think it's expecting `//===--- your-filename - test input file for iwyu ------------------------------===//` the ` - description` part is missing.

tests/cxx/namespace_macro-indirect.h
BlackWaters: Done

tests/cxx/namespace_macro.cc
BlackWaters: Done

tests/cxx/namespace_macro-direct.h
BlackWaters: Done

BlackWaters:
Ops... sorry for that mistake, I just thought `iwyu-check-license-header.py --add ... ` would add the right license for me.

I run `iwyu-check-license-header.py ...` for my test files, and no error occurs. I hope everything works fine this time.

Thanks for the review!

kimgr:
@BlackWaters `iwyu-check-license-header.py --add` should probably do that better :). Thanks!

Looks good to me now, thanks for contributing!

Issue #1654, opened 2024-11-17
[doc] Unclear documentation for `--cxx17ns`


carlosgalvezp:
The documentation only says:

```suggests the more concise syntax introduced in C++17```

However it does not specify "what syntax" it's referring to. It assumes that the user already knows that `ns` is a short form of `namespaces`. It's also unclear why IWYU would be related to nested namespaces, since IWYU is about header inclusions - one has to realize IWYU may also suggest forward declarations, and _those_ are the ones to be placed in the nested namespace.

Something like this would be clearer IMHO:

```
use C++17 nested namespaces when suggesting additions of forward declarations.
```

PR #1659, opened 2024-11-24
Clarify documentation for --cxx17ns

kimgr:
Issue #1654 righfully complained that the docs were opaque; somehow we forgot to mention it controls namespace declarations. Take the suggested doc string, and apply it to both --help output and manpage.


kimgr:
@carlosgalvezp Thanks for the suggestion, does this look good to you?

include-what-you-use.1
```diff
@@ -63,7 +63,7 @@ default.
 .RE
 .TP
 .B \-\-cxx17ns
-Suggest the more concise syntax for nested namespaces introduced in C++17.
+Use C++17 nested namespaces when suggesting additions of forward declarations.
```
carlosgalvezp, line 66: Ultra-nit: Here "Use" starts with capital U whereas in the other file it's small u.

carlosgalvezp:
Looks great, thank you!

include-what-you-use.1
kimgr: That's intentional, to vibe with the surrounding code

PR #1663, opened 2024-12-02
CMake: Only need a C++ compiler

EQt:
Makes it easier to configure a different compiler because you only need to set CXX environment variable.
Before you would also set a CC compiler.

See [CMake's `project` documentation](https://cmake.org/cmake/help/latest/command/project.html) for details.


kimgr:
It looks like LLVM's CMake system requires the C compiler. Don't know if there's a way to work around that?

EQt:
Indeed. Did not expect that. Then my proposal does not make sense, sorry.

PR #1676, opened 2025-01-21
Don't attempt to clear elaborated type duplicates

### LINKED ISSUE #1674: iwyu mangles anonymous unions
This is very similar to #1336 but is present as of iwyu 0.23 (i am unable to test latest master):

```cpp
int thing()
{
    union {
        int v1;
    } a, b;
    return 0;
}
```
```bash
$ include-what-you-use.exe stuff.cpp

stuff.cpp should add these lines:

stuff.cpp should remove these lines:
-   // lines 3-3

The full include-list for stuff.cpp:
---
```

Applying the suggestion would produce a syntaxically incorrect cpp, which is clearly undesirable
```cpp
int thing()
{
        int v1;
    } a, b;
    return 0;
}
```

The issue goes away if only a single variable is declared as a type of the union (i.e. if we change `union { .. } a, b;` to `union { .. } a;` or `union { .. } b;` ), and the test cases for #1336 (#1342 ) don't cover this multi-variable setup.

bolshakov-a:
It was assumed in #1232 that at most one elaborated type can "own" a tag declaration (i.e. none of possible following similar types redeclares that declaration), hence there should be no duplicates in `fwd_decl_map`. Even if that assumption is true, an `ElaboratedTypeLoc` may be handled more than once if it corresponds to different AST nodes. It happens when that type is used in a declaration of several variables.

Elaborated types should obviously not be removed in any case because they are part of user's main code and not of `#include` or fwd-decl lists.

This fixes #1674.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1790,7 +1790,8 @@ void ClearDesiredForSurplusIncludesOrForwardDeclares(ContainerType& container) {
     typename ContainerType::iterator v = ++container.lower_bound(k->first);
     typename ContainerType::iterator vend = container.upper_bound(k->first);
     for (; v != vend; ++v) {
-      v->second->clear_desired();
+      if (!v->second->is_elaborated_type())
+        v->second->clear_desired();
     }
   }
 }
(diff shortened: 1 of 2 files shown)


kimgr:
LGTM, thank you!

PR #1678, opened 2025-01-27
Don't report type sugar from implicit code

### LINKED ISSUE #1613: `<ext/type_traits.h>` suggested for no apparent reason
```cpp
#include <algorithm>
#include <string>
#include <vector>

class C {
public:
    const std::string &str() const;
};

void f()
{
    std::vector<const C*> f;
    std::vector<const C*> c;
    bool b = std::equal(f.begin(), f.end(), c.begin(), [](const C* t1, const C* t2) {
        return t1->str() == t2->str();
    });
}
```

```
#include <ext/type_traits.h>  // for __gnu_cxx::__enable_if

#include <ext/type_traits.h>  // for __gnu_cxx::__enable_if
#include <algorithm>          // for std::equal
#include <string>             // for std::operator==, std::basic_string, std::string
#include <vector>             // for std::vector
---
```

If I change `str()` to return `bool` instead it is no longer suggested.

Happens with 0.23dev and I can reproduce it with back to 0.20 (no earlier binaries to  test with).

bolshakov-a:
Skip any sugar type nodes in any implicitly generated code, because type sugar is used only when it is explicitly written.

The additional methods have been placed into `IwyuAstConsumer` and not into `InstantiatedTemplateVisitor` because the latter should not report type sugar in any case (it should not even have any methods doing this).

The change in `VisitType` has been made because an exact type from an exception specification (e.g. `ElaboratedType`) may be skipped now.

This fixes #1613.


kimgr:
Thank you!

PR #1679, opened 2025-01-31
Fix libc++ mappings to allow <iosfwd> as a valid public include

### LINKED ISSUE #1656: `<iosfwd>` suggested to be replaced with full stream header with libc++
```cpp
#include <iosfwd>

void f(std::ostream&);
```

```
$ include-what-you-use -Xiwyu --update_comments -stdlib=libc++ input.cpp

input.cpp should add these lines:
#include <ostream>  // for ostream

input.cpp should remove these lines:
- #include <iosfwd>  // lines 1-1

The full include-list for input.cpp:
#include <ostream>  // for ostream
---
```

```
$ include-what-you-use -Xiwyu --update_comments input.cpp

input.cpp should add these lines:

input.cpp should remove these lines:

The full include-list for input.cpp:
#include <iosfwd>  // for ostream
---
```

This started with 0.23.

kon72:
Fixes #1656

### DIFF
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -1183,23 +1183,23 @@ const IncludeMapEntry libcxx_include_map[] = {
 
 
     // For the following entries:
-    // cd llvm-project/libcxx/include ; find __fwd -type f -name "*.h" | sort | sed -e 's#__fwd/\(.*\).h#  { "<__fwd/\1.h>", kPrivate, "<\1>", kPublic },#'
+    // cd llvm-project/libcxx/include ; find __fwd -type f -name "*.h" | sort | sed -E 's#__fwd/(.*).h#  { "<__fwd/\1.h>", kPrivate, "<\1>", kPublic },# ; s#<(fstream|ios|istream|ostream|sstream|streambuf)>#<iosfwd>#'
     //
     // tweak hash, pair, subrange entries, and comment out bit_reference, get
     {"<__fwd/array.h>", kPrivate, "<array>", kPublic},
     //{"<__fwd/bit_reference.h>", kPrivate, "<bit_reference>", kPublic},
-    {"<__fwd/fstream.h>", kPrivate, "<fstream>", kPublic},
+    {"<__fwd/fstream.h>", kPrivate, "<iosfwd>", kPublic},
     //{"<__fwd/get.h>", kPrivate, "<get>", kPublic},
     {"<__fwd/hash.h>", kPrivate, "<functional>", kPublic},
-    {"<__fwd/ios.h>", kPrivate, "<ios>", kPublic},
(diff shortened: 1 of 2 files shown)


kimgr:
@kon72 Thanks!

The libstdc++ mappings always make me a little uncomfortable, as we don't have good test coverage for specific standard library impls (we can't guarantee that they're on the machine running the tests, etc).

This looks like all `__fwd/*.h` that happen to be mapped to the io subsystem headers will now _only_ be mapped to `<iosfwd>`? Won't that drop a bunch of bona fide includes of `<ostream>` and friends?

kimgr:
I tested this with:
```
// t.hh
#include <iosfwd>
void fofo(std::ostream& os);

// t.cc
#include "t.hh"
#include <ostream>

void fofo(std::ostream& os) {
  os << "fofo\n";
}
```
and it seems to be behaving nicely --

* with libc++ from llvm-16, there are no complaints on master or this PR branch
* with libc++ from llvm-17, there is indeed a suggestion to replace `<iosfwd>` by `<ostream>` in `t.hh` on master, but not on the PR branch

IWYU tests seem to be similarly broken with libc++ before and after the change. So I'm inclined to get this merged. Any other ideas for better real-world testing?

kimgr:
Also added `-stdlib=libc++` to the IWYU dogfood script, and it produces the same result before/after, so that seems sane.

firewave:
> Any other ideas for better real-world testing?

I came across these in the Cppcheck CI which runs IWYU with `libstdc++` and `libc++`. It also runs `clang-include-cleaner`.

It is a small codebase which only targets C++11 but sufficient as a smoketest.

Since I also want to run it on Windows in that job at some point I would need to build my own binary. I could probably also add optional support to build the latest development version.

kon72:
> This looks like all __fwd/*.h that happen to be mapped to the io subsystem headers will now only be mapped to <iosfwd>? Won't that drop a bunch of bona fide includes of <ostream> and friends?
>
> and it seems to be behaving nicely

Thanks for your check!
IWYU has default mappings which include `{ "<iosfwd>", kPublic, "<ios>", kPublic }` and `{ "<ios>", kPublic, "<ostream>", kPublic }`--that's why it still correctly handles `<ostream>` as a public header.

https://github.com/include-what-you-use/include-what-you-use/blob/463d1d300b32e0db332f90abe9dca44b8585a7ab/iwyu_include_picker.cc#L639-L658

firewave:
> IWYU has default mappings which include `{ "<iosfwd>", kPublic, "<ios>", kPublic }` and `{ "<ios>", kPublic, "<ostream>", kPublic }`--that's why it still correctly handles `<ostream>` as a public header.

They are also interchangeable - both are not a good thing IMO. See also #1540.

kimgr:
I've spent some time trying to worry about things that might break, but come up short :)

I think this is good to go. Thank you!

PR #1683, opened 2025-02-10
removed public-to-public mappings for streams

firewave:


firewave:
The failure of `cxx.test_no_char_traits` is expected. That is a behavior which was masked by these mappings and will lead to unnecessary stream includes being suggested. It also applies to `cxx.test_expl_inst_macro`.

`cxx.test_badinc` shows what is tried to being achieved here. That `<iosfwd>` is sufficient and the whole stream headers are not needed. Still experiences the same behavior as the other failures that unnecessary stream headers are being suggested.

firewave:
I rebased and the test failures look expected and reasonable. I will run it against Cppcheck later.

firewave:
This no longer seems to have any effect on Cppcheck: https://github.com/cppcheck-opensource/cppcheck/actions/runs/27867921277.

When I ran it a while ago I led to these changes: https://github.com/cppcheck-opensource/cppcheck/pull/6684.

firewave:
A shortcoming I see is that it suggests `<ostream>` for `std::ostringstream` when `<sstream>` already exists.

bolshakov-a:
@firewave, @kimgr, I think this can be rebased and merged now.

bolshakov-a:
> this can be rebased and merged now.

Actually, not. `operator<<(basic_ostream<char, traits>&, const char*)` is not mentioned in the `<ostream>` header synopsis, so `iwyu-mapgen-std-symbol.py` misses it, and IWYU requires `<ostream>` even if there is `<iostream>` included. That operator should probably be added into `HANDWRITTEN_MAPPING`.

It is also not fully clear to me if `<iostream>` is guaranteed to provide definitions of `std::char_traits` specializations, or `<string>` should be included for portability.

kimgr:
> It is also not fully clear to me if <iostream> is guaranteed to provide definitions of std::char_traits specializations, or <string> should be included for portability.

The kludge here seems to assume it implicitly: https://github.com/include-what-you-use/include-what-you-use/blob/b9853a1927e8a51032386ed7ce4a557d5f71d18a/iwyu_include_picker.cc#L392 -- if `<istream>` and `<ostream>` are allowed to provide for `std::char_traits`, it seems natural that `<iostream>` would include both `<istream>` and `<ostream>`, and thereby also provide it. But it's a bit sketchy, I agree.

bolshakov-a:
Asked here: https://discourse.llvm.org/t/do-i-o-stream-headers-provide-std-char-traits/91519.

bolshakov-a:
> That operator should probably be added into `HANDWRITTEN_MAPPING`.

I've opened #2092.

firewave:
> @firewave, @kimgr, I think this can be rebased and merged now.

Before merging this we need to add tests (seperately - not in this PR) which covers the shortcomings this exposes. Some were already mentioned here but it needs a proper look at the previous results (probably without the latest changes). I would like to see this merged so it makes the release but I am currently short on time for open source stuff.

bolshakov-a:
I hope the shortcomings can be eliminated by proper symbol mapping, but it is still not completely clear to me what it should look like: see [the aforementioned LLVM forum thread](https://discourse.llvm.org/t/do-i-o-stream-headers-provide-std-char-traits/91519).

PR #1684, opened 2025-02-10
[doc] Refresh the README install instructions

kimgr:
Now that e046d23571876e72719ef96a36ff0cb1cb2e64b6 has landed, and there's a complex but consistent policy for how to configure IWYU to use different resource dirs for built-in headers, try to actually explain it.

Issue #100 is now closed, so remove the link. It should no longer answer any questions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,31 +148,93 @@ This builds all of LLVM, Clang and IWYU in a single tree.
 
 ### How to install ###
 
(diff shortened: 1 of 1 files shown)


kimgr:
Rendered markdown here: https://github.com/kimgr/include-what-you-use/blob/install-readme/README.md#how-to-install

Issue #1614, opened 2024-09-07
`<iterator>` suggested when using `std::set::rbegin()`


firewave:
```cpp
#include <set>

void f()
{
    std::set<int> configurations;
    if (*configurations.rbegin()) {}
}
```

```
test.cpp should add these lines:
#include <iterator>  // for std::reverse_iterator

test.cpp should remove these lines:

The full include-list for test.cpp:
#include <iterator>  // for std::reverse_iterator
#include <set>       // for std::set
---
```

Happens with latest trunk. Can also reproduce back to at 0.20 (I have currently no older binaries available).

kimgr:
Fixed on master as of 76bc5422971b24174eb2f48780690c27dd60ed04.

Issue #111, opened 2015-06-15
iwyu suggests too much to add (inherited += operator)


vsapsai:
Originally reported on Google Code with ID 111

```
    // an example:
    // main.cc
    #include "mystring.hh"

    int main()
    {
        mystring s;
        s += "hello"; // triggers spurious include
        return 0;
    }
    //mystring.hh
    #include <string>
    struct mystring : public std::basic_string<char> {};

due to '+=' usage iwyu wrongly suggests pulling '#include <string>'

While this one does not trigger warning:

    // main.cc
    #include "mystring.hh"

    int main()
    {
        mystring s;
        //s += "hello"; // triggers spurious include
        return 0;
    }

Thanks!
```

Reported by `slyich` on 2013-09-18 15:13:09

Issue #138, opened 2015-06-15
 iwyu suggests too much to add: shared_ptr


vsapsai:
Originally reported on Google Code with ID 138

```
For the following example it suggests <memory> even though it is only used through the
typedef.

//sharedptr_test.cc
#include "myptr.h"
int main() {
    myptr p;
    *p; //no error without this line
}
//myptr.h
#include <memory>
typedef std::shared_ptr<char> myptr;

The problem seems to be with the return type of "*p". Even though it simply evaluates
to "char" the template expression seems to be confusing IWYU. The problem is with 3.4.1
and HEAD and with libstdc++ and libc++.

Originally added as further example to #111 but moved to new issue because it isn't
the same problem

Related to #135.
```

Reported by `roland@rschulz.eu` on 2014-05-30 02:58:29

kimgr:
Fixed on master as of 76bc5422971b24174eb2f48780690c27dd60ed04, scheduled for IWYU 0.24.

It didn't even take 10 years! :-)

PR #1686, opened 2025-02-26
Improve operand type determination

### LINKED ISSUE #1657: unnecessary `<functional>` suggested for `std::set` comparison
```cpp
#include <set>

void f()
{
    std::set<int> s1;
    std::set<int> s2;
    (void)(s1 == s2);
}
```

```
$ include-what-you-use -Xiwyu --update_comments input.cpp

input.cpp should add these lines:
#include <functional>  // for less

input.cpp should remove these lines:

The full include-list for input.cpp:
#include <functional>  // for less
#include <set>         // for set, operator==, allocator
---
```

bolshakov-a:
clang inserts a "no-op" implicit cast nodes under the operator call expression node for some reason, and it hindered the determination of the written operand type. It caused bugs like e.g. when `std::set<int>` was compared with another one, it was considered as if its default template arguments (`std::less<int>` and `std::allocator<int>`) were explicitly written. `ReplayClassMemberUsesFromPrecomputedList` method reports all of them in that case.

`IgnoreParenImpCasts()` has been called to avoid this, as it has been done already in several other places where `GetTypeOf` function is used.

This fixes #1657.


kimgr:
Wow, high-impact patch! Looks like this also fixes https://github.com/include-what-you-use/include-what-you-use/issues/111

bolshakov-a:
That's good! Should I add a corresponding test case?

firewave:
This also fixes #1614.

firewave:
#138 is fixed by this as well.

kimgr:
> That's good! Should I add a corresponding test case?

If you have some spare energy to reduce them, that would be great.

bolshakov-a:
The reason of #111 and #138 is an implicit derived-to-base conversion which caused the target (base) type to be reported instead of the derived one before this patch. The reason of #1614 is hiding a type alias and is covered in my test cases. The exact case of #1657 is _not_ covered because I don't want to pull standard library headers in the test to launch "precomputed list" logic.

kimgr:
Feels like it would be good to have coverage for the derived-to-base conversion, that comes up all the time.

#138 looked more like aliasing to me, so that's presumably already covered?

And yeah, good call on avoiding standard library headers, that's not worth the additional coverage.

bolshakov-a:
> https://github.com/include-what-you-use/include-what-you-use/issues/138 looked more like aliasing to me, so that's presumably already covered?

Type aliases and class inheritance are different things and are handled by IWYU, obviously, differently. I don't think there may be an "author intent" policy for the inheritance, because a derived class author should include the base class declaration, there is no choice here. However, all the mentioned issues have the common cause: an `ImplicitCastExpr` AST node which hided the reasonable type.

tests/cxx/overloaded_operator.cc
```diff
@@ -0,0 +1,45 @@
...
+
+#include "tests/cxx/overloaded_operator-direct.h"
+
+void Fn() {
+  ProvidingAlias t1, t2;
+  // Test that IWYU determines the first argument type correctly
+  // as the providing alias and doesn't suggest the underlying type.
+  (void)(t1 == t2);
+  ProvidingRefAlias t3 = t1;
+  // Test that the extra parentheses doesn't hide the type as-written, and IWYU
```
kimgr, line 22: typo: `extra parentheses doesn't` -> `don't`

tests/cxx/overloaded_operator.cc
```diff
@@ -0,0 +1,45 @@
...
+  // doesn't suggest the underlying type, again.
+  (void)((t3) == t2);
+
+  // IWYU: Derived is...*indirect.h
+  Derived d1, d2;
+  // Test that IWYU ignores devived-to-base implicit conversion and reports
```
kimgr, line 28: typo: `devived-to-base` -> `derived-to-base`

kimgr:
A few typos, but looks good to me with them fixed! Thank you!

firewave:
> #138 looked more like aliasing to me, so that's presumably already covered?
> 
> And yeah, good call on avoiding standard library headers, that's not worth the additional coverage.

I ran 0.23 and 0.24dev against Cppcheck and looked at the diff and I can see the difference in the output:

```diff
@@ -2313,7 +2311,7 @@
 #include <cstddef>        // for std::size_t
 #include <functional>     // for std::function
 #include <map>            // for std::map
-#include <memory>         // for std::shared_ptr, std::__shared_ptr_access, std::operator!=, std::operator==
+#include <memory>         // for std::shared_ptr, std::operator!=, std::operator==
 #include <string>         // for std::string
 #include <unordered_map>  // for std::unordered_map
 #include <utility>        // for std::move
```

So that somewhat seconds as some kind of (manual) system test. I will try to do such comparisons for future releases as well.

bolshakov-a:
I'm afraid, however, that reporting an internal library stuff is unavoidable as IWYU sensitivity would raise. Clang sometimes loses the exact user-written type when building AST.

kimgr:
> I'm afraid, however, that reporting an internal library stuff is unavoidable as IWYU sensitivity would raise. Clang sometimes
> loses the exact user-written type when building AST.

Sorry, I didn't follow this. Do you mean this is a problem with this patch, or just a general concern? What do you mean by "IWYU sensitivity would raise"?

I would like to merge this, but if you think further refinements are necessary, I can wait.

bolshakov-a:
No, not with this patch, but in general. A saw such internal stuff suggestions on my local branches with other fixes, related to templates, in particular, and had no idea how to avoid them.

PR #1687, opened 2025-03-01
Improve C-style array handling

bolshakov-a:
The code added to `CanForwardDeclareType` avoids reporting array element types in cases like declaring external array variables. The change in `ReportTypeUseInternal` makes them reported when used actually.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -175,6 +175,7 @@ using clang::ASTConsumer;
 using clang::ASTContext;
 using clang::ASTFrontendAction;
 using clang::ArraySubscriptExpr;
+using clang::ArrayType;
 using clang::Attr;
 using clang::BinaryOperator;
 using clang::CXXBaseSpecifier;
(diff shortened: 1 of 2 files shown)


bolshakov-a:
The `TODO` in the test is easy to fix (it is just needed to ignore `ParenType` nodes). I've decided, however, to separate it to another PR for clarity.

I'm not sure whether it is worth to test recursion inside `ReportTypeUseInternal`, i.e. write arrays of "providing" and "non-providing" type aliases. Some extra test files would be required for that.

kimgr:
> I'm not sure whether it is worth to test recursion inside ReportTypeUseInternal, i.e.
> write arrays of "providing" and "non-providing" type aliases. Some extra test files
> would be required for that.

I don't think that's necessary at this point.

Thanks, LGTM!

bolshakov-a:
FWIW, another approach without extending `CanForwardDeclareType` might be to add `VisitArrayType` calling `set_in_forward_declare_context(true)`, but that may be somewhat confusing, as it seems to me.

kimgr:
@bolshakov-a Aha, I was actually wondering a little why array reporting was handled in `ReportTypeUseInternal` and not in a `VisitArray...` override. But the forward-declarabilty didn't trigger me.

Instinctively, I like the forward-declare context (I've been toying with the idea to make it a bitmask, so we can have context information for e.g. `ignore all types`, `traversing a node of type X`, etc.), but `CanForwardDeclareType` appears to have been more appreciated:
```
  // TODO(csilvers): get rid of in_forward_declare_context() and make
  // this the canonical place to figure out if we can forward-declare.
```
I'm fine with it as-is, anyway.

PR #1688, opened 2025-03-01
[clang compat] Handle new HLSL cast kinds

kimgr:
Two new cast kinds have been added in Clang:

CK_HLSLAggregateSplatCast:
https://github.com/llvm/llvm-project/commit/4d2d0afceeb732a5238c2167ab7a6b88cc66d976

CK_HLSLElementwiseCast:
https://github.com/llvm/llvm-project/commit/3f8e2802069aabe4384ecd4575d50fd4457dae51

I can't see that they would affect IWYU analysis, so just handle them silently to placate -Wswitch.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1869,6 +1869,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
                                               // the full type.
       case clang::CK_HLSLArrayRValue:
       case clang::CK_HLSLVectorTruncation:
+      case clang::CK_HLSLElementwiseCast:
+      case clang::CK_HLSLAggregateSplatCast:
         break;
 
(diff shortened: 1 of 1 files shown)


kimgr:
Note that these cast kinds were added after the Clang 20 branch on the LLVM side, so adding these is a breaking change vs. Clang 20. I will hold off merging (and live with the warnings) until we've branched `clang_20` for IWYU 0.24.

firewave:
FYI https://apt.llvm.org/ has not yet introduced the 20 branch.

firewave:
> FYI https://apt.llvm.org/ has not yet introduced the 20 branch.

Turns out the branch already existed and it just wasn't announced.

kimgr:
IWYU 0.24 is now released and `clang_20` has been created without this patch. Merging to master now.

PR #1692, opened 2025-03-08
Denote rvalue-referenced types as fwd-declarable

bolshakov-a:
IWYU should handle rvalue references as lvalue ones in most cases.


bolshakov-a:
Discovered while working on type traits.

kimgr:
Looks good, thanks!

Issue #1578, opened 2024-07-27
forward declaration not suggested for type returned from function


firewave:
`test.h`
```cpp
#include "s.h"

S f();
```

`s.h`
```cpp
class S {};
```

Analyzing a file including `test.h` will not suggest a forward declaration which would be sufficient.

But if you are already using a forward declaration instead of the include it will report that everything is correct.

If the type is used as a function parameter it will suggest the forward declaration.

I am quite surprised but apparently it is legal: https://stackoverflow.com/questions/47817012/incomplete-types-as-function-parameters-and-return-values.

I am using include-what-you-use 0.22 on Manjaro Linux.

bolshakov-a:
This is the default IWYU policy. Please read the docs: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhatIsAUse.md

I think it would be nice to add command-line flags to switch these policies off.

firewave:
> This is the default IWYU policy. 

Interesting. This is basically the "mission statement" I was referring to in https://github.com/include-what-you-use/include-what-you-use/pull/1568#issuecomment-2224983328.

> I think it would be nice to add command-line flags to switch these policies off.

I second that.

I doubt that each author will properly express their intend. And even if that is covered in code reviews there might still be cases slipping through.

And with the function parameter case suggesting the forward declaration, depending on when you started to use include-what-you-use and the evolution of the code, you might end with either of the cases.

I think it would be nice to be able to enforce a certain policy for consistency across the code.

firewave:
> I think it would be nice to be able to enforce a certain policy for consistency across the code.

I think there should be three different policies: `permissive` (current / default), `incomplete` and `full`.

firewave:
> I think there should be three different policies: `permissive` (current / default), `incomplete` and `full`.

It could be more complex than that though as we currently have two different policies in place. We allow either if any of them is already in place and suggest the full type.

So there would be more cases:
- `incomplete` - enforce usage of incomplete types
- `full` - enforce usage of full types
- `suggest-incomplete` - do not enforce and use incomplete types for suggestions (if necessary)
- `suggest-full` - do not enforce and use full types for suggestions (the current default)

PR #1693, opened 2025-03-09
Make "assignable" trait handling more fine-grained

### LINKED ISSUE #1615: include suggested when forward declaration is (was) sufficient
I reduced this from much bigger code (`lib/pathanalysis.cpp` in the `Cppcheck` project) and the warning about the unnecessary include in `test.cpp` is intentional. It is probably possible to reduce it more but most changes I tried to caused the issues to disappear.

`test.cpp`
```cpp
#include "test.h"
#include "inc.h"
```

`test.h`
```cpp
#include "types.h"

#include <functional>
#include <list>
#include <string>

class Token;

struct PathAnalysis {
    struct Info {
        const Token* tok;
        ErrorPath errorPath;
    };

    void forward(const std::function<void(const Info&)>& f) const;

    Info forwardFind(std::function<bool(const Info&)> pred) const {
        Info result{};
        forward([&](const Info& info) {
            result = info;
        });
        return result;
    }
};
```

`inc.h`
```cpp
#include "token.h"
```

`token.h`
```cpp
class Token{};
```

`types.h`
```cpp
#include <list>
#include <string>
#include <utility>

class Token;

using ErrorPathItem = std::pair<const Token *, std::string>;
using ErrorPath = std::list<ErrorPathItem>;
```

```
#include "token.h"     // for Token

- class Token;  // lines 7-7

#include "token.h"     // for Token
#include "types.h"     // for ErrorPath
#include <functional>  // for std::function
#include <list>        // for std::list
#include <string>      // for std::basic_string
---

- #include "inc.h"  // lines 2-2

#include "test.h"
---
```

This started with 0.23dev. I cannot reproduce it with 0.22.

```

- #include <string>  // lines 5-5

#include "types.h"     // for ErrorPath
#include <functional>  // for std::function
#include <list>        // for std::list
class Token;  // lines 7-7
---

- #include "inc.h"  // lines 2-2

#include "test.h"
---
```

Here's the files in an archive to ease reproducing it: [example.zip](https://github.com/user-attachments/files/16919779/example.zip)

bolshakov-a:
The test refactored and extended.

This patch is not consistent in what should be done with _possible_ conversions. That means: if the types don't satisfy the trait at the time of analysis, should they be nevertheless complete to prevent a UB if some conversion operators appear in the future? It has been generally answered "yes" for this patch, but not for the pointer assignments. The cases when a type starts inheriting another type, previously unrelated, seem to be rare (or not so rare?). However, this has not been applied to reference assignments, when the full rhs class type is always required. This is because I'm really not sure which
approach should be preferred... But I think IWYU hardly can (and should) prevent any possible UB after user's code future refactoring, especially in cases like described in [here](https://github.com/include-what-you-use/include-what-you-use/pull/745#issuecomment-2697923215), otherwise it might require completely unrelated types.

Nevertheless, this fixes #1615.


iwyu.cc
```diff
@@ -2043,21 +2043,87 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    // important not to suggest a full type when it is redundant for computing
+    // the correct value.
+    switch (expr->getTrait()) {
+      case TypeTrait::BTT_IsSame:
+        return true;
+      case TypeTrait::BTT_IsAssignable:
+      case TypeTrait::BTT_IsNothrowAssignable:
+        // TODO(bolshakov): require complete argument types here when
+        // CanForwardDeclareType stops doing it.
+        if (RemoveReference(lhs_type)->isRecordType()) {
```
kimgr, line 2069: I think you want to check for `TagType` here to also cover unions. Or `isRecordType() || isUnionType()`.

iwyu.cc
```diff
@@ -2043,21 +2043,87 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+          // type.
+          ReportTypeUse(CurrentLoc(), rhs_type, DerefKind::RemoveRefs);
+        }
+        return true;
+      case TypeTrait::BTT_IsTriviallyAssignable: {
+        // Trivial assignment doesn't involve any user-defined operators but may
+        // involve a derived-to-base conversion.
+        const Type* lhs_deref_type = RemoveReference(lhs_type);
+        if (lhs_deref_type->isStructureOrClassType()) {
+          // Rhs may be the same or derived class. In both cases, its type info
```
kimgr, line 2100: All-upper `RHS`, please.

kimgr: Hmm, or maybe that makes the other all-lower `rhs` look weird. Maybe compromise with `The rhs...`?

iwyu_ast_util.cc
```diff
@@ -1683,6 +1689,28 @@ vector<const Type*> GetCanonicalArgComponents(
   return res;
 }
 
+bool IsReferenceToModifiableLValue(const Type* type) {
+  return type->isLValueReferenceType() &&
+         !type->getPointeeType().getQualifiers().hasConst();
+}
+
+bool DerivedToBasePtrConvertible(const Type* derived_ptr_type,
```
kimgr, line 1697: Can we fit and `Is` prefix here, so `IsDerivedTo...`?

iwyu.cc
```diff
@@ -2043,21 +2043,87 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    if (expr->getNumArgs() == 2)
+      rhs_type = expr->getArg(1)->getType().getTypePtr();
+
+    // Most of the binary type traits require the types to be complete types or
+    // arrays of unknown bound for compilable code. It means that compilers
+    // accept pointers or references to fwd-declared types. However, in certain
+    // cases it may cause a UB (C++20 [meta.rqmts]p5) because the full
+    // pointed-to type is needed to determine the trait value correctly, so it's
+    // better for IWYU to require the full type info in such situations. It is
+    // important not to suggest a full type when it is redundant for computing
```
kimgr, line 2060: I don't understand the last sentence, particularly "redundant for computing the correct value"?

iwyu.cc
kimgr: Wait, what? `RecordType` also covers unions? I was under the impression that unions were the difference between `Tag` and `Record` types. But it doesn't look like it, so I guess this works fine.

firewave:
After all the comments have been addressed I will do a comparison run.

iwyu.cc
bolshakov-a: Yes, `RecordType` covers classes, structs, and unions, and `TagType` also covers enums.

iwyu.cc
bolshakov-a: The lowercase abbreviations seem to be more common in variable names ([an example in the C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c83-for-value-like-types-consider-providing-a-noexcept-swap-function)), but it may because the uppercase ones would look like macro names...

iwyu.cc
bolshakov-a: I mean: in cases like `std::is_assignable_v<Token*&, Token*>`, the result is always `true` regardless of what `Token` actually is (like in #1615). No full type info is needed.

bolshakov-a:
@kimgr, @firewave, @selat, I would like to know your opinion about whether IWYU should suggest the full type info in cases like `std::is_assignable<Class1&, Class2&>` just because a conversion operation _may_ appear _in the future_, or `Class2` _may_ start to inherit `Class1`? Or this is not very important? At least for future PRs to be more consistent.

firewave:
I think this is "author intent" vs "IWYU intent". Similar to other cases where incomplete and full policy are both allowed but not enforced, it should be lenient (until there is a switch to control the policy) to go for the incomplete if none is currently done and keep the full intact if that is already used. But I somehow have the feeling this is the actual crux here.

I hope this made any sense at all. I am not really a person to document on issues on this level.

bolshakov-a:
Thanks! However, the "author intent" analysis has its costs. I'd say that it causes much of the IWYU complexity. It may be not very difficult to implement only on the user's side, though, as it should be with type traits.

However, I'm not sure if all of this _really_ should be done. I don't still realize which bad things would happen in real-world code if one just removes `VisitTypeTraitExpr`.

firewave:
So maybe it is time to toss anything about "intent" out of the window and enforce an actual policy. See https://github.com/include-what-you-use/include-what-you-use/issues/1578#issuecomment-2283764995 for a suggestion.

iwyu.cc
kimgr: Yeah, let's go with `The rhs may be the same...`, to avoid the casing challenge.

kimgr:
@bolshakov-a 
> I would like to know your opinion about whether IWYU should suggest the full type info in cases like `std::is_assignable<Class1&, Class2&>` just because a conversion operation may appear in the future, or Class2 may start to inherit Class1? 

It's a good question. I feel like it would be more consistent not to assume the worst and instead leave that as a lurking bug (maybe with a little warning on high verbosity level).

Do I understand you correctly that this is the risky behavior:

* T0: `Class1` and `Class2` are unrelated and not assignable
* T1: We run IWYU, and it suggests that forward-declares are enough for `Class1` and `Class2` when used with `is_assignable`
* T2: We make a change to `Class1` or `Class2` so there's an implicit conversion which makes them assignable. But that will only fire with `is_assignable` if `Class1` and `Class2` are complete types. If not, behavior is undefined?

And this is only a problem if they are lvalue references, right? Otherwise the complete type for `Class1` and `Class2` would already be required?

I wonder if longer-term we could find a way to make this part of Sema reusable: https://github.com/llvm/llvm-project/blob/7f415e444197025180d786c5e212f13b738179cb/clang/lib/Sema/SemaExprCXX.cpp#L5746, so we don't end up reimplementing it poorly?

iwyu.cc
bolshakov-a: Sorry, I forgot. We decided to use the word "redundant" when an `#include` is already somewhere in the tree, didn't we? If so, "unnecessary" fits better here, indeed.

bolshakov-a:
> I feel like it would be more consistent not to assume the worst and instead leave that as a lurking bug

Do you mean that the full type-info suggestions should be avoided whenever possible (without introducing an UB in a present user's code)? Or just keep my variant of the handling?

> If not, behavior is undefined?

Yes, correct. But it seems that in real-world code a type-trait check should be followed by an explicit use of the checked property, and IWYU should report it anyway, so IWYU doesn't need the explicit type-trait handling with all of its complexity...

> And this is only a problem if they are lvalue references, right? Otherwise the complete type for Class1 and Class2 would already be required?

Lvalue and rvalue references, yeah. And, more likely, pointers to members need a similar consideration. The standard states that the types should be complete, cv-qualified `void`, or unbound arrays (note that pointers and references are always complete, even when pointed-to type is incomplete).

> I wonder if longer-term we could find a way to make this part of Sema reusable

Of course, it would be good to reuse the compiler logic somehow instead of reimplementing it... However, as regards `CheckConvertibilityForTypeTraits`, it doesn't fully correspond to the "assignable" property.

kimgr:
> > I feel like it would be more consistent not to assume the worst and instead leave that as a lurking bug
>
> Do you mean that the full type-info suggestions should be avoided whenever possible (without introducing an UB in a present
> user's code)? Or just keep my variant of the handling?

I'm not sure :-)

As I understand it, the new logic in this patch will never (barring mistakes) suggest a change that introduces UB, right?

So the concern is that the user might make changes to the types involved in the traits expression in a way that the traits expression will now require the complete type, but they don't notice? If that's the case, I'm inclined to think that's the user's problem, and mostly unrelated to IWYU. But I guess there's some subtlety that I'm missing.

bolshakov-a:
Yeah, correct. But note that a stricter policy leads to a simpler implementation, in general. If we'd decide to report any pointed-to type which is different from the lhs pointed-to type (to prevent further user's bugs, as we could say), then I would have to remove `isDerivedFrom` check from `IsDerivedToBasePtrConvertible` and keep there only the qualifier check (and rename the function to `CouldBeDerived...`). If we'd decide to report only the actually needed types at the moment of analysis, then some more code should be written to do the same derived-to base check for references, and to scan the lhs for a suitable assignment operator, etc. (However, the latter should anyway be done some day to find that same "3rd type"...) Indeed, the simplest version from the master branch causes #1615 because it is exceedingly strict.

> However, as regards CheckConvertibilityForTypeTraits, it doesn't fully correspond to the "assignable" property.

Moreover, `CheckConvertibilityForTypeTraits` and similar parts of `Sema` code don't say which types should be complete but just compute the trait value.

kimgr:
@bolshakov-a So the tradeoff seems to be between:

* Require too much in general, at the cost of unwanted changes to user code and longer compile times
* Require too little in some cases, at the cost of potentially setting user code up for UB

In my mind, the only case we want to avoid is where IWYU will _introduce_ UB by suggesting that we remove includes that in fact are necessary for traits requirements. It's not fully clear to me if any of the variants have that behavior... But it seems to me the [three-part conversion example](https://github.com/include-what-you-use/include-what-you-use/pull/745#issuecomment-2697923215) might trigger something with the patch in this PR?

bolshakov-a:
> It's not fully clear to me if any of the variants have that behavior...

None of the suggested alternatives (except the complete removal of the handler) is supposed to introduce an UB.

> But it seems to me the https://github.com/include-what-you-use/include-what-you-use/pull/745#issuecomment-2697923215 might trigger something with the patch in this PR

`C` is reported neither on my branch nor on the master. It is not so easy to fix it in IWYU. Also note that attempts to prevent _possible future_ UBs would lead to requiring `C` full type even if `C` would not have a conversion ctor from `B`. And it doesn't look like a good idea. However, it is still a synthetic example. I don't know if it matters for any real-world code.

bolshakov-a:
> I think it's fine if the heuristics err in the favor of complete types.

So, would it be better to remove the check that `derived` is actually derived from `base` from `IsDerivedToBasePtrConvertible`?

iwyu.cc
```diff
@@ -2043,21 +2043,87 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+          // Scalar types can be assigned to only if they're modifiable lvalues.
+          const Type* lhs_deref_type = RemoveReference(lhs_type);
+          const Type* rhs_deref_type = RemoveReference(rhs_type);
+          // For derived-to-base pointer conversion to be allowed, the derived
+          // type should be complete.
+          // TODO(bolshakov): what about member pointers?
+          if (IsDerivedToBasePtrConvertible(rhs_deref_type, lhs_deref_type)) {
+            ReportTypeUse(CurrentLoc(), rhs_type, DerefKind::RemoveRefsAndPtr);
+            return true;
+          }
```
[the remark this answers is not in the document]
bolshakov-a: It could, but I think the logic is more clearly expressed in this form, because it is rather a special case vs. other alternatives than the two complementary alternatives.

kimgr:
> > I think it's fine if the heuristics err in the favor of complete types.
>
> So, would it be better to remove the check that derived is actually derived from base from IsDerivedToBasePtrConvertible?

I've been a little busy the past weeks so I keep coming back to this patch for 10 minutes at a time, and I honestly have not been able to get my head around larger-scale effects of one policy or the other.

I'm happy to merge this as-is based on a superficial look at the tests -- it seems to be doing the right thing, and is an improvement on the status quo -- and I guess we can figure out more edge-case problems as they surface.

I have to start preparing a 0.24 release for Clang 20 soon, so it would be nice to land this first.

bolshakov-a:
Ok, let's proceed with it as-is. I don't think all these cases are very important, and even the complete removal of the trait handler would probably go, as I've already mentioned.

bolshakov-a:
> I've been a little busy the past weeks so I keep coming back to this patch for 10 minutes at a time

That's pretty sad that such a unique and important (IMHO) project as IWYU has no paid maintainer...

firewave:
I did  a comparison run and my reported issue has indeed bene fixed, and I see no other changes. 👍

> That's pretty sad that such a unique and important (IMHO) project as IWYU has no paid maintainer...

Welcome to open source. I think not even clang-tidy has a dedicated paid contributor. And even if you have paid people their contributions might still be erratic. Not every project is like DXVK which basically went into overdrive when somebody started putting a wage onto it...

bolshakov-a:
I'm not sure about clang-tidy specifically, but the main clang (as well as gcc, I think) maintainers definitely work on it as part of their full-time job. That's why it succeeded in competition with commercial compilers (at least until Google, particularly Richard Smith, stopped active contributing). Communism is a utopia, as the history of my country showed, but people and organizations can develop mutually beneficial cooperation and make free products for anyone...

Sorry about turning this PR into a useless chatter.

kimgr:
I think IWYU has always been a labor of love. csilvers, who started it, just happened to be Google's CTO, and could spend some company time on it.

I have no affiliation with Google, never have, and my current day job does not use C++, so I just look after IWYU because it makes me happy. I'm not so sure paid maintainership would make me happier or better equip me to understand Clang and IWYU. Maybe I just need to retire :o)

bolshakov-a:
> csilvers, who started it, just happened to be Google's CTO, and could spend some company time on it.

And it looks suspicious that he left commercial programming in 2012, just after a couple of years trying to write IWYU...

> I just look after IWYU because it makes me happy

One can envy you!

Just some thoughts appear occasionally: maybe, I'm terribly wrong about IWYU importance, and if no one currently wants to pay for it (money or time, which is nearly the same) then probably no one really needs it... The company I worked for didn't want to give me even unpaid time for working on IWYU.

Ah, nevermind. Let's do what one wants or thinks he should do.

firewave:
> if no one currently wants to pay for it (money or time, which is nearly the same) then probably no one really needs it...

That is the crux of open source. Nobody wants to buy for it but everybody wants to profit from it.

> maybe, I'm terribly wrong about IWYU importance,

There's several reasons you might want to apply to your project. So importance may vary.

One might be build times and that might be an important issue to tackle. I read something crazy at some point in the LLVM issue tracker. It was someone who worked in the gaming industry and the person stated that nobody is using the actual STL includes because they are way too heavy and they are using custom thin layers to make compilation fast and only bring the actual library in later. I have no idea how that might even be working with all the template madness and unfortunately I do not have link to it either. But that shows IWYU is tackling things which are very important but it is not only about "excessive includes" but also "includes which are excessive". LLVM is actually tackling both by making their standards library very granular and light on transitive includes. So unsurprisingly there are a lot of cogs in play in just this one scenario.

And now we are waaaaaay off track ...

bolshakov-a:
Even worse, if you refactor one of your headers or just realize that it doesn't actually need some of its external library includes, you often cannot just remove them because you find out that a half of your project depends on it transitively. If I remember correctly, it took about a month (!) for our team to clean up one of such headers. The IWYU tool didn't correctly work with it those days. (Surprisingly, it didn't give a lot of compile time improvement; [jumbo build](https://en.wikipedia.org/wiki/Unity_build) helped much better, but it looks like a kind of ugliness for me.) Due to this, the company included the IWYU principle into an obligatory style guide. But then I noticed how extremely inefficient my work became when figuring out by eyes which includes are still needed in a several-thousand-lines-of-code file after removing a couple of the lines. So I think it must be automated.

But, maybe, it would be better to rewrite all the code into C++20 modules, or just stop writing C++ completely, [as the previous US administration suggested](https://www.theregister.com/2024/11/08/the_us_government_wants_developers/)... Personally, I've left the commercial C++ development and don't want to get back, at least until IWYU would work fine.

firewave:
> If I remember correctly, it took about a month (!) for our team to clean up one of such headers.

The amount required could be quite arbitrary as it highly depends on your design. If it affects APIs it might take years until you can wrap it up, even tough it might just be a few hours of actual work.

> it didn't give a lot of compile time improvement

In conjunction with something like `ccache` it might still help. It might also help with the run-time of something like static code analysis tools which might be way less efficient than a compiler.

> maybe, it would be better to rewrite all the code into C++20 modules, or just stop writing C++ completely

Don't get me started on modern C++.

> at least until IWYU would work fine.

Because of that some people just drop these tools because they are not worth the time spend on maintaining the integration or dealing with false positives.

And for others that's how they start committing boatloads of their time for free to an open source/pet project. That's how I got so involved in Cppcheck. I wanted it to be usable and for years I am basically just trying to wrap up all my open work but I keep getting involved in bigger or other tasks because everything is convoluted or fundamentally broken ...

Having it clean running against Cppcheck seems to be getting close if the current pace of improvements is kept up (maybe another year or two). At least of the CLI code. For the GUI code I finally need to start looking into adjusting the mapping generator.

bolshakov-a:
> maybe another year or two

Sounds good. (If you really believe it.)

I just wanted to say that a couple of people working in a non-obligatory way in spare time is seemingly not enough even for IWYU to escape from the alpha-quality stage in the near future. (It should be even more relevant for comprehensive static analysis tools like Cppcheck.) But if the mankind doesn't consider this as a problem, it probably isn't.

firewave:
> for IWYU to escape from the alpha-quality stage

A big problem in perception of the quality and the adaption of IWYU is that depending on your platform you might be stuck with a very outdated version (same with Cppcheck) - including the Compiler Explorer integration (see #1358). And there's also the issue that it might break when the underlying default Clang version is being updated and you are required to manually specify the proper includes.

Even on ArchLinux-based systems where it is in AUR you are dependent on your distro already having moved to the matching Clang version. That might lead to failing updates of the package and might also lead to people dropping the latest version.

Another pain point may also understanding how it decides what to add/remove, The hints are useful in most cases but for the rest of them it is hard to understand where those symbols are coming from (sometimes I wish it would just tell me the line of code it thinks it needs that include for). And there's the issue with bigger third-party libraries (Qt *cough* Boost *cough*).

And without proper feedback it is also hard to improve things. That also requires another group of dedicated people (another common issue across the open source space).

That all makes it feel like it might not be production-ready although it is a very solid piece of software with people involved that really know what it is all about (I wish I had that knowledge but to be honest I do not long or intend to achieve that). It took me quite a while until I got a hold on how to do that.

But IWYU definitely did make an impact. LLVM started their own `clang-include-cleaner` and they adopted the existing IWYU annotations. And there have been discussions and conses when those were extended. So it is not like nobody cared.

PR #1699, opened 2025-04-05
Bump version to 0.25 on master

kimgr:


kimgr:
Build failure seems to be due to broken packaging for LLVM 21, so merging with fingers crossed.

PR #1702, opened 2025-04-05
Minor fixes to man page

aaronpuchert:
For the experimental flags, we distinguish between (bold) terminals, italic non-terminals, and regular syntax. We also refer to the non- terminal in the text.

In the section of mapping files, we move the quotes around "visibility" up to the syntax summary line like quotes around the other entities. This seems more consistent.

### DIFF
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -4,7 +4,7 @@
 .\" This file is distributed under the University of Illinois Open Source
 .\" License. See LICENSE.TXT for details.
 .\"
-.TH INCLUDE-WHAT-YOU-USE 1 "2022-02-21" include-what-you-use "User Commands"
+.TH INCLUDE-WHAT-YOU-USE 1 "2025-04-05" include-what-you-use "User Commands"
 .SH NAME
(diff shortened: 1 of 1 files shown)


include-what-you-use.1
```diff
@@ -213,13 +215,13 @@ Includes the contents of another
...
 .TP
-.IB "header\fR := " \(dq include-spec "\(dq, " \(dqvisibility\(dq
+.IB "header\fR := " \(dq include-spec "\(dq, \(dq" visibility \(dq
 Describes a header file. The
 .I include-spec
 specifies the header file and
 .I visibility
 specifies whether the header is
-.BR \(dqpublic\(dq " or " \(dqprivate\(dq .
+.BR public " or " private .
```
kimgr, line 224: Could you do similar quoting on line 241 where symbol mappings are described to have *private* by default?

kimgr: Sorry, I had my diffs backward. It's consistent after your change, so no change needed.

kimgr:
Thanks!

PR #1703, opened 2025-04-06
[ci] Work around broken llvm-21 CMake modules

kimgr:
Somehow the LLVMConfig.cmake and ClangConfig.cmake modules compute the install prefix by stepping up from the current file, and they end up one step too far up.

So when using this faulty install prefix to compute the CMake dir used for includes, they wind up in /usr/lib/lib/cmake instead of /usr/lib/llvm-21/lib/cmake.

Use a terrifying sed hack to remove the second get_filename_component and set the world straight again.


kimgr:
Now that this works, we uncover compiler errors due to broken APIs in Clang. Merging past that.

Issue #1700, opened 2025-04-05
IWYU master is broken vs. Clang 21


kimgr:
After patching CMake issues in packaging, I managed to build IWYU against latest Clang mainline.

There are two (possibly related issues):

First, `DependentTemplateSpecializationType` no longer has a `getQualifier()` method (removed in https://github.com/llvm/llvm-project/commit/dc17429ae6961a6783371dcf6749eea657b5446a)

Second, fixing that by using `...->getDependentTemplateName().getQualifier()` instead (not sure that's the right fix) makes it possible to build IWYU, but test runs fail on pointer-to-member access.

There have been several changes to pointer-to-member code in Clang recently, so something probably changed semantics in a way that we're not ready for. I'm currently building LLVM from source to try and bisect which change broke us.

kimgr:
CMake issue patched in https://github.com/include-what-you-use/include-what-you-use/commit/2ea2f9fe78c5b930517bffeae021bb3877ef0d5e.

bolshakov-a:
https://github.com/llvm/llvm-project/commit/14f7bd63b95d0f61a6f47119ac66398ca230559a introduces `NestedNameSpecifier` node between `MemberPointerType` and its class type, whereas `NestedNameSpecifier` turns off forward-declarability in IWYU:
https://github.com/include-what-you-use/include-what-you-use/blob/d436a81a7c7a49ac2e15b75c1deddf1fb76d1cbf/iwyu.cc#L2572

kimgr:
@bolshakov-a Ah, thanks. I was just about to start digging :-)

Indeed, this patch fixes the test suite:
```diff
diff --git a/iwyu.cc b/iwyu.cc
index 4b56b2b..5b1c13e 100644
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2566,10 +2566,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
     if (!Base::VisitNestedNameSpecifier(nns))
       return false;
-    // If we're in an nns (e.g. the Foo in Foo::bar), we're never
-    // forward-declarable, even if we're part of a pointer type, or in
-    // a template argument, or whatever.
-    current_ast_node()->set_in_forward_declare_context(false);
+
+    ASTNode* ast_node = current_ast_node();
+    if (!ast_node->ParentIsA<clang::MemberPointerType>()) {
+      // If we're in an nns (e.g. the Foo in Foo::bar), we're never
[...]

PR #1704, opened 2025-04-06
Add mapping generator for macOS

### LINKED ISSUE #1701: include-what-you-use 0.24 suggesting _stdio.h on macOS 15
I ran `include-what-you-use` against [this file](https://github.com/mohd-akram/todo/blob/master/src/todo.c) and I get this result:

```
src/main.c should add these lines:
#include <_printf.h>  // for printf
#include <_stdio.h>   // for fprintf, fclose, stderr, fopen, stdout
#include <_stdlib.h>  // for atoi, getenv, strtol, EXIT_FAILURE, EXIT_SUCCESS
#include <_string.h>  // for strlen, strcat

src/main.c should remove these lines:
- #include <string.h>  // lines 5-5

The full include-list for src/main.c:
#include <_printf.h>  // for printf
#include <_stdio.h>   // for fprintf, fclose, stderr, fopen, stdout
#include <_stdlib.h>  // for atoi, getenv, strtol, EXIT_FAILURE, EXIT_SUCCESS
#include <_string.h>  // for strlen, strcat
#include <stdbool.h>  // for bool, false, true
#include <stddef.h>   // for NULL, size_t
#include <stdio.h>    // for FILE
#include <stdlib.h>   // for free, malloc, realloc
#include "todo.h"     // for load_tasks, move_task, remove_task, add_task
```

Apple seems to have introduced these private headers in macOS 15 (see [here](https://github.com/apple-oss-distributions/Libc/tree/main/include)).

mohd-akram:
Fixes #1701


kimgr:
Very nice! Python is so modern it makes my head spin :-)

I'll need some time to review this in detail, but overall looks great.

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,123 @@
...
+            public_path = p.parent.with_name(p.name)
+            if public_path in paths:
+                p = public_path
+
+        # _stdio.h -> stdio.h
+        while p.name.startswith('_'):
+            public_path = p.with_name(p.name[1:])
+            if public_path in paths:
+                p = public_path
+            else:
```
kimgr, line 57: I'm not sure I follow the loop here --

* if `p` starts with `_`, we strip off the first `_`
* if the stripped path exists in the public paths, we update `p` to the stripped name and goto 1
* if it doesn't, we break out of the loop

Ah, so for `___wctype.h`, we'll keep stripping off `_` until we're at `wctype.h`.

But for `_static_assert.h` there is no public header `static_assert.h`, so that will be left alone?

It seems a little sensitive if there's a case like `___header.h` (three underscores), `_header.h` (one underscore) and `header.h` (none, public header), but I suppose that doesn't exist.

I wonder if this would be easier to express as:
```
public_path = p.parent.with_name(p.name.lstrip('_'))
if public_path in paths:
    p = public_path
```
 or does that do something else?

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,123 @@
...
+                break
+
+        # arm -> machine
+        if p.parent.name in {'arm', 'arm64', 'i386'}:
+            public_path = p.parent.with_name('machine') / p.name
+            if public_path in paths:
```
kimgr, line 63: But if not, should that be an error/warning?

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,123 @@
...
+                p = public_path
+
+        # sys/time.h -> time.h
```
kimgr, line 66: This comment looks like it no longer applies.

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,123 @@
...
+        # map other headers
+        for glob, repl in mappings.items():
+            if p.relative_to(root).full_match(glob):
+                public_path = root / repl
+                if public_path in paths:
+                    p = public_path
+                break
+
+        # we don't know this one
+        if p.name.startswith('_'):
```
kimgr, line 90: Again, maybe a little warning might be nice, so users notice when new headers are overlooked? There's a recent example in the libstdcxx generator, that prints to stderr

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,123 @@
...
+            includes.append([
+                f'<{initial.relative_to(root)}>',
+                'private',
+                f'<{p.relative_to(root)}>',
+                'public'
+            ])
+
+    includes.sort(key=itemgetter(0))
+
+    print(f'# Generated using {Path(__file__).name}')
```
kimgr, line 103: Might as well use entire sys.argv here so we can see which source directory for reproducibility. Fancy example in libstdcxx mapgen, which does shell line wrapping, probably not necessary here.

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,123 @@
...
+def main(dir):
+    """ Entry point. """
+
+    root = Path(dir)
+
+    paths = set(root.glob('**/*.h'))
+
+    includes = []
+
+    mappings = {
```
kimgr, line 33: Can make this a global constant, eg PATTERN_MAPPINGS or something

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def main(dir):
+    """ Entry point. """
+
+    root = Path(dir)
+
+    paths = set(root.glob('**/*.h'))
+
+    includes = []
```
kimgr, line 31: I feel like this would be better named 'mappings'

mapgen/iwyu-mapgen-macos.py
kimgr: (These warning suggestions can probably be dropped if they're very noisy by default. But it sounded like these were both things that should never happen?)

mapgen/iwyu-mapgen-macos.py
mohd-akram: Only reason I didn't do that is because the include path on macOS...is a bit long (`/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include`) :D

mapgen/iwyu-mapgen-macos.py
mohd-akram: It should only be a warning if all the attempts fail since a subsequent step might manage to map the header.

mapgen/iwyu-mapgen-macos.py
mohd-akram: It still applies, but that's just one example. I've clarified that this is just an example.

mapgen/iwyu-mapgen-macos.py
kimgr: Ouch! Well, I guess it's worth overflowing the 80-col limit for traceability 🙂

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,129 @@
...
+from pathlib import Path
+
+
+PATTERN_MAPPINGS = {
+    'pthread_*': 'pthread.h',
+    'sys/_types/_fd_*': 'sys/select.h',
+    'sys/_types/_in_*': 'netinet/in.h',
+    'sys/_types/_o_*': 'fcntl.h',
+    'sys/_types/_s_*': 'sys/stat.h',
+    'sys/_types/_seek_*': 'stdio.h',
```
mohd-akram, line 30: We can get rid of all of these special cases by adding more symbols to iwyu_include_picker.cc, since they're all POSIX symbols with standard locations. The benefit of that is that it would also work on other systems. See #1706.

mapgen/iwyu-mapgen-macos.py
kimgr: Ah, right, the subsequent rules might do something. Alright, makes sense.

mapgen/iwyu-mapgen-macos.py
kimgr: I wonder if it makes sense to move ahead with these independently, and merge this first?

I'd like to see some test output from #1706, and that's currently blocked by #1700.

mohd-akram: Sure, sounds good to me.

kimgr: I'll try to get to it tomorrow.

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env python3
+
+##===--- iwyu-mapgen-macos.py ---------------------------------------------===##
```
kimgr, line 3: Sorry for thinking about naming so late, but I think this should be `iwyu-mapgen-apple-libc.py` -- the Apple Libc seems to be used across macOS, iOS, tvOS and watchOS.

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,129 @@
...
+#
+#                     The LLVM Compiler Infrastructure
+#
+# This file is distributed under the University of Illinois Open Source
+# License. See LICENSE.TXT for details.
+#
+##===----------------------------------------------------------------------===##
+
+""" Generate mappings for macOS libc.
```
kimgr, line 12: macOS libc -> Apple Libc

mapgen/iwyu-mapgen-macos.py
```diff
@@ -0,0 +1,129 @@
...
+        json.dump(dict(include=mapping), sys.stdout)
+        print(',')
+    print(']')
+
+    return 0
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Generate IWYU mappings for macOS libc.'
```
kimgr, line 119: Apple Libc

mapgen/iwyu-mapgen-macos.py
kimgr: And let's update commit message title too

mapgen/iwyu-mapgen-macos.py
mohd-akram: You're right, that definitely is more accurate.

kimgr:
Thank you!

firewave:
I integrated it into the Cppcheck CI (see [#7465](https://github.com/danmar/cppcheck/pull/7465)) but it fails to run:

```
+ python3 iwyu-mapgen-apple-libc.py /Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include
  File "/Users/runner/work/cppcheck/cppcheck/iwyu-mapgen-apple-libc.py", line 106
    print(f'# Generated using {' '.join(sys.argv)}')
                                                   ^
SyntaxError: f-string: expecting '}'
```
https://github.com/danmar/cppcheck/actions/runs/14433185399/job/40470455870

mohd-akram:
I think this is because of the nested quotes in the f-string. It requires Python 3.12+. Can be fixed by just using different quotes internally.

firewave:
That seems a bit high reaching given 3.13 was just released. Python 3.9 is still in support until October 2025: https://devguide.python.org/versions/.

I am using the GitHub `macos-13` runner which is still supported. According to the documentation that should be using 3.13 though: https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md ... and I confirmed that is indeed the case:

```
-- Python_VERSION =        3.13.2
```

firewave:
On a side note - maybe there should be build steps for testing the generators with the various available runners.

mohd-akram:
> That seems a bit high reaching given 3.13 was just released. Python 3.9 is still in support until October 2025: https://devguide.python.org/versions/.

Yeah, it should probably be changed. I didn't notice it actually.

> I am using the GitHub `macos-13` runner which is still supported. According to the documentation that should be using 3.13 though: https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md ... and I confirmed that is indeed the case:
> 
> ```
> -- Python_VERSION =        3.13.2
> ```

It should definitely work with that. I tested with Python 3.13.3.

firewave:
After removing that override the mapping generator is working.

But the analysis fails with an assertion:
```
/private/tmp/include-what-you-use-20250406-7581-b0uutx/include-what-you-use/iwyu_include_picker.cc:1467: Assertion failed: (*map)[key] == visibility Same file seen with two different visibilities: <sys/time.h> Old vis: 1 New vis: 2
```
https://github.com/danmar/cppcheck/actions/runs/14442622565/job/40495968319

[macos.imp.zip](https://github.com/user-attachments/files/19733484/macos.imp.zip)

----

A nit: the first line is a comment and IIRC mappings are JSON files. And JSON files do not allow comments. Some parsers actually do but the file is not valid.

kimgr:
@firewave The .imp format is not json, it's yaml, see the last paragraph of https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUMappings.md#mapping-files

kimgr:
@mohd-akram Maybe we could demodernize the python a bit for wider compat. How new are the Path API features in use? Those seem nice and useful, but I could live without the `f` formatting.

firewave:
> The .imp format is not json, it's yaml

Good to know before I had back into the Qt mapping generator.

FYI the LLVM YAML parser is not complete and ignored things silently in the past (like multi-line fields).

firewave:
> Maybe we could demodernize the python a bit for wider compat. How new are the Path API features in use? Those seem nice and useful, but I could live without the `f` formatting.

Both are already available in Python 3.6. The issue is some extension of the fstring syntax which was apparently only introduced in 3.12.

mohd-akram:
Thanks @firewave, good catch. I've simplified the script quite a bit in #1714 and moved more of the mappings to #1706.

@kimgr It should be widely compatible now in my PR. The earliest non-EOL Python is 3.9, that's also what comes with macOS and I've verified that the script runs with it well with the new changes.

firewave:
I did a comparison run and it looks good. Thanks!

mapgen/iwyu-mapgen-apple-libc.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # architecture/byte_order.h -> machine/byte_order.h
+        if p.parent.name == 'architecture':
+            public_path = p.parent.with_name('machine') / p.name
+            if public_path in paths:
+                p = public_path
+
+        # map other headers
+        for glob, repl in PATTERN_MAPPINGS.items():
+            if p.relative_to(root).full_match(glob):
```
kimgr, line 84: I noticed when testing the revised version that `Path.full_match` is not available until Python 3.13, so this fails to run on my fairly recent Ubuntu 24.04 system. Let's see if we can relax that requirement a bit when fixing the quotes issue too?

mohd-akram: It can be changed to `Path.match` which is less strict but maybe is good enough.

PR #1705, opened 2025-04-06
[clang compat] Adjust GetQualifier for changed Clang API

kimgr:
Clang removed DependentTemplateSpecializationType::getQualifier() and moved it into a DependentTemplateStorage accessible via getDependentTemplateName() in https://github.com/llvm/llvm-project/commit/dc17429ae6961a6783371dcf6749eea657b5446a

Fix our only use so it does the two-step extraction of qualifier.


kimgr:
Test failures are due to another upstream Clang change, so I'll address those separately.

PR #1706, opened 2025-04-06
Add more POSIX mappings

### LINKED ISSUE #1627: implementation header suggested for `locale_t`
```cpp
#include <locale.h>

void f()
{
    locale_t l;
}
```

```
#include <bits/types/locale_t.h>  // for locale_t

- #include <locale.h>  // lines 1-1

#include <bits/types/locale_t.h>  // for locale_t
---
```

https://man7.org/linux/man-pages/man3/locale_t.3type.html.

Happens with 0.23dev and can reproduce back to at least 0.20 (have no earlier binaries available).

### LINKED ISSUE #1628: implementation header suggested for `mbstate_t`
```cpp
#include <wchar.h>

void f()
{
    mbstate_t s;
}
```

```
#include <bits/types/mbstate_t.h>  // for mbstate_t

- #include <wchar.h>  // lines 1-1

#include <bits/types/mbstate_t.h>  // for mbstate_t
---
```

https://man7.org/linux/man-pages/man3/mbstate_t.3type.html
https://en.cppreference.com/w/c/string/multibyte/mbstate_t
https://en.cppreference.com/w/cpp/string/multibyte/mbstate_t

Happens with 0.23dev and can reproduce back to at least 0.20 (have no earlier binaries available).

mohd-akram:
These additions come from running `include-what-you-use` against all the .c files in the Android project's [POSIX tests](https://github.com/aosp-mirror/platform_bionic/tree/main/tests/headers/posix), then filtering the results to only include the ones that would require hard-coding in the macOS generator (#1704). The benefit of this approach is that this will work on other systems too (eg. BSDs) without requiring a mapping file for each one.

Fixes #1627
Fixes #1628

The script might be useful to run on other systems as well:

<details>

<summary>Script</summary>

```shell
#!/bin/sh
# usage: iwyu-posix file

set -euo pipefail

src=$(sed '
/#include "sys_stat_h_mode_constants.h"/r sys_stat_h_mode_constants.h
/#include "sys_stat_h_file_type_test_macros.h"/r sys_stat_h_file_type_test_macros.h
' "$1")
shift

while :; do
	out="$(printf '%s\n' "$src" | include-what-you-use "$@" \
		-Wno-error=incompatible-function-pointer-types \
		-Wno-error=implicit-int \
		-xc - 2>&1 || :)"
	errors=$(printf '%s\n' "$out" | grep ': error:' || :)

	if [ "$errors" ]; then
		lines=$(printf '%s\n' "$errors" | cut -d: -f2 | sed 's/$/d/')
		src=$(printf '%s\n' "$src" | sed "$lines")
	else
		printf '%s\n' "$out"
		break
	fi
done
```

</details>

<details>

<summary>Script output</summary>

```
#include <_abort.h>          // for abort
#include <_ctermid.h>              // for L_ctermid, ctermid
#include <_printf.h>               // for printf
#include <_types/_locale_t.h>  // for locale_t
#include <_types/_nl_item.h>  // for nl_item
#include <_types/_wctrans_t.h>  // for wctrans_t
#include <_types/_wctype_t.h>       // for wctype_t
#include <_types/_wctype_t.h>   // for wctype_t
#include <arm/_mcontext.h>                 // for mcontext_t
#include <malloc/_malloc.h>  // for posix_memalign
#include <sys/_pthread/_pthread_attr_t.h>        // for pthread_attr_t
#include <sys/_pthread/_pthread_attr_t.h>  // for pthread_attr_t
#include <sys/_pthread/_pthread_cond_t.h>        // for pthread_cond_t
#include <sys/_pthread/_pthread_condattr_t.h>    // for pthread_condattr_t
#include <sys/_pthread/_pthread_key_t.h>         // for pthread_key_t
#include <sys/_pthread/_pthread_mutex_t.h>       // for pthread_mutex_t
#include <sys/_pthread/_pthread_mutexattr_t.h>   // for pthread_mutexattr_t
#include <sys/_pthread/_pthread_once_t.h>        // for pthread_once_t
#include <sys/_pthread/_pthread_rwlock_t.h>      // for pthread_rwlock_t
#include <sys/_pthread/_pthread_rwlockattr_t.h>  // for pthread_rwlockattr_t
#include <sys/_pthread/_pthread_t.h>             // for pthread_t
#include <sys/_pthread/_pthread_t.h>       // for pthread_t
#include <sys/_select.h>             // for select
#include <sys/_types/_fd_clr.h>      // for FD_CLR
#include <sys/_types/_fd_isset.h>    // for FD_ISSET
#include <sys/_types/_fd_set.h>      // for FD_SET
#include <sys/_types/_fd_setsize.h>  // for FD_SETSIZE
#include <sys/_types/_fd_zero.h>     // for FD_ZERO
#include <sys/_types/_in_addr_t.h>    // for in_addr_t
#include <sys/_types/_in_addr_t.h>  // for in_addr_t
#include <sys/_types/_in_port_t.h>    // for in_port_t
#include <sys/_types/_in_port_t.h>  // for in_port_t
#include <sys/_types/_mbstate_t.h>  // for mbstate_t
#include <sys/_types/_o_dsync.h>   // for O_DSYNC
#include <sys/_types/_o_sync.h>    // for O_SYNC
#include <sys/_types/_posix_vdisable.h>  // for _POSIX_VDISABLE
#include <sys/_types/_s_ifmt.h>    // for S_IFBLK, S_IFCHR, S_IFDIR, S_IFIFO, S_IFLNK, S_IFMT, S_IFREG, S_IFSOCK, S_IRGRP, S_IROTH, S_IRUSR, S_IRWXG, S_IRWXO, S_IRWXU, S_ISGID, S_ISUID, S_ISVTX, S_IWGRP, S_IWOTH, S_IWUSR, S_IXGRP, S_IXOTH, S_IXUSR
#include <sys/_types/_s_ifmt.h>  // for S_IFBLK, S_IFCHR, S_IFDIR, S_IFIFO, S_IFLNK, S_IFMT, S_IFREG, S_IFSOCK, S_IRGRP, S_IROTH, S_IRUSR, S_IRWXG, S_IRWXO, S_IRWXU, S_ISGID, S_ISUID, S_ISVTX, S_IWGRP, S_IWOTH, S_IWUSR, S_IXGRP, S_IXOTH, S_IXUSR
#include <sys/_types/_sa_family_t.h>  // for sa_family_t
#include <sys/_types/_seek_set.h>        // for SEEK_CUR, SEEK_END, SEEK_SET
#include <sys/_types/_seek_set.h>  // for SEEK_CUR, SEEK_END, SEEK_SET
#include <sys/_types/_ucontext.h>          // for ucontext_t
```

</details>


kimgr:
Slick! I didn't know about the Android POSIX tests, that's quite a useful resource.

I tried to play around with this locally (Ubuntu system) in checkout of platform_bionic, but I couldn't get any meaningful results. How did you invoke `iwyu-posix.sh`?

mohd-akram:
@kimgr You should be able to just do eg. `./iwyu-posix float_h.c` in the `platform_bionic/tests/headers/posix` directory. For me, that outputs:

```
#include <sys/_types/_o_dsync.h>   // for O_DSYNC
#include <sys/_types/_o_sync.h>    // for O_SYNC
#include <sys/_types/_s_ifmt.h>    // for S_IFBLK, S_IFCHR, S_IFDIR, S_IFIFO
#include <sys/_types/_seek_set.h>  // for SEEK_CUR, SEEK_END, SEEK_SET
#include <sys/fcntl.h>             // for flock, AT_EACCESS, AT_FDCWD, AT_RE...
#include <sys/types.h>             // for off_t, mode_t, pid_t
#include "/tmp/posix/sys_stat_h_mode_constants.h"

- #include <fcntl.h>  // lines 4-4

#include <sys/_types/_o_dsync.h>   // for O_DSYNC
#include <sys/_types/_o_sync.h>    // for O_SYNC
#include <sys/_types/_s_ifmt.h>    // for S_IFBLK, S_IFCHR, S_IFDIR, S_IFIFO
#include <sys/_types/_seek_set.h>  // for SEEK_CUR, SEEK_END, SEEK_SET
#include <sys/fcntl.h>             // for flock, AT_EACCESS, AT_FDCWD, AT_RE...
#include <sys/types.h>             // for off_t, mode_t, pid_t
#include "/tmp/posix/sys_stat_h_mode_constants.h"
#include "header_checks.h"         // for MACRO, MACRO_VALUE, STRUCT_MEMBER
#include "sys_stat_h_mode_constants.h"
---
```

I invoked it like this to get all the headers that contain an underscore in them:

```
printf '%s\n' *.c | xargs -I{} ./iwyu-posix {} -Xiwyu --max_line_length=1000 | grep '[</]_' | sort | uniq
```

mohd-akram:
Yeah, I referred to the POSIX standard and mapped them manually.

kimgr:
I see, thanks! I thought there was something more automated going on, and couldn't connect the dots.

kimgr:
Hmm, actually... I wonder if we could go wild with regex and generate a full set of POSIX mappings from the Android POSIX tests. Since every public header has a structured listing of all symbols it owns (the `MACRO`, `FUNCTION`, `TYPE` macros), it should be possible to do a structured translation. I might look into that at some point, just thinking out loud.

mohd-akram:
I very much had the same crazy thought :D 

And before finding the Android tests I thought about doing a parsing of the POSIX specification (since I had [just](https://github.com/mohd-akram/man-posix) done on a project with it).

The advantage of the latter would be that the spec mentions which header is the canonical one for a given symbol. The Android tests seem to test for symbols even in non-canonical headers. But of course parsing the tests would be far easier since they're incredibly regular in the way they're written. But maybe that metadata can be obtained somehow, or with some heuristics, or possibly the instances of headers redefining things are small enough that they can be filtered out.

kimgr:
@mohd-akram Yeah, I like the spec better as a source, too. But that Makefile looks terrifying :-) Food for thought, anyway, maybe we can figure something out.

mohd-akram:
> @mohd-akram Yeah, I like the spec better as a source, too. But that Makefile looks terrifying :-)

Haha, I'm sorry you had to see that. That's part of the reason I didn't want to get into parsing the spec again. It won't be too bad if there's a proper HTML parser with a DOM (Python only has a very [basic](https://docs.python.org/3/library/html.parser.html) parser), but presumably you'd want to avoid an additional dependency like beautifulsoup.

There should be some repo with this kind of POSIX metadata so projects can use it easily. I've searched but couldn't find anything other than the Android tests. I'll look into it some more.

mohd-akram:
Looks like [cython_posix](https://github.com/earowley/cython_posix/tree/main) might be useful for this.

iwyu_include_picker.cc
```diff
@@ -95,21 +95,29 @@ const IncludeMapEntry libc_symbol_map[] = {
   // an option for this type.  That's the preferred #include all else
   // equal.  The visibility on the symbol-name is ignored; by convention
   // we always set it to kPrivate.
+  { "_POSIX_VDISABLE", kPrivate, "<unistd.h>", kPublic },
```
kimgr, line 98: Looks like we have this whole map duplicated in https://github.com/include-what-you-use/include-what-you-use/blob/7b3205c3ca548463edf502045fea9c6c8d9d43c3/gcc.libc.imp -- can you make the same changes there for now?

At some point it would obviously be better to generate this whole thing from the same source, but I'd prefer to take some cheap wins first, and then think of the whole POSIX generator separately.

(the external mapping is useful when running with `-Xiwyu --no_default_mappings` e.g. for some baremetal setup, and then being able to selectively add back the builtin sets with e.g. `-Xiwyu --mapping_file=gcc.libc.imp`)

(I'm also aware that the file is probably misnamed, I think the libc is not directly associated with GCC)

kimgr:
> Looks like [cython_posix](https://github.com/earowley/cython_posix/tree/main) might be useful for this.

Yeah, maybe. It's like two steps removed from the actual standard, but I guess it would be much easier to introspect Python than parse C or HTML, sadly.

iwyu_include_picker.cc
mohd-akram: Did you mean to link to `gcc.symbols.imp`? The other file has header mappings not symbol ones.

mohd-akram:
> > Looks like [cython_posix](https://github.com/earowley/cython_posix/tree/main) might be useful for this.
> 
> Yeah, maybe. It's like two steps removed from the actual standard, but I guess it would be much easier to introspect Python than parse C or HTML, sadly.

Yeah. I started on it a bit. Basically you can use `TreeFragment.parse_from_strings` from Cython to parse those pxd files and then introspect the nodes. They include some GCC-specific stuff though. I think I'd rather just parse the spec. Been trying it out and you can get most of the way with a simple shell script so I think it should be doable, but it will require some tweaking to get the output fully right.

iwyu_include_picker.cc
kimgr: Oof, yes -- sorry! I meant `gcc.symbols.imp`.

kimgr:
@mohd-akram Part of the reason the mapping generators are Python is for portability. I'd like to keep them runnable on Windows too, so I've tended to avoid shell scripts other than for internal dev tools. I left Windows before the WSL, so maybe things are easier to get running these day.

mohd-akram:
For sure. I was thinking that I could have the parsed symbols as simple text/JSON files in a separate repo (which could also be used by other projects) and then IWYU could consume those. Thinking more though, the whole idea of a spec parser/mapping generator for POSIX might be overkill :D The POSIX spec isn't exactly fast changing, and it's not too difficult to open the page for a [header](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/stdio.h.html) and just scan the symbols manually and add them all. What might be helpful, whether a generator is used or not, would be to group the entries in `gcc.symbols.imp`/`iwyu_include_picker.cc` by header. Then it's easier to see if things are mapped correctly header by header.

kimgr:
This should fix #1627 and #1628.

kimgr:
Both of these pass on Ubuntu:
```
./run_iwyu_tests.py
IWYU_EXTRA_ARGS="-Xiwyu --no_default_mappings -Xiwyu --mapping_file=iwyu.gcc.imp" ./run_iwyu_tests.py
```
So the built-in and external mappings appear to agree as far as the test suite is concerned.

LGTM!

PR #1708, opened 2025-04-08
[clang compat] Handle new MemberPointerType structure

kimgr:
Clang was modified to sometimes put a NestedNameSpecifier under MemberPointerType, for the purposes of preserving sugar, in: https://github.com/llvm/llvm-project/commit/14f7bd63b95d0f61a6f47119ac66398ca230559a

This broke our test suite for pointer-to-member forward decls, since VisitNestedNameSpecifier disables forward declarations unconditionally.

Peek up the tree and only disable forward decls if we're not part of a MemberPointerType.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -220,6 +220,7 @@ using clang::FunctionType;
 using clang::LateParsedTemplate;
 using clang::LinkageSpecDecl;
 using clang::MemberExpr;
+using clang::MemberPointerType;
 using clang::NamedDecl;
 using clang::NamespaceAliasDecl;
 using clang::NestedNameSpecifier;
@@ -2566,10 +2567,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
     if (!Base::VisitNestedNameSpecifier(nns))
       return false;
-    // If we're in an nns (e.g. the Foo in Foo::bar), we're never
-    // forward-declarable, even if we're part of a pointer type, or in
-    // a template argument, or whatever.
-    current_ast_node()->set_in_forward_declare_context(false);
+
+    // MemberPointerType allows forward-declaration of their nested names, e.g.
+    // Class in Class::*member...
+    ASTNode* ast_node = current_ast_node();
+    if (!ast_node->ParentIsA<MemberPointerType>()) {
+      // ... but for other nns (e.g. the Foo in Foo::bar), we're never
+      // forward-declarable, even if we're part of a pointer type, or in a
+      // template argument, or whatever.
+      ast_node->set_in_forward_declare_context(false);
+    }
     return true;
   }
 


kimgr:
@bolshakov-a Could you see if this is a fair characterization in comments and commit message?

iwyu.cc
```diff
@@ -2566,10 +2566,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-    // a template argument, or whatever.
-    current_ast_node()->set_in_forward_declare_context(false);
+
+    ASTNode* ast_node = current_ast_node();
+    if (!ast_node->ParentIsA<clang::MemberPointerType>()) {
+      // If we're in an nns (e.g. the Foo in Foo::bar), we're never
+      // forward-declarable, even if we're part of a pointer type, or in a
+      // template argument, or whatever. The exception is MemberPointerType,
+      // which uses nns to carry additional sugar, without actually having nns
+      // semantics.
```
bolshakov-a, line 2576: This sounds a little strange for me. `::` is not an additional sugar but the canonical way to denote a pointer-to-member type. Or, do you mean that `NestedNameSpecifier` node is actually redundant in the AST and thus is a sugar for AST reader (as opposed to a language sugar)?

> without actually having nns semantics

Is it because there is no (nested) written name? Seems like it actually is the semantics of "pointer to something that is nested inside a class".

Personally, I think that `The exception is MemberPointerType.` without any further explanation would be sufficient.

iwyu.cc
```diff
@@ -2566,10 +2566,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
     if (!Base::VisitNestedNameSpecifier(nns))
       return false;
-    // If we're in an nns (e.g. the Foo in Foo::bar), we're never
-    // forward-declarable, even if we're part of a pointer type, or in
-    // a template argument, or whatever.
-    current_ast_node()->set_in_forward_declare_context(false);
+
+    ASTNode* ast_node = current_ast_node();
+    if (!ast_node->ParentIsA<clang::MemberPointerType>()) {
```
bolshakov-a, line 2571: `using clang::MemberPointerType;`?

iwyu.cc
bolshakov-a: I see the note of sugar in the clang commit message (https://github.com/llvm/llvm-project/pull/132401). As far as I understand, nns is meant to be better suited to preserve, indeed, any additional written qualification like in `int ns1::ns2::Class::*` (instead of using `ElaboratedType` referring to nns in such cases, I think). There is also a change in preserving sugar when a `MemberPointerType` refers in fact to the base class members whereas it has been got from a reference to a derived class member (like in `decltype(&Derived::base_member)`), but I think it is unrelated. All of this is somewhat obscure for me...

kimgr: Thanks -- I'll admit that I just wrote something in the hope that you would suggest something better :-)

Now I've looked into it a bit more, and it looks to me like the NNS is always inserted -- but you said it's not always there? I softened the language in commit message and comments to gloss over whether it's always/sometimes there.

When I originally read the Clang patch, I somehow understood it as abusing NNS to represent the type + metadata, but on reflection it seems to be clear that the NNS in fact accurately represents the nested name + type information better than a plain type. That's where the "semantics" misunderstanding came from. Now removed.

iwyu.cc
kimgr: Done

iwyu.cc
bolshakov-a: > it looks to me like the NNS is always inserted -- but you said it's not always there?

According to the current `RecursiveASTVisitor` code, it may be absent in some cases, but I have not succeeded in achieving this. (See [the comment](https://github.com/llvm/llvm-project/pull/132551#issuecomment-2782712579), to which the author did not respond for some reason.)

But I've realized that it doesn't really matter because `VisitMemberPointerType` should anyway be added to enable forward-declarations of member own types, like in:
```cpp
class Class{};
// IndirectClass may be forward-declared.
IndirectClass Class::* p;
```

PR #1709, opened 2025-04-12
Handle member pointer types in "assignable" traits

bolshakov-a:
This has been done before fixing forward-declarability in member pointer types to avoid regression. Due to this, the test cases have been written using aliased types. (If the direct class types were used, they would be reported in all cases.)

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2084,9 +2084,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
           const Type* rhs_deref_type = RemoveReference(rhs_type);
           // For derived-to-base pointer conversion to be allowed, the derived
           // type should be complete.
-          // TODO(bolshakov): what about member pointers?
           if (IsDerivedToBasePtrConvertible(rhs_deref_type, lhs_deref_type)) {
             ReportTypeUse(CurrentLoc(), rhs_type, DerefKind::RemoveRefsAndPtr);
+            return true;
+          }
+          if (IsBaseToDerivedMemPtrConvertible(rhs_deref_type, lhs_deref_type,
+                                               compiler()->getSema())) {
+            const auto* lhs_mem_ptr_type =
+                lhs_deref_type->castAs<MemberPointerType>();
+            // TODO(bolshakov): check whether the type is provided by member
+            // pointer type alias, probably?
+            ReportTypeUse(CurrentLoc(),
+                          lhs_mem_ptr_type->getQualifier()->getAsType(),
+                          DerefKind::None);
             return true;
           }
           // Otherwise, report the rhs record type (ReportTypeUse doesn't report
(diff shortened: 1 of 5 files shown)


iwyu.cc
```diff
@@ -2084,11 +2084,26 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
           const Type* rhs_deref_type = RemoveReference(rhs_type);
           // For derived-to-base pointer conversion to be allowed, the derived
           // type should be complete.
-          // TODO(bolshakov): what about member pointers?
           if (IsDerivedToBasePtrConvertible(rhs_deref_type, lhs_deref_type)) {
             ReportTypeUse(CurrentLoc(), rhs_type, DerefKind::RemoveRefsAndPtr);
             return true;
           }
+          if (const auto* rhs_mem_ptr_type =
```
kimgr, line 2091: I wonder if this could go into a `HandleTraitsMemberPointerType` or something? The block is almost identical between here and `BTT_IsTriviallyAssignable`. I have grown to prefer open-coded functions over highly decomposed, but this is so complex that it breaks the main flow.

Actually, maybe a compromise is to change `IsBaseToDerivedMemPtrConvertible` to take two `Type*`s, and do the downcast inside? That would remove a lot of incidental indentation. Oh, or the `lhs_mem_ptr_type->getQualifier()` needs a `MemberPointerType`?

kimgr: On the other hand, if `IsBaseToDerivedMemPtrConvertible` returns true, you know that `lhs_deref_type` is-a `MemberPointerType`, so you could cast it again, unconditionally. It's annoying to cast twice, but I think it would be easier to follow and look more like https://github.com/include-what-you-use/include-what-you-use/pull/1709/files#diff-7ca107c8b836c1765e8b735b74f06833bec48f308a76017886cee1d5736002d9L2088

iwyu.cc
```diff
@@ -2084,11 +2084,26 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
           const Type* rhs_deref_type = RemoveReference(rhs_type);
           // For derived-to-base pointer conversion to be allowed, the derived
           // type should be complete.
-          // TODO(bolshakov): what about member pointers?
           if (IsDerivedToBasePtrConvertible(rhs_deref_type, lhs_deref_type)) {
             ReportTypeUse(CurrentLoc(), rhs_type, DerefKind::RemoveRefsAndPtr);
             return true;
           }
```
kimgr, line 2090: `BTT_IsTriviallyAssignable` case has an `else if` here, that would work here too, right?

iwyu_ast_util.cc
```diff
@@ -1713,6 +1715,33 @@ bool IsDerivedToBasePtrConvertible(const Type* derived_ptr_type,
   return false;
 }
 
+bool IsBaseToDerivedMemPtrConvertible(
```
kimgr, line 1718: Could you add a small example of two types/functions being base-to-derived member pointer convertible in the doc comment? Just something to trigger the brain what this looks like in actual code.

iwyu.cc
bolshakov-a: `else` is senseless when the previous branch ends with `return` statement.

kimgr: Ah. Yes.

iwyu.cc
bolshakov-a: Done. Furthermore, the contents of that `if` statement branch can actually be moved into a separate handler method ("One-to-five-line functions should be considered normal." — C++ Core Guidelines). Or even the whole `if - else if` statement into something like `HandlePtrInheritanceInAssignableTypeTrait`...

kimgr: Looks good! As I said, I've begun to change my mind on function size. Decomposing into one-to-five-line functions turns reading code from a challenge of understanding what code does, to a challenge of understanding what a name means -- and I find people (particularly myself) are much worse at inventing good names than at writing working code.

bolshakov-a: LOL

PR #1711, opened 2025-04-12
Add support for the cleanup attribute

### LINKED ISSUE #1710: __attribute__((__cleanup__(x))) support is missing
https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-cleanup-variable-attribute

> The cleanup attribute runs a function when the variable goes out of scope. This attribute can only be applied to auto function scope variables; it may not be applied to parameters or variables with static storage duration. The function must take one parameter, a pointer to a type compatible with the variable. The return value of the function (if any) is ignored. 

It seems like include-what-you-use skips over cleanup `__attribute__` nodes, causing includes to get removed when symbols from those includes are only used in cleanup `__attribute__` nodes.

I would hope that supporting this use case is as simple as checking for cleanup `__attribute__` nodes and keeping track of the functions used within those.

DaanDeMeyer:
https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-cleanup-variable-attribute

This attribute takes a function which should be tracked by IWYU.

Fixes #1710

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -220,6 +220,7 @@ using clang::FunctionType;
 using clang::LateParsedTemplate;
 using clang::LinkageSpecDecl;
 using clang::MemberExpr;
+using clang::MemberPointerType;
 using clang::NamedDecl;
 using clang::NamespaceAliasDecl;
 using clang::NestedNameSpecifier;
@@ -2566,10 +2567,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
     if (!Base::VisitNestedNameSpecifier(nns))
       return false;
-    // If we're in an nns (e.g. the Foo in Foo::bar), we're never
-    // forward-declarable, even if we're part of a pointer type, or in
-    // a template argument, or whatever.
-    current_ast_node()->set_in_forward_declare_context(false);
+
+    // MemberPointerType allows forward-declaration of their nested names, e.g.
+    // Class in Class::*member...
+    ASTNode* ast_node = current_ast_node();
+    if (!ast_node->ParentIsA<MemberPointerType>()) {
+      // ... but for other nns (e.g. the Foo in Foo::bar), we're never
+      // forward-declarable, even if we're part of a pointer type, or in a
+      // template argument, or whatever.
+      ast_node->set_in_forward_declare_context(false);
+    }
     return true;
   }
 


tests/c/var_cleanup_attr-d1.h
```diff
@@ -0,0 +1,11 @@
+//===--- elaborated_struct-d1.h - test input file for iwyu ----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+static inline void freep(void *p) {
```
bolshakov-a, line 10: `static` looks strange in a header.

iwyu.cc
```diff
@@ -1728,6 +1729,11 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         ReportDeclUse(CurrentLoc(), redecl, nullptr, UF_DefinitionUse);
       }
     }
+
+    if (const CleanupAttr *attr = decl->getAttr<CleanupAttr>()) {
+      ReportDeclUse(CurrentLoc(), attr->getFunctionDecl(), nullptr, UF_None);
```
bolshakov-a, line 1734: I'd prefer not to explicitly specify the last two arguments.

iwyu.cc
```diff
@@ -1728,6 +1729,11 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         ReportDeclUse(CurrentLoc(), redecl, nullptr, UF_DefinitionUse);
       }
     }
+
+    if (const CleanupAttr *attr = decl->getAttr<CleanupAttr>()) {
```
bolshakov-a, line 1733:
```suggestion
    if (const auto* attr = decl->getAttr<CleanupAttr>()) {
```
Besides, it's worth to run clang-format on the changes.

tests/c/var_cleanup_attr.c
```diff
@@ -0,0 +1,23 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// The cleanup attribute runs a function when the variable goes out of scope. This
+// test makes sure that we track that function.
+
+#include "var_cleanup_attr-d1.h"
```
bolshakov-a, line 13: It would be better to put `freep` into an indirectly included header (`...-i1.h`) so that the test would show explicit diagnostics.

bolshakov-a:
It seems to me that the C language test folder is primarily for things that are specific for the C language (as opposed to the common ones for C and C++), and GCC attributes aren't such things, AFAIU. @kimgr should be able to clarify.

tests/c/var_cleanup_attr-d1.h
DaanDeMeyer: Suggestions for alternatives? static inline is the way to avoid the ODR warning

DaanDeMeyer:
> It seems to me that the C language test folder is primarily for things that are specific for the C language (as opposed to the common ones for C and C++), and GCC attributes aren't such things, AFAIU. @kimgr should be able to clarify.

Moved to cxx tests

tests/c/var_cleanup_attr-d1.h
bolshakov-a: `inline` is sufficient to satisfy the ODR, at least in C++:
> Every program shall contain exactly one definition of every **non-inline** function or variable that is odr-used in that program outside of a discarded statement; no diagnostic required.

https://timsong-cpp.github.io/cppwp/n4659/basic.def.odr#4

> There can be more than one definition of a class type, enumeration type, inline function with external linkage ([dcl.inline](https://timsong-cpp.github.io/cppwp/n4659/dcl.inline)), inline variable with external linkage ([[dcl.inline]](https://timsong-cpp.github.io/cppwp/n4659/dcl.inline)), ...

https://timsong-cpp.github.io/cppwp/n4659/basic.def.odr#6

In fact, it is the primary aim of `inline` nowadays.

Anyway, for the purpose of the test, it would be probably sufficient just to declare `freep` instead of defining it.

tests/cxx/var_cleanup_attr.cc
```diff
@@ -0,0 +1,33 @@
...
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+// The cleanup attribute runs a function when the variable goes out of scope.
+// This test makes sure that we track that function.
+
+#include "tests/cxx/var_cleanup_attr-d1.h"
+
+void cleanup(void) {
```
bolshakov-a, line 17: Nit: `(void)` is a C-ism.
```suggestion
void cleanup() {
```

iwyu.cc
```diff
@@ -1728,6 +1729,15 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    return true;
+  }
+
+  bool VisitCleanupAttr(CleanupAttr *attr) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+
+    ReportDeclUse(CurrentLoc(), attr->getFunctionDecl());
     return true;
   }
```
bolshakov-a, line 1742: Handlers are arranged here in groups for declarations, statements, types. The better place for `VisitCleanupAttr` is outside of them, probably in a new separate group.

tests/c/var_cleanup_attr-d1.h
kimgr: This threw me too when I switched between C++ and C, but I think the rules and conventions are slightly different between the languages (if I remember correctly, it's been a few years).

The way I understood it then is that `static inline` is frowned upon for C++, but normal for C. But again, I could be misremembering.

kimgr:
> It seems to me that the C language test folder is primarily for things that are specific for the C language (as opposed to the
> common ones for C and C++), and GCC attributes aren't such things, AFAIU. @kimgr should be able to clarify.

I think of `tests/c` as "constructs that require at least C semantics (but possibly apply to C++ too)". But it was added very late in development, so there's a bunch of tests in `tests/cxx` that could probably live in `tests/c`.

That said, I feel like the right place for this testcase is actually `tests/c`, because C++ code will probably not use `__attribute__((cleanup))` .

iwyu.cc
```diff
@@ -1728,6 +1729,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         ReportDeclUse(CurrentLoc(), redecl, nullptr, UF_DefinitionUse);
       }
     }
+
```
kimgr, line 1732: Please revert unrelated whitespace changes

tests/cxx/var_cleanup_attr.cc
```diff
@@ -0,0 +1,33 @@
...
+
+// IWYU_ARGS: -I .
+
+// The cleanup attribute runs a function when the variable goes out of scope.
+// This test makes sure that we track that function.
+
+#include "tests/cxx/var_cleanup_attr-d1.h"
+
+void Cleanup() {
+  // IWYU: Freep is defined in...*which isn't directly #included.
```
kimgr, line 18: I think you want to check that it actually does recommend `var_cleanup_attr-i1.h` here, so;
```
// IWYU: Freep is...*var_cleanup_attr-i1.h
```
is the more typical way to cover that.

tests/cxx/var_cleanup_attr-d1.h
```diff
@@ -0,0 +1,10 @@
+//===--- var_cleanup_attr-d1.h - test input file for iwyu ----------------===//
```
kimgr, line 1: These header lines should all be padded with `-` to fill 80 col limit

tests/cxx/var_cleanup_attr.cc
```diff
@@ -0,0 +1,33 @@
+//===--- var_cleanup_attr.c - test input file for iwyu -------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
```
kimgr, line 9: As mentioned in discussion, I agree that these tests are better suited for `tests/c`. Sorry about the back and forth.

kimgr (commented):
Some nits inline.

Unfortunately, we don't take patches on the release branches, so please retarget to master branch.

bolshakov-a:
Oh, @DaanDeMeyer, sorry. It seems that you have to revert the changes suggested by me back.

PR #1712, opened 2025-04-13
Add support for the cleanup attribute

### LINKED ISSUE #1710: __attribute__((__cleanup__(x))) support is missing

DaanDeMeyer:
https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-cleanup-variable-attribute

This attribute takes a function which should be tracked by IWYU.

Fixes #1710

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -197,6 +197,7 @@ using clang::CXXTypeidExpr;
 using clang::CallExpr;
 using clang::CastExpr;
 using clang::ClassTemplateSpecializationDecl;
+using clang::CleanupAttr;
 using clang::CompilerInstance;
 using clang::ConceptSpecializationExpr;
 using clang::Decl;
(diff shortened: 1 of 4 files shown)


DaanDeMeyer:
Apologies @bolshakov-a @kimgr I messed up the retargeting so creating a new PR with all the requested changes targeting the master branch.

kimgr:
@DaanDeMeyer Beautiful, thanks!

DaanDeMeyer:
@kimgr Is there a policy for backporting stuff to the clang compat branches? Or is that simply something I have to do locally until there's a new release of llvm?

kimgr:
@DaanDeMeyer Yeah, you'd have to figure it out in your local build, we don't backport to released versions

PR #1714, opened 2025-04-14
Improve Apple Libc mapping generator

mohd-akram:
The POSIX headers have been added to the default mappings so avoid
mapping them incompletely here. This resolves an accidental mapping from
sys/time.h to time.h since both are valid POSIX headers. Also, do not
strip the architecture from the path when it is intentional (eg.
mach-o/reloc.h does not include mach-o/arm/reloc.h). Lastly, do not map
byte_order.h since both forms are deprecated and there's no preference.

### DIFF
--- a/mapgen/iwyu-mapgen-apple-libc.py
+++ b/mapgen/iwyu-mapgen-apple-libc.py
@@ -21,15 +21,7 @@ from operator import itemgetter
 from pathlib import Path
 
 
-PATTERN_MAPPINGS = {
-    'pthread_*': 'pthread.h',
-    'sys/_types/_fd_*': 'sys/select.h',
-    'sys/_types/_in_*': 'netinet/in.h',
-    'sys/_types/_o_*': 'fcntl.h',
-    'sys/_types/_s_*': 'sys/stat.h',
-    'sys/_types/_seek_*': 'stdio.h',
-    'sys/ttycom.h': 'sys/ioctl.h',
-}
+PATTERN_MAPPINGS = {}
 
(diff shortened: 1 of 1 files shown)


kimgr:
I'd like to see this split into two commits with more explanatory commit messages. I can take care of the Python compat patch if you want, but I'm not sure what's going on with the redundant/inaccurate rules. Do you think you could expand a little in the commit message?

Also, do any of these changes address the assertion failure reported here: https://github.com/include-what-you-use/include-what-you-use/pull/1704#issuecomment-2801142565? I think that may be a collision between built-in mappings and generated mappings, but I'm not sure.

mohd-akram:
> I'd like to see this split into two commits with more explanatory commit messages. I can take care of the Python compat patch if you want, but I'm not sure what's going on with the redundant/inaccurate rules. Do you think you could expand a little in the commit message?

Sure, I removed the compat patch to keep it simple and expanded the commit message.

> Also, do any of these changes address the assertion failure reported here: https://github.com/include-what-you-use/include-what-you-use/pull/1704#issuecomment-2801142565? I think that may be a collision between built-in mappings and generated mappings, but I'm not sure.

Yeah, it should address it. The script was erroneously marking `sys/time.h` as private, but that's not correct, which seemed to be causing the collision.

firewave:
I can confirm that this fixes the assert when used with 0.24.

firewave:
> I can confirm that this fixes the assert when used with 0.24.

I also compared the results and it looks good.

PR #1715, opened 2025-04-15
added workflow to test mapping generators in CI

firewave:


firewave:
I think we could limit this to PRs only.

PR #1717, opened 2025-04-15
Avoid full type for assignable from void trait

bolshakov-a:
'void' cannot be assigned to an object, so the complete object type is not needed for proper evaluation of such a trait.

The case when 'void' appears at the LHS should already be discarded by the logic.


kimgr:
Cool, thanks!

PR #1718, opened 2025-04-15
Remove questionable IsSystemIncludeFile

kimgr:
This would return true for special filenames, such as "\<built-in\>", "\<command-line\>", etc.

We have library functions to check for system headers and special files now, so use them directly.

Check for both special files and system headers to preserve semantics, but we might consider only checking IsSystemHeader now.


firewave:
You missed the diff in the dogfood result. Maybe the diff should not be in a `<details>` section for visibility.

kimgr:
@firewave Thanks -- opened https://github.com/include-what-you-use/include-what-you-use/pull/1723 to do that.

PR #1719, opened 2025-04-16
Add mappings for WEOF and winsize

mohd-akram:


kimgr:
The "macOS" in the commit subject rubs me the wrong way, a bit -- these are mappings for GNU libc. I suppose your changes are also valid for GNU libc users?

And now that there's a mapping generator for Apple Libc, it feels wrong to make macOS tweaks to the libc mappings at all. Seems like it would be better to;

* make the mapping generator also produce C++
* fold Apple Libc mappings into `include_picker.cc`
* detect Darwin and dynamically pick Apple Libc instead of GNU libc

All that said, the GNU libc mappings are a messy result of many years of hand-editing, so they probably no longer represent libc accurately. So if these changes improve the status quo across Linux and Mac, I think we should take them, but let's phrase it in terms of GNU libc, not macOS, maybe?

mohd-akram:
Actually, the "macOS" was just to refer to where the problem was encountered, as that's the motivation for the change. It could very well be fixing the same issue on GNU and BSD especially (macOS libc is mostly from FreeBSD, which shares code with other BSDs). The stuff that's in the generator is largely Apple-specific as there's not a ton that wouldn't also be applicable to eg. AIX, Solaris, and BSDs.

> * make the mapping generator also produce C++

For C++ we use the mappings bundled with Clang and haven't found an issue with that yet tbh since Apple uses libc++.

> * fold Apple Libc mappings into `include_picker.cc`
> * detect Darwin and dynamically pick Apple Libc instead of GNU libc

This does not seem needed yet AFAICT. And Apple changes their stuff a lot. Adding more POSIX symbols to `include_picker.cc` and possibly the [xlocale](https://man.freebsd.org/cgi/man.cgi?query=xlocale) APIs which FreeBSD also picked up would confine the generator further to just Apple mappings.

> So if these changes improve the status quo across Linux and Mac, I think we should take them, but let's phrase it in terms of GNU libc, not macOS, maybe?

In this case, I can split out the `ttycom.h` change? That's a BSD-only header AFAICT.

kimgr:
> > make the mapping generator also produce C++
>
> For C++ we use the mappings bundled with Clang and > haven't found an issue with that yet tbh since Apple uses libc++.

Sorry, I meant produce C++ code in addition to the yaml, so the Apple Libc mappings can be built into the include-what-you-use executable.

PR #1721, opened 2025-04-18
Fix copy-pasted test case

bolshakov-a:


bolshakov-a:
Hey @kimgr, something strange appears in the dogfood results...

kimgr:
`<string>` is something I overlooked in https://github.com/include-what-you-use/include-what-you-use/pull/1718#issuecomment-2807235708.

I think the `DiagnosticFrontendInterface.inc` one is something that changed in LLVM recently, and will need an override.

PR #1726, opened 2025-04-18
Add pragmas to keep right DiagnosticFrontend.h include

kimgr:
Clang reorganized the diagnostics headers in
https://github.com/llvm/llvm-project/commit/6c74fe9087fd85059158719ad1ab67e0f5098300

We don't seem to have a rule for .inc files, and I'm not entirely sure what it should be. Add pragmas to keep the public header.

### DIFF
--- a/iwyu_driver.cc
+++ b/iwyu_driver.cc
@@ -22,7 +22,7 @@
 #include <vector>
 
 #include "clang/Basic/Diagnostic.h"
-#include "clang/Basic/DiagnosticFrontend.h"
+#include "clang/Basic/DiagnosticFrontend.h"  // IWYU pragma: keep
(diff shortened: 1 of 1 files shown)


bolshakov-a:
It looks a little strange that that commit was merged two weeks ago, but the IWYU dogfood results didn't contain that change 4 days ago. Probably, the direct cause was another LLVM commit. However, I'm not sure it is important.

kimgr:
Our CI pipeline uses the Debian packages from https://apt.llvm.org, and they have a significant delay vs. LLVM mainline. So I think/hope it adds up.

bolshakov-a:
Got it, thanks!

PR #1727, opened 2025-04-19
Fix sys/ttycom.h mapping

mohd-akram:
Some BSDs define winsize in sys/ttycom.h, so it was originally mapped to
termios.h. A mapping for the winsize symbol has since been added and the
remaining symbols should be imported via sys/ioctl.h. While here, group
it with other system headers that are not present on Linux.

### DIFF
--- a/gcc.libc.imp
+++ b/gcc.libc.imp
@@ -218,11 +218,12 @@
   { "include": [ "<asm/unistd.h>", "private", "<sys/syscall.h>", "public"] },
   { "include": [ "<linux/limits.h>", "private", "<limits.h>", "public"] },   # PATH_MAX
   { "include": [ "<linux/prctl.h>", "private", "<sys/prctl.h>", "public"] },
+  { "include": [ "<sys/ucontext.h>", "private", "<ucontext.h>", "public"] },
+  # System headers available on AIX, BSD, Solaris and other Unix systems
   { "include": [ "<sys/dtrace.h>", "private", "<dtrace.h>", "public"] },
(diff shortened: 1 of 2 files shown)


kimgr:
A little summary of what the fix amounts to would be welcome in the commit message.

There's some parts to it, right?

* all symbols with special semantics are already explicitly mapped (`winsize`, `WEOF`, ...)
* ... and the rest of it is fronted by `sys/ioctl.h`
* Note `sys/ttycom.h` is a FreeBSD libc header

... or something like that.

gcc.libc.imp
```diff
@@ -221,7 +221,7 @@
   { "include": [ "<sys/dtrace.h>", "private", "<dtrace.h>", "public"] },
   { "include": [ "<sys/paths.h>", "private", "<paths.h>", "public"] },
   { "include": [ "<sys/syslimits.h>", "private", "<limits.h>", "public"] },
-  { "include": [ "<sys/ttycom.h>", "private", "<termios.h>", "public"] },
+  { "include": [ "<sys/ttycom.h>", "private", "<sys/ioctl.h>", "public"] },
```
kimgr, line 224: Might be worth adding a comment here that `sys/ttycom.h` is a FreeBSD libc header, so people don't clean it up after finding that the header doesn't exist on a Linux system.

kimgr:
Maybe the FreeBSD libc comment is better placed in the code than the commit message.

mohd-akram:
Good points, thanks. Updated the commit and added a comment.

kimgr:
Thank you!

PR #1729, opened 2025-04-20
Improve "convertible" trait handling

bolshakov-a:
Because `__is_convertible_to` is a synonym for `__is_convertible`, and already has some test coverage, test cases for it have not been added. OTOH, `__is_nothrow_convertible` has also been handled similarly in IWYU, so the test cases for it may be redundant as well...

The change in the `ptr_ref_aliases` test is because the only `IndirectBase` use that was considered as a full use by IWYU was inside `__is_convertible_to` traits.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2146,6 +2146,61 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         }
         return true;
       }
+      case TypeTrait::BTT_IsConvertible:
+      case TypeTrait::BTT_IsConvertibleTo:
+      case TypeTrait::BTT_IsNothrowConvertible: {
+        // No conversion to/from void or to a function type.
+        if (lhs_type->isVoidType() || rhs_type->isVoidType())
+          return true;
+        if (rhs_type->isFunctionType())
+          return true;
+        // Only non-reference-related types can involve user-defined conversions
+        // (C++17 [dcl.init.ref]).
+        if (GetCanonicalType(RemoveReference(lhs_type)) ==
+            GetCanonicalType(RemoveReference(rhs_type))) {
+          return true;
+        }
+        const bool rhs_can_bind_to_temp = RefCanBindToTemp(rhs_type);
+        // If the rhs is a reference that can bind to its class temporary
+        // object, the ctors of the class are to be considered.
+        if (rhs_can_bind_to_temp)
+          ReportTypeUse(CurrentLoc(), rhs_type, DerefKind::RemoveRefs);
+        // Pointer and pointer-to-member conversions should be taken into
+        // account if the rhs type is a pointer or pointer-to-member type
+        // directly, or a reference which can bind to a temporary pointer
+        // object.
+        if (!rhs_type->isReferenceType() || rhs_can_bind_to_temp) {
+          const Type* lhs_deref_type = RemoveReference(lhs_type);
+          const Type* rhs_deref_type = RemoveReference(rhs_type);
+          if (IsDerivedToBasePtrConvertible(lhs_deref_type, rhs_deref_type)) {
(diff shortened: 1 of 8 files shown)


iwyu.cc
```diff
@@ -2146,6 +2146,61 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      case TypeTrait::BTT_IsNothrowConvertible: {
+        // No conversion to/from void or to a function type.
+        if (lhs_type->isVoidType() || rhs_type->isVoidType())
+          return true;
+        if (rhs_type->isFunctionType())
+          return true;
+        // Only non-reference-related types can involve user-defined conversions
+        // (C++17 [dcl.init.ref]).
+        if (GetCanonicalType(RemoveReference(lhs_type)) ==
+            GetCanonicalType((RemoveReference(rhs_type)))) {
```
kimgr, line 2160: Looks like an extra pair of parentheses around `RemoveReference(...)` here.

iwyu.cc
```diff
@@ -2146,6 +2146,61 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        const bool rhs_can_bind_to_temp = RefCanBindToTemp(rhs_type);
+        // If the rhs is a reference that can bind to its class temporary
+        // object, the ctors of the class are to be considered.
+        if (rhs_can_bind_to_temp)
+          ReportTypeUse(CurrentLoc(), rhs_type, DerefKind::RemoveRefs);
+        // Pointer and pointer-to-member conversions should be taken into
+        // account if the rhs type is a pointer or pointer-to-member type
+        // directly, or a reference which can bind to a temporary pointer
+        // object.
+        if (!rhs_type->isReferenceType() || RefCanBindToTemp(rhs_type)) {
```
kimgr, line 2172: Can use `rhs_can_bind_to_temp` instead of invoking `RefCanBindToTemp` again

kimgr:
I wonder if it would be a good idea to move the `is_convertible` cases out of `ptr_ref_aliases.cc` and into `type_trait.cc`? Unrelated to this PR, but maybe worth looking into as a cleanup before or after.

kimgr:
Also, I wanted to ask: these improvements for traits are useful on their own, but are they part of a bigger plan? Any chance they are a foundational building block to help IWYU improve its understanding of STL code?

bolshakov-a:
> I wonder if it would be a good idea to move the is_convertible cases out of ptr_ref_aliases.cc and into type_trait.cc? Unrelated to this PR, but maybe worth looking into as a cleanup before or after.

The primary reason to introduce `ptr_ref_aliases` test was to test that correct `DerefKind`s are used with `ReportTypeUse` calls. But I decided that test cases for type traits may overflow `ptr_ref_aliases` and should better be placed into the `type_trait` case. So maybe it is a good idea to remove them completely from `ptr_ref_aliases` as redundant. I'd prefer to do it after this PR.

> Also, I wanted to ask: these improvements for traits are useful on their own, but are they part of a bigger plan? Any chance they are a foundational building block to help IWYU improve its understanding of STL code?

They are of course related to the STL type traits (because the implementations use these builtins), but I don't know any real world problem which this PR fixes. This work is just for completeness, as we decided [here](https://github.com/include-what-you-use/include-what-you-use/pull/745#issuecomment-2701988676), because no one was able to confirm that `VisitTypeTraitExpr` is not really needed in real-world code (or the opposite). The trait handling looks like a can of worms, however I hope the handling of the remaining trait builtins will be easier.

iwyu.cc
bolshakov-a: Fixed, thanks!

kimgr:
Thanks, LGTM!

PR #1731, opened 2025-04-20
Make sure CodeAuthorWantsJustAForwardDeclare() also considers identical locations

### LINKED ISSUE #1730: Forward declaration of struct with typedef results in the full definition being included
The following code in a C header:

```c
typedef struct Manager Manager;
```

Results in the following output from IWYU:

```
../src/core/dynamic-user.h:12:16: (1) [ RecordDecl ] struct Manager
../src/core/dynamic-user.h:12:24: (1) [ TypedefDecl ] typedef struct Manager Manager
../src/core/dynamic-user.h:12:9: (2, fwd decl) [ ElaboratedTypeLoc ] struct Manager
Found owning elaborated type: ../src/core/dynamic-user.h:12-12: struct Manager
../src/core/dynamic-user.h:12:16: (3, fwd decl) [ RecordTypeLoc ] struct Manager
Marked full-info use of decl Manager (from ../src/core/manager.h:175:8) at ../src/core/dynamic-user.h:12:16
```

This is wrong since the typedef itself is a forward declaration already. The IWYU argument that the header should include Manager in this case doesn't make sense, because we're defining the type to the same name. This is a very common idiom in C to allow using structs without having to prefix the name with `struct` all the time.

I'm willing to write a PR to fix this, but it's really complicated figuring out where to best make the required change. `CodeAuthorWantsJustAForwardDeclare()` does not seem to have sufficient information available. I'm looking at `IsProvidedTypeComponent()` since I can access the parent AST node there, but still not sure if that's the right place to handle this.

DaanDeMeyer:
If the locations are the same, we can also consider one location to
be before the other. This matters in case where an elaborated type
appears on the same line as the record declaration
(typedef struct Manager Manager).

Fixes https://github.com/include-what-you-use/include-what-you-use/issues/1730

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1430,7 +1430,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // Check if the author forward-declared the class in the same file.
     bool found_earlier_forward_declare_in_same_file = false;
     for (const NamedDecl* redecl : redecls) {
-      if (IsBeforeInSameFile(redecl, use_loc)) {
+      // In case of something like "typedef struct Manager Manager", the
+      // redeclaration and the type will have the same location.
+      if (GetLocation(redecl) == use_loc ||
+          IsBeforeInSameFile(redecl, use_loc)) {
         found_earlier_forward_declare_in_same_file = true;
         break;
       }
--- a/tests/c/elaborated_struct-d2.h
+++ b/tests/c/elaborated_struct-d2.h
@@ -1,0 +1,16 @@
+//===--- elaborated_struct-d2.h - test input file for iwyu ----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+typedef struct Struct Typedef;
+
+/**** IWYU_SUMMARY
+
+(tests/c/elaborated_struct-d2.h has correct #includes/fwd-decls)
+
+***** IWYU_SUMMARY */
--- a/tests/c/elaborated_struct.c
+++ b/tests/c/elaborated_struct.c
@@ -7,8 +7,9 @@
 //
 //===----------------------------------------------------------------------===//
 
-// IWYU_ARGS: -I .
+// IWYU_ARGS: -I . -Xiwyu --check_also=tests/c/elaborated_struct-d2.h
 
+#include "tests/c/elaborated_struct-d2.h"
 #include "tests/c/elaborated_struct-d1.h"
 
 // C basically never requires an explicit forward declaration, all uses of
@@ -43,6 +44,7 @@ struct Struct;
 
 tests/c/elaborated_struct.c should remove these lines:
 - #include "tests/c/elaborated_struct-d1.h"  // lines XX-XX
+- #include "tests/c/elaborated_struct-d2.h"  // lines XX-XX
 
 The full include-list for tests/c/elaborated_struct.c:
 struct ForwardDeclared;  // lines XX-XX


bolshakov-a:
Thanks! I agree that it is better to avoid the complete type `#include` for `typedef` underlying type because it can be forward-declared (or de-facto is, in C). However, IWYU has [a specific policy for it](https://github.com/include-what-you-use/include-what-you-use/blob/0.24/docs/WhatIsAUse.md#automatic-re-export-typedefs), stating that if the typedef author places the direct `#include` in the `typedef` defining file, then it should be kept, and the `typedef` should be considered as "providing" the underlying type, so its users should not bother about including the underlying type header. I think it is better to adhere to that policy, at least to avoid a noise from internal stuff of some libraries.

And in the present form, the PR contains a bug that the underlying type is required neither on typedef nor on user side, so you'll end up with uncompilable code.

I think a more correct fix should be, indeed, in `CodeAuthorWantsJustAForwardDeclare` or somewhere near it, but it really seems to be tricky. So I suggest to find out why the elaborated type `struct Manager` is not considered to introduce a forward-declaration _before_ the `Manager` tag type. Some work has already been done to handle elaborated types as fwd-declarations, so you can see `Found owning elaborated type` line in the log.

Besides, you may be interested that there is another problem with typedefs named similarly to their underlying types: #1665.

kimgr:
I think the author intent rule says that if the type is forward-declared at typedef declaration site, the user of the type becomes responsible (i.e. typedef author pushes the responsibility to the user).

It's interesting with C, because _all_ typedef declarations contain a "forward decl" by virtue of being elaborated;
```
typedef struct Foo Foo;
        ^^^^^^^^^^
```

so I wonder if we could have a language-dependent special-case in `CodeAuthorWantsJustAForwardDeclare`. I think the current "found-forward-declare-earlier-in-same-file" probably doesn't hit the elaborated type that is part of the declaration.

kimgr:
Ah, but then I guess we'd need to turn the logic in `CodeAuthorWantsJustAForwardDeclare` upside down, so the presence of an `#include` actually overrides the ever-present forward-decl.

bolshakov-a:
And it seems unclear for me what is specific here for the case of name coincidence. Looks like IWYU behaves similarly whether the names coincide or not, i.e. for the case
```cpp
#include "tests/cxx/direct.h"
typedef class IndirectClass Indirect;
```
IWYU outputs:
```
#include "tests/cxx/indirect.h"  // for IndirectClass

- #include "tests/cxx/direct.h"  // lines 1-1

#include "tests/cxx/indirect.h"  // for IndirectClass
```
which looks undesirable.

> Ah, but then I guess we'd need to turn the logic in CodeAuthorWantsJustAForwardDeclare upside down, so the presence of an #include actually overrides the ever-present forward-decl.

It already does, if I've understood you correctly. If the typedef-defining file contains the direct `#include`, the typedef is considered as "providing" regardless of the presence of fwd-declaration.

kimgr:
@DaanDeMeyer Aha, just saw your comment on the issue:

> I'm willing to write a PR to fix this, but it's really complicated figuring out where to best make the required change. CodeAuthorWantsJustAForwardDeclare() does not seem to have sufficient information available. 

Is it the "decl is part of a typedef" property that's lost at this point?

bolshakov-a:
> so I wonder if we could have a language-dependent special-case in CodeAuthorWantsJustAForwardDeclare.

I think it would be good to fix it in the C++ case as well, because it is also valid code in C++.

kimgr:
@bolshakov-a 
> > so I wonder if we could have a language-dependent special-case in CodeAuthorWantsJustAForwardDeclare.
>
> I think it would be good to fix it in the C++ case as well, because it is also valid code in C++.

Yes. Question is if a typedef declaration with an elaborated type is as strong a signal as a separate forward-decl. I think it might be in the case of typedefs, but e.g. for function decls, the scales might tip differently;
```
typedef class Foo FooType;
void func(class Foo*);
```

I tried some permutations of your C++ example:
```
$ cat tests/1731/t.cc 
#include "tests/cxx/direct.h"
class IndirectClass;
typedef class IndirectClass Indirect;

$ include-what-you-use -I. tests/1731/t.cc 

tests/1731/t.cc should add these lines:

tests/1731/t.cc should remove these lines:
- #include "tests/cxx/direct.h"  // lines 1-1
- class IndirectClass;  // lines 2-2

The full include-list for tests/1731/t.cc:
---
```
Hmm, remove everything, including the forward-decl we just added? OK:
```
$ cat tests/1731/t.cc 
typedef class IndirectClass Indirect;

$ include-what-you-use -I. tests/1731/t.cc 

(tests/1731/t.cc has correct #includes/fwd-decls)
```
At first I thought it would lead to inconsistent results, but it looks like maybe the only issue is that the elaborated type inside typedef is not seen as a forward-decl. If it were, I guess your original example would reduce to my last example.

DaanDeMeyer:
@kimgr So to fix this we have to make sure the elaborated type inside the the typedef is seen as a fwd-decl? I'm a bit out of my league here, any advice on where to fix this would be great. (Or feel free to just do your own PR if you know where this needs to be fixed if that's quicker)

kimgr:
@DaanDeMeyer That's the theory :slightly_smiling_face: -- unfortunately I don't really have a clue how to figure that out inside `CodeAuthorJustWantsAForwardDeclare` either... It might be possible to special-case typedefs wherever it's called and conditionally pass the elaborated type or the canonical type, depending.

bolshakov-a:
@DaanDeMeyer, you can check if this helps in your case:
```patch
--- a/iwyu_location_util.h
+++ b/iwyu_location_util.h
@@ -217,7 +217,7 @@ inline bool IsBeforeInTranslationUnit(const T& a, const U& b) {
   if ((IsInMacro(a_loc) || IsInMacro(b_loc)) &&
       GetInstantiationLoc(a_loc) == GetInstantiationLoc(b_loc))
     return true;
-  return GlobalSourceManager()->isBeforeInTranslationUnit(a_loc, b_loc);
+  return a_loc == b_loc || GlobalSourceManager()->isBeforeInTranslationUnit(a_loc, b_loc);
 }
```
However, it doesn't work when the translation unit contains a declaration of the `typedef`ed class prior to the `typedef`, because the elaborated type doesn't introduce an "owning declaration" in that case.

DaanDeMeyer:
Hmm, so the check is fine, the problem is that when we do the check for struct TypedeffedStruct, `GetTagRedecls()` returns an empty set.

bolshakov-a:
> trivial to test by taking the test case from this PR

No, I meant not from this PR but from your original issue the logs from which are in #1730.

DaanDeMeyer:
> > trivial to test by taking the test case from this PR
> 
> No, I meant not from this PR but from your original issue the logs from which are in #1730.

Ah indeed it doesn't add the include anymore in that case. That's good enough for my use case for now.

bolshakov-a:
> `GetTagRedecls()` returns an empty set.

It returns exactly one item (at least on my machine): either the elaborated type owned declaration, or the full declaration from the `-d1.h` header, if present. And in your original case, I suspect, the complete declaration appears _after_ the `typedef` in the translation unit, so the owned declaration is still present. This is why I've said that my fix doesn't work in all cases: analysis of elaborated type owned declarations, which was adopted in IWYU, is fragile.

> I was a bit concerned about [this -Wvisibility quirk we have for C](https://github.com/include-what-you-use/include-what-you-use/blob/86874f649a117ddeb17e6f68d47bd84cd83acd79/iwyu.cc#L4547), but I guess that's only interesting for use reporting, and we don't need to maintain the same behavior for author intent hints.

Thanks, I forgot about this C language peculiarity. But I think it is irrelevant for the author intent analysis because the fwd-declaration which IWYU suggests to add would be considered as just the second (or first, strictly speaking) fwd-declaration, and the author-intent logic doesn't depend on the number of fwd-declarations: if there is a direct `#include`, no one of them changes the state of affairs.

DaanDeMeyer:
> It returns exactly one item (at least on my machine): either the elaborated type owned declaration, or the full declaration from the -d1.h header, if present. And in your original case, I suspect, the complete declaration appears after the typedef in the translation unit, so the owned declaration is still present. This is why I've said that my fix doesn't work in all cases: analysis of elaborated type owned declarations, which was adopted in IWYU, is fragile.

The difference was that in the test case the include with the full definition was already present and is kept in that case, whereas in the case from my original issue, all that was present is the typedef with the forward declaration and in that case GetTagRedecls() does return a single item as expected with the same location, so your suggestion works and fixes the issue. It means the header is still kept if it's already included, but that might also be OK depending on how you interpret the IWYU policy.

I've reworked the PR now to just implement your fix. I'll leave any bigger changes for later as I'm too familiar with IWYU to be able to implement those without spending a lot of time on this.

bolshakov-a:
The name of the `IsBeforeInTranslationUnit` function should be changed now. Or, maybe better, the check for equality should be placed in `CodeAuthorWantsJustAForwardDeclare`, to avoid any unexpected influence to the `ReportExplicitInstantiations` method and other functions. Not sure.

And a test case is worth writing, despite it is somewhat tricky, i.e. the file structure should be like this:
```c
// ...-d1.h
struct Struct{};

// ...-d2.h
typedef struct Struct Typedef;

// test.c
#include "...-d2.h"
#include "...-d1.h"  // Note the order of includes

// IWYU_ARGS: -I . -Xiwyu --check_also=...-d2.h
```

kimgr:
It feels a little icky to use only location info to fix this... But the patch is nice and clean, so I guess let's go with it for now. Thanks!

PR #1732, opened 2025-04-21
Use consistent naming in C tests

kimgr:
The C tests have been ambivalent around naming; some use IWYU-style PascalCase names and some use hacker-style snake_case names.

While the latter looks like nicer C style in general, prefer IWYU-style names to make it easier to move/copy details between C and C++ suites, and to avoid name collisions with standard library types.


tests/c/var_def_is_use-i1.h
```diff
@@ -7,4 +7,4 @@
 //
 //===----------------------------------------------------------------------===//
 
-extern int extern_global;
+extern int ExternGlobal;
```
bolshakov-a, line 10: Do you think global variables should also be in PascalCase? The C++ tests use snake_case, like this:
https://github.com/include-what-you-use/include-what-you-use/blob/182147d52d8c7e610c03dce83b67185fe2ba858b/tests/cxx/extern_variable.cc#L17

kimgr: :facepalm: Thanks! I got a little over-excited.

kimgr: Reverted

PR #1736, opened 2025-04-24
Improve reference to temporary trait handling

bolshakov-a:
These cases share the main coded logic with "convertible" traits, so they have not been thoroughly tested.

A test case has been added for "convertible" traits to nail down the difference from "reference to temporary" ones in function reference type handling.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2149,6 +2149,22 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         }
         return true;
       }
+      case TypeTrait::BTT_ReferenceBindsToTemporary:
+      case TypeTrait::BTT_ReferenceConstructsFromTemporary:
+      case TypeTrait::BTT_ReferenceConvertsFromTemporary:
+        // Seems like __reference_binds_to_temporary differs from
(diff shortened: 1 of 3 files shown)


iwyu.cc
```diff
@@ -2149,6 +2149,22 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        // type, whereas __reference_converts_from_temporary considers whether
+        // a conversion function is not explicit. All of this is irrelevant to
+        // IWYU because the complete type is requested in either case.
+        if (!RefCanBindToTemp(lhs_type))
+          return true;
+        // No temporaries of function type.
+        if (RemoveReference(lhs_type)->isFunctionType())
+          return true;
+        swap(lhs_type, rhs_type);
+        [[fallthrough]];
```
kimgr, line 2167: Elegant!

bolshakov-a: Not so sure...

kimgr: It's just so exceedingly rare that `swap` expresses exactly what you want to do :slightly_smiling_face:

PR #1739, opened 2025-04-26
[ci] Remove packaging workaround for llvm-21

kimgr:
The packages now seem to be internally consistent.


kimgr:
Hmm? The dogfood results are surprising, but have nothing to do with this change.

Issue #1740, opened 2025-04-28
improve out-of-the-box LLVM selection in CMake


firewave:
Currently the lookup of LLVM/Clang in CMake is performed without a version:

```
  find_package(LLVM CONFIG REQUIRED)
  find_package(Clang CONFIG REQUIRED)
```

On my system I have lots of LLVM/Clang versions installed which leads to the following:

```
-- IWYU: configuring for LLVM 10.0.0 from /usr/lib/llvm-10/cmake
```

This obviously will not compile so I had to adjust the lookup to ad the current version

```
  find_package(LLVM 20.1 CONFIG REQUIRED)
  find_package(Clang 20.1 CONFIG REQUIRED)
```

which will select the desired version

```
-- IWYU: configuring for LLVM 20.1.4 from /usr/lib/llvm-20/cmake
```

This is not optimal and should be improved.

kimgr:
You don't mention what your `cmake` command looks like. As the README says [here](https://github.com/include-what-you-use/include-what-you-use/#:~:text=%23%20For%20IWYU%200.11/Clang%207%20and%20later), you have to provide a `CMAKE_PREFIX_PATH` to tell it what you intend.

While it might be convenient to ignore that, it's not going to work consistently.

firewave:
> You don't mention what your `cmake` command looks like.

I have no additional options specified.

> As the README says [here](https://github.com/include-what-you-use/include-what-you-use/#:~:text=%23%20For%20IWYU%200.11/Clang%207%20and%20later), you have to provide a `CMAKE_PREFIX_PATH` to tell it what you intend.

Having this specified in CMake removes the need of documenting this and the user having to manually looking up a path (it might be different on each distro) and specifying it manually. That logic exists in the CMake extension for a reason.

Also don't you usually specify the location for packages via something like `*_ROOT`?

kimgr:
The `CMAKE_PREFIX_PATH` variable, as I understand it, is a search path for `find_packages`. If you don't constrain it, it will search pretty freely, as you've seen.

> Having this specified in CMake removes the need of documenting this and the user having to manually looking up a path (it might be different on each distro) and specifying it manually. That logic exists in the CMake extension for a reason.

My exposure to different distros is limited, but I'm thinking something like:
```
cmake -G Ninja -DCMAKE_PREFIX_PATH="$(llvm-config --prefix)" ../../include-what-you-use
```
would be a normal way to avoid manually looking up paths.

[...]

PR #1741, opened 2025-04-28
bail out in CMake when minimum required LLVM version is not satisfied

firewave:
refs #1740


kimgr:
I want to do as little as humanly possible, but I want to leave the code base open to experiments.

PR #1742, opened 2025-04-28
Always use "use location" as canonical use location for a decl

### LINKED ISSUE #1737: iwyu marks full use of function used in macro but doesn't add include for it to header
```

### PR DESCRIPTION
**Pull request by DaanDeMeyer**
Currently, the headers that IWYU suggests to add to a header file defining a macro depend on whether that macro is used in the corresponding implementation file that causes the header file to be analyzed. So if a file a.h defines a macro that uses various declarations and the corresponding file a.c does not use that macro, then IWYU will suggest removing any headers that provide those decls from a.h. On the other hand, if a.c does use the macro, then IWYU will suggest keeping the headers.

This means that if a.c does not use the macro, then any other files that include a.h and use the macro will end up with missing header files as the required header files to use the macro will not be added to a.h and will not be added to the file using the macro either, causing compilation failures.

To fix this problem, let's always use the "use location" for decls within a macro as the canonical location, so that the required headers are always added to the file that's expanding the macro.

Fixes #1737

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -135,7 +135,6 @@
 #include "iwyu_port.h"  // for CHECK_
 #include "iwyu_preprocessor.h"
 #include "iwyu_stl_util.h"
-#include "iwyu_string_util.h"
 #include "iwyu_use_flags.h"
 #include "iwyu_verrs.h"
 #include "llvm/ADT/ArrayRef.h"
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by DaanDeMeyer** [comment]
The test doesn't quite work yet but I'm not exactly sure what's missing. For some reason it's not suggesting to include the indirect header into macro.c

**Comment by DaanDeMeyer** [comment]
```
➜  include-what-you-use git:(clang_20) ✗ mkosi sandbox -- build/bin/include-what-you-use -Xiwyu --verbose=6 -I. tests/c/macro.c
Setting verbose-level to 6
Search path: /home/daandemeyer/projects/include-what-you-use/ (user)
Search path: /usr/lib/clang/20/include/ (system)
Search path: /usr/local/include/ (system)
Search path: /usr/include/ (system)
Found include: <built-in>:-1 -> tests/c/macro.c
Added to main compilation unit: tests/c/macro.c
[ Exiting to  ] tests/c/macro.c:1:1
[ #include    ] "tests/c/macro-d1.h" (tests/c/macro-d1.h)
Found include: tests/c/macro.c:15 -> tests/c/macro-d1.h
Added an #include: tests/c/macro.c -> "tests/c/macro-d1.h"
Found include: tests/c/macro-d1.h:10 -> tests/c/macro-i1.h
[ Exiting to  ] tests/c/macro.c:16:1
[ Use macro   ] tests/c/macro.c:18:3: MACRO (from tests/c/macro-d1.h:12:9)
Marked full-info use of macro MACRO (from tests/c/macro-d1.h) at tests/c/macro.c:18:3
Intends-to-provide for tests/c/macro.c:
   tests/c/macro.c
   tests/c/macro-d1.h
Intends-to-provide for tests/c/macro-d1.h:
   tests/c/macro-d1.h
   tests/c/macro-i1.h
Intends-to-provide for tests/c/macro-i1.h:
   tests/c/macro-i1.h
tests/c/macro.c:17:6: (1) [ FunctionDecl ] void cleanup(void)
tests/c/macro.c:17:1: (2) [ FunctionProtoTypeLoc ] void (void)
tests/c/macro.c:17:1: (3) [ BuiltinTypeLoc ] void
tests/c/macro.c:17:20: (2) [ CompoundStmt ] CompoundStmt 0x33fae708
`-CallExpr 0x33fae6b8 'void'
  |-ImplicitCastExpr 0x33fae6a0 'void (*)(int)' <FunctionToPointerDecay>
  | `-DeclRefExpr 0x33fae628 'void (int)' Function 0x33fae3c0 'myfunc' 'void (int)'
  `-IntegerLiteral 0x33fae648 'int' 5

Marked full-info use of decl myfunc (from tests/c/macro-i1.h:10:6) at tests/c/macro.c:18:3 <Spelling=./tests/c/macro-d1.h:12:18>
tests/c/macro.c:18:3 <Spelling=tests/c/macro.c:18:9>: (4) [ IntegerLiteral ] IntegerLiteral 0x33fae648 'int' 5

Marked full-info use of decl myfunc (from tests/c/macro-i1.h:10:6) at tests/c/macro.c:18:3 <Spelling=./tests/c/macro-d1.h:12:18>
--- Calculating IWYU violations for tests/c/macro.c ---
Mapped tests/c/macro-d1.h to "tests/c/macro-d1.h" for MACRO (only candidate)
Ignoring full use of MACRO (tests/c/macro.c:18:3): #including dfn from "tests/c/macro-d1.h"

(tests/c/macro.c has correct #includes/fwd-decls)
```

This is the output I get, for some reason only the `MACRO` use is still intact by the time violations are calculated? So the `myfunc` use is attributed to the wrong file maybe?

**Comment by bolshakov-a** [comment]
Seems like the `used_loc` is in the macro, so `GetCanonicalUseLocation` is still needed but in the truncated form like:
```cpp
if (!use_loc.isMacroID())
      return use_loc;

SourceManager* sm = GlobalSourceManager();
return sm->getExpansionLoc(use_loc);
```

DaanDeMeyer:
> Seems like the `used_loc` is in the macro, so `GetCanonicalUseLocation` is still needed but in the truncated form like:
> 
> ```c++
> if (!use_loc.isMacroID())
>       return use_loc;
> 
> SourceManager* sm = GlobalSourceManager();
> return sm->getExpansionLoc(use_loc);
> ```

Thank you! Pushed another version

DaanDeMeyer:
Removed new test since the existing cxx tests cover this pretty well already. I adapted those to these changes instead.

kimgr:
This change makes me a little nervous -- `GetCanonicalUseLocation` has been tuned over years to try and suggest the right side of the macro boundary.

> So if a file a.h defines a macro that uses various declarations and the corresponding file a.c does not use that macro, then
> IWYU will suggest removing any headers that provide those decls from a.h. On the other hand, if a.c does use the macro, then
> IWYU will suggest keeping the headers.

Yes. But I would venture to guess that most files defining macros are freestanding headers, not typical associated headers part of a component (in the `a{.h,.c}` sense). Maybe the scales tip slightly differently in C vs. C++.

The reduced example posted in the issue (https://github.com/include-what-you-use/include-what-you-use/issues/1737#issuecomment-2832004206) demonstrate some annoying behaviors in IWYU, but I think of them as "well, just don't do that" examples:

* don't use `--check_also=macros.h` if `macros.h` contains unexpanded macros
* don't use `include-what-you-use -xc macros.h` if `macros.h` contains unexpanded macros

All that said, I'm curious about the effects. All changes in the test suite look worse to me, but that's because we mostly cover the interesting cases, and they may constitute <0.01% of all macro uses. And if we can guarantee that there are zero conflicting/macro-state-dependent behaviors with this simpler model, that might be a good reason to dumb it down.

bolshakov-a:
> But I would venture to guess that most files defining macros are freestanding headers, not typical associated headers part of a component (in the a{.h,.c} sense).

Freestanding headers are also worth checking with IWYU. Maybe, checking headers is even more important than checking source files...

DaanDeMeyer:
> Yes. But I would venture to guess that most files defining macros are freestanding headers, not typical associated headers part of a component (in the a{.h,.c} sense). Maybe the scales tip slightly differently in C vs. C++.

In systemd this is absolutely not the case. I did a quick search and in 188 of our 1040 header files we define function like macros. Almost all of those files have an associated source file.

> All that said, I'm curious about the effects. All changes in the test suite look worse to me, but that's because we mostly cover the interesting cases, and they may constitute <0.01% of all macro uses. And if we can guarantee that there are zero conflicting/macro-state-dependent behaviors with this simpler model, that might be a good reason to dumb it down.

With this change I get no compilation failures in systemd anymore after running IWYU that are due to missing header includes for stuff in macros.

The one thing we could potentially do when checking the corresponding source file of the header file defining the macro is check if the required header to use the macro is already in the header file defining macro and if it is, keep it there. That way you keep the existing headers in place without requiring pragmas but you also don't end up with uncompilable code after running IWYU.

DaanDeMeyer:
> The one thing we could potentially do when checking the corresponding source file of the header file defining the macro is check if the required header to use the macro is already in the header file defining macro and if it is, keep it there. That way you keep the existing headers in place without requiring pragmas but you also don't end up with uncompilable code after running IWYU.

Frankly thinking more about I still don't like this because it seems like an extremely arbitrary rule to decide which includes we keep or not.

There's also an argument to be made that moving includes to the implementation helps keep headers lean and mean. The only reason macros are defined in the macro is because they have to be, yet by including the necessary headers in the same header, we force all source files to parse those headers, even if they don't use the macro. Moving the includes to the source files means only those source files that use the macro pay the cost of including the necessary header.

bolshakov-a:
> Frankly thinking more about I still don't like this because it seems like an extremely arbitrary rule to decide which includes we keep or not.

Moreover, that way you would introduce a dependency of a header interface from the associated source file implementation details (whether the macro from the header is expanded in it or not), and the change of those details could break the compilation of your project.

DaanDeMeyer:
Hmm so now that I'm actually starting to prepare the IWYU results to start submitting PRs, I'm of course running into the issue that I don't want all the headers added for the expanded macros after all. Yet I don't want to export the headers from the header defining the macro either.

So the only remaining solution seems to be keep pragmas in the header defining the macro.

The problem with the current solution is that IWYU will still always suggest including the headers in the files where the macro is expanded.

But the existing situation where the headers added to the header file defining the macro depend on whether the macro is used in the corresponding source file or not still seems rather bad.

Maybe I'd actually prefer to go the opposite way then and have IWYU not add any header files for expanded macros at all, neither to the header file defining the macro nor the source file where the macro is expanded.

But I suppose this is not a change that we can make unless it's gated behind an option;

bolshakov-a:
> So the only remaining solution seems to be keep pragmas in the header defining the macro.
>
> The problem with the current solution is that IWYU will still always suggest including the headers in the files where the macro is expanded.

Do you mean even with the pragmas? `pragma export` was intended to be used for such cases.

DaanDeMeyer:
> > So the only remaining solution seems to be keep pragmas in the header defining the macro.
> > The problem with the current solution is that IWYU will still always suggest including the headers in the files where the macro is expanded.
> 
> Do you mean even with the pragmas? `pragma export` was intended to be used for such cases.

@bolshakov-a I think pragma export is the wrong solution for such cases. I don't want every header that's used by the macro to be exported, since that means if any symbol from the exported headers is used directly in the file that expands the macro, IWYU won't add an explicit include for it, which seems wrong to me.

kimgr:
OK, I'm a little lost as to what would be a good next step.

It seems like dumbing down `GetCanonicalUseLocation` will cause a lot of noisy includes on the expansion side of things.

And its current behavior is causing a lot of false removals on the macro definition side.

It sounds like an new command-line option is the only recourse, but I can't even imagine what to call it/how it would work... What are the semantics we're looking for?

It seems to me the problem is that when running something like `include-what-you-use -xc macros.h` (from the small repro), IWYU doesn't see what's happening inside the macros, so it can't really know which includes are necessary. And I'm not sure if we can do anything about that.

DaanDeMeyer:
Yeah I don't have a good solution here either, let's close this as it doesn't seem like a good way forward.

PR #1746, opened 2025-04-30
Add some missing mappings to the default mappings

DaanDeMeyer:
I found these trying to get IWYU running on systemd's codebase


kimgr:
Nominally looks good, but I don't have some of the headers on my Ubuntu 24.04 system. My GNU libc version is 2.39.

Is yours older or newer?
```
grep -Er -A1 "#\s*define\s+__GLIBC__" /usr/include
```

DaanDeMeyer:
Much newer, I'm running Fedora rawhide.

kimgr:
Thanks, that explains it. Unfortunately these mapping definitions exist in two places: can you also mirror the changes in `include_picker.cc`, `libc_include_map`: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_include_picker.cc#L346?

DaanDeMeyer:
> Thanks, that explains it. Unfortunately these mapping definitions exist in two places: can you also mirror the changes in `include_picker.cc`, `libc_include_map`: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_include_picker.cc#L346?

Done

PR #1748, opened 2025-04-30
Make sure C arrays result in includes and not a forward declaration

### LINKED ISSUE #1747: IWYU forward declares type used in array which causes compilation failure
If I have the following declaration in a header (C):

```
int sd_dhcp_server_set_lpr(sd_dhcp_server *server, const struct in_addr lpr[], size_t n);
```

Then IWYU chooses to forward declare `struct in_addr;`, which then causes the following compilation failure:

```
../src/systemd/sd-dhcp-server.h:72:47: error: array has incomplete element type 'const struct in_addr'
   72 |                 const struct in_addr addresses[],
      |                                               ^
../src/systemd/sd-dhcp-lease.h:30:8: note: forward declaration of 'struct in_addr'
   30 | struct in_addr;
```

DaanDeMeyer:
If an array type is used in a function declaration, the full type has to be available, and not just a forward declaration.

Fixes #1747

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2673,6 +2673,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  bool VisitArrayType(ArrayType* type) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+    if (!compiler()->getLangOpts().CPlusPlus) {
+      const Type* elem_type = type->getElementType().getTypePtr();
+      ReportTypeUse(CurrentLoc(), elem_type, DerefKind::None);
+    }
+    return true;
+  }
+
   //------------------------------------------------------------
   // Visitors of attributes.
 
--- a/tests/c/array.c
+++ b/tests/c/array.c
@@ -1,0 +1,35 @@
+//===--- array.c - test input file for iwyu -------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+#include "tests/c/direct.h"
+
+// IWYU: Indirect is defined in...*indirect.h
+// IWYU: Indirect needs a declaration
+void array(const struct Indirect array[]);
+
+// IWYU: Indirect is defined in...*indirect.h
+typedef struct Indirect (*ArrayPtr)[5];
+
+// IWYU: Indirect is defined in...*indirect.h
+extern const struct Indirect extern_array[];
+
+/**** IWYU_SUMMARY
+
+tests/c/array.c should add these lines:
+#include "tests/c/indirect.h"
+
+tests/c/array.c should remove these lines:
+- #include "tests/c/direct.h"  // lines XX-XX
+
+The full include-list for tests/c/array.c:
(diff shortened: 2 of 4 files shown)


bolshakov-a:
> If an array type is used in a function declaration

Not only in a function declaration but also in an `extern` variable declaration, `typedef` declaration, under a pointer, or wherever:
```
$ ./clang -fsyntax-only -x c -
typedef struct Undef (*ArrayPtr)[5];

<stdin>:1:33: error: array has incomplete element type 'struct Undef'
    1 | typedef struct Undef (*ArrayPtr)[5];
      |                                 ^
<stdin>:1:16: note: forward declaration of 'struct Undef'
    1 | typedef struct Undef (*ArrayPtr)[5];
      |                ^
1 error generated.

$ ./clang -fsyntax-only -x c++ -
typedef struct Undef (*ArrayPtr)[5];

$
```
 Some of these cases are probably worth being placed in the test as well.

Also, please recall that the conventional way of writing tests for IWYU is to place the used stuff in an `-indirect.h` header to produce explicit diagnostics.

iwyu.cc
```diff
@@ -2673,6 +2674,22 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
 
+  bool VisitArrayType(ArrayType* type) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+    if (!compiler()->getLangOpts().CPlusPlus)
+      ReportTypeUse(CurrentLoc(), &*type->getElementType(), DerefKind::None);
+    return true;
+  }
+
+  bool VisitIncompleteArrayType(IncompleteArrayType *type) {
```
bolshakov-a, line 2685: What is it needed for? `IncompleteArrayType` is derived from `ArrayType`, so the previous handler should be sufficient.

DaanDeMeyer: nvm turns out this isn't necessary after all

tests/c/array.c
```diff
@@ -0,0 +1,35 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+#include "tests/c/array-direct.h"
+
+// IWYU: Struct is defined in...*array-indirect.h
+// IWYU: Struct needs a declaration
+void array(const struct Struct array[]);
```
kimgr, line 16: Since you just want an arbitrary struct type, you can use the reusable `struct Indirect` from `tests/c/direct.h`/`tests/c/indirect.h`.

I wrote up some design guidelines for tests: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#regression-tests, they may be useful.

iwyu.cc
```diff
@@ -2673,6 +2673,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return true;
   }
 
+  bool VisitArrayType(ArrayType* type) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+    if (!compiler()->getLangOpts().CPlusPlus)
+      ReportTypeUse(CurrentLoc(), &*type->getElementType(), DerefKind::None);
```
kimgr, line 2680: I think this would be clearer with a named variable;
```
const Type* elem_type = type->getElementType().getTypePtr();
ReportTypeUse(CurrentLoc(), elem_type, DerefKind::None);
```

kimgr:
Looks great, thanks!

bolshakov-a:
@DaanDeMeyer, thanks for the help with improving IWYU!

PR #1749, opened 2025-05-01
Add all reproducible bugs to bugs suite

kimgr:
Now that the bugs suite exists to do negative regression testing
(i.e. notice when bugs are fixed), add repro cases for all bugs where
I've been able to produce a fully contained repro, or one that only
depends on libc, libstdc++ or libc++.

So far these are probably only reproducible in a Linux environment with
GCC and Clang available.

Once this is merged, our CI environment will fail the build if a PR
fixes a bug -- not that there's anything wrong with that, but we need to
make sure it comes with fully reduced regression tests. Once that's
taken care of, we can remove the bugs test and close the bug.

Roughly 60 bugs are getting test coverage with this patch.

### DIFF
--- a/tests/bugs/1000/1000.cc
+++ b/tests/bugs/1000/1000.cc
@@ -1,0 +1,21 @@
+//===--- 1000.cc - iwyu test ----------------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_XFAIL
+#include <cmath>
+
+static void f() {
+  (void)std::abs(-12);
+}
+
+/**** IWYU_SUMMARY
+
+(tests/bugs/1000/1000.cc has correct #includes/fwd-decls)
+
+***** IWYU_SUMMARY */
--- a/tests/bugs/1061/1061.cc
+++ b/tests/bugs/1061/1061.cc
@@ -1,0 +1,22 @@
+//===--- 1061.cc - iwyu test ----------------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_XFAIL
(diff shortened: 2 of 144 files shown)


kimgr:
cc @bolshakov-a @firewave

kimgr:
Two unexpected passes. I'm guessing it's because I run a newer Ubuntu with newer libstdc++ on my local system.

bolshakov-a:
Wow, well done!

firewave:
es muy loco...

firewave:
> * I'd like to run these in CI, but separate from the normal test suite (e.g. behind a flag to `run_iwyu_tests.py`), so we can see when bugs are fixed

The matrix could have two different approaches. Either use a set of specific compiler/library versions or a set of distros.

> * (for now, I only "allow" standard library deps, but we could extend this to popular libraries, e.g. Qt)

For libraries that could be integrated as the "smoketest" step into the mapgen job proposed in #1715.

kimgr:
Original PR description:

> Here's a draft for an idea I had -- create a suite of all reported bugs containing a contained-enough repro case to run on a system with libc++ and libstdc++ installed.
>
> I've triaged every single open bug, and extracted 63 test cases (also have some other interesting statistics, but that's for later)
>
> Here's a plan:
>
> * All test cases are marked as `// IWYU_XFAIL`, so they will fail if a bug is fixed to signal that the issue needs to be closed
> * I'd like to run these in CI, but separate from the normal test suite (e.g. behind a flag to `run_iwyu_tests.py`), so we can see when bugs are fixed
> * These tests have more lax requirements than normal test cases. As long as we can reasonably set up an environment to run them, library dependencies are OK (for now, I only "allow" standard library deps, but we could extend > this to popular libraries, e.g. Qt)
> * When a bug-test is fixed, it should be covered by library-independent tests for the C++ constructs that triggered the original bug. That way we keep the main test suite fast and portable, but can still add test coverage for pending > bugs
> * Probably need a little CONTRIBUTING note on how to think about, and write tests for this suite
>
> This PR is very big, and very unpolished, but I think it gets the idea across. Let me know what you think about this.

kimgr:
cc @graveljp I've subsumed some of your testcases from https://github.com/graveljp/iwyu-bugs here (the ones linked to from IWYU issues). Let me know if this makes sense to you as a general framework.

PR #1752, opened 2025-05-02
Don't look for C++ stdlib if we're compiling C code

DaanDeMeyer:
Otherwise we end up mapping from `<string.h>` to `<cstring>`.

### DIFF
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -450,8 +450,9 @@ static CStdLib DeriveCStdLib() {
   return CStdLib::Glibc;
 }
 
-static CXXStdLib DeriveCXXStdLib(const ToolChain& toolchain) {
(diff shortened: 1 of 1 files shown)


kimgr:
Very nice! Cool that we can now let the compiler flags/state affect which built-in mappings are chosen.

Do you think you could add a new test to the tests/c suite that checks the string.h vs. cstring quality? Let me know if you need a hand and I can sketch something up. 

Might be a good idea to test it first without the patch to make sure it fails/succeeds for the right reasons.

kimgr:
I'll give it a shot. I think we even have a bug logged for c++ c headers in c mode, but I'm afk now, so will check maybe tomorrow.

kimgr:
I haven't managed to repro either. I'll see if I can find some time to try a little harder, but otherwise let's just move on with it, I guess.

PR #1754, opened 2025-05-02
[clang compat] Use new CompilerInstance constructor

kimgr:
Clang removed setInvocation in favor of a constructor parameter in:
https://github.com/llvm/llvm-project/commit/b69dcb873476cd8e7d3f6f9ffd5b6d0bbe1a3a17

Adjust our only use in IWYU driver code.

### DIFF
--- a/iwyu_driver.cc
+++ b/iwyu_driver.cc
@@ -389,8 +389,8 @@ bool ExecuteAction(int argc,
   // FIXME: This is copied from cc1_main.cpp; simplify and eliminate.
 
   // Create a compiler instance to handle the actual work.
-  unique_ptr<CompilerInstance> compiler(new CompilerInstance);
-  compiler->setInvocation(invocation);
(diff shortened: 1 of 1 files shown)


kimgr:
This is a little prescient, but I noticed that clang/llvm packages for 22.04 already contain the clang change, whereas clang/llvm packages for 24.04 has not yet been released (the debian package build is huge, and takes time).

So I'm going to let this sit until tomorrow, and hopefully we should see a build failure on `master` to confirm I have this right, and hopefully I'll also be able to reproduce and test this locally before merging.

kimgr:
Nightly build failed: https://github.com/include-what-you-use/include-what-you-use/actions/runs/14807183386

kimgr:
The Ubuntu 24.04 LLVM package builds are failing, so won't be able to test on that platform. But the Clang change appears to be on main still, so let's merge this so our master builds green on our CI.

PR #1755, opened 2025-05-05
Handle array references in type traits

bolshakov-a:
Due to array to pointer implicit conversion, arrays may be used where pointers to their elements are acceptable. Direct or typedefed array types in type traits should currently be reported unconditionally, because `CanForwardDeclareType` returns false for them. This change adds handling of arrays under reference types.

`ASTContext::getAsArrayType` desugars the given type to the array type and merges the cv-qualifier with the one of the array element type.

One comment has been corrected because `ReportTypeUse` performs reporting not only for record types but also for elements of arrays of them.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2116,10 +2116,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
                           DerefKind::None);
             return true;
           }
-          // Otherwise, report the rhs record type (ReportTypeUse doesn't report
-          // other kinds of types) because they may have a conversion operator
-          // to the lhs type.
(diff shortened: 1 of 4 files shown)


kimgr:
Phew! The patch is very hard to follow with so many small type edits, I wonder if it would be possible/better to do the `Type*` -> `QualType` as a refactor first, and then introduce the array handling on top of that?

bolshakov-a:
Ok, sure.

bolshakov-a:
I've opened #1756.

kimgr:
Merged #1756 -- please rebase

kimgr:
Thank you, looks good!

bolshakov-a:
Thank you!

PR #1762, opened 2025-05-30
Add array "constructible" trait handling

bolshakov-a:
Since C++20 (apparently after [P0960](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0960r3.html)), "constructible" traits may evaluate to true when testing array construction. This changeset adds reporting of types on which the result depends in such cases.

As for the case of ctors with multiple args handling of which was added earlier, firstly the argument types are checked whether they might be converted to the array element type. For example:
```cpp
__is_constructible(int[2], Class&, int*)
```
is always false regardless of whether `Class` has a conversion function to `int` because `int*` cannot be implicitly converted to `int`, hence the complete `Class` type is not needed. So as not to write all the cases by hand, `IsConvertible` function has been introduced, the logic of which has been taken from the clang source. Some of its code (`EnterExpressionEvaluationContext`, `sfinae_trap.hasErrorOccurred()` check) is not covered by tests and has been added only for generality. It seems that it makes sense only with record types for which this function is not used currently. (Maybe, a unit test is worth writing somehow?) OTOH, `ContextRAII` has not been used because `Sema` should already be in the global (translation unit) context at the time of the IWYU analysis.


iwyu_ast_util.h
```diff
@@ -902,6 +902,14 @@ bool IsBaseToDerivedMemPtrConvertible(const clang::Type* base_mem_ptr_type,
                                       const clang::Type* derived_mem_ptr_type,
                                       clang::Sema&);
 
+// Checks whether an object of the type 'to' may be copy-initialized from
+// an object of the type 'from'. SourceLocation must be valid, otherwise clang
+// may crash on a class template specialization.
+bool IsConvertible(clang::QualType from,
+                   clang::QualType to,
+                   clang::SourceLocation,
```
kimgr, line 910: It would be nice to give this parameter a name for documentation purposes. It looks like `from_loc`, based on how it's used in the impl, but maybe there's a better name.

bolshakov-a: Yeah, the problem may appear when `from` is an uninstantiated class template, but "conversion location" looks more honest with regard to what is passed from `VisitTypeTraitExpr`. It is quite sad that a location is needed for the function that should be location-agnostic...

kimgr:
Thanks, and sorry for the delay!

bolshakov-a:
No problem, thank you!

PR #1768, opened 2025-06-12
Fix member pointer class alias handling

bolshakov-a:
`GetCanonicalType` removes the sugar, hence information that the class type was provided by a typedef was lost.


kimgr:
Thank you!

PR #1769, opened 2025-06-13
[clang compat] Ignore type trait deprecation

bolshakov-a:
https://github.com/llvm/llvm-project/commit/208e3b09564ea7ab2b10004742e553938a2dc60f deprecated `__reference_binds_to_temporary` builtin type trait, so the warnings appeared on running the test.


kimgr:
Thanks!

PR #1770, opened 2025-06-14
Handle __is_trivially_constructible builtin trait

bolshakov-a:
Some more non-trivial code added.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2334,6 +2334,149 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         }
         return true;
       }
+      case TypeTrait::TT_IsTriviallyConstructible: {
+        if (lhs_type->isRecordType()) {
+          // Seems like __is_trivially_constructible may return true only for
+          // default and copy construction before C++20. In C++20, apparently
+          // after P0960, it may return true for aggregate initialization even
+          // with multiple argument types. It is better to report the types even
+          // in the older language modes to avoid possible UB in user's code
+          // after switching to a newer standard.
+          // Unions cannot have multiple initializer expressions.
+          if (expr->getNumArgs() > 2 && lhs_type->isUnionType())
+            return true;
+          // Check arguments.
+          for (const TypeSourceInfo* arg : expr->getArgs()) {
+            if (arg->getType()->isVoidType())
+              return true;
+          }
+          // Report arguments.
+          for (const TypeSourceInfo* arg : expr->getArgs()) {
+            const Type* arg_type = arg->getType().getTypePtr();
+            // Unions don't take part in inheritance, and user-defined
+            // conversions of unions are not trivial operations.
+            if (RemovePointersAndReferences(arg_type)
+                    ->isStructureOrClassType()) {
+              ReportTypeUse(CurrentLoc(), arg_type,
+                            DerefKind::RemoveRefsAndPtr);
+            }
+          }
+          return true;
+        }
+
+        // The trait may return true in C++20 for the array aggregate init.
+        if (const ArrayType* array_type = lhs_type->getAsArrayTypeUnsafe()) {
+          if (const auto* const_array_type =
+                  dyn_cast<ConstantArrayType>(array_type)) {
+            if (const_array_type->getZExtSize() + 1 < expr->getNumArgs())
+              return true;
+          }
+          const QualType elem_type = array_type->getElementType();
+          // There are 3 kinds of trivially constructible arrays which should be
+          // considered: array of classes, of pointers to classes, and of member
+          // pointers.
+          // Class type elements may be trivially initialized from arguments
+          // of the same or a derived type. Unions cannot have derived types.
+          if (elem_type->isStructureOrClassType()) {
+            // Check arguments.
+            for (const TypeSourceInfo* arg : drop_begin(expr->getArgs())) {
+              if (!RemoveReference(arg->getType())->isStructureOrClassType())
+                return true;
+            }
+            // Report arguments.
+            for (const TypeSourceInfo* arg : drop_begin(expr->getArgs())) {
+              const Type* arg_type = arg->getType().getTypePtr();
+              ReportTypeUse(CurrentLoc(), arg_type, DerefKind::RemoveRefs);
+            }
+            return true;
+          }
+          // Any object pointer is convertible to void*, no full type needed.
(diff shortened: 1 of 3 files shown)


iwyu.cc
```diff
@@ -2334,6 +2334,149 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+          // default and copy construction before C++20. In C++20, apparently
+          // after P0960, it may return true for aggregate initialization even
+          // with multiple argument types. It is better to report the types even
+          // in the older language modes to avoid possible UB in user's code
+          // after switching to a newer standard.
+          // Unions cannot have multiple initializer expressions.
+          if (expr->getNumArgs() > 2 && lhs_type->isUnionType())
+            return true;
+          // Check arguments.
+          for (const TypeSourceInfo* arg : expr->getArgs()) {
```
kimgr, line 2349: Why does the presence of _one_ `void` type in the trait-expr automatically make all other types unused? I see the presence of `void` is undefined behavior, but if the compiler doesn't diagnose it, shouldn't we pretend like all the other types are still evaluated?

I'm a little confused by this whole traits handling in general, because it looks like you're almost evaluating the trait expression -- and I don't fully understand what the implication is on use of their arguments (except in the cases where the traits semantics do/don't require complete types). It makes all the early returns look a bit suspicious.

bolshakov-a: The standard and compilers usually require only that the types in the trait arguments are complete. It means that pointers and references to incomplete types are acceptable despite the evaluated trait value may become wrong. IWYU currently reports those types not by `VisitTypeTraitExpr` but by `VisitTagType`, `VisitTypedefType`, etc. (because `CanForwardDeclareType` returns `false` for them). Due to this, the early returns from `VisitTypeTraitExpr` don't affect reporting of types which should be reported according to the standard (and I wrote some test cases to check that). Some day IWYU should start to consider the type traits as a fwd-decl context. Then, some code should be added to report such mandatory types explicitly before the early returns.

`VisitTypeTraitExpr` currently just tries to report types which are not mandatory but the absence of which may cause the wrong evaluation result. I think that a `void` in the trait args causes defined behaviour because it is explicitly allowed in the standard. And because you cannot have a `void` expression in an initializer list, the trait is guaranteed to be `false` in such case, so no one additional type is needed.

I noted previously that the correct handling of the traits is not easy, but the majority of the work has been already done. Maybe, the logic is better to be moved from `iwyu.cc` to a separate file...

bolshakov-a: Maybe, your question is: "Could the trait erroneously evaluate to `true` due to an absence of some type not reported because of the early return?" I can't imagine such an example.

bolshakov-a: > and I wrote some test cases to check that

It turns out that there is no exactly that test case in this PR, but I've checked that such a line:
```cpp
bool b = __is_trivially_constructible(Class, void, Struct, void, Union1);
```
does trigger the reporting of `Class`, `Struct`, and `Union1`, as it must be. Should I add a test case like this?

kimgr: > IWYU currently reports those types not by VisitTypeTraitExpr but by VisitTagType,

Ah, that's the key thing I overlooked.

> Maybe, your question is: "Could the trait erroneously evaluate to true due to an absence of
> some type not reported because of the early return?"

No, I went to read up `std::is_constructible` and found this explanation: https://en.cppreference.com/w/cpp/types/is_constructible.html#:~:text=(possibly%20cv%2Dqualified)%20void%2C

And I assumed that by returning true there, we would miss all other argument types.

kimgr: > bool b = __is_trivially_constructible(Class, void, Struct, void, Union1);
> does trigger the reporting of Class, Struct, and Union1, as it must be. Should I add a test case like this?

I think these two cover it OK:
https://github.com/include-what-you-use/include-what-you-use/pull/1770/files#diff-82d056aca7eb54bbad882e32e4e03dd90fb1925019bb017f396ebf52f8dc8f64R1575-R1577

even if I guess that only covers void either in `T` or in `...Args`, but not `void` interleaved with other types in `Args...`. Anyway, I think the tests look good!

iwyu.cc
```diff
@@ -2334,6 +2334,149 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+          // Direct reference binding cases. See [dcl.init.ref].
+          // Only (maybe, potential) derived-to-base binding cases should be
+          // taken into account.
+          if (GetCanonicalType(lhs_deref_type.getTypePtr()) ==
+              GetCanonicalType(rhs_deref_type.getTypePtr())) {
+            return true;
+          }
+          const bool lvalue_bindable = lhs_type->isLValueReferenceType() &&
+                                       rhs_type->isLValueReferenceType();
+          const bool rvalue_bindable =
```
kimgr: I think you can drop these `const` for locals.

kimgr: Actually, maybe I'm internally inconsistent :-) -- I think `const T*` looks perfectly normal but `const T` for a value type looks excessive. So to follow that logic, maybe it makes sense to drop `const` for the `QualType` locals above too.

bolshakov-a: One guy from the company I worked for said similar things until he read Core Guidelines, and then he changed his mind and started to insist on `const`ness almost everywhere.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con1-by-default-make-objects-immutable

I don't suggest following it blindly, and I agree that in a small block scope it is rather a waste of space. OTOH, developers of some tools like clang-tidy are implementing Core Guideline checks, and it may be better to follow the rules as strictly as possible to facilitate automated checks...

kimgr: Yeah, I don't have strong feelings either way, it just looks a little noisy to me.

iwyu.cc
bolshakov-a: Btw, the wording on the Cppreference ("... is not a complete type, (possibly cv-qualified) void, or an array of unknown bound, the behavior is undefined.") may be misleading. The wording from the standard makes it clear that `void` and unknown bound arrays are acceptable.
http://eel.is/c++draft/meta.unary.prop#tab:meta.unary.prop

kimgr: Oh, I see the ambiguity now. Thanks for clarifying!

kimgr:
Thank you!

bolshakov-a:
Thanks for the review!

PR #1772, opened 2025-06-18
Compatible with clang-19.

jiapei100:


kimgr:
We don't backport to released versions, the `clang_19` branch only exists in case Clang should break us in a minor 19.x version.

Feel free to keep a backport branch for yourself (or share from your fork), but we can't take it in the project.

PR #1778, opened 2025-06-29
Exclude list works without source list specification too

AJIOB:


kimgr (commented):
Thanks! This makes a lot of sense.

Unfortunately we have an 80-column limit for Python, so it's hard to make this patch fit the max line length with the extra indentation. Rather than try to come up with a way to line-wrap, I'm going to try and rewrite the patch in a flatter form with the same semantics.

AJIOB:
Hi @kimgr ,

Thanks, I've tried to rewrite the script by myself.

One more suggestion (not for this PR) - add the linter, that will automatically check the code style & line length.

kimgr:
@AJIOB Nice, thanks!

Yeah, I should get around to adding a linter at some point, but I suspect it's going to have issues with some of the older IWYU Python code. And maybe some of the newer :slightly_smiling_face:

PR #1779, opened 2025-07-11
Set fwd-decl context in binary and multiarg traits

bolshakov-a:
This change removes full-type suggestions for `__is_same` trait, for which the argument types need not to be complete. In fact, it does not affect handling common implementations of `std::is_same`, because they look approximately like this:
```cpp
template <class T1, class T2>
struct is_same : std::integral_constant<bool, __is_same(T1, T2)> {};
```
where `__is_same` appears already in a fwd-decl context (inside template arguments). OTOH, this change fixes handling of e.g. `std::is_assignable` in such STL implementations because it forces reporting of required types.

Besides, this fixes unwanted reporting of the element types of arrays of unknown bound in some cases. Some code has been written to preserve the reportings for the first argument of "constructible" traits. It is not needed for the cases like
`__is_trivially_constructible(Struct[], Class&)` and `__is_trivially_constructible(Struct[], Struct&)`, because the second type is reported, and the element type should be available through it (be the same or a base class) for the trait to evaluate to true (if it didn't evaluate to 'false' unconditionally).

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2056,6 +2056,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (expr == nullptr || expr->getNumArgs() < 2)
       return true;
 
+    current_ast_node()->set_in_forward_declare_context(true);
+
+    auto require_complete_arg_types = [expr, this] {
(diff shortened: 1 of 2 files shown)


kimgr:
Beautiful, thanks!

PR #1781, opened 2025-07-12
Handle "is base of" type traits

bolshakov-a:


bolshakov-a:
Maybe, some test cases with cv-qualified types are needed?

kimgr:
@bolshakov-a The tests are already very comprehensive, so I'd prefer to leave them out (I guess the relevant canonical-type comparisons are already covered by alias tests). If something shows up in practice, we can always add more special test cases later.

PR #1784, opened 2025-07-15
Handle unary type traits

bolshakov-a:
The only remaining trait is the C language binary type trait `__builtin_types_compatible_p`.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2053,7 +2053,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreCurrentASTNode())
       return true;
 
-    if (expr == nullptr || expr->getNumArgs() < 2)
+    if (expr == nullptr)
       return true;
 
     current_ast_node()->set_in_forward_declare_context(true);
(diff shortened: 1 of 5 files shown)


iwyu.cc
```diff
@@ -2083,7 +2083,120 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      case TypeTrait::UTT_StructuredBindingSize:
+        // This trait compiles only for decomposable types, which simplifies
+        // analysis. Constant-size arrays are always decomposable, so
+        // the complete element type is not needed. For class types, it is
+        // probably better to require the complete type despite its
+        // fwd-declaration is sufficient to define the std::tuple_size
+        // specialization in the case of tuple-like classes.
+        if (!lhs_type->isArrayType())
+          ReportTypeUse(CurrentLoc(), lhs_type, DerefKind::None);
+        return true;
```
bolshakov-a, line 2132: In principle, this may be united with the group started with `UTT_IsFinal` (in fact, as `UTT_IsAggregate` and `UTT_IsImplicitLifetime` have been united with `UTT_IsFinal` and `UTT_IsSealed`.

iwyu.cc
```diff
@@ -2083,7 +2083,120 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      case TypeTrait::UTT_IsNothrowDestructible:
+      case TypeTrait::UTT_IsTriviallyDestructible:
+        // For these traits, unbounded array element types are not needed.
+        if (!lhs_type->isIncompleteArrayType())
+          ReportTypeUse(CurrentLoc(), lhs_type, DerefKind::None);
+        return true;
+      case TypeTrait::UTT_IsIntangibleType:
+      case TypeTrait::UTT_IsTypedResourceElementCompatible:
+      case TypeTrait::BTT_IsScalarizedLayoutCompatible:
+        CHECK_UNREACHABLE_("HLSL is not supported.");
```
bolshakov-a: Just to be sure: does IWYU really not support it?

kimgr: I don't know what HLSL looks like, or how it ties into the physical/logical relationships that IWYU cares about. But it seems a bit harsh to abort. Why not just add a TODO and return true?

iwyu.cc
kimgr: I guess they're ordered and grouped like this because they follow the Clang Sema order? That seems good to me.

OTOH, if you ordered them by outcomes instead (like you're suggesting), it might be easier to add handlers for new traits.

I don't have a strong preference, do what you feel is best.

iwyu.cc
```diff
@@ -2083,7 +2083,120 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // better for IWYU to require the full type info in such situations. It is
     // important not to suggest a full type when it is decidedly unnecessary
     // for computing the correct value.
+    // See CheckUnaryTypeTraitTypeCompleteness function from
+    // clang/lib/Sema/SemaTypeTraits.cpp for the reference on unary type traits.
     switch (expr->getTrait()) {
+      case TypeTrait::UTT_IsCompleteType:
+        // TODO(bolshakov): would it be better to require the type here because
+        // static_assert(__is_complete_type(Type)) is more probable to appear
```
kimgr: On the other hand, doesn't the `__is_complete_type` check signal that you want conditional behavior depending on whether the full type is available or not? Having IWYU make the full type be available (by adding an `#include`) would make the check redundant in that case...?

iwyu.cc
```diff
@@ -2053,7 +2053,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreCurrentASTNode())
       return true;
 
-    if (expr == nullptr || expr->getNumArgs() < 2)
```
kimgr, line 2056: This feels a little scary, as `expr->getNumArgs() < 2` implies that `rhs_type` will be null throughout the whole trait switch, for both unary and binary traits.

I wonder if it would make sense to split it into two switch statements, one for unary and one for binary. That way you could do lhs/rhs/count validation between them.

tests/cxx/type_trait.cc
```diff
@@ -7,7 +7,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-// IWYU_ARGS: -I . -Wno-deprecated-builtins -Wno-c++20-extensions
+// IWYU_ARGS: -I . -Wno-deprecated-builtins -Wno-c++20-extensions \
+//            -fms-extensions
```
kimgr, line 11: I'm not sure what the status is now, but `-fms-extensions` used to change language semantics in subtle ways (e.g. the delayed template expansion). So I wonder if the tests would be stronger if we had the native Clang traits tests in `type_trait.cc`, and then add a separate `type_trait_ms.cc` for any constructs needing MSVC extensions?

kimgr: If that sounds like a good idea, but causes a lot of churn, it might be a good idea to split this PR to skip over the MS-specific traits, and then add those in a separate PR with the separate test file.

iwyu.cc
bolshakov-a: `rhs_type` already was null on `TT_...` traits with more than two arguments. For the `BTT_...` traits, clang checks that the argument number is exactly 2 before the IWYU analysis.

iwyu.cc
bolshakov-a: I still cannot imagine such use cases. I think, they would introduce some sort of UB. But I'd prefer don't require the complete type here for now, too. Do you think the comment should be removed?

iwyu.cc
kimgr: I see. I overlooked the n-ary `TT_` traits. OK, I guess this is good.

iwyu.cc
kimgr: Your imagination is bigger than mine, so I'm sure you're right :-)

Yeah, I think we can remove the TODO, it's not 100% clear that there's anything to do.

iwyu.cc
bolshakov-a: I've checked that IWYU can be launched on an HLSL code. (You may need to add `-target dxil-pc-shadermodel6.3-library` command line flag, as well as `-x hlsl` if you type the input in the console). I've somehow concerned that if someone actually uses IWYU for HLSL, the change may introduce a regression. OTOH, support for HLSL in IWYU is not declared anywhere. There is even no `hlsl` folder in the tests.

I also have (almost) zero experience with shader programming despite over 6 years of working for a computer graphics company. I talked to some people and I was told that IWYU is hardly useful for HLSL because there is no translation units and linkage between them there.

tests/cxx/type_trait.cc
bolshakov-a: Looks like `-fms-extensions` doesn't change the way templates are parsed. You need `-fms-compatibility` or `-fdelayed-template-parsing` to compile this: https://godbolt.org/z/vhKdqvhhf

iwyu.cc
bolshakov-a: Yeah, they are mostly ordered as in `CheckUnaryTypeTraitTypeCompleteness` from `SemaTypeTraits.cpp`, except `UTT_HasNothrowAssign`, `UTT_HasNothrowMoveAssign`, `UTT_HasTrivialAssign`, and `UTT_HasTrivialMoveAssign`, which are separated into a distinct group due to MSVC behavior. Moreover, `UTT_StructuredBindingSize` is somewhat special trait, so let's leave the order as-is.

tests/cxx/type_trait.cc
bolshakov-a: A link to the clang docs on that: https://clang.llvm.org/docs/UsersManual.html#microsoft-extensions

kimgr: Ah, thanks. I had it confused with `-fms-compatibility`.

kimgr:
Thank you!

bolshakov-a:
Thank you!

PR #1786, opened 2025-07-18
Scan variable template implicit instantiations

bolshakov-a:
This is especially useful in conjunction with improved type trait handling because many people write now e.g. `std::is_base_of_v<T1, T2>` instead of `std::is_base_of<T1, T2>::value`.

`current_ast_node_` has been made protected so that `CurrentASTNodeUpdater` can be used inside `InstantiatedTemplateVisitor`. It is needed for putting `VarDecl` node into the stack correctly so that `CanForwardDeclareType` evaluates to true on the types of variable template specializations marked as `extern`. In fact, I'm not very sure that the handling of such variable templates is really needed: clang warns on them (hence the `-Wno-undefined-var-template` flag in the test) and suggests to use explicit instantiation declarations.

The change in `InstantiatedTemplateVisitor::VisitRecordType` has been made because clang inserts `ElaboratedType` node for some reason between `SubstTemplateTypeParmTypeLoc` and `RecordType` nodes in the case of the pointer variable definition inside the lambda initializer of `ptr_use_in_partial_spec`.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -274,6 +274,7 @@ using clang::UsingShadowDecl;
 using clang::UsingType;
 using clang::ValueDecl;
 using clang::VarDecl;
+using clang::VarTemplateSpecializationDecl;
 using clang::driver::ToolChain;
 using llvm::cast;
 using llvm::drop_begin;
@@ -347,8 +348,8 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
   // We need to create implicit ctor/dtor nodes, which requires
   // non-const methods on CompilerInstance, so the var can't be const.
   explicit BaseAstVisitor(CompilerInstance* compiler)
-      : compiler_(compiler),
-        current_ast_node_(nullptr) {}
+      : current_ast_node_(nullptr), compiler_(compiler) {
+  }
 
   virtual ~BaseAstVisitor() = default;
 
@@ -867,15 +868,16 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
     return compiler_;
   }
 
- private:
-  template <typename T> friend class BaseAstVisitor;
-  CompilerInstance* const compiler_;
-
   // The currently active decl/stmt/type/etc -- that is, the node
   // being currently visited in a Visit*() or Traverse*() method.  The
   // advantage of ASTNode over the object passed in to Visit*() and
   // Traverse*() is ASTNode knows its parent.
   ASTNode* current_ast_node_;
+
+ private:
+  template <typename T>
+  friend class BaseAstVisitor;
+  CompilerInstance* const compiler_;
 };
 
 // ----------------------------------------------------------------------
@@ -3452,6 +3454,8 @@ class InstantiatedTemplateVisitor
   // tpl_type_args_of_interest, which are the types we care about, and
   // usually explicitly written at the call site.
   //
+  // ScanInstantiatedVariable is similar, but it is used for variable templates.
+  //
   // ScanInstantiatedType() is similar, except that it looks through
   // the definition of a class template instead of a statement.
   //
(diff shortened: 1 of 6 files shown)


tests/cxx/variable_template.cc
```diff
@@ -0,0 +1,193 @@
...
+
+template <typename>
+// IWYU: GetInt is...*-indirect.h
+int refers_to_fn_in_init = GetInt();
+
+template <typename T>
+T typed_as_param_expl_spec;
+
+template <>
+// IWYU: IndirectClass is...*/indirect.h
```
bolshakov-a: `*/` is not very good combination of symbols in C++ comments, but it is needed to distinguish between `indirect.h` and `variable_template-indirect.h`.

iwyu.cc
```diff
@@ -3842,10 +3865,8 @@ class InstantiatedTemplateVisitor
     // CanForwardDeclareType function relies on the specific placement of
     // the type node in the AST. An intermediate SubstTemplateTypeParmType could
     // break that logic. However, such cases don't even need to be considered
-    // here, because they are handled in VisitSubstTemplateTypeParmType. But
-    // maybe using HasAncestorOfType, or replacing VisitSubst...Type with
-    // TraverseSubst...Type and Traverse...TypeLoc would be more reliable.
-    if (!current_ast_node()->ParentIsA<SubstTemplateTypeParmType>())
+    // here, because they are handled in VisitSubstTemplateTypeParmType.
+    if (!current_ast_node()->HasAncestorOfType<SubstTemplateTypeParmType>())
```
bolshakov-a: An alternative could be `MostElaboratedAncestor`.

tests/cxx/variable_template.cc
kimgr: You could work around it by naming the new file `variable_template-i1.h`, even if there's just the one.

iwyu.cc
```diff
@@ -4099,6 +4132,31 @@ class InstantiatedTemplateVisitor
...
 
+  bool TraverseExpandedTemplateVariableHelper(VarDecl* decl) {
+    if (!decl || ContainsKey(traversed_decls_, decl))
+      return true;  // avoid recursion and repetition
+    // TODO(bolshakov): is this really needed? BaseAstVisitor::Traverse* methods
+    // already have protection from recursion.
+    traversed_decls_.insert(decl);
+
+    AstFlattenerVisitor nodeset_getter(compiler());
+    // This gets to the decl for the (uninstantiated) template-as-written:
```
kimgr, line 4144: This is significantly simpler than the recursive push/pop thing going on in TraverseExpandedTemplateFunctionHelper: https://github.com/include-what-you-use/include-what-you-use/pull/1786/files#diff-7ca107c8b836c1765e8b735b74f06833bec48f308a76017886cee1d5736002d9L4076.

I have a very shallow understanding of variable templates, but are they just less nested than function templates by definition? Or why is the `ValueSaver` not necessary here?

iwyu.cc
kimgr: Yeah, let's do that if we know it's only elaborated types getting in the way.

iwyu.cc
bolshakov-a: I've found in the history that `ValueSaver` was added there by me for a very specific task: so as not to block the second analysis of a function from `HandleFunctionCall` after the first one has been made from `VisitDeclRefExpr`. It is important for reporting the function return type from `HandleFunctionCall`. This is irrelevant to variable templates. But I can add `ValueSaver` here as well to be consistent and maybe invent some specific test case, despite I doubt that it would be useful in real life...

bolshakov-a: In fact, I have an idea that `AstFlattenerVisitor` should be replaced some day with some more advanced approach to block reporting of entities used at the template definition side... It would be needed to report types not only from template arguments.

kimgr: Thanks for explaining. I think it's good as-is without `ValueSaver` if it's just there for a specific function-call quirk.

kimgr:
Thanks!

bolshakov-a:
Thank you!

PR #1792, opened 2025-08-03
Add `--no_fwd_decls_source_only` option to disable forward declarations in source files

JoHuang:
Ref: [this thread](https://groups.google.com/g/include-what-you-use/c/OmqYn2MELRg/m/QKIDBzC_AwAJ?utm_medium=email&utm_source=footer)

- Added `--no_fwd_decls_source_only` command-line option to suppress forward declarations in `.cpp` files only.
- Updated `PrintHelp` to document the new option.
- Modified `CommandlineFlags` to include the new flag.
- Enhanced `ProcessForwardDeclare` to respect this setting.
- Added tests to verify `no_fwd_decls_source_only` behavior.

### Background

Our project uses complex macros, which sometimes obscure symbol usage from IWYU. As a result, some required symbols are incorrectly treated as candidates for forward declarations. However, in `.cpp` files, we typically include headers directly and do not benefit from forward declarations. This option allows us to avoid unnecessary or incorrect forward declarations in source files, while still retaining them in headers.

### DIFF
--- a/iwyu_globals.cc
+++ b/iwyu_globals.cc
@@ -117,6 +117,9 @@ static void PrintHelp(const char* extra_msg) {
          "   --update_comments: update and insert 'why' comments, even if no\n"
          "        #include lines need to be added or removed.\n"
          "   --no_fwd_decls: do not use forward declarations.\n"
+         "   --no_fwd_decls_source_only: do not use forward declarations\n"
+         "        in source files only (header files still allow forward\n"
+         "        declarations).\n"
          "   --verbose=<level>: the higher the level, the more output.\n"
          "   --quoted_includes_first: when sorting includes, place quoted\n"
          "        ones first.\n"
@@ -217,6 +220,7 @@ CommandlineFlags::CommandlineFlags()
       update_comments(false),
       comments_with_namespace(false),
       no_fwd_decls(false),
+      no_fwd_decls_source_only(false),
       quoted_includes_first(false),
       cxx17ns(false),
       exit_code_error(EXIT_SUCCESS),
@@ -242,6 +246,7 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
     {"no_comments", no_argument, nullptr, 'o'},
     {"update_comments", no_argument, nullptr, 'u'},
     {"no_fwd_decls", no_argument, nullptr, 'f'},
+    {"no_fwd_decls_source_only", no_argument, nullptr, 'g'},
     {"quoted_includes_first", no_argument, nullptr, 'q' },
     {"cxx17ns", no_argument, nullptr, 'C'},
     {"error", optional_argument, nullptr, 'e'},
@@ -278,6 +283,7 @@ int CommandlineFlags::ParseArgv(int argc, char** argv) {
         }
         break;
       case 'f': no_fwd_decls = true; break;
+      case 'g': no_fwd_decls_source_only = true; break;
       case 'x':
         if (strcmp(optarg, "add") == 0) {
           prefix_header_include_policy = CommandlineFlags::kAdd;
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1206,7 +1206,10 @@ void ProcessForwardDeclare(OneUse* use,
   // the headers, suggest that header, and recategorize as a full use. If we can
   // only find a decl in this file, it must be a self-sufficent decl being used,
   // so we can just let IWYU do its work, and there is no need to recategorize.
-  if (!use->ignore_use() && GlobalFlags().no_fwd_decls) {
+  // Note: --no_fwd_decls_source_only only applies to source files.
+  if (!use->ignore_use() && 
+      (GlobalFlags().no_fwd_decls || 
+       (GlobalFlags().no_fwd_decls_source_only && !IsHeaderFile(GetFilePath(use->use_loc()))))) {
     bool promote_to_full_use = true;
     for (const Decl* decl = use->decl(); decl != nullptr;
          decl = decl->getPreviousDecl()) {
--- a/iwyu_globals.h
+++ b/iwyu_globals.h
@@ -105,6 +105,7 @@ struct CommandlineFlags {
   bool update_comments; // Force 'why' comments. No short option.
   bool comments_with_namespace; // Show namespace in 'why' comments.
   bool no_fwd_decls;  // Disable forward declarations.
+  bool no_fwd_decls_source_only;  // Disable forward declarations in source files only.
   bool quoted_includes_first; // Place quoted includes first in sort order.
   bool cxx17ns; // -C: C++17 nested namespace syntax
   int exit_code_error;   // Exit with this code for iwyu violations.
--- a/tests/cxx/no_fwd_decls_source_only-d1.h
+++ b/tests/cxx/no_fwd_decls_source_only-d1.h
@@ -1,0 +1,19 @@
+//===--- no_fwd_decls_source_only-d1.h - test input file for iwyu -------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_NO_FWD_DECLS_SOURCE_ONLY_D1_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_NO_FWD_DECLS_SOURCE_ONLY_D1_H_
+
+
+template <typename T>
+class FinalTemplate final {};
+
+class FinalClass final {};
+
+#endif  // INCLUDE_WHAT_YOU_USE_TESTS_CXX_NO_FWD_DECLS_SOURCE_ONLY_D1_H_
--- a/tests/cxx/no_fwd_decls_source_only.cc
+++ b/tests/cxx/no_fwd_decls_source_only.cc
@@ -1,0 +1,31 @@
+//===--- no_fwd_decls_source_only.cc - test input file for iwyu ---------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --no_fwd_decls_source_only -I .
+
+#include "tests/cxx/no_fwd_decls_source_only.h"
+
+void FwdDeclFinal::testFinalTemplate(FinalTemplate<int>* finalTemplate) {
+}
+
+void FwdDeclFinal::testFinalClass(FinalClass* finalClass) {
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/no_fwd_decls_source_only.cc should add these lines:
+#include "tests/cxx/no_fwd_decls_source_only-d1.h"
+
+tests/cxx/no_fwd_decls_source_only.cc should remove these lines:
+
+The full include-list for tests/cxx/no_fwd_decls_source_only.cc:
+#include "tests/cxx/no_fwd_decls_source_only.h"
+#include "tests/cxx/no_fwd_decls_source_only-d1.h"  // for FinalClass, FinalTemplate
+
+***** IWYU_SUMMARY */
--- a/tests/cxx/no_fwd_decls_source_only.h
+++ b/tests/cxx/no_fwd_decls_source_only.h
@@ -1,0 +1,36 @@
+//===--- no_fwd_decls_source_only.h - test input file for iwyu ----------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_NO_FWD_DECLS_SOURCE_ONLY_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_NO_FWD_DECLS_SOURCE_ONLY_H_
+
+#include "tests/cxx/no_fwd_decls_source_only-d1.h"
+
+class FwdDeclFinal {
+public:
+  void testFinalTemplate(FinalTemplate<int>* finalTemplate);
+  void testFinalClass(FinalClass* finalClass);
+};
+
+#endif  // INCLUDE_WHAT_YOU_USE_TESTS_CXX_NO_FWD_DECLS_SOURCE_ONLY_H_
+
+/**** IWYU_SUMMARY
+
+tests/cxx/no_fwd_decls_source_only.h should add these lines:
+class FinalClass;
+template <typename T> class FinalTemplate;
+
+tests/cxx/no_fwd_decls_source_only.h should remove these lines:
+- #include "tests/cxx/no_fwd_decls_source_only-d1.h"  // lines XX-XX
+
+The full include-list for tests/cxx/no_fwd_decls_source_only.h:
+class FinalClass;
+template <typename T> class FinalTemplate;
+
+***** IWYU_SUMMARY */


bolshakov-a:
Thanks for working on this! However, from my point of view, fwd-declarable uses in source files should just be ignored, without recategorizing to full uses. For example:
```cpp
// funcs.h
class Class;

Class* GetPtr();
void TakePtr(Class*);

// source.cpp
#include "funcs.h"

void TransmitPtr() {
    Class* p = GetPtr();
    TakePtr(p);
}
```
IWYU currently suggests to add `class Class;` declaration to `source.cpp`, but replacing it with `#include` is hardly better. Note that a fwd-declaration should already be provided to `source.cpp` by `funcs.h`. Considering your test case, do you really want to include the full definitions of `FinalClass` and `FinalTemplate` in your source file despite not using them? Maybe, you have just faced with some bug when IWYU doesn't report complete type definitions where they are needed in your code?

JoHuang:
Indeed, we ran into a situation where IWYU didn’t expand a macro as expected — possibly a bug, though I haven’t been able to reproduce it in a minimal test case so far. At the moment, this option seems to serve as a workaround for that edge case.

Also, in our project, it’s very rare for a source file to use only pointers or references without needing the full definition. So in most cases, we actually prefer to include the full C++ header directly.

That said, if this behavior doesn’t seem appropriate to merge into the main branch, I totally understand.

bolshakov-a:
> possibly a bug, though I haven’t been able to reproduce it in a minimal test case so far

That's a pity. Note that if you e.g. used the "almost always auto" approach, you might not have type names written in your source code files, so IWYU would not report even fwd-decl uses, and recategorizing them as full uses would not help you.

I just wanted to add an option to suppress fwd-declarations in source files some day with the meaning described by me, and two similar options may be confusing to users...

> That said, if this behavior doesn’t seem appropriate to merge into the main branch, I totally understand.

I don't make decisions here, @kimgr does it.

JoHuang:
Just to clarify — the option you had in mind to suppress forward declarations in source files, would it behave the same way as the one I implemented here? Or did you intend something different?

bolshakov-a:
It would cause ignoring fwd-declarable usages so that for `source.cpp` in my example IWYU would suggest neither a fwd-declaration of `Class` nor `#include` of a file containing its full definition. Maybe IWYU should scan included files to check that some of them actually provides a declaration, maybe it is redundant in real-world code. Probably, it should be even not an option but the default behavior.

JoHuang:
Is this CI failure related to the Clang version (looks like Clang 22)?
Just wanted to check if this is expected or if I should make changes to adapt to it. Thanks!

bolshakov-a:
The same failure is present on the master branch. Fixed here: #1793.

kimgr:
> IWYU currently suggests to add class Class; declaration to source.cpp, but replacing it with #include is hardly better. Note that a 
> fwd-declaration should already be provided to source.cpp by funcs.h.

It seems pretty un-IWYU to remove explicit dependencies in favor of just hoping the symbol makes it through from some included header.

The revised Google style guide recommends against forward declarations entirely:

* https://google.github.io/styleguide/cppguide.html#Include_What_You_Use
* https://google.github.io/styleguide/cppguide.html#Forward_Declarations

And I think this is the behavior we're aiming for with `--no_fwd_decls`.

(Side note: to me, their policy looks a little too pessimistic. Build times are usually the main issue with growing C++ programs, and forward declarations are a well-established technique that make a huge difference. Maybe Google has better build tools/caching that tip the scales, but I don't agree with the pro/con tradeoff for C++ in general.)

iwyu_globals.cc
```diff
@@ -117,6 +117,9 @@ static void PrintHelp(const char* extra_msg) {
          "   --update_comments: update and insert 'why' comments, even if no\n"
          "        #include lines need to be added or removed.\n"
          "   --no_fwd_decls: do not use forward declarations.\n"
+         "   --no_fwd_decls_source_only: do not use forward declarations\n"
```
kimgr: I think I would prefer to make `--no_fwd_decls` take an optional argument, and:

* Add an `enum class NoFwdDecls` with enumerators `Disabled`, `Everywhere` and `SourcesOnly`
* Let it default to `Disabled`
* If  `--no_fwd_decls=everywhere` or `--no_fwd_decls` is provided without argument, set it to `Everywhere`
* If `--no_fwd_decls=src-only` is provided, set it to `SourcesOnly`
* If any other argument, fail with an error

It even feels like we could generalize `--no_fwd_decls` to some kind of `--policy_tweak`, but I'm not sure how well that scales. Small steps.

I'm not loving my name suggestions, but something to that effect would be nicer than a whole new switch.

iwyu_output.cc
```diff
@@ -1206,7 +1206,10 @@ void ProcessForwardDeclare(OneUse* use,
   // the headers, suggest that header, and recategorize as a full use. If we can
   // only find a decl in this file, it must be a self-sufficent decl being used,
   // so we can just let IWYU do its work, and there is no need to recategorize.
-  if (!use->ignore_use() && GlobalFlags().no_fwd_decls) {
+  // Note: --no_fwd_decls_source_only only applies to source files.
+  if (!use->ignore_use() && 
+      (GlobalFlags().no_fwd_decls || 
+       (GlobalFlags().no_fwd_decls_source_only && !IsHeaderFile(GetFilePath(use->use_loc()))))) {
```
kimgr: I'd prefer not looking at the path name here, but at the file entry. We didn't have great support for that, so I added some improvement in #1796.

bolshakov-a:
> It seems pretty un-IWYU to remove explicit dependencies in favor of just hoping the symbol makes it through from some included header.

Do you know a real-world case when a fwd-declaration in a .cpp-file is useful? Here was a complaint about them: https://github.com/include-what-you-use/include-what-you-use/issues/1115#issuecomment-1314727796

CC @justusranvier

kimgr:
@bolshakov-a Ah, I see what you're saying now -- the payoff for forward-decls is primarily beneficial in shared headers, because they are included over and over, and so forward-decls help make them smaller and faster to parse.

As far as "useful" in the sense "makes things faster to compile", no. But I think expressing the shallow dependency with a forward-declare is a boon for readability.

justusranvier:
> Do you know a real-world case when a fwd-declaration in a .cpp-file is useful? Here was a complaint about them: [#1115 (comment)](https://github.com/include-what-you-use/include-what-you-use/issues/1115#issuecomment-1314727796)

After posting that comment a couple years ago I learned what was causing the undesired forward declarations in our case.

    Factory.hpp
    Class.hpp
    Class.cpp: #include "Class.hpp" // IWYU pragma: associated

`Class.cpp` has the implementation for the class defined in `Class.hpp` and it also has the implementation of a factory function which returns `Class` but that function was declared in `Factory.hpp`.

Because `Factory.hpp` wasn't marked with `// IWYU pragma: associated` when it was included in in `Class.cpp` this would cause IWYU to add unnecessary and undesired forward declarations related to the implemented factory function in `Class.cpp` that were already forward declared in `Factory.hpp`

Ultimately rather than constantly fighting IWYU by adding pragmas to suppress the undesired behavior we adapted our project structure to avoid the problem based on inspiration from the [Pitchfork Layout](https://joholl.github.io/pitchfork-website/):

    Factory.hpp
    Class.hpp
    Class.cpp: #include "Class.hpp" // IWYU pragma: associated
    Class.factory.cpp : #include "Factory.hpp" // IWYU pragma: associated
    
After making that change this particular issue is no longer a problem for us however I still think it's wrong for IWYU to add forward declarations in cpp files. Forward declarations only belong in header files.

iwyu_globals.cc
JoHuang: I've made the changes you suggested - replaced the separate boolean flags with a `NoFwdDecls` enum and made `--no_fwd_decls` accept optional arguments (`everywhere` or `src-only`). The old usage still works exactly the same way. Thanks!

iwyu_output.cc
JoHuang: Should I merge #1796 into my branch to use the file entry improvements, or wait for it to be merged to master first?

kimgr: I think it would be nice to get master green before building up any more work-in-progress. There's been a huge breaking change in Clang, so that might be a while... But once that's sorted, I'll merge #1796 asap, and you can rebase on master.

kimgr:
My primary concern is the case where a type `T` is incidentally provided by a large unrelated header `foobar.h`. And removing `#include "foobar.h"` is inhibited because it provides `T`. Or in the world you seem to be suggesting, compilation breaks because the forward-declaration of `T` disappears when `#include "foobar.h"` is removed.

But yeah, I guess it does look like there are some code bases where these forward-decls are always redundant, because they're already defined in a shared header by convention. Question is if it's better to export the forward-decls automatically or to just pretend name-only uses are magically covered. Maybe those are two different `--no_fwd_decls` policies?

justusranvier:
The definitions in a source file are either defined elsewhere in an associated header file or they aren't.

In the former case then all necessary forward declarations belong in the associated header file (which IWYU may or may not be capable of identifying without explicit annotations).

In the latter case I'm struggling to imagine a situation where a forward declaration is both necessary and sufficient: in every case I can think of you should need an include, not a forward declaration.

The problem I originally commented on was where a single source file implemented symbols defined in multiple header files and we fixed that by splitting it into multiple source files where each source file has exactly one associated header. We could have kept everything in a single source file and marked more than one header with `IWYU pragma: associated` but that seemed more likely to cause other problems.

kimgr:
@justusranvier I'm not sure if these are universal truths or just something that happens to be true in your code bases?

As for real-world examples, here's one: https://github.com/include-what-you-use/include-what-you-use/blob/82e6005f7387a76831324438458eda12aea45191/iwyu.cc#L167

* iwyu.cc only uses `ToolChain` by name, to funnel it through a bunch of factory wiring
* iwyu_driver.h only uses `ToolChain` by name too, so there's no strict provider of a forward-decl for the type

In this case it seems to me a fwd-decl is the right mechanism to avoid including a bunch of unnecessary headers. Sure, there may already be a declaration of `ToolChain` available in the transitive header closure of iwyu.cc or iwyu_driver.cc, but that's just a coincidence.

bolshakov-a:
@kimgr, `iwyu.cc` uses `ToolChain` only to get it from `ExecuteAction` internals and pass it to `InitGlobals` function, hence both of `iwyu_driver.h` and `iwyu_globals.h` headers should provide at least the fwd-declaration, and this is not a coincidence. (Ok, `InitGlobals` might take `void*` instead, but `iwyu.cc` might also use just `void*` in this case.) So, the `ToolChain` fwd-declaration inside `iwyu.cc` is obviously redundant.

One case that came into my mind is with explicit instantiation definitions of templates when the template doesn't require the complete argument type, like this:
```cpp
// Class.h
class Class {};

// tpl.h
template <typename T>
T* GetAddress(T& t);

// tpl.cpp
#include "tpl.h"

template <typename T>
T* GetAddress(T& t) {
    T* p = &t;
    return p;
}

class Class;

template Class* GetAddress(Class&);

// main.cpp
#include "tpl.h"
#include "Class.h"

int main() {
    Class c;
    Class *p = GetAddress(c);
}
```
This is exactly the case when the `Class` fwd-declaration inside `tpl.cpp` is both necessary and sufficient. This technique may be useful if you want e.g. to hide the template implementation in a .cpp-file. Also, such cases may be occasionally encountered when working with Boost.Serialization library, IIRC. However, they are very rare.

kimgr:
@bolshakov-a Thanks, I completely missed the forward-decls in both headers. Took some time to set up a bigger screen, and now I see them :-).

So, just so I understand:

* @JoHuang's patch will currently promote all forward-decl uses in non-headers to full uses
* You're suggesting to instead ignore forward-decl uses in non-headers, on the assumption that they must be provided by some header we're already including

?

justusranvier:
> ignore forward-decl uses in non-headers, on the assumption that they must be provided by some header we're already including

If all the headers in a project are fully IWYU-clean then this is a valid assumption.

It took us a while to realize that IWYU doesn't work perfectly on a header unless there is at least one source file which marks that header as "associated". To make sure this is the case we have many source files in our projects which have no content at all except for a single `#include ... // IWYU pragma: associated` just to make sure that IWYU processes that header and to make sure the header can compile successfully in isolation.

bolshakov-a:
@kimgr, yeah, I'm suggesting to add an opt-in or opt-out option to ignore them. I remember that there was some irritation (including from me, but not only) in the company I worked for when someone commited fwd-decls suggested by IWYU in source files. So I think it would be good to have such an option. But yeah, it may be just another `--no_fwd_decls` policy.

kimgr:
Feels like the policy should be reversed, e.g. `--fwd_decls[=policy]`

* `none` => no forward decls ever
* `headers` => only in headers
* `all` => wherever the language allows

I think that might easier to think about than `--no_fwd_decls`...?

bolshakov-a:
@justusranvier, ["Include Toolbox" VS plugin](https://github.com/Wumpf/IncludeToolbox) makes a temporary source file when the user launches IWYU through it on a header. You might do something similar with your build system... Btw, there is one strange thing in your first message:
> `Class.cpp: #include "Class.hpp" // IWYU pragma: associated`

You shouldn't be required to place the pragma in such case because `Class.hpp` is an associated header for `Class.cpp` by default.

@kimgr, looks like there should be four of them, not three: the two existing ones and the two new ones. (However, I'm still in doubt if the policy suggested in this PR is very useful...)

justusranvier:
> You shouldn't be required to place the pragma in such case because `Class.hpp` is an associated header for `Class.cpp` by default.

In many cases IWYU does detect the relationship correctly, for example when both files are in the same directory.

There are enough false positives and false negatives though that we just always explicitly mark it as a matter of policy.

iwyu_output.cc
JoHuang: Got it.

bolshakov-a:
> In many cases IWYU does detect the relationship correctly, for example when both files are in the same directory.

It is a very common practice when headers and their corresponding source files are in different folders, so IWYU should support it. I've experimented for a while and noted that IWYU does support it... to some extent. Namely, paths to headers included from headers should be relative to any of the include search paths explicitly specified in the command line. Anyway, it would be good to open issues related with associated header detection, at least to track them.

> ["Include Toolbox" VS plugin](https://github.com/Wumpf/IncludeToolbox) makes a temporary source file when the user launches IWYU through it on a header.

In fact, temporary source files just including headers are not needed because IWYU can be launched on a header file directly.

bolshakov-a:
> I completely missed the forward-decls in both headers.

For the headers to be compilable by itself, a declaration (either a forward decl or the full definition) should anyway be present in the header or be included in it directly or indirectly.

kimgr:
@JoHuang Master is now compatible with upstream Clang again, and I merged the IsInHeader thing too. Could you rebase to trigger another build?

JoHuang:
@kimgr 
I’ve rebased and fixed `IsInHeader`. I’m not fully sure I understood everything correctly, so I’d really appreciate your review

kimgr:
@JoHuang I'll circle back to this in a week or so, I'm in the middle of a busy few weeks, but haven't forgotten.

kimgr:
The suggestion to avoid forward-decls entirely in source files still makes me a little uncomfortable, so I'd like to split the discussion to see if we can lay a foundation for future work in that direction separately:

* Add a `--fwd_decls=<policy>` switch to get rid of all the confusing double negations. In my mind, the policies would be something like `always`, `never`
* Make `--no_fwd_decls` an alias for `--fwd_decls=never`
* Once that's in place, we can add a new policy `headers`, and basically implement what's in this patch

So if we want to follow the "ignore-in-source-files-instead-of-making-them-full-uses" train of thought suggested by @justusranvier and @bolshakov-a, we should be able to do that as a fourth policy `ignore-in-source`.

I'm not very happy with the naming I suggested above, so I think we should brainstorm a little on the command-line design. Once we've picked something, we're stuck with it basically forever, so let's make it good.

bolshakov-a:
`on`, `off`, `ignore_in_source`, `transform_to_include_in_source`?

bolshakov-a:
This may be more consistent: `allowed`, `transform_to_includes`, `transform_to_includes_in_source_files`, `suppress_in_source_files`.

kimgr:
Maybe two switches, so they can cover all permutations? 

* on, off, suppress
* headers, sources, all

Naming is hard.

JoHuang:
To me:
`always`
`never`
`headers-only` -> my patch
`use-if-needed` -> avoid redundant fwd-decls (https://github.com/include-what-you-use/include-what-you-use/pull/1792#issuecomment-3148627950) without affecting necessary fwd-decls (https://github.com/include-what-you-use/include-what-you-use/pull/1792#issuecomment-3172863564)

kimgr:
@JoHuang So would you say `use-if-needed` implies `headers-only`?

I've been trying to come up with more examples, but keep confusing myself... These are the choices, right:

* use forward-decls wherever the language allows (current default)
* force full includes always (current `--no_fwd_decl`)
* only use forward-decls in headers, force full includes in source files
* only use forward-decls in headers, **ignore** forward-declare uses in source files where possible

? I find it unusually hard to come up with words that describe a progression of more or less forward-declary-ness... It's like the last bullet is a bit to the side.

bolshakov-a:
I still suppose the last option should be the most frequently used one if implemented. Consider your example with `clang::driver::ToolChain`: if `iwyu_driver.h` and `iwyu_globals.h` stop naming it, `iwyu.cc` should be changed anyway, be the fwd-decl present there or not. Even in the case if it is replaced with some base/derived class, the fwd-decl should be replaced with `#include` to make derived-to-base conversion possible.

JoHuang:
@kimgr 

use-if-needed is broader than headers-only.
	•	headers-only means source files must never use forward declarations.
	•	use-if-needed still allows them in source files, but only when strictly required — avoiding the extra decls that always would introduce.

That's what I meant.

kimgr:
I'm still thinking about this when I have some spare time. Naming is the worst.

This is my latest take:
```
--fwd_decls=disabled              # alias for --no_fwd_decls
--fwd_decls=enabled               # current default 
--fwd_decls=enabled-headers-only  # JoHuang's new mode
--fwd_decls=enabled-minimized     # Bolshakov-a's future drop mode
```

I guess this would translate into an enum in `iwyu_globals`, which we can then condition on to use the respective policies throughout IWYU.

We shouldn't add `enabled-minimized` yet, but I think it's important to think about how it will fit into the naming scheme when/if it happens.

@JoHuang @bolshakov-a Thoughts?

bolshakov-a:
I don't mind. In fact, I don't care very much about naming because, in a typical use case, a user has to specify it only once in a build script.

But honestly, "minimizing" may be misleading. IWYU has [some policies about "type provision"](https://github.com/include-what-you-use/include-what-you-use/blob/0.25/docs/WhatIsAUse.md#automatic-re-export) and it would be good to be able to customize them (https://github.com/include-what-you-use/include-what-you-use/issues/1578#issuecomment-2283764995, CC @firewave). "Minimize" may be read as "force `#include` for typedefs, function return types, parameter types" despite these policies should be orthogonal.

JoHuang:
I feel the term `enabled` shows up a bit repetitively, but that’s fine.  

For new users, `minimized` may sound like “very few fwd decls,” while in fact `headers-only` uses even fewer.  
Since the `minimized` mode won’t be implemented right away, this naming can be discussed later — it might eventually become the preferred default, but backward compatibility would need to be considered.  

For me, I’d just like to see this PR move forward.

bolshakov-a:
> For new users, minimized may sound like “very few fwd decls,” while in fact headers-only uses even fewer.

I think they would differ only in `#include` count, not in fwd-decl count.

> Since the minimized mode won’t be implemented right away, this naming can be discussed later

I agree.

firewave:
I still think my suggested policies basically cover it. Having the permissive modes allow it to incrementally adjust existing code bases and taking full types away will obviously have ripple effects and you probably do not want to address them all at once especially if you are dealing with a big code base.

It might also be worth to look what effect switching between the types has on the `clang-include-cleaner` output. If this would lead to flip-flops we should file an upstream ticket as such behavior will most certainly annoy user and even lead them to drop one of the tools.

Maybe we should also allow this to be configurable via a pragma as well to allow mixing of the policies (as a future improvement). In case of API headers you might want to have them always providing full types to ease their usage while the actual code base would rely on incomplete types. I have no idea if this use case is valid though. But let's stay on actually defining the policies for now.

kimgr:
@firewave so these are your proposed names, right?

> * `incomplete` - enforce usage of incomplete types
> * `full` - enforce usage of full types
> * `suggest-incomplete` - do not enforce and use incomplete types for suggestions (if necessary)
> * `suggest-full` - do not enforce and use full types for suggestions (if necessary / current default)

And you're thinking these are values of something like a `--types=` switch, right? So;

* `incomplete` ~= current default
* `full` ~= current `--no_fwd_decls`
* I'm not sure I understand the semantics of `suggest-incomplete` and `suggest-full`

But it looks overall like this is losing the headers vs. source files distinction?

kimgr:
@JoHuang
> For me, I’d just like to see this PR move forward.

Yeah, sorry this got blocked. If you want to try some things out, you could move ahead with adding a `--fwd_decls` switch and making `--no_fwd_decls` (without arguments) an alias of `--fwd_decls=<whatever name we pick>`.

But I would really like to settle on strong names before merging, because it's easier to understand and extend that way, and we can't easily change our minds without breaking users.

firewave:
> * `incomplete` ~= current default

`suggest-full` is the current default based on the context I posted it in. It allows full and incomplete types (i.e. no enforcement) but will suggest full types. Actually I think this is consistent and in some cases it might suggest incomplete type!? That probably needs to be checked and made consistent as the first step.

> * I'm not sure I understand the semantics of `suggest-incomplete` and `suggest-full`

Those would be the permissive modes where those would only be applied to suggestions (i.e. new code) but not to existing code which already adheres to either (i.e. old/legacy code). That way you would accumulate less technical debt while you incrementally update your code base.

> But it looks overall like this is losing the headers vs. source files distinction?

That's what I meant with "basically".

Maybe this should be addressed incrementally. First make sure the current fixed behavior is consistent (as mentioned above). And as the next step make it (internally) configurable which to suggest. Then expose that to the user. Afterwards the ability to enforce it. And then the header/source distinction. And potentially the pragma stuff. Some steps might need to be switched around but I think the steps in itself seem reasonable.

bolshakov-a:
Let's not get messed up. The policies about re-exporting in certain contexts and about suppressing forward-declarations are separate policies. The first one affects not only the presence of a fwd-decl at a function/typedef declaration site but also the presence or absence of a `#include` at use sites. The second one is about suppressing (or recategorization to full use) fwd-decls independently of which kind of declaration triggers reporting of them. I just wanted to draw attention that another command-line switch with slightly similar meaning may appear in the future, so the naming should be as less confusing as possible, ideally. The switches/policies _could_ be united somehow, but it should be a distinct decision.

kimgr:
@JoHuang Please rebase on latest master before moving forward, I've put some effort into cleaning up the `IsInHeader` semantics to be quite a bit better. I'd still love to see a first draft with `--fwd_decl=<whatever names>` to be able to play with it.

PR #1793, opened 2025-08-05
Fix clang compatibility

bolshakov-a:
This makes IWYU compatible with https://github.com/llvm/llvm-project/commit/4a9eaad9e1283b872788832d5bce7e7945b97c78 and https://github.com/llvm/llvm-project/commit/91cdd35008e9ab32dffb7e401cdd7313b3461892, meanwhile fixing a couple of minor bugs.


kimgr:
> dyn_cast_or_null helper function is deprecated now, so dyn_cast_if_present has been used instead.

How sad. I'm increasingly finding short evocative names are better than long, precise ones :-)

iwyu.cc
```diff
@@ -3141,8 +3142,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   bool VisitNestedNameSpecifier(NestedNameSpecifier* nns) {
     // If this is a NamespaceAlias, then mark the alias as used,
     // requiring whichever file(s) declare the alias.
-    if (nns->getKind() == NestedNameSpecifier::NamespaceAlias) {
-      ReportDeclUse(CurrentLoc(), nns->getAsNamespaceAlias());
+    if (auto* alias =
```
kimgr: Could you make this `const auto*`?

kimgr:
Yikes, this looks like it's going to be fun to adjust to: https://github.com/llvm/llvm-project/commit/91cdd35008e9ab32dffb7e401cdd7313b3461892

iwyu.cc
bolshakov-a: I just recalled our conversation that `const` is not worth being inserted except pointed-to and referred-to types. I think the reason here is as follows: you cannot compile the code if you try to assign a pointer to const to a pointer to non-const. Moreover, even if the function returns a pointer or a reference to non-const, placing `const` makes the code more robust against refactoring when the function becomes returning ref or ptr to const.

However, this becomes irrelevant when using `auto*` because `auto*` may be deduced to `const Type*` if the function returns it.

kimgr: Yeah, but I find type deduction hides relevant information here. The constness is an interesting property whether we spell out the type or not.

bolshakov-a:
> Yikes, this looks like it's going to be fun to adjust to: https://github.com/llvm/llvm-project/commit/91cdd35008e9ab32dffb7e401cdd7313b3461892

Maybe, somebody will revert it...

iwyu_ast_util.cc
```diff
@@ -1610,7 +1564,7 @@ bool HasImplicitConversionConstructor(const Type* type) {
     return false;  // can't implicitly convert to a non-const reference
 
   type = RemoveReferenceAsWritten(type);
-  const NamedDecl* decl = TypeToDeclAsWritten(type);
+  const NamedDecl* decl = GetDefinitionAsWritten(TypeToDeclAsWritten(type));
```
bolshakov-a: This problem manifested itself in the dogfood results (`Twine` for `iwyu_driver.cc` and `iwyu_path_util.cc`). Probably, it is worth writing a test case?

kimgr: Yeah, that might be nice. What's the trigger then? `TypeToDeclAsWritten` returned a forward-decl of `Twine`?

kimgr:
> -#include <functional>

Interesting. I haven't been able to find any uses of `<functional>` where they were removed, so maybe this is an improvement? It's a bit hard to verify, as the other STL headers probably provide any used parts of `<functional>`.

> +#include "clang/Basic/CustomizableOptional.h"

This comes up in iwyu_preprocessor.cc also, I've just added a;
```
// IWYU pragma: no_include "clang/Basic/CustomizableOptional.h"
```

to suppress it in dogfood for now.

iwyu_ast_util.cc
bolshakov-a: Seemingly it is. It called under the hood `RecordType::getDecl()` which returned the definition, but now it calls `RecordType::getOriginalDecl()` which returns just some declaration. I think all such places in IWYU need reviewing.

bolshakov-a:
It is strange that removing of `<functional>` was not suggested in the first dogfood run. I think it would better be investigated along with the `CustomizableOptional` appearance because it may be a manifestation of some other bugs. However, I'm afk today, so feel free to finalize this PR, but note that [our conversation](https://github.com/llvm/llvm-project/pull/147835#issuecomment-3179438371) with @mizvekov suggests that passing `false` to `TraverseType` may be wrong, and the PR needs reworking. OTOH, `BaseAstVisitor` does not pass `QualifiedTypeLoc`s:
https://github.com/include-what-you-use/include-what-you-use/blob/82e6005f7387a76831324438458eda12aea45191/iwyu.cc#L444

bolshakov-a:
> I haven't been able to find any uses of <functional> where they were removed

You may just run the previous IWYU revision and see why it suggested the header.

kimgr:
> > I haven't been able to find any uses of  where they were removed
> 
> You may just run the previous IWYU revision and see why it suggested the header.

Ah, yes, I was just doing that. Had some errands to run while LLVM rebuilt.at llvm/llvm-project@28417e6459bb5174b9502f440e3dbb86f7a0046e.

Turns out it's for `std::function` [here](https://github.com/include-what-you-use/include-what-you-use/blob/82e6005f7387a76831324438458eda12aea45191/iwyu.cc#L5437) and [here](https://github.com/include-what-you-use/include-what-you-use/blob/82e6005f7387a76831324438458eda12aea45191/iwyu_cache.cc#L95).

While I think it's correct that `<functional>` is not required (should be provided by the callee decls), I'm not sure it's omitted for the right reasons now.

iwyu_ast_util.cc
bolshakov-a: Probably, this change is the reason for the `<functional>` removing suggestions? `TypeToDeclAsWritten` may probably return not the full definition for templates like `std::function` both earlier and now?

kimgr: I'll see if I can fold the GetDefinitionAsWritten into TypeToDeclAsWritten and see what happens.

kimgr: Nope, that broke 25 testcases on my branch before NNS changes.

kimgr: But yes, removing/adding `GetDefinitionAsWritten` here in `HasImplicitConversionConstructor` is what triggers `<functional>` being required or not.

kimgr: OK, I see what's happening now:

* without `GetDefinitionAsWritten`, `HasImplicitConversionCtor(const NamedDecl*)` does not think the `std::function` decl is an explicit specialization, and so resolves it to the specialized template templated decl, which, for some reason does NOT have an implicit ctor
* with `GetDefinitionAsWritten`, `HasImplicitConversionCtor(const NamedDecl*)` thinks the `std::function` decl is an explicit specialization, and so leaves the incoming `cxx_class` alone, which HAS an implicit ctor

So if `std::function` indeed has an implicit conversion from a lambda (which seems reasonable), and the autocast rule should hold, the removal is correct. So the patch behavior is more correct. But the fact that there's ambiguity at all seems to indicate something's wrong in `HasImplicitConversionCtor` or `HasImplicitConversionConstructor` in how they reason about template specializations.

bolshakov-a: Good investigation, thanks!

One thing came into my mind: maybe this constructor enumeration is redundant at all at the call site because, if the code compiles, an implicit conversion ctor should be. And if there isn't any, the user should write the type explicitly, and IWYU should report the type anyway.

The main thing is that the call site logic should be compatible to the function declaration site logic.

bolshakov-a: Anyway, this change in the IWYU behavior looks correct to me as well. But if you want to restore the old behavior, you may wrap into `GetDefinitionAsWritten` directly `getOriginalDecl()` calls. Maybe, it is the most reliable way to adjust IWYU to the changes in Clang.

bolshakov-a: > But if you want to restore the old behavior, you may wrap into GetDefinitionAsWritten directly getOriginalDecl() calls.

Ah yes, 26 broken tests, I see...

bolshakov-a:
As for `CustomizableOptional.h`: the cause is not 40395c3fc but 409e09d65. The non-member `operator<` for `OptionalFileEntryRef` (i.e. `CustomizableOptional`) is suggested now for `~IwyuPreprocessorInfo()`.

The following description is valid for libc++-12. When scanning the implicitly generated `IwyuPreprocessorInfo` destructor, IWYU analyzes the `std::map` destructor for e.g. `map<clang::OptionalFileEntryRef, IwyuFileInfo> iwyu_file_info_map_` field, which in turn calls the destructor of its internal field of `_Rb_tree` templated type. `~_Rb_tree()` calls `_Rb_tree::_M_erase` which definition __is out of line__ and thus contains `_Rb_tree<...>::` prefix which causes now IWYU to scan the whole `_Rb_tree` class template. It contains a nested structure `_Rb_tree::_Auto_node`, and IWYU traverses it. One of its methods is `_M_insert` which eventually causes the use of `operator<` of the map key type (`OptionalFileEntryRef`) to determine the correct place to insert new nodes.

Note that `std::map` fields of `IwyuPreprocessorInfo` don't cause by itself `_Rb_tree` scanning because `std::map` is one of "precomputed list" templates, internals of which are not usually scanned by IWYU. Probably, IWYU should not traverse their methods as well...

Besides the issue that `OptionalFileEntryRef` type alias should probably provide not only the aliased type but also `operator<` for it, __the scanning of `_Rb_tree::_Auto_node` is a bug__. Compilers don't require template argument types to be complete when used inside a nested class, and not only in its methods but also in its fields. For example, in such case:
```cpp
#include "tests/cxx/direct.h"

template <class T>
struct Tpl {
  struct Nested {
    T t;
   };
};
Tpl<IndirectClass> t;
```
the complete `IndirectClass` type is not needed, but IWYU requires it for the variable `t` (if somewhere else in the translation unit `Tpl<IndirectClass>::Nested` is used, otherwise Clang doesn't instantiate it, and IWYU doesn't see it.)

 The fix is simple:
```patch
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -4334,7 +4334,7 @@ class InstantiatedTemplateVisitor
 
     for (DeclContext::decl_iterator it = class_decl->decls_begin();
          it != class_decl->decls_end(); ++it) {
-      if (isa<CXXMethodDecl>(*it) || isa<FunctionTemplateDecl>(*it))
+      if (isa<CXXMethodDecl>(*it) || isa<FunctionTemplateDecl>(*it) || isa<RecordDecl>(*it))
         continue;
```
However, in this case, we should invent a way to report `IndirectClass` on `Tpl<IndirectClass>::Nested` use.

Maybe, it would be better to remove 409e09d65 from this PR (despite I still think that that commit is a correct change) and to find a way to keep `ASTNode::DynCast` for `NestedNameSpecifier`s?

iwyu_ast_util.cc
bolshakov-a: > But if you want to restore the old behavior, you may wrap into GetDefinitionAsWritten directly getOriginalDecl() calls.

The correct way is the following:
```patch
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1519,7 +1519,9 @@ static const NamedDecl* TypeToDeclImpl(const Type* type, bool as_written) {
     // A template type alias
     return template_decl;
   } else if (const RecordType* record_type = type->getAs<RecordType>()) {
-    return record_type->getOriginalDecl();
+    const RecordDecl* declaration = record_type->getOriginalDecl();
+    const RecordDecl* definition = declaration->getDefinition();
+    return definition ? definition : declaration;
   } else if (const TagType* tag_type = DynCastFrom(type)) {
```
For enums (i.e. the `TagType` branch below) it is not currently needed. Not sure about `InjectedClassNameType`s. For `ReportDeclUse` inside `IwyuAstConsumer::VisitTagType`, it is also not needed because `GetDefinitionAsWritten` is called later inside `ReportFullSymbolUse`.

Maybe, it would be actually better not to affect templates in this PR even if the removing of `<functional>` would be a real improvement, just to separate changes from each other.

UPD: I overlooked a simpler way:
```patch
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1519,7 +1519,7 @@ static const NamedDecl* TypeToDeclImpl(const Type* type, bool as_written) {
     // A template type alias
     return template_decl;
   } else if (const RecordType* record_type = type->getAs<RecordType>()) {
-    return record_type->getOriginalDecl();
+    return record_type->getOriginalDecl()->getDefinitionOrSelf();
   } else if (const TagType* tag_type = DynCastFrom(type)) {
```

bolshakov-a: As regards `HasImplicitConversionCtor`: the code of `std::function` (in libstdc++ implementation) is in [its _partial_ specialization](https://github.com/gcc-mirror/gcc/blob/b125eee2d8526538476834773fb8216035d08598/libstdc%2B%2B-v3/include/bits/std_function.h#L324).
> without GetDefinitionAsWritten, HasImplicitConversionCtor(const NamedDecl*) does not think the std::function decl is an explicit specialization...

Because `TypeToDeclAsWritten` (with or without `getDefinition()` inside) returns `ClassTemplateSpecializationDecl` which corresponds to the _implicit_ instantiation (instantiated from the _explicit_ partial specialization).

> ... and so resolves it to the specialized template templated decl, which, for some reason does NOT have an implicit ctor

It is [just empty](https://github.com/gcc-mirror/gcc/blob/b125eee2d8526538476834773fb8216035d08598/libstdc%2B%2B-v3/include/bits/std_function.h#L112-L113).

> with GetDefinitionAsWritten, HasImplicitConversionCtor(const NamedDecl*) thinks the std::function decl is an explicit specialization...

Because `GetDefinitionAsWritten` transforms the given implicitly instantiated decl into `ClassTemplateSpecializationDecl` corresponding to the (explicit) partial specialization (i.e. definition).

> But the fact that there's ambiguity at all seems to indicate something's wrong in HasImplicitConversionCtor or HasImplicitConversionConstructor in how they reason about template specializations.

Yeah, I haven't completed the getting to definition inside `HasImplicitConversionCtor` and leaved `TODO` about partial specializations there. But it looks like just reimplementing `GetDefinitionAsWritten`, so it should probably be removed in favor of the latter. Maybe, not in this PR... However, if we leave `GetDefinitionAsWritten` in this PR, my code from `HasImplicitConversionCtor` becomes redundant and may be removed.

bolshakov-a: I've added a test case about the original problem with `Twine`.

kimgr: > Maybe, it would be actually better not to affect templates in this PR even if the removing of `<functional>` would be a real improvement, just to separate changes from each other.

Yeah, I think it would be nice to address that separately, if possible.

kimgr: > UPD: I overlooked a simpler way:
> [... TypeToDeclImpl ...]

I made a change in that spirit on current IWYU master + llvm before the breaking patch:
```diff
@@ -223,6 +223,18 @@ static bool IsImplicitConversionCtor(const CXXConstructorDecl* ctor) {
          !ctor->isCopyOrMoveConstructor();
 }
 
+static const NamedDecl* GetDefinitionOrSelf(const TagDecl* decl) {
+  const NamedDecl* definition = decl->getDefinition();
+  if (definition) {
+    return definition;
+  }
+  return decl;
+}
+
+static const NamedDecl* GetDefinitionOrFirst(const TagType* type) {
+  return GetDefinitionOrSelf(type->getDecl());
+}
+
 }  // anonymous namespace
 
 //------------------------------------------------------------
@@ -1565,7 +1577,7 @@ static const NamedDecl* TypeToDeclImpl(const Type* type, bool as_written) {
     // A template type alias
     return template_decl;
   } else if (const RecordType* record_type = type->getAs<RecordType>()) {
-    return record_type->getDecl();
+    return GetDefinitionOrFirst(record_type);
   } else if (const TagType* tag_type = DynCastFrom(type)) {
     return tag_type->getDecl();    // probably just enums
   } else if (template_decl) {
```

All tests still pass and dogfood results are unchanged... so it seems like a good thing?

I think we can leave `HasImplicitConversion...` as-is for the purposes of this PR, and then follow up with a cleanup?

kimgr:
> As for CustomizableOptional.h [...]

Yikes, what a deep-dive! OK, that makes a lot of sense.

> Maybe, it would be better to remove https://github.com/include-what-you-use/include-what-you-use/commit/409e09d654c121bd6201c673cbd77501ba7b33ec from this PR (despite I still think that that commit is a correct change) and to find a way to keep ASTNode::DynCast for NestedNameSpecifiers?

Yeah, I think it would be nice to make the most focused possible change to cover for the Clang changes.

If this is the only place we use `DynCast` for `NestedNameSpecifier`, is there a chance we could do it some other way? I'm thinking something like (very rough sketch):
```
if (const auto* method = ast_node->GetAncestorAs<CXXMethodDecl>()) {
  if (method->getQualifier() != nullptr) {
    return true;
  }
}
```

kimgr:
I tried that on IWYU master+llvm pre-break, and there's one test failure:
```
FAIL: test_reexport_overridden (__main__.cxx.test_reexport_overridden)
  File "/home/kimgr/code/iwyu/include-what-you-use/./run_iwyu_tests.py", line 93, in <lambda>
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kimgr/code/iwyu/include-what-you-use/iwyu_test_util.py", line 696, in TestIwyuOnRelativeFile
AssertionError: 
tests/cxx/reexport_overridden.cc:52: Unmatched regex:
FwdRetType is...*reexport_overridden-i2.h

tests/cxx/reexport_overridden.cc:55: Unmatched regex:
FwdRetType is...*reexport_overridden-i2.h

tests/cxx/reexport_overridden.cc:77: Unmatched regex:
IndirectClass is...*indirect.h
```
So probably some more detail is necessary.

kimgr:
This appears to work in my pre-environment:
```diff
diff --git a/iwyu.cc b/iwyu.cc
index eaa91913..86ce2052 100644
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3880,10 +3880,9 @@ class InstantiatedTemplateVisitor
     // |-NNS------^
     // |-CXXMethodDecl--^
     ASTNode* ast_node = current_ast_node();
-    if (const auto* nns = ast_node->GetParentAs<NestedNameSpecifier>()) {
-      if (nns->getAsType() == type) {
-        if (const auto* method = ast_node->GetAncestorAs<CXXMethodDecl>(2)) {
-          CHECK_(nns == method->getQualifier());
+    if (const auto* method = ast_node->GetAncestorAs<CXXMethodDecl>()) {
+      if (const NestedNameSpecifier* nns = method->getQualifier()) {
+        if (nns->getAsType() == type) {
           return true;
         }
       }
```

I'm guessing it should be easier to port to new NNS design?

bolshakov-a:
I've decided to spend some time on trying to accommodate passed-by-value NNS into `ASTNode`.

kimgr:
@bolshakov-a The way it's no longer really part of the AST makes me a bit uncomfortable with forcing it into `ASTNode`. It feels like the `RecursiveASTVisitor` would be much more consistent if they just broke existing code and ripped `NestedNameSpecifier` out of the traversal entirely.

But yeah, as long as it _is_ traversed (and now Clang does it by default, if I understand things correctly) maybe it makes sense to fold it into `ASTNode`.

iwyu_ast_util.h
```diff
@@ -79,6 +81,16 @@ using std::string;
...
 // (such as a function parameter), meaning all types seen below
 // that node are legal to fowrard-declare according to c++.
+
+namespace detail {
+template <typename T>
+const T* null_value = nullptr;
+
+template <>
+inline clang::NestedNameSpecifier null_value<clang::NestedNameSpecifier> =
+    std::nullopt;
```
bolshakov-a, line 91: Braced initialization would be more robust against possible refactoring if the NNS constructor suddenly becomes explicit. But maybe, copy-initialization looks more readable.

iwyu_ast_util.h
```diff
@@ -96,11 +108,19 @@ class ASTNode {
...
+    CHECK_(*nns);
+  }
   ASTNode(const clang::NestedNameSpecifierLoc* nnsloc)
-      : kind_(kNNSLocKind), as_nnsloc_(nnsloc),
-        parent_(nullptr), in_fwd_decl_context_(false) { }
+      : kind_(kNNSLocKind),
+        as_nnsloc_(nnsloc),
+        parent_(nullptr),
+        in_fwd_decl_context_(false) {
+    CHECK_(nnsloc->getNestedNameSpecifier());
```
bolshakov-a: Hoping that it is always `true`. It would be good to hear from @mizvekov whether null NNS may appear in the AST or not.

iwyu_ast_util.h
```diff
@@ -96,11 +108,19 @@ class ASTNode {
       : kind_(kTypelocKind), as_typeloc_(typeloc),
         parent_(nullptr), in_fwd_decl_context_(false) { }
   ASTNode(const clang::NestedNameSpecifier* nns)
-      : kind_(kNNSKind), as_nns_(nns),
-        parent_(nullptr), in_fwd_decl_context_(false) { }
+      : kind_(kNNSKind),
+        as_nns_(nns),
+        parent_(nullptr),
+        in_fwd_decl_context_(false) {
+    CHECK_(*nns);
```
kimgr: Dogfood indicates that this is somehow attributed to the definition of `CHECK_` :weary:

bolshakov-a: No, the erroneous fwd-decl suggestions is the newly uncovered bug related to template variable. Let it be a template function instead.

bolshakov-a: But yeah, `CHECK_` also causes erroneous attribution of NNS to the macro location (`iwyu_port.h`). If so, let's not check.

iwyu_ast_util.cc
bolshakov-a: Ok, `GetDefinitionAsWritten` replaced by `getDefinitionOrSelf`.

I've [posted a question](https://github.com/llvm/llvm-project/pull/147835#discussion_r2280520266) about `InjectedClassNameType`.

iwyu.cc
```diff
@@ -454,32 +454,32 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
...
-    return Base::TraverseTypeLoc(typeloc);
+    return Base::TraverseTypeLoc(typeloc, traverse_qualifier);
   }
 
-  bool TraverseNestedNameSpecifier(NestedNameSpecifier* nns) {
+  bool TraverseNestedNameSpecifier(NestedNameSpecifier nns) {
     if (!nns)
       return true;
-    ASTNode node(nns);
+    ASTNode node(&nns);
```
kimgr, line 463: I'm a bit surprised the address of the argument `nns` is valid throughout the whole traversal of the subtree under `NestedNameSpecifier`. But it has to be, because `node` has the same lifetime, and that's definitely alive... Right?

bolshakov-a: Yeah, correct. The same trick already was in `TraverseNestedNameSpecifierLoc` below.

iwyu_ast_util.h
mizvekov: Since the comment is outdated I can't look at the specific code this was talking about.

But a NNS of kind null appears in the same places a nullptr NNS * would before my patch, when the type in question was written with no / empty name qualifier.

iwyu_ast_util.h
bolshakov-a: But I'm not sure that it is the right way. Probably, it would be better to place pragmas instead so as not to forget about the bugs?

iwyu.cc
bolshakov-a: But maybe NNS should better be stored in the `ASTNode` by value as well.

iwyu_ast_util.h
bolshakov-a: Ah, sorry! IWYU has [explicit checks](https://github.com/include-what-you-use/include-what-you-use/pull/1793/files#diff-7ca107c8b836c1765e8b735b74f06833bec48f308a76017886cee1d5736002d9R476) that NNS is not null before constructing `ASTNode`, hence a null NNS should simply not get here.

iwyu_ast_util.h
kimgr: I'm sure they'll resurface :-)

iwyu.cc
kimgr: I think it's fine as-is. There doesn't appear to be a copy-ctor on NNS, so it takes some shuffling to store it by-val.

bolshakov-a: Sorry? This compiles:
```cpp
NestedNameSpecifier nns2 = nns;
```
or I haven't understood you correctly.

iwyu_ast_util.h
```diff
@@ -151,34 +165,6 @@ class ASTNode {
   // current node knows its location, returns an invalid SourceLocation.
   clang::SourceLocation GetLocation() const;
 
-  // Returns true if this node points to the exact same
```
kimgr, line 154: Could you avoid moving these? The patch is already huge, so it would be nice to minimize unnecessary changes.

bolshakov-a: No, it doesn't compile otherwise.

iwyu.cc
kimgr: Oh, it does? I had problems earlier when I was playing with it, but I think I was targeting the wrong llvm tree or something. Anyway, I think the current impl is fine.

iwyu_ast_util.h
kimgr: Gotcha. `auto` at work.

kimgr:
Alright, I think this looks good for the most part. I wonder if the new tests could be added in a separate commit afterwards, because it's nice to be able to back/forward-port clang-compat patches, and the smaller they are, the easier that becomes.

bolshakov-a:
> I wonder if the new tests could be added in a separate commit afterwards

A couple of them fixes the changes in the behavior, as stated in the commit message. Maybe, remaining should be moved into a separate commit _before_ the compatibility commit to emphasize that they fix the old behavior?

iwyu_ast_util.cc
bolshakov-a: I remember that a couple of days ago removing of `InjectedClassNameType` branch inside `TypeToDeclImpl` caused `badinc` test failure inside `I2_TemplateClass<FOO>::CcFileFn`, but now the test passes. That's strange.

Anyway, it is safer to put `getDefinitionOrSelf` there along with a `TODO`.

iwyu_ast_util.h
bolshakov-a: In principle, it may be circumvented by replacing `auto` with `decltype(detail::GetNullValue<To>())`. Or just by separation of the moving into a distinct commit before the compatibility commit.

iwyu.cc
```diff
@@ -3454,13 +3453,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-    // analysis when passing 'ElaboratedType' into it.
-    if (const auto* elaborated =
-            current_ast_node()->template GetParentAs<ElaboratedType>()) {
-      ReportTypeUse(CurrentLoc(), elaborated, DerefKind::None);
-    } else {
-      ReportTypeUse(CurrentLoc(), type, DerefKind::None);
-    }
+    // analysis when passing elaborated type into it.
+    ReportTypeUse(CurrentLoc(), type, DerefKind::None);
   }
```
bolshakov-a, line 3458: `ReportWrittenTypeAlias` function just delegates now to `ReportTypeUse`. I'm not sure if it should be leaved here.

iwyu_ast_util.cc
bolshakov-a: > That's strange.

The test fails on the revision prior to https://github.com/llvm/llvm-project/commit/91cdd35008e9ab32dffb7e401cdd7313b3461892. After that Clang change, `InjectedClassNameType` inherits `TagType` and can be handled by the corresponding branch.

iwyu.cc
kimgr: We can always remove it in a follow-up.

kimgr:
@bolshakov-a Sorry for the delay, I've been busy with the rest of my life :-). Are we still waiting for something (e.g. `InjectedClassNameType`), or should we merge this to get master back to green?

bolshakov-a:
I have nothing against merging this. I don't currently expect any regression.

Probably, just the issue about `TraverseQualifier` needs clarification:
https://github.com/llvm/llvm-project/pull/147835#issuecomment-3185811000

kimgr:
@bolshakov-a Ah, right. Yeah, if something changes I guess we can follow up later.

PR #1794, opened 2025-08-05
[clang compat] Change type_info tag kind

bolshakov-a:
https://github.com/llvm/llvm-project/commit/2e404d1f800c804b5088ad1f54f24738efbdf9cd changed the kind of `type_info` implicitly added in the MSVC compatibility mode from class to struct.

This should be merged after #1793.

### DIFF
--- a/tests/cxx/clmode.cc
+++ b/tests/cxx/clmode.cc
@@ -27,13 +27,13 @@ type_info* pt = nullptr;
 
 tests/cxx/clmode.cc should add these lines:
 #include "tests/cxx/indirect.h"
-class type_info;
+struct type_info;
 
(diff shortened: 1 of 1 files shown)


bolshakov-a:
This is not needed after https://github.com/llvm/llvm-project/commit/16766b3aef281a24acad08292ffe69554b44805b. Maybe it is worth merging and then reverting specifying the LLVM commit that fixed it, maybe not.

kimgr:
Thanks for the detective work! If it's fixed on llvm mainline now, I don't think it's worth recording in our commit log that this was broken upstream for a day.

bolshakov-a:
Ok, closing.

PR #1798, opened 2025-08-23
Fix opaque declarations of elaborated enums

bolshakov-a:
C and C++ don't allow elaborated types to introduce new enumerations, so opaque declarations of enumerations with fixed underlying types are still needed. If the underlying type is not fixed, the full definition is required, but in this case CanForwardDeclareType returns `false`.

Prior to this, for elaborated types naming enumerations with fixed underlying type, IWYU required neither the full definition nor an opaque declaration.


kimgr:
Looks good, thanks!

Issue #1763, opened 2025-06-03
<tuple> suggested for `std::array`


hebasto:
Here is a minimum reproducible example ( `array_tuple.cpp`):
```c++
#include <array>
#include <string>

void f(std::array<std::array<int, 2>, 3> a) {}

void g(std::string s) {}
```

Toolchain details:
```
$ include-what-you-use --version       
include-what-you-use 0.24 (git:6e08906) based on Ubuntu clang version 20.1.7 (++20250528073343+7cf14539b644-1~exp1~20250528193358.125)
$ clang++-20 -v
Ubuntu clang version 20.1.7 (++20250528073343+7cf14539b644-1~exp1~20250528193358.125)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-20/bin
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/13
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/13
Candidate multilib: .;@m64
Selected multilib: .;@m64
```

IWYU output:
```
$  include-what-you-use array_tuple.cpp 

array_tuple.cpp should add these lines:
#include <tuple>   // for array

array_tuple.cpp should remove these lines:

The full include-list for array_tuple.cpp:
#include <array>   // for array
#include <string>  // for basic_string, string
#include <tuple>   // for array
---
```

Conditions required to reproduce:
1. Clang uses a GCC-13 installation.
2. `<string>` is included.
3. A 2-dimensional array is used.

The issue disappears under other conditions, for example after installing `g++-14`.

bolshakov-a:
Slightly reduced:
```cpp
#include <array>
#include <string>

std::array<int,2> a, *p;
```
IWYU reports forward-declaration of `std::array` from `<tuple>` (line 2019) for some reason, likely because names from `std` cannot be forward-declared in user's code, according to the standard.

bolshakov-a:
Reduced (not standard-conformant code):
```cpp
// Class.h
namespace std {
    template<int>
    class Class{};
}

// fwd.h
namespace std {
    template<int>
    class Class;
}

// fwd-direct.h
#include "fwd.h"

// main.cpp
#include "Class.h"
#include "fwd-direct.h"

std::Class<0> c, *p;
```
```
main.cpp should add these lines:
#include "fwd.h"    // for Class

main.cpp should remove these lines:
[...]

PR #1801, opened 2025-08-23
Handle __builtin_types_compatible_p

bolshakov-a:
`__builtin_types_compatible_p` is a C language trait that was almost like "is same" before C23. However, [N3037](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3037.pdf) allowed different tag type to be compatible in the same translation unit provided the tags are identical and the contents are similar. Types with identical tags may be different when they are in different scopes. The interesting for IWYU case is when one declaration is in the file scope and the other is in a block scope. The file scope declaration may be in a header file and may be required to correctly determine type compatibility.

The types to report may be buried under an arbitrary number of derived types like pointers or function types. Moreover, field types of structs and unions should be recursively handled as well. For example:
```c
struct Inner { int i; };
struct Outer { struct Inner** p; };
extern struct Outer o;

int Fn() {
  struct Inner { int i; };
  struct Outer { struct Inner** p; };
  return __builtin_types_compatible_p(struct Outer, typeof(o));
}
```
Not only `Outer` but also `Inner` should be reported at the trait location for the trait to be evaluated correctly. OTOH, if `Outer` structures contained another fields which differ in type or name, reporting `Inner` would better to be avoided because the trait is guaranteed to be 0. The only way to achieve this is to (almost) evaluate the trait inside IWYU.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -2670,15 +2670,38 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         // is already there, and the aliased template should be able to be just
         // fwd-declared by the author.
         return true;
-      default:
-        for (const TypeSourceInfo* arg : expr->getArgs()) {
-          QualType qual_type = arg->getType();
-          const Type* type = qual_type.getTypePtr();
-          ReportTypeUse(CurrentLoc(), type, DerefKind::RemoveRefsAndPtr);
-        }
-
+      case TypeTrait::BTT_TypeCompatible: {
+        // __builtin_types_compatible_p is a C language trait that checks if
+        // the types are compatible. Since C23, two different record types
+        // defined in the same translation unit can be compatible.
+        // Types with different cvr-qualification are incompatible according to
+        // the C standard, but this trait ignores top-level qualifiers.
+        // getUnqualifiedArrayType is used to remove qualifiers from the element
+        // type if the argument is an array type.
+        QualType lhs_unqual_type =
+            compiler()->getASTContext().getUnqualifiedArrayType(lhs_qual_type);
+        QualType rhs_unqual_type =
+            compiler()->getASTContext().getUnqualifiedArrayType(rhs_qual_type);
+        auto provided_getter = [this](const Type* type) {
+          if (const auto* typedef_type = type->getAs<TypedefType>()) {
+            const TypedefNameDecl* decl = typedef_type->getDecl();
+            return GetProvidedTypes(decl->getUnderlyingType().getTypePtr(),
+                                    GetLocation(decl));
+          }
+          return set<const Type*>{};
+        };
+        // ReportTypeUse doesn't report enumerations. But for this trait,
+        // the enum content may be important, hence ReportDeclsUse is used
+        // instead.
+        CompatibilityChecker chkr{provided_getter, compiler()->getASTContext()};
+        if (chkr.CouldBeCompatible(lhs_unqual_type, rhs_unqual_type))
+          ReportDeclsUse(CurrentLoc(), chkr.GetDeclsToReport());
         return true;
+      }
     }
+    VERRS(4) << "Unexpected type trait: " << getTraitName(expr->getTrait())
+             << '\n';
+    return true;
   }
 
   // Mark that we need the full type info for the thing we're taking
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -53,16 +53,20 @@
 #include "iwyu_verrs.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/PointerUnion.h"
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/Casting.h"
 
 // TODO: Clean out pragmas as IWYU improves.
 // IWYU pragma: no_include "clang/AST/StmtIterator.h"
 // IWYU pragma: no_include "clang/Basic/CustomizableOptional.h"
+// IWYU pragma: no_include <iterator>
+// IWYU pragma: no_include <tuple>
 
 using clang::ASTContext;
 using clang::ASTDumper;
 using clang::ArrayType;
+using clang::AtomicType;
 using clang::BlockPointerType;
 using clang::CXXConstructExpr;
 using clang::CXXConstructorDecl;
@@ -77,6 +81,7 @@ using clang::CallExpr;
 using clang::ClassTemplateDecl;
 using clang::ClassTemplatePartialSpecializationDecl;
 using clang::ClassTemplateSpecializationDecl;
+using clang::ConstantArrayType;
 using clang::Decl;
 using clang::DeclContext;
 using clang::DeclRefExpr;
@@ -94,6 +99,7 @@ using clang::Expr;
 using clang::ExprResult;
 using clang::ExprWithCleanups;
 using clang::FunctionDecl;
+using clang::FunctionProtoType;
 using clang::FunctionTemplateDecl;
 using clang::FunctionTemplateSpecializationInfo;
 using clang::FunctionType;
@@ -161,8 +167,11 @@ using llvm::dyn_cast;
 using llvm::dyn_cast_or_null;
 using llvm::errs;
 using llvm::isa;
+using llvm::range_size;
 using llvm::raw_string_ostream;
+using llvm::zip_equal;
 using std::function;
+using std::pair;
 using std::vector;
 
 namespace include_what_you_use {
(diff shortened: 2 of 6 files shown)


iwyu_ast_util.cc
```diff
@@ -223,6 +230,142 @@ static bool IsImplicitConversionCtor(const CXXConstructorDecl* ctor) {
...
+// is used.
+static bool CouldBeCompatible(
+    QualType lhs,
+    QualType rhs,
+    bool full_type_guaranteed,
+    set<pair<const RecordDecl*, const RecordDecl*>>& handled,
+    set<const Type*> provided_types,
+    function<set<const Type*>(const Type*)> provided_getter,
+    ASTContext& ctx,
+    set<const NamedDecl*>& decls_to_report) {
```
bolshakov-a: Too many parameters. Probably, it should be a class?

iwyu_ast_util.cc
```diff
@@ -223,6 +230,142 @@ static bool IsImplicitConversionCtor(const CXXConstructorDecl* ctor) {
...
+  provided_types.merge(provided_getter(lhs.getTypePtr()));
+  rhs = rhs.getDesugaredType(ctx);
+  lhs = lhs.getDesugaredType(ctx);
+  if (lhs == rhs)
+    return true;
+  if (lhs.getQualifiers() != rhs.getQualifiers())
+    return false;
+  if (const auto* rhs_record = rhs->getAsRecordDecl()) {
+    if (const auto* lhs_record = lhs->getAsRecordDecl()) {
+      // TODO(bolshakov): handle unnamed and anonymous structs and unions?
```
bolshakov-a: There is an issue about unnamed structs:
https://github.com/llvm/llvm-project/issues/151394

For the case of anonymous ones, IWYU seems to work currently fine. Should I add some test cases on it?

kimgr: I can't say I know the difference between anonymous and unnamed structs...?

It might be good to have some test cases in place in case Clang eventually changes. But I guess we're reimplementing most of Clang's policy here, so we won't necessarily notice changes in Clang's implementation?

iwyu_ast_util.cc
kimgr: I'm fine with this as a function, but yeah, the param list is a little wild.

Sometimes it helps to split into a "launcher" and a "recursive" part to avoid the recursive worklist stuff in the launcher.

I think my instinct says keep the function as is, and we'll see what happens. All parameters are distinct types, so there's no major risk of passing the wrong thing, as far as I can see.

iwyu_ast_util.cc
bolshakov-a: > I can't say I know the difference between anonymous and unnamed structs...?

```c
struct Example {
  struct {  // anonymous
    int a;  // these behave as if they were direct members of struct Example
    int b;  // hence can be accessed as obj.a
  };
  struct {  // unnamed
    int a;
    int b;  // these can be accessed only as obj.field.a
  } field;
} obj;

struct {  // also unnamed
  int c, d;
} obj2;
```

> But I guess we're reimplementing most of Clang's policy here, so we won't necessarily notice changes in Clang's implementation?

`static_assert`s would fail if that happens. As for unnamed struct and union case, it seems to be not clang but a core language issue, so I'd prefer not to write tests on it unless it is resolved.

iwyu_ast_util.cc
bolshakov-a: > Sometimes it helps to split into a "launcher" and a "recursive" part to avoid the recursive worklist stuff in the launcher.

Launcher already is on the line 1940.

kimgr: :facepalm: Maybe I should've read the whole patch before offering advice :slightly_smiling_face: 

Anyway, the current function interface is fine by me, but if you want to play with it, maybe push some alternatives as commits on top, so we can evaluate them more easily?

bolshakov-a: > Maybe I should've read the whole patch before offering advice

No problem, I think it's Ok to write questions while reading a PR.

iwyu_ast_util.cc
kimgr: Ah, there we go, thanks.

I'm just thinking, at least for myself, I'm not monitoring core language revisions day-to-day, and I typically don't notice when Clang adopts them. So it might be nice with an alarm bell in the tests that lets us know what happened. Sounds like the current `static_assert`s (in the current tests?) would catch changes, so that's good enough for me.

bolshakov-a: > Sounds like the current static_asserts (in the current tests?) would catch changes

As I've noticed, I haven't written test cases on anonymous or unnamed structures. I can do that to fix the status quo and leave a comment that things may change.

bolshakov-a: I just try to follow the principle that every line of code should be covered by tests. If there are no lines of code, I don't know exactly what cases should be tested.

bolshakov-a:
`<iterator>` has been suggested for `std::pair` due to the same bug as described in #1763.

`<tuple>` has been suggested for range-based `for` loops over results of `zip_equal`. Besides it raises a couple of questions about type provision (#1802 and #1803), it should not be reported, IMHO, just because `std::tuple` is a name from the `std` namespace which should not be forward-declared according to the standard, hence full definitions for entities from `std` should be provided where explicitly written (and only there, just like e.g. typedef names).

So I've put pragmas to suppress these erroneous suggestions.

iwyu_ast_util.cc
bolshakov-a: I've added several test cases on anonymous and unnamed entities.

iwyu.cc
```diff
@@ -2670,15 +2670,38 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        set<const NamedDecl*> decls_to_report;
+        if (CouldBeCompatible(lhs_unqual_type, rhs_unqual_type, provided_getter,
+                              compiler()->getASTContext(), decls_to_report)) {
+          ReportDeclsUse(CurrentLoc(), decls_to_report);
         }
-
         return true;
+      }
     }
+    CHECK_UNREACHABLE_("Unexpected type trait.");
```
kimgr: Does this mean we now cover all known traits? That's awesome!

But also, let's replace this with a level 4 log so IWYU doesn't crash when a new trait is added.

iwyu_ast_util.cc
```diff
@@ -223,6 +232,142 @@ static bool IsImplicitConversionCtor(const CXXConstructorDecl* ctor) {
...
 
+// In certain contexts, the type is guaranteed by the language to be complete,
+// e.g. when it is the type of a record field, or an array element type.
+// To prevent redundant suggestion in such cases, full_type_guaranteed parameter
+// is used.
+static bool CouldBeCompatible(
+    QualType lhs,
+    QualType rhs,
+    bool full_type_guaranteed,
+    set<pair<const RecordDecl*, const RecordDecl*>>& handled,
```
kimgr: Output and input/output parameters are generally pointers in Google style (at least as it was defined in IWYU times) so that mutability is visible on the call site, i.e.
```
CouldBeCompatible(t1, t2, false, &handled, provided_types, a_getter, ast_context, &decls_to_report);
```
vs.
```
CouldBeCompatible(t1, t2, false, handled, provided_types, a_getter, ast_context, decls_to_report);
```
So please change from mutable reference to pointer for `handler` and `decls_to_report`.

iwyu_ast_util.cc
```diff
@@ -223,6 +232,142 @@ static bool IsImplicitConversionCtor(const CXXConstructorDecl* ctor) {
...
+           zip_equal(lhs_record->fields(), rhs_record->fields())) {
+        if (lhs_field->getIdentifier() != rhs_field->getIdentifier())
+          return false;
+        // No check for alignment specifiers because if the fields differ only
+        // in them, it is probably just a bug.
+        // The rules are applied to the field types recursively.
+        // Field types are guaranteed to be complete.
+        if (!CouldBeCompatible(lhs_field->getType(), rhs_field->getType(), true,
+                               handled, provided_types, provided_getter, ctx,
+                               decls_to_report))
```
kimgr: Usually easier to read multi-line-predicate if statements with braces.

iwyu_ast_util.cc
bolshakov-a: > if you want to play with it, maybe push some alternatives as commits on top, so we can evaluate them more easily?

I've added a commit transforming `CouldBeCompatible` function into a class. To be honest, I have poor knowledge and understanding of object-oriented design principles, so not sure it has been done correctly.

iwyu.cc
bolshakov-a: Compiler should produce a warning in the case if a new enum value unhandled in the switch has been added. We will probably notice it before a release?

kimgr: Probably, but there's no reason to crash for something we don't control.

iwyu_ast_util.cc
kimgr: > I just try to follow the principle that every line of code should be covered by tests. If there are no lines of code,
> I don't know exactly what cases should be tested.

Yeah, I like that. I guess there are some language constructs for which we don't have lines of code _in IWYU_, but that are still implicitly handled by Clang. As I understood it that was the case for anonymous types, so it made sense to me to record in tests what the current expected behavior is, in case Clang changes, and we need to add code to IWYU.

So I tend to think not only of lines of code, but also what the desired behavior is. That behavior might exist even without code lines. Anyway, judging from the great quality of your tests so far, you already agree :-)

iwyu.cc
```diff
@@ -2670,15 +2670,37 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        // ReportTypeUse doesn't report enumerations. But for this trait,
+        // the enum content may be important, hence ReportDeclsUse is used
+        // instead.
+        CompatibilityChecker chkr{provided_getter, compiler()->getASTContext()};
+        if (chkr.CouldBeCompatible(lhs_unqual_type, rhs_unqual_type))
+          ReportDeclsUse(CurrentLoc(), chkr.GetDeclsToReport());
         return true;
+      }
     }
+    VERRS(4) << "Unexpected type trait.\n";
```
kimgr: Let's include the trait name:
```
VERRS(4) << "Unexpected type trait: " << getTraitName(expr->getTrait()) << "\n";
```

iwyu_ast_util.cc
```diff
@@ -225,6 +234,133 @@ static bool IsImplicitConversionCtor(const CXXConstructorDecl* ctor) {
...
+        return false;
+      if (!full_type_guaranteed) {
+        if (!provided_types.count(rhs.getCanonicalType().getTypePtr()))
+          decls_to_report_.insert(rhs_record);
+        if (!provided_types.count(lhs.getCanonicalType().getTypePtr()))
+          decls_to_report_.insert(lhs_record);
+      }
+
+      // This recursion check has been placed after record type reporting
+      // because it might be not reported at previous occurences due to the type
```
kimgr: spelling nit: occurence -> occurrence

iwyu_ast_util.cc
```diff
@@ -225,6 +234,133 @@ static bool IsImplicitConversionCtor(const CXXConstructorDecl* ctor) {
...
+                               provided_types);
+    }
+    return false;
+  } else if (const auto* rhs_func = rhs->getAs<FunctionProtoType>()) {
+    if (const auto* lhs_func = lhs->getAs<FunctionProtoType>()) {
+      if (lhs_func->getNumParams() != rhs_func->getNumParams())
+        return false;
+      for (auto [lhs_param, rhs_param] :
+           zip_equal(lhs_func->param_types(), rhs_func->param_types())) {
+        // Parameter qualifications are ignored.
```
kimgr: A bit surprising that e.g. these two are compatible:
```
// immutable parameter
void(const int*);
// mutable parameter
void(int*);
```

But I'm guessing the standard/implementation says so?

iwyu_ast_util.cc
```diff
@@ -225,6 +234,133 @@ static bool IsImplicitConversionCtor(const CXXConstructorDecl* ctor) {
...
+                               provided_types)) {
+          return false;
+        }
+      }
+      return CouldBeCompatible(lhs_func->getReturnType(),
+                               rhs_func->getReturnType(), false,
+                               provided_types);
+    }
+    return false;
+  } else if (auto* rhs_atomic = rhs->getAs<AtomicType>()) {
```
kimgr: Could these be `const auto*` too, or do we use non-const members? Same question for enum types below.

iwyu_ast_util.h
```diff
@@ -929,6 +932,38 @@ bool IsConvertible(clang::QualType from,
                    clang::SourceLocation conv_loc,
                    clang::Sema&);
 
+class CompatibilityChecker {
+ public:
+  CompatibilityChecker(
+      std::function<set<const clang::Type*>(const clang::Type*)>
+          provided_getter,
+      clang::ASTContext& ctx)
+      : provided_getter_{provided_getter}, ctx_{ctx} {
```
kimgr, line 943: I'm still too old for initialization syntax, I don't get why it's better than simply `provided_getter_(provided_getter)`. But I guess this is where C++ is moving...?

iwyu.cc
```diff
@@ -2670,15 +2670,37 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+            const TypedefNameDecl* decl = typedef_type->getDecl();
+            return GetProvidedTypes(decl->getUnderlyingType().getTypePtr(),
+                                    GetLocation(decl));
+          }
+          return set<const Type*>{};
+        };
+        // ReportTypeUse doesn't report enumerations. But for this trait,
+        // the enum content may be important, hence ReportDeclsUse is used
+        // instead.
+        CompatibilityChecker chkr{provided_getter, compiler()->getASTContext()};
```
kimgr, line 2696: Iniitialization syntax looks weird to me here too. But I may not be representative :-)

iwyu_ast_util.cc
bolshakov-a: No, these are not compatible. These are compatible: `void(int* const)` and `void(int*)`.

https://godbolt.org/z/Ye1c8EETE

iwyu.cc
kimgr, line 2696: I think my instinct prefers the plain function, but the helper class does help clarify the parameter list(s) quite a bit. What's your feeling?

iwyu_ast_util.cc
bolshakov-a: > it that was the case for anonymous types

I would not say so. IWYU handling of records just occurred to be perfectly valid for anonymous records as well, as it is explicitly specified in the C standard, 6.2.7:
> For determining type compatibility, anonymous structures and unions are considered a regular member of the containing structure or union type, and the type of an anonymous structure or union is considered compatible to the type of another anonymous structure or union, respectively, if their members fulfill the preceding requirements.

so [the line](https://github.com/include-what-you-use/include-what-you-use/pull/1801/files#diff-7ec045b2b5739efe5504c9d31e7e935fe9405be09ae655d067bda89fdd9cf6f9R256)
```cpp
if (lhs_record->getIdentifier() != rhs_record->getIdentifier())
```
works both for named and for anonymous record types.

However, this doesn't hold for unnamed record types in the strict interpretation of the standard: they are always incompatible. GCC follows that, Clang does not. There is an opinion that Clang is more right than GCC, and the wording of the standard needs a change. But if it is decided that they indeed should be incompatible, IWYU will need some handling of this special case, and in that case care must be taken not to break the behavior for anonymous types, and corresponding lines of code would probably appear. OTOH, it is probably better to write tests in advance.

iwyu_ast_util.h
bolshakov-a: Yeah, it is:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-list

iwyu.cc
bolshakov-a: In fact, list-initialization here allows `CompatibilityChecker` constructor to be removed completely. But in that case, `private:` access specifier should be removed as well so as to transform `CompatibilityChecker` into an aggregate. It would hardly be a good design...

bolshakov-a: The helper class interface looks better for me because there is no need to pass dependencies (`provided_getter` and `ASTContext`) in every recursive call, and there are no out-parameters. (However, the latter could be partially achieved with function as well by returning `pair<bool, set<const NamedDecl*>>` or just `optional<set<const NamedDecl*>>`). OTOH, the class interface doesn't look good because there is a dependency between methods: `GetDeclsToReport()` should be called after `CouldBeCompatible`. Moreover, if `CouldBeCompatible` returns `false`, `decls_to_report_` is leaved in a strange state (but this may be said about the out-parameter too).

As I've said, I don't have good knowledge of basic OOP principles like when classes should be introduced, or what their interface should look like, because I have no professional programmer's education. So if you think the function is better, I'd prefer it.

kimgr: I also think the total end result is easier to read, so let's keep it.

> As I've said, I don't have good knowledge of basic OOP principles like when classes should be introduced, or what their interface should look like, because I have no professional programmer's education

In my experience, practically nobody does (self included) :-)

iwyu_ast_util.cc
kimgr: Ah, I see, thanks! And now I see that this is covered here too, nice: https://github.com/include-what-you-use/include-what-you-use/pull/1801/files#diff-0b7594fb27cc55e48e14a4074de7038928d884adb12df51712e4e4cf2cb95876R136

iwyu_ast_util.h
kimgr: I can accept it, but I don't have to like it, I guess :)

iwyu_ast_util.cc
kimgr: Thanks!

kimgr:
I'm ready to merge this if you want to take it out of review. Or are you waiting for some update on the Clang side?

iwyu_ast_util.h
bolshakov-a: I can replace it with parentheses, if you want. I agree that in many cases the parentheses are as fine as the curly braces.

bolshakov-a:
I think the member function implementations of `CompatibilityChecker` should be placed next to one another, and probably the commits should be squashed into a single one?

I don't expect any significant Clang changes in the near future.

iwyu_ast_util.h
kimgr: No, it's fine. It's good for me to exercise modern C++ a bit.

kimgr:
Ah, yes, missed both of those things. Please ping me when done.

bolshakov-a:
@kimgr, done.

kimgr:
Thank you!

Issue #1563, opened 2024-07-07
Wishlist for diagnostics


kimgr:
This is an umbrella issue for various ideas for diagnostics improvements that have nothing to do with IWYU's core analysis, but just want to tweak its diagnostic format or fidelity.

* Report explicitly that removal is caused by duplicate include: https://github.com/include-what-you-use/include-what-you-use/issues/1557#issuecomment-2172854324
* Control output verbosity: https://github.com/include-what-you-use/include-what-you-use/issues/190
* Empty output should be removed: https://github.com/include-what-you-use/include-what-you-use/issues/107
* Add --output option to export to file: https://github.com/include-what-you-use/include-what-you-use/pull/888

This is all good market research for a new design for IWYU diagnostics. I'll close these issues for now, and link new requests here, so we can weigh needs and wants if we ever look into a different diagnostic design.

kimgr:
* Add color output: https://github.com/include-what-you-use/include-what-you-use/issues/1680

Issue #949, opened 2021-09-03
Question: How do you silence "correct #includes/fwd-decls"


rianquinn:
Any idea on how to silence all of the "correct #includes/fwd-decls" statements so that I can just see what needs to be changed?

rianquinn:
I was able to hack my way through it a bit, but it would be great if there an option for this as there is simply too much output that is insane to sift through for large projects. What I did is the following:

Comment out:
```cpp
output = "\n(" + filename + " has correct #includes/fwd-decls)\n";
```

From there, you will notice that a lot of newlines are added by the Python script. To remove these, change the following:
```python
print(formatter(proc.get_output()))
```
to
```python
print(formatter(proc.get_output()), end = "")
```

In my script there was two of the above lines that I had to change. Again, it would be great if there was an option for this, but this should work for now.

rianquinn:
I am going to close this as there were a number of changes that I just made to our own fork that addressed a bunch of issues include:
- Cleaning up the output to be less verbose. 
- Returning a proper error code so that IWYU can be integrated into CI. 
- Stopping the python script on error to prevent endless output from compound errors. 

IWYU is AMAZING, and it should simply be integrated directly into the Clang Compiler IMO. If you are interested in my changes, they can be seen here:
https://github.com/Bareflank/include-what-you-use

And how we are using it can be seen here:
https://github.com/rianquinn/bsl/blob/master/cmake/find_programs.cmake#L62

The result looks like the following:
![image](https://user-images.githubusercontent.com/5932003/132101838-7fbeb3c8-d466-4e03-85f9-b4a1d22a60bc.png)

PR #1806, opened 2025-09-02
Add clang-warning-without-note format

HannaWAR:


iwyu_tool.py
```diff
@@ -53,16 +54,17 @@
 GENERAL, ADD, REMOVE, LIST = range(4)
 
 
-def clang_formatter(output, style):
+def clang_formatter(output, style, logLevel):
```
kimgr: Naming style is mixed across different python scripts, but newer ones like `iwyu_tool.py` use `hacker_case_with_underscores`.

iwyu_tool.py
```diff
@@ -53,16 +54,17 @@
...
     formatted = []
 
     state = (GENERAL, None)
     for line in output.splitlines():
         match = CORRECT_RE.match(line)
         if match:
-            formatted.append('%s:1:1: note: #includes/fwd-decls are correct' %
-                             match.groups(1))
+            if logLevel is LogLevel.NOTE:
+                formatted.append('%s:1:1: note: #includes/fwd-decls are correct' %
```
kimgr: Actually I'm beginning to wonder if this note is useful at all. I don't use these custom diagnostics myself, so I don't have a feeling for how they behave.

@svenpanne You added this back in 2016, do you have a feeling for how valuable the "everything's OK" diagnostics are? I'm considering if we should just remove them.

iwyu_tool.py
```diff
@@ -100,12 +102,16 @@ def clang_formatter(output, style):
...
+    WARNING = 2
 
 DEFAULT_FORMAT = 'iwyu'
 FORMATTERS = {
     'iwyu': lambda output: output,
-    'clang': lambda output: clang_formatter(output, style="error"),
-    'clang-warning': lambda output: clang_formatter(output, style="warning"),
+    'clang': lambda output: clang_formatter(output, style="error", logLevel=LogLevel.NOTE),
+    'clang-warning': lambda output: clang_formatter(output, style="warning", logLevel=LogLevel.NOTE),
+    'clang-warning-without-note': lambda output: clang_formatter(output, style="warning", logLevel=LogLevel.WARNING),
```
kimgr: I feel like it might be nice to avoid the extra enum and just fold the behavior into a new style. But I see how `style` is just used to say `error` or `warning` in the right place in diagnostics, so it will probably get messier.

iwyu_tool.py
svenpanne: Well, the underlying original diagnostic _"foobar has correct #includes/fwd-decls"_ is not really helpful, either, see e.g. #949. IIRC I just added the line above to keep a 1:1 correspondence. Personnally, I fully agree with Eric S. Raymond's "Rule of Silence": When a program has nothing surprising to say, it should say nothing.

So in a nutshell: Let's nuke the line mentioned above from both the wrapper and the underlying IWYU, at least in the default config. Adding an option to re-add it might make sense, but OTOH I can't really imagine a scenario where one actually needs that output. Usually e.g. C++ compilers don't tell you that they successfully compiled a file, either. 😉

kimgr: Yeah, that seems right to me too. @HannaWAR would you consider simplifying this patch by simply removing the note unconditionally?

kimgr: > Let's nuke the line mentioned above from both the wrapper and the underlying IWYU

Oh, I didn't see this. I think this should be added to #1563 if it's not already there, the core IWYU summary diagnostics are pretty terrible, but it takes some careful design to get them better, imho. Let's not do that right now.

AJIOB:
Looks like the @HannaWAR nothing change in C/C++ sources. @kimgr , should we merge this variant?

kimgr:
@AJIOB Yes, this looks good to me.

PR #1807, opened 2025-09-05
Do not suggest forward-decls for implicit code

### LINKED ISSUE #1774: IWYU suggests to fwd-declare types not explicitly written
For such an example:
```cpp
// fn.h
class Class;

template <typename>
class Tpl;

Class& GetClass();
Tpl<int>& GetTpl();

// source.cpp
#include "fn.h"

int main() {
    auto& class_ref = GetClass();
    auto& tpl_ref = GetTpl();
    [&class_ref, &tpl_ref]{}();
}
```
IWYU produces the output:
```
source.cpp should add these lines:
class Class;
template <typename> class Tpl;

source.cpp should remove these lines:

The full include-list for source.cpp:
#include "fn.h"  // for GetClass, GetTpl
class Class;
template <typename> class Tpl;
```
Forward-declaring `Class` and `Tpl` in `source.cpp` makes no sense because they are not explicitly written there.

The problem obviously arises because `VisitTagType` and `VisitTemplateSpecializationType` handle implicit AST nodes (namely, the declarations of the capture variables) as well as explicit ones.

bolshakov-a:
There is no need to change `VisitUsingType` because using-types inside implicit code are already skipped in `TraverseType` and `TraverseTypeLoc`.

Another way to fix it might be to move `ReportDeclForwardDeclareUse` calls into `Visit...TypeLoc` functions which are not called for implicit code (due to `IwyuAstConsumer::TraverseTypeLoc` logic). But this breaks fwd-declaring of types from dynamic exception specifications because they are `Type`s in the AST and not `TypeLoc`s (there is [a `TODO` inside `RecursiveASTVisitor`](https://github.com/llvm/llvm-project/blob/a549e73cad60336d8e9c0622ae7ad86aa65ef4ce/clang/include/clang/AST/RecursiveASTVisitor.h#L1417)).

This fixes #1774.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3102,7 +3102,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
     // If we are forward-declarable, so are our template arguments.
     if (CanForwardDeclareType(current_ast_node())) {
-      ReportDeclForwardDeclareUse(CurrentLoc(), decl);
+      if (!current_ast_node()->template GetAncestorAs<Decl>()->isImplicit())
+        ReportDeclForwardDeclareUse(CurrentLoc(), decl);
       current_ast_node()->set_in_forward_declare_context(true);
     } else {
       if (type->isTypeAlias())
@@ -5098,6 +5099,10 @@ class IwyuAstConsumer
     // needed: just forward-declare.
     if (CanForwardDeclareType(current_ast_node()) && !is_provided) {
       current_ast_node()->set_in_forward_declare_context(true);
+
+      if (current_ast_node()->GetAncestorAs<Decl>()->isImplicit())
+        return Base::VisitTagType(type);
+
       if (compiler()->getLangOpts().CPlusPlus) {
         // In C++, if we're already elaborated ('class Foo x') but not
         // a qualified name ('class ns::Foo x', 'class Class::Nested x') or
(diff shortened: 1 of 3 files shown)


iwyu.cc
```diff
@@ -3102,7 +3102,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
     // If we are forward-declarable, so are our template arguments.
     if (CanForwardDeclareType(current_ast_node())) {
-      ReportDeclForwardDeclareUse(CurrentLoc(), decl);
+      if (!current_ast_node()->template GetAncestorAs<Decl>()->isImplicit())
```
bolshakov-a: This construct is used now in 3 places. Maybe, it should be a function `InImplicitCode`?

kimgr: Sounds good!

tests/cxx/no_fwd_for_implicit-direct.h
```diff
@@ -1,4 +1,4 @@
-//===--- fn.h - iwyu test -------------------------------------------------===//
+//===--- no_fwd_for_implicit-direct.h - test input file for iwyu ----------===//
```
kimgr, line 1: Very cool -- it hadn't even occurred to me that we could just repurpose the bug tests as normal tests if they have no ~external~ library dependencies. Nice!

bolshakov-a: Bug tests look like a good idea, indeed!

iwyu_ast_util.h
```diff
@@ -468,6 +468,8 @@ bool IsAutocastExpr(const ASTNode*);
 // Returns the decl-context of the deepest decl in the ast-chain.
 const clang::DeclContext* GetDeclContext(const ASTNode* ast_node);
 
+bool InImplicitCode(const ASTNode*);
```
bolshakov-a, line 471: Not sure if a comment is needed. The name seems already self-explanatory to me.

PR #1808, opened 2025-09-06
Add test on spurious fwd-decl for var template arg

bolshakov-a:
The bug was noticed [here](https://github.com/include-what-you-use/include-what-you-use/pull/1793#discussion_r2280505322) and has been fixed in https://github.com/llvm/llvm-project/commit/1cb47c19f8eca4badd8fb5e1a1b1cf4aaab607b8.

### DIFF
--- a/tests/cxx/variable_template-i1.h
+++ b/tests/cxx/variable_template-i1.h
@@ -22,3 +22,18 @@ int both_args_used_def_provided = [] {
 using ProvidingRefAlias = IndirectClass&;
 
 int GetInt();
+
+template <typename>
+int var_tpl_in_header;
+
+template <typename T>
+void UseVarTplInHeader() {
+  // IWYU should not suggest a forward-declaration of DeclaredInCC here.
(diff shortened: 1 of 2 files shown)


bolshakov-a:
The test should pass when the LLVM packages are updated to some actual version.

firewave:
> The test should pass when the LLVM packages are updated to some actual version.

See https://github.com/llvm/llvm-project/issues/155585.

bolshakov-a:
Thanks, interesting. Btw, I intentionally pushed the commit early to make it clear that the test actually shows a bug.

reneleonhardt:
LLVM noble has been fixed / updated a few hours ago, can you run your CI again?

bolshakov-a:
@kimgr

reneleonhardt:
It would be nice if the `Capture LLVM major version and install prefix` step would actually print the found version before checking out, or even include it in the next step title so developers don't have to search the logs to find that the development branch has been used.

PR #1812, opened 2025-09-15
[clang compat] Remove unused using-declaration

bolshakov-a:
`DependentTemplateSpecializationType` was removed in https://github.com/llvm/llvm-project/commit/ba9d1c41c41d568a798e0a8c38a89d294647c28d. It was not referred to in IWYU except a using-declaration.


kimgr:
If only they were all this straightforward :slightly_smiling_face:

PR #1816, opened 2025-09-19
Turn on C++20 mode in type_trait test

bolshakov-a:
It is desirable for proper testing of the new `***_synthesises_from_spaceship` builtin traits because the three-way comparison ("spaceship") operator is a C++20 feature.

No functional change.


kimgr:
:+1:

PR #1818, opened 2025-09-21
wip: Put the mechanics into place for BTT_xxSynthesisesFromSpaceship

kimgr:
This almost certainly does not do/test the right things, but these should be the moving parts.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -116,6 +116,7 @@
 #include "clang/AST/TemplateName.h"
 #include "clang/AST/Type.h"
 #include "clang/AST/TypeLoc.h"
+#include "clang/AST/UnresolvedSet.h"
 #include "clang/Basic/Diagnostic.h"
 #include "clang/Basic/FileEntry.h"
 #include "clang/Basic/LangOptions.h"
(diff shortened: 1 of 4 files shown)


iwyu.cc
```diff
@@ -2698,6 +2699,23 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      case TypeTrait::BTT_LeSynthesisesFromSpaceship:
+      case TypeTrait::BTT_GtSynthesisesFromSpaceship:
+      case TypeTrait::BTT_GeSynthesisesFromSpaceship: {
+        require_complete_arg_types();
+
+        UnresolvedSet<16> spaceship_decls;
+        Sema& sema = compiler()->getSema();
+        sema.LookupBinOp(sema.TUScope, CurrentLoc(), clang::BO_Cmp,
+                         spaceship_decls);
+        for (const NamedDecl* decl : spaceship_decls) {
```
bolshakov-a, line 2712: This is too broad. Such a lookup doesn't even check the actual argument types, so even unrelated operators should be reported here. I prefer just copy-paste the logic from SemaTypeTraits.cpp.

iwyu.cc
```diff
@@ -2698,6 +2699,23 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
           ReportDeclsUse(CurrentLoc(), chkr.GetDeclsToReport());
         return true;
       }
+      case TypeTrait::BTT_LtSynthesisesFromSpaceship:
+      case TypeTrait::BTT_LeSynthesisesFromSpaceship:
+      case TypeTrait::BTT_GtSynthesisesFromSpaceship:
+      case TypeTrait::BTT_GeSynthesisesFromSpaceship: {
+        require_complete_arg_types();
```
bolshakov-a, line 2706: There are cases when it is guaranteed that the `operator<=>` is absent: when one of the args is `void`, or when both of them are arrays. There is no need for the complete type info in such cases.

kimgr:
Superseded by #1819. Closing.

PR #1819, opened 2025-09-21
Handle new *synthesises_from_spaceship type traits

### LINKED ISSUE #1814: Handle new `__builtin_xx_synthesises_from_spaceship` traits
https://github.com/llvm/llvm-project/commit/b1a8089f1cf67a6038a6b6d958eff1bd2feb4f8d introduced a set of new traits, and IWYU's build is now warning that they're unhandled:

```
/home/runner/work/include-what-you-use/include-what-you-use/iwyu.cc:2090:13: warning: 4 enumeration values not handled in switch: 'BTT_LtSynthesisesFromSpaceship', 'BTT_LeSynthesisesFromSpaceship', 'BTT_GtSynthesisesFromSpaceship'... [-Wswitch]
 2090 |     switch (expr->getTrait()) {
```
https://github.com/include-what-you-use/include-what-you-use/actions/runs/17775333585/job/50521474222

Judging from the Clang tests, it looks like they require the full type (to be able to know if there's a spaceship operator):
https://github.com/llvm/llvm-project/blob/b1a8089f1cf67a6038a6b6d958eff1bd2feb4f8d/clang/test/SemaCXX/type-trait-synthesises-from-spaceship.cpp

@bolshakov-a Does that make sense? I can take a stab at adding support if it sounds right.

bolshakov-a:
The operator lookup logic has mostly been copy-pasted from `EvaluateBinaryTypeTrait` function in `clang/lib/Sema/SemaTypeTraits.cpp`.

This fixes #1814.


bolshakov-a:
I've found a case where both of the argument types _must_ be complete even when the overloaded operator is a non-member function. Maybe, a test case on unions and/or pointers is worth writing as well?

kimgr:
I see this does more or less what Clang Sema does for the traits at hand, so LGTM.

I'm still a bit confused about Clang's implementation -- the trait names seem to indicate that plain `<`, `<=`, `>=`, `>` operators are synthesized _from_ `<=>`, whereas language text only talks about synthesis in the other direction: https://en.cppreference.com/w/cpp/language/default_comparisons.html#Synthesized_three-way_comparison

It looks like in Clang the synthesis happens when do an actual lookup; an `operator<=>` is dynamically generated by `BuildSynthesizedThreeWayComparison`. So I guess the trait emulates that and sees if it can generate an `operator<=>` from whatever plain operator it's able to look up?

I wonder if it's just the `From` in `BTT_xxSynthesisesFromSpaceship` that is confusing me. To me it seems like the semantics are `BTT_xxSynthesisesSpaceship`.

bolshakov-a:
There is some confusion in terms. The synthesized comparison you are talking about is defined in such case:
```cpp
#include <compare>

struct Inner {
    bool operator<(const Inner&) const;
    bool operator==(const Inner&) const;
};

struct Outer {
    std::strong_ordering operator<=>(const Outer&) const = default;

    Inner m_inner;
};
```
To provide the defaulted `operator<=>` for `Outer`, a three-way comparison in synthesized for `Inner` (which may be some nonmodifiable legacy class). This behavior was proposed in [P1186](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1186r3.html).

The new traits are not about it, they are about if the _overload resolution_ for e.g. `a < b` finds a _rewritten candidate_ (e.g. `a <=> b < 0`. The corresponding `operator<=>` should be already present for that, it is not synthesized, what is synthesized (not the correct term in this case) here is `operator<`.

kimgr:
Thank you, that makes things a lot clearer!

> The new traits are not about it, they are about if the overload resolution for e.g. a < b finds a rewritten candidate (e.g. a <=> b < 0.

And I guess that's maintained as an `Expr`, to be able to group the call and the comparison, and so that's why we go digging for `CXXOperatorCallExpr`s to find the relevant `operator<=>`?

> The corresponding operator<=> should be already present for that, it is not synthesized, what is synthesized (not the correct term in this case) here is operator<.

I guess it's correct in some sense, the indirection to dynamically implement `a < b` in terms of `(a <=> b) < 0` seems like synthesis. It's just unfortunate that the wording conflicts.

Anyway, thank you for this patch! Let's get rid of the warnings on master.

bolshakov-a:
> And I guess that's maintained as an Expr, to be able to group the call and the comparison, and so that's why we go digging for CXXOperatorCallExprs to find the relevant operator<=>?

`CXXRewrittenBinaryOperator` (which is also `Expr`) groups the operator call and the comparison. The first `dyn_cast<CXXOperatorCallExpr>` branch intercepts cases when non-rewritten relational operators are selected by overload resolution (and hence they are to be reported and not any `operator<=>`).

bolshakov-a:
> and not any operator<=>

But this is questionable, because if e.g. a distinct `operator>` is removed after some user's code change, but there is an `operator<=>` in some other file... Anyway, these cases are too unlikely, and I think all the type trait handling is of little real-world importance, if of any at all.

PR #1823, opened 2025-09-24
Report fn decl specifying missing default argument

bolshakov-a:
If you write
```cpp
void Fn(int = 0, int);
```
there should be exactly one `Fn` declaration prior to this declaration that specifies the default argument value for the second parameter. If it is in another header, report that header for inclusion.

This change affects only the function declaration side, not the function use side.

`UF_DefinitionUse` has been renamed because it is used now not only at definitions but also at function declarations.


bolshakov-a:
I've transformed `def_arg_specified` into a global function because I'm going to use it outside of `VisitFunctionDecl` too.

iwyu_use_flags.h
```diff
@@ -17,7 +17,7 @@ typedef unsigned UseFlags;
 
 const UseFlags UF_None = 0;
 const UseFlags UF_InCxxMethodBody = 1;       // use is inside a C++ method body
-const UseFlags UF_DefinitionUse = 2;         // use is itself a definition
+const UseFlags UF_Exact = 2;                 // don't remap to another decl
```
kimgr: Hehe, I know I say this every time anybody touches these flags, which indicates that maybe I'm doing it backwards :slightly_smiling_face::

The flags are intended to say something about the context in which the use occurs, not about what you want to happen in use reporting.

I'm not sure it's possible to generalize both definition-is-use-of-earlier-declaration and decl-has-a-prior-decl-with-default-values-for-later-params into the same concept. Might be better to add a new flag and check both.

bolshakov-a: I think that having two flags doing exactly the same would be a useless code bloating. But I agree that mixing the principles of naming doesn't look good. So I suggest to rename it to `UF_DeclarationUse` because definitions are a subset of declarations.

Maybe there is a broader contradiction between your view that the analysis should better be done in `iwyu_output.cc` (hence some information about the use context needs to be carried from AST visitors to `iwyu_output.cc`) and my preference to do it in `iwyu.cc` where the required information is present.

kimgr: `UF_DeclarationUse` or maybe `UF_RedeclUse` sounds good to me.

The flags were originally a single boolean `in_cxx_method_body`, which was used to forward contextual knowledge from `iwyu.cc` to `iwyu_output.cc` (see aa2bc1f9e0b664065b5e1cbe45637fbd0f9e327e). So I think we only use them where there's reasoning that spans both.

But yeah, there's quite a bit of analysis going on in iwyu.cc (particularly for author-intent), so policy is not only in output.

iwyu_ast_util.h
```diff
@@ -741,6 +741,10 @@ bool IsImplicitlyInstantiatedDfn(const clang::FunctionDecl*);
 // argument.
 const clang::CXXMethodDecl* GetFromLeastDerived(const clang::CXXMethodDecl*);
 
+// Returns true if the given function declaration explicitly specifies a default
+// argument value for the i-th parameter.
+bool DefArgSpecified(unsigned i, const clang::FunctionDecl*);
```
kimgr: Maybe `IsDefArgSpecified`? Or will that make caller lines too long?

iwyu_use_flags.h
bolshakov-a: Renamed to `UF_RedeclUse`.

iwyu.cc
```diff
@@ -4974,6 +4974,28 @@ class IwyuAstConsumer
...
+    // values for the last parameters.
+    for (unsigned i = 1, ie = decl->getNumParams(); i < ie; ++i) {
+      if (IsDefArgSpecified(i - 1, decl) && !IsDefArgSpecified(i, decl)) {
+        // There should be a fn redecl specifying the i-th param default value.
+        for (const FunctionDecl* redecl : decl->redecls()) {
+          if (IsDefArgSpecified(i, redecl)) {
+            ReportDeclUse(CurrentLoc(), redecl, nullptr, UF_RedeclUse);
+            return Base::VisitFunctionDecl(decl);
+          }
+        }
```
kimgr, line 4993: Could you break after here? Or is it possible to see something like:
```
void foo(int, int, int, int);
void foo(int, int=0, int, int=0);
void foo(int, int, int=0, int);
void foo(int=0, int, int, int)
```
?

I'm reading the semantics as each declaration can default params from the end of the list, but earlier declarations may have "shifted" the end by defaulting later params. So I'm guessing what I suggested above isn't allowed, but this would be:
```
void foo(int, int, int, int);
void foo(int, int, int, int=0);
void foo(int, int, int=0, int); // effectively foo(int, int, int=0, int=0)
void foo(int=0, int=0, int, int)   // effectively foo(int=0, int=0, int=0, int=0)
```

bolshakov-a: > So I'm guessing what I suggested above isn't allowed

Yeah, the second decl doesn't compile.

There is a `return` above. As the comment suggests, it should be reached, so I'd better place here `CHECK_UNREACHABLE_` instead of `break;`.

kimgr: Oh, I didn't see the return.

So the outer loop finds the first defaulted argument, and then searches for the first redecl where the argument after is defaulted. Will that work with multi-default-arg examples like `foo` above? I would have expected the algorithm to be more like:
```
for (int i = 0; i < num_params; ++i) {
  if (!IsDefaultedArg(i, decl)) {
    for (const FunctionDecl* redecl : decl->redecls()) {
       if (IsDefaultedArg(i, redecl) {
          // report a decl use
       }
    }
  }
}
```

This doesn't care about the precise semantics so much, it just says that if an earlier decl provides a default value for a param where we do not, we want to keep that redecl around.

It seems a bit simpler, but also more correct...?

bolshakov-a: > Will that work with multi-default-arg examples like foo above?

Yeah, it works:
```cpp
//  direct.h
#include "indirect1.h"
#include "indirect2.h"
#include "indirect3.h"

//  indirect1.h
void foo(int, int, int, int);

//  indirect2.h
void foo(int, int, int, int=0);

//  indirect3.h
void foo(int, int, int=0, int);

//  last_decl.cpp
#include "direct.h"

void foo(int=0, int=0, int, int);
```
```
$ include-what-you-use last_decl.cpp -Xiwyu --check_also=indirect1.h -Xiwyu --check_also=indirect2.h -Xiwyu --check_also=indirect3.h

(indirect1.h has correct #includes/fwd-decls)

(indirect2.h has correct #includes/fwd-decls)

indirect3.h should add these lines:
#include "indirect2.h"  // for foo

indirect3.h should remove these lines:

The full include-list for indirect3.h:
#include "indirect2.h"  // for foo
---

last_decl.cpp should add these lines:
#include "indirect3.h"  // for foo

last_decl.cpp should remove these lines:

The full include-list for last_decl.cpp:
#include "indirect3.h"  // for foo
---
```
Note that `indirect3.h` should `#include "indirect2.h"` to be compilable by itself. So, there is no need to `#include "indirect2.h"` in `last_decl.cpp` Otherwise, the `#include` order in `last_decl.cpp` becomes important.

kimgr: Very nice! Alright, in that case LGTM!

bolshakov-a: Maybe, you suggest the following change which is NFC in fact:
```diff
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -4983,12 +4983,11 @@ class IwyuAstConsumer
     // there should be (a) prior declaration(s) that specifies default argument
     // values for the last parameters.
     for (unsigned i = 1, ie = decl->getNumParams(); i < ie; ++i) {
-      if (IsDefArgSpecified(i - 1, decl) && !IsDefArgSpecified(i, decl)) {
+      if (IsDefArgSpecified(i - 1, decl)) {
         // There should be a fn redecl specifying the i-th param default value.
         for (const FunctionDecl* redecl : decl->redecls()) {
           if (IsDefArgSpecified(i, redecl)) {
             ReportDeclUse(CurrentLoc(), redecl, nullptr, UF_RedeclUse);
-            return Base::VisitFunctionDecl(decl);
           }
         }
       }
```
In this case IWYU just reports the current decl for itself if it specifies `i`-th argument as well as `i-1`-th.

kimgr: Oops, saw this after merge. Yeah, that seems harder to understand and worse :-)

PR #1824, opened 2025-09-25
Allow forward-declarations in __gnu_cxx namespace

bolshakov-a:
The standard explicitly forbids only declarations in the std namespace in user's code (in the general case).

The reasoning not to expose system header internals seems to me to be wrong. If a user doesn't write a type name explicitly, IWYU should not suggest to forward-declare it anyway (it would be a sign of a bug). And if he does, it is probably normal to require a forward-declaration.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -998,11 +998,8 @@ set<string> CalculateMinimalIncludes(
 // A1) If not a class or a templated class, recategorize as a full use.
 // A2) If a templated class with default template params, recategorize
 //     as a full use (forward-declaring in that case is too error-prone).
-// A3) If a symbol in std, __gnu_cxx, or another system namespace,
-//     recategorize as a full use.  This is entirely a policy
-//     decision: we've decided never to forward-declare anything in
-//     a system namespace, because it's best not to expose the internals
-//     of system headers in user code, if possible.
+// A3) If a symbol in std, recategorize as a full use. The C++ standard forbids
+//     declarations in std namespace in user's code, in general.
 // A4) If the file containing the use has a pragma inhibiting the forward
 //     declaration of the symbol, change the use to a full info use in order
 //     to make sure that the compiler can see some declaration of the symbol.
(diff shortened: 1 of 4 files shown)


bolshakov-a:
In fact, I'm still [not very confident in this change](https://github.com/include-what-you-use/include-what-you-use/issues/1763#issuecomment-3264007135). However, it looks like the libstdc++ authors [have nothing against declaring things in `__gnu_cxx` namespace in user's code](https://gcc.gnu.org/onlinedocs/gcc-15.2.0/libstdc++/api/a04133.html):
> The reason this class exists is so users can specialize it. Classes in namespace std may not be specialized for fundamental types, but classes in namespace __gnu_cxx may be.

Class template explicit specializations had to be declared in the namespace of the primary template (`__gnu_cxx` in this case) prior to C++11. And the comment was added back in 2003. So, users had to add their own `__gnu_cxx` namespace definition to specialize the template. Hence, defining a `__gnu_cxx` namespace to add some forward-declarations is probably Ok as well... OTOH, C++98 and C++03 standards differentiate between original and extension namespace definitions... Probably, a question should be asked on StackOverflow...

iwyu_output.cc
```diff
@@ -1117,15 +1114,13 @@ void ProcessForwardDeclare(OneUse* use,
...
+  // (A3) If it is in namespace std, recategorize as a full use.
   // TODO(csilvers): if someone has specialized a class in std, the
   // specialization should be treated as in user-space and
   // forward-declarable.  Check for that case.
-  if (StartsWith(use->symbol_name(), "std::") ||
-      StartsWith(use->symbol_name(), "__gnu_cxx::")) {
+  if (StartsWith(use->symbol_name(), "std::")) {
     VERRS(6) << "Moving " << use->symbol_name()
-             << " from fwd-decl use to full use: in a system namespace "
+             << " from fwd-decl use to full use: in the namespace std"
```
kimgr: I'd leave out "the" in this message `in namespace std` reads very normal for a compiler diagnostic.

tests/cxx/system_namespaces.cc
```diff
@@ -10,7 +10,7 @@
 // IWYU_ARGS: -I .
 
 // Tests that we correctly replace forward declares with #includes for
-// items in a system namespace: std or __gnu_cxx or the like.
+// items in the namespace std.
```
kimgr: Similarly here, I think "the" can be removed.

tests/cxx/system_namespaces.cc
```diff
@@ -23,22 +23,23 @@ __gnu_cxx::SystemClass* bar = 0;
 typedef int __system_type;
 notsys_ns::TplClass<__system_type>* baz = 0;
 
-
-// The summary is where all the action is on this test.  Only TplClass
-// should be replaceable by a forward-declare, because only it is
-// exposing a symbol in a non-system namespace.
+// The summary is where all the action is on this test. Only std::StdClass
+// should be not replaceable by a forward-declare, because only it is
+// exposing a symbol in the std namespace.
```
kimgr, line 28: ... but just to counter: "the std namespace" sounds better, so make sure we keep this one.

kimgr:
Yeah, libstdc++ encouraging user declarations in `__gnucxx` seems questionable. But if they explicitly encourage specializations, I guess it's not up to us to judge. I'm fine with merging this.

PR #1827, opened 2025-10-02
Select fn redecl with def arg used at call site

bolshakov-a:
The duplicating reporting of function declaration from `VisitDeclRefExpr` has been avoided because `DeclRefExpr` may point not to the correct redeclaration. However, if the function name is introduced by a using- declaration, that using-decl is still to be reported.

The change in `IsDefArgSpecified` is to avoid an assertion failure on a templated constructor of `std::set` reported from `HandleAliasedClassMethods`. In principle, the default argument handling is not needed for class methods.


iwyu_output.cc
```diff
@@ -169,6 +169,17 @@ std::string PrintableUnderlyingType(const EnumDecl* enum_decl) {
   return std::string();
 }
 
+static bool FnDeclHasDefArg(const NamedDecl* decl) {
+  const auto* fn_decl = decl->getAsFunction();
```
kimgr: Let's spell out `FunctionDecl` here. I think `auto` is OK if the type name is spelled out on the right side, but not otherwise.

iwyu_ast_util.cc
```diff
@@ -1381,6 +1397,16 @@ bool IsDefArgSpecified(unsigned i, const FunctionDecl* func) {
                                               func->getEndLoc());
 }
 
+FunctionDecl* GetRedeclSpecifyingDefArg(unsigned i, FunctionDecl* func) {
+  if (i < func->getNumParams()) {
```
kimgr: I don't really like how you can pass any unsigned value, and the function will silently fall through and return `func`.

I'd rather handle this at the caller side, I'll see if I can suggest something more readable there too.

iwyu.cc
```diff
@@ -801,6 +812,18 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
     const Type* parent_type = expr->getAllocatedType().getTypePtrOrNull();
     // 'new' calls operator new in addition to the ctor of the new-ed type.
     if (FunctionDecl* operator_new = expr->getOperatorNew()) {
+      // If the placement-new call makes use of default arguments, select
```
kimgr, line 814: I find these for loops really hard to read. I suspect a less compact version would be more straighforward:
```
for (unsigned i = 0; i < expr->getNumPlacementArgs(); ++i) {
  if (isa<CXXDefaultArgExpr>(expr->getPlacementArg(i)) {
    CHECK_((i + 1) < expr->getNumPlacementArgs());
    operator_new = GetRedeclSpecifyingDefArg(i + 1, operator_new);
    break;
  }
}
```

That would also allow you to get rid of the "allow-out-of-bounds" thing in `GetRedeclSpecifying...`

iwyu.cc
```diff
@@ -714,9 +714,20 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
       return false;
     if (CanIgnoreCurrentASTNode())
       return true;
-    return this->getDerived().HandleFunctionCall(expr->getDirectCallee(),
-                                                 TypeOfParentIfMethod(expr),
-                                                 expr);
+    FunctionDecl* func = expr->getDirectCallee();
+    if (!func)
+      return true;
+    // If the call makes use of default arguments, select exactly that
```
kimgr, line 720: Same as for the new-expr, should be able to make this a simpler loop

iwyu.cc
kimgr: Or maybe the check needs to be a condition instead, it's not fully clear to me why we take the next one after the defaultargexpr. Is it because the placement args are counted separately?

I guess the argument expressions for new-expr are not available, so this can be the same as in `TraverseCallExpr`.

iwyu_output.cc
bolshakov-a: I wanted to replace with `dyn_cast<FunctionDecl>` and forgot. Handling function templates is too questionable.

iwyu.cc
bolshakov-a: If you declare an allocation function like this:
```cpp
void* operator new(std::size_t size, int a, int b);
```
then for a call like `new(1, 2) int` the first parameter (`size`) is initialized with the result of `sizeof(int)`, the second parameter (`a`) is initialized with the first argument expression (`1`), and the third parameter (`b`) is initialized with the second argument (`2`): https://godbolt.org/z/aeGc6eME1

I left a comment below, but maybe it is not clear enough. Also, there is a couple of test cases.

iwyu.cc
bolshakov-a: It looks now really better, thanks!

iwyu.cc
kimgr: Looks good to me!

kimgr:
Looks like the new tests are passing, so we're only blocked by the `__builtin_is_implicit_lifetime` issue in Clang. Would you rather merge this now (if you have other work queued up), or are you OK to wait until the mainline build is green?

bolshakov-a:
I have no strong opinion here. Merging to a red master branch is not good in general. However, I have run the IWYU dogfood locally and don't see any problems except `#include <tuple>` suggestion in `iwyu.cc`, but this is present on the master branch as well.

bolshakov-a:
> ... except `#include <tuple>` suggestion in `iwyu.cc` ...

Moreover, this appears to be my local problem, maybe due to some other libstdc++ version.

bolshakov-a:
It would be good to add standard library version number into the log to make investigation of such issues easier.

kimgr:
Ah, the dogfood and bugs suite didn't run. Yeah, then let's wait.

kimgr:
Ooh, nice. I ran the bugs suite locally on this branch, and:
```
$ ./run_iwyu_tests.py --extra-suite=bugs bugs
...
UNEXPECTED SUCCESS: test_1763 (__main__.bugs.test_1763)
UNEXPECTED SUCCESS: test_1763b (__main__.bugs.test_1763b)
Ran 67 tests in 26.613s

FAILED (expected failures=65, unexpected successes=2)
```

So it seems like it fixes #1763.

Incidentally, my `__type_pack_element` branch in #1834 also gets unexpected success for 1763 and 1763b. Unfortunately neither of the branches fix 1763c, which seems to actually deal with smeared function default args...?

bolshakov-a:
I don't see any tests on #1763 in the repo. Have you committed and pushed them onto the server? So I cannot say anything but that having them succeeded simultaneously on two unrelated branches looks like due to your local environment, probably different standard library version. Have you tested them on the master branch?

kimgr:
:facepalm: -- ok, they're unversioned in my working tree. I was probably working on it and got interrupted mid-way. Sorry about the confusion.

kimgr:
Please rebase on latest master so we can see the whole CI end-to-end. Thanks!

bolshakov-a:
I think you could just re-run the build. I saw in the log that it checks out not the branch but the result of merging it with master.

bolshakov-a:
Anyway, rebased.

kimgr:
Thanks!

kimgr:
> I think you could just re-run the build. I saw in the log that it checks out not the branch but the result of merging it with master.

For reference, I tried this on another PR just now, and rerunning the build jobs doesn't appear to run the PR on top of the latest master:

* 14:29: First build: https://github.com/include-what-you-use/include-what-you-use/actions/runs/18803081331/attempts/1?pr=1841
* 14:31: Merge other PR to make master green: https://github.com/include-what-you-use/include-what-you-use/pull/1840
* 14:39: Second attempt: https://github.com/include-what-you-use/include-what-you-use/actions/runs/18803081331

I think we changed the github checkout action at some point to have it build on latest master, but it doesn't seem to have helped.

I'm fine with the current rebase workflow, so I don't really care, just thought I'd share what I saw.

bolshakov-a:
Yeah, it still merged afd879c353e for some reason:
https://github.com/include-what-you-use/include-what-you-use/actions/runs/18803081331/job/53653387732#step:4:86

PR #1828, opened 2025-10-04
[clang compat] Agree with Clang on spelling of synthesize

kimgr:
https://github.com/llvm/llvm-project/commit/b9e41ae6f9efb068e1fe137357c3d8ded31cdd02 changed the spelling of the builtins from "synthesises" to "synthesizes", and this affects both enumerators and the exposed builtin names.

Update to agree with new spelling.


kimgr:
Tests still fail because `__builtin_is_implicit_lifetime` changed semantics. But the test suite succeeds if I remove those static assertions from `type_trait.cc`, so I'll merge this unrelated change with a red build so we can address the lifetime intrinsic separately.

PR #1832, opened 2025-10-14
[ci] Print LLVM config to build log

kimgr:
Print the values of LLVM_MAJOR and LLVM_PREFIX after they are calculated, to make it easier to identify which version was used.


kimgr:
@reneleonhardt Was this what you had in mind for https://github.com/include-what-you-use/include-what-you-use/pull/1808#issuecomment-3288143694?

reneleonhardt:
Looks much better, thank you!

What I meant was that you could also use env variables inside step names, so no one would have to drill down to a step log to see the version 😉

```yaml
- name: Build include-what-you-use with LLVM ${{ env.LLVM_MAJOR }}
```

kimgr:
@reneleonhardt Nice! Done.

Issue #1617, opened 2024-09-13
Forward declare suggested when include needed when using structured binding on std::tuple with libc++


bungeman:
A minimal reproduction in [iwyutest.zip](https://github.com/user-attachments/files/16999467/iwyutest.zip).

1. With
llvm [llvmorg-19.1.0-rc4](https://github.com/llvm/llvm-project/tree/llvmorg-19.1.0-rc4)
iwyu 5ac52caf1529630e6aa7932e0ea9382fca206fb4

`item.h`:
```
#ifndef item_DEFINED
#define item_DEFINED

class Item {
public:
    void process();
};

#endif
```

`process.cpp`:
```
#include "item.h"
#include <tuple>

void process(std::tuple<Item*> foo) {
    auto [item] = foo;
    item->process();
}
```

2. iwyu on process.cpp
`~/src/clang_output/bin/include-what-you-use -stdlib=libc++ -c process.cpp -o process.o -Xiwyu --comment_style=long -Xiwyu --quoted_includes_first -Xiwyu --verbose=3`
```
process.cpp should add these lines:
class Item;

process.cpp should remove these lines:
- #include "item.h"  // lines 1-1

The full include-list for process.cpp:
#include <tuple>  // for std::tuple
class Item;
---
```
However, changing `Item` to a forward declaration here is incorrect as this
will not compile. Building with the co-built clang with `-MD` shows this using
`~/src/clang_output/bin/../include/c++/v1/` for `tuple`. Dropping the
`-stdlib=libc++` from the above command, one gets

```
(process.cpp has correct #includes/fwd-decls)
```

and building with the co-built clang with `-MD` shows this using the local
`include/c++/13/` for `tuple`. It isn't clear yet if this is because there is difficulty
with the use of the type through the structured binding or because libc++ tuple
does something interesting.

kimgr:
I tried the following:
```
$ include-what-you-use -c tests/1617/process.cpp -Xiwyu --comment_style=long -Xiwyu --quoted_includes_first -Xiwyu --verbose=7 2>&1 | ./scrub-logs.py > 1617-libstdcxx.txt
$ include-what-you-use -stdlib=libc++ -c tests/1617/process.cpp -Xiwyu --comment_style=long -Xiwyu --quoted_includes_first -Xiwyu --verbose=7 2>&1 | ./scrub-logs.py > 1617-libcxx.txt
```
and then diffed them. There's a bunch of noise, but this portion looks relevant:
[...]

PR #1834, opened 2025-10-18
Handle builtin template arguments better

kimgr:
The std::tuple implementation in libc++ (and lately in libstdc++ as
well) relies on a compiler builtin to extract the nth arg from a
template parameter pack: __type_pack_element.

In Clang, __type_pack_element is fully represented as a template and
falls under the normal template processing in IWYU. However, its
specializations also flag as a being substituted from a type alias.

We used to assume that means there is a TypeAliasTemplateDecl associated
with the type, and did some extra type provision lookups for the alias
type itself.

Now also handle the other case, and report types for non-alias
templates. As far as I know, this only applies to
BuiltinTemplateDecl (like __type_pack_element), but could potentially
pick up reporting for other alias template kinds.

This should also improve reporting for __make_integer_seq similarly.

Remove now-fixed bug test.

Fixes https://github.com/include-what-you-use/include-what-you-use/issues/1617


kimgr:
@bolshakov-a This appears to do the right thing, but do you think there's a need to expand provided types for further blocking? I'm not sure what that would look like or what might create a need... Something like:
```
typedef Foo FooT;
__type_pack_element(0, FooT) foo;  // should Foo be provided/blocked?
```

kimgr:
This seems to work as expected:
```
// type.h
struct UserType {
  void method();
};

// def.h
#include "type.h"

typedef UserType MyType;

// t3.cc
#include "def.h"

void bar() {
  auto x = __type_pack_element<0, MyType*>{};
  x->method();
}
```
And:
```
$ include-what-you-use t3.cc

(t3.cc has correct #includes/fwd-decls)
```

bolshakov-a:
I have performed some debugging and see that the sugar is preserved well so, yeah, I don't think any additional provided type collecting is needed here. `GetAliasTemplateProvidedTypes` call in the first branch collects types provided by a type alias declaration itself, not by some template arguments at the use site. And builtin templates don't provide anything.

One problem I have noticed is that IWYU requires `IndirectClass` in cases like this:
```cpp
#include "tests/cxx/direct.h"

__type_pack_element<0, IndirectClass>* pic;
```
It is reported due to
```cpp
if (is_provided || type->isTypeAlias())
  ReportDeclUse(CurrentLoc(), TypeToDeclAsWritten(type), comment);
```
inside `IwyuAstConsumer::VisitTemplateSpecializationType`. The intent was to report alias template declarations, like typedef declarations, not their underlying type. But `TypeToDeclAsWritten` returns underlying type declaration in this case. The fix may be to add `BuiltinTemplateDecl` handling inside `TypeToDeclImpl`:
```diff
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1544,7 +1544,7 @@ static const NamedDecl* TypeToDeclImpl(const Type* type, bool as_written) {
     // useful at all?
     return icn_type->getOriginalDecl()->getDefinitionOrSelf();
   } else if (as_written && template_decl &&
-             isa<TypeAliasTemplateDecl>(template_decl)) {
+             (isa<TypeAliasTemplateDecl>(template_decl) || isa<clang::BuiltinTemplateDecl>(template_decl))) {
     // A template type alias
     return template_decl;
   } else if (const RecordType* record_type = type->getAs<RecordType>()) {
```
or just to short-circuit `BuiltinTemplateDecl` case in the calling code. Maybe, all the places where `isTypeAlias()` is called need reviewing.

kimgr:
Pushed a testcase loosely based on `alias_template.cc` because it had a lot of nice variations. I've marked up the assertions I don't fully understand with `TODO`, let me know what you think.

kimgr:
Ah, applied your patch, and all but one TODO disappeared.

tests/cxx/type_pack_elem.cc
```diff
@@ -0,0 +1,96 @@
...
+    // IWYU: IndirectClass is...*indirect.h
+    alias;
+
+// Template nesting
+template <class T>
+struct Tpl {
+  T t;
+};
+
+// TODO: not sure about the full-use on __type_pack_element here.
```
bolshakov-a: Why do you doubt? It actually is a full use like in `Tpl<Tpl<Tpl<IndirectClass>>> nested;` declaration because a class template implicit instantiation requires its field types to be complete, i.e. instantiates them recursively until `IndirectClass`: https://godbolt.org/z/cPP3W4fT8

kimgr: I started to wonder, because it makes sense to me that the `nested` on L69 creates a complete-type dependency on `IndirectClass`. But I wasn't sure about the instantiation expression `Tpl<Tpl<Tpl<IndirectClass>>>`, I would have expected `IndirectClass` to be forward-declarable in this context.

As you can see in godbolt, the diagnostic is on the `nested` variable, not the `__type_pack_element` expression.

kimgr: Maybe a more convincing argument is https://godbolt.org/z/zKx4b3rso -- traits checks can be done without `IndirectClass` being defined.

bolshakov-a: The first warning is from `TraverseTypeLoc` (in fact, it is reported even three (!) times, two times from `InstantiatedTemplateVisitor` and one more time from `IwyuAstConsumer`; not sure if such a redundancy can cause any problems) because `CanForwardDeclareType` returns `false`, and hence `__type_pack_element<...>` type sugar is reported, which causes the warning on its underlying type.

The second warning is from `VisitCXXConstructExpr`.

This is just like
```cpp
IndirectClass
ic;
```
causes warnings on the both lines.

In your last example with `__is_same`, the type (I mean the whole `__type_pack_element<...>` construct) is in the fwd-decl context and hence is not reported.

kimgr: Cool, thanks. So I'm starting to think that these are behaviors that are already a bit fuzzy in IWYU, and fixing said fuzziness should also fix it for `__type_pack_element`.

I think it's fair to say, then, that `Tpl<IndirectClass> t0` and `__type_pack_element<0, Tpl<IndirectClass>> t1` should behave the same in this example:
```c++
Tpl<IndirectClass>
    t0;

__type_pack_element<0, Tpl<IndirectClass>>
    t1;
```

which they do on the current branch.

Nesting doesn't really complicate things more than that, as far as I can see.

bolshakov-a: Yeah, correct.

There are cases when a variable type is explicitly written and should be reported, but the constructor is not called (like at class field declarations). And there are cases when a variable type is not written but nonetheless should be reported (like with `auto`); it is reported currently from `VisitCXXConstructExpr`.

kimgr: Thanks for the details, that really helps!

kimgr:
@bolshakov-a Cleaned up the patch, please review when you have a chance.

iwyu.cc
```diff
@@ -3519,12 +3519,17 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
           InsertAllInto(
               GetAliasTemplateProvidedTypes(template_spec_type, al_tpl_decl),
               &blocked_types);
-          const Type* type = template_spec_type->getAliasedType().getTypePtr();
+          ReportTypeUseInternal(used_loc, type, blocked_types, deref_kind);
+        } else {
+          // Builtin templates like __type_pack_element<0, Class*> are marked as
+          // type alias substitutions, but have no associated alias template
+          // decl.
           ReportTypeUseInternal(used_loc, type, blocked_types, deref_kind);
```
bolshakov-a: `ReportTypeUseInternal` may be moved outside the if-statement, thus leaving inside only provided types harvesting.

tests/cxx/type_pack_elem.cc
```diff
@@ -0,0 +1,97 @@
...
+template <class T>
+struct Tpl {
+  T t;
+};
+
+// IWYU: IndirectClass is...*indirect.h
+// IWYU: IndirectClass needs a declaration
+Tpl<IndirectClass>
+    // IWYU: IndirectClass is...*indirect.h
+    t0;
```
bolshakov-a, line 69: Have you intentionally placed here a case without `__type_pack_element` or just forget to remove after experimentation?

kimgr: Good point. I did put it there to convince myself and others that it's symmetrical with __type_pack_element, but now that that's established, I'm not sure it's very valuable.

bolshakov-a: May be worth leaving here for better clarity that there is nothing specific in `__type_pack_element` with double-reporting.

kimgr: I'll add a comment to explain it's there just for comparison.

kimgr: Done

iwyu.cc
kimgr: Done, thanks!

PR #1836, opened 2025-10-20
Simplify IsDefArgSpecified

bolshakov-a:
There is no need to analyze source locations: there already is a nice `hasInheritedDefaultArg()` method.

One more test case has been added to check the presence of `param->hasDefaultArg()` call.

No functional change is intended.


bolshakov-a:
Maybe, it is better to separate the test change in a distinct commit to make clear that it doesn't show any actual improvement?

kimgr:
Very nice! I think it's OK to bundle these together.

PR #1842, opened 2025-10-25
Avoid wrong template fwd-decl due to explicit spec

bolshakov-a:
Prior to this, `SpecializedClassTpl<int>* pscti;` followed by an explicit specialization of the template caused a forward-declaration suggestion even if the primary template has default template arguments, which resulted in uncompilable code. Despite at A2 stage it was recategorized to full use, at B2 stage it was recategorized back to fwd-decl use because the referred declaration was after the use. An analogous problem with explicit instantiations was solved in #1556 by means of `GetInstantiatedFromDecl` call, but it did not work with explicit specializations because they are not instantiated.

The main change is in `VisitTemplateSpecializationType`. It reports now primary template declarations for fwd-decl uses and not any specialization. For the full use reporting, the `GetInstantiatedFromDecl` removal should not change behavior because it is anyway called later inside `GetDefinitionAsWritten` call from `ReportFullSymbolUse`.

The change in `iwyu_output.cc` is just to maintain the behavior proposed in #735. Including a file containing any template specialization definition is considered like including the primary template defining file for the purpose of determining if a forward-declaration should be suggested. Note that explicit instantiation declarations are considered as definitions. Not sure if a short-circuiting of implicit instantiations is worth placing for optimization.
`DeclIsVisibleToUseInSameFile` call has been replaced with `IsBeforeInSameFile` because the former returns true also when the locations coincide, and class template explicit specializations would be considered as primary template forward-decl providers for itself.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3203,21 +3203,18 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreCurrentASTNode() || CanIgnoreType(type))
       return true;
 
-    // If not a full explicit specialization, report the original template
-    // or a partial specialization from which the specialization is instantiated
-    // (and don't report explicit instantiations here).
-    const NamedDecl* decl = GetInstantiatedFromDecl(TypeToDeclAsWritten(type));
-
     // If we are forward-declarable, so are our template arguments.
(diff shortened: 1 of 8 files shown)


kimgr:
LGTM, thanks!

PR #1844, opened 2025-10-26
StringRef-ize utility modules

kimgr:
LLVM has a nice internal version of `std::string_view` called `llvm::StringRef`. Its API is much richer than `std::string`, and it's more efficient in general (avoids allocations for temporaries) at the cost of some lifetime safety.

Many of LLVM's APIs are expressed in terms of `StringRef`, so use it for our base string and path modules to make it easier to interoperate without temporary `std::string`s.

### DIFF
--- a/iwyu_location_util.h
+++ b/iwyu_location_util.h
@@ -51,7 +51,6 @@
 #include "clang/Lex/Token.h"
 #include "iwyu_globals.h"
 #include "iwyu_path_util.h"
-#include "llvm/ADT/StringRef.h"
 
 namespace clang {
 class Decl;
@@ -83,11 +82,11 @@ bool IsInScratchSpace(clang::SourceLocation loc);
 
 // Resolve canonical file path from various file entry types.
 inline string GetFilePath(clang::OptionalFileEntryRef file) {
-  return (!file ? "<built-in>" : NormalizeFilePath(file->getName().str()));
+  return (!file ? "<built-in>" : NormalizeFilePath(file->getName()));
 }
 
 inline string GetFilePath(clang::FileEntryRef file) {
-  return NormalizeFilePath(file.getName().str());
+  return NormalizeFilePath(file.getName());
 }
 
 //------------------------------------------------------------
--- a/iwyu_path_util.cc
+++ b/iwyu_path_util.cc
@@ -15,7 +15,6 @@
 #include "iwyu_port.h"
 #include "iwyu_string_util.h"
 #include "llvm/ADT/SmallString.h"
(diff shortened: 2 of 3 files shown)


bolshakov-a:
> Its API is much richer than std::string

It doesn't sound good, according to Herb Sutter's point of view:
http://www.gotw.ca/gotw/084.htm

iwyu_string_util.h
```diff
@@ -133,40 +132,39 @@ inline void StripWhiteSpace(string* str) {
...
   // If max_segs is 0, the first part of the condition will always be true.
   while (retval.size() + 1 != max_segs &&
          (pos = str.find(divider)) != string::npos) {
-    retval.push_back(str.substr(0, pos));
-    str = str.substr(pos + divider.length());
+    retval.emplace_back(str.substr(0, pos));
+    str = str.substr(pos + divider.size());
   }
-  retval.push_back(str);
+  retval.emplace_back(str);
```
bolshakov-a, line 155: FYI, replacing `push_back` with `emplace_back` doesn't make code more effective in such cases:
https://godbolt.org/z/vYcdYsGTT

kimgr: Ah, I mostly replaced it mechanically since I saw most of the other cases were now rvalues.

Should be possible to wrap this in `std::move` now to avoid a copy?

kimgr:
> It doesn't sound good, according to Herb Sutter's point of view:
> http://www.gotw.ca/gotw/084.htm

Yeah, that makes sense. But ironically, `std::string` is still picking up new method in C++20 and 23: https://en.cppreference.com/w/cpp/string/basic_string.html#Operations.

So if we can transition to `StringRef` (which makes an effort to mirror `std::string` present and future), it will be much easier to switch to `std::string_view` and newer `std::string` methods as they become available.

iwyu_string_util.h
bolshakov-a: Even with rvalue cases, `emplace_back` is not better than `push_back` when just a copy/move is occurred, as shown in my example. But I don't mind if you just write `emplace_back` by default so as not to think about the difference.

If you think the performance gain outweighs the danger of use-after-move here, you may insert `std::move`.

bolshakov-a:
`std::string_view` is already available in C++17.

I think the logic of the standard committee members is that it is better to adhere to the internal interface and extend it than mix it with external interface functions. But in fact, different people have different opinions on such questions:
https://martinfowler.com/bliki/HumaneInterface.html
https://martinfowler.com/bliki/MinimalInterface.html

PR #1846, opened 2025-10-26
Make '<stdin>' a special filename

kimgr:
Clang sets the filename for source text coming from stdin to `<stdin>`. In most places where IsSpecialFilename was checked, stdin was also checked separately.

Consider it a special filename and simplify callers. Move
IsSpecialFilename out-of-line.

Allow any special filename in OneUse constructor, not just `<stdin>`.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -310,7 +310,7 @@ OneUse::OneUse(const string& symbol_name, OptionalFileEntryRef dfn_file,
       is_iwyu_violation_(false) {
   CHECK_(dfn_file && "OneUse: dfn_file must be set");
   CHECK_(!decl_filepath_.empty() && "OneUse: dfn_file must have a name");
-  CHECK_(decl_filepath_ == "<stdin>" || !IsQuotedInclude(decl_filepath_))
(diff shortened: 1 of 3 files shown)


bolshakov-a:
This has broken taking the input from stdin, actually. This causes an assertion failure now:
```
$ include-what-you-use -xc -
int i;
/home/bols/progs/iwyu/include-what-you-use/iwyu_preprocessor.cc:1070: Assertion failed: main_file_ && "Main file should be present"
```
The reason is seemingly here:
https://github.com/include-what-you-use/include-what-you-use/blob/f18eea6656070407e8f6b968f3b9ddb7ecb19d13/iwyu_preprocessor.cc#L856

kimgr:
Yikes! Good catch. I'll look into that when I have a minute. I was just thinking of ways to test stdin input, it doesn't really fit in our test framework.

kimgr:
So I wrote up a minimal test framework on the side for stdin tests. And it fails on master, but also fails if I revert this PR, because I built even more assumptions on it here: https://github.com/include-what-you-use/include-what-you-use/pull/1847.

I'll try to find a good compromise and push it together with the new testing.

PR #1848, opened 2025-10-26
Force including fwd-decls with default tpl args

bolshakov-a:
A class template declaration specifying a default template argument should be included regardless of whether it is the complete definition or just a forward-declaration. Otherwise, two forward-declarations specifying the same default template argument may appear in a translation unit (e.g. when the header with the original fwd-decl is included for some other reason), and this would break compilation.


bolshakov-a:
@kimgr, your bugs suite works!

PR #1852, opened 2025-10-26
Select tpl redecl with def arg needed at use site

bolshakov-a:
Mostly similar to https://github.com/include-what-you-use/include-what-you-use/commit/f39823818b962824c178873fa2fe0c3243f9fb42, but this is for class template and alias template default arguments. This doesn't fix behavior with variable and function templates.

This fixes the case reported in https://github.com/include-what-you-use/include-what-you-use/issues/1763#issuecomment-2949495499.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3205,10 +3205,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
     // If we are forward-declarable, so are our template arguments.
     if (CanForwardDeclareType(current_ast_node())) {
-      if (!InImplicitCode(current_ast_node())) {
-        const NamedDecl* decl = type->getTemplateName().getAsTemplateDecl();
-        ReportDeclForwardDeclareUse(CurrentLoc(), decl);
-      }
       current_ast_node()->set_in_forward_declare_context(true);
     } else {
       if (type->isTypeAlias())
@@ -5318,12 +5314,28 @@ class IwyuAstConsumer
     if (CanIgnoreCurrentASTNode())
       return true;
 
+    bool can_fwd_decl = CanForwardDeclareType(current_ast_node());
     // If we're not in a forward-declare context, use of a template
     // specialization requires having the full type information.
-    if (!CanForwardDeclareType(current_ast_node()) || type->isTypeAlias()) {
+    if (!can_fwd_decl || type->isTypeAlias()) {
       const TemplateInstantiationData data = GetTplInstData(type);
       instantiated_template_visitor_.ScanInstantiatedType(
           current_ast_node(), data.resugar_map, data.provided_types);
+    }
+
+    if (!InImplicitCode(current_ast_node())) {
+      const TemplateDecl* tpl = type->getTemplateName().getAsTemplateDecl();
+      size_t num_written_args = type->template_arguments().size();
+      if (num_written_args < tpl->getTemplateParameters()->size()) {
+        // If the type as-written makes use of default arguments, select exactly
+        // that redeclaration which specifies the first default argument used.
+        ReportDeclForwardDeclareUse(
+            CurrentLoc(), GetRedeclSpecifyingDefArg(num_written_args, tpl));
+      } else if (can_fwd_decl || type->isTypeAlias()) {
+        // TODO: instead of GetIndependentRedecl call, figure out how to print
+        // a fwd-decl without default arguments if there isn't any.
+        ReportDeclForwardDeclareUse(CurrentLoc(), GetIndependentRedecl(tpl));
+      }
     }
 
     const auto [is_provided, comment] =
@@ -5331,7 +5343,7 @@ class IwyuAstConsumer
     // Don't call ReportTypeUse here, because scanning of template instantiation
     // internals should be avoided: it is allowed not to provide some of
     // template args.
-    if (is_provided || type->isTypeAlias())
+    if (is_provided)
       ReportDeclUse(CurrentLoc(), TypeToDeclAsWritten(type), comment);
 
     return Base::VisitTemplateSpecializationType(type);
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -130,6 +130,7 @@ using clang::RecordDecl;
 using clang::RecordType;
 using clang::RecursiveASTVisitor;
 using clang::Redeclarable;
+using clang::RedeclarableTemplateDecl;
 using clang::ReferenceType;
 using clang::Sema;
 using clang::SourceLocation;
@@ -234,6 +235,15 @@ static const Type* DesugarAliasTypes(const Type* type) {
 static bool IsImplicitConversionCtor(const CXXConstructorDecl* ctor) {
   return !ctor->isExplicit() && ctor->getNumParams() == 1 &&
          !ctor->isCopyOrMoveConstructor();
+}
+
+static const RedeclarableTemplateDecl* GetRedeclNotSpecifyingAnyDefArg(
+    const RedeclarableTemplateDecl* decl) {
+  for (const RedeclarableTemplateDecl* redecl : decl->redecls()) {
+    if (none_of(*redecl->getTemplateParameters(), IsDefTplArgSpecified))
+      return redecl;
+  }
+  return nullptr;
 }
 
 }  // anonymous namespace
@@ -1254,7 +1264,7 @@ bool IsNestedClass(const TagDecl* decl) {
 
 bool HasDefaultTemplateParameters(const TemplateDecl* decl) {
   TemplateParameterList* tpl_params = decl->getTemplateParameters();
-  return tpl_params->getMinRequiredArguments() < tpl_params->size();
+  return any_of(*tpl_params, IsDefTplArgSpecified);
 }
 
 template <class T>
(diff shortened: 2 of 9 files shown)


bolshakov-a:
I've added a test case representing the case from https://github.com/include-what-you-use/include-what-you-use/issues/1763#issuecomment-2949495499.

bolshakov-a:
`InImplicitCode` call has been moved to a more correct place.

iwyu_ast_util.cc
```diff
@@ -1422,6 +1432,31 @@ FunctionDecl* GetRedeclSpecifyingDefArg(unsigned i, FunctionDecl* func) {
...
+  // If this template isn't redeclarable, then the given declaration should
+  // specify default arguments, if any.
+  if (!redeclarable)
+    return decl;
+  for (const RedeclarableTemplateDecl* redecl : redeclarable->redecls()) {
+    const TemplateParameterList* params = redecl->getTemplateParameters();
+    if (IsDefTplArgSpecified(params->getParam(i)))
+      return redecl;
+  }
+  CHECK_UNREACHABLE_("There should be a redecl specifying the default arg");
```
kimgr, line 1447: Could you add a little comment above this check to describe what I did wrong if it fails? I'm guessing this function can only be called with `TemplateDecl`s for which we know there are default parameters, or something?

iwyu_ast_util.cc
```diff
@@ -1422,6 +1432,31 @@ FunctionDecl* GetRedeclSpecifyingDefArg(unsigned i, FunctionDecl* func) {
...
+}
+
+const NamedDecl* GetIndependentRedecl(const NamedDecl* decl) {
+  const auto* redeclarable = dyn_cast<RedeclarableTemplateDecl>(decl);
+  if (!redeclarable)
+    return decl;
+  if (const TemplateDecl* found = GetRedeclNotSpecifyingAnyDefArg(redeclarable))
```
kimgr, line 1454: It might be nice to carry some of the information from doc comment here too, it took me a long time to get my head around what it was doing.

iwyu_output.cc
```diff
@@ -1516,9 +1516,12 @@ void CalculateIwyuForForwardDeclareUse(
   CHECK_(tag_decl && "Non-tag types should have been handled already");
 
   vector<const NamedDecl*> dfns;
-  if (const NamedDecl* dfn = GetTagDefinition(use->decl()))
+  // A definition in any of desired headers makes the forward-declaration
+  // redundant, but not when the fwd-decl specifies a default template argument.
+  bool has_def_arg = tpl_decl && HasDefaultTemplateParameters(tpl_decl);
+  if (const NamedDecl* dfn = GetTagDefinition(use->decl()); dfn && !has_def_arg)
```
kimgr: This packed condition is a little too much for me (even though I've been doing a lot of Golang lately where this is normal style). How about splitting it into:
```
if (!tpl_decl || !HasDefaultTemplateParameters(tpl_decl)) {
  if (const NamedDecl* dfn = GetTagDefinition(use->decl()) {
     dfns.push_back(dfn);
  }

  if (tpl_decl) {
    // ...
  }
}
```

The diff gets a little worse, but I think it might be easier to follow.

kimgr: OTOH, it's fairly unclear that `if (tpl_decl)` is only true if there are no default template params. So maybe worth keeping a little boolean around anyway.

iwyu_ast_util.cc
bolshakov-a: But the comment in the header already states that the function has a precondition that there should be a redeclaration giving a default argument value to the i-th parameter.

iwyu_output.cc
bolshakov-a: Probably, it will be more readable if the declaration of `dfn` variable is moved out of the if-statement. But then it would have longer scope than needed, and another `dfn`-named variables below would hide it.

iwyu.cc
```diff
@@ -5318,20 +5314,36 @@ class IwyuAstConsumer
...
     // If we're not in a forward-declare context, use of a template
     // specialization requires having the full type information.
-    if (!CanForwardDeclareType(current_ast_node()) || type->isTypeAlias()) {
+    if (!can_fwd_decl || type->isTypeAlias()) {
       const TemplateInstantiationData data = GetTplInstData(type);
       instantiated_template_visitor_.ScanInstantiatedType(
           current_ast_node(), data.resugar_map, data.provided_types);
     }
 
+    const TemplateDecl* tpl = type->getTemplateName().getAsTemplateDecl();
```
kimgr: Both of these vars are only used inside the `!InImplicitCode(...)` branch, is it possible to move them in? Or does that hit the 80-col limit and get noisy?

iwyu_ast_util.cc
kimgr: I see, I hadn't gotten that far in the diff when I read this.

iwyu_ast_util.cc
```diff
@@ -1422,6 +1432,34 @@ FunctionDecl* GetRedeclSpecifyingDefArg(unsigned i, FunctionDecl* func) {
...
+    return found;
+  // If there is no redeclaration without any explicitly written default
+  // arguments, return the one specifying the last argument because it doesn't
+  // depend on any other redeclaration.
+  unsigned last_param_idx = redeclarable->getTemplateParameters()->size() - 1;
+  return GetRedeclSpecifyingDefArg(last_param_idx, redeclarable);
```
kimgr, line 1460: Wait, how do we know this doesn't hit the the `CHECK_` above?

iwyu_output.cc
kimgr: I think I prefer the outer nesting I suggested above, it makes it much clearer where the default-argument comment applies as well.

iwyu.cc
bolshakov-a: I have simply overlooked that, thanks!

iwyu_ast_util.cc
bolshakov-a: Otherwise, the code wouldn't compile. Only function templates allow specifying some template arguments without the last one, but `GetIndependentRedecl` is not called for function templates.

kimgr: :exploding_head: OK, I think I see now.

I wonder if `GetIndependentRedecl` is too general a name for the assumptions that it makes. E.g. it's not obvious to me (if I forget about what you just showed me) that I can't call this function if `(num_written_args < tpl->getTemplateParameters()->size())`.

So I wonder if it would be better to inline it, even if it's quite noisy. That way we don't have to give it an overly specific name.

bolshakov-a: > that I can't call this function if (num_written_args < tpl->getTemplateParameters()->size()).

You can call it, why not? It doesn't depend on `TemplateSpecializationType` and its written arguments in any way.

It still seems to me that having it a distinct function makes the code more clear.

In fact, ideally, it should not search for the redecl specifying the last default argument but instead somehow construct a new one not specifying any default argument, as the TODO in `VisitTemplateSpecializationType` states. Because if all the template arguments are explicitly written, the template may be forward-declared, and this should not depend on whether there is a redeclaration convenient for IWYU somewhere or not.

kimgr: I was thinking that if `num_written < tpl->getTemplateParameters()->size()`, then the:
```
unsigned last_param_idx = redeclarable->getTemplateParameters()->size() - 1;
return GetRedeclSpecifyingDefArg(last_param_idx, redeclarable);
```
portion will be looking for someone who specified the last param on L1444: https://github.com/include-what-you-use/include-what-you-use/pull/1852/files#diff-7ec045b2b5739efe5504c9d31e7e935fe9405be09ae655d067bda89fdd9cf6f9R1444.

Since not all arguments were written, I was thinking no redecl will be found.

But I think I see now -- the `num_written` constraint is on the decl we're looking at, but there MUST exist redecls that specify `num_written..size - 1` args, otherwise the user code wouldn't compile.

I still think it's pretty subtle, but I don't know how to make it clearer. At least I've convinced myself it shouldn't crash :-)

bolshakov-a: Note that all the redeclarations should specify the same number of template _parameters_, so you cannot e.g. write:
```cpp
template <int, int = 0>
class Tpl;

template <int>  // compilation error
class Tpl;
```
Hence `redeclarable->getTemplateParameters()->size() == tpl->getTemplateParameters()->size()` for all the redeclarations.

One more time: `GetIndependentRedecl` is unrelated to any `TemplateSpecializationType`, it may be called even there is no one, just template declarations without usages. The only difference could in theory be what exact "redeclaration not specifying any def arg" is returned may depend on the exact given template redeclaration, which in turn may depend on a template specialization type.

kimgr: I can't take it in, too tired :-).

But I understand it better now than before, thank you for explaining!

bolshakov-a: I completely understand you, so I had nothing against postponing the review until e. g. the weekend, as long as you cannot work on IWYU in business hours.

> But I understand it better now than before

TBH, most of the test cases marked by me as TODO pass on the previous master, so it may be seen as some regression. This is because clang lookup finds the last redecl which is in the .cc-file, and hence IWYU did not report anything, but now `GetIndependentRedecl` remaps them to declarations from the header. So, it is a trade-off.

PR #1853, opened 2025-10-29
Visit concept reference instead of spec expr

bolshakov-a:
It is more general and fixes reporting of concepts referenced in auto type constraints.


kimgr:
Nice, thanks!

PR #1854, opened 2025-10-29
Avoid traversing deduced type of auto type

bolshakov-a:
Prior to this, IWYU required complete type for the cases like
```cpp
[&ref1 = ref2]{}();
```
It was because IWYU traversed the (implicit) type in the ref1 declaration that appeared to be `AutoType` with the corresponding deduced underlying type, and then traversed that deduced type, but the presence of `AutoType` node broke the analysis inside `CanForwardDeclareType`.

For the most of the other auto type use cases, the AST nodes corresponding to a written auto type, surprisingly, don't have any deduced type, hence the problem did not arise.

I've decided not to special-case `CanForwardDeclareType` but to make `TraverseAutoTypeLoc` more similar to other sugar type traverse methods.

I'm not sure whether `TraverseAutoType` should be overloaded similarly.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -180,6 +180,7 @@ using clang::ASTFrontendAction;
 using clang::ArraySubscriptExpr;
 using clang::ArrayType;
 using clang::Attr;
+using clang::AutoTypeLoc;
 using clang::BinaryOperator;
 using clang::BinaryOperatorKind;
 using clang::CXXBaseSpecifier;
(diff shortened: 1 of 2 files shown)


bolshakov-a:
> I'm not sure whether TraverseAutoType should be overloaded similarly.

I tried the following:
```cpp
bool TraverseAutoType(clang::AutoType* type, bool) {
  CHECK_(type->getDeducedType().isNull());
  return true;
}
```
and the IWYU tests pass, and the dogfood ran successfully. I verified that the method was actually called in some cases.

FWIW, I also tried to remove deduced type traversing from `RecursiveASTVisitor`, and all the clang tests pass. However, there may be external users relying on the current behavior, so I'm not sure about making a PR in clang.

kimgr:
It would feel balanced to override `TraverseAutoType` as well. Maybe add a log at low verbosity level so we notice if the deduced type starts being non-null due to changes in Clang?

kimgr:
Otherwise, LGTM! Very nice.

bolshakov-a:
There is another concern: we should somehow monitor changes being made to the original versions of these methods and copy-paste them into our versions. E.g., the original version of `TraverseAutoType` doesn't contain concept reference traversing but instead contains traversing of concept template arguments. I don't currently know how to test it. I'd prefer to keep the original version at work until it is obvious that it doesn't fit to IWYU.

PR #1855, opened 2025-10-29
Treat stdin separately from special files where appropriate

kimgr:
In 492f292b42365d4be2630e17e7bab84006595fe7, I folded `"<stdin>"` into the
set of special filenames recognized by IsSpecialFilename.

This broke the following IWYU invocation:

    $ echo "int i;" | include-what-you-use -xc -
    .../iwyu_preprocessor.cc:1070: Assertion failed: main_file_ && "Main file should be present"
    Aborted (core dumped)

Remove `"<stdin>"` from the special filenames and add a new predicate,
IsSpecialFilenameOrStdin, which includes it. Add similar wrappers
IsSpecialFileOrStdin and IsInSpecialFileOrStdin.

Carefully pick one or the other for all existing uses of Is.*SpecialFile
to get the best semantics. Include stdin if there's any uncertainty.

Consistently treat stdin as not a headerfile and not a quoted-include.

Our current test suite has no mechanism for testing inputs read from
stdin. Add a new Bash-based test runner and a new tests/stdin suite
containing inputs suitable for triggering issues with stdin inputs.

There is no facility for content assertions or anything, we just check
that the include-what-you-use invocation doesn't fail.

The tests/stdin/minimal.cc testcase was failing on master before this
patch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,10 @@ jobs:
           cd ./build
           CTEST_OUTPUT_ON_FAILURE=1 ctest -j 6
 
+      - name: Test include-what-you-use (stdin)
+        run: |
+          ./iwyu-run-stdin-tests.bash ./build/bin/include-what-you-use
+
       - name: Check if any known bugs were fixed
         run: |
           ./run_iwyu_tests.py --extra-suite=bugs bugs -- ./build/bin/include-what-you-use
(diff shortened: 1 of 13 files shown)


bolshakov-a:
Thanks!

kimgr:
I think I'll squash this PR into one commit after review to make sure we don't have failing tests on master. Or alternatively, reorder the commits.

@bolshakov-a Let me know what you think!

bolshakov-a:
At a glance, I don't see any problems.

PR #1856, opened 2025-10-30
Improve blocking implicit typedef reports

bolshakov-a:
The only reason why it has not been done by means of replacing `VisitTypedefType` with `VisitTypedefTypeLoc` is dynamic exception specifications, which use `Type`s and not `TypeLoc`s. But they were removed in C++17, so I'm not sure if testing them is needed.


kimgr:
Thanks!

PR #1869, opened 2025-11-22
Differentiate explicit specializations in mappings

bolshakov-a:
Prior to this, IWYU took into account only template names when looking for a symbol in mapping files. So, users could not distinguish between a primary template and its explicit or partial specializations despite they can be placed in separate files. Moreover, some libraries allow users to write their own specializations of some library templates, but IWYU erroneously remapped a use of such a specialization to the primary template if it is mentioned in a mapping file for the library. As an example, if someone writes a specialization of `std::allocator` for his class, IWYU would not report the specialization header because `std::allocator` is mapped to the standard library headers inside `stdlib_cxx_symbol_map` from `iwyu_include_picker.cc`.

Therefore, this change is needed to write the standard library symbol mapping correctly and not to break users who specialize some of its class templates.

This change makes template arguments included in a symbol name. For partial specializations, template parameters used inside template arguments are replaced with `:N`, where `N` is the partial specialization template parameter number. Template parameter depth is not currently included in the placeholder.

Only class template specializations have been supported, not function or variable ones.

Note that if some IWYU users rely on the old behavior, it is a breaking change for them because IWYU will stop mapping explicit specializations to the correct headers until fixing the corresponding mapping files.

The change in `explicit_instantiation2.cc` test reveals that IWYU reports fwd-decl uses from prototypes of implicitly defined copy-ctors (`Template(const Template<char>&)` in this case). I'm not sure if it causes some bug in general, but this PR should not make any regression with this.

### DIFF
--- a/docs/IWYUMappings.md
+++ b/docs/IWYUMappings.md
@@ -132,6 +132,42 @@ syntax in the first entry. Track the [following
 bug](https://github.com/include-what-you-use/include-what-you-use/issues/233)
 for updates.
 
+#### Template specializations ####
+
+A bare template name in a symbol mapping corresponds to the primary template,
+i.e. any use of its implicit specialization (and not any implicit specialization
+of its partial specialization) matches the mapping. To map an explicit or
+partial specialization, append the template argument list to the template name.
+Note that, currently, IWYU expects a specific format of template argument lists.
+For example, a list consisting of a pointer to `int` should be written as
+`<int *>` and not as e.g. `<int*>`. When in doubt, run IWYU with verbosity level
+6 on some code using the specialization, and look for a string like `Marked
+full-info use of decl template-name<...> (from ...` to see how IWYU spells
+the template arguments.
+
+To denote a partial specialization, you should replace all the template
+parameters in the template argument list with `:N`, where `N` is the parameter
+number, starting from 0. Note that it is the number of the partial
+specialization template parameter, not of any primary template parameter. For
+example, for such a code:
+
+    template <typename, typename, int>
+    class Tpl;
+
+    template <typename T1, typename T2>
+    class Tpl<T1, T2, 5> {};
+
+    template <typename T>
+    class Tpl<T, T, 5> {};
+
+the first partial specialization can be referred to in a mapping file as
+`Tpl<:0, :1, 5>`, and the second one as `Tpl<:0, :0, 5>`. IWYU does not
+currently have a proper support for mappings of partial specializations
+of templates being members of other templates.
+
+At present, only class template specializations are supported, not function
+or variable ones.
+
 
 ### Mapping refs ###
 
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -11,6 +11,8 @@
 
 #include "iwyu_ast_util.h"
 
+#include <cstddef>
+#include <regex>
 #include <set>                          // for set
 #include <string>                       // for string, operator+, etc
 #include <utility>                      // for pair
@@ -177,6 +179,8 @@ using llvm::raw_string_ostream;
 using llvm::zip_equal;
 using std::function;
 using std::pair;
+using std::regex;
+using std::regex_replace;
 using std::vector;
 
 namespace include_what_you_use {
@@ -533,7 +537,43 @@ string GetWrittenQualifiedNameAsString(const NamedDecl* named_decl) {
       named_decl->getASTContext().getPrintingPolicy();
   printing_policy.SuppressUnwrittenScope = true;
   named_decl->printQualifiedName(ostream, printing_policy);
-  return ostream.str();
+
+  if (const auto* spec =
+          dyn_cast<ClassTemplateSpecializationDecl>(named_decl)) {
+    if (spec->isExplicitSpecialization()) {
+      ostream << '<';
+      const TemplateArgumentList& args = spec->getTemplateArgs();
+      const TemplateParameterList* params =
+          spec->getSpecializedTemplate()->getTemplateParameters();
+      for (unsigned i = 0, ie = args.size(); i < ie; ++i) {
+        if (i > 0)
+          ostream << ", ";
+        bool include_type = TemplateParameterList::shouldIncludeTypeForArgument(
+            printing_policy, params, i);
+        const TemplateArgument& arg = args.get(i);
+        // Clang embraces arguments corresponding to a parameter pack in angle
+        // brackets, so cannot use the default 'print' for them.
+        if (arg.getKind() == TemplateArgument::Pack) {
+          ArrayRef<TemplateArgument> pack_args = arg.getPackAsArray();
+          for (size_t j = 0, je = pack_args.size(); j < je; ++j) {
+            if (j > 0)
+              ostream << ", ";
+            pack_args[j].print(printing_policy, ostream, include_type);
+          }
+        } else {
+          args.get(i).print(printing_policy, ostream, include_type);
+        }
+      }
+      ostream << '>';
+
+      // Replace clang placeholders in partial specialization arguments with :N,
+      // where N is the parameter index (depth is ignored for now).
+      static regex tpl_param{"(?:type|value|template)-parameter-\\d+-(\\d+)"};
+      retval = regex_replace(retval, tpl_param, ":$1");
+    }
+  }
+
+  return retval;
 }
 
 // --- Utilities for Template Arguments.
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -325,8 +325,23 @@ const IncludeMapEntry stdlib_cxx_symbol_map[] = {
   // argument, and sometimes it bleeds through when clang desugars the
   // string/ostream/istream type.
   { "std::char_traits", kPrivate, "<string>", kPublic },
+  { "std::char_traits<char>", kPrivate, "<string>", kPublic },
+  { "std::char_traits<char8_t>", kPrivate, "<string>", kPublic },
+  { "std::char_traits<char16_t>", kPrivate, "<string>", kPublic },
+  { "std::char_traits<char32_t>", kPrivate, "<string>", kPublic },
+  { "std::char_traits<wchar_t>", kPrivate, "<string>", kPublic },
   { "std::char_traits", kPrivate, "<ostream>", kPublic },
+  { "std::char_traits<char>", kPrivate, "<ostream>", kPublic },
+  { "std::char_traits<char8_t>", kPrivate, "<ostream>", kPublic },
+  { "std::char_traits<char16_t>", kPrivate, "<ostream>", kPublic },
+  { "std::char_traits<char32_t>", kPrivate, "<ostream>", kPublic },
+  { "std::char_traits<wchar_t>", kPrivate, "<ostream>", kPublic },
   { "std::char_traits", kPrivate, "<istream>", kPublic },
+  { "std::char_traits<char>", kPrivate, "<istream>", kPublic },
+  { "std::char_traits<char8_t>", kPrivate, "<istream>", kPublic },
+  { "std::char_traits<char16_t>", kPrivate, "<istream>", kPublic },
+  { "std::char_traits<char32_t>", kPrivate, "<istream>", kPublic },
+  { "std::char_traits<wchar_t>", kPrivate, "<istream>", kPublic },
 
   // std::ptrdiff_t is often an architecture specific definition, force the
   // canonical location.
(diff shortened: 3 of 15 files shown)


bolshakov-a:
@aaronballman, @endilll, this is what I talked about, if you have a chance to look into. I decided to use tentatively the question mark as a placeholder for now. Things that are valid identifiers in C++ (like `_`) should better be avoided, as well as C++ operators (`<* *>` for the partial specialization for pointers would at least look strange).

I don't expect any confusion of template template parameters with other template parameter kinds because the kind of some argument is unambiguously specified by the primary template (or I haven't got your idea). Regarding default arguments, a mapping file writer should specify all the parameters explicitly, in the presented approach.

Clashes are possible in cases like this:
```cpp
template <typename, typename, int>
class Tpl;

template <typename T1, typename T2>
class Tpl<T1, T2, 1> {};

template <typename T> 
class Tpl<T, T, 1> {};

Tpl<int, int, 1> t1;
Tpl<int, char, 1> t2;
```
In the presented solution, a user cannot distinguish between the two partial specializations in a mapping. Probably, template parameter indices should be kept in the placeholders...

kimgr:
I think it's quite elegant. I was a bit concerned with the `std::regex` at first, but I see what you're doing there and it's not more crazy than the other decl rewriting we're doing in `MungedForwardDeclare...` for example.

It's a bit annoying that both `?` (and `*` for pointers) need to be escaped in tests, but I think it also looks very clear in mappings. You might consider something like `%` to get out of both C++ identifiers and regex syntax, but it wouldn't look as obvious.

I was looking at the potential clash example, and I'm not fully seeing how to avoid it. Could you show an example of what the conflicting mappings would look like, and how parameter indices might help? I'm thinking `%0, %1` etc might actually read quite nicely, but it looks a lot more complex by default in mappings.

Endilll:
> Probably, template parameter indices should be kept in the placeholders...

If this is meant to be a solution for the general problem, then you also need to deal with this:
```cpp
template<typename T1, typename T2>
struct S {
  template<typename, typename>
  struct Inner {};

  template<typename U>
  struct Inner<U, T1> {}; // how would you name this using indicies?
};
```
Generally, I'm hinting at the existence of Clang's `MultiLevelTemplateArgumentList` (and equivalent entities in other implementations) to represent this.

bolshakov-a:
> Could you show an example of what the conflicting mappings would look like, and how parameter indices might help?

In the presented approach, the both partial template specializations from my example are named `Tpl<?, ?, 1>`; whereas with indices, the first one would be named `Tpl<?0, ?1, 1>` and the second one - `Tpl<?0, ?0, 1>` (or with `%` instead of `?`). Note that these are indices of partial spec template parameters, not of the primary template parameters.

> If this is meant to be a solution for the general problem, then you also need to deal with this:

At first I thought that member templates should not be mapped to a header different from their hosting template header, so the problem doesn't exist. But now I see that specializations of member templates may be defined outside of the hosting class, like this:
```cpp
template <typename T1, typename T2>
template <typename U>
struct S<T1, T2>::Inner<U, T1> {
};


template <typename T1, typename T2>
template <typename U>
struct S<T1, T2>::Inner<U, U> {
};
```
and placed in different files.

So maybe just leave clang-generated names as-is? However, `type-parameter-0-0` looks too verbose to me, and I'm still not sure if the parameter kind matters or not.

Once a decision would be made, it would be hard to change it later.

bolshakov-a:
One more question to pay attention is the spacing in types. Clang makes `int *`, although `int*` may look more conventional. However, Clang doesn't currently allow to change it by means of the policy.

AaronBallman:
> One more question to pay attention is the spacing in types. Clang makes `int *`, although `int*` may look more conventional. However, Clang doesn't currently allow to change it by means of the policy.

Traditionally, we leave formatting decisions like that up to clang-format rather than trying to handle it in printing policies. (Aside: it would be neat if Clang's fix-it machinery was hooked up to clang-format so that we'd automatically apply settings from the `.clang-format` config file.)

bolshakov-a:
@AaronBallman, if you mean that `TypePrinter` output should depend from a (somewhere found) config file, it would be bad for IWYU mapping files because they would become non-portable. As for pointer or reference sign spacing, it may probably be changed by simple substring replacement. Or just leaved as-is, with a note in the docs.

kimgr:
> So maybe just leave clang-generated names as-is? However, `type-parameter-0-0` looks too verbose to me, and I'm still not sure if the parameter kind matters or not.

Yeah, that full placeholder syntax is going to get very noisy in mappings.

I wonder if it would be possible to "compact" indices also, so they're only spelled out when there's a need to disambiguate? I think 95% of the cases are going to be relatively simple specializations, and then there's a long tail of extremely niche constructs. It seems like a shame to force a fully general syntax on the majority, but it would obviously be nice to be fully general for the minority.

bolshakov-a:
I started to think that the best choice would probably be `0-0` format: it is both concise and general (if parameter kinds are actually not needed). AFAIU, there should be no ambiguity with arithmetic subtraction (or some other delimiter may be taken).

kimgr:
The `0-0` syntax is nice and compact, but it's also quite hard to explain the (depth,index) two-dimensionality. Particularly since I guess most specializations will be at depth 0. I wonder if prefixing with colon might give a syntax that works to "stack", i.e.:
```
:0   - index only
:1:0 - depth + index 
```

bolshakov-a:
> I wonder if prefixing with colon might give a syntax that works to "stack"

Do you mean "If all the depths equal to 0, skip them" or "If a depth equals to 0, skip it"? Anyway, looks like a good idea!

kimgr:
I hadn't really thought about it like that, but I guess it would be clearest to only skip depth if all depths are zero. I was trying to experiment with this yesterday, but it's still not fully clear to me how it works end to end...

Is it something like: 
* We render symbol names for template specializations including tpl parameters
* For _partial_ specializations, this special syntax kicks in where we print placeholders instead of actual types
* (Something something where the parameter indices are necessary to disambiguate)
* Once we can do unambiguous printing of symbol names, we can let users add mappings for these symbols without any changes to the mapping logic?

bolshakov-a:
I think you have understood it correctly. Keys of the internal symbol map are just strings (as opposed to `clang::Type`s), in the current implementation. Hence, users should write type names in their mappings exactly like IWYU prints them.

bolshakov-a:
I have discovered that types nested inside templates cannot currently be represented correctly in mappings anyway, because they are considered to be inside the context of the full specialization (not partial or the primary template). Hence, `printQualifiedName` call prints exact arguments used, without placeholders, which makes no sense in mappings for implicit specializations. Due to this, I think, I have not succeeded in obtaining any depth in partial template spec arguments other than 0.

We could implement now only placeholders with indices and add the depth part to the stacked placeholder some day later if required.

bolshakov-a:
I've made the changes and added some documentation. Not sure if any of the mapping files delivered with IWYU needs fixing.

bolshakov-a:
I've found an issue with parameter packs. They cause incorrect output.

bolshakov-a:
> I've found an issue with parameter packs. They cause incorrect output.

Fixed. I think that `TplParamPack2<int, :0, :1...>` is less surprising than `TplParamPack2<int, <:0, :1...>>`.

iwyu_ast_util.cc
```diff
@@ -533,7 +536,43 @@ string GetWrittenQualifiedNameAsString(const NamedDecl* named_decl) {
...
+        if (arg.getKind() == TemplateArgument::Pack) {
+          ArrayRef<TemplateArgument> pack_args = arg.getPackAsArray();
+          for (size_t j = 0, je = pack_args.size(); j < je; ++j) {
+            if (j > 0)
+              ostream << ", ";
+            pack_args[j].print(printing_policy, ostream, include_type);
+          }
+        } else {
+          args.get(i).print(printing_policy, ostream, include_type);
+        }
```
bolshakov-a, line 565: Template argument printing might be a recursive function, but `include_type` hinders. It should have `variant<const TemplateParameterList*, bool>` parameter or something else...

docs/IWYUMappings.md
```diff
@@ -132,6 +132,42 @@ syntax in the first entry. Track the [following
 bug](https://github.com/include-what-you-use/include-what-you-use/issues/233)
 for updates.
 
+#### Template specializations ####
+
+A bare template name in a symbol mapping correspond to the primary template,
```
kimgr: typo: correspond -> corresponds

docs/IWYUMappings.md
```diff
@@ -132,6 +132,42 @@ syntax in the first entry. Track the [following
...
+i.e. any use of its implicit specialization (and not any implicit specialization
+of its partial specialization) matches the mapping. To map an explicit or
+partial specialization, append the template argument list to the template name.
+Note that, currently, IWYU expects a specific format of template argument lists.
+For example, a list consisting of a pointer to `int` should be written as
+`<int *>` and not as e.g. `<int*>`. When in doubt, run IWYU with verbosity level
+6 on some code using the specialization, and look for a string like `Marked
+full-info use of decl `*`template-name`*`<...> (from ...` to see how IWYU spells
```
kimgr: The asterisks look a little strange there in plain text. I see they make the `template-name` italic, but they also split the code-span into three parts. I think it's understandable even without italic styling.

iwyu_ast_util.cc
```diff
@@ -533,7 +536,43 @@ string GetWrittenQualifiedNameAsString(const NamedDecl* named_decl) {
...
+      }
+      ostream << '>';
+
+      // Replace clang placeholders in partial specialization arguments with :N,
+      // where N is the parameter index (depth is ignored for now).
+      regex tpl_param{"(?:type|value|template)-parameter-\\d+-(\\d+)"};
```
kimgr: It would be nice to compile this regex once and cache it. Can't think of a nice C++ way to do it, other than a function with a static storage duration local regex:
```
const std::regex& BuildTplParamRegex() {
  static regex r{"(?:type|value|template)-parameter-\\d+-(\\d+)"};
  return r;
}
```

(not the name I would choose, but haven't found it yet :)).

Maybe there are more idiomatic ways to compile and reuse?

bolshakov-a: Maybe, just denote it `static` right here, without another function?

kimgr: Oh, yeah. I was hung up on some old c++03 thread safety pattern, I think (but that would have needed extra synchronization anyway, so not sure what I was thinking).

Static local sounds good to me.

kimgr:
LGTM! I see you have a comment open; I have no opinion on that. Let me know when you're ready to merge.

bolshakov-a:
That code just looks ugly. But if you don't mind about it, I'm ready to merge.

kimgr:
> That code just looks ugly. But if you don't mind about it, I'm ready to merge.

We'll have to leave some beauty for the next year. Thank you!

bolshakov-a:
Thanks to all!

bolshakov-a:
One more case where the implemented simple approach doesn't work is C++20 constrained templates:
```cpp
template <typename>
struct Tpl;

template <typename T> requires (sizeof(T) > sizeof(int))
struct Tpl<T> {};

template <typename T> requires (sizeof(T) <= sizeof(int))
struct Tpl<T> {};

int main() {
    Tpl<long long> t1;
    Tpl<int> t2;
}
```
The both partial specializations are currently printed as `Tpl<:0>`. Not sure how important this case is.

bolshakov-a:
> Not sure how important this case is.

It is. There is a clash between
```cpp
template <class T, class charT> requires is-vector-bool-reference <T>
struct formatter<T, charT>;
```
from `<vector>` and
```cpp
template <ranges::input_range R, class charT>
  requires (format_kind<R> != range_format::disabled) &&
           formattable<ranges::range_reference_t<R>, charT>
struct formatter<R, charT>;
```
from `<format>`.

bolshakov-a:
Another clash:
```cpp
template<class, auto> struct Tpl;

template<int I> struct Tpl<int, I> {};

template<auto I> struct Tpl<int, I> {};
```
For the both specializations, Clang prints `value-parameter-0-0`.

By the way, @AaronBallman, @Endilll, please note that Clang disagrees with GCC about which specialization should be selected for chars: https://godbolt.org/z/d7xP1bKba. I'm too lazy to figure out which behavior is correct and to file a bug, but looks like GCC is more right.

bolshakov-a:
> Regarding default arguments, a mapping file writer should specify all the parameters explicitly, in the presented approach.

If the following is true, it would be better to skip defaulted template arguments in the standard library mapping:

> ... this allows vendors to provide implementations of the standard library that have more template parameters for library templates than the standard requires (suitably defaulted, of course, to remain compatible).

http://www.gotw.ca/gotw/034.htm

bolshakov-a:
No, implementations are not actually allowed to add unspecified template parameters: https://discourse.llvm.org/t/additional-parameters-in-stl-templates/89569

Issue #1603, opened 2024-08-30
Better support for Windows.h


justanotheranonymoususer:
There are several mappings you have, but nothing for Windows.h.
Even something as simple as `BYTE` says that `minwindef.h` is missing, even if `Windows.h` is included.

I found this mapping from Azure, perhaps it can be used as a starting point:
https://github.com/Azure/c-build-tools/blob/master/iwyu/rules.imp

Some of the entries:
```python
[
    { include: ["<basetsd.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<consoleapi.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<errhandlingapi.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<fileapi.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<handleapi.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<libloaderapi.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<minwinbase.h>", "public", "\"windows.h\"", "public"] },
    { include: ["<minwindef.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<processenv.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<processthreadsapi.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<securitybaseapi.h>", "public", "\"windows.h\"", "public"] },
    { include: ["<synchapi.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<threadpoolapiset.h>", "public", "\"windows.h\"", "public"] },
    { include: ["<winerror.h>", "private", "\"windows.h\"", "public"] },
    { include: ["<winscard.h>", "public", "\"windows.h\"", "public"] },
    { include: ["<wtypes.h>", "public", "\"windows.h\"", "public"] },
    { include: ["<winreg.h>", "public", "\"windows.h\"", "public"] },
    ...
]
```

There are other mappings online, for example:
https://github.com/Agrael1/BuildIWYU/blob/ecd5efcf0360b109cdae9e5a603c9ae8927604da/msvc.imp#L1692

kimgr:
Hey @justanotheranonymoususer, thanks for the links, I hadn't been keeping track.

[...]

PR #1870, opened 2025-11-27
Add support for generating mappings of Windows.h

itzexpoexpo:
This PR adds a mapping generator for the Win32 API header `<Windows.h>`, including the resulting `windowsh.imp` mapping file.

The mappings are generated by recursively going through all files this header includes, excluding those that are guarded by the `WIN32_LEAN_AND_MEAN` macro so as not to conflict with users that define it and want to cherry-pick the headers explicitly.

Fixes issue #1603.


windowsh.imp
```diff
@@ -0,0 +1,178 @@
...
+  {"include": ["<specstrings.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<specstrings_strict.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<specstrings_undef.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<stdarg.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<stddef.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<stdint.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<stdio.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<stdlib.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<stralign.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<string.h>", "private", "<Windows.h>", "public"]},
```
bolshakov-a: I think it is wrong to mark C standard library headers as private and to map them to `<Windows.h>`.

itzexpoexpo: For some reason I didn't think Windows.h would include the standard library. I added a list-based exclusion for those. Albeit I still see some internal CRT headers being remapped (namely `corecrt.h` and variants) which is probably unavoidable for sake of automation

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,208 @@
...
+import re
+import sys
+import json
+import argparse
+from pathlib import Path
+from collections.abc import Generator
+
+
+LEAN_AND_MEAN_RE = re.compile(r"#ifndef*\s+WIN32_LEAN_AND_MEAN")
+INCLUDE_RE = re.compile(r"#include\s+<(.+)>")
```
bolshakov-a: Nit:
```suggestion
INCLUDE_RE = re.compile(r"#\s*include\s+<(.+)>")
```
Not sure that such forms are used in Windows headers but they are possible.

bolshakov-a:
In fact, I always wondered what statements like "Header 	winuser.h (include Windows.h)" from MS documentation really mean. Does it mean that `winuser.h` is a private header that should not be included directly? Why is it documented in that case? Or, probably, it is a public-to-public mapping?

itzexpoexpo:
A lot of Windows headers assume that certain macros are defined or headers included before they're included themselves.
An example would be the symbol `HANDLE`, including only `winnt.h` which defines it wouldn't compile, as the header requires macros like `CONST`, `UNALIGNED`, `_AMD64_`/`_X86_` to be defined before it's inclusion.

*`winuser.h` fails with "No target architecture" and complains of undefined types `HWND`, `HDC`, `PCONTEXT` and `PSLIST_HEADER`.*

`Windows.h` automatically defines and includes all the necessary internals while presenting a single header for every Win32 API.
I assume the original header is documented either for tooling or backwards compatibility purposes, but as far as I know, they're not meant to be used directly.

bolshakov-a:
Makes sense, thanks!

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,210 @@
...
+import re
+import sys
+import json
+import argparse
+from pathlib import Path
+from collections.abc import Generator
+
+IF_RE = re.compile(r"#\s*if")
+ENDIF_RE = re.compile(r"#\s*endif")
+LEAN_AND_MEAN_RE = re.compile(r"#\s*ifndef\s+WIN32_LEAN_AND_MEAN")
```
kimgr: Are they consistent, or are there any instances of things like `#if !defined(WIN32_LEAN_AND_MEAN)` in the Windows headers?

itzexpoexpo: I've looked and they are consistent in only using `#ifndef WIN32_LEAN_AND_MEAN`, they do however use the alternate form for other macros, so to future-proof I've added support for it too.

kimgr:
I meant to say: thanks for this! I'll try to get back to a more complete review.

I was hoping to find a package of the Windows Kit for my Ubuntu machine so I could try the mapper, but that doesn't seem to be available.

itzexpoexpo:
No worries!

I've looked into it, and unfortunately the SDK is only available through Windows-only installers.
However, I did find a "mirror" on GitHub: https://github.com/thnetii/windows-sdk-headers.
If you put its contents under `.../Windows Kits/10/Include/10.0.18362.0/` (matching the structure it would have on Windows), you can test the mapper on them.

They are older headers, but I've checked, and the mapper understands them just fine.

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,213 @@
...
+
+    if not include_path.exists():
+        print(f"error: '{include_path}' is not a valid Windows Kits directory")
+
+    available_versions: list[list[str]] = [
+        entry.name.split(".", 4)
+        for entry in include_path.iterdir()
+        if entry.name.startswith("10")
+    ]
+    kit_root = include_path / find_latest_version(available_versions)
```
kimgr: Is there a reason to do all this searching? Why not let the user specify the include path they want to use rather than forcing the latest version?

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,213 @@
...
+    formatting, it always prefers to wrap dicts onto multiple lines.
+
+    Cheat, and use json.dumps for escaping each line.
+    """
+
+    def jsonline(mapping, indent):
+        return (indent * " ") + json.dumps(mapping)
+
+    for name in sorted(include_names):
+        # Regex-escape period and build a regex matching both "" and <>.
```
kimgr: Comment doesn't match what's happening, the below only creates `<>` mappings, right?

On that note, maybe it _should_ allow both?

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,213 @@
...
+    simple way to convince Python's json library to generate a "packed"
+    formatting, it always prefers to wrap dicts onto multiple lines.
+
+    Cheat, and use json.dumps for escaping each line.
+    """
+
+    def jsonline(mapping, indent):
+        return (indent * " ") + json.dumps(mapping)
+
+    for name in sorted(include_names):
```
kimgr: It was already sorted before calling this function

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,213 @@
...
+
+def fill_descendant_includes_of(include_root: Path, header_name: str) -> None:
+    """Recursively find all includes of `header_name`.
+
+    Found includes are added into the `accum_includes` set.
+    """
+    file = find_file_in(include_root, header_name)
+    if file is None:
+        return
+    accum_includes.add(header_name)
```
kimgr: I'm not too happy about the mutable global. It would be nicer to pass the include set in as a mutable parameter.

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
```
kimgr, line 1: Could you mark the file as executable so that makes it into Git metadata? On Linux it would be `chmod +x mapgen/iwyu-mapgen-windowsh.py`, not sure how it's done on Windows.

mapgen/iwyu-mapgen-windowsh.py
itzexpoexpo: I originally thought it'd be more user-friendly, but thinking about it now, forcing a version could indeed cause problems in the future. Removed the search

mapgen/iwyu-mapgen-windowsh.py
itzexpoexpo: I think considering only the `<>` form is more correct since they're system headers. It additionally allows the user to opt-out if they have a header with a colliding name (e.g. a local header named `"consoleapi.h"` sounds plausible).

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,190 @@
...
+    if not kit_path.exists() or not find_file_in(kit_path, "Windows.h"):
+        print(
+            f"error: '{kit_path}' does not exist or does not contain Windows.h",
+        )
+        return 1
+
+    accum_includes: set[str] = set()
+    fill_descendant_includes_of(accum_includes, kit_path, "Windows.h")
+    accum_includes.remove("Windows.h")
+    accum_includes.remove("windows.h")
```
kimgr: I tested this on the windows-sdk-headers checkout you linked to -- and this failed with a KeyError. Would it make sense to use `discard` instead of `remove` to remove-if-exists?

windowsh.imp
```diff
@@ -0,0 +1,150 @@
...
+  {"include": ["<ntstatus.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<oaidl.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<objbase.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<objidl.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<objidlbase.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<ole2.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<oleauto.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<oleidl.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<pmmintrin.h>", "private", "<Windows.h>", "public"]},
+  {"include": ["<poppack.h>", "private", "<Windows.h>", "public"]},
```
kimgr: The `packon.h`, `packoff.h`, `pshpack*.h` and `poppack.h` headers should probably not be mapped -- I think they can be used in user code to do Windows-friendly struct packing.

Raymond Chen talks a bit about them: https://devblogs.microsoft.com/oldnewthing/20241216-00/?p=110645, even though the text mostly explains why they are headers at all.

More importantly, they don't define any symbols, they are just textual includes to "execute some pragmas".

I think I'd put in a `TEXTUAL_HEADERS = [ ... ]` list and use it to ignore such includes.

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+
+    if_count = 0
+    lean_mean_idx: int | None = None
+
+    with headerpath.open("r", encoding="utf-8") as fobj:
+        for line in fobj.readlines():
+            if IF_RE.search(line) is not None:
+                # Check if we entered a WIN32_LEAN_AND_MEAN block
+                if (
```
kimgr: I find the parenthesized if-statements quite hard to read. This works better, IMHO:
```
                # Check if we entered a WIN32_LEAN_AND_MEAN block
                if LEAN_AND_MEAN1_RE.search(line) is not None or \
                   LEAN_AND_MEAN2_RE.search(line) is not None:
                    lean_mean_idx = if_count
```

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,190 @@
...
+
+                # Check if we escaped a WIN32_LEAN_AND_MEAN block
+                if if_count == lean_mean_idx:
+                    lean_mean_idx = None
+
+                continue
+
+            if lean_mean_idx is None:
+                include_match = INCLUDE_RE.search(line)
+                if (
```
kimgr: Same for the parentheses, try line-continuation instead

mapgen/iwyu-mapgen-windowsh.py
bolshakov-a: Maybe, `is not None` is just redundant?

kimgr: Yeah, I also prefer to rely on ~standard~ conventional Python truthyness or falsyness.

itzexpoexpo: I think using `is not None` is completely valid here, even if a bit redundant. It emphasizes the condition being "the thing wasn't found" instead of "the result of this call was falsy". It's the [recommended best practice](https://peps.python.org/pep-0008/#programming-recommendations) in modern Python.

firewave:
> I was hoping to find a package of the Windows Kit for my Ubuntu machine so I could try the mapper, but that doesn't seem to be available.

Wine and MinGW/MSYS also offer the Windows headers. That might be feasible for local testing (and should probably also be supported). The packages are `libwine-dev` and `mingw-w64-common`.

itzexpoexpo:
> Wine and MinGW/MSYS also offer the Windows headers. That might be feasible for local testing (and should probably also be supported). The packages are libwine-dev and mingw-w64-common.

Thanks for the reminder. After the review changes this was trivial to implement. Latest commit tested as working under Gentoo with MinGW headers.

Output was missing the intrinsic headers (`zmmintrin.h` and others) like the older mirror headers did, maybe it'd be better to exclude these from the mapping too?

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,197 @@
...
+    # Textual headers
+    "packon.h",
+    "packoff.h",
+    "pshpack1.h",
+    "pshpack2.h",
+    "pshpack4.h",
+    "pshpack8.h",
+    "poppack.h",
+    # MinGW32 headers
+    "_mingw.h",
```
kimgr: Oh, but shouldn't these be mapped to Windows.h? The leading underscore looks like a hint that these are private to mingw. I pulled the package down and looked at the files, and it mostly looks like a bunch of support macros. But I don't know how they're used in practice.

kimgr:
> Output was missing the intrinsic headers (zmmintrin.h and others) like the older mirror headers did, maybe it'd be better to exclude these from the mapping too?

Good call on the intrin headers, I think those are often used directly as well. So I agree that they should be "excluded", i.e. not mapped to `Windows.h`

mapgen/iwyu-mapgen-windowsh.py
itzexpoexpo: From what I understand, some of the macros like `__cdecl` or `__volatile__` are treated as keywords by MSVC, and MinGW chooses to imitate it by making them macros in these headers.
This makes me think they could possibly be used standalone without `Windows.h` - on the other hand your point is also correct, the underscore prefix suggests they're not meant to be used by the user. I couldn't find any information on using the header directly either, so I re-added them to the mapping for now.

It's possible (and likely) this was done by the MinGW developers to just avoid editing the Win32 headers too much, and user-written code should use MinGW-specific alternatives. I think their headers are taken from Windows directly, as opposed to a re-implementation like Wine does.

mapgen/iwyu-mapgen-windowsh.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+def fill_descendant_includes_of(
+    into_set: set[str],
+    include_root: Path,
+    header_name: str,
+) -> None:
+    """Recursively find all includes of `header_name`.
+
+    Found includes are added into the `accum_includes` set.
```
kimgr: `into_set`

kimgr:
One more little docstring nit, then LGTM!

kimgr:
I'm actually a little doubtful whether we should include the generated `windowsh.imp` here. People will always have better results if they generate it from scratch in their local environment, and we avoid getting bugfixes for the generated .imp file rather than for the generator itself.

kimgr:
@itzexpoexpo Alright, nice! LGTM. Do you have anything else, or should I merge it?

itzexpoexpo:
You can merge it. Thanks!

kimgr:
Squashed and merged. Thank you!

Issue #1863, opened 2025-11-06
<format> suggested for `std::vector` in C++20


hebasto:
MRE:
```c++
#include <chrono>  // critical for this MRE
#include <vector>
using namespace std::chrono_literals;
void f(std::vector<int>& v) { v.resize(42); }
```

Output:
```
$ include-what-you-use -std=c++20 mre.cpp 
/usr/include/c++/15/bits/unicode-data.h:1:1: warning: no private include name for @headername mapping

mre.cpp should add these lines:
#include <format>  // for vector

mre.cpp should remove these lines:

The full include-list for mre.cpp:
#include <chrono>  // for chrono_literals
#include <format>  // for vector
#include <vector>  // for vector
---
```

Tools:
```
$ include-what-you-use --version
include-what-you-use 0.25 (git:791e69e) based on Debian clang version 21.1.4 (8)
$ clang-21 -v
Debian clang version 21.1.4 (8)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-21/bin
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/15
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/15
Candidate multilib: .;@m64
Selected multilib: .;@m64
```

bolshakov-a:
I don't currently have libstdc++-15 installed and can reproduce this neither on 12 nor on 13 nor on 14. However, I suppose this is a duplicate of #1763 and should be fixed by the explicit mapping:
```patch
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -339,6 +339,8 @@ const IncludeMapEntry stdlib_cxx_symbol_map[] = {
   { "std::size_t", kPrivate, "<ctime>", kPublic },
   { "std::size_t", kPrivate, "<cuchar>", kPublic },
   { "std::size_t", kPrivate, "<cwchar>", kPublic },
+
+  { "std::vector", kPrivate, "<vector>", kPublic },
 };
```

bolshakov-a:
Seems like after 2d4e4f4b9eedd773b47b5fbc824a92c992364a5a, this should be reproduced with all the `vector` template arguments explicitly written:
```cpp
void f(std::vector<int, std::allocator<int>>& v) { v.resize(42); }
```
#1874 fixes this on my machine with libstdc++-15.2.0.

kimgr:
[...]

Issue #785, opened 2020-04-17
Mapping for all symbols defined in the C and C++ standard


amurzeau:
Hi,

I'm wondering, why all defined symbols by standards are not mapped to their header instead of trying to map private header to public ones ?
Like for example strcmp symbol to string.h ?
I've dumped the C18 latest draft standard into a mapping file and this use 68KB.
Maybe for C++ it will be too large, but at least perfectly accurate for all compilers.

See here for what I've come up: [C18.imp.txt](https://github.com/include-what-you-use/include-what-you-use/files/4490349/C18.imp.txt)

The C standard happen to have a Library summary in annex B which list all defined functions and macros per headers. I used that to create that file.

Is this something IWYU would want for C/C++ in its files ?
I've not finished on the C++ side, but I'm working to try to retrieve symbols from the C++ latest draft in latex format from http://github.com/cplusplus/draft.

kimgr:
Nice! I've thought about this before (alternatively, parsing http://en.cppreference.com), but never got around to building anything.

I think the reason it wasn't originally done this way is because it was easier to build mappings from header relationships than from actually locating symbol definitions (would require parsing C++ again to do it from standard library headers). Most of the original mappings were built with shell one-liners.

I wonder if we could distill this to a Python script we could put in the IWYU repo? And maybe combine with @trixirt's long dormant idea to generate IWYU built-in mappings from mapping files: https://github.com/include-what-you-use/include-what-you-use/pull/671. I've been meaning to pick this up some time.

amurzeau:
Indeed parsing cppreference is probably a better idea, as I had to retrieve the text from the C18 PDF which isn't script friendly. But https://en.cppreference.com/w/cpp/symbol_index is interesting as it sum up all symbols.
When comparing the file I've (C18.imp) and cppreference, I found some discrepancies (2 typos and 2 missing functions) that I've fixed on cppreference side.

kimgr:
Cool! I would've expected the stdc++ latex to be easier to parse, but if there's a big page with everything in it, that makes for a good target!

amurzeau:
Yes I will see if what I can do, but parsing both can be a way to double check that cppreference doesn't contains errors (like typos) as in the C symbol index ^^
Also, the stdc++ latex won't have removing functions that were available in previous standards, so using cppreference is a good idea. (like `gets` in C, which isn't in C18).

amurzeau:
[...]

PR #1874, opened 2025-12-20
Add C++ standard library symbol mapping

### LINKED ISSUE #1616: false positive: IWYU sometimes suggests replacing `<utility>` with `<iterator>` for std::pair
### Example
```cpp
#include <utility>
#include <span>

template <typename T>
std::pair<T, T> foo(std::span<int>);
```

### Steps to reproduce
```console
$ include-what-you-use -std=c++20 repr.cpp

repr.cpp should add these lines:
#include <iterator>  // for pair

repr.cpp should remove these lines:
- #include <utility>  // lines 1-1

The full include-list for repr.cpp:
#include <iterator>  // for pair
#include <span>      // for span
---
```

### Versions
include-what-you-use 0.23 (git:5ac52ca) based on clang version 20.0.0git
GNU libstdc++ version 14.2.1

bolshakov-a:
The mapping has been generated from C++23 final draft (N4950) LaTeX sources downloaded from https://github.com/cplusplus/draft repository.

Headers described in [input.output] clause are skipped because declarations from `<iosfwd>` cannot at now be marked as forward-declarations in the mapping. By the same reason, some other classes are explicitly excluded.

Functions overloaded in different headers, as well as operators, are currently excluded from the mapping so as not to cause any regression because IWYU cannot currently distinguish them in mappings. Probably, they could be included when the mapping is complete. In that case, IWYU would at least not to recommend to remove a required header if it is already included in a user's source code.

Preprocessor macro names have been excluded by a somewhat similar reason: so that IWYU does not suggest to replace C headers defining them with C++ ones.

Formally, this fixes #1616, #1763, and probably #1863. However, similar bugs may appear until all the mapping has been written.


mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,461 @@
...
+            return self.act_on_operator()
+        if s.startswith(TEMPLATE_KWD, cp):
+            return self.act_on_template()
+        if s.startswith(DECLTYPE_KWD, cp):
+            self.skip_decltype()
+            return self.act_on_identifier_or_kwd()
+
+        id = self.read_qualified_id()
+        cp = self.__cur_pos
+        if s[cp] in '=:;{(':
```
bolshakov-a: Not sure how readable it is.

bolshakov-a:
I accidentally pushed the complete mapping, and a bug test showed that it would fix #1616.

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,461 @@
...
+            self.skip_till_semicolon()
+            return
+
+        assert self.__syn[self.__cur_pos] == '{'
+        self.__cur_pos += 1
+        self.skip_spaces()
+        while self.__syn[self.__cur_pos] != '}':
+            self.parse_top_level()
+            self.skip_spaces()
+        del self.__namespace_stack[num_stack_items_at_enter:]
```
bolshakov-a: Maybe, add a context manager class for this?

std_symbol_map.inc
```diff
@@ -0,0 +1,421 @@
...
+{ "std::pmr", kPrivate, "<list>", kPublic },
+{ "std::pmr::list", kPrivate, "<list>", kPublic },
+{ "std", kPrivate, "<vector>", kPublic },
+{ "std::vector", kPrivate, "<vector>", kPublic },
+{ "std::pmr", kPrivate, "<vector>", kPublic },
+{ "std::pmr::vector", kPrivate, "<vector>", kPublic },
+{ "std::vector<bool, :0>", kPrivate, "<vector>", kPublic },
+{ "std::hash", kPrivate, "<vector>", kPublic },
+{ "std::hash<std::vector<bool, :0>>", kPrivate, "<vector>", kPublic },
+{ "std::formatter<:0, :1>", kPrivate, "<vector>", kPublic },
```
bolshakov-a: Probably, this should be excluded too. See https://github.com/include-what-you-use/include-what-you-use/pull/1869#issuecomment-3678116207.

bolshakov-a: Done.

bolshakov-a:
I've excluded `std::hash` primary template as well until all the headers declaring them are included in the mapping.

kimgr:
Very cool! This seems like exactly the right source for symbol mappings.

std_symbol_map.inc
```diff
@@ -0,0 +1,417 @@
+{ "std", kPrivate, "<algorithm>", kPublic },
```
kimgr, line 2: It seems like a shame to attribute `namespace std;` to all of these expensive-to-include headers, but I guess it would be very unlikely that someone has;
```
using namespace std;
```
and nothing else from `std` in their source code. I think in that unlikely case, IWYU would force an include of `<algorithm>`, since it's first among the mappings.

Off the top of my head, I don't think we need to do anything about this, seems good enough that `std` is provided by all standard headers.

std_symbol_map.inc
```diff
@@ -0,0 +1,417 @@
...
+{ "std::find_if", kPrivate, "<algorithm>", kPublic },
+{ "std::find_if_not", kPrivate, "<algorithm>", kPublic },
+{ "std::ranges::find", kPrivate, "<algorithm>", kPublic },
+{ "std::ranges::find_if", kPrivate, "<algorithm>", kPublic },
+{ "std::ranges::find_if_not", kPrivate, "<algorithm>", kPublic },
+{ "std::ranges::find_last", kPrivate, "<algorithm>", kPublic },
+{ "std::ranges::find_last_if", kPrivate, "<algorithm>", kPublic },
+{ "std::ranges::find_last_if_not", kPrivate, "<algorithm>", kPublic },
+{ "std::find_end", kPrivate, "<algorithm>", kPublic },
+{ "std::ranges::find_end", kPrivate, "<algorithm>", kPublic },
```
kimgr: Is it possible to sort by (header, qualified-name), so symbols show up in alphabetic order header-by-header?

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,470 @@
...
+    .*?                     # Search non-greedily.
+    \\begin\{codeblock\}
+    (?P<code>.*?)
+    \\end\{codeblock\}''',
+    re.DOTALL | re.VERBOSE)
+
+# Functions overloaded in different headers may cause problems because IWYU
+# doesn't consider function parameters in mappings. Hence they are excluded
+# for a while.
+OVERLOADED_FUNCS = {'std::abs', 'std::acos', 'std::acosh', 'std::asin',
```
kimgr: I usually prefer for listings like this to be formatted like:
```
OVERLOADED_FUNCS = {
  'std::abs',
  ...
  'std::vprint_unicode',
}
```
to make it easier to scan for elements (and with trailing comma to minimize the diff when adding/removing).

But this set is large enough that it would eat up a lot of vertical space, so I think it's probably better as-is.

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,470 @@
...
+    'std::sinh', 'std::sqrt', 'std::swap', 'std::tan', 'std::tanh',
+    'std::to_string', 'std::tolower', 'std::toupper', 'std::vprint_nonunicode',
+    'std::vprint_unicode'}
+
+# Some classes and template specializations are excuded as well.
+# The specializations defined in different headers differ only in C++20
+# constraints, and hence cannot be distinguished in the mapping. The classes are
+# provided by different headers, and some of the headers provide only forward-
+# declarations, therefore some additional consideration is required.
+DUPLICATED_CLASSES = {'std::allocator', 'std::atomic',
```
kimgr: This one is short enough that you could do the element-by-line formatting instead.

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,470 @@
...
+    'std::formatter<:0, :1>', 'std::hash', 'std::tm', 'std::tuple'}
+
+class Parser:
+    """ Parses a standard library header synopsis from a LaTeX file and fills in
+        found_names list. This stupid pseudo-C++ parser has been written with
+        a lot of assumptions about the synopsis syntax. Use carefully and check
+        the output!
+    """
+    def __init__(self, syn):
+        self.__syn = syn
```
kimgr: It took me a while to understand that `syn` was short for `synopsis`. How about just `text`/`__text`?

kimgr: (I think of parsers in general as turning text into some more desirable data structure)

mapgen/iwyu-mapgen-std-symbol.py
kimgr: I think that would be a good idea!

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,470 @@
...
+        elif s[cp:cp+3] == '...':
+            self.__cur_pos += 3
+        elif s[cp:cp+2] == '::':
+            self.__cur_pos += 2
+        self.act_on_tpl_param_id_or_kwd()
+
+    def skip_requires_clause(self):
+        self.__cur_pos += len(REQUIRES_KWD)
+        self.skip_primary_expr()
+        while self.__syn.startswith('&&', self.__cur_pos)\
```
kimgr: Space before line continuation char, please

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,470 @@
...
+            return
+        if id not in self.found_names:
+            self.found_names.append(id)
+
+def process_synopsis(headername, syn):
+    p = Parser(syn)
+    p.parse()
+    for symbol in p.found_names:
+        escaped = symbol.replace('"', r'\"')
+        print(f'{{ "{escaped}", kPrivate, "<{headername}>", kPublic }},')
```
kimgr: It might be nice to allow both `.imp` format and inline C++ format, like the libstdcxx mapper: https://github.com/include-what-you-use/include-what-you-use/blob/b00eeed3497688a1e6cc0a23dbaf14840a1c2fe2/mapgen/iwyu-mapgen-libstdcxx.py#L271

The idea I had around this was to allow users regenerate mappings locally and replace the internal mappings with something like `-Xiwyu -no_internal_mappings -Xiwyu --mapping_file=mything.imp` without having to rebuild IWYU.

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,470 @@
...
+            # Avoid names containing TeX commands.
+            return
+        if id not in self.found_names:
+            self.found_names.append(id)
+
+def process_synopsis(headername, syn):
+    p = Parser(syn)
+    p.parse()
+    for symbol in p.found_names:
+        escaped = symbol.replace('"', r'\"')
```
kimgr: Could you print a little header comment that this is generated code? Doesn't need to be as fancy as https://github.com/include-what-you-use/include-what-you-use/blob/b00eeed3497688a1e6cc0a23dbaf14840a1c2fe2/mapgen/iwyu-mapgen-libstdcxx.py#L157, but just a little something to indicate that it's generated code, and where it came from.

mapgen/iwyu-mapgen-std-symbol.py
kimgr: It looks like a chaotic angry emoji, but I think it's pretty clear what's being tested. I like literals.

bolshakov-a: It turns out that the synopsis of `<iterator>` header contains some functions having a body (and hence not ended with a semicolon), so the function case (`(`) should be handled separately. I will probably do it in the next PR.

std_symbol_map.inc
bolshakov-a: If `std` is attributed to all the headers, IWYU accepts e.g. such a code without complaining:
```cpp
#include <vector>
using namespace std;
```
which I thought is correct.

In fact, I added namespaces to the mapping to make the name lookup work, so that e.g. lookup for `filesystem` in the template argument:
```cpp
template<> struct hash<filesystem::path>;
```
finds `std::filesystem` namespace added to `found_names`, and the script adds the full prefix. I decided not to introduce a separate list for namespaces, because adding them to the output might be useful. However, `std` being the top-level namespace doesn't help with name lookup, hence may be excluded from the output if desirable.

std_symbol_map.inc
bolshakov-a: Of course, but it is easier to review (at least by myself) and compare the output with the actual text of the standard when the order resembles the order in the synopses.

mapgen/iwyu-mapgen-std-symbol.py
bolshakov-a: I've decided not to postpone and pushed the fix.

std_symbol_map.inc
kimgr: Sorry, yes, I was thinking out loud in the comment. I think it's good as-is.

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,522 @@
...
+    'std::isblank', 'std::iscntrl', 'std::isdigit', 'std::isgraph',
+    'std::islower', 'std::isprint', 'std::ispunct', 'std::isspace',
+    'std::isupper', 'std::isxdigit', 'std::log', 'std::log10',
+    'std::make_error_code', 'std::make_error_condition', 'std::move',
+    'std::pow', 'std::print', 'std::println', 'std::remove', 'std::sin',
+    'std::sinh', 'std::sqrt', 'std::swap', 'std::tan', 'std::tanh',
+    'std::to_string', 'std::tolower', 'std::toupper', 'std::vprint_nonunicode',
+    'std::vprint_unicode'}
+
+# Some classes and template specializations are excuded as well.
```
kimgr: typo: excuded -> excluded

kimgr:
This looks good to me! I haven't looked into all the details, but the output looks very comprehensive. Do you want to do the full generation before merging?

Also, could you expand on the docs here to include the steps for how to run the script: https://github.com/include-what-you-use/include-what-you-use/blob/fd49ecbad794ecd7716a0d5d9c29e8620d597d1c/docs/IWYUMappings.md#generating-mapping-files?

std_symbol_map.inc
kimgr: I see, that makes sense.

bolshakov-a:
> Do you want to do the full generation before merging?

I think it is Ok to merge this in parts.

bolshakov-a:
Well, I've added the rest except the headers from [input.output] and deleted the bug test for #1616. Not sure if any test should be written.

bolshakov-a:
> Not sure if any test should be written.

I've added a small test about `std::pair`.

bolshakov-a:
It turns out that the standard allows some variable templates (e.g. `std::ranges::enable_view`) to be specialized by users. So, #1875 should be merged prior to this to avoid a regression.

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,539 @@
...
+    'std::formatter<:0, :1>',
+    'std::hash',
+    'std::tm',
+    'std::tuple',
+}
+
+ERRONEOUS_SYMBOL = 'std::formatter<std::chrono::hh_mm_ss<duration<:0, :1>>, :2>'
+CORRECT_SYMBOL = \
+    'std::formatter<std::chrono::hh_mm_ss<std::chrono::duration<:0, :1>>, :2>'
+
```
kimgr: I would prefer to inline these literals since they're only used in one place, and can't be extended to cover more symbols. But I see now that they're very long, so they probably won't fit line width inline.

Maybe make it "table-driven" instead so it's at least extensible:
```
ERRONEOUS_SYMBOLS = dict(
    ('std::formatter<std::chrono::hh_mm_ss<duration<:0, :1>>, :2>':
     'std::formatter<std::chrono::hh_mm_ss<std::chrono::duration<:0, :1>>, :2>'),
)

# and when used
id = ERRONEOUS_SYMBOLS.get(id, id)
```

But yeah... It looks a bit noisy still with the long symbol names.

docs/IWYUMappings.md
```diff
@@ -273,3 +273,13 @@ version and target match, but they should port pretty well.
 
 There is no strong policy for updating the internal mappings, we try to use a
 mainstream target and a middle-aged libstdc++ version.
+
+Another one, `mapgen/iwyu-mapgen-std-symbol.py`, is used to generate
+`std_symbol_map.inc` file containing the C++ standard library symbol mapping.
+That file is automatically included into `iwyu_include_picker.cc` by
+a `#include` directive. To regenerate it, you should have a C++ standard LaTeX
```
kimgr, line 280: I think we can include the `git clone https://github.com/cplusplus/draft.git` command here, since it's public and stable.

bolshakov-a:
I've removed some stuff that became redundant. I have kept the mapping from `std::size_t` to `<cstddef>` so that `<cstddef>` remains its canonical header. Hence, this mapping entry is duplicated at now, in fact. By the same reason, I've moved `#include "std_symbol_map.inc"` up so that `std::char_traits -> <string>` mapping is kept prior to the mappings to `<ostream>` and `<istream>`.

I also think that `std_symbol_map.inc` has indeed to be sorted by symbol names, so that such rearrangements to place canonical mappings before others can easily be done in the script. But, probably, it should be a separate PR? Or, it would be better to make it right now to minimize changes stored in the repo history?

UPD: I have thought about implementation, and it seems that the canonical mapping will be present as an explicit list in the script. That list may just be printed prior to the rest. Sorting would not help.

bolshakov-a:
@kimgr, are you sure that the script should have the ability to produce `.imp` file? I'm asking because I want to add a flag to denote forward-declaration mappings for `<iosfwd>` (probably as a new enumerator of `IncludeVisibility` or a replacement of it for symbol mappings). If it is exposed to users, we have to answer the questions about how it should work with user's types, which is not obvious.

kimgr:
Cool, I think the current version looks quite nice.

> I also think that std_symbol_map.inc has indeed to be sorted by symbol names, so that such rearrangements to place canonical mappings before others can easily be done in the script. But, probably, it should be a separate PR? Or, it would be better to make it right now to minimize changes stored in the repo history?

If it's not a very complex and error-prone thing to do, I think we might as well fix it up-front.

> are you sure that the script should have the ability to produce .imp file? I'm asking because I want to add a flag to denote forward-declaration mappings for <iosfwd> (probably as a new enumerator of IncludeVisibility or a replacement of it for symbol mappings). If it is exposed to users, we have to answer the questions about how it should work with user's types, which is not obvious.

It would be nice, since it's the only way for users to tweak IWYU mapping policy to their local environment. Also your new internal mappings will already be translated to .imp by `-Xiwyu --export_mappings`, so I guess you can't escape the challenge...

Yeah, I think that's the right way to think about it: `.imp` and `IncludeMapEntry` records are just two formats of the same data model. If we will have special rules for forward-only mappings, I think they need to be described/enforced independent of data format.

std_symbol_map.inc
```diff
@@ -0,0 +1,2374 @@
...
+{ "std::mblen", kPrivate, "<cstdlib>", kPublic },
+{ "std::mbrlen", kPrivate, "<cwchar>", kPublic },
+{ "std::mbrtoc16", kPrivate, "<cuchar>", kPublic },
+{ "std::mbrtoc32", kPrivate, "<cuchar>", kPublic },
+{ "std::mbrtoc8", kPrivate, "<cuchar>", kPublic },
+{ "std::mbrtowc", kPrivate, "<cwchar>", kPublic },
+{ "std::mbsinit", kPrivate, "<cwchar>", kPublic },
+{ "std::mbsrtowcs", kPrivate, "<cwchar>", kPublic },
+{ "std::mbstate_t", kPrivate, "<cuchar>", kPublic },
+{ "std::mbstate_t", kPrivate, "<cwchar>", kPublic },
```
bolshakov-a, line 1268: I'm not sure about this. Cppreference also [places](https://en.cppreference.com/w/cpp/string/multibyte/mbstate_t.html) `<cuchar>` before `<cwchar>` for `mbstate_t`. However, [it is noted](https://eel.is/c++draft/char.traits.specializations.char#1) in [char.traits.specializations.char] that

> The type `mbstate_t` is defined in [`<cwchar>`](https://eel.is/c++draft/cwchar.syn#header:%3ccwchar%3e)

So maybe, a canonical header list should be added to the script right now...

kimgr: What would that list look like? It would have to be hand-maintained for every symbol?

kimgr: Or it can be parsed out of the standards text, and then used to order in output?

bolshakov-a: Canonical header notion is derived from plain English text in the standard, and I don't want to parse it. So, the list (or, rather, dictionary) may look like this:
```python
canonical_headers = {
  'std::size_t': 'cstddef',
  'std::mbstate_t': 'cwchar',
  'std::char_traits': 'string',
  'std::char_traits<char>': 'string',
  ...
}
```
and should be maintained manually.

kimgr: But that mapping will essentially list all symbols, right? Or only the special ones where there's ambiguity? It seems like it's going to turn into quite a big hand-maintained thing.

bolshakov-a: No, why should it contain all the symbols? It should contain only those symbols which are defined in different headers, and for which the standard gives a preference to one of the headers. I cannot currently remember something else but `size_t`, `mbstate_t`, `char_traits`, and maybe some names from `DUPLICATED_CLASSES`.

bolshakov-a: There are also `std::tuple_size` and `std::tuple_element`. I got stupid and thought that because the primary templates are not defined, IWYU should always suggest the correct header for a specialization. But actually, when a user _defines its own specialization_, it probably wants `<tuple>` to be suggested for his specialization header and not `<array>`.

kimgr: I see. I thought there was always a canonical header that we would need to list. But it's only in the cases where symbols are defined in different headers, then. That sounds reasonable.

bolshakov-a: Besides, there are `sorted_equivalent`, `sorted_equivalent_t`, `sorted_unique`, and `sorted_unique_t`, which are defined both in `<flat_map>` and in `<flat_set>`.

For `wint_t`, I don't see which of `<cwchar>` and `<cwctype>` headers should be preferred.

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,564 @@
...
+        with open(path, 'r') as f:
+            symbols_and_headernames += process_tex_file(f.read())
+
+    for symbol in sorted({item[0] for item in symbols_and_headernames}):
+        escaped = symbol.replace('"', r'\"')
+        if symbol in CANONICAL_HEADERS:
+            print_line(escaped, CANONICAL_HEADERS[symbol], lang)
+        # Print all the rest of symbol headers except the canonical one.
+        headers = sorted(item[1] for item in symbols_and_headernames \
+            if item[0] == symbol and item[1] != CANONICAL_HEADERS.get(symbol))
```
bolshakov-a: Yeah, this is O(N^2) (more likely, even multiplied by a logarithm, looking closer on the outer `for`), but I'm not sure if some fix is justified for this script which still runs in less than a second on my machine.

firewave:
CC @alejandro-colomar as this essentially supersedes #1536.

std_symbol_map.inc
alejandro-colomar: ISO C documents <wchar.h> as the canonical header for `wint_t`.

alejandro-colomar: If you have doubts about types from C, you can check the manual pages; I've documented most of them, and have differentiated the canonical header from others.

alejandro-colomar:
```
$ man -w wint_t | xargs mansect 'SYNOPSIS|NOTES' | man /dev/stdin | cat
wint_t(3type)                                                     wint_t(3type)

SYNOPSIS
     #include <wchar.h>

     typedef /* ... */  wint_t;

     #define WEOF        /* ... */

     #include <stdint.h>

     #define WINT_WIDTH  /* ... */
     #define WINT_MAX    /* ... */
     #define WINT_MIN    /* ... */

NOTES
     The following header also provides wint_t and WEOF: <wctype.h>.

Linux man‐pages 6.16‐434‐gaf802... 2025‐12‐25                     wint_t(3type)
```

alejandro-colomar:
> CC @alejandro-colomar as this essentially supersedes #1536.

@firewave  Thanks!

@bolshakov-a 

Would it be possible to have some "strict" flag that ignores non-canonical headers?  I'd like to require <stddef.h> in my programs if they use NULL, even if <string.h> is included.

Could you clarify what this PR achieves?  I assume it will partially revert what I did with the reduction of mappings, and that it will now be fine to include <cstring> for std::size_t.  I would prefer to not do that, or at least allow users to be stricter and enforce canonical headers only.

std_symbol_map.inc
bolshakov-a: > ISO C documents <wchar.h> as the canonical header for `wint_t`.

Maybe, you have made this conclusion just because `<wchar.h>` is documented before `<wctype.h>` in the standard? Anyway, in the proposed mapping, `<cwchar>` is placed before `<cwctype>`, according to the alphabetical order.

bolshakov-a:
> Would it be possible to have some "strict" flag that ignores non-canonical headers?

It would, but I currently don't plan to implement it because I'd prefer not to include `<cstddef>` if some other header that should provide it is already included in my source code file. Probably, @bungeman might express his opinion as well because he originally added the mappings for `size_t`.

> Could you clarify what this PR achieves? I assume it will partially revert what I did with the reduction of mappings, and that it will now be fine to include for std::size_t.

The main goal of this PR is to provide the standard mapping of the C++ standard library symbols (declared in `std` namespace). It not only fixes some bugs related to forward-declarations but also should improve support of MS STL which usually defines symbols in private headers (`<x...>`), as well as of any other possible standard-conforming implementation. Note that it should not change the work with the C standard library.

As for `std::size_t`, this PR changes nothing relatively to the present master branch. For `mbstate_t` and `wint_t` from `std` namespace, there was no mapping at all.

mapgen/iwyu-mapgen-std-symbol.py
kimgr: Actually, could you line-break after the leading `{` and before the closing `}`?
```
OVERLOADED_FUNCS = {
    'std:abs', 'std::acos', ...,
    'std::asinh', 'std::atan', ...
    ...,
    'std::vprint_unicode',
}
```

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,564 @@
...
+    'std::basic_common_reference<:0, :1, :2, :3>',
+    'std::compare_three_way',
+    'std::formatter<:0, :1>',
+    'std::hash',
+    'std::tm',
+    'std::tuple',
+}
+
+ERRONEOUS_SYMBOLS = dict(
+    (('std::formatter<std::chrono::hh_mm_ss<duration<:0, :1>>, :2>',
```
kimgr: Let's move the leading paren to the `dict(`, i.e.

```
ERRO...OLS = dict((
    ('std::formatter<std::chrono::hh_mm_ss<duration<:0, :1>>, :2>',
    'std::formatter<std::chrono::hh_mm_ss<std::chrono::duration<:0, :1>>, :2>'),
))
```

... and possibly align the second line by the quotes -- I think that's what Python editors prefer -- if you can live with overflowing the 80-col limit (I think it's fine in this case, since we don't have any control over the name length)

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,564 @@
...
+def print_content(std_source_path, lang):
+    symbols_and_headernames = []
+    for path in sorted(glob.glob(os.path.join(std_source_path, '*.tex'))):
+        if path.endswith('iostreams.tex'):
+            # Skip until <iosfwd> can be handled correctly.
+            continue
+        with open(path, 'r') as f:
+            symbols_and_headernames += process_tex_file(f.read())
+
+    for symbol in sorted({item[0] for item in symbols_and_headernames}):
```
kimgr: The `item[N]` unpacking makes this (and the canonical headers check below) a bit hard to understand...

I wonder if it would be easier if you represented `symbols_and_headernames` as a dict from `symbol: [header, header, header...]`.

Then you could say:
```
for symbol, headers in sorted(symbols_and_headernames.items()):
```
here, and
```
canonical_header = CANONICAL_HEADERS.get(symbol)
if canonical_header:
    print_line(escaped, canonical_header, lang)

for headername in headers:
    if headername != canonical_header:
        print_line(escaped, headername, lang)
```
to do the filtering.

But I'm not sure if I'm missing some subtlety in the sorting there...?

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,564 @@
...
+        escaped = symbol.replace('"', r'\"')
```
kimgr: Could/should the escaping move into `print_line`?

std_symbol_map.inc
alejandro-colomar: Nope; that conclusion come from the fact that 7.32.1 (<wctype.h>) in C23 says:

```
The types declared are wint_t described in 7.31.1;
```

where 7.31.1 is <wchar.h>.

In general, the C standard defines identifiers in one header, and then the other headers declare it as described in the canonical header.

See <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf#subsection.7.32.1>.

alejandro-colomar:
> As for `std::size_t`, this PR changes nothing relatively to the present master branch. For `mbstate_t` and `wint_t` from `std` namespace, there was no mapping at all.

It doesn't change anything because we haven't merged the patches for size_t and std::size_t yet, but we were planning to do it.  We did the same with most other standard types, where we removed all non-canonical headers from the mappings.

mapgen/iwyu-mapgen-std-symbol.py
kimgr: Ah, I see you would be re-escaping the symbol for every (symbol, header) pair if you do. I don't have a strong opinion either way.

std_symbol_map.inc
bolshakov-a: > The types declared are wint_t described in 7.31.1;

Looks reasonable, so as just not to repeat itself. It would be good to hear the opinion of @AaronBallman or some other C standard committee member.

mapgen/iwyu-mapgen-std-symbol.py
bolshakov-a: I also think about that. In that case, `+=` is to be replaced by a hand-written merging function, or a function to convert a list to a dictionary should be introduced. What would be better here? Or, maybe, there is some appropriate function in the Python standard library?

mapgen/iwyu-mapgen-std-symbol.py
bolshakov-a: Yeah, `print_line` is a more appropriate place for escaping.

bolshakov-a:
@alejandro-colomar, I think it would be wrong to incorporate your patch into mine, even if I agreed with it.

std_symbol_map.inc
alejandro-colomar: I'm actually member of the C Committee.  :-)

bolshakov-a: Ah, that's good! However, I suppose you didn't take part in the standardization of the majority of the standard library, so another opinions would still be appreciated.

mapgen/iwyu-mapgen-std-symbol.py
kimgr: I can't think of anything off-hand. `dict.update` does dict merging, but it doesn't have enough smarts to merge list values.

If I'm reading it right, `process_tex_file` now returns a `list(tuple(symbol,header))`, but `header` is always the same? I wonder if it would be simpler to return a `tuple(header, list(symbol))` and then use that to incrementally build a `dict(symbol, list(header))`, something like:

```
headers_by_symbol = {}
for path in sorted(glob.glob(os.path.join(std_source_path, '*.tex'))):
    # ...
    with open(path, 'r') as f:
        headername, symbols = process_tex_file(f.read())
        for symbol in symbols:
            headers_by_symbol.setdefault(symbol, []).append(headername)
```
The same list-to-dict transform technique should work for your `list(tuple(...))` shape too.

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,566 @@
...
+
+def process_tex_file(tex_file):
+    res = []
+    for m in SYNOPSIS.finditer(tex_file):
+        symbols = process_synopsis(m.group('code'))
+        res += zip(symbols, repeat(m.group('headername')))
+    return res
+
+def print_line(symbol, headername, lang):
+    escaped = symbol.replace('"', r'\"')
```
kimgr, line 522: Hmm, actually, when would the symbol ever have quotes in it?

bolshakov-a: It's for user-defined literal operators (like [`operator""s`](https://www.en.cppreference.com/w/cpp/string/basic_string/operator%2522%2522s.html)), which are explicitly excluded from the mapping at now.

mapgen/iwyu-mapgen-std-symbol.py
bolshakov-a: > If I'm reading it right, process_tex_file now returns a list(tuple(symbol,header)), but header is always the same?

No, not necessarily the same. One TeX file corresponds to one standard clause and may contain several header synopses, e.g. `<type_traits>` and `<ratio>` in [`meta.tex`](https://timsong-cpp.github.io/cppwp/n4950/#meta).

bolshakov-a: Maybe, just print all the symbols having canonical headers on top of the mapping file, and then
```python
for symbol, header in symbols_and_headernames:
  if header != CANONICAL_HEADERS.get(symbol):
    print_line(symbol, header, lang)
```
?

bolshakov-a: I've succeeded with your suggestion modified a little.

kimgr: Nice, looks good to me!

kimgr:
I'm happy with the patch as-is now. If anything, maybe remove `Part 1` from the commit message, it already explains the limitations quite clearly.

@alejandro-colomar @bolshakov-a I wonder if we'll ever reach consensus on the only-canonical vs. laissez-faire modes. I'm tempted to pretend #1536 doesn't exist, since it's been dormant for a year, and having auto-generated complete symbol mappings for C++ is a major improvement for users.

What if we reduce #1536 to focus on C mappings? Ideally it would be nice at some point to generate the C include and symbol mappings too, like we do for C++, but I don't know that there are any good sources of truth.

mapgen/iwyu-mapgen-std-symbol.py
kimgr: I think it's down to `O(N * log(N))` or something now, right? Anyway, the input is not expected to grow unbounded, so this is obviously fine no matter what.

std_symbol_map.inc
kimgr: The C++ C-compatibility headers are quite annoying because their behavior is/used to be poorly specified, and sometimes in conflict with the corresponding C headers. I wonder if maybe it's even a bad idea to consider the C standard for C++ headers? If we follow the letter of the C++ standard, at least we should be internally consistent.

alejandro-colomar:
> I'm happy with the patch as-is now. If anything, maybe remove `Part 1` from the commit message, it already explains the limitations quite clearly.
> 
> @alejandro-colomar @bolshakov-a I wonder if we'll ever reach consensus on the only-canonical vs. laissez-faire modes. I'm tempted to pretend #1536 doesn't exist, since it's been dormant for a year, and having auto-generated complete symbol mappings for C++ is a major improvement for users.
> 
> What if we reduce #1536 to focus on C mappings?

Since I only program in C, I would be happy as long as I get the strict mappings in C only.  :)

Since I don't expect reaching consensus, I think the best thing would be to provide a `--canonical-only` flag to select the stricter behavior.  However, I expect that to be non-trivial to implement, so a temporary workaround can be to keep the C mappings strict with the C++ ones being more lenient.

> Ideally it would be nice at some point to generate the C include and symbol mappings too, like we do for C++, but I don't know that there are any good sources of truth.

The main source of truth is the C standard, and then POSIX.1.  They disagree on a few identifiers, though.  And then there's a lot of non-standard stuff.

For now, I guess manually updating them is easier.

std_symbol_map.inc
alejandro-colomar: I don't remember all of the details, but IIRC it was preferable to use the C (.h) headers rather than the C++ (c*) headers in C++ code.  The c* headers were underspecified historically, as you say, while the .h headers are guaranteed to work, because of compatibility with old programs.

So, for C identifiers, the C standard is indeed what should be followed, and the C++ standard should be taken with a pinch of salt.

mapgen/iwyu-mapgen-std-symbol.py
bolshakov-a: Yes, it is already irrelevant.

bolshakov-a:
Of course, it would be preferable to have consistent C and C++ mappings. I think the flag is a good idea, but not for this PR (and it is not in my priorities).

> Ideally it would be nice at some point to generate the C include and symbol mappings too, like we do for C++, but I don't know that there are any good sources of truth.

I have an idea of generating the C library mapping from the synopses of the corresponding C++ headers, with necessary hand-made additions. Probably, POSIX specification is also a good source, if there is a convenient way to parse it.

std_symbol_map.inc
kimgr: I've personally preferred the C headers in the past, because it's clearer how they work. But there are plenty of coding standards that encourage the C++ wrapper headers.

The theory seems to be, from C++'s perspective, to pretend like the C headers don't exist. If you need C functionality, use the headers sanctioned by the C++ standard. So it's not necessarily relevant to discuss the C headers in a C++ context.

I don't really agree with that take, or know if it's fully practical, but if it's at all possible I think we should treat C and C++ as completely different modes in IWYU.

kimgr:
> > Ideally it would be nice at some point to generate the C include and symbol mappings too, like we do for C++, but I don't know that there are any good sources of truth.
>
> The main source of truth is the C standard, and then POSIX.1. They disagree on a few identifiers, though. And then there's a lot of non-standard stuff.

Right, but do they exist somewhere in parseable form?

> I have an idea of generating the C library mapping from the synopses of the corresponding C++ headers, with necessary hand-made additions. Probably, POSIX specification is also a good source, if there is a convenient way to parse it.

Not sure about using C++ to reason about C, see my response in the other thread. I worry that both languages are internally consistent, but not cross-consistent, if that makes sense. It would be nice to treat C entirely separate to avoid weird cross-pollination. But of course it's worth a shot to see what falls out :-)

std_symbol_map.inc
bolshakov-a: > So, for C identifiers, the C standard is indeed what should be followed, and the C++ standard should be taken with a pinch of salt.

I agree with this, but I'm still not fully confident that "described in ..." always means "the canonical header is ..."

If the C++ standard had any clear indication, I'd take that.

alejandro-colomar:
> > > Ideally it would be nice at some point to generate the C include and symbol mappings too, like we do for C++, but I don't know that there are any good sources of truth.
> > 
> > 
> > The main source of truth is the C standard, and then POSIX.1. They disagree on a few identifiers, though. And then there's a lot of non-standard stuff.
> 
> Right, but do they exist somewhere in parseable form?

<https://port70.net/~nsz/c/> had transformed old versions of the standard to plain text, which I've used for parsing in the past.  However, the formatting of C23 seems to have changed significantly, and the author of those conversions gave up.

I don't know of any parseable versions of the C standard.  And the sources are not public, so one can't work on the sources directly.

I wish the editor would publish plain-text (and maybe also HTML) versions of the standard.  Cc: @ThePhD

> 
> > I have an idea of generating the C library mapping from the synopses of the corresponding C++ headers, with necessary hand-made additions. Probably, POSIX specification is also a good source, if there is a convenient way to parse it.
> 
> Not sure about using C++ to reason about C, see my response in the other thread. I worry that both languages are internally consistent, but not cross-consistent, if that makes sense. It would be nice to treat C entirely separate to avoid weird cross-pollination. But of course it's worth a shot to see what falls out :-)

Yes, C++ is not compatible with C, and when in doubt --even in C++ code--, the C standard should have preference, as it's more stable.  I'd like to avoid using C++ for determining the C mappings.

Also, I'd reject anything that bring the whole set of mappings before a --canonical-only flag is implemented.  As long as we don't have such a flag, I'd like to keep the mappings manual.

bolshakov-a:
@amurzeau managed to get the C library mapping some time ago, see #785.

std_symbol_map.inc
alejandro-colomar: Well, nothing in the standard says so explicitly, but if you compare those descriptions with the traditional implementations of libc, you'll see that with a few exceptions, the identifiers are defined in those headers, and the rest only provide them through inclusion.

bolshakov-a: I've taken a look at glibc v. 2.36, and it defines `wint_t` in a separate header `<bits/types/wint_t.h>`. Moreover, `<wctype.h>` doesn't include `<wchar.h>`, which is not so small. Hence, requiring a user to include `<wchar.h>` when he only uses `wint_t` and e.g. `towlower`, you'd actually make worse.

alejandro-colomar: I'd say that's more an exception than the rule.

alejandro-colomar: But I agree in some cases having the more lenient rules can be good, and thus I think it's make sense to have `--canonical-only` and default to not using it.

alejandro-colomar: Hmmmm, I've had an idea.

If iwyu(1) had a concept of canonical header and another of alternative header, we could have the following rule:

-  Every header included in a file must be the canonical header of at least one identifier used in the file.

This would allow including wint_t in a file that uses towlower(), but wouldn't allow it in a file that only uses <wchar.h> stuff.

Does that sound like a good thing to you?

Cc: @kimgr

kimgr: I guess that makes sense. I don't have a clear idea of how to implement it, particularly since we don't have a strong concept of "canonical header". And I'm also not sure how it interplays with the current set-cover algorithm used to decide the desired headers. But it rings true!

kimgr:
OK, are we ready to move ahead with this? I don't want to block this PR on trying to solve the C standard problem, because that seems like a separate endeavor altogether.

iwyu_include_picker.cc
```diff
@@ -302,8 +302,13 @@ const IncludeMapEntry libc_symbol_map[] = {
   { "offsetof", kPrivate, "<stddef.h>", kPublic },
 };
 
-// Common kludges for C++ standard libraries
 const IncludeMapEntry stdlib_cxx_symbol_map[] = {
+#include "std_symbol_map.inc"  // IWYU pragma: keep
+
+  { "std::size_t", kPrivate, "<cstdio>", kPublic },
```
kimgr, line 308: Does this qualify as a "common kludge" too? That is, should the comment move up above this line?

bolshakov-a: This is not a kludge but a documented source of `std::size_t`. It is just not yet included to the generated mapping.

std_symbol_map.inc
bolshakov-a: > This would allow including wint_t in a file that uses towlower(), but wouldn't allow it in a file that only uses <wchar.h> stuff.

I haven't understood it, tbh. What do you exactly mean "including wint_t"?

As for your overall idea, it sounds good. If a file used to use `size_t` and `strcpy` and included `<string.h>`, and then stopped to use `strcpy`, it would be good if IWYU suggested to replace `<string.h>` with `<stddef.h>`.

alejandro-colomar:
> OK, are we ready to move ahead with this? I don't want to block this PR on trying to solve the C standard problem, because that seems like a separate endeavor altogether.

Yup, I'm okay with the idea of this PR for the moment.  I don't want to block it.

std_symbol_map.inc
alejandro-colomar: > > This would allow including wint_t in a file that uses towlower(), but wouldn't allow it in a file that only uses <wchar.h> stuff.
> 
> I haven't understood it, tbh. What do you exactly mean "including wint_t"?
> 

Oops, sorry, my brain didn't print some words.  :)

I wanted to say "This would allow including <wctype.h> for wint_t in a file that uses towlower(), ...".

> As for your overall idea, it sounds good. If a file used to use `size_t` and `strcpy` and included `<string.h>`, and then stopped to use `strcpy`, it would be good if IWYU suggested to replace `<string.h>` with `<stddef.h>`.

Yup, that's the idea.

alejandro-colomar: Seems we have an agreement.  :-)

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -0,0 +1,566 @@
...
+    else:
+        print_content(std_source_path, lang)
+
+if __name__ == '__main__':
+    argparser = argparse.ArgumentParser(
+        description='Generate IWYU mappings for the C++ standard library.')
+    argparser.add_argument('std_source_path',
+        help='Path to TeX sources of the standard ("source" subdirectory)')
+    argparser.add_argument('--lang', choices=["c++", "imp"], default="c++",
+        help="Output language")
```
kimgr: You seem to be using single quotes consistently, but some of these strings got double quotes when copied from the other mapper.

kimgr:
@bolshakov-a One tiny nit, otherwise looks ready to merge!

mapgen/iwyu-mapgen-std-symbol.py
bolshakov-a: One of my ex-colleagues said: "Copy-pasting is a crime against humanity"

bolshakov-a:
One more side note about C library mappings. If I understand correctly "7.1.3 Reserved identifiers", the C standard allows users to declare their own entities with names coinciding with standard library names unless the corresponding standard header is included. As an extreme case, a user may define its own `size_t` and `NULL` in its own header and not include `<stddef.h>` in any way. So, IWYU should probably have a flag to opt-out C standard library mappings, so as to support such cases and not remap user's symbols to standard library headers.

kimgr:
> So, IWYU should probably have a flag to opt-out C standard library mappings, so as to support such cases and not remap user's symbols to standard library headers.

That's one of the intended use cases of `--no_internal_mappings`. It might be nice to be more fine-grained somehow.

kimgr:
> Formally, this fixes https://github.com/include-what-you-use/include-what-you-use/issues/1616, https://github.com/include-what-you-use/include-what-you-use/issues/1763, and probably https://github.com/include-what-you-use/include-what-you-use/issues/1863. However, similar bugs may appear until all the mapping has been written.

I closed these issues tagged with the 0.26 milestone.

alejandro-colomar:
> One more side note about C library mappings. If I understand correctly "7.1.3 Reserved identifiers", the C standard allows users to declare their own entities with names coinciding with standard library names unless the corresponding standard header is included. As an extreme case, a user may define its own `size_t` and `NULL` in its own header and not include `<stddef.h>` in any way. So, IWYU should probably have a flag to opt-out C standard library mappings, so as to support such cases and not remap user's symbols to standard library headers.

That's an annoyance of the standard, and something that programs shouldn't rely on, because it's just so easy to include one of those headers accidentally through another include.

It's there essentially to allow introducing new identifiers in the standard in new headers, such as when bool was added in <stdbool.h> to avoid breaking existing bool definitions, but only as a transitory thing.

std_symbol_map.inc
AaronBallman: > I don't remember all of the details, but IIRC it was preferable to use the C (.h) headers rather than the C++ (c*) headers in C++ code. The c* headers were underspecified historically, as you say, while the .h headers are guaranteed to work, because of compatibility with old programs.
> 
> So, for C identifiers, the C standard is indeed what should be followed, and the C++ standard should be taken with a pinch of salt.

The situation is a bit more nuanced than that.

WG21 *redefines the contents* of C standard library headers. IOW, including `<string.h>` does not mean you're getting the one from your C standard library, instead you're usually getting the header injected by the STL: https://github.com/llvm/llvm-project/blob/main/libcxx/include/string.h This way, C++ ends up with overload sets for some C APIs, removes C syntactic features that C++ doesn't support (like static array extents, `restrict`, etc), and other adjustments. What you end up with is morally equivalent but not the exact same thing as what you get in C. The `<c*>` variant of the header includes the underlying `<*.h>` file but then does things which are not valid in C, like adding symbols to `namespace std`, etc. Then there's the fact that C++ nails down what symbols live in which headers more tightly than C does.

So which one the user wants to use depends on context. If the user is writing a header file intended to be shared between C and C++, then they almost certainly want the `<*.h>` file. If they're writing an implementation file for C++, they possibly want the `<c*>` file instead.

PR #1875, opened 2025-12-23
Support variable template specs in mappings

bolshakov-a:
The change in `GetInstantiatedFromDecl` and in `GetDefinitionAsWritten` is required because otherwise, for partially specialized variable templates, an implicitly instantiated specialization of the partial specialization comes into `GetWrittenQualifiedNameAsString` instead of the partial specialization itself.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -11,7 +11,6 @@
 
 #include "iwyu_ast_util.h"
 
-#include <cstddef>
 #include <regex>
 #include <set>                          // for set
 #include <string>                       // for string, operator+, etc
@@ -56,6 +55,7 @@
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/PointerUnion.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/Casting.h"
 
@@ -166,8 +166,11 @@ using clang::UnresolvedLookupExpr;
 using clang::UsingDirectiveDecl;
 using clang::ValueDecl;
 using clang::VarDecl;
+using clang::VarTemplateDecl;
+using clang::VarTemplatePartialSpecializationDecl;
 using clang::VarTemplateSpecializationDecl;
 using llvm::ArrayRef;
+using llvm::ListSeparator;
 using llvm::PointerUnion;
 using llvm::cast;
 using llvm::dyn_cast;
@@ -175,6 +178,7 @@ using llvm::dyn_cast_or_null;
 using llvm::errs;
 using llvm::isa;
 using llvm::range_size;
+using llvm::raw_ostream;
 using llvm::raw_string_ostream;
 using llvm::zip_equal;
 using std::function;
@@ -187,7 +191,7 @@ namespace include_what_you_use {
 
 namespace {
 
-void DumpASTNode(llvm::raw_ostream& ostream, const ASTNode* node) {
+void DumpASTNode(raw_ostream& ostream, const ASTNode* node) {
   if (const Decl *decl = node->GetAs<Decl>()) {
     ostream << "[" << GetKindName(decl) << "] " << PrintableDecl(decl);
   } else if (const Stmt *stmt = node->GetAs<Stmt>()) {
@@ -248,6 +252,36 @@ static const RedeclarableTemplateDecl* GetRedeclNotSpecifyingAnyDefArg(
       return redecl;
   }
   return nullptr;
+}
+
+template <typename T>
+static void PrintExplicitSpecTplArgs(const T* spec,
+                                     const PrintingPolicy& printing_policy,
+                                     raw_ostream& ostream) {
+  if (spec->isExplicitSpecialization()) {
+    ostream << '<';
+    const TemplateArgumentList& args = spec->getTemplateArgs();
+    const TemplateParameterList* params =
+        spec->getSpecializedTemplate()->getTemplateParameters();
+    ListSeparator LS;  // ', ' by default.
+    for (unsigned i = 0, ie = args.size(); i < ie; ++i) {
+      bool include_type = TemplateParameterList::shouldIncludeTypeForArgument(
+          printing_policy, params, i);
+      const TemplateArgument& arg = args.get(i);
+      // Clang encloses arguments corresponding to a parameter pack in angle
+      // brackets, so cannot use the default 'print' for them.
+      if (arg.getKind() == TemplateArgument::Pack) {
+        for (TemplateArgument pack_arg : arg.getPackAsArray()) {
+          ostream << LS;
+          pack_arg.print(printing_policy, ostream, include_type);
+        }
+      } else {
+        ostream << LS;
+        args.get(i).print(printing_policy, ostream, include_type);
+      }
+    }
+    ostream << '>';
+  }
 }
 
 }  // anonymous namespace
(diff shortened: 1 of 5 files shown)


iwyu_ast_util.cc
```diff
@@ -250,6 +253,38 @@ static const RedeclarableTemplateDecl* GetRedeclNotSpecifyingAnyDefArg(
...
+    const TemplateArgumentList& args = spec->getTemplateArgs();
+    const TemplateParameterList* params =
+        spec->getSpecializedTemplate()->getTemplateParameters();
+    for (unsigned i = 0, ie = args.size(); i < ie; ++i) {
+      if (i > 0)
+        ostream << ", ";
+      bool include_type = TemplateParameterList::shouldIncludeTypeForArgument(
+          printing_policy, params, i);
+      const TemplateArgument& arg = args.get(i);
+      // Clang embraces arguments corresponding to a parameter pack in angle
```
kimgr: embraces -> encloses maybe?

iwyu_ast_util.cc
```diff
@@ -250,6 +253,38 @@ static const RedeclarableTemplateDecl* GetRedeclNotSpecifyingAnyDefArg(
...
+static void PrintExplicitSpecTplArgs(const T* spec,
+                                     const PrintingPolicy& printing_policy,
+                                     raw_ostream& ostream) {
+  if (spec->isExplicitSpecialization()) {
+    ostream << '<';
+    const TemplateArgumentList& args = spec->getTemplateArgs();
+    const TemplateParameterList* params =
+        spec->getSpecializedTemplate()->getTemplateParameters();
+    for (unsigned i = 0, ie = args.size(); i < ie; ++i) {
+      if (i > 0)
```
kimgr: This looks like a nice utility: https://github.com/llvm/llvm-project/blob/a666d1fd9a1b06e9122531b7e139471f67c1a1d8/llvm/include/llvm/ADT/StringExtras.h#L520

iwyu_ast_util.cc
```diff
@@ -250,6 +253,38 @@ static const RedeclarableTemplateDecl* GetRedeclNotSpecifyingAnyDefArg(
...
+        ostream << ", ";
+      bool include_type = TemplateParameterList::shouldIncludeTypeForArgument(
+          printing_policy, params, i);
+      const TemplateArgument& arg = args.get(i);
+      // Clang embraces arguments corresponding to a parameter pack in angle
+      // brackets, so cannot use the default 'print' for them.
+      if (arg.getKind() == TemplateArgument::Pack) {
+        ArrayRef<TemplateArgument> pack_args = arg.getPackAsArray();
+        for (size_t j = 0, je = pack_args.size(); j < je; ++j) {
+          if (j > 0)
```
kimgr: And using the ListSeparator here would make it possible to simplify the loop into a range loop, I think.

iwyu_ast_util.cc
```diff
@@ -1183,34 +1196,45 @@ TemplateInstantiationData GetTplInstDataForVariable(
...
       return partial_spec_decl;
     }
   }
-  // class_decl is not instantiated from a template.
-  return class_decl;
+  if (const auto* tpl_sp_decl = dyn_cast<VarTemplateSpecializationDecl>(decl)) {
+    // An instantiated variable template.
+    PointerUnion<VarTemplateDecl*, VarTemplatePartialSpecializationDecl*>
+        instantiated_from = tpl_sp_decl->getInstantiatedFrom();
+    if (const VarTemplateDecl* tpl_decl =
```
kimgr: Could use `const auto*` here since we have the `VarTemplateDecl` in the rhs cast expression

iwyu_ast_util.cc
```diff
@@ -1183,34 +1196,45 @@ TemplateInstantiationData GetTplInstDataForVariable(
   return TemplateInstantiationData{resugar_map, provided_types};
 }
 
-const NamedDecl* GetInstantiatedFromDecl(const NamedDecl* class_decl) {
+const NamedDecl* GetInstantiatedFromDecl(const NamedDecl* decl) {
   if (const ClassTemplateSpecializationDecl* tpl_sp_decl =
-      DynCastFrom(class_decl)) {  // an instantiated class template
+          DynCastFrom(decl)) {  // an instantiated class template
```
kimgr: Let's switch this away from `DynCastFrom` to `const auto* tpl_sp_decl = dyn_cast<ClassTemplateSpecializationDecl>(...)`

iwyu_ast_util.cc
```diff
@@ -1183,34 +1196,45 @@ TemplateInstantiationData GetTplInstDataForVariable(
...
 }
 
 const NamedDecl* GetDefinitionAsWritten(const NamedDecl* decl) {
   // First, get to decl-as-written.
-  if (const CXXRecordDecl* class_decl = DynCastFrom(decl)) {
-    decl = GetInstantiatedFromDecl(class_decl);
-    if (const ClassTemplateDecl* tpl_decl = DynCastFrom(decl))
-      decl = tpl_decl->getTemplatedDecl();  // convert back to CXXRecordDecl
-  } else if (const FunctionDecl* func_decl = DynCastFrom(decl)) {
+  if (const FunctionDecl* func_decl = DynCastFrom(decl)) {
```
kimgr: Let's abandon `DynCastFrom` as part of rewrite here too

iwyu_ast_util.cc
```diff
@@ -1219,6 +1243,11 @@ const NamedDecl* GetDefinitionAsWritten(const NamedDecl* decl) {
     const FunctionDecl* tp_decl = func_decl->getTemplateInstantiationPattern();
     if (tp_decl && tp_decl->getFriendObjectKind() == Decl::FOK_None)
       decl = tp_decl;
+  } else {
+    decl = GetInstantiatedFromDecl(decl);
+    if (const ClassTemplateDecl* tpl_decl = DynCastFrom(decl))
```
kimgr: Let's abandon `DynCastFrom` as part of rewrite here too

iwyu_ast_util.cc
```diff
@@ -538,41 +573,19 @@ string GetWrittenQualifiedNameAsString(const NamedDecl* named_decl) {
...
-      static regex tpl_param{"(?:type|value|template)-parameter-\\d+-(\\d+)"};
-      retval = regex_replace(retval, tpl_param, ":$1");
-    }
+    PrintExplicitSpecTplArgs(cls_spec, printing_policy, ostream);
+  } else if (const auto* var_spec =
+                 dyn_cast<VarTemplateSpecializationDecl>(named_decl)) {
+    PrintExplicitSpecTplArgs(var_spec, printing_policy, ostream);
   }
 
+  // Replace clang placeholders in partial specialization arguments with :N,
```
kimgr, line 583: Hmm, don't we need to `ostream.flush();` before using `retval` below? Maybe there's never any buffering in `raw_ostring_stream`?

iwyu_ast_util.cc
bolshakov-a: Thanks!

iwyu_ast_util.cc
bolshakov-a: Yeah, `there is no need to call flush().`

https://github.com/llvm/llvm-project/blob/4e44e8761795d4672121368929b9c99f6dbaffa7/llvm/include/llvm/Support/raw_ostream.h#L659

kimgr:
Nice, thanks!

iwyu_ast_util.cc
```diff
@@ -250,6 +254,36 @@ static const RedeclarableTemplateDecl* GetRedeclNotSpecifyingAnyDefArg(
...
+template <typename T>
+static void PrintExplicitSpecTplArgs(const T* spec,
+                                     const PrintingPolicy& printing_policy,
+                                     raw_ostream& ostream) {
+  if (spec->isExplicitSpecialization()) {
+    ostream << '<';
+    const TemplateArgumentList& args = spec->getTemplateArgs();
+    const TemplateParameterList* params =
+        spec->getSpecializedTemplate()->getTemplateParameters();
+    ListSeparator LS;  // ', ' by default.
```
kimgr, line 266: Oops, this ended up with an LLVM-style name. Fixed in https://github.com/include-what-you-use/include-what-you-use/commit/f81729654195586cc1f9b62b4d044cd03d2727c6.

bolshakov-a: Thank you!

PR #1337, opened 2023-11-11
Support libc++ 17

kon72:
This change updates `libcxx.imp` to match Clang 17.

PR #1876, opened 2025-12-28
Update LLVM libc++ mappings to LLVM 21 / 22.

jashank:
Keeping these mappings up-to-date avoids unfortunate surprises when upstream changes the layout of the headers.  Previously (in #1337), we updated the LLVM libc++ internal mappings for LLVM v17.  But the passage of time, and intervening LLVM versions, have meant new header divisions have appeared; for example, `<__ostream/*>`, which provides `<<` for `std::basic_ostream`.

Therefore, update the mappings to the current LLVM mainline. (For our purposes, this is close enough to LLVM 21.1.8.)

Whilst here, improve the update commands to more closely match the current codebase format, and improve the documentation directions to more clearly note the required changes.  (It may be that this has reached the point where a simple generator tool would be a good longer-term solution.)

### DIFF
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -1189,14 +1189,14 @@ const IncludeMapEntry libcxx_symbol_map[] = {
 };
 
 const IncludeMapEntry libcxx_include_map[] = {
-    {"<__mutex_base>", kPrivate, "<mutex>", kPublic},
     {"<__tree>", kPrivate, "<set>", kPublic},
     {"<__tree>", kPrivate, "<map>", kPublic},
 
     // For the following entries:
-    // cd llvm-project/libcxx/include ; find -type d -name "__*" | sort | sed -e 's#./__\(.*\)#  { "@<__\1/.*>", kPrivate, "<\1>", kPublic },#'
+    // cd llvm-project/libcxx/include ; find . -type d -name "__*" | grep -v cxx03 | LC_ALL=C sort | sed -e 's#./__\(.*\)#    {"@<__\1/.*>", kPrivate, "<\1>", kPublic},#'
     //
-    // tweak locale_dir entry, and comment out debug_utils, fwd, pstl, support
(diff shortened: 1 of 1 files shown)


kimgr:
Thanks!

See also: https://github.com/include-what-you-use/include-what-you-use/issues/1271 -- it seems upstream libc++ ships with generated mappings, so I've been meaning to try and make IWYU pick up .imp files from the include search path too, but there's a bunch of invariants to rethink.

> (It may be that this has reached the point where a simple generator tool would be a good longer-term solution.)

Definitely! Would you like to take a stab at it? I started at some point, but don't use libc++ myself, so not really sure what conventions to rely on.

jashank:
Aha! #1271 is exactly the breadcrumb I needed. The link there is now stale: `libcxx.imp` is now generated from the build process — https://github.com/llvm/llvm-project/blob/e2e2c50bebc03d2e70eb1b33827cf6258496ed2c/libcxx/include/CMakeLists.txt#L1700-L1707 — and, moreover, the packaging on both FreeBSD (in the base system) and Fedora (in `libcxx-devel`, part of the `llvm` srpm) install the generated product into `/usr/include/c++/v1/libcxx.imp`. I strongly suspect that adopting this is actually the correct answer, certainly easier than maintaining a generator tool here.

Maybe the right answer here is, as other tools do, to establish some sort of either canonical "site-[tool]" / "vendor-[tool]" directories for additional .imp files to be installed into, or to establish some sort of search path for existing directories in which one might find other `.imp` files.

kimgr:
@jashank Yeah, that's essentially what #1271 suggests, the idea was to pick up `.imp` files from the normal header search path. But it's not entirely clear how to make it more opt-in. It might be nice to play on your vendor directory, and only look in `*/.iwyu` for each header search path. But let's continue discussion on #1271 if you have more ideas, I'd love to move forward with that, but I'm a bit overwhelmed :)

PR #1877, opened 2025-12-30
Improve forward-declare printing for templates

### LINKED ISSUE #1862: requires requires results in forward declaration with syntax error
Input:

* `a.h`:
  ```c++
  template <typename T>
  requires requires(T a) { *a; }
  class X {};
  ```
* `b.h`:
  ```c++
  #include "a.h"

  bool foo(X<int*>& value);
  ```

Running `include-what-you-use --std=c++20 b.h` results in:
```text
error: invalid argument '--std=c++20' not allowed with 'C'

b.h should add these lines:
template <typename T> requires requires (T X;

b.h should remove these lines:
- #include "a.h"  // lines 1-1

The full include-list for b.h:
template <typename T> requires requires (T X;
---
```
As you can see, the forward declaration `template <typename T> requires requires (T X;` has an obvious syntax error.

(Sorry for the "error: invalid argument", I didn't figure out how to get rid of this... I'm getting a similar error in an existing project with more complex headers and templates and without that error, the above was my attempt to produce a minimal reproducer.)

```console
include-what-you-use 0.25 based on Homebrew clang version 21.1.4
```

kimgr:
Issue #1862 describes a problem with ad-hoc requires expressions in templates. IWYU prints a forward-declaration for:

    template <typename T>
    requires requires(T a) { *a; }
    class X {};

as:

    template <typename T> requires requires (T X;

The string-chomping logic to isolate the template-prefix and tag kind ended up deleting too much.

Instead, add and use a ScopedRemoveDefinition helper to temporarily ignore the body and super classes of the declaration for printing, and then do more focused cleanup of the 'final' attribute.

Add a test bed for the fwd decl printer with basic coverage of tag decls and templates, including some more advanced requires expression. This uncovered a few other bugs, left as TODOs in the test.

Remove the fwd_decl_final test case in favor of this new one.

Fixes #1862.


tests/cxx/fwd_decl_print.cc
```diff
@@ -0,0 +1,114 @@
...
+// Derived templates
+DerivedTemplate<int>* dt1;
+
+// Templates with explicit and implicit attributes (attrs should not be included
+// in fwd decl).
+TemplateWithStdAttr<int>* at1;
+TemplateWithGNUAttr<int>* at2;
+// TODO: this results in:
+//     template <class T> class alignas(8)  TemplateWithAlignAs;
+// but the alignas attribute should not be included.
```
bolshakov-a: They are acceptable in forward-declarations, they should just be the same if present.

bolshakov-a:
I haven't reviewed the test in detail, but overall this LGTM!

tests/cxx/fwd_decl_print.cc
kimgr: Ah, thanks. I know `final` is prohibited for forward decls, so I thought the same was true for `alignas()`. But that makes sense. I guess I'll try to drop it in the interest of being consistent with non-templates.

kimgr: Updated the commentary to make that clearer. I might follow up with a later fix for `alignas` consistency.

PR #1878, opened 2026-01-02
Print template forward-decls without alignas keyword

kimgr:
C++ has two specifiers in c++11 (some new may arrive in C++26): final
and alignas(...). They are modelled as attributes in Clang, and printed
by the DeclPrinter.

Generalize our forward-decl printer to replace both 'alignas(...)' and
'final' using a regular expression.

Collapse repeated spaces to cover for ugly formatting from Clang.

Try to improve the comments in this function to make it clearer what
we're doing.

### DIFF
--- a/tests/cxx/fwd_decl_print-d1.h
+++ b/tests/cxx/fwd_decl_print-d1.h
@@ -26,6 +26,7 @@ class DerivedClass : public Class {};
 class [[nodiscard]] ClassWithStdAttr {};
 class __attribute__((warn_unused_result)) ClassWithGNUAttr {};
 
-// Class types with implicit attributes.
+// Class types with keyword attributes.
 class alignas(8) ClassWithAlignAs {};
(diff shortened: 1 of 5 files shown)


tests/cxx/fwd_decl_print-d1.h
```diff
@@ -26,6 +26,7 @@ class DerivedClass : public Class {};
 class [[nodiscard]] ClassWithStdAttr {};
 class __attribute__((warn_unused_result)) ClassWithGNUAttr {};
 
-// Class types with implicit attributes.
+// Class types with keyword attributes.
 class alignas(8) ClassWithAlignAs {};
+class alignas(sizeof(bool)) ClassWithAlignAsExpr {};
```
bolshakov-a: Nit: `alignof` instead of `sizeof` would look more like a real-world case.

PR #1879, opened 2026-01-02
Distinguish function overloads in mappings

bolshakov-a:
This change adds function parameter types to names searched in IWYU symbol mapping files. In order not to break existing mappings, and to provide a way to map functions whose parameter types are not strictly specified, IWYU searches the bare function name in the mappings if it fails to find the name with parameter types specified.


bolshakov-a:
There is at least a couple of issues:

- It would be definitely better to allow such typedefs as `std::size_t` in mappings, whose underlying types are platform-dependent. With the presented approach, which uses canonical types, at least `operator new` cannot be mapped in a portable way. OTOH, canonical types seem to be more robust in other cases, and the presented behavior is consistent with the _current_ behavior with template specialization arguments.
- There is no distinction between a function and a function template with the same arguments, like in this case:
  ```cpp
  void Fn();
  template <typename> void Fn();
  ```
  Forcing mapping file writers to enumerate all the template arguments (like `TplFn<:0, :1, :2>(arg1, :1, arg2, :0 *, ...)`) looks like an excessive burden. I thought of designating template functions with just empty angle brackets (`<>`), but it would look like a specialization, not the primary template. I'm not sure how important such cases are.

bolshakov-a:
> at least `operator new` cannot be mapped in a portable way

Looks like it should not actually be done, because users may provide their own [replacement functions](https://timsong-cpp.github.io/cppwp/n4950/replacement.functions#2), and the ordinary placement-new is handled explicitly [in `VisitCXXNewExpr`](https://github.com/include-what-you-use/include-what-you-use/blob/309f9be69030b708bc4193755ddd4985d0947b22/iwyu.cc#L3068-L3091).

kimgr:
Phew, that's a big one! But also interesting, as it opens up for more fine-grained mappings -- nice work!

* I'm a bit concerned about diagnostics in general now. Most standard library functions are very focused and limited in the number of parameters, but user code typically runs wild with many parameters and long type names, e.g.
  ```
  TemplateInstantiationData GetTplInstDataForClass(
    llvm::ArrayRef<clang::TemplateArgument> written_tpl_args,
    const clang::ClassTemplateSpecializationDecl* cls_tpl_decl,
    std::function<set<const clang::Type*>(const clang::Type*)> provided_getter);
  ```
  How would that one render in an IWYU diagnostic?
* I'm thinking it would be nice to only print argument lists in diagnostics if there's any ambiguity. But that seems kind of hard to do as we'd need to keep an index of all encountered symbols before attempting any diagnostics?

I wonder if we're reaching the limits of what we can encode in the primitive mapping key.

Hmm, actually, this reminds me of C++ name decoration/mangling. That seems like a good way to generate a unique key, since it's well-defined and covers enough to work for C++ toolchains in general. But I wouldn't want to expose users to decorated names, so not sure if it's immediately useful.

iwyu_include_picker.h
```diff
@@ -192,7 +196,10 @@ class IncludePicker {
 
   // Returns the headers which the symbol is mapped to. If none, returns
   // the headers which decl_filepath is mapped to.
-  vector<string> GetMappedPublicHeaders(const string& symbol_name,
+  vector<string> GetMappedPublicHeaders(const string& decl,
```
kimgr: I think the `symbol_name` might also refer to a macro, so `decl` might narrow things too much.

bolshakov-a: Accidental renaming, thanks!

bolshakov-a:
> Phew, that's a big one!

The main change in tests is the new `overload_fn_mapping` test.

> How would that one render in an IWYU diagnostic?

Like this:
```
Marked full-info use of decl include_what_you_use::GetTplInstDataForClass(llvm::ArrayRef<clang::TemplateArgument>, const clang::ClassTemplateSpecializationDecl *, std::function<std::set<const clang::Type *> (const clang::Type *)>)
```
However, note that, in `#include` comments, IWYU prints short names by default. Function arguments are displayed in verbose diagnostics, which often are already quite big (> 5 000 000 lines for `iwyu.cc` at level 6). But they should be printed so that mapping file writers may see how exactly they are spelled (and the writers might want to add the arguments to a mapping entry even if currently there is no ambiguity).

> But I wouldn't want to expose users to decorated names, so not sure if it's immediately useful.

Neither would I. Moreover, every implicit template specialization has a unique mangled name, but only explicit ones are relevant for mappings. Implicit specs should be covered by the primary template mapping entry.

Also recall that modern linkers have demanglers to produce diagnostics in a human-readable form, with all the function arguments printed, like this:

```
/usr/bin/ld: CMakeFiles/include-what-you-use.dir/iwyu.cc.o: в функции «include_what_you_use::IwyuAstConsumer::GetTplInstData(llvm::ArrayRef<clang::TemplateArgument>, clang::ClassTemplateSpecializationDecl const*) const»:
/home/bols/progs/iwyu/include-what-you-use/iwyu.cc:5547: неопределённая ссылка на «include_what_you_use::GetTplInstDataForClass(llvm::ArrayRef<clang::TemplateArgument>, clang::ClassTemplateSpecializationDecl const*, std::function<std::set<clang::Type const*, std::less<clang::Type const*>, std::allocator<clang::Type const*> > (clang::Type const*)>)»
```
(Sorry for Russian. Not sure how to switch off the localization.) Btw, looks like IWYU currently prints _less_ than it should print in the diagnostics: the default template args of `std::set` are skipped.

bolshakov-a:
> It would be definitely better to allow such typedefs as `std::size_t` in mappings...

An example of function overloads taking a typedef with some implementation-defined underlying type is `abs(intmax_t)` and `div(intmax_t, intmax_t)` from `<cinttypes>`.

Also, `<print>` header has overloads of `print`, `println`, `vprint_unicode`, and `vprint_nonunicode`, which take `FILE*`, but `FILE` may be an alias of some unspecified type. However, this may be solved by denoting `<print>` as a fallback header of that function names, and specifying parameter types explicitly only for `<ostream>` overloads.

An idea came into my mind: what if IWYU would search the both options (with sugared and canonical parameter type) in mappings? But this would increase search time by 2^N, where N is the parameter number...

kimgr:
> However, note that, in #include comments, IWYU prints short names by default. 

Ah, I saw the `--comment_style=long` example in the tests, but that's an outlier. Thanks!

> Moreover, every implicit template specialization has a unique mangled name, but only explicit ones are relevant for mappings. Implicit specs should be covered by the primary template mapping entry.

I'm sure that's just one of many complications :-). I only had a burst of inspiration that mangled names are a perfect two-way mapping between raw detailed readable names and more compact mangled unreadable names, and they might be useful somehow. I assumed LLVM already has code to mangle/demangle... That said, I don't have a clear idea of where/how to use mangling, so let's pretend I didn't bring them up :-)

> > It would be definitely better to allow such typedefs as std::size_t in mappings...
> An example of function overloads taking a typedef with some implementation-defined underlying type is abs(intmax_t) and div(intmax_t, intmax_t) from <cinttypes>.

Those are good examples... You mentioned earlier that canonical types seem more robust -- what are the kinds of scenarios you're thinking of? I'm wondering if preferring sugared types would actually make more sense here (particularly since the function is called `GetWrittenQualifiedNameAsString` with `Written` indicating that we want sugared names).

Back to the PR, I have to say, I'm not happy about how the tests become so much more detailed. The regex escaping is particularly painful. I'm trying to think of ways to reduce the level of detail where there are no overloads to disambiguate.

Overall, while this would potentially unlock a lot of mapping opportunities, I fear that it might complicate things to a point where it's hard to reason about the overall behavior. Throwing out another half-baked thought: are we basically reimplementing [name lookup/overload resolution](https://en.cppreference.com/w/cpp/language/overload_resolution.html)? It seems a bit overwhelming.

Do you have any data on the percentage/count of overloads in `std` vs. unique function names? I wonder if there's a simpler model we could look at if the number is relatively small.

bolshakov-a:
> You mentioned earlier that canonical types seem more robust -- what are the kinds of scenarios you're thinking of?

A couple of cases are covered in `overload_fn_mapping` test. The most unexpected one is that in the following code:
```cpp
namespace ns {
  class Class;
  void FnFromNs(Class);
}
```
the function parameter type `Class` appears to be non-canonical despite being `RecordType`. The canonical type is the qualified type `ns::Class`, and I'd expect it to be acceptable in mappings, like it works with template specialization mapping. Also there is somewhat similar case with using-declaration. Moreover, function templates aren't handled as expected without resugaring (written parameter names like `T` or `U` are retained and hence not replaced with `:N`).

Canonicalization makes mappings more robust against how exactly parameter types are written in a header (with or without qualification, or aliased). However, you may argue that keeping sugared names makes mappings more robust against changes in underlying types of typedefs.

> I'm trying to think of ways to reduce the level of detail where there are no overloads to disambiguate.

Function parameters may be denoted as `\(.*\)` when not interesting.

> Throwing out another half-baked thought: are we basically reimplementing [name lookup/overload resolution](https://en.cppreference.com/w/cpp/language/overload_resolution.html)?

I think IWYU should support function overloading, at some reasonable extent. Exactly because of that this PR has been opened.

> Do you have any data on the percentage/count of overloads in std vs. unique function names?

I've counted 1047 unique function names in C++23, 49 of which are overloaded in different headers. But note that among the overloaded functions are such widely used ones as `std::move`, `std::swap`, `std::begin`, `std::end`, `std::remove`, `std::get`, and the new `std::erase`. I tried the following with MS STL:
```cpp
#include <algorithm>
#include <string>

int main() {
  std::string s1, s2;
  swap(s1, s2);                               // line 6
  (void)std::move(s2);                        // line 7
  std::move(begin(s1), end(s1), begin(s2));   // line 8
  erase(s2, 'a');   // C++20
  s1.erase(std::remove(s1.begin(), s1.end(), 'b'), s1.end()); // line 10
}
```
The IWYU output is terrible:
```
main.cpp:5:15: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.
main.cpp:5:19: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.
main.cpp:5:3: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.
main.cpp:6:3: warning: std::_Construct_in_place is defined in <xutility>, which isn't directly #included.
main.cpp:6:3: warning: std::addressof is defined in <memory>, which isn't directly #included.
main.cpp:6:3: warning: std::forward is defined in <utility>, which isn't directly #included.
main.cpp:6:3: warning: std::swap is defined in <utility>, which isn't directly #included.
main.cpp:6:3: warning: std::swap is defined in <xstring>, which isn't directly #included.
main.cpp:7:9: warning: std::move is defined in <type_traits>, which isn't directly #included.
main.cpp:8:13: warning: std::addressof is defined in <memory>, which isn't directly #included.
main.cpp:8:13: warning: std::begin is defined in <xutility>, which isn't directly #included.
main.cpp:8:13: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.
main.cpp:8:24: warning: std::addressof is defined in <memory>, which isn't directly #included.
main.cpp:8:24: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.
main.cpp:8:24: warning: std::end is defined in <xutility>, which isn't directly #included.
main.cpp:8:33: warning: std::addressof is defined in <memory>, which isn't directly #included.
main.cpp:8:33: warning: std::begin is defined in <xutility>, which isn't directly #included.
main.cpp:8:33: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.
main.cpp:8:3: warning: std::declval is defined in <utility>, which isn't directly #included.
main.cpp:8:3: warning: std::forward is defined in <utility>, which isn't directly #included.
main.cpp:8:3: warning: std::move is defined in <type_traits>, which isn't directly #included.
main.cpp:8:3: warning: std::move is defined in <xutility>, which isn't directly #included.
main.cpp:9:3: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.
main.cpp:9:3: warning: std::erase is defined in <xstring>, which isn't directly #included.
main.cpp:10:12: warning: std::_String_iterator is defined in <xstring>, which isn't directly #included.
main.cpp:10:12: warning: std::remove is defined in <xmemory>, which isn't directly #included.
main.cpp:10:27: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.
main.cpp:10:39: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.
main.cpp:10:55: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.
main.cpp:10:6: warning: std::conjunction_v is defined in <type_traits>, which isn't directly #included.

#include <memory>       // for addressof
#include <type_traits>  // for conjunction_v, move
#include <utility>      // for forward, declval, swap
#include <xmemory>      // for remove
#include <xstring>      // for _String_iterator, erase, swap
#include <xutility>     // for begin, _Construct_in_place, end, move

- #include <algorithm>  // lines 1-1

#include <memory>       // for addressof
#include <string>       // for string
#include <type_traits>  // for conjunction_v, move
#include <utility>      // for forward, declval, swap
#include <xmemory>      // for remove
#include <xstring>      // for _String_iterator, erase, swap
#include <xutility>     // for begin, _Construct_in_place, end, move
```
IWYU annoyingly reports some internal stuff of `std::string` like `addressof` or `forward`. (After 309f9be69030b708bc4193755ddd4985d0947b22, they are at least attributed to their documented headers and not to `<type_traits>`. Thank you, @bolshakov-a!) But note that one-arg version of `std::move` should be attributed to `<utility>` and not to `<type_traits>` where it is defined in MS STL; the documented header for algorithmic version of `std::move`, as well as of `std::remove`, is `<algorithm>` and not `<xutility>` or `<xmemory>`, and so on. `std::swap` taking `string`s should probably be [mapped to nothing](https://github.com/include-what-you-use/include-what-you-use/issues/785#issuecomment-3703302972), but not the overload(s) taking fundamental types.

kimgr:
> > I'm trying to think of ways to reduce the level of detail where there are no overloads to disambiguate.
>
> Function parameters may be denoted as \(.*\) when not interesting.

Yeah, but even that is quite noisy. And functions with parameter lists tend to have a lot of chars that need escaping (`()`, `*`, `.`, `[]`)

I have an idea for a test framework improvement to make it behave more like [LLVM's `FileCheck`](https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-regex-matching-syntax), which does literal checking by default and regex with special syntax. I'll see if I can hack something up quickly that vibes with IWYU's conventions.

> I think IWYU should support function overloading, at some reasonable extent. Exactly because of that this PR has been opened.

Yes, I think it's super valuable. I just want to make sure the mappings can be made stable, convenient and unambiguous in the long-term. Doing something hard-coded inside IWYU is "better" in that sense as a first step, because we don't expose users to new mapping syntax until we've seen how it plays out in practice.

But let's iterate on this. I'll start with the test framework matching syntax, so the test changes become less verbose.

bolshakov-a:
> I just want to make sure the mappings can be made stable, convenient and unambiguous in the long-term.

Because of this I marked this PR as a draft, hoping that some qualified people will look at this and leave their comments...

> Doing something hard-coded inside IWYU is "better" in that sense as a first step, because we don't expose users to new mapping syntax until we've seen how it plays out in practice.

I haven't fully understood what you mean. Some internal rules for internal mappings? But you said that users can export the internal mappings with `--export_mappings`...

kimgr:
> > Doing something hard-coded inside IWYU is "better" in that sense as a first step, because we don't expose users to
> > new mapping syntax until we've seen how it plays out in practice.
>
> I haven't fully understood what you mean. Some internal rules for internal mappings? But you said that users can
> export the internal mappings with --export_mappings...

Internal-only constructs don't have to follow the same rules as general mappings; I'm thinking something like `iwyu_cache.cc` which is a hard-coded lookup table.

kimgr:
Here's what I had in mind for more beautiful tests: https://github.com/include-what-you-use/include-what-you-use/pull/1881. I think it should make the more common cases read nicely, e.g.
```
// IWYU: printf\(const char \*, \.\.\.\) is...*<stdio.h>
// IWYU: I2_Class::~I2_Class\(\) is...*badinc-i2-inl.h
// IWYU: puts\(const char \*\) is...*stdio.h
// IWYU: Fn\(const unsigned int \(\*\)\[5\]\[7\]\) is...*-i22.h
```
will become
```
// IWYU: printf(const char *, ...) is...*<stdio.h>
// IWYU: I2_Class::~I2_Class() is...*badinc-i2-inl.h
// IWYU: puts(const char *) is...*stdio.h
// IWYU: Fn(const unsigned int (*)[5][7]) is...*-i22.h
```

This is at the expense of the relatively few places we use `.*` to wild-match, e.g.
```
 // IWYU: I2_TemplateClass::I2_TemplateClass<.*> is...*badinc-i2-inl.h
```
will turn into:
```
// IWYU: I2_TemplateClass::I2_TemplateClass<{{.*}}> is...*badinc-i2-inl.h
```

Let's discuss the details over in #1881, but for me this removes most concerns about spelling out the whole function signature in tests.

bolshakov-a:
I think if the mapping is fully internal, it might be not a mapping but some (ugly) hard-code like in `MapPrivateDeclToPublicType`. Or clang AST matchers may be used probably... So `GetWrittenQualifiedNameAsString` and hence the tests would not affected at all.

kimgr:
> I think if the mapping is fully internal, it might be not a mapping but some (ugly) hard-code like in MapPrivateDeclToPublicType. 
> Or clang AST matchers may be used probably... So GetWrittenQualifiedNameAsString and hence the tests would not affected 
> at all.

Exactly, yeah. Users would not be able to do anything for user-defined overloads, but also users would not be exposed to the complexity.

But let's explore the approach in this PR a bit more and see what we can learn.

bolshakov-a:
Rebased.

iwyu_include_picker.h
```diff
@@ -195,6 +199,9 @@ class IncludePicker {
   vector<string> GetMappedPublicHeaders(const string& symbol_name,
                                         const string& use_path,
                                         const string& decl_filepath) const;
+  vector<string> GetMappedPublicHeaders(const clang::NamedDecl* decl,
+                                        const string& use_path,
+                                        const string& decl_filepath) const;
```
kimgr, line 204: Looks like you could get rid of `decl_filepath` now, it is always `GetFilePath(decl)`. OTOH, maybe it's nice to be able to override it?

bolshakov-a: Despite the fact that the tests pass if `decl_filepath` is replaced by `GetFilePath(GetInstantiationLoc(GetLocation(decl)))`, I think it may be wrong because `ReportFullSymbolUse` [canonicalizes the location in some cases](https://github.com/include-what-you-use/include-what-you-use/blob/73eb686d8cb26e25532bb321756c140e9970868e/iwyu_output.cc#L720) and doesn't do it in others. (However, that logic is questionable.)

kimgr: I forgot about the loc canonicalization. Makes sense as-is for now, then.

kimgr:
Alright, so what's the status?

You've raised:
- No clear strategy for sugared vs. canonical types
- Cannot distinguish between function and function templates with the same arguments

My concerns:
- :heavy_check_mark: Tests shouldn't require gratuitous escaping
- Possibly gold-plating: it would be nice if decls were canonicalized, so spacing doesn't matter for what's in the mapping vs. what's looked up

And what are next steps? Is it feasible to generate mappings with param-lists from the standard text?

bolshakov-a:
> Possibly gold-plating: it would be nice if decls were canonicalized, so spacing doesn't matter for what's in the mapping vs. what's looked up

Do you mean that e.g. ` *` would be automatically replaced by `*` both in mappings and in a looked-up name? That's an interesting idea. However, I think it may be a follow-up improvement.

> Is it feasible to generate mappings with param-lists from the standard text?

I think I could do it, although it requires some more elaborated parsing, and overloads using aliased parameter types should be explicitly excluded or handled somehow if we decide the parameter types should be canonical.

kimgr:
> Do you mean that e.g.  * would be automatically replaced by * both in mappings and in a looked-up name? That's an interesting idea. However, I think it may be a follow-up improvement.

Something like that, yeah. Ideally I'd like to spin it through something like clang-format, or even `clang -ast-print` to get a fully canonical string, but that's a little over-the-top. And yes, can definitely be attacked later.

> I think I could do it, although it requires some more elaborated parsing, and overloads using aliased parameter types should be explicitly excluded or handled somehow if we decide the parameter types should be canonical.

Yeah, it seems quite a lot harder to parse. So did you plan to add overloads as an internal mapping set by hand? With 49 functions, that might be doable.

bolshakov-a:
> Ideally I'd like to spin it through something like clang-format, or even `clang -ast-print` to get a fully canonical string, but that's a little over-the-top.

This is what @AaronBallman suggested [here](https://github.com/include-what-you-use/include-what-you-use/pull/1869#issuecomment-3596373093). As for `-ast-print`, I think it uses the same `TypePrinter` under the hood as `print` methods:
```
$ echo 'int Fn(int*);' | ../llvm-project/build/bin/clang -xc++ - -Xclang -ast-print -fsyntax-only
int Fn(int *);
```
> So did you plan to add overloads as an internal mapping set by hand? With 49 functions, that might be doable.

This may be an option.

kimgr:
This looks contained enough: https://github.com/llvm/llvm-project/blob/c6db8f4229d70169949430185e156fb2e6e4150e/clang/lib/Index/CommentToXML.cpp#L590 -- but I don't know what the performance profile of that operation is. It may be painful on startup for thousands of symbols, and maybe even worse on lookup. But it might be a fun test.

The challenge with the Type/DeclPrinters is we'd have to somehow parse it into a Decl/Type first, which is probably more complicated and expensive.

bolshakov-a:
That looks interesting, thanks! I think your idea is to normalize somehow both a pretty-printed declaration and strings from mappings to some common format before comparing? That would be good.

> The challenge with the Type/DeclPrinters is we'd have to somehow parse it into a Decl/Type first, which is probably more complicated and expensive.

I'd say it is impossible because if you have in a mapping file a symbol `a::b::c`, you don't know if `a` and `b` are class or namespace names. So you cannot parse it without context, but probably you can lex it and produce some normalized string from the lexed sequence of tokens. I think it is what clang-format does, and what `CommentASTToXMLConverter::formatTextOfDeclaration` does. But I remember that clang-format can produce some weird results when given not enough context.

Anyway, I don't think this PR depends on such a decision. We can just say to users to look into the logs for correct symbol form. I think it is even desirable for function argument handling to correspond to what is done for explicit template specialization arguments at the moment.

bolshakov-a:
Well, looks like nobody else wants to take part in the discussion, so my thoughts are:

- A function with the same name and parameter list as some function template is rare case, which may be ignored for now. If required, some differentiation like `<>` may be added later, keeping the proposed in this PR behavior as a fallback.
- The standard library mapping should not depend on whether `std::` is spelled in a parameter type in some implementation or not, therefore some canonicalization is needed anyway. As for `intmax_t` overloads, they are defined in `<cinttypes>` only when `intmax_t` designate an extended integer type, otherwise they are defined in `<cstdlib>` and `<cmath>`. So, mapping e.g. `abs(intmax_t)` to `<cinttypes>` would be just plain wrong in general (and for all the three main implementations). I don't want to introduce a special case for `FILE` in IWYU code. Perhaps, distinct platform-specific mappings which would use its underlying type is a better solution.

UPD: https://github.com/include-what-you-use/include-what-you-use/pull/1879#issuecomment-3707278753 contains a better proposal regarding the functions from `<print>` taking `FILE*`: just not to specify parameters for them at all.

kimgr:
@bolshakov-a So how would you like to move forward with this PR? I'm not exactly sure what the next logical step is...?

bolshakov-a:
If we agree that the overloads of the standard library functions taking `FILE*` as a parameter may be specified in separate platform-specific mappings using the underlying types of `FILE`, or just mapped to nothing in the common mapping, this PR can be merged.

kimgr:
I'm a bit conflicted about timing;

* part of me thinks we should wait with this until after the 0.26 release (LLVM branch is taken, so we may need to branch for clang_22 soon)
* ... but part of me thinks we might as well get this in 0.26 to get some feedback from users

The only visible change in this PR is that diagnostics are "noisier", right, i.e. with parameter types? I'm not sure how angry that might make people, or if they would like to be able to turn it off with a flag or something.

bolshakov-a:
> this PR can be merged.

But note that I looked only at C++23. I don't know if C++26 adds some functions which are distinguished by `size_t` in parameters or something like that. Neither I can say about the future, of course.

> The only visible change in this PR is that diagnostics are "noisier", right, i.e. with parameter types?

Correct. But if this PR is merged now, maybe we can merge some useful standard function mappings before the release. However, I don't have strong opinion.

bolshakov-a:
> But if this PR is merged now, maybe we can merge some useful standard function mappings before the release.

In fact, before writing overloaded function mappings, I want to answer [the question](https://github.com/include-what-you-use/include-what-you-use/issues/785#issuecomment-3780271209) about mapping to nothing.

bolshakov-a:
I've found a problem, indicated by TODO.

bolshakov-a:
> I've found a problem, indicated by TODO.

The fix is here: #1900.

kimgr:
Alright, I feel ready to merge this. I did some limited performance testing, and couldn't see any significant regressions. @bolshakov-a You don't have anything blocking either, right? Besides all potential future work on top of this?

bolshakov-a:
I don't see any blockers except that I found an easy way to obtain the language options inside `GetMappedPublicHeaders` method and just have pushed the change.

However, if you are afraid of long function signatures in the log messages, maybe it is worth to keep the old output format and try to perform some normalization on parameter type strings, as you suggested, so that users should not look at the verbose log to find out the required spelling. OTOH, it may be done later if users complain for some reasons.

bolshakov-a:
Oh, wait. One blocker is the docs.

kimgr:
Let's roll it out and see! Thanks for humoring a long and slow review cycle!

kimgr:
(oh, and nice catch on the language, hadn't occurred to me that this is only an option for C++)

bolshakov-a:
Thanks for the review and the help with format of the checks in tests!

PR #1880, opened 2026-01-03
Add bug test for issue 1810

kimgr:


kimgr:
Oh, wait. We're arguing that it's not a bug. Probably don't want to add this test.

bolshakov-a:
With names from `std`, this is more likely a bug. Just as with the comparison operators of `std::string`: there should be `#include <array>` somewhere in the translation unit, in a standard-conforming program.

kimgr:
I think it's more like an improvement on the existing design. It currently works as intended, so I don't think we need to be told if it stops working :-)

PR #1881, opened 2026-01-05
Add explicit regex syntax for test diagnostic assertions

kimgr:
99+% of our assertions follow this form:

    // IWYU: SomeType is...*someheader.h
    // IWYU: SomeType needs a declaration

Under the hood the expected diagnostic text after '// IWYU:' is
turned into a regular expression which is then matched against actual
IWYU diagnostics.

We've had some minor friction with this in the past, e.g. having to
escape parentheses and periods, such as:

    // IWYU: Unknown or malformed pragma \(foo\)
    // IWYU: I1_TemplateClass is...*badinc-i1.h.*#included\.

This has not been a pressing issue so far, but we're now considering
reporting full function signatures instead of just function names, and
escaping C++ function argument lists makes them nearly unreadable:

    // IWYU: printf\(const char \*, \.\.\.\) is...*<stdio.h>
    // IWYU: I2_Class::~I2_Class\(\) is...*badinc-i2-inl.h
    // IWYU: puts\(const char \*\) is...*stdio.h
    // IWYU: Fn\(const unsigned int \(\*\)\[5\]\[7\]\) is...*-i22.h

So turn the implementation around, and treat everything as literal text,
except:

- `{{pattern}}`
- `...*`, which is now an alias for `{{.*}}`

The `{{pattern}}` syntax and some of its implementation is inspired by
LLVM's FileCheck tool, which is used to verify most of Clang's
diagnostics, and so has a good track record for C++ code.

The examples above would now be spelled as:

    // IWYU: SomeType is...*someheader.h
    // IWYU: SomeType needs a declaration
    // IWYU: Unknown or malformed pragma (foo)
    // IWYU: I1_TemplateClass is...*badinc-i1.h...*#included.
    // IWYU: printf(const char *, ...) is...*<stdio.h>
    // IWYU: I2_Class::~I2_Class() is...*badinc-i2-inl.h
    // IWYU: puts(const char *) is...*stdio.h
    // IWYU: Fn(const unsigned int (*)[5][7]) is...*-i22.h

We had a limited number of wildcard matches, and mostly for template
argument lists. They used to be spelled:

    // IWYU: I2_TemplateClass::I2_TemplateClass<.*> is...*badinc-i2-inl.h

but now require a bit more syntax:

    // IWYU: I2_TemplateClass::I2_TemplateClass<{{.*}}> is...*badinc-i2-inl.h
    // or, with alias
    // IWYU: I2_TemplateClass::I2_TemplateClass<...*> is...*badinc-i2-inl.h

Update all affected tests to use the new format.

### DIFF
--- a/iwyu_test_util.py
+++ b/iwyu_test_util.py
@@ -252,12 +252,8 @@ def _GetExpectedDiagnosticRegexes(spec_loc_to_line):
     line = spec_loc_to_line[loc]
     m = _EXPECTED_DIAGNOSTICS_RE.match(line.strip())
     assert m is not None, "Input should contain only matching lines."
-    regex = m.group(1)
-    if not regex:
-      # Allow the regex to be omitted if we are uninterested in the
-      # diagnostic message.
-      regex = r'.*'
-    regexes.append(re.compile(regex))
+    pattern = _ParseMatchPattern(m.group(1))
+    regexes.append(re.compile(pattern))
     # Do we have a spec on the next line?
     path, line_num = loc
     next_line_loc = path, line_num + 1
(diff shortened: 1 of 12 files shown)


bolshakov-a:
Cool, thank you! However, it may be surprising that `...*` is the only allowed regular expression pattern besides the braced form...

kimgr:
Yeah, I'm on the fence about that one too. But `...*` reads so much better than `{{...*}}` (or `{{.*}}` as I've chosen to interpret it). I'm very sensitive to syntactic noise :o)

kimgr:
Plus now it's straightforward to search-and-replace all tests to remove the `...*` alias if we change our minds.

kimgr:
@bolshakov-a Updated commit message, please take another look.

iwyu_test_util.py
```diff
@@ -268,6 +264,35 @@ def _GetExpectedDiagnosticRegexes(spec_loc_to_line):
...
+  """ Parses an IWYU diagnostic assertion and returns a regex string. """
+  if not pattern:
+    # Allow the pattern to be omitted if we are uninterested in the
+    # diagnostic message.
+    return '.*'
+
+  # First normalize the conventional ...* pattern to {{.*}}
+  pattern = pattern.replace('...*', '{{.*}}')
+
+  # Then interleave literal literal string segments with regex segments,
```
bolshakov-a:
```suggestion
  # Then interleave literal string segments with regex segments,
```

bolshakov-a:
One typo, otherwise LGTM!

iwyu_test_util.py
kimgr: Thanks, fixed.

PR #1885, opened 2026-01-13
Set up infrastructure for IWYU unittests

kimgr:
Add iwyu-unittests and run-unittests targets.

Add a shared iwyu_unittest_main.cc which sets up IWYU globals before
running tests.

We can now build and run unittests with either of these commands:

    ninja iwyu-unittests && ./bin/iwyu-unittests
    ninja iwyu-unittests && ctest [--output-on-failure] -R "unittest"
    ninja run-unittests

Unittests will automatically be picked up in CI by ctest.

No tests registered for execution yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,6 +265,32 @@ if (MSVC)
   )
 endif()
 
+# Add unittest target.
+add_llvm_executable(iwyu-unittests
+  unittests/iwyu_unittest_main.cc
+)
+
+target_include_directories(iwyu-unittests PRIVATE
+  ${CMAKE_CURRENT_SOURCE_DIR}
+)
+
+target_link_libraries(iwyu-unittests PRIVATE
+  iwyu
+  iwyu-gtest
+)
+
+enable_testing()
+add_test(NAME iwyu-unittests
+  COMMAND $<TARGET_FILE:iwyu-unittests>
+)
+
+# Expose a custom run-unittests target in the generated buildsystem.
+add_custom_target(run-unittests
+  DEPENDS iwyu-unittests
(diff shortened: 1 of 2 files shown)


kimgr:
@bolshakov-a Can I ask you to try this one in Windows too? Would be interesting to see behavior both with and without GTest available.

bolshakov-a:
I currently don't have GTest installed, so I got "GTest not found, unittests not built" line in the CMake output. In other aspects, IWYU still compiles and works.

Tbh, I don't have a proper knowledge of CMake, so I don't know how to install GTest correctly so that CMake will see it.

kimgr:
Hmm, yeah. I'm not sure how this works on windows. I installed libgtest-dev on Ubuntu, and cmake magically found it.

I'll do some research and see if I can find some procedure for you to test, if possible.

bolshakov-a:
Perhaps it is sufficient to be able to run unittests in Unix environment. The other tests don't pass on Windows anyway.

firewave:
This should not be a custom target but added via [add_test](https://cmake.org/cmake/help/latest/command/add_test.html) so it is available via CTest as well.

kimgr:
@firewave Could you spell that out for me? I haven't managed to use `add_test` in a way that it both becomes available to CTest and shows up as a build system target (which I prefer).

firewave:
```cmake
enable_testing()
add_test(NAME iwyu-unittests COMMAND $<TARGET_FILE:iwyu-unittests>)
```

kimgr:
@firewave Thanks, but how do I run that from my build system (edit: as generated by `cmake -G Ninja` or even `cmake -G "Unix Makefiles"`)?

kimgr:
I ended up adding both.

PR #1887, opened 2026-01-15
Bundle googletest v1.17.0

kimgr:
The GoogleTest CMake quickstart recommends declaring a dependency on
googletest GitHub, which is presumably resolved at configure-time.

This seems quite fragile;

- if github.com is not reachable, our build fails
- if the googletest project is taken down, our build fails

Instead, add a simple bash script to let us manually clone and minify
the googletest project into the IWYU tree. Build a private iwyu-gtest
library from the resulting source code.

The vendor/googletest directory is the unabridged output of

    ./vendor/iwyu-roll-googletest.bash v1.17.0

Exclude the new vendor/ directory from license header checks.

The resulting libiwyu-gtest.a is unused at this point, but will be
linked with unittests as soon as we are able.


kimgr:
@bolshakov-a Could you give this a shot on Windows?

bolshakov-a:
It fails: C++17 version of the standard should be explicitly specified, because the default one in VS2022 is still C++14. The error message is:
```
vendor\googletest\include\gtest\internal\gtest-port.h(273,1): error C1189: #error:  C++ versions less than C++17 are not supported.
```

kimgr:
Now setting C++17 explicitly.

bolshakov-a:
It still fails with some strange errors. Looks like `gtest-all.cc` includes `vendor\googletest\VERSION` somehow...

bolshakov-a:
Looks like it is here: https://github.com/google/googletest/blob/v1.17.0/googletest/include/gtest/internal/gtest-port.h#L291. If I rename `VERSION` to `VERSION1`, compilation succeeds.

kimgr:
@bolshakov-a Reworked a bit, both the c++17 and `VERSION` issues should be resolved now.

bolshakov-a:
No, it doesn't still work. Even iwyu and include-what-you-use VS projects have now lost C++17 flag.

kimgr:
So in fact this didn't work as I expected: https://github.com/include-what-you-use/include-what-you-use/pull/1888#issuecomment-3764225490 -- I'll see if I can find an idiomatic way to do that.

kimgr:
@bolshakov-a Now rebased, could you give this another shot too?

bolshakov-a:
Eventually, this compiles! However, one warning appears, namely C4530, suggesting `/EHsc` flag to be added.

kimgr:
> However, one warning appears, namely C4530, suggesting /EHsc flag to be added.

Excellent, I'll throw that in among the compile options.

bolshakov-a:
Now it builds fine.

bolshakov-a:
Btw, `CHECK_UNREACHABLE_` inside `DeriveCXXStdLib` causes [C4715 warning](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4715), but it is on the master branch as well.

kimgr:
Ah, we should probably have a conditional `__declspec(noreturn)` on `~FatalMessageEmitter()`

PR #1888, opened 2026-01-17
[cmake] Set common C++ properties globally

kimgr:


firewave:
You could just set it globally.

kimgr:
@firewave Oh, are they just variables? You mean something like:
```
set(CXX_STANDARD 17)
set(CXX_STANDARD_REQUIRED ON)
set(CXX_EXTENSIONS OFF)
```
?

kimgr:
Updated to do that. Seems to work.

PR #1889, opened 2026-01-18
[cmake] Set common C++ property defaults globally

kimgr:
I tried to do this in 8f035c618998022db2c3aca3d2a15de9fbdf5327, but it didn't bite.

The target properties themselves don't seem to be variables, but they default to the value of CMAKE_$name.

Only set the variables unless they are already set, because users should be able to override.


kimgr:
@bolshakov-a Sorry to use you as a Windows buildbot, but could you check this on your MSVC rig?

bolshakov-a:
This works!

PR #1892, opened 2026-01-22
Fix build with GCC

bolshakov-a:
`llvm_update_compile_flags` CMake function called from `llvm_add_library` adds some compiler flags which are required to build correctly with GCC.

The problem was introduced in 9c7a6a3c08af3ecfc6e0a8716455435c696b7bb2.


bolshakov-a:
Superseded by #1894.

kimgr:
Oh. I'm sorry, I didn't see this PR before I started.

PR #1893, opened 2026-01-22
Fix reporting type under forwarding reference

### LINKED ISSUE #1886: IWYU suggests forward decl which breaks compilation for template instantiation with forwarding references
IWYU incorrectly suggests replacing an `#include` with a forward declaration for a type passed to a function template that uses a forwarding reference (`T&&`). Following this suggestion breaks the build because the template body accesses a member of the type, requiring the full definition at the point of instantiation.

This issue only occurs with forwarding references (`T&&`). Changing the template parameter to an lvalue reference (`T&`) or by-value (`T`) causes IWYU to correctly identify the need for the header.

**Reproduction Steps:**
- `test.cpp `:
```c++
#include "a.h"

template <typename T>
void f(T&& t) { t.h(); }

void g(A& a) { f(a); }
```
- `a.h`:
```c++
struct A { void h() {} };
```
- output:
```
$ include-what-you-use test.cpp 

struct A;

- #include "a.h"  // lines 1-1

struct A;
---
```

**Environment:**
```
include-what-you-use 0.25 (git:791e69e) based on Ubuntu clang version 21.1.8 (++20251221033104+2078da43e25a-1~exp1~20251221153121.57)
$ clang-21 -v
Ubuntu clang version 21.1.8 (++20251221033104+2078da43e25a-1~exp1~20251221153121.57)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-21/bin
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/11
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/12
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/13
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/14
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/15
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/15
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Candidate multilib: x32;@mx32
Selected multilib: .;@m64
```

bolshakov-a:
Function template deduced template argument types may be more structurally complex than the corresponding function argument type, at least in the case of forwarding references. For example:
```cpp
template <class T> int Fn(T&&);

class Class {} c;
int i = Fn(c);
```
Here, the function argument type is `Class`, but the template argument type is deduced to `Class&`. Hence, it should be disassembled to components so that IWYU can find a match for the function argument type later in `GetTplInstDataForFunction` to build the final resugaring map.

This fixes #1886.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -963,7 +963,7 @@ static map<const Type*, const Type*> GetTplTypeResugarMapForFunctionNoCallExpr(
       }
     }
   }
-  return retval;
+  return ResugarTypeComponents(retval);
 }
 
 static map<const Type*, const Type*> GetTplTypeResugarMapForExplicitTplArgs(
--- a/tests/cxx/derived_function_tpl_args.cc
+++ b/tests/cxx/derived_function_tpl_args.cc
@@ -33,6 +33,11 @@ template<typename T> void FnWithReference(const T& t) {
   (void)dummy;
 }
 
+template <typename T>
+void FnWithForwardingReference(T&& t) {
+  (void)sizeof(t);
+}
+
 template<typename T> struct TplClass {
   T t;
 };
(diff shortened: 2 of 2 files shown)


kimgr:
Nice, thanks!

Issue #1838, opened 2025-10-21
Missing complete use in template method expansion


bungeman:
Minimized example in  [3.zip](https://github.com/user-attachments/files/23024342/3.zip).

#### foo.h
```
struct Foo {
    void done();
};
```
#### test.cpp
```
#include "foo.h"

template <typename T> struct sp {
    ~sp() { ptr->done(); }
    T* ptr;
};

sp<Foo> MakeFoo();

void foo() {
    MakeFoo();
}
```
With llvm/llvm-project@3afbda091a049 and include-what-you-use 423301855beec
`./include-what-you-use -std=c++17 -c test.cpp -o test.o`
```
test.cpp should add these lines:
struct Foo;

test.cpp should remove these lines:
- #include "foo.h"  // lines 1-1

The full include-list for test.cpp:
struct Foo;
```
However, doing so will result in errors as `~sp()` must be instantiated which will use `Foo`.

bungeman:
Note that this may be related to (or a simpler reproduction of) #1626.

bolshakov-a:
There is a draft commit that should fix it: https://github.com/bolshakov-a/include-what-you-use/commit/37394c9fb792756fbae0000c6eba13c3a0e47c58.

kimgr:
Fixed on master in c0d9955.

PR #1896, opened 2026-01-23
Scan destructor of returned object at call site

### LINKED ISSUE #837: Auto and templated return type
Hello,

I found a case where IWYU makes a recommendation which does not compile when it encounters "auto". I observed this behavior at HEAD (a95999b244f551b719d2f38cfb2d2b7b1cc8fe34) of master compiled with clang-11.

Expected output when specifying the type of pFooB:
```
$./build/bin/include-what-you-use /home/vagrant/main.cpp
/home/vagrant/main.cpp should add these lines:
#include <memory>  // for unique_ptr

/home/vagrant/main.cpp should remove these lines:

The full include-list for /home/vagrant/main.cpp:
#include <memory>  // for unique_ptr
#include "fooA.h"  // for getPointerFooB
#include "fooB.h"  // for fooB
```
Actual output when declaring pFooB with auto:
```
$./build/bin/include-what-you-use /home/vagrant/main.cpp
/home/vagrant/main.cpp should add these lines:

/home/vagrant/main.cpp should remove these lines:
- #include "fooB.h"  // lines 1-1

The full include-list for /home/vagrant/main.cpp:
#include "fooA.h"  // for getPointerFooB
```
In the case where auto is used, it would be acceptable to me to remove the inclusion of fooB.h from main.cpp if it were also recommended to add an inclusion of fooB.h to fooA.h. Using the --no_forward_decls flag had no effect on the output of IWYU in this case. 

 I tried to debug this on my own but had a hard time following IWYU as worked through the tree. I did increase the verbosity to 7 and compared the output between the expected result and the erroneous result with auto. I noticed the following difference:

Good:
```
|-DeclStmt 0x3026f68
| `-VarDecl 0x30259b0  pFooB 'std::unique_ptr<fooB>':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >' listinit destroyed
|   `-ExprWithCleanups 0x3026f50 'std::unique_ptr<fooB>':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >'
|     `-CXXConstructExpr 0x3026f20 'std::unique_ptr<fooB>':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >' 'void (class std::unique_ptr<class fooB, struct std::default_delete<class fooB> > &&) noexcept' elidable list
|       `-MaterializeTemporaryExpr 0x3026f08 'std::unique_ptr<fooB>':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >' xvalue
|         `-CXXBindTemporaryExpr 0x3025af0 'std::unique_ptr<fooB>':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >' (CXXTemporary 0x3025af0)
|           `-CallExpr 0x3025ac8 'std::unique_ptr<fooB>':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >'
|             `-ImplicitCastExpr 0x3025ab0 'std::unique_ptr<fooB> (*)(void)' <FunctionToPointerDecay>
|               `-DeclRefExpr 0x3025a60 'std::unique_ptr<fooB> (void)' lvalue Function 0x2f2fef0 'getPointerFooB' 'std::unique_ptr<fooB> (void)'
```
Bad:
```
|-DeclStmt 0x41194f8
| `-VarDecl 0x4117e88  pFooB 'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >' listinit destroyed
|   `-ExprWithCleanups 0x41194e0 'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >'
|     `-CXXConstructExpr 0x41194b0 'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >' 'void (class std::unique_ptr<class fooB, struct std::default_delete<class fooB> > &&) noexcept' elidable list
|       `-MaterializeTemporaryExpr 0x4119498 'std::unique_ptr<fooB>':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >' xvalue
|         `-CXXBindTemporaryExpr 0x4117fd0 'std::unique_ptr<fooB>':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >' (CXXTemporary 0x4117fd0)
|           `-CallExpr 0x4117fa8 'std::unique_ptr<fooB>':'class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >'
|             `-ImplicitCastExpr 0x4117f90 'std::unique_ptr<fooB> (*)(void)' <FunctionToPointerDecay>
|               `-DeclRefExpr 0x4117f38 'std::unique_ptr<fooB> (void)' lvalue Function 0x4021f20 'getPointerFooB' 'std::unique_ptr<fooB> (void)'
```

I'm not sure what is being printed here, but using context clues, it looks like in the good case where the type is explicitly defined, IWYU picks up the type of pFooB as std::unique_ptr<fooB>. In the case where auto is used, IWYU finds the type of pFooB to be class std::unique_ptr<class fooB, struct std::default_delete<class fooB> >. In the case where The way I interpret this is that IWYU is not correctly identifying pFooB's type. The text of the source files is below:

```C++
//fooB.h
#include <memory>
#ifndef FOOB_H
#define FOOB_H

class fooB
{
    public:
        int fooBVar;
};
#endif
```
```C++
//fooA.h
#include <memory>
class fooB;

std::unique_ptr<fooB> getPointerFooB(){
    return std::make_unique<fooB>();
}
```
```C++
//main.cpp
#include "fooB.h"
#include "fooA.h"

int main()
{
    auto pFooB{ getPointerFooB()};

    //std::unique_ptr<fooB> pFooB{ getPointerFooB()};
    return 0;
}
```

Any help here would be much appreciated. I am looking at some input code where this example is a pattern used throughout the code.

bolshakov-a:
A function returning by value returns a temporary object whose destructor is eventually called and thus should compile.

Ideally, this handling should be in `TraverseCXXBindTemporaryExpr` because the corresponding node is inserted when a destructor is to be called. However, it would require some more code to avoid function templates, because destructors of types returned by function template implicit instantiations should be handled on scanning those instantiations, partly for taking into account type aliases and default template arguments. Moreover, the case of braced initialization of aggregate classes should be handled separately anyway, because for the braced return statement (`return {};`) clang doesn't insert `CXXBindTemporaryExpr` node. In fact, it doesn't even require the destructor to be compilable in this case, but gcc does (correctly). This handling has not been done in this changeset. In fact, IWYU doesn't currently even handle construction of aggregates through initializer lists. However, some test cases have been written in advance.

This fixes #837 and #1838.


kimgr:
LGTM!

I was a little surprised to see such focus on templates when the commit message indicates that this is mostly not about templates :) -- I would have expected some test cases returning plain class instances. But I suppose that's already covered by the `iwyu_stricter_than_cpp*` test?

bolshakov-a:
There is just no point in scanning destructors for non-template classes: all the needed types should be reported at the destructor definition site, not at any function call site.

bolshakov-a:
But wait. Are you saying about out-of-class inline destructor definitions? That may be worth testing...

kimgr:
> There is just no point in scanning destructors for non-template classes: all the needed types should be reported at the destructor definition site, not at any function call site.

OK, that sounds reasonable. But I find the lack of mention of templates in the code and the commit message vs. the exclusive use of templates for return type in the tests to be a little surprising.

> But wait. Are you saying about out-of-class inline destructor definitions? That may be worth testing...

I wasn't thinking about that specifically, no, but I guess that should have similar behavior? You're thinking something like:
```
// val.h
struct Val {
  ~Val();
}

// val-dtor.h
Val::~Val() {
  puts("goodbye");
}

// func.h
#include "val.h"

Val func() {
  return {}
}

// main.cc
#include "func.h"

void t() {
  (void)func();
}
```
I can't quite make that example add up :-)

bolshakov-a:
I have found a test (with some wrong style) and added this case to it. It fails on master branch.

kimgr:
Love it, thanks!

PR #1900, opened 2026-01-24
Fix dependent array size expression in mappings

bolshakov-a:
Symbol mappings should not rely on exact names of partial specialization template parameters. To avoid them in dependent array size expressions, the printing policy should be adjusted to canonicalize types.


kimgr:
Looks straightforward enough!

PR #1906, opened 2026-01-26
Report enumerations for __PRETTY_FUNCTION__

bolshakov-a:
Into some function-local predefined variables, compilers insert function template arguments. Some libraries like [MagicEnum](https://github.com/Neargye/magic_enum) make use of this to perform compile-time reflection on enumerations, allowing e.g. getting the written names of enumeration items. To actually have item names instead of number values, the complete enum type info must be present.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -233,6 +233,8 @@ using clang::PPCallbacks;
 using clang::ParenType;
 using clang::ParmVarDecl;
 using clang::PointerType;
+using clang::PredefinedExpr;
+using clang::PredefinedIdentKind;
 using clang::Preprocessor;
 using clang::QualType;
 using clang::QualifiedTypeLoc;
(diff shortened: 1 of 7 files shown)


bolshakov-a:
There is an article about that: https://blog.rink.nu/2023/02/12/behind-the-magic-of-magic_enum/

iwyu.cc
```diff
@@ -4246,6 +4248,48 @@ class InstantiatedTemplateVisitor
...
+                // way to forward-declare them.
+                if (enum_decl->isFixed())
+                  ReportDeclUse(caller_loc(), enum_decl);
+              }
+            }
+          }
+        }
+        break;
+      case PredefinedIdentKind::PrettyFunctionNoVirtual:
+        CHECK_UNREACHABLE_("Unexpected value");
```
bolshakov-a: I've pushed this to see the dogfood results, but probably it would be better to replace this with some logging?

tests/cxx/enums.cc
```diff
@@ -81,6 +81,111 @@ struct Struct2 {
 
 enum class Struct2::Nested { A, B, C };
 
+// Some of function-local predefined variables allow to output enum item names.
+// This trick is used in some enum reflection libraries. Test that IWYU detects
+// such uses correctly and requires complete enum types so that item names are
+// available to a compiler.
```
kimgr: Could you add a link to that blog post here too?

tests/cxx/enums.cc
```diff
@@ -7,7 +7,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-// IWYU_ARGS: -I .
+// IWYU_ARGS: -I . -fms-extensions
```
kimgr: I think of `-fms-extensions` as turning on a whole new compiler mode, and I don't want it to affect any other semantics in the normal enum tests. I realize it's inconvenient, but I wonder if it would be better to test MS extensions separately? I'm guessing it's their special predefined names?

bolshakov-a: > I'm guessing it's their special predefined names?

Yeah, it is `__FUNCSIG__` that is used in MSVC: https://github.com/Neargye/magic_enum/blob/5b6e12fa1b43f24a124fc1eaee5a5a2a3fa4dda4/include/magic_enum/magic_enum.hpp#L580-L586

iwyu.cc
kimgr: Wouldn't it be better to just treat it as the other PrettyFunction variants? The doc string says it's like PrettyFunction just without `virtual`. That said, I have no clue how to write test code to provoke this one :-)

tests/cxx/enums.cc
bolshakov-a: > I realize it's inconvenient, but I wonder if it would be better to test MS extensions separately?

Do you mean to introduce a new test named `ms_enums`?

kimgr: Yeah, I think limited coverage in a separate `ms_enums.cc` might be warranted.

iwyu.cc
bolshakov-a: There is no corresponding token. I looked at Clang sources, and it seems that `PredefinedExpr` is never created with such kind.

kimgr: Yeah, I only found it in some VTableBuilder debug-print-like dumps.

Not sure, the semantics are pretty clear: it's the same as `PrettyFunction`, just without a `virtual` keyword, but if it's never used... Feels like that enum value is in the wrong enum :-)

I guess it's either:

* Keep the CHECK_ so it explodes if this comes into use
* Make it a VERRS(2) log, so it prints something annoying if it comes into use
* Ignore it completely. Users will have to notice and report if it's eventually exposed as `__PRETTY_FUNCTION_NV__` or something, and IWYU doesn't see through it

Pick whatever you think is convenient.

kimgr:
Looks good to me, ping back when you've decided on the PrettyFunctionNoVirtual handling, and we can merge. Thanks!

iwyu.cc
bolshakov-a: In my previous work I was taught to assert aggressively (it was an architectural CAD project; there was a principle "It's better for a CAD to crash than for a building"). However, a source analysis tool is somewhat different thing, so let it logs.

bolshakov-a:
@kimgr

Issue #300, opened 2016-05-10
private header __hash_table is suggested for inclusion when using #include <unordered_set>


ggeorgiev:
(src/dom/cxx/cxx_library.h has correct #includes/fwd-decls)

src/dom/cxx/cxx_library.cpp should add these lines:
# include <__hash_table>               // for __hash_const_iterator

EugeneZelenko:
Very likely duplicated of #141. LLVM STL has __hash_table, GCC's doesn't.

firewave:
Still seeing this with 0.22.

firewave:
This might not actually be about missing libc++ mappings but possibly requiring headers for implicit symbols.

```
#include <__hash_table>   // for std::__1::operator!=
```

Using `-D_LIBCPP_REMOVE_TRANSITIVE_INCLUDES` that gets simplified to

```
#include <__hash_table>   // for std::operator!=
```

But using that define I am also seeing an additional `std::operator<` as reason for `<string>`.

I wonder if these operators should actually should be considered.

kimgr:
There's no contained repro here, so hard to say what's going on. This small example no longer suggests any private headers on mainline:
```
$ cat t.cc
#include <unordered_set>

void foo() {
  std::unordered_set<int> s;
  int sum = 0;
  for (int a : s) {
    sum += a;
  }
}

$ include-what-you-use t.cc

(t.cc has correct #includes/fwd-decls)
```

Closing.

firewave:
```cpp
#include <string>
#include <unordered_set>

static const std::unordered_set<std::string> us;

void f()
{
    if (us.find("") != us.end()) {}
}
```

```
$ include-what-you-use -Xiwyu --update_comments -stdlib=libc++ a.cpp

a.cpp should add these lines:
#include <__hash_table>   // for operator!=

a.cpp should remove these lines:

The full include-list for a.cpp:
#include <__hash_table>   // for operator!=
#include <string>         // for basic_string, string
#include <unordered_set>  // for unordered_set
---
```

[iwyu.zip](https://github.com/user-attachments/files/19924540/iwyu.zip)


Using 0.23 with Clang 19.1.7 on Manjaro Linux.

firewave:
I can also reproduce it with 0.24. Unfortunately I am not able to build the trunk right now.

kimgr:
Thanks!

firewave:
[...]

PR #1907, opened 2026-01-27
Handle `UnaryTransformType`s in typedef decls

### LINKED ISSUE #169: Fails to see through std::remove_pointer

bolshakov-a:
In libc++, some type transformation templates are defined using `UnaryTransformType`s, approximately like this:
```cpp
template <typename T>
struct remove_pointer {
  using type = __remove_pointer(T);
};
```
It broke IWYU analysis of type provision because `UnaryTransformType` desugars to its canonical (output) type, hence `SubstTemplateTypeParmType` node was skipped in `IsSubstTemplateTypeParmType`, and the underlying type was considered as provided by the alias.

After this change, `UnaryTransformType` is replaced by its base (input) type on collecting typedef type components, thus preserving detection of `SubstTemplateTypeParmType`.

This fixes #169.


bolshakov-a:
By the way, @kimgr, the bug test failed on the server not because of the bug but because the server has not libc++ installed. The same applies for tests for #300 and #1250.

iwyu_ast_util.cc
```diff
@@ -722,6 +723,14 @@ class TypeEnumeratorWithoutSubstituted
   bool TraverseTypeHelper(QualType qual_type, bool traverse_qualifier) {
     CHECK_(!qual_type.isNull());
     const Type* type = qual_type.getTypePtr();
+
+    // UnaryTransformType desugars by default to its canonical type which,
```
kimgr: Could you add a few examples of unary transforms, e.g. `A UnaryTransformType, such as __add_pointer or __remove_extents, desugars by default to its canonical type...`

iwyu_ast_util.h
```diff
@@ -840,6 +840,10 @@ set<const clang::Type*> GetCanonicalComponentsOfType(const clang::Type* type);
 // template <class T> struct Tpl1 { typedef Tpl2<T, A> Alias; };
 // the returned set for the type 'Tpl1<B>::Alias' contains 'A' but not 'B'
 // or any sugar for 'B'.
+// This not necessarily returns all the complex components but should return at
```
kimgr: Phrasing: `This does not necessarily return all the ...`

bolshakov-a: Thanks!

kimgr:
Looks great, thanks!

PR #1908, opened 2026-01-27
Fix bug test for #1250

bolshakov-a:
It is libstdc++ on which IWYU output is not ideal, not libc++. The test failed due to absence of libc++ on the server.

### DIFF
--- a/tests/bugs/1250/1250.cc
+++ b/tests/bugs/1250/1250.cc
@@ -7,7 +7,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-// IWYU_ARGS: -stdlib=libc++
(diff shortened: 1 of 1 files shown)


kimgr:
Ouch, thanks!

Looks like there are two more bug tests failing for the wrong reasons:
```
$ git grep "libc\+\+" -- tests/bugs/
tests/bugs/1250/1250.cc:// IWYU_ARGS: -stdlib=libc++
tests/bugs/169/169.cc:// IWYU_ARGS: -stdlib=libc++
tests/bugs/300/300.cc:// IWYU_ARGS: -stdlib=libc++
... unrelated ...
```

I'll look into those separately.

Issue #323, opened 2016-05-30
Using boost variant and c++14 boost apply_visitor does not recognize the need of types


ggeorgiev:
Say you have a: `boost::variant<Foo, Bar> baz` if used like this:

`boost::apply_visitor([](auto const& obj) { obj.something() },  baz)` 

The types `Foo` and `Bar` are needed, but iwyu is missing it.

kimgr:
Thanks, that makes sense. I wonder if the lambda body is even visited by IWYU. Can you run it with higher verbosity level and try to glean what it's doing?

bolshakov-a:
On the current IWYU master and with Boost 1.74, IWYU detects the full use of `Foo` and `Bar` for `boost::apply_visitor`. However, for the definition of `baz` variable, it reports that only `Foo` should have a complete type, which is a bug...

bolshakov-a:
Now IWYU works correctly for both Boost.Variant and Boost.Variant2. However, for `std::variant`, it doesn't report types for `visit` function (standard counterpart of `apply_visitor`).

PR #1910, opened 2026-01-28
Add parameter pack elements to resugar_map

bolshakov-a:
`InstantiatedTemplateVisitor` only reports types presented in `resugar_map`. This change appends template arguments passed to a class template parameter pack to `resugar_map`, thus allowing them to be reported.

This fixes the case reported in https://github.com/include-what-you-use/include-what-you-use/issues/323#issuecomment-3812413057, because the second parameter of `boost::variant` is a parameter pack.

Besides, this fixes some crashes in `GetRedeclSpecifyingDefArg` on handling templates with parameter packs. `GetTplParamNumberWithoutPack` has been introduced so that detecting of defaulted arguments may be performed correctly.

`llvm::iterator_range` is reported now for `zip_equal` return type used in a range-based for loop in `iwyu_ast_util.cc`. It exactly has a template parameter pack, hence its argument types are now reported.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -57,6 +57,7 @@
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/StringRef.h"
+#include "llvm/ADT/iterator_range.h"
 #include "llvm/Support/Casting.h"
 
 // TODO: Clean out pragmas as IWYU improves.
@@ -1525,8 +1526,15 @@ const NamedDecl* GetIndependentRedecl(const NamedDecl* decl) {
   // If there is no redeclaration without any explicitly written default
   // arguments, return the one specifying the last argument because it doesn't
   // depend on any other redeclaration.
-  unsigned last_param_idx = redeclarable->getTemplateParameters()->size() - 1;
+  unsigned last_param_idx = GetTplParamNumberWithoutPack(redeclarable) - 1;
   return GetRedeclSpecifyingDefArg(last_param_idx, redeclarable);
+}
+
+unsigned GetTplParamNumberWithoutPack(const TemplateDecl* tpl) {
+  const TemplateParameterList* params = tpl->getTemplateParameters();
+  unsigned size = params->size();
+  CHECK_(size > 0);
+  return params->hasParameterPack() ? size - 1 : size;
 }
 
 // --- Utilities for Type.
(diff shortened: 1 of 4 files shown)


iwyu_ast_util.cc
```diff
@@ -1761,15 +1761,20 @@ static TemplateInstantiationData GetTplInstDataForClassNoComponentTypes(
...
+  vector<TemplateArgument> tpl_args = written_tpl_args;
 
   if (cls_tpl_decl) {
     const TemplateArgumentList& tpl_arg_list =
         cls_tpl_decl->getTemplateInstantiationArgs();
     tpl_args = tpl_arg_list.asArray();
-    num_args = tpl_arg_list.size();
+
+    CHECK_(!tpl_args.empty());
+    TemplateArgument last = tpl_args.back();
```
kimgr, line 1781: Any reason not to make this a `const TemplateArgument&`?

iwyu_ast_util.cc
```diff
@@ -1761,15 +1761,20 @@ static TemplateInstantiationData GetTplInstDataForClassNoComponentTypes(
...
-  unsigned num_args = tpl_args.size();
+  vector<TemplateArgument> tpl_args = written_tpl_args;
 
   if (cls_tpl_decl) {
     const TemplateArgumentList& tpl_arg_list =
         cls_tpl_decl->getTemplateInstantiationArgs();
     tpl_args = tpl_arg_list.asArray();
-    num_args = tpl_arg_list.size();
+
+    CHECK_(!tpl_args.empty());
```
kimgr, line 1779: This makes me a little uneasy, but I guess a class template specialization can never have zero args?

iwyu_ast_util.cc
bolshakov-a: It may be invalidated when performing `insert` on `tpl_args`, I think.

bolshakov-a: But I'm not sure.

iwyu_ast_util.cc
bolshakov-a: I think so.

iwyu_ast_util.cc
kimgr: Oh, I didn't see the circularity there. Yeah, I guess if appending reallocates the vector it might get messy. Maybe add a little comment to explain so nobody gets tempted to optimize in the future?

kimgr:
By the way, funny that `llvm/ADT/iterator_range.h` is now desired. I guess there's a range loop somewhere that now expands a little differently. I'd be OK with adding the include, but I'm not sure it should really be there? Do you have a gut feeling one way or the other?

bolshakov-a:
The report of `llvm::iterator_range` is from here: https://github.com/include-what-you-use/include-what-you-use/blob/bc4fcd04759ebce0670c8b0d05951536228e80d8/iwyu_ast_util.cc#L2026-L2027

I suspect that it is from `zip_equal` return type, `zippy`, which exactly has a parameter pack in its template parameters. But I'm still trying to debug. Maybe, I'll postpone it until tomorrow...

bolshakov-a:
> I suspect that it is from `zip_equal` return type, `zippy`, which exactly has a parameter pack in its template parameters.

Yes, it is [reported for range-based for loop initializer](https://github.com/include-what-you-use/include-what-you-use/blob/bc4fcd04759ebce0670c8b0d05951536228e80d8/iwyu.cc#L1853). The exact type of range-initializer is `detail::zippy<detail::zip_first, class llvm::iterator_range<class clang::DeclContext::specific_decl_iterator<class clang::FieldDecl> >, class llvm::iterator_range<class clang::DeclContext::specific_decl_iterator<class clang::FieldDecl> > >`. Class template `zippy` [contains `std::tuple` field](https://github.com/llvm/llvm-project/blob/d8da229d7172700f4f93fafe3ac22144ca04f6e9/llvm/include/llvm/ADT/STLExtras.h#L783) which requires full argument types. The info that `iterator_range` is provided by fn-return is lost. I think that even if the for-loop were rewritten using explicit iterator variables and the old-style for-loop, the type returned by the initializer would anyway be reported for `.begin()` and `.end()` member call expressions. So we hardly can suppress such reportings in IWYU, hence it's better to apply the suggestion, I think. Anyway, LLVM people can decide to forward-declare `iterator_range` in the future for the `RecordDecl::fields()` method declaration, so including it explicitly in IWYU code can make it more robust.

iwyu_ast_util.cc
bolshakov-a: I added an explicit `pop_back`. Though it should not change observable behaviour, it seems to be more correct to _replace_ parameter pack argument with it's content.

iwyu_ast_util.cc
bolshakov-a: I have written some additional test cases and found a couple of crashes. I incorporated the fixes in this PR. I suspect, however, that they are present on master branch because they are somehow unrelated.

iwyu_ast_util.cc
kimgr: Ah, nice. I was actually about to suggest that, but I wasn't sure if there was a reason to keep it.

kimgr:
Thanks, this LGTM!

PR #1911, opened 2026-01-28
Extract and use shared ArrayDiff in unittests

kimgr:
ArrayDiff returns a string representation of the difference between two ArrayRefs, which may in turn wrap vectors, plain arrays, initializer lists, etc.

This is directly based on the VectorDiff in iwyu_include_picker_test.cc, and will replace it when/if I get around to enabling those tests.

The iwyu_string_util_test suite for various Split functions is much better this way.

### DIFF
--- a/unittests/iwyu_string_util_test.cc
+++ b/unittests/iwyu_string_util_test.cc
@@ -14,6 +14,7 @@
 #include <string>
 
 #include "gtest/gtest.h"
+#include "iwyu_test_helpers.h"
 
 namespace include_what_you_use {
 using std::string;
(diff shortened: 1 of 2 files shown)


kimgr:
@bolshakov-a Would love it if you could take this for a spin on your MSVC system. I don't trust my template specializations to be very portable :-/

bolshakov-a:
After this change, the unittests compile and pass:
```diff
--- a/unittests/iwyu_unittest_main.cc
+++ b/unittests/iwyu_unittest_main.cc
@@ -7,6 +7,7 @@
 //
 //===----------------------------------------------------------------------===//
 #if defined(_WIN32)
+#define NOMINMAX
 #include <windows.h>
```

PR #1915, opened 2026-01-29
Remove whitespace collapsing for forward-declares

kimgr:
I managed to fix this in Clang's DeclPrinter instead: https://github.com/llvm/llvm-project/commit/41079dac98ece04a3ee555049c5dd1d50523bb61

Once that lands in Debian packages, we can remove the whitespace cleanup in MungedForwardDeclareLineForTemplates and remove the now-unused function.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -510,9 +510,6 @@ string MungedForwardDeclareLineForTemplates(const TemplateDecl* decl) {
   // rather on the exact form printed by Clang, so it's safe to make certain
   // assumptions about format.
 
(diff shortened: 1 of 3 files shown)


kimgr:
Leaving this as a draft until the Debian packages have caught up.

Also note that my Clang patch made it onto the llvm-project `release/22.x` branch, and so is safe to merge to master before we take our release branch for LLVM 22.

kimgr:
Forgot about this before I took the branch. Will merge it on mainline when the upstream apt.llvm.org snapshot packages are stable.

PR #1916, opened 2026-01-29
Use new Clang SuppressDeclAttributes printing policy

kimgr:
With this policy we no longer need to rely on regex hacks to clean keyword attributes from forward declarations.

The new policy was added upstream in:
https://github.com/llvm/llvm-project/commit/4010cbb7cc3a7b48f958ac19d747a52abbd13190


kimgr:
Leaving this as a draft until the Debian packages have caught up.

Also note that the Clang patch DID NOT make it onto the llvm-project release/22.x branch, so we can't merge this until after we've taken the IWYU branch for LLVM 22.

PR #1917, opened 2026-01-29
Add function parameter pack items to resugar_map

bolshakov-a:
This fixes the case reported in https://github.com/include-what-you-use/include-what-you-use/issues/323#issuecomment-3819662241, because `std::visit` uses parameter pack for variant arguments.


kimgr:
Nice and clean, thanks!

PR #1921, opened 2026-01-30
Improve member call class type detection

### LINKED ISSUE #1061: False Positive: reports `<ratio>` is required when doing `std::chrono::duration` calculations
MRE:

```c++
#include <chrono>

auto calc(std::chrono::seconds d)
{
    return d + std::chrono::seconds(1);
}
```

IWYU reports `<ratio>` is required at the `+` operation (the source location reported by `-v3` and above points to the `+` character). This is an `operator +` between two `std::chrono::duration`s, but it also happens on other `+`s and `-`s between `std::chrono::duration`s and/or `std::chrono::time_point`s.

I suspect this problem comes from the construction of a "new" `std::ratio` type for the common measurement of two `duration`s; it will convert both `duration`s to this common measurement then add them up. (I quote "new" because the result type is constructed anew regardless of whether the result is identical to either one of them.)

If this is the case, I think this situation is kind of like `std::priority_queue` in that by default it implicitly uses another (independent) component as part of its template argument (like `std::priority_queue` uses `std::vector` by default), but also different in that this implicit usage is made "public" by the construction of similar type, while `std::priority_queue` doesn't derive other templated types from `std::vector`.

Possible workaround: If I'm not mistaken, although `<ratio>` is nominally an independent component, its use in standard library is almost exclusively in `<chrono>`. If one only uses `<chrono>` but not using `<ratio>` explicitly, one can map `<ratio>` to `<chrono>` in the custom mappings to workaround this.

bolshakov-a:
Prior to this, class template member functions were not scanned in some cases because clang may insert intermediate `ImplicitCastExpr` nodes which prevent detection of `TemplateSpecializationType` (it was desugared to `RecordType`).

The `#include` added to `iwyu_cache.cc` is for `NamedDecl` which is now reported due to `DynCastFrom` call; more precisely, from internals of the conversion function of its return type (there is a call to `ClassTemplateSpecializationDecl::classof(const Decl*)` inside it with `NamedDecl*` passed as an argument; conversion from `NamedDecl*` to `Decl*` requires complete type). Prior to this change, the conversion function was likely not considered as templated and hence not scanned.

Also, this fixes #1061 for some reason.


bolshakov-a:
> Also, this fixes https://github.com/include-what-you-use/include-what-you-use/issues/1061 for some reason.

Looks like `std::ratio` is now detected as provided by `TemplateSpecializationType` of the first operand as a default argument.

kimgr:
Very nice!

PR #1922, opened 2026-01-30
Add missing <bits/types/struct_rusage.h> mapping to <sys/resource.h>

hebasto:


iwyu_include_picker.cc
```diff
@@ -466,6 +466,7 @@ const IncludeMapEntry libc_include_map[] = {
   { "<bits/types.h>", kPrivate, "<sys/types.h>", kPublic },
   { "<bits/types/siginfo_t.h>", kPrivate, "<signal.h>", kPublic },
   { "<bits/types/siginfo_t.h>", kPrivate, "<sys/wait.h>", kPublic },
+  { "<bits/types/struct_rusage.h>", kPrivate, "<sys/resource.h>", kPublic },
```
kimgr, line 469: Hmm, this doesn't add up on my system; `struct_rusage.h` is only included by `sys/wait.h`, which does not appear to be included by `sys/resource.h`. Is your libc very new? I'm on glibc 2.39.

kimgr: Ah, I see. `sys/resource.h` includes `x86_64-linux-gnu/bits/resource.h`, which includes `bits/types/struct_rusage.h`.

And looking at the contents and [docs](https://man7.org/linux/man-pages/man2/getrusage.2.html), it makes perfect sense. Sorry about the confusion.

hebasto: glibc 2.42 on Fedora 43.

The include chain is:
`sys/resource.h` --> `bits/resource.h` --> `bits/types/struct_rusage.h`

Issue #1809, opened 2025-09-08
libc mapping lacks symbols from `sys/auxv.h`


hebasto:
For example, [`libc_symbol_map`](https://github.com/include-what-you-use/include-what-you-use/blob/bbf45363aef84ed9bed09b461ebf646b9ba2e34d/iwyu_include_picker.cc#L88) contains neither `AT_HWCAP` nor `AT_HWCAP2`, both of which are defined in `sys/auxv.h`.

kimgr:
Thanks for the report! The glibc symbol mapping is annoying in general, because there doesn't seem to be any canonical truth. Whatever is there seems to be a hand-edited shell pipeline from ~2011, and I don't have a good idea for something more rigorous.

We could just keep adding individual symbols as people report them, but it feels kinda sad.

PR #1923, opened 2026-01-30
Add symbol mappings for auxiliary vector entry types

hebasto:
This PR adds new symbol mappings for auxiliary vector entry types (as described in [`getauxval(3)`](https://www.man7.org/linux/man-pages/man3/getauxval.3.html)).

These symbols are defined in `elf.h`.

Fixes https://github.com/include-what-you-use/include-what-you-use/issues/1809.


kimgr:
Thank you!

PR #1924, opened 2026-01-31
Test non-member operator on templates with def arg

bolshakov-a:
aee0d831b5cc7abe76da491681642743af733d6e unintentionally fixed detection of provided default template arguments of template classes on handling calls to non-member templated operators (bug #1061), although that change was for member functions. This adds some dedicated test cases.

No functional change.


kimgr:
Thanks!

PR #1926, opened 2026-02-01
Make all library headers providing all transitives

bolshakov-a:
IWYU has some facilities to avoid exposing external library implementation details to users. Namely, `PublicHeaderIntendsToProvide` function is used to determine if a declaration used inside a template is meant to be provided by a template, especially when it is an external library template. Prior to this change, this usually meant that some of the declarations in the template instantiation chain should be in a public file, or in a file marked as private in some include mapping. Thus, that might not work when where is only a symbol mapping.

It was noticed in https://github.com/include-what-you-use/include-what-you-use/pull/1879#issuecomment-3708365409 that after 309f9be69030b708bc4193755ddd4985d0947b22 the reported internal stuff may be mapped to some other public header. This may be considered as a regression because those headers may provide more than needed. E.g., for the example in that comment, IWYU suggested to include `<memory>`. However, it was not included even transitively in the original code using MS STL.

The issue has been fixed by marking all the headers under any public header as providing all their transitive includes. (Prior to this, only the public headers themselves were considered to do it.) I think it is reasonable because an external library may include only their own headers or headers from some other external library.

Another way to fix the issue could probably be to check headers from symbol mappings inside `PublicHeaderIntendsToProvide`.

A couple of comments in `iwyu_include_picker.h` have been fixed on the way.

### DIFF
--- a/iwyu_preprocessor.cc
+++ b/iwyu_preprocessor.cc
@@ -992,7 +992,13 @@ void IwyuPreprocessorInfo::PopulateIntendsToProvideMap() {
       continue;
     intends_to_provide_map_[file].insert(file);  // Everyone provides itself!
     if (picker.IsPublic(file)) {
-      AddAllIncludesAsFileEntries(file, &intends_to_provide_map_[file]);
+      // It is assumed that every external library header (public or private)
+      // provides all its transitively included headers.
+      for (OptionalFileEntryRef included_file_or_self :
+           transitive_include_map_.at(file)) {
+        InsertAllInto(transitive_include_map_.at(included_file_or_self),
+                      &intends_to_provide_map_[included_file_or_self]);
+      }
     } else {
       const set<OptionalFileEntryRef>& direct_includes =
           fileinfo.second.direct_includes_as_fileentries();
@@ -1032,6 +1038,8 @@ void IwyuPreprocessorInfo::PopulateIntendsToProvideMap() {
   //   a templated function or class in i1.h, you see the need for
   //   symbol Foo which isn't a template argument, don't worry about
   //   it.'  Double check whether that's true.
+  // TODO(bolshakov): is it still needed after making private includes providing
+  // all their transitively included headers?
   for (const auto& header_map : private_headers_behind) {
     OptionalFileEntryRef public_header = header_map.first;
     for (OptionalFileEntryRef private_header : header_map.second) {
@@ -1072,8 +1080,9 @@ void IwyuPreprocessorInfo::HandlePreprocessingDone() {
   }
   MutableGlobalIncludePicker()->FinalizeAddedIncludes();
   FinalizeProtectedIncludes();
-  PopulateIntendsToProvideMap();
+  // PopulateIntendsToProvideMap uses results of PopulateTransitiveIncludeMap.
   PopulateTransitiveIncludeMap();
+  PopulateIntendsToProvideMap();
 }
 
 bool IwyuPreprocessorInfo::BelongsToMainCompilationUnit(
--- a/iwyu_include_picker.h
+++ b/iwyu_include_picker.h
@@ -281,12 +281,12 @@ class IncludePicker {
 
   // A map of all quoted-includes to whether they're public or private.
   // Files whose visibility cannot be determined by this map nor the one
-  // below are assumed public.
+  // below may be assumed public (see GetVisibility usage).
   VisibilityMap include_visibility_map_;
 
   // A map of paths to whether they're public or private.
   // Files whose visibility cannot be determined by this map nor the one
-  // above are assumed public.
+  // above may be assumed public (see GetVisibility usage).
   // The include_visibility_map_ takes priority over this one.
   VisibilityMap path_visibility_map_;
 
--- a/tests/cxx/block_template_internals-d1.h
+++ b/tests/cxx/block_template_internals-d1.h
@@ -1,0 +1,10 @@
+//===--- block_template_internals-d1.h - test input file for iwyu ---------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/block_template_internals-i1.h"
--- a/tests/cxx/block_template_internals-i1.h
+++ b/tests/cxx/block_template_internals-i1.h
@@ -1,0 +1,15 @@
+//===--- block_template_internals-i1.h - test input file for iwyu ---------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/block_template_internals-i2.h"
+
+template <typename T>
+int TplFn(T t) {
+  return InternalFn(t);
+}
--- a/tests/cxx/block_template_internals-i2.h
+++ b/tests/cxx/block_template_internals-i2.h
@@ -1,0 +1,10 @@
+//===--- block_template_internals-i2.h - test input file for iwyu ---------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/block_template_internals-i3.h"
--- a/tests/cxx/block_template_internals-i3.h
+++ b/tests/cxx/block_template_internals-i3.h
@@ -1,0 +1,10 @@
+//===--- block_template_internals-i3.h - test input file for iwyu ---------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+int InternalFn(int);
--- a/tests/cxx/block_template_internals.cc
+++ b/tests/cxx/block_template_internals.cc
@@ -1,0 +1,25 @@
+//===--- block_template_internals.cc - test input file for iwyu -----------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --mapping_file=tests/cxx/block_template_internals.imp -I .
+
+// Test that IWYU does not report template implementation details, especially
+// for templates from external libraries.
+
+#include "tests/cxx/block_template_internals-d1.h"
+
+// IWYU should not report InternalFn here. It is assumed to be provided
+// by the public header for TplFn.
+int x = TplFn(1);
+
+/**** IWYU_SUMMARY
+
+(tests/cxx/block_template_internals.cc has correct #includes/fwd-decls)
+
+***** IWYU_SUMMARY */
--- a/tests/cxx/block_template_internals.imp
+++ b/tests/cxx/block_template_internals.imp
@@ -1,0 +1,3 @@
+[
+  { "symbol": ["TplFn", "private", "\"tests/cxx/block_template_internals-d1.h\"", "public"] },
+]


kimgr:
> The issue has been fixed by marking all the headers under any public header as providing all their transitive includes. (Prior to this, only public headers were considered to do it.)

This seems to be circular, is there a typo in there?

bolshakov-a:
I don't see. I mean: before this change, only headers marked as public were considered to provide all their transitively included headers, whereas private headers behind them were considered to provide only their directly included headers (which may in turn appear to be public and provide all their own transitive `#include`s, but this is irrelevant). After this change, all external library headers, both marked as public in mappings and private headers included by them, are considered to provide all their transitively included headers.

Look at the test case. If you move `TplFn` definition from `-i1.h` to `-d1.h`, the test passes on master because `-d1.h` is marked as a public header in the mapping file and hence is considered to provide `-i3.h`, but `-i1.h` header included by `-d1.h` was not considered to do it.

kimgr:
Ah, thank you, now I see!

Can I suggest something like:

> The issue has been fixed by marking all the headers under any public header as providing all their transitive includes. (Prior to this, only the public headers themselves were considered to do it.)

?

iwyu_preprocessor.cc
```diff
@@ -959,6 +959,16 @@ void IwyuPreprocessorInfo::AddAllIncludesAsFileEntries(
   }
 }
 
+void IwyuPreprocessorInfo::
+    MakeAllTransitivelyIncludedProvideTheirOwnTransitives(
```
kimgr: I wonder if it would be nicer to just inline the method to avoid the long and complicated name? It's only called once, so it makes sense to me.

kimgr:
So does this only affect headers explicitly marked public in mappings? I guess that makes sense, but I wonder to what extent it masks bugs in IWYU's core analysis of types/decls/uses?

For example, will IWYU be able to suggest `<cstddef>` if someone is including `<string>` but only using `std::size_t`? I suppose in that case symbol mappings take over, but how does it behave if we don't have symbol mappings (most mappers only produce include mappings)?

Does this make it harder to debug template spill bugs? I suppose we can still use the verbose logging to see where templates are reported even if IWYU does not complain in the end.

Oh, I see now that this is only used from template analysis? I guess that narrows the blast radius a bit.

bolshakov-a:
> Oh, I see now that this is only used from template analysis?

Yeah, should not affect `std::size_t` handling.

bolshakov-a:
> Can I suggest something like:

Done.

kimgr:
Right, size_t was a bad example. But it seems like any template in a library will be provided by whatever pubic header is already included? Or is that an oversimplification?

bolshakov-a:
> But it seems like any template in a library will be provided by whatever pubic header is already included?

It already is (and not necessarily a template but e.g. an ordinary function like `InternalFn` from test), if the template on the top of the template instantiation stack appears to be in a public header, or in a header explicitly marked as private in an include mapping. See [the `for` loop](https://github.com/include-what-you-use/include-what-you-use/pull/1926/files#diff-0f5b5adee052d338a07c9a285158ade85b6ba3e61a1c85ad5312030ffd560aa4R1043) after my TODO, or just try to add the following line to the mapping on master:
```yaml
{ "include": ["\"tests/cxx/block_template_internals-i1.h\"", "private", "\"tests/cxx/block_template_internals-d1.h\"", "public"] },
```

bolshakov-a:
So in fact, this PR makes internal library headers provide _even less_ than if they were explicitly marked as private in an include mapping, because in the latter case a private header is considered to provide all transitively included headers _of its public header(s)_.

bolshakov-a:
For example, imagine you have the following library with two public interface headers:
```cpp
// guts.h
#pragma once

template <typename Fun, typename Arg>
void Call(Fun fn, const Arg& arg) {
    fn(arg);
}

class OtherClass;
void CustomizationPoint(const OtherClass&);

// intervening.h
#include "guts.h"

// tpl_fn_priv.h
#include "intervening.h"

template <typename T>
void TplFn(const T& t) {
    Call([](auto x) { CustomizationPoint(x); }, t);
}

// public1.h
#include "tpl_fn_priv.h"

class Class {};

// predef_custom_publ.h
// Provides some predefined customization points.

#include "guts.h"

class Class;
void CustomizationPoint(const Class&);
```
so you have the following `#include` graph:
scheme
If you use the library like this:
```cpp
#include "public1.h"
#include "predef_custom_publ.h"

void User() {
    Class c;
    TplFn(c);
}
```
with a mapping file that maps all the private headers to the public ones, like this:
```yaml
[
  {"include": ["\"tpl_fn_priv.h\"", "private", "\"public1.h\"", "public"]},
  {"include": ["\"intervening.h\"", "private", "\"public1.h\"", "public"]},
  {"include": ["\"guts.h\"", "private", "\"public1.h\"", "public"]},
  {"include": ["\"guts.h\"", "private", "\"predef_custom_publ.h\"", "public"]},
]
```
then `guts.h` is considered to provide the union of all the headers transitively included into `public1.h` and `predef_custom_publ.h`, i.e. all the library. Hence, IWYU erroneously gives:
```

- #include "predef_custom_publ.h"  // lines 2-2

#include "public1.h"  // for Class, TplFn
```
After applying this suggestion, the code doesn't compile because it lacks `CustomizationPoint(const Class&)`.

This is what Craig wrote about in the comment. This PR doesn't fix this bug. However, if you _replace_ the include mapping with the symbol mapping:
```yaml
[
  {"symbol": ["TplFn", "private", "\"public1.h\"", "public"]},
  {"symbol": ["CustomizationPoint", "private", "\"predef_custom_publ.h\"", "public"]},
]
```
the output on my branch is correct.

kimgr:
Thank you for the nice example!

> This is what Craig wrote about in the comment. This PR doesn't fix this bug. However, if you replace the include mapping with the symbol mapping:
>
> ```
> [
>  {"symbol": ["TplFn", "private", "\"public1.h\"", "public"]},
>  {"symbol": ["CustomizationPoint", "private", "\"predef_custom_publ.h\"", "public"]},
> ]
> ```
> the output on my branch is correct.

And what is the behavior on master if this symbol mapping is present?

bolshakov-a:
> And what is the behavior on master if this symbol mapping is present?

It suggests to include `guts.h` for `Call`:
```
#include "guts.h"                // for Call

#include "guts.h"                // for Call
#include "predef_custom_publ.h"  // for CustomizationPoint
#include "public1.h"             // for Class, TplFn
```
because `tpl_fn_priv.h` is not considered to provide `Call` due to the presence of `intervening.h` and because `tpl_fn_priv.h` is not mapped to a public header.

In fact, the correct behavior may be achieved on master by means of removing `guts.h` entries from the include map, but this is counterintuitive.

kimgr:
Thanks, that's good contrast.

I was still having a bit of diffculty seeing how the explicit loop over transitive includes is different from the recursive `AddAllIncludesAsFileEntries`. I tried adding some logging, and the following came out:

```diff
diff --git a/master.txt b/branch.txt
index ab59cf4f..24f1f012 100644
--- a/master.txt
+++ b/branch.txt
@@ -6,9 +6,16 @@ AddAllIncludesAsFileEntries: adding: tests/cxx/block_template_internals.cc -> te
 AddAllIncludesAsFileEntries: adding: tests/cxx/block_template_internals.cc -> tests/cxx/block_template_internals-i2.h
 AddAllIncludesAsFileEntries: adding: tests/cxx/block_template_internals.cc -> tests/cxx/block_template_internals-i3.h
 PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-d1.h -> tests/cxx/block_template_internals-d1.h
-AddAllIncludesAsFileEntries: adding: tests/cxx/block_template_internals-d1.h -> tests/cxx/block_template_internals-i1.h
-AddAllIncludesAsFileEntries: adding: tests/cxx/block_template_internals-d1.h -> tests/cxx/block_template_internals-i2.h
-AddAllIncludesAsFileEntries: adding: tests/cxx/block_template_internals-d1.h -> tests/cxx/block_template_internals-i3.h
+PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-d1.h -> tests/cxx/block_template_internals-d1.h
+PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-d1.h -> tests/cxx/block_template_internals-i1.h
+PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-d1.h -> tests/cxx/block_template_internals-i2.h
+PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-d1.h -> tests/cxx/block_template_internals-i3.h
+PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-i1.h -> tests/cxx/block_template_internals-i1.h
+PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-i1.h -> tests/cxx/block_template_internals-i2.h
+PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-i1.h -> tests/cxx/block_template_internals-i3.h
+PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-i2.h -> tests/cxx/block_template_internals-i2.h
+PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-i2.h -> tests/cxx/block_template_internals-i3.h
+PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-i3.h -> tests/cxx/block_template_internals-i3.h
 PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-i1.h -> tests/cxx/block_template_internals-i1.h
 PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-i1.h -> tests/cxx/block_template_internals-i2.h
 PopulateIntendsToProvideMap: adding tests/cxx/block_template_internals-i2.h -> tests/cxx/block_template_internals-i2.h
@@ -28,55 +35,17 @@ Intends-to-provide for tests/cxx/block_template_internals-d1.h:
 Intends-to-provide for tests/cxx/block_template_internals-i1.h:
    tests/cxx/block_template_internals-i1.h
    tests/cxx/block_template_internals-i2.h
+   tests/cxx/block_template_internals-i3.h
 Intends-to-provide for tests/cxx/block_template_internals-i2.h:
    tests/cxx/block_template_internals-i2.h
    tests/cxx/block_template_internals-i3.h
 Intends-to-provide for tests/cxx/block_template_internals-i3.h:
    tests/cxx/block_template_internals-i3.h
...
```
So there seems to be a lot of new provision mappings between the intermediary headers, but somehow the only effective difference is that `-i1.h` now provides `-i3.h`. I think it mirrors your `public1.h`, `intervening.h`, `guts.h` example.

I'm not questioning that it works better, but it's quite a subtle difference with big impact. I wonder if there are tradeoffs worth mentioning in the commit message, for example if there are cases where this new policy performs worse.

I think the whole "intends-to-provide" mechanism rubs me the wrong way, because it's a rather blunt mechanism to mask deeper attribution problems.

bolshakov-a:
> So there seems to be a lot of new provision mappings between the intermediary headers, but somehow the only effective difference is that `-i1.h` now provides `-i3.h`.

Direct `#include` mappings of internal (non-public) headers are just duplicated later.

> I wonder if there are tradeoffs worth mentioning in the commit message, for example if there are cases where this new policy performs worse.

I'd say that it can make worse if some customization point function appears to be indirectly included in some private header just accidentally, so it should be explicitly included by users so as not to be broken with future library releases... However, the master branch shows the similar behavior anyway if the template on top of an instantiation stack (`TplFn`) appears to be in its documented public header, because headers marked as public are already considered to provide all its transitively included headers. It looks obvious to me that IWYU behavior should not depend on whether a template is placed in its public header or in some internal header behind it. So, I don't still see any significant drawback of the proposed change. On the contrary, it should make IWYU behavior more stable and predictable, by improving suppression of library internals, in accordance with intention of the original IWYU authors, I suppose...

> I think the whole "intends-to-provide" mechanism rubs me the wrong way, because it's a rather blunt mechanism to mask deeper attribution problems.

It is rather just an attempt to cope with the fact that many libraries which are not under user's control don't follow the IWYU principle.

bolshakov-a:
> I think the whole "intends-to-provide" mechanism rubs me the wrong way, because it's a rather blunt mechanism to mask deeper attribution problems.

I also have the following thought. A strict implementation of the IWYU principle would lead to exposition of lots of external library implementation details to user's code. As I have said, many libraries don't follow the IWYU principle and include some stuff (which is not intended to be explicitly included by users) not directly into their headers but "somehow". IWYU tries to detect it and to avoid annoying implementation detail suggestions. These are not critical bugs because such implementation detail headers should be already included in user's translation units anyway through some `#include` chain, but it looks like IWYU reports some garbage.

For types, IWYU has the "author intent" mechanism that checks not only the presence of a direct `#include` but also the absence of any forward-declaration (if there is a forward-declaration and no direct `#include`, then the library authors probably want users to include the type explicitly; otherwise the default expectation is that the authors provide the type). However, for functions,  there is no forward-declarations. You cannot even redeclare a function if you don't know exact parameter types. So the only way to make sure that the library authors don't want to provide a function used in some dependent expression inside some library template is to check all the `#include` graph for the absence of a declaration of that function. (Note that, for types, IWYU currently reports only those types which are specified in the template arguments).

kimgr:
> I also have the following thought. A strict implementation of the IWYU principle would lead to exposition of lots of external library implementation details to user's code.

Yeah, definitely. But I think it would be nicer if the mechanism to work around that was mappings (maybe in a more advanced form than what we currently have), not hardcoded mapping-like rules like "intends-to-provide". This is not a criticism of this patch, just a nagging feeling that we're doing similar things in more than one way.

Anyway, you've helped me get my head around this and since there seem to be no obvious downsides, it definitely LGTM! Thank you!

bolshakov-a:
Maybe you are right, but writing mapping files is already a burden for users. And library implementation details may be unstable, hence mappings for them would be fragile. Or I haven't got your idea.

PR #1927, opened 2026-02-02
Add `timeval` symbol mappings for alternative headers

hebasto:
According to [POSIX](https://www.man7.org/linux/man-pages/man3/timeval.3type.html), in addition to `<sys/time.h>`, `struct timeval` is also provided by the following headers: `<sys/resource.h>`, `<sys/select.h>`, and `<utmpx.h>`.

This prevents IWYU from suggesting including `<sys/time.h>` when `timeval` is already available via these headers.

Noticed in https://github.com/bitcoin/bitcoin/pull/34448#discussion_r2754677021.


kimgr:
Thank you!

PR #1928, opened 2026-02-03
Add symbol mappings for fastest and smallest integer types

hebasto:
On master (commit 9f560199e8080d2a7051b9f94e19f2530e5e2a5d), IWYU incorrectly suggests including an internal glibc header:
```
$ cat test.c
#include <stdint.h>

int main(void)
{
    int_least64_t i = 42;
    return i;
}
$ include-what-you-use test.c 

#include <bits/stdint-least.h>  // for int_least64_t

- #include <stdint.h>  // lines 1-1

#include <bits/stdint-least.h>  // for int_least64_t
---
```

This PR fixes this issue.


bolshakov-a:
I think that the symbol mapping for the C standard library is needed anyway. However, I'm not sure if I do that before the upcoming release.

hebasto:
> I think that the symbol mapping for the C standard library is needed anyway.

Thanks! Reworked.

> However, I'm not sure if I do that before the upcoming release.

No worries. Our [project](https://github.com/bitcoin/bitcoin) maintains its own patch set and mappings regardless.

bolshakov-a:
LGTM. I think that symbol mappings should in general be preferred over include mappings because they are cross-platform and don't require any knowledge about details of a specific library implementation which may change in the future. Btw, @amurzeau managed to obtain a symbol mapping for C18 several years ago: https://github.com/include-what-you-use/include-what-you-use/issues/785#issuecomment-615524242.

kimgr:
Thank you!

PR #1929, opened 2026-02-04
Suppress inline namespaces completely

bolshakov-a:
This is required for standard library symbol mappings because implementers may place declarations into some unspecified inline namespaces inside `std`. Moreover, this makes IWYU more stable because, prior to this, it put or did not put inline namespace names into symbol names depending on whether there is any name conflict or not.


kimgr:
Neat!

PR #1939, opened 2026-03-09
Ensure --dry_run returns the exit code according to its documentation

nmushtak-netflix:
In the documentation of `fix_includes.py`, it says

    -n, --dry_run         Do not actually edit any files; just print diffs. Return code is 0 if no changes are needed,
                          else min(the number of files that would be modified, 100)

However, `main` always returns the value 0, so this non-zero exit code is not actually implemented in `fix_includes.py`. This PR implements a fix so `main` returns a non-zero exit code when files would be modified according to the above documentation, and adds two unit tests to ensure the return value of `main` works as expected.

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -2484,14 +2484,17 @@ def main(argv):
   if flags.update_comments:
     flags.comments = True
 
+  exit_code = 0
   if flags.sort_only:
     if not files_to_modify:
       sys.exit('FATAL ERROR: -s flag requires a list of filenames')
     SortIncludesInFiles(files_to_modify, flags)
   else:
-    ProcessIWYUOutput(sys.stdin, files_to_modify, flags, cwd=os.getcwd())
+    process_ret = ProcessIWYUOutput(sys.stdin, files_to_modify, flags, cwd=os.getcwd())
+    if flags.dry_run:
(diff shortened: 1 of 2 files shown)


kimgr:
@nmushtak-netflix Thanks! I see I changed this policy in d0ad0858932d4c3d6fc1d027a78d5fcae759dcb6, but forgot to update the help text.

I wonder, is there any reason to have dry-run behave differently from live runs with regard to the exit code? I mean, do you have a use case for this behavior, or could we just change the dry-run help text?

nmushtak-netflix:
I would like to use `fix_includes.py` in CI checks, and for my CI to detect when `fix_includes.py` would need to make a change, it would be helpful for me to have the behavior where `--dry_run` outputs a nonzero exit code if a change would be made.

kimgr:
I see, makes sense. But you don't really care about the number of files edited, right? I'm thinking we could simplify that thing a bit, but I can do that as a followup.

nmushtak-netflix:
Yes the number of files edited doesn't matter for the exit code, I only did that because that is what the help text says

kimgr:
I might get around to simplifying this further then. But this PR LGTM!

Issue #1938, opened 2026-03-09
Assertion failed: There should be a redecl specifying the default arg


martin-eder-zeiss:
Hi,

The current `master` branch triggers the following assertion:

```
/tmp/iwyu/iwyu_ast_util.cc:1545: Assertion failed: There should be a redecl specifying the default arg
```

Minimal reproducer:

```cpp
template<typename T>
void
f(int = 0);

template<typename T>
void
f(int)
{
}

int
main()
{
  f<char>();
  return 0;
}
```

kimgr:
Thanks! Will make sure to get that fixed before we take the release branch.

bolshakov-a:
#1940 fixes the crash. However, there remains a case with wrong behavior when there is no template definition in the translation unit, like this:
```cpp
// f-def_arg.h
template <typename>
void f(int = 0);

// f.h
#include "f-def_arg.h"

template <typename>
void f(int);

// main.cpp
#include "f.h"

int main() {
  f<char>();
}
```
```
f.h should add these lines:

f.h should remove these lines:
- #include "f-def_arg.h"  // lines 1-1

The full include-list for f.h:
---

(main.cpp has correct #includes/fwd-decls)
```
The code doesn't compile after applying the suggestion. Remapping of the implicit instantiation to the first template declaration inside `GetRedeclSpecifyingDefArg` would probably be wrong because even in this case the implicit instantiation should be scanned because a template parameter may be fully used in a function template signature like this:
```cpp
template <typename T>
T::Internal f(int = 0);
```

kimgr:
Closing this as fixed, feel free to create a new issue from the missing-definition comment (`...` -> `Reference in a new issue`)

PR #1940, opened 2026-03-10
Fix crash on function template with default args

bolshakov-a:
`GetRedeclSpecifyingDefArg` returns now the given decl for function template specializations because it should anyway be passed into `HandleFunctionCall` for scanning implicit instantiations and reporting the complete template definition for them. Reporting the template redeclaration specifying the default argument at the call site is questionable because it should be the first redeclaration in a translation unit and hence the resulting code may not compile depending on the order of include directives.

This fixes the crash reported in #1938.


kimgr:
Thanks!

PR #1949, opened 2026-03-13
Add standard overloaded functions to mapping

bolshakov-a:
`OVERLOADED_FUNCS` set has been kept in the script to minimize changes in the mapping. This has a drawback that the set should be maintained manually on updating to newer revisions of the standard.


bolshakov-a:
Now the output for [my example](https://github.com/include-what-you-use/include-what-you-use/pull/1879#issuecomment-3708365409) is almost ideal:
```
main.cpp:7:9: warning: std::move(:0 &&) is defined in <utility>, which isn't directly #included.
main.cpp:8:13: warning: std::begin(:0 &) is defined in <iterator>, which isn't directly #included.
main.cpp:8:24: warning: std::end(:0 &) is defined in <iterator>, which isn't directly #included.
main.cpp:8:33: warning: std::begin(:0 &) is defined in <iterator>, which isn't directly #included.
main.cpp:10:12: warning: std::_String_iterator is defined in <xstring>, which isn't directly #included.

#include <iterator>   // for begin, end
#include <utility>    // for move
#include <xstring>    // for _String_iterator

#include <algorithm>  // for move, remove
#include <iterator>   // for begin, end
#include <string>     // for string, erase, swap
#include <utility>    // for move
#include <xstring>    // for _String_iterator
```

kimgr:
I just want to say that this is very cool, and a significant body of work! Having IWYU behave portably based on generated symbol mappings is a milestone. Thank you!

bolshakov-a:
> Having IWYU behave portably based on generated symbol mappings is a milestone.

Note that, besides operator mappings, there still remain [[input.output]](https://eel.is/c++draft/#input.output) entities unmapped, as well as macros and significant part of the C standard library.

Issue #1629, opened 2024-09-19
False Positive: Include required for default template parameter not explicitly used


carlosgalvezp:
Consider the following example:

```
$ cat foo.h

#include <vector>

template <typename T, typename Alloc = typename std::vector<T>::allocator_type>
class Vector final : private std::vector<T, Alloc>
{
 public:
    Vector() = default;
};
```

```
$ cat main.cpp

#include "foo.h"

int main()
{
    Vector<int> v;
}
```

Running IWYU in C++14 mode gives what I believe is a FP, or at least undesirable:

```
$ include-what-you-use -std=c++14 main.cpp 

main.cpp should add these lines:
#include <memory>  // for allocator

main.cpp should remove these lines:

The full include-list for main.cpp:
#include <memory>  // for allocator
#include "foo.h"   // for Vector
---
```

The issue goes away in C++17. This happens on latest trunk.

carlosgalvezp:
@kimgr Did you have a chance to look into this? :)

kimgr:
@carlosgalvezp Unfortunately not, life got in the way. Will have to get 0.23 out the door, then maybe look into bugs.

carlosgalvezp:
Understandable, thank you!

carlosgalvezp:
I've looked a bit into this, and if I remove this part of the code, I no longer get warnings:
https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_ast_util.cc#L1620

Obviously this might lead to False Negatives. I don't recall what command to use to run all the unit tests, what was it?

kimgr:
`./run_iwyu_tests.py` assuming `include-what-you-use` is on the PATH.

akien-mga:
I believe I have a similar issue with a default template parameter, though in my case it's not resolved by using C++17 (or later).

See the attached reproduction project: [iwyu-template-default.zip](https://github.com/user-attachments/files/25868574/iwyu-template-default.zip)

That simplified test doesn't make too much sense but it's a mix of @carlosgalvezp's test and Godot's `HashMap` implementation ([`hash_map.h`](https://github.com/godotengine/godot/blob/c7dd32929369947876dbf0fdf798b0743b01c7ba/core/templates/hash_map.h#L55-L69)).

[...]

PR #1950, opened 2026-03-14
Take account of template base specifier location

bolshakov-a:
It is needed to determine dependent base provision status correctly inside `GetLocOfTemplateThatProvides`. Otherwise, the template instantiation ("caller") location is erroneously considered as the base specifier location, which may cause exposition of implementation details to users.

Currently, however, a base type should be mentioned in the template arguments so that IWYU can report it at the instantiation site, ~~so the fix is hardly useful~~. But it may change in the future.

The test case with a default template argument has been placed into `default_tpl_arg` test and not into `block_template_internals` so as not to be spoiled by the fact that `block_template_internals-d1.h` is public.

This fixes https://github.com/include-what-you-use/include-what-you-use/issues/1629.


tests/cxx/block_template_internals-d1.h
```diff
@@ -8,3 +8,6 @@
 //===----------------------------------------------------------------------===//
 
 #include "tests/cxx/block_template_internals-i1.h"
+
+template <typename T1, typename T2>
+class DerivedTpl : BaseTpl<T1> {};
```
kimgr: Is it worth spelling out `private` here? As far as I can tell from the code change, both public and private bases are treated the same, but it might be nice to be more explicit.

tests/cxx/default_tpl_arg-d2.h
```diff
@@ -52,6 +52,9 @@ class TplProvidingDefArg {
 template <typename T = IndirectClass>
 struct DerivedTplProvidingDefArg : TplProvidingDefArg<T> {};
 
+template <typename T1, typename T2 = IndirectTemplate<T1>>
+class DerivedFromProvidedDefArg : T2 {};
```
kimgr: public/private might be nice here too

tests/cxx/block_template_internals-d1.h
bolshakov-a: It is irrelevant to the bug. If you insist, I'd prefer to put `public` here because it looks like a more common case.

kimgr: Yeah, let's do `public` then, to keep the code mainstream. Private bases is something I would think twice about if I saw it in live code, so I don't want to make it a distraction in the tests if it has no impact on behavior.

PR #1955, opened 2026-03-14
Add standard operator mapping

bolshakov-a:


bolshakov-a:
Perhaps, some tests should be added for literal operators? But this would involve some complicated chain of supporting definitions (like `std::ratio` and `std::chrono::duration` for `operator""s` returning `std::chrono::seconds`, or `std::char_traits` and `std::allocator` for `std::string`) to be consistent with the standard.

bolshakov-a:
Nevermind. Looks like IWYU doesn't support user-defined literals at all.

kimgr:
So I already took the `clang_22` branch yesterday. There's no hard reason to cherry-pick this onto the branch too, right? It might be nice to have the extra support for operators, but it can also wait until 0.27?

bolshakov-a:
It is slightly sad that IWYU would probably still suggest `<xstring>` for (`str1 + str2`) alongside `<string>` with MS STL, but yeah, it may be postponed. Anyway, string iterators still cause `<xstring>` suggestion, and significant parts of standard libraries are not still mapped, as I noted.

kimgr:
Yeah, then let's continue with the release branch as-is, gotta make the cut somewhere. Thanks!

kimgr:
Very cool, thanks!

PR #1959, opened 2026-03-22
Find provided types behind sugar on construction

bolshakov-a:
This improves provided type determination for source types of copy or move construction inside `SourceOrTargetTypeIsProvided`.


kimgr:
Nice, thanks!

PR #1963, opened 2026-03-25
Scan alias templates in GetProvidedTypeComponents

bolshakov-a:
Getting provided type components for top-level type has been moved out from `GetTplInstDataForClass` so as to avoid infinite recursion `GetTplInstData` -> `GetTplInstDataForClass` -> `GetProvidedTypeComponents` -> `GetProvidedByTplArg` -> `GetTplInstData`.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3434,6 +3434,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     return retval;
   }
 
+  set<const Type*> GetAliasTemplateProvidedTypes(
+      const TemplateSpecializationType* type,
+      const TypeAliasTemplateDecl* al_tpl_decl) const {
+    const TypeAliasDecl* al_decl = al_tpl_decl->getTemplatedDecl();
+    set<const Type*> res = GetProvidedTypes(type->getAliasedType().getTypePtr(),
+                                            GetLocation(al_decl));
+    RemoveAllFrom(GetCanonicalArgComponents(type), &res);
+    return res;
+  }
+
   set<const Type*> blocked_types_;
   set<const Decl*> blocked_for_fwd_decl_;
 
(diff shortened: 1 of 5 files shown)


tests/cxx/alias_template.cc
```diff
@@ -90,6 +90,21 @@ struct TplWithUsingArgInternals {
...
+
+void TestProvisionByTplArg() {
+  // IWYU: IndirectClass needs a declaration
+  // IWYU: IndirectClass is...*indirect.h
+  Identity<IndirectClass> iic;
+  // IWYU: IndirectClass is...*indirect.h
+  TakeByCopyInTplFn(iic);
+
+  Identity<Providing> ip;
+  // Test collecting provided types for SourceOrTargetTypeIsProvided function.
```
bolshakov-a, line 104: I'm not sure about referring to the IWYU source in tests. Source code may change, and the comment would become outdated.

iwyu.cc
```diff
@@ -3434,6 +3434,16 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
 
+  set<const Type*> GetAliasTemplateProvidedTypes(
+      const TemplateSpecializationType* type,
+      const TypeAliasTemplateDecl* al_tpl_decl) const {
+    const TypeAliasDecl* al_decl = al_tpl_decl->getTemplatedDecl();
+    set<const Type*> res = GetProvidedTypes(type->getAliasedType().getTypePtr(),
+                                            GetLocation(al_decl));
+    RemoveAllFrom(GetCanonicalArgComponents(type), &res);
+    return res;
+  }
```
bolshakov-a, line 3445: Just a move from `private` to `protected` section.

tests/cxx/alias_template.cc
kimgr: Yeah, it feels a little fragile. But it's also very useful when coming in from the outside trying to understand the role of a function. I think it's good to keep.

PR #1964, opened 2026-03-28
[cmake] Use vanilla add_library for static libraries

### LINKED ISSUE #1961: [0.26] Missing libiwyu.so.22.1 library
The `include-what-you-use` binary builds and links with `libiwyu.so.22.1`, but the CMake doesn't install it, resulting in errors when attempting to run the installed package:

```
can't install iwyu:
  - nothing provides libiwyu.so.22.1()(64bit) needed by iwyu-0.26-1.fc45.x86_64
```

The regression introduced by https://github.com/include-what-you-use/include-what-you-use/commit/9c7a6a3c08af3ecfc6e0a8716455435c696b7bb2 and https://github.com/include-what-you-use/include-what-you-use/commit/e9a08a9e6cb32eff8120473ccc51e035d5942451 commits.

Full ldd output:
```
$  ldd ./include-what-you-use
        linux-vdso.so.1 (0x00007f511846c000)
        libiwyu.so.22.1 => not found
        libLLVM.so.22.1 => not found
        libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f5118000000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f5117e0d000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f511834e000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f5118320000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f511846e000)
```

> We didn't plan to install these libraries, they are only for internal use.

Then these libraries need to be `STATIC` or `OBJECT` to prevent linking against them dynamically.

UPD. Fedora build system always explicitly sets the `-DBUILD_SHARED_LIBS:BOOL=ON` build flag. This may cause `BUILDTREE_ONLY` to be treated as `SHARED`.
UPD2. Explicitly setting the `-DBUILD_SHARED_LIBS:BOOL=OFF` can be used as a workaround. Most Linux distributions will need it.

kimgr:
The add_llvm_library macro has a lot of customization and
parameterization to fit libraries into the LLVM conventions.

Unfortunately the macro forces libraries to be built as SHARED if the
BUILD_SHARED_LIBS option is on, no matter if we explicitly specify
STATIC. This in turn led to packaging failures, because the shared
libraries were not installed.

Switch to using vanilla add_library, and make the libraries OBJECT
libraries, which is the cheapest kind (we have no use for proper static
libraries either). Use the llvm_update_compile_flags functions to make
exception and RTTI flags consistent with LLVM.

This reverts b828e17 and also e9a08a9, which is no longer necessary.

Fixes #1961.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,8 @@ set(LLVM_LINK_COMPONENTS
   AllTargetsInfos
 )
 
-add_llvm_library(iwyu BUILDTREE_ONLY
+add_library(iwyu
+  OBJECT
   iwyu.cc
   iwyu_ast_util.cc
   iwyu_cache.cc
@@ -152,6 +153,7 @@ add_llvm_library(iwyu BUILDTREE_ONLY
   iwyu_regex.cc
   iwyu_verrs.cc
 )
+llvm_update_compile_flags(iwyu)
 
 add_llvm_executable(include-what-you-use
   iwyu_main.cc
@@ -233,9 +235,11 @@ if (WIN32)
 endif()
 
 # Build vendored gtest library
-add_llvm_library(iwyu-gtest BUILDTREE_ONLY
+add_library(iwyu-gtest
+  OBJECT
   vendor/googletest/src/gtest-all.cc
 )
+llvm_update_compile_flags(iwyu-gtest)
 
 target_include_directories(iwyu-gtest
   PUBLIC


kimgr:
@bolshakov-a Do you still have a Windows/MSVC environment where you could try this? I'm not sure how OBJECT libraries are translated in MSVC style. Thanks!

bolshakov-a:
> I'm not sure how OBJECT libraries are translated in MSVC style.

Looks like they become just static libraries (`.lib`).

We have to wait until the actual clang is built on my Windows machine.

bolshakov-a:
Hey @kimgr, `add_library` were deliberately replaced with `add_llvm_library` due to build failures on my Debian-12 machine with gcc-12.2.0: https://github.com/include-what-you-use/include-what-you-use/commit/b828e1741656a7bd3eec8b9423577578477a6866. I have tried your branch on it, and those failures have come back.

kimgr:
@bolshakov-a Incredible. Good thing we have a Git log, my memory is not working anymore :slightly_smiling_face:. I'll see if I can get those flags on there some other way.

kimgr:
@bolshakov-a Could you try the latest rev of this branch on your Debian system?

bolshakov-a:
Sure. Now it works, unit-tests pass.

kimgr:
Also tested a build with `-DBUILD_SHARED_LIBS=on`, and no shared libraries were produced/installed.

PR #1966, opened 2026-03-28
Collect argument types from nested name specifiers

bolshakov-a:
Prior to this, IWYU didn't report `Arg1` for instantiated member templates like `Host<Arg1>::Nested<Arg2>` when `Arg1` was fully used inside `Nested` template but not inside `Host` directly.

This improves only handling of (possibly indirect) member templates of other templates. Non-templated members of templates aren't still scanned correctly.


kimgr:
Beautiful, thanks!

PR #1986, opened 2026-04-08
Don't scan instantiated specialization arguments

bolshakov-a:
They should be handled only where specified, otherwise it may cause unwanted reports about entities used to specify them. For example, if a variable
```cpp
extern IntArgTpl<GetInt()> iat;
```
is declared in a header, then `GetInt` function should be reported for that header and not for any other source file that fully uses the variable and hence instantiates the template.

`TraverseTemplateSpecializationType` function causes template specialization argument traversing, whereas
`TraverseTemplateSpecializationTypeHelper` does not.

The call of `ReportExplicitInstantiations` has been inserted because this function is called from `VisitTemplateSpecializationType` which in turn is called from `TraverseTemplateSpecializationType` but not from `TraverseTemplateSpecializationTypeHelper`.

Besides, `set_in_forward_declare_context` is called from `VisitTemplateSpecializationType` and hence should now be called explicitly for alias templates. An instantiated alias template can always be scanned as if its top-level type is in a fwd-decl context because even if it is not it is anyway reported from `VisitTemplateSpecializationType` (perhaps in `IwyuAstConsumer`), and doing it from `TraverseInstantiatedAliasTemplateHelper` was just a redundant double-reporting.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3815,11 +3815,11 @@ class InstantiatedTemplateVisitor
     }
 
     if (const auto* tpl_spec_type = type->getAs<TemplateSpecializationType>()) {
-      TraverseTemplateSpecializationType(
-          const_cast<TemplateSpecializationType*>(tpl_spec_type), false);
+      TraverseTemplateSpecializationTypeHelper(tpl_spec_type);
     } else if (const auto* record_type = type->getAs<RecordType>()) {
       TraverseRecordTypeHelper(record_type);
     }
+    ReportExplicitInstantiations(type);
   }
 
   void ScanInstantiatedClass(ClassTemplateSpecializationDecl* decl,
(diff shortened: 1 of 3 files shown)


bolshakov-a:
> and doing it from `TraverseInstantiatedAliasTemplateHelper` was just a redundant double-reporting

In fact, it was even triple-reporting: one from `InstantiatedTemplateVisitor::VisitTemplateSpecializationType`, one from `InstantiatedTemplateVisitor::TraverseInstantiatedAliasTemplateHelper`, and one more from `IwyuAstConsumer::VisitTemplateSpecializationType`.

iwyu.cc
```diff
@@ -3815,11 +3815,11 @@ class InstantiatedTemplateVisitor
     }
 
     if (const auto* tpl_spec_type = type->getAs<TemplateSpecializationType>()) {
-      TraverseTemplateSpecializationType(
-          const_cast<TemplateSpecializationType*>(tpl_spec_type), false);
+      TraverseTemplateSpecializationTypeHelper(tpl_spec_type);
     } else if (const auto* record_type = type->getAs<RecordType>()) {
       TraverseRecordTypeHelper(record_type);
     }
```
bolshakov-a, line 3821: Are curly braces still needed here? Google style guide [requires using them](https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching), LLVM style guide [requires omitting them](https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements).

kimgr: I think we're technically more Google than LLVM style, and I prefer more braces lately, so I say keep them.

kimgr:
Looks good to me!

PR #1987, opened 2026-04-08
Avoid reporting types without name

### LINKED ISSUE #1789: Unwanted suggestions for unnamed types
For the following code:
```c
// indirect.h
extern struct { int i; } var_indirect;

// direct.h
#include "indirect.h"

extern typeof(var_indirect) var_direct;

// source.c
#include "direct.h"

int main() {
    return var_direct.i;
}
```
IWYU outputs:
```
source.c should add these lines:
#include "indirect.h"  // for (anonymous)

source.c should remove these lines:

The full include-list for source.c:
#include "direct.h"    // for var_direct
#include "indirect.h"  // for (anonymous)
```
Suggesting an `#include` for unnamed structure definition seems to be pointless because there is no way to forward-declare it, hence the complete definition should be included along with a `typedef`, or with a variable referenced to get the type, as in the example.

OTOH, the explicit `#include "indirect.h"` may facilitate refactoring when the struct is given a name, and the `#include "indirect.h"` in `direct.h` is removed.

@kimgr, do you agree that this is a bug that should better be fixed?

The issue primarily concerns the C language, more precisely C23 which allows a definition of a previously declared variable with unnamed type like this: `typeof(var_direct) var_direct;`. When the type is `typedef`ed, IWYU should consider the `typedef` as "providing" and not suggest the `#include`.

kimgr:
Add tests for deduced unnamed types using typeof() in C23 and decltype()
in C++.

Fixes #1789.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3606,6 +3606,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (blocked_types.count(GetCanonicalType(type)))
       return;
     if (const NamedDecl* decl = TypeToDeclAsWritten(type)) {
+      if (!decl->getIdentifier()) {
+        VERRS(6) << "Ignoring unnamed type: " << PrintableType(type) << "\n";
+        return;
+      }
       decl = GetDefinitionAsWritten(decl);
       VERRS(6) << "(For type " << PrintableType(type) << "):\n";
       IwyuBaseAstVisitor<Derived>::ReportDeclUse(used_loc, decl);
--- a/tests/c/unnamed-direct.h
+++ b/tests/c/unnamed-direct.h
@@ -1,0 +1,10 @@
+//===--- unnamed-direct.h - iwyu test -------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/c/unnamed-i2.h"
--- a/tests/c/unnamed-i1.h
+++ b/tests/c/unnamed-i1.h
@@ -1,0 +1,12 @@
+//===--- unnamed-i1.h - iwyu test -----------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+extern struct {
+  int i;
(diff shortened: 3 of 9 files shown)


kimgr:
@bolshakov-a I thought I'd try to fix #1789. But I've completely lost my self-confidence for IWYU analysis, so I'm humbly assuming this is not what you had in mind?

If I give the respective structs in `unnamed-i1.h` a name, they are reported in `unnamed.cc`, which seems to be a bug in itself -- shouldn't `decltype` and `typeof` ideally provide the underlying type?

bolshakov-a:
> so I'm humbly assuming this is not what you had in mind?

Why do you think so? This LGTM!

> If I give the respective structs in unnamed-i1.h a name, they are reported in unnamed.cc, which seems to be a bug in itself -- shouldn't decltype and typeof ideally provide the underlying type?

No, they should not. Their referred-to variables may be `extern`, so their types can be forward-declared. The variables declared with `decltype` or `typeof` as the type specifier may itself be `extern` (or be made such in future), so `decltype`s may be replaced by names of forward-declared types. And after all, I'm not very happy with IWYU provision policies and would prefer stricter ones (requiring full types at call sites, not at declaration sites), at least for user's declarations. So, as long as users don't complain, I'd prefer not to extend the provision policies to `decltype`s.

Unluckily, I don't remember why I claimed that this issue affects the C language more than C++...

bolshakov-a:
> Unluckily, I don't remember why I claimed that this issue affects the C language more than C++...

Obviously, it should be because declaring variables of unnamed types as `extern` and then defining them like `decltype(var) var;` is useless in C++ (you cannot link such an "extern" variable with similarly declared in other translation units because their classes have no name for linkage). But in C it works.

In fact, in C++, even `inline` variables of unnamed types in headers don't work as expected (they are still _different_ variables for every TU).

kimgr:
> > Unluckily, I don't remember why I claimed that this issue affects the C language more than C++...
>
> Obviously, it should be because declaring variables of unnamed types as extern and then defining them like decltype(var)
> var; is useless in C++ (you cannot link such an "extern" variable with similarly declared in other translation units because 
> their classes have no name for linkage). But in C it works.
>
> In fact, in C++, even inline variables of unnamed types in headers don't work as expected (they are still different variables 
> for every TU).

So are you saying that the C++ testcase is not representative of real code? Any idea how to make it better?

bolshakov-a:
No, I'm not saying this. I just said that there are lesser chances to find such constructs in real-world code. Perhaps, it would be better to mark the variables in the C++ test as `static`, not `extern` or `inline`.

kimgr:
Updated the tests and commit message.

PR #1988, opened 2026-04-09
Search for autocast constructors in definitions

bolshakov-a:
Prior to this, a forward-declaration of a template appeared prior to the definition spoiled IWYU analysis of "autocast" constructors because the canonical template declaration is the first (empty in such cases) declaration. Thus, the "autocast" constructors were not found.

`GetTagDefinition` does the right thing even with uninstantiated template specializations, so the self-written code has been replaced with it.

Moreover, IWYU finds now "autocast" constructors inside partial specializations, including the one of `std::function`.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -90,7 +90,6 @@
 
 #include <cstdio>
 #include <cstdlib>                      // for atoi, exit
-#include <functional>
 #include <map>                          // for map, swap, etc
 #include <memory>                       // for unique_ptr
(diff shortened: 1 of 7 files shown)


kimgr:
> Moreover, IWYU finds now "autocast" constructors inside partial specializations, including the one of std::function.

Ooh, this is nice. Could I convince you to write up a reduced testcase for it?

bolshakov-a:
> Could I convince you to write up a reduced testcase for it?

Makes sense. Done.

Issue #1064, opened 2022-06-18
Prefer C <std*.h> headers over <cstd*> ones


alejandro-colomar:
<cstd*> headers provide C-compatible stuff in C++, but in the `std::` namespace, which makes it incompatible with C.  So the only thing it does is bring libc functionality to C++ under a different name.

Instead, the good old C headers provide real C-compatible stuff, with the same exact semantics as in C.  They also provide POSIX stuff.

C headers <std*.h> have been _un_deprecated: <https://open-std.org/JTC1/SC22/WG21/docs/papers/2018/p0619r4.html#3.5>, so now they are at the same level than C++ headers from a standards point of view, and they are more useful, so IMO, they should be preferred.

Most coding styles recommend against this.  But they are outdated, since most of them recommend <cstd*> headers because of ISO C++ having deprecated them.  Let's ignore them.

On top of that, iwyu(1) has problems identifying C and C++ code, which triggers the recommendation of some C++ headers in C code.  Having preference of the C headers even in C++ code would fix that issue as a side effect. :)

alejandro-colomar:
Status of C headers in C++:

<https://open-std.org/JTC1/SC22/WG21/docs/papers/2018/p0619r4.html#3.5>
<https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2340r1.html>

danielh2942:
surely you could fix this by not having the stl.c.headers.imp file included and/or altering the imp file you have for the cpp standard library

bolshakov-a:
@alejandro-colomar, your words make sense, but you should at first convince WG21 that `<name.h>` headers should be preferred over `<cname>` (or even that the latter should be deprecated). As long as the C++ standard claims the opposite, I don't think IWYU should contradict it.

alejandro-colomar:
@AlisdairM, @StephanTLavavej, @kaminski-tomasz

What is the status of p0619 in WG21?  Is it still being worked on?  Was it voted?

<https://wg21.link/p0619>

StephanTLavavej:
[P0619R4](https://wg21.link/P0619R4) was formally accepted for C++20, and was [implemented in MSVC's STL](https://github.com/microsoft/STL/issues/28) back in 2020.

alejandro-colomar:
[...]

PR #1990, opened 2026-04-12
Always suggest canonical headers

bolshakov-a:
This implements the logic proposed in https://github.com/include-what-you-use/include-what-you-use/pull/1874#discussion_r2653633124: every header included in a file should be canonical for at least one symbol used in the file.

### DIFF
--- a/tests/cxx/std_size_t.cc
+++ b/tests/cxx/std_size_t.cc
@@ -20,7 +20,9 @@
 #include <cstdio>
 #include <stdio.h>
 
+// IWYU: std::size_t is...*cstddef
 std::size_t f() {
+  // IWYU: size_t is...*cstddef
   const size_t x = 100;
   printf("%zu\n", x);
   return sizeof(int);
@@ -29,11 +31,13 @@ std::size_t f() {
 /**** IWYU_SUMMARY
 
 tests/cxx/std_size_t.cc should add these lines:
+#include <cstddef>
 
 tests/cxx/std_size_t.cc should remove these lines:
-- #include <stdio.h>  // lines XX-XX
+- #include <cstdio>  // lines XX-XX
 
 The full include-list for tests/cxx/std_size_t.cc:
-#include <cstdio>  // for printf, size_t
+#include <stdio.h>  // for printf
+#include <cstddef>  // for size_t
 
 ***** IWYU_SUMMARY */
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -2181,6 +2181,12 @@ vector<string> IncludePicker::GetMappedPublicHeaders(
   return GetCandidateHeadersForFilepathIncludedFrom(decl_filepath, use_path);
 }
 
+vector<string> IncludePicker::GetMappedPublicHeaders(
+    const string& quoted_header, const string& use_path) const {
+  return BestQuotedIncludesForIncluder(
+      GetPublicValues(filepath_include_map_, quoted_header), use_path);
+}
+
 // Parses a YAML/JSON file containing mapping directives of various types:
 //  symbol   - symbol name -> quoted include
 //  include  - private quoted include -> public quoted include
--- a/iwyu_include_picker.h
+++ b/iwyu_include_picker.h
@@ -205,6 +205,9 @@ class IncludePicker {
                                         const string& use_path,
                                         const string& decl_filepath) const;
 
+  vector<string> GetMappedPublicHeaders(const string& quoted_header,
+                                        const string& use_path) const;
+
  private:
   // Private implementation of mapping file parser, which takes
   // mapping file search path to allow recursion that builds up
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -915,7 +915,6 @@ set<string> CalculateMinimalIncludes(
   // - Includes in direct_includes that are also already in
   //   desired_headers.
   // - Includes in desired_headers.
-  // - Includes in direct_includes.
   // Picking in this order minimizes the number of #includes we add,
   // while allowing us to remove #includes if need be.
   for (OneUse& use : *uses) {
@@ -957,15 +956,6 @@ set<string> CalculateMinimalIncludes(
         use.set_suggested_header(choice);
         desired_headers.insert(use.suggested_header());
         LogIncludeMapping("#include already needed", use);
-      }
-    }
-    for (const string& choice : public_headers) {
-      if (use.has_suggested_header())
-        break;
-      if (ContainsKey(direct_includes, choice)) {
-        use.set_suggested_header(choice);
-        desired_headers.insert(use.suggested_header());
-        LogIncludeMapping("#include already present", use);
       }
     }
   }
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -963,27 +963,27 @@ set<string> CalculateMinimalIncludes(
   // Step (3): Now we have a set-cover problem: we need to end up with
   // a set of headers, called cover, so that for every i:
   //    intersection(cover, public_headers[i]) != empty_set
-  // We do this greedily: we find the header that's listed the most
-  // often.  Among those, we prefer the one that's listed first in
+  // We do this greedily: we find the header that's listed first in
   // public_headers[i] the most often (each list is in approximate
-  // best-fit order).  Among those, we choose arbitrarily.  We repeat
-  // until we cover all sets.
+  // best-fit order). Among those, we prefer the one that's listed the most
+  // often totally. Among those, we choose arbitrarily.  We repeat until we
+  // cover all sets.
   set<OneUse*> unmapped_uses;
   for (OneUse& use : *uses) {
     if (use.NeedsSuggestedHeader())
       unmapped_uses.insert(&use);
   }
   while (!unmapped_uses.empty()) {
-    map<string, pair<int,int>> header_counts;   // total appearances, 1st's
+    map<string, pair<int, int>> header_counts;  // 1st's, total appearances
     for (OneUse* use : unmapped_uses) {
       CHECK_(!use->has_suggested_header());
       const vector<string>& public_headers = use->public_headers();
       for (const string& choice : public_headers) {
         if (use->has_suggested_header())
           break;
-        ++header_counts[choice].first;  // increment total count
+        ++header_counts[choice].second;  // increment total count
         if (choice == use->public_headers()[0])
-          ++header_counts[choice].second;  // increment first-in-list count
+          ++header_counts[choice].first;  // increment first-in-list count
       }
     }
     pair<string, pair<int, int>> best = *header_counts.begin();
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -375,6 +375,7 @@ void OneUse::SetPublicHeaders() {
   if (use_kind_ == UseKind::FwdDecl) {
     public_headers_ = GlobalIncludePicker().GetCandidateHeadersForSymbolFwdDecl(
         symbol_name_, GetFilePath(use_loc_));
+    SetCanonicalHeaders();
     return;
   }
 
@@ -387,6 +388,15 @@ void OneUse::SetPublicHeaders() {
   }
   if (public_headers_.empty())
     public_headers_.push_back(ConvertToQuotedInclude(decl_filepath()));
+  SetCanonicalHeaders();
+}
+
+void OneUse::SetCanonicalHeaders() {
+  if (public_headers_.empty())
+    return;
+  canonical_headers_ = GlobalIncludePicker().GetMappedPublicHeaders(
+      public_headers_[0], GetFilePath(use_loc_));
+  canonical_headers_.insert(canonical_headers_.begin(), public_headers_[0]);
 }
 
 const vector<string>& OneUse::public_headers() {
@@ -396,6 +406,10 @@ const vector<string>& OneUse::public_headers() {
         << "Full uses should always have at least one hdr";
   }
   return public_headers_;
+}
+
+const vector<string>& OneUse::canonical_headers() {
+  return canonical_headers_;
 }
 
 bool OneUse::PublicHeadersContain(const string& elt) {
@@ -947,8 +961,7 @@ set<string> CalculateMinimalIncludes(
   //   that are not going away, since they should be calculated already).
   // - Includes in direct_includes that are also already in
   //   desired_headers.
-  // - Includes in desired_headers.
-  // - Includes in direct_includes.
+  // - Canonical headers in direct_includes.
   // Picking in this order minimizes the number of #includes we add,
   // while allowing us to remove #includes if need be.
   for (OneUse& use : *uses) {
@@ -983,6 +996,22 @@ set<string> CalculateMinimalIncludes(
         LogIncludeMapping("#include already present and needed", use);
       }
     }
+    for (const string& choice : use.canonical_headers()) {
+      if (use.has_suggested_header())
+        break;
+      if (ContainsKey(direct_includes, choice)) {
+        use.set_suggested_header(choice);
+        desired_headers.insert(use.suggested_header());
+        LogIncludeMapping("#include already present", use);
+      }
+    }
+  }
+
+  // Clean out unmapped uses that are satisfied by desired_headers.
+  for (OneUse& use : *uses) {
+    if (!use.NeedsSuggestedHeader())
+      continue;
+    const vector<string>& public_headers = use.public_headers();
     for (const string& choice : public_headers) {
       if (use.has_suggested_header())
         break;
@@ -990,15 +1019,6 @@ set<string> CalculateMinimalIncludes(
         use.set_suggested_header(choice);
         desired_headers.insert(use.suggested_header());
         LogIncludeMapping("#include already needed", use);
-      }
-    }
-    for (const string& choice : public_headers) {
-      if (use.has_suggested_header())
-        break;
-      if (ContainsKey(direct_includes, choice)) {
-        use.set_suggested_header(choice);
-        desired_headers.insert(use.suggested_header());
-        LogIncludeMapping("#include already present", use);
       }
     }
   }
@@ -1027,6 +1047,16 @@ set<string> CalculateMinimalIncludes(
         ++header_counts[choice].first;  // increment total count
         if (choice == use->public_headers()[0])
           ++header_counts[choice].second;  // increment first-in-list count
+      }
+    }
+    // IWYU should not suggest headers which aren't canonical for at least one
+    // symbol used in a file. Only the first mapped header should be considered,
+    // headers having include-mapping to it are not desirable. Clean up the map.
+    for (auto it = header_counts.begin(); it != header_counts.end();) {
+      if (it->second.second == 0) {
+        it = header_counts.erase(it);
+      } else {
+        ++it;
       }
     }
     pair<string, pair<int, int>> best = *header_counts.begin();
--- a/iwyu_output.h
+++ b/iwyu_output.h
@@ -124,12 +124,14 @@ class OneUse {
 
   string PrintableUseLoc() const;
   const vector<string>& public_headers();  // not const because we fill lazily
+  const vector<string>& canonical_headers();
   bool PublicHeadersContain(const string& elt);
   bool NeedsSuggestedHeader();
   int UseLinenum() const;
 
  private:
   void SetPublicHeaders();         // sets based on decl_filepath_
+  void SetCanonicalHeaders();
 
   string symbol_name_;             // the symbol being used
   string short_symbol_name_;       // 'short' form of the symbol being used
@@ -142,6 +144,7 @@ class OneUse {
   UseFlags use_flags_;             // flags describing features of the use
   string comment_;                 // If not empty, append to clang warning msg
   vector<string> public_headers_;  // header to #include if dfn hdr is private
+  vector<string> canonical_headers_;  // preferred public headers
   string suggested_header_;        // header that allows us to satisfy use
   bool ignore_use_;                // set to true if use is discarded
   bool is_iwyu_violation_;         // set to false when we figure out it's not
--- a/tests/cxx/badinc-inl.h
+++ b/tests/cxx/badinc-inl.h
@@ -16,7 +16,9 @@
 
 typedef int (*InlH_FunctionPtr)(int);
 
+// IWYU: NULL is...*stddef.h
 InlH_FunctionPtr InlH_Function(class InlH_Class* c=NULL) {
+  // IWYU: NULL is...*stddef.h
   return NULL;
 }
 
@@ -27,6 +29,15 @@ HPrivate_Enum InlH_PrivateFunction() { return HP1; }
 
 /**** IWYU_SUMMARY
 
-(tests/cxx/badinc-inl.h has correct #includes/fwd-decls)
+tests/cxx/badinc-inl.h should add these lines:
+#include <stddef.h>
+
+tests/cxx/badinc-inl.h should remove these lines:
+- #include <locale.h>  // lines XX-XX
+
+The full include-list for tests/cxx/badinc-inl.h:
+#include <stddef.h>  // for NULL
+#include "tests/cxx/badinc-private.h"  // for HPrivate_Enum
+#include "tests/cxx/badinc-private2.h"
 
 ***** IWYU_SUMMARY */
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -172,6 +172,7 @@ Cc_OnlySpecializedStruct<int>* cc_onlyspecializedstruct_ptr;
 
 // Try a type that only needs a forward-declare because of the language linkage.
 extern "C" struct Cc_C_Struct;
+// IWYU: NULL is...*stddef.h
 struct Cc_C_Struct* cc_c_struct_ptr = NULL;
 
 // The types.
@@ -438,6 +439,7 @@ int H_Class::f(I2_Enum i2_enum) { return 1; }
 /*static*/ int H_Class::static_out_of_line(I2_Enum i2_enum) {
   // IWYU: I1_Class needs a declaration
   I1_Class* i1_class;
+  // IWYU: NULL is...*stddef.h
   i1_class = NULL;
   return 1;
 }
@@ -468,6 +470,7 @@ template<typename FOO>
 FOO H_TemplateClass<FOO>::static_out_of_line(I2_Enum i2_enum) {
   // IWYU: I1_Class needs a declaration
   I1_Class* i1_class;
+  // IWYU: NULL is...*stddef.h
   i1_class = NULL;
   return FOO();
 }
@@ -476,6 +479,7 @@ template<typename FOO>
 int H_TemplateClass<FOO>::H_TplNestedStruct::tplnested(I2_Enum i2_enum) {
   // IWYU: I1_Class needs a declaration
   I1_Class* i1_class;
+  // IWYU: NULL is...*stddef.h
   i1_class = NULL;
   return 1;
 }
@@ -484,6 +488,7 @@ template<typename FOO>
 FOO H_TemplateClass<FOO>::H_TplNestedStruct::static_tplnested(I2_Enum i2_enum) {
   // IWYU: I1_Class needs a declaration
   I1_Class* i1_class;
+  // IWYU: NULL is...*stddef.h
   i1_class = NULL;
   return FOO();
 }
@@ -693,6 +698,7 @@ I1_TemplateClass<I1_Enum> i1_templateclass(I11);
 I1_TemplateClass<int> i1_templateclass2(10);
 // IWYU: I1_Class needs a declaration
 // IWYU: I1_TemplateClass is...*badinc-i1.h
+// IWYU: NULL is...*stddef.h
 I1_TemplateClass<I1_Class*> i1_templateclass3(NULL);
 // We need full type info for i1_templateclass because we never
 // fwd-declare a class with default template parameters.
@@ -718,6 +724,7 @@ I1_TemplateClass<std::vector<I1_Class> >* i1_nested_templateclass_ptr;
 // IWYU: I1_TemplateClass is...*badinc-i1.h
 // IWYU: I1_Class needs a declaration
 // IWYU: std::vector is...*<vector>
+// IWYU: NULL is...*stddef.h
 I1_TemplateClass<std::vector<I1_Class>*> i1_nested_templateclass_ptr2(NULL);
 // IWYU: I1_TemplateSubclass is...*badinc-i1.h
 // IWYU: I1_Class needs a declaration
@@ -742,11 +749,13 @@ I1_TemplateClass<std::vector<I1_Enum> >* i1_nested_templateclass_enum_ptr;
 // IWYU: I1_TemplateClass is...*badinc-i1.h
 // IWYU: I1_Enum is...*badinc-i1.h
 // IWYU: std::vector is...*<vector>
+// IWYU: NULL is...*stddef.h
 I1_TemplateClass<std::vector<I1_Enum>*> i1_nested_templateclass_enum_ptr2(NULL);
 // Note: we have to avoid the most vexing parse, here!
 // IWYU: I1_TemplateClass is...*badinc-i1.h
 I1_TemplateClass<D1_I1_Typedef> i1_tplclass_with_typedef((D1_I1_Typedef()));
 // IWYU: I1_TemplateClass is...*badinc-i1.h
+// IWYU: NULL is...*stddef.h
 I1_TemplateClass<D1_I1_Typedef*> i1_templateclass_with_typedef_ptr(NULL);
 // IWYU: I1_TemplateClass is...*badinc-i1.h
 I1_TemplateClass<D1_Enum> i1_templateclass_d(D11);
@@ -876,6 +885,7 @@ float SimpleFunction(I3_ForwardDeclareClass* a, I3_ForwardDeclareStruct* b)
     throw(I3_ForwardDeclareClass*, I1_Class*) {
   return 1.0;
 }
+// IWYU: NULL is...*stddef.h
 float simple_function = SimpleFunction(NULL, NULL);
 
 // IWYU: I3_ForwardDeclareStruct needs a declaration
@@ -1169,6 +1179,7 @@ int main() {
   i1_template_method_only_class.e(local_i2_class_ptr);
   // IWYU: std::vector is...*<vector>
   // IWYU: I2_Class needs a declaration
+  // IWYU: NULL is...*stddef.h
   std::vector<I2_Class>* i2_class_vector = NULL;
   // IWYU: I2_Class is...*badinc-i2.h
   // IWYU: I1_TemplateMethodOnlyClass is...*badinc-i1.h
@@ -1429,6 +1440,7 @@ int main() {
   // IWYU: I1_Class needs a declaration
   // IWYU: I1_Class is...*badinc-i1.h
   // IWYU: I1_const_ptr is...*badinc-i1.h
+  // IWYU: NULL is...*stddef.h
   I1_const_ptr<I1_Class> local_i1_const_ptr(NULL);
   // Needs I1_const_ptr because it calls its operator*().
   // IWYU: I1_const_ptr is...*badinc-i1.h
@@ -1453,6 +1465,7 @@ int main() {
   // IWYU: I1_Class needs a declaration
   // IWYU: I1_Class is...*badinc-i1.h
   // IWYU: I1_const_ptr is...*badinc-i1.h
+  // IWYU: NULL is...*stddef.h
   I1_const_ptr<I1_Class> local_i1_const_ptr2(NULL);
   // IWYU: I1_const_ptr is...*badinc-i1.h
   local_i1_const_ptr2 = local_i1_const_ptr;
@@ -1461,9 +1474,11 @@ int main() {
   // IWYU: I1_Class needs a declaration
   // IWYU: I1_Class is...*badinc-i1.h
   // IWYU: I1_const_ptr is...*badinc-i1.h
+  // IWYU: NULL is...*stddef.h
   (void)I1_const_ptr<I1_Class>(NULL);
   // IWYU: I1_Class needs a declaration
   // IWYU: I1_const_ptr is...*badinc-i1.h
+  // IWYU: NULL is...*stddef.h
   I1_const_ptr<I1_Class*> local_i1_const_ptrptr(NULL);
   // TODO(csilvers): IWYU: I1_Class needs a declaration
   // IWYU: I1_const_ptr is...*badinc-i1.h
@@ -1596,6 +1611,7 @@ int main() {
   // IWYU: I1_TemplateClass is...*badinc-i1.h
   // IWYU: I1_Enum is...*badinc-i1.h
   // IWYU: I2_Class needs a declaration
+  // IWYU: NULL is...*stddef.h
   I1_TemplateClass<I1_Enum, I2_Class>* local_i1_template_class_for_inl = NULL;
   // IWYU: I2_Class::~I2_Class() is...*badinc-i2-inl.h
   // IWYU: I1_TemplateClass is...*badinc-i1.h
@@ -1700,9 +1716,11 @@ int main() {
 
   // Also test while and if initializers.
   // IWYU: I1_Class needs a declaration
+  // IWYU: NULL is...*stddef.h
   while (I1_Class* i = NULL) {
   }
   // IWYU: I1_Class needs a declaration
+  // IWYU: NULL is...*stddef.h
   if (I1_Class* i = NULL) i = NULL;
 
   // Test some macros, including one we shouldn't mark as in-use.
@@ -1809,7 +1827,6 @@ int main() {
 
 tests/cxx/badinc.cc should add these lines:
 #include <ctype.h>
-#include <stddef.h>
 #include <list>
 #include "tests/cxx/badinc-i1.h"
 class D2_Class;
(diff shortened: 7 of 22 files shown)


bolshakov-a:
@alejandro-colomar, please take a look.

At first, I thought that the desired behavior can be achieved just by removing the code which pinned any public header if it is already included directly (1st commit). The change in `badinc-inl.h` shows the desired effect. But then I've noticed that public headers are ranked by the total use count and only then by the canonical use count, so if you have such a mapping file:
```yaml
[
  { "symbol": ["A", "private", "\"a.h\"", "public"] },
  { "symbol": ["A", "private", "\"provides_both.h\"", "public"] },
  { "symbol": ["B", "private", "\"b.h\"", "public"] },
  { "symbol": ["B", "private", "\"provides_both.h\"", "public"] },
]
```
and use both `A` and `B` in your source, IWYU would suggest `provides_both.h` despite it is canonical neither for `A` nor for `B`. So I've added the second commit. Note the change in `std_size_t` test (the comments in the test are rather obsolete). There are three symbols used: `std::size_t`, `size_t`, and `printf`. `<cstdio>` header provides all of them but is not canonical for any of them, so IWYU now suggests replacing it with `<stdio.h>` and `<cstddef>`. Is it the desired behavior?

bolshakov-a:
I'll probably investigate the dogfood results tomorrow.

bolshakov-a:
Well, the changes don't look good. As Aaron noted, C++-style headers of the C standard library should be preferred in C++ code, but there are cases when the C-style header names should be kept. Consider the following mixed C/C++ program:
```cpp
// get_size.h
#include <stddef.h>
#ifdef __cplusplus
extern "C"
{
#endif
size_t GetSize();
#ifdef __cplusplus
}
#endif

// get_size.cpp
#include "get_size.h"

#include <cstdio>
#include <vector>

using std::printf;

size_t GetSize() {
    printf("In GetSize.\n");
    std::vector<int> v{1,2,3};
    return v.size();
}

// main.c
#include "get_size.h"

#include <stdio.h>

int main(void) {
    printf("Size = %d.\n", GetSize());
}
```
Here, `get_size.cpp` includes C++-style header `<cstdio>`, but the C-style `#include <stddef.h>` should be kept in its _associated_ header `get_size.h`.

Perhaps, there may be more than one canonical header for a symbol. Perhaps, IWYU lacks some other concept. Perhaps, the behavior should be tunable by means of command-line flags to satisfy different programming needs...

alejandro-colomar:
> @alejandro-colomar, please take a look.
> 
> At first, I thought that the desired behavior can be achieved just by removing the code which pinned any public header if it is already included directly (1st commit). The change in `badinc-inl.h` shows the desired effect. But then I've noticed that public headers are ranked by the total use count and only then by the canonical use count, so if you have such a mapping file:
> 
> ```yaml
> [
>   { "symbol": ["A", "private", "\"a.h\"", "public"] },
>   { "symbol": ["A", "private", "\"provides_both.h\"", "public"] },
>   { "symbol": ["B", "private", "\"b.h\"", "public"] },
>   { "symbol": ["B", "private", "\"provides_both.h\"", "public"] },
> ]
> ```
> 
> and use both `A` and `B` in your source, IWYU would suggest `provides_both.h` despite it is canonical neither for `A` nor for `B`. So I've added the second commit. Note the change in `std_size_t` test (the comments in the test are rather obsolete). There are three symbols used: `std::size_t`, `size_t`, and `printf`. `<cstdio>` header provides all of them but is not canonical for any of them, so IWYU now suggests replacing it with `<stdio.h>` and `<cstddef>`. Is it the desired behavior?

Yes, that's the desired behavior.

alejandro-colomar:
> Well, the changes don't look good. As Aaron noted, C++-style headers of the C standard library should be preferred in C++ code, but there are cases when the C-style header names should be kept. Consider the following mixed C/C++ program:
> 
> ```c++
> // get_size.h
> #include <stddef.h>
> #ifdef __cplusplus
> extern "C"
> {
> #endif
> size_t GetSize();
> #ifdef __cplusplus
> }
> #endif
> 
> // get_size.cpp
> #include "get_size.h"
> 
> #include <cstdio>
> #include <vector>
> 
> using std::printf;
> 
> size_t GetSize() {
>     printf("In GetSize.\n");
>     std::vector<int> v{1,2,3};
>     return v.size();
> }
> 
> // main.c
> #include "get_size.h"
> 
> #include <stdio.h>
> 
> int main(void) {
>     printf("Size = %d.\n", GetSize());
> }
> ```
> 
> Here, `get_size.cpp` includes C++-style header `<cstdio>`, but the C-style `#include <stddef.h>` should be kept in its _associated_ header `get_size.h`.

What does your patched iwyu suggest in this case?

> 
> Perhaps, there may be more than one canonical header for a symbol. Perhaps, IWYU lacks some other concept. Perhaps, the behavior should be tunable by means of command-line flags to satisfy different programming needs...

bolshakov-a:
> What does your patched iwyu suggest in this case?

It suggests replacing `<cstdio>` with `<stdio.h>` in `get_size.cpp`. I just had an idea that C++-style headers should be canonical in C++ mode. But it that case IWYU would suggest replacing `<stddef.h>` with `<cstddef>` in the header, which would definitely be wrong.

alejandro-colomar:
> > What does your patched iwyu suggest in this case?
> 
> It suggests replacing `<cstdio>` with `<stdio.h>` in `get_size.cpp`.

Hmmm, I see.  That's because we have both `printf()` and `std::printf()`.  I guess that's okay.  It's the user's problem if they want to use both `printf()` and `std:printf()`.  They should just not do that.

> I just had an idea that C++-style headers should be canonical in C++ mode. But it that case IWYU would suggest replacing `<stddef.h>` with `<cstddef>` in the header, which would definitely be wrong.

Yeah, that's wrong.  I think your current patch is good.

bolshakov-a:
>  That's because we have both `printf()` and `std::printf()`. I guess that's okay. It's the user's problem if they want to use both `printf()` and `std:printf()`. They should just not do that.

No, the using-declaration `using std::printf;` is just a way not to repeat `std::` on every `printf` call. (In fact, due to that, IWYU should not remove `#include <cstdio>` at least; I've opened #2009). Moreover, it is a way to make the code portable because
> It is unspecified whether these names ... are first declared within the global namespace scope and are then injected into namespace `std` by explicit using-declarations
>
> https://timsong-cpp.github.io/cppwp/n4950/headers#5

But many programmers don't place such using-decls because the common C++ standard library implementations actually provide the names in the global namespace with `<c...>` headers, as well as injected into `std`. Anyway, there may be not using-declarations but a using-directive:
```cpp
#include <cstdio>
using namespace std;
int main() {
  printf("String");  // Is it ::printf or std::printf?
}
```
In fact, the C++ standard is pretty clear:
> Source files that are not intended to also be valid ISO C should not use any of the C headers.
>
> https://timsong-cpp.github.io/cppwp/n4950/support.c.headers.general

and in the [footnote 157](https://timsong-cpp.github.io/cppwp/n4950/headers#footnote-157):
> ... the newer forms are the preferred forms for all uses except for C++ programs which are intended to be strictly compatible with C.

For IWYU, that means: C++-style headers should be preferred, but C-style headers should be acceptable even in C++ code.

alejandro-colomar:
> > That's because we have both `printf()` and `std::printf()`. I guess that's okay. It's the user's problem if they want to use both `printf()` and `std:printf()`. They should just not do that.
> 
> No, the using-declaration `using std::printf;` is just a way not to repeat `std::` on every `printf` call.

Ahh, sorry, I don't know much C++, as you can see.

> (In fact, due to that, IWYU should not remove `#include <cstdio>` at least; I've opened #2009). Moreover, it is a way to make the code portable because
> 
> > It is unspecified whether these names ... are first declared within the global namespace scope and are then injected into namespace `std` by explicit using-declarations
> > https://timsong-cpp.github.io/cppwp/n4950/headers#5
> 
> But many programmers don't place such using-decls because the common C++ standard library implementations actually provide the names in the global namespace with `<c...>` headers, as well as injected into `std`. Anyway, there may be not using-declarations but a using-directive:
> 
> ```c++
> #include <cstdio>
> using namespace std;
> int main() {
>   printf("String");  // Is it ::printf or std::printf?
> }
> ```
> 
> In fact, the C++ standard is pretty clear:
> 
> > Source files that are not intended to also be valid ISO C should not use any of the C headers.
> > https://timsong-cpp.github.io/cppwp/n4950/support.c.headers.general
> 
> and in the [footnote 157](https://timsong-cpp.github.io/cppwp/n4950/headers#footnote-157):
> 
> > ... the newer forms are the preferred forms for all uses except for C++ programs which are intended to be strictly compatible with C.
> 
> For IWYU, that means: C++-style headers should be preferred, but C-style headers should be acceptable even in C++ code.

As long as it doesn't affect headers shared with C, I guess I don't care too much.

kimgr:
This looks like a good idea in general. I think the special rules were mostly there originally to avoid too many spurious changes to Google code (e.g. force `<stddef.h>` for `size_t` everywhere). But I also think it makes sense for IWYU to be more idealistic than pragmatic at this point.

Some observations and questions from the dogfood results and test changes:

> ```diff
> + #include <getopt.h>                           // for no_argument, required_a...
> - #include "iwyu_getopt.h"  // lines 32-32
> ```

`iwyu_getopt.h` is a portability header, so collapsing to the underlying canonical header will not be desirable. I wonder if this will happen for all portability wrapper headers like this?

> ```diff
> +#include <unistd.h>
> ```

Same here, `<unistd.h>` is not (or at least was not) portable, and `iwyu_getopt.h` is supposed to provide for it.

> ```diff
> + #include "clang/AST/TypeBase.h"                 // for Type, EnumType (ptr o...
> ```

Hmm, I think the intention in Clang was for people to include `Type.h` and treat `TypeBase.h` as a private header. I added https://github.com/llvm/llvm-project/commit/081fe1d517f13f204ff06f50cb68fb6508dba0d7 to support IWYU when the split happened. But maybe that should be an `// IWYU pragma: private, include "clang/AST/Type.h"` inside `TypeBase.h` instead.

> ```diff
>   tests/cxx/badinc.cc should add these lines:
>   #include <ctype.h>
> - #include <stddef.h>
> ```

I don't understand why `<stddef.h>` is removed here, there's several new diagnostics that shows it's needed for `NULL`?

bolshakov-a:
> But I also think it makes sense for IWYU to be more idealistic than pragmatic at this point.

I think that the idealistic way would be to do that it is said in the C++ standard and prefer `<c...>` headers in the C++-code, or at least not to replace them with C headers (as it works on master).

> `iwyu_getopt.h` is a portability header, so collapsing to the underlying canonical header will not be desirable. I wonder if this will happen for all portability wrapper headers like this?

It is the effect of the changes: `iwyu_getopt.h` doesn't provide anything new in Unix environment, so IWYU suggests to replace it with the canonical headers.

> But maybe that should be an `// IWYU pragma: private, include "clang/AST/Type.h"` inside `TypeBase.h` instead.

Yes, probably.

> I don't understand why `<stddef.h>` is removed here, there's several new diagnostics that shows it's needed for `NULL`?

It should now be provided by an associated header of `badinc.cc`, namely `badinc-inl.h`.

kimgr:
> It is the effect of the changes: iwyu_getopt.h doesn't provide anything new in Unix environment, so IWYU suggests to replace it with the canonical headers

Yeah, but I don't think that works very well. I wonder if it would be possible to exempt included with an export pragma from the canonical-only rule? That would also fix the Type/TypeBase regression.

alejandro-colomar:
> > I don't understand why `<stddef.h>` is removed here, there's several new diagnostics that shows it's needed for `NULL`?
> 
> It should now be provided by an associated header of `badinc.cc`, namely `badinc-inl.h`.

I don't like this.  badinc.cc should include its own stuff.  If the associated header defined NULL itself, it would be okay, but I don't want to rely on the associated header including another one that provides NULL.

kimgr:
@alejandro-colomar That's normal IWYU policy -- for a component X, x.h defines the interface and x.c or x.cc defines the implementation. x.c or x.cc typically won't compile unless the associated x.h is already included, so it makes sense to let x.h provide all its includes for this relation.

alejandro-colomar:
> @alejandro-colomar That's normal IWYU policy -- for a component X, x.h defines the interface and x.c or x.cc defines the implementation. x.c or x.cc typically won't compile unless the associated x.h is already included, so it makes sense to let x.h provide all its includes for this relation.

Hmmm, then I guess it's fine.

alejandro-colomar:
> > @alejandro-colomar That's normal IWYU policy -- for a component X, x.h defines the interface and x.c or x.cc defines the implementation. x.c or x.cc typically won't compile unless the associated x.h is already included, so it makes sense to let x.h provide all its includes for this relation.
> 
> Hmmm, then I guess it's fine.

BTW, how does iwyu(1) know that badinc-inl.h is associated with badinc.cc?  I don't see any IWYU pragma, and the name is different.  Is `-inl` implicitly understood by iwyu(1)?

kimgr:
@alejandro-colomar 

> BTW, how does iwyu(1) know that badinc-inl.h is associated with badinc.cc? I don't see any IWYU pragma, and the name is different. Is `-inl` implicitly understood by iwyu(1)?

Yes, see https://github.com/include-what-you-use/include-what-you-use/blob/52f85e1f4d990f55fc6556d543eb051d79364a16/iwyu_preprocessor.cc#L1088 and https://github.com/include-what-you-use/include-what-you-use/blob/52f85e1f4d990f55fc6556d543eb051d79364a16/iwyu_path_util.cc#L92

bolshakov-a:
> Yeah, but I don't think that works very well. I wonder if it would be possible to exempt included with an export pragma from the canonical-only rule?

Probably, it will be not very straightforward, but if done, it could help with C standard library symbols as well, if the logic is as follows: replace used headers with canonical headers, but not when the used header has an include mapping to the desired canonical header.

Btw, I've noticed that for code like this:
```cpp
#include "some_header_including_cstdio"

using namespace std;

int main() {
  printf("Test");
}
```
The "set-cover" algorithm from the master branch produces the desired result:
```
source.cpp should add these lines:
#include <cstdio>  // for printf, std

source.cpp should remove these lines:
- #include "hdr.h"  // lines 1-1

The full include-list for source.cpp:
#include <cstdio>  // for printf, std
```
i.e. IWYU doesn't suggest `#include <stdio.h>` along with, probably, `<algorithm>` for `std`. It is very desirable not to break that logic.

bolshakov-a:
> replace used headers with canonical headers, but not when the used header has an include mapping to the desired canonical header.

I've added a draft commit that should implement this logic, but it still seems inconsistent. Why should IWYU keep some header having a public-to-public mapping to the reported canonical header? It also might provide some extra definitions not needed for the analyzed file, and, probably, should better be replaced by the canonical header by the same logic an implementation of which this PR pursues?

Moreover, I still don't like the changes in `std_size_t` test. I'd prefer the only `#include <cstdio>`, as on the master branch.

bolshakov-a:
Well, it seems that there actually is some asymmetry between symbol and include mappings: libraries often provide "umbrella headers" like e.g. `<boost/geometry.hpp>`, which are encouraged for use in tutorials, and IWYU should accept if a user decides to include such a header instead of a bunch of underlying public headers, although they may provide more than needed for a specific user's source file.

I've also noticed that the headers are ranked by use count and not by different symbol count in the "set cover" algorithm (issue #2048), but it is usually hidden because the headers are sorted by total use count first (hence if e.g. a file uses `A` and `B`, and `b.h` is canonical for `B` and public for `A`, IWYU suggests only `b.h` regardless of the use count of `A` because the total use count of `A` and `B` is greater anyway). But when I swapped the sorting criteria, suggestions became strongly dependent on the use count, so I've reverted that commit and implemented exactly what @alejandro-colomar suggested: excluding the headers that are canonical for none of the used symbols from the candidate header ~~set~~ map (`header_counts`).

I hope this PR is now good enough.

alejandro-colomar:
I haven't tested, by from what you say, LGTM.  Thanks!!  :)

bolshakov-a:
> i.e. IWYU doesn't suggest #include `<stdio.h>` along with, probably, `<algorithm>` for std. It is very desirable not to break that logic.

Of course it is broken:
```console
source.cpp should add these lines:
#include <stdio.h>    // for printf
#include <algorithm>  // for std

source.cpp should remove these lines:
- #include "hdr.h"  // lines 1-1

The full include-list for source.cpp:
#include <stdio.h>    // for printf
#include <algorithm>  // for std
```
A possible (somewhat ugly) fix is:
```patch
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1035,29 +1035,30 @@ set<string> CalculateMinimalIncludes(
       unmapped_uses.insert(&use);
   }
   while (!unmapped_uses.empty()) {
-    map<string, pair<int,int>> header_counts;   // total appearances, 1st's
+    map<string, std::tuple<int, int, int>> header_counts;   // total appearances, 1st's, canonical
     for (OneUse* use : unmapped_uses) {
       CHECK_(!use->has_suggested_header());
       const vector<string>& public_headers = use->public_headers();
       for (const string& choice : public_headers) {
         if (use->has_suggested_header())
           break;
-        ++header_counts[choice].first;  // increment total count
+        ++std::get<0>(header_counts[choice]);  // increment total count
         if (choice == use->public_headers()[0])
-          ++header_counts[choice].second;  // increment first-in-list count
+          ++std::get<1>(header_counts[choice]);  // increment first-in-list count
+        if (llvm::count(use->canonical_headers(), choice))
+          ++std::get<2>(header_counts[choice]);
       }
     }
     // IWYU should not suggest headers which aren't canonical for at least one
-    // symbol used in a file. Only the first mapped header should be considered,
-    // headers having include-mapping to it are not desirable. Clean up the map.
+    // symbol used in a file. Clean up the map.
     for (auto it = header_counts.begin(); it != header_counts.end();) {
-      if (it->second.second == 0) {
+      if (std::get<2>(it->second) == 0) {
         it = header_counts.erase(it);
       } else {
         ++it;
       }
     }
-    pair<string, pair<int, int>> best = *header_counts.begin();
+    pair<string, std::tuple<int, int, int>> best = *header_counts.begin();
```
But this means that e.g. for the following code:
```cpp
// hdr.h
#include <boost/variant/variant.hpp>
#include <boost/variant/static_visitor.hpp>
#include <boost/variant/apply_visitor.hpp>

// src.cpp
#include "hdr.h"

struct Visitor : boost::static_visitor<> {
  void operator()(int) const {}
  void operator()(char) const {}
};

void Fn() {
  boost::variant<int, char> v{};
  boost::apply_visitor(Visitor{}, v);
}
```
with the following mapping file:
```yaml
[
{ "symbol": ["boost::variant", "full", "<boost/variant/variant.hpp>", "public"]},
{ "symbol": ["boost::apply_visitor", "full", "<boost/variant/apply_visitor.hpp>", "public"]},
{ "symbol": ["boost::static_visitor", "full", "<boost/variant/static_visitor.hpp>", "public"]},
{ "include": ["<boost/variant/apply_visitor.hpp>", "public", "<boost/variant.hpp>", "public"] },
{ "include": ["<boost/variant/static_visitor.hpp>", "public", "<boost/variant.hpp>", "public"] },
{ "include": ["<boost/variant/variant.hpp>", "public", "<boost/variant.hpp>", "public"] },
]
```
IWYU suggests to add the umbrella header `<boost/variant.hpp>` into `src.cpp` instead of the three specific public headers. Is it Ok? I suspect that different programmers have different opinions.

bolshakov-a:
As far as Boost.Variant is considered as an example, it would be nice to hear an opinion from @apolukhin.

bolshakov-a:
Even worse, with the above-mentioned patch applied, IWYU would still suggest `<iostream>` instead of `<istream>` and `<ostream>` for [my example](https://github.com/include-what-you-use/include-what-you-use/pull/2028#issuecomment-4575056261). So we have to choose either to place them manually in those cases, or to place `<cstdio>` manually in [other cases](https://github.com/include-what-you-use/include-what-you-use/pull/1990#issuecomment-4276668914). Tuning IWYU to produce the desired behavior in all cases seems to be hard. Note that, once the desired header is included directly, IWYU should honor the choice. CC @firewave.

Of course, there is an option to remove the public-to-public mapping of `<istream>` and `<ostream>` to `<iostream>`, but it would contradict with the C++ standard, and you probably don't want to include `<ostream>` directly when writing `std::cout << "Text";`.

bolshakov-a:
But maybe the concept of multiple canonical headers introduced here is wrong and should be replaced with an explicit umbrella header notion, which would be set in the mappings by users? Umbrella headers would be considered as interchangeable with their underlying headers, whereas the usual public-to-public header mapping would mean that a header provides another header and introduces some more entities, so replacing the underlying headers with it is undesirable until those additional symbols are used as well. `iwyu_getopt.h` should then be considered as an umbrella header for `<unistd.h>` and `<getopt.h>`. (I had an idea that they should be considered as private for IWYU, but it would lead to a visibility conflict.)

Standard C library symbols deserve some special consideration. `<c*>` headers are not, in fact, public headers for their `<*.h>` counterparts (it is just not guaranteed by the C++ standard). For the cases like [this](https://github.com/include-what-you-use/include-what-you-use/pull/1990#issuecomment-4276668914), IWYU should detect `using namespace std;` and remap `printf` use to `std::printf` if the AST node refers to the declaration from `<stdio.h>` actually. Maybe always, maybe only when `<cstdio>` is already included directly. But this looks like a separate piece of work.

bolshakov-a:
Or maybe `<c*>` headers are just to be made canonical in C++ mode, as discussed in [#1064](https://github.com/include-what-you-use/include-what-you-use/issues/1064#issuecomment-4315483024)? Strictly speaking, it would be wrong for global-namespace names (functions and types, in contrast to macros) but in practice, the common implementations do provide them with `<c*>` headers.

bolshakov-a:
> But maybe the concept of multiple canonical headers introduced here is wrong

Maybe not. Consider e.g. `offsetof` macro. `<cstddef>` should be its canonical header in the C++ mode. But if a user wants to include `<stddef.h>` instead, IWYU should accept that, even if no other symbol from `<stddef.h>` is used. But note that `<stddef.h>` cannot be considered as a public header for `<cstddef>` (it doesn't provide things like `std::byte`). Such cases with C standard library symbols can be hard-coded in IWYU, but probably some external libraries require similar handling, and users should be able to do it in mapping files?

firewave:
At the core IWYU should not care about umbrella headers. IMO that is something which needs to be controlled by either mappings or annotations on a library-level.

The umbrella headers of Boost are quite inconsistent and the libraries have different implementations (this is only based on the libraries I have worked with in the past):
- Boost.Regex only contains internal/private headers and you have to use the umbrella one
- BEAST contains all the public headers for the full feature set but they are not all required. In this case you might cut down the headers if you are willing to maintain them yourself.
- others (Boost.Containers? Boost.Math? I do not recall right now) have umbrella headers for certain categories but also lower tier umbrella headers which are only for certain containers/functionality. In this case you should use the lower tier ones.

bolshakov-a:
@firewave, thanks for taking part in the discussion!

> At the core IWYU should not care about umbrella headers. IMO that is something which needs to be controlled by either mappings or annotations on a library-level.

There is exactly the question how they should be exposed in mappings (as usual public headers or as some distinct type). The first version of the present PR made using of umbrella headers impossible because they provide no new symbols, so IWYU would probably always replace `boost/beast/http.hpp` with a set of underlying public headers. I suspect many programmers would not like it and want IWYU to accept the umbrella header at least as an option.

But there are cases when there is a public-to-public mapping, but the mapped-to header is not an umbrella header, and it is undesirable to treat such headers as interchangeable, like `<istream>` and `<iostream>`. So looks like it should be possible to distinguish such cases in a mapping file.

> Boost.Regex only contains internal/private headers and you have to use the umbrella one

In that case, it is not an umbrella header in my terminology, it is just a simple case of a public header for several private headers, no problem here.

bolshakov-a:
I was almost confident that the trouble with `using namespace std;` should be solved by some proper handling of `using namespace std;`, because in [this case](https://github.com/include-what-you-use/include-what-you-use/pull/1990#issuecomment-4276668914) `<cstdio>` is the only acceptable public header (the file is obviously not intended to be valid ISO C), so no best-choice guessing is needed. An even more obvious situation is with using-decls like `using std::printf;`. When there is no `using namespace std;` or using-decls, then unqualified C standard library names should be attributed to C-style headers, which is also the only correct option because it is not guaranteed that C++-style headers provide global-namespace names.

But then I started to think about `ReportTypeUse` when the type as-written may not be handy. I realized that some non-trivial heuristics would be required to figure out what header to report for, say, `struct tm`, because e.g. `std::gmtime` is likely refers to the global-namespace `gmtime` and hence returns `struct tm*` and not `std::tm*`. The situation becomes even more unclear when a header intended for C interoperability provides a global-namespace struct declaration without a definition, and then a C++ source file uses it, like this:

```cpp
// hdr.h
extern struct timespec* ts;

// src.cpp
#include "hdr.h"
#include <ctime>

auto sec = ts->tv_sec;  // Does this compile with <ctime> instead of <time.h>???
```
Just to be sure: @alejandro-colomar, do I understand correctly that `hdr.h` is correct ISO C in that it doesn't require `#include <time.h>`?

This looks like an LWG issue because `src.cpp` is not intended for interoperability with C and therefore should include `<ctime>` and not `<time.h>`, but it is not guaranteed that the provided `std::timespec` is in the global namespace as well. (But they can answer: "Just put `#include <time.h>` into `hdr.h`")

alejandro-colomar:
> ```c++
> // hdr.h
> extern struct timespec* ts;
> 
> // src.cpp
> #include "hdr.h"
> #include <ctime>
> 
> auto sec = ts->tv_sec;  // Does this compile with <ctime> instead of <time.h>???
> ```
> 
> Just to be sure: @alejandro-colomar, do I understand correctly that `hdr.h` is correct ISO C in that it doesn't require `#include <time.h>`?

Yes, that is valid ISO C code:

```sh
alx@devuan:~$ echo 'extern struct timespec* ts;' | gcc -Wpedantic -x c - -S
alx@devuan:~$
```

bolshakov-a:
Thanks! Also asked here: https://discourse.llvm.org/t/are-struct-tm-and-std-tm-the-same-type/91110.

bolshakov-a:
Well, based on [that answer](https://discourse.llvm.org/t/are-struct-tm-and-std-tm-the-same-type/91110/2), for the purposes of `ReportTypeUse` (i.e. when the type to report is not explicitly spelled in the analyzed part of the source code), `<ctime>` and `<time.h>` are interchangeable because they both provide the same types.

It means that in cases like this:
```cpp
// hdr1.h
#include <time.h>

extern struct tm* t;

static inline int GetSec() {
  return t->tm_sec;
}

// cpp_header_not_intended_for_interoperability.h
#include "hdr1.h"
#include <ctime>

inline int GetMin() {
  return t->tm_min;
}
```
`<time.h>` is required in `hdr1.h` (`<ctime>` is not acceptable), but `<ctime>` should be required/acceptable in `cpp_header_not_intended_for_interoperability.h`. I don't know how to achieve such behavior (i.e. replacing `<ctime>` with `<time.h>` in `hdr1.h` if it were there, but not doing this in the second header). Note that IWYU cannot assume that `hdr1.h` provides `<time.h>` because `GetSec` might not be there, so `hdr1.h` might not include any library header, or might include `wchar.h` which provides `tm` declaration but not the definition.

alejandro-colomar:
> Well, based on [that answer](https://discourse.llvm.org/t/are-struct-tm-and-std-tm-the-same-type/91110/2), for the purposes of `ReportTypeUse` (i.e. when the type to report is not explicitly spelled in the analyzed part of the source code), `<ctime>` and `<time.h>` are interchangeable because they both provide the same types.
> 
> It means that in cases like this:
> 
> ```c++
> // hdr1.h
> #include <time.h>
> 
> extern struct tm* t;
> 
> static inline int GetSec() {
>   return t->tm_sec;
> }
> 
> // cpp_header_not_intended_for_interoperability.h
> #include "hdr1.h"
> #include <ctime>
> 
> inline int GetMin() {
>   return t->tm_min;
> }
> ```
> 
> `<time.h>` is required in `hdr1.h` (`<ctime>` is not acceptable), but `<ctime>` should be required/acceptable in `cpp_header_not_intended_for_interoperability.h`. I don't know how to achieve such behavior (i.e. replacing `<ctime>` with `<time.h>` in `hdr1.h` if it were there, but not doing this in the second header). Note that IWYU cannot assume that `hdr1.h` provides `<time.h>` because `GetSec` might not be there, so `hdr1.h` might not include any library header, or might include `wchar.h` which provides `tm` declaration but not the definition.

Maybe the solution is to accept that C headers are the way to go.  It might be controversial to C++ purists, but it's simple, and it works.  :)

bolshakov-a:
But what if a source file looks like this?
```cpp
#include <cstddef>

using namespace std;

size_t s;
byte b;
```
Do you suggest to place `#include <stddef.h>` along with `#include <cstddef>`? But this would be obviously redundant. Note that you cannot replace the latter with the former due to `std::byte`.

alejandro-colomar:
> But what if a source file looks like this?
> 
> ```c++
> #include <cstddef>
> 
> using namespace std;
> 
> size_t s;
> byte b;
> ```
> 
> Do you suggest to place `#include <stddef.h>` along with `#include <cstddef>`?

Hmmm, this is an interesting case.  C++ added C++-only declarations to headers derived from C; I wasn't aware of this.  That seems to have been a bad decision.  It indeed makes this more problematic than it would have been.

> But this would be obviously redundant. Note that you cannot replace the latter with the former due to `std::byte`.

I think it could be reasonable to include both `<stddef.h>` and `<cstddef>`, each for one different type.  This is an odd corner case.

firewave:
I keep wondering if it would make sense to ask about a `libc++` option which doesn't export the `std` symbols to the global namespace...if just for the consistency of having them prefixed in user code.

bolshakov-a:
@firewave, strictly speaking, they don't export `std` symbols to the global namespace, but do the opposite: export glibc stuff to `std` namespace with using-decls. So I don't think they want to redeclare (or even redefine) C library stuff instead. See also a declined proposal: [P0657](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0657r2.pdf).

kimgr:
> > Do you suggest to place #include <stddef.h> along with #include <cstddef>?
>
> Hmmm, this is an interesting case. C++ added C++-only declarations to headers derived
> from C; I wasn't aware of this. That seems to have been a bad decision. It indeed makes
> this more problematic than it would have been.

Ohmygosh, yeah, putting `std::byte` in `<cstddef>` seems like a poor choice to me.

Re the discourse thread, I think jwakely's response:

> The correct way to write that code is to refer to std::timespec, or to include <time.h> or add using
> std::timespec to the global namespace before referring to timespec.

can be read as `include <time.h> _in src.cpp_`? Then there's no extra cost of including `<time.h>` and setting up a bunch of C++ wrapping in `hdr.h`, etc. 

I wonder if we're circling back to the fact that command-line switch would be better for users? Everybody has their own opinion about `<std*.h>` vs. `<cstd*>` and having IWYU force a codebase one way or the other doesn't seem like a good play.

I think with recent IWYU changes, we have a lot more control over which mappings are set up at startup, so I suppose it would be possible to choose different canonical headers based on user policy for the cstd-header subset?

bolshakov-a:
> can be read as `include <time.h> _in src.cpp_`

No, that's not true. You may still have a compilation error if `std::timespec` declaration with the following using-decl (which may appear even inside `<time.h>`) doesn't strictly precede `hdr.h` content: https://godbolt.org/z/16fxMjqrf.

> I wonder if we're circling back to the fact that command-line switch would be better for users?

And I'm thinking about making the changes of this PR an optional experimental feature.

> I suppose it would be possible to choose different canonical headers based on user policy for the cstd-header subset?

But don't forget that there are cases when users may want both `<c*>` and `<*.h>` headers to be acceptable, namely when a `.cpp` file uses some C++ stuff, but its associated header should be usable as a C header in a mixed C/C++ program.

kimgr:
> But don't forget that there are cases when users may want both <c*> and <*.h> headers to be acceptable, namely when
> a .cpp file uses some C++ stuff, but its associated header should be usable as a C header in a mixed C/C++ program.

Hmm, yeah, maybe it would be nice if we could also "disassociate" associated headers to allow IWYU to be used independently on `a.h` vs. `a.cpp`.

kimgr:
> > can be read as include <time.h> _in src.cpp_
>
> No, that's not true. You may still have a compilation error if std::timespec declaration with the following
> using-decl (which may appear even inside <time.h>) doesn't strictly precede hdr.h content:
> https://godbolt.org/z/16fxMjqrf.

Oh wow, the `<c*>` headers really are a minefield.

> And I'm thinking about making the changes of this PR an optional experimental feature.

I wonder if it would make sense to think about ways to handle the `<c*>` headers specifically first, and then think about canonical headers? Or accept that this PR makes the already-imperfect handling of `<c*>` headers a bit worse/different?

Making it an experimental feature/toggled behavior may mean we need to maintain bugfixes across two modes, which could be a bit painful.

bolshakov-a:
> I wonder if it would make sense to think about ways to handle the `<c*>` headers specifically first, and then think about canonical headers?

Yeah, it also seems to me that it is the correct way. What about mapping the C standard library symbols only to C headers in C language mode, and only to `<c*>` headers in the C++ language mode, approximately as proposed in https://github.com/include-what-you-use/include-what-you-use/issues/1064#issuecomment-4305058055? It would be not correct according to the C++ standard, but Ok with the common real-world implementations. IWYU should then be launched by users separately (from their asscociated cpp-files) on headers intended for interoperability as on C headers. The effect should then be in line with the current C++ standard wording.

I see two ways of doing this:
* writing two separate C standard library symbol mappings, or
* making [`stdlib_c_include_map`](https://github.com/include-what-you-use/include-what-you-use/blob/0d18628409cb7b9290d4238d3a165b31d3f3e75b/iwyu_include_picker.cc#L650) a private-to-public mapping. It would cause some visibility conflicts which could be likely resolved by some hack.

kimgr:
I would like to make it possible for users to force use of the basic C headers for C++ too, using a command-line switch. I'm pretty sure there are C++ projects where the plain C headers are still preferred, even if it's still in direct conflict with the standard. Edit: So I'm proposing a mode where IWYU suggests C headers even for source files that clearly contains C++ (`-xc++` or one of the cxx suffixes).

I think that pushes the scales toward two separate C standard library symbol mappings. Not sure how to deal with the `std::byte` debacle, but that might be a third separate set of mappings (C++-only symbols exposed from wannabe C headers).

bolshakov-a:
> I think that pushes the scales toward two separate C standard library symbol mappings.

I just thought that the present PR should be merged prior to writing C library symbol mapping (with multiple-header symbols included), otherwise it would rollback @alejandro-colomar's work: https://github.com/include-what-you-use/include-what-you-use/pull/1874#issuecomment-3697620160. However, multiple-header symbol mappings can be kept as-is for a while.

> Not sure how to deal with the `std::byte` debacle

It is already mapped, along with C library names introduced into `std` namespace:
https://github.com/include-what-you-use/include-what-you-use/blob/0d18628409cb7b9290d4238d3a165b31d3f3e75b/std_symbol_map.inc#L419

alejandro-colomar:
> > I think that pushes the scales toward two separate C standard library symbol mappings.
> 
> I just thought that the present PR should be merged prior to writing C library symbol mapping (with multiple-header symbols included), otherwise it would rollback @alejandro-colomar's work: [#1874 (comment)](https://github.com/include-what-you-use/include-what-you-use/pull/1874#issuecomment-3697620160).

I would find it acceptable to change my include style to be less strict, in that `<string.h>` is an acceptable provider of NULL as long as I use `<string.h>`-only features.  Is that what worried you?

bolshakov-a:
> I would find it acceptable to change my include style to be less strict, in that `<string.h>` is an acceptable provider of NULL as long as I use `<string.h>`-only features.

This PR already implies exactly that behavior.

> I'm pretty sure there are C++ projects where the plain C headers are still preferred, even if it's still in direct conflict with the standard. ... I think that pushes the scales toward two separate C standard library symbol mappings.

@kimgr, this can be achieved just by turning off `stdlib_c_include_map` completely when the corresponding command line flag (we should choose a name for it) is in use.

kimgr:
@bolshakov-a Yes, that's what I was hoping for!

> (we should choose a name for it) 

`-Xiwyu --use_c_headers`, and then we can potentially add an enum value for modes if necessary later on? Stuff like `--use_c_headers={c_language,always,blah}`.

bolshakov-a:
> Yes, that's what I was hoping for!

So, let's try to proceed with `stdlib_c_include_map` manipulations?

bolshakov-a:
> So, let's try to proceed with stdlib_c_include_map manipulations?

Done in #2062.

kimgr:
So now that #2062 is merged, is this ready for consideration?

bolshakov-a:
I think so. Because `<cstdio>` is now the only candidate for `printf` in the C++ mode, [the case with `using namespace std;`](https://github.com/include-what-you-use/include-what-you-use/pull/1990#issuecomment-4276668914) is not affected by this PR anymore. I've resolved the conflicts in `badinc` (there is no changes there now), and fixed `std_size_t` test.

The CI failure is somewhat strange. In my fork, CI passes: https://github.com/bolshakov-a/include-what-you-use/actions/runs/28335028997.

tests/cxx/std_size_t.cc
```diff
@@ -7,15 +7,8 @@
...
-// size_t is such a commonly-used type that it deserves a dedicated test to
-// demonstrate how this behaves in general. Note how the use of 'printf' is
-// attributed to <cstdio> and <stdio.h> is removed. This happens because
-// std::size_t is mapped to <cstdio> before 'printf' is seen and already-needed
-// includes take precedence if a symbol is available from multiple sources.
-//
-// (Note that if 'printf' was seen before std::size_t, <stdio.h> would still be
-// required.)
+// In this test, IWYU picks <cstdio> for printf and size_t although it is
+// not guaranteed that <c...> headers provide names from the global namespace.
```
bolshakov-a, line 11: In fact, the comment change should be attributed to #2062 and not to this PR.

bolshakov-a:
I've just discovered that the latest C++ standard draft makes one more difference between `<cstddef>` and `<stddef.h>` from C23: it is explicitly stated that `<cstddef>` doesn't provide the macro `unreachable`: https://eel.is/c++draft/cstddef.syn#1 (seemingly, because C++23 defined [its own `unreachable`](https://timsong-cpp.github.io/cppwp/n4950/utility.unreachable) in `<utility>` header). I don't know if C++ standard library implementers will explicitly `#undef` the C-library `unreachable`. (In my `glibc` version, it is still absent.)

I hope that it won't cause any trouble with IWYU, because if that macro is `#undef`ed, C++ programmers cannot make use of it and will use `std::unreachable` from `<utility>` instead.

bolshakov-a:
Also note that the present PR doesn't change the behavior with cases when there is an include mapping between standard library headers, like `<stdint.h>` -> `<inttypes.h>` mapping:
https://github.com/include-what-you-use/include-what-you-use/blob/146292c5224f204166bb855ba8bf392685f92691/iwyu_include_picker.cc#L647-L648

For this example:
```c
#include <inttypes.h>
int8_t i;
```
IWYU outputs that `#include`s are correct although it is better to replace `<inttypes.h>` with more specific `<stdint.h>`. This is why I wrote that include mappings probably need a way to distinguish between interchangeable (umbrella) and non-interchangeable headers.

A workaround is to specify symbol mappings for all the symbols from mapped-from header to mapped-to header instead of the include mappings.

kimgr:
@bolshakov-a Could you rebase this PR so we get a clean build?

tests/cxx/std_size_t.cc
kimgr: Then let's merge it as a separate refactor to keep this PR as small as possible.

bolshakov-a: I've opened #2081.

kimgr:
I'm feeling a little worried about this patch, because it feels like it will have a dramatic impact on IWYU behavior. But I'm trying to be more optimistic -- maybe the dramatic impact will be truly positive :slightly_smiling_face:.

I was hoping you could clarify some things in the commit message:

> Canonical headers should be preferred regardless of whether there is
> another public header for a symbol included or not.

Maybe it would be nice to define exactly what "canonical header" means here. In my mind it's just the first mapped header, but I feel like there's more going on now.

> Headers mapped to the first public header of a symbol are now also
> considered as canonical ones. This has been done to support "umbrella
> headers". IWYU should honor users' decisions to include e.g.
> `<boost/geometry.hpp>` instead of specific headers.

If it helps for umbrella headers, that's good, but it sounds like a strange policy in general... So if I have `#include <boost/geometry/some_private.h>` in my source code, and `some_private.h` is mapped to `geometry.hpp`, the private header will be considered canonical? I'm guessing that's not what's happening, so maybe you could clarify this section a bit?

> The check for public headers that are in desired_headers has been moved
> out from the cycle over the uses in `CalculateMinimalIncludes` function

`cycle` -> `loop`?

> so that if some of the canonical headers of a symbol is directly
> included, and it is at the same time a public header of some unmapped
> symbol, IWYU attributes the use of the latter to that header, regardless
> of the order of the uses.

That makes sense.

bolshakov-a:
> So if I have `#include <boost/geometry/some_private.h>` in my source code, and `some_private.h` is mapped to `geometry.hpp`, the private header will be considered canonical? I'm guessing that's not what's happening, so maybe you could clarify this section a bit?

"Headers" replaced with "Public headers" in the commit message.

> `cycle` -> `loop`?

Thanks! A "false friend of a translator": "loop" is "cycle" in Russian in programming context.

kimgr:
Thank you! So what kind of impact should we expect here? IWYU dogfood does not appear to suggest any changes, so I'm guessing... not so much?

I wonder if it would make sense to spell out a concrete example of before/after in the commit message too, in terms of standard library headers, which is where most people will be hit, I guess? The `boost/geometry.hpp` example is great for public headers, so I think the narrowing to canonical-always could use a similar explanatory example.

kimgr:
Thanks for the additional example. Let's do this.

PR #2004, opened 2026-04-15
fix: add parentheses to IWYU_ARRAYSIZE macro expression

### LINKED ISSUE #1993: IWYU_ARRAYSIZE macro missing parentheses around result expression
The `IWYU_ARRAYSIZE` macro in `iwyu_port.h` is defined as:
```c
#define IWYU_ARRAYSIZE(arr) sizeof(arr) / sizeof(*arr)
```

This is missing outer parentheses around the entire expression. If the result is used in a multiplication or addition context (e.g. `2 * IWYU_ARRAYSIZE(arr)`), operator precedence causes incorrect evaluation.

The safe idiomatic form is:
```c
#define IWYU_ARRAYSIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
```

The macro is used in 9 call sites in `iwyu_include_picker.cc`, all as direct function arguments, so no call sites need to change.

pierluigilenoci:
Fixes #1993

The `IWYU_ARRAYSIZE` macro in `iwyu_port.h` was missing outer parentheses around the division expression, making it unsafe when used in multiplication or addition contexts (e.g. `2 * IWYU_ARRAYSIZE(arr)` would evaluate as `(2 * sizeof(arr)) / sizeof(*arr)`).

This PR adds the missing parentheses and uses the more conventional `(arr)[0]` instead of `*arr`.

No call sites need to change since all current usages pass the result as a direct function argument.

### DIFF
--- a/iwyu_port.h
+++ b/iwyu_port.h
@@ -17,7 +17,7 @@
 #include "llvm/Support/raw_ostream.h"
 
 // Count of statically allocated array.
-#define IWYU_ARRAYSIZE(arr) sizeof(arr) / sizeof(*arr)
+#define IWYU_ARRAYSIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
 namespace include_what_you_use {
 


pierluigilenoci:
Hi @kimgr, @bolshakov-a — this is a one-line fix to the `IWYU_ARRAYSIZE` macro in `iwyu_port.h`. Would you be willing to review? Thank you!

bolshakov-a:
LGTM, thanks!

PR #2005, opened 2026-04-15
fix(iwyu_tool): reduce CPU usage in job polling loop

### LINKED ISSUE #1995: iwyu_tool.py: polling loop uses time.sleep(0.0001) causing CPU-wasteful busy-wait
In `iwyu_tool.py`, the main job scheduling loop yields CPU with:
```python
# Yield CPU.
time.sleep(0.0001)
```

This wakes up 10,000 times per second regardless of how long IWYU jobs take. For analysis jobs that run for seconds, this is a CPU-wasteful busy-wait that increases battery drain and CPU usage unnecessarily.

Changing to `time.sleep(0.1)` (100ms, 10 polls/second) is sufficient for jobs that take seconds, with no perceptible latency difference.

pierluigilenoci:
Fixes #1995

The job scheduling loop in `iwyu_tool.py` used `time.sleep(0.0001)` (100us), causing 10,000 wakeups per second while waiting for IWYU analysis jobs that typically run for seconds.

This PR changes the sleep interval to `0.1` seconds (10 polls/sec), which is more than sufficient granularity for long-running jobs and significantly reduces CPU and battery usage during analysis runs.

### DIFF
--- a/iwyu_tool.py
+++ b/iwyu_tool.py
@@ -438,8 +438,8 @@ def execute(invocations, verbose, formatter, jobs, max_load_average=0):
         pending.extend(i.start(verbose) for i in invocations[:capacity])
         invocations = invocations[capacity:]
 
-        # Yield CPU.
-        time.sleep(0.0001)
+        # Yield CPU between job polls.
+        time.sleep(0.1)
     return exit_code
 
 


pierluigilenoci:
Hi @kimgr, @bolshakov-a — this is a small fix to reduce CPU usage in the job polling loop in `iwyu_tool.py`. Would you be willing to review? Thank you!

bolshakov-a:
Looks reasonable, but I have no idea how to find the optimum value.

kimgr:
LGTM. I remember experimenting with that poll delay (it was originally zero), but I see now that we're in no rush since we're just monitoring comparatively long-running child processes. Makes sense.

Issue #2008, opened 2026-04-15
IWYU suggests including non-existent <math> header.


danmccormick-ca:
I'm using a home-brew installed 0.26, paired with a home-brew installed llvm 22 on macOS

```
// foo.cpp
#include <cmath>

int main() {
  float f = -3.14159;
  return std::sqrt(f) != 0.0f;
}

// command line for invocation
/opt/homebrew/bin/include-what-you-use  -std=c++20 -stdlib=libc++ -resource-dir /opt/homebrew/opt/llvm/lib/clang/22 -isystem /opt/homebrew/opt/llvm/include/c++/v1 -nostdinc++ -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/ foo.cpp

// output
foo.cpp should add these lines:
#include <math>   // for sqrt
...
```

So the issue here is that the suggested include of `<math>` is not a header that exists. 

This was added here: https://github.com/include-what-you-use/include-what-you-use/commit/f74eebfcef2ec437bd5c6e10ac4b7147aa5e6ff4#diff-da97d292096b0c4a128af80c00a916d91bb0589f201fe818d5294ce11826f8fbR1223 

In this commit: https://github.com/include-what-you-use/include-what-you-use/commit/f74eebfcef2ec437bd5c6e10ac4b7147aa5e6ff4#diff-da97d292096b0c4a128af80c00a916d91bb0589f201fe818d5294ce11826f8fbR1223 

I suspect this was just a typo, and should be `<cmath>`, but I'm not 100% sure that I'm not missing something.

danmccormick-ca:
Tagging @jashank

kimgr:
@danmccormick-ca Thanks! Yeah, that shell pipeline in the PR:
```
cd llvm-project/libcxx/include ; find . -type d -name "__*" | grep -v cxx03 | LC_ALL=C sort | sed -e 's#./__\(.*\)#    {"@<__\1/.*>", kPrivate, "<\1>", kPublic},#'
```

seems to just map the directory prefix to a header of the same name, e.g. `__atomic/.*` -> `<atomic>`.

The PR mentions a few tweaks, https://github.com/include-what-you-use/include-what-you-use/pull/1876/changes#diff-da97d292096b0c4a128af80c00a916d91bb0589f201fe818d5294ce11826f8fbL1199, so I guess the output was manually edited, but `__math`/`<math>` were overlooked.

[...]

PR #2010, opened 2026-04-19
Fix libcxx mapping for __math/ to cmath

kimgr:
The shell pipeline finds all directories called '__xxx' and blindly generates a mapping from `__xxx/.*` to `<xxx>`.

This works for most of the support dirs containing private headers, but there are some exceptions; the math subsystem is one of them.

Add a little note about tweaking, and update it manually.

Fixes issue #2008.


kimgr:
I also did an exhaustive check and extracted all mapped-to filenames and checked that they exist in libcxx/include, something like:
```
cd ~/code/llvm-project/libstdcxx/include
# ... manually copy the mapping records from iwyu_include_picker.cc to ~/code/scratch/xx...
cat ~/code/scratch/xx |
  # remove comments
  sed -e '/ *\/\/.*/d' |
  # remove blank lines
  sed -e '/^ *$/d' |
  # extract filename
  sed -Ee 's/.*kPrivate, "<([^>]+)>",.*/\1/' |
  # remove duplicates
  sort -u |
  # check that each file exists
  xargs ls -alh
```
No surprises there anymore.

PR #2011, opened 2026-04-19
Add mapping for some [input.output] headers

bolshakov-a:
The symbols from [[input.output]](https://timsong-cpp.github.io/cppwp/n4950/input.output) clause of the standard were excluded from the standard symbol mapping because there is currently no way to map forward-declarations from `<iosfwd>` correctly. However, contents of some [input.output] headers don't overlap with `<iosfwd>` contents, so there is no need to support forward-declaration mapping for them. This change adds the C++23 contents of those headers to the mapping.


kimgr:
Nice increment, thanks!

PR #2012, opened 2026-04-19
Remove `<print>` mapping

bolshakov-a:
Adding it in d1e42bf8bebae7269ae83e999b49f11144afeeda was wrong because the overloads from `<ostream>` are still not mapped, so any use of them would be erroneously attributed to `<print>`.


bolshakov-a:
@kimgr

PR #2013, opened 2026-04-19
Add std::hash primary template mapping

### LINKED ISSUE #2007: libstdc++ makes incorrect suggestion for std::hash
Minimal example:

```cpp
#include <cstddef>
#include <memory>

struct Foo {};
using FooRef = std::shared_ptr<const Foo>;

template<>
struct std::hash<FooRef> {
    std::size_t operator()(const FooRef&) const = delete;
};
```

```sh
include-what-you-use -std=c++20 example.cpp

/tmp/test_hash.cpp should add these lines:
#include <string_view>  // for hash
```

I believe the correct suggestion should be `<functional>`: https://en.cppreference.com/w/cpp/header/functional.html

Tested on Ubuntu 25.10 with IWYU built from source for clang 22: 

```
include-what-you-use --version
include-what-you-use 0.26 (git:01a091d) based on Ubuntu clang version 22.1.1 (++20260305112603+e693f0fe6390-1~exp1~20260305232734.42)
clang --version
Ubuntu clang version 20.1.8 (0ubuntu4)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-20/bin
```

Originally found in a slightly different CI environment, where it incorrectly suggested `#include <variant>`: https://github.com/bitcoin/bitcoin/actions/runs/24406181800/job/71290035323?pr=33922

bolshakov-a:
After d1e42bf8bebae7269ae83e999b49f11144afeeda, it can be done because no one of the standard library headers providing std::hash is excluded from the mapping. It has been also noted that a couple of places where std::hash is declared are not mentioned in the corresponding header synopses. The mapping for them has been added manually.

This fixes #2007.


hebasto:
> This fixes #2007.

I can confirm that this [resolves](https://github.com/bitcoin/bitcoin/pull/35101#discussion_r3109296011) the issue in our project.

Thank you for addressing this so quickly!

bolshakov-a:
You're welcome!

Sjors:
I tested that this patch does the job in https://github.com/bitcoin/bitcoin/pull/35101. Can you backport it to the `clang_22` branch?

bolshakov-a:
Sorry, but we don't make backports. @kimgr, right? (But if actually backported, it should be done along with the two preceding commits.)

kimgr:
Nope, we have enough trouble keeping mainline working.

Sjors:
@bolshakov-a just to double check, you mean these two?

- https://github.com/include-what-you-use/include-what-you-use/commit/28d70e6a776c735fec8fe5ad630657533000591b
- https://github.com/include-what-you-use/include-what-you-use/commit/d1e42bf8bebae7269ae83e999b49f11144afeeda

bolshakov-a:
Correct, if you want to run the mapping generator by yourself. But in fact, all you need to get rid of this bug are `std::hash` mappings.

PR #2014, opened 2026-04-22
Handle user-defined literals in function-call traversal

l0rinc:
### Problem
IWYU could miss that a user-defined literal requires the header that declares its literal operator and could suggest removing a direct include even though that include was still needed for literal lookup.

This showed up in Bitcoin Core during review of [bitcoin/bitcoin#34435](https://github.com/bitcoin/bitcoin/pull/34435#discussion_r3115756052): headers using literals such as `16_MiB` could lose their direct `#include <util/byte_units.h>`.

### Fix
IWYU already has special traversal for ordinary function calls so it can record the callee declaration as a use and apply the usual call-site handling.
User-defined literals were missing that same treatment.
This change adds matching traversal for `UserDefinedLiteral` and routes it through the existing function-call handling path.

### Reproducers
A regression test has been added for a minimal direct/indirect include case:
a source file includes a wrapper header, uses a user-defined literal, and IWYU must replace the wrapper include with the header that directly declares the literal operator.

The test also covers a literal operator whose return type is only forward-declared at the call site, so IWYU must report the return type through the same function-call handling path.

The original Bitcoin Core failure is reproduced in https://github.com/l0rinc/bitcoin/pull/142.
The same Bitcoin Core state passes with this IWYU change applied in https://github.com/l0rinc/bitcoin/pull/141.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -218,6 +218,7 @@ using clang::FriendTemplateDecl;
 using clang::FunctionDecl;
 using clang::FunctionProtoType;
 using clang::FunctionTemplateDecl;
+using clang::UserDefinedLiteral;
 using clang::FunctionType;
 using clang::LateParsedTemplate;
 using clang::LinkageSpecDecl;
@@ -728,6 +729,17 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
     }
     return this->getDerived().HandleFunctionCall(
         func, TypeOfParentIfMethod(expr), expr);
+  }
+
+  bool TraverseUserDefinedLiteral(UserDefinedLiteral* expr) {
+    if (!Base::TraverseUserDefinedLiteral(expr))
+      return false;
+    if (CanIgnoreCurrentASTNode())
+      return true;
+    FunctionDecl* func = expr->getDirectCallee();
+    if (!func)
+      return true;
+    return this->getDerived().HandleFunctionCall(func, nullptr, expr);
   }
 
   bool TraverseCXXMemberCallExpr(CXXMemberCallExpr* expr) {
--- a/tests/cxx/user_defined_literal-direct.h
+++ b/tests/cxx/user_defined_literal-direct.h
@@ -1,0 +1,18 @@
+//===--- user_defined_literal-direct.h - test input file for iwyu ---------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// This file is meant to be directly included by the test .cc file.  It only
+// re-exports the header that declares the literal operator.
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_USER_DEFINED_LITERAL_DIRECT_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_USER_DEFINED_LITERAL_DIRECT_H_
+
+#include "user_defined_literal-indirect.h"
+
+#endif  // INCLUDE_WHAT_YOU_USE_TESTS_CXX_USER_DEFINED_LITERAL_DIRECT_H_
--- a/tests/cxx/user_defined_literal-indirect.h
+++ b/tests/cxx/user_defined_literal-indirect.h
@@ -1,0 +1,19 @@
+//===--- user_defined_literal-indirect.h - test input file for iwyu -------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_USER_DEFINED_LITERAL_INDIRECT_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_USER_DEFINED_LITERAL_INDIRECT_H_
+
+constexpr unsigned long long operator""_x(unsigned long long value) {
+  return value;
+}
+
+class IndirectClass {};
+
+#endif  // INCLUDE_WHAT_YOU_USE_TESTS_CXX_USER_DEFINED_LITERAL_INDIRECT_H_
--- a/tests/cxx/user_defined_literal.cc
+++ b/tests/cxx/user_defined_literal.cc
@@ -1,0 +1,39 @@
+//===--- user_defined_literal.cc - test input file for iwyu ---------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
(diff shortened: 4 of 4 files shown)


tests/cxx/user_defined_literal_include.cc
```diff
@@ -0,0 +1,26 @@
+//===--- user_defined_literal_include.cc - test input file for iwyu -------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -std=c++17 -Xiwyu --check_also="tests/cxx/*-d1.h" -I .
```
bolshakov-a:
```suggestion
// IWYU_ARGS: -Xiwyu --check_also="tests/cxx/*-d1.h" -I .
```
I think that specifying C++17 is pointless because user-defined literals were already in C++11.

iwyu.cc
```diff
@@ -730,6 +731,18 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
...
+  bool TraverseUserDefinedLiteral(UserDefinedLiteral* expr) {
+    if (!Base::TraverseUserDefinedLiteral(expr))
+      return false;
+    if (CanIgnoreCurrentASTNode())
+      return true;
+    FunctionDecl* func = expr->getDirectCallee();
+    if (!func)
+      return true;
+    return this->getDerived().HandleFunctionCall(
+        func, TypeOfParentIfMethod(expr), expr);
```
bolshakov-a:
```suggestion
        func, nullptr, expr);
```
AFAIK, user-defined literals cannot be class methods.

bolshakov-a (approved):
Thanks for working on this! A couple of nits inline, but generally looks good! However, the test case looks somewhat strange and overcomplicated. The common way to write IWYU tests is roughly as follows:
```cpp
// ...-indirect.h
constexpr unsigned long long operator""_x(unsigned long long value) {
  return value;
}

// ...-direct.h
#include "...-indirect.h"

// test.cc
#include "...-direct.h"

// IWYU: operator""_x(unsigned long long) is...*-indirect.h
constexpr unsigned long long kLiteralValue = 1_x;

// IWYU summary: test.cc should add ...-indirect.h and remove ...-direct.h
```

kimgr:
Yeah, what @bolshakov-a said about the testcase. It's nice if you can write the test case not so that IWYU doesn't complain, but so that it _does_ suggest changes, because then you will also get diagnostics emitted at the use locations. See https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#force-diagnostics for more test guidelines.

tests/cxx/user_defined_literal_include-d1.h
```diff
@@ -0,0 +1,28 @@
...
+// Tests that a user-defined literal in a header counts as a direct use of the
+// header that declares the literal operator.
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_USER_DEFINED_LITERAL_INCLUDE_D1_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_USER_DEFINED_LITERAL_INCLUDE_D1_H_
+
+#include "user_defined_literal_include-i1.h"
+#include "user_defined_literal_include-i2.h"
+
+constexpr int kTransitiveUse = kIndirectSize;
```
kimgr: I'm not sure what the `kTransitiveUse` and referred-to `indirect` symbols are used for here. Do they relate at all to the user-defined literals, or could they be removed?

iwyu.cc
```diff
@@ -730,6 +731,18 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
         func, TypeOfParentIfMethod(expr), expr);
   }
 
+  bool TraverseUserDefinedLiteral(UserDefinedLiteral* expr) {
```
kimgr, line 734: Do we need to do this in `Traverse...` as opposed to `Visit...`?

My rule of thumb is that `Traverse...` overrides are only necessary if we want to modify the traversal of the AST, whereas `Visit...` is preferred if we just want to inspect the node. But I made that rule up, so I might be missing something :-)

kimgr: Ah, I see that `HandleFunctionCall` triggers its own custom traversal, so that's why everybody else is doing this too.

tests/cxx/user_defined_literal_include-d1.h
l0rinc: All scaffolding removed in latest push, thanks for the comments

tests/cxx/user_defined_literal_include.cc
l0rinc: Removed with most of the args, thanks

iwyu.cc
l0rinc: Double checked the standard wording with an LLM and `nullptr` is indeed correct here.

https://eel.is/c++draft/over.literal#2 says that a declaration with a literal-operator-id must declare a function or function template that belongs to a namespace, though it may be a friend function. https://eel.is/c++draft/over.literal#7 then says literal operators are ordinary namespace-scope functions.
And https://eel.is/c++draft/class.mem.general#4 says a friend declaration does not itself declare a class member.

So even the friend-function case does not make a literal operator a class method, which means there is no meaningful parent type to pass here.

l0rinc (commented):
Force-pushed and updated the PR description, thanks for the quick reviews.

The testcase is now written in the more usual IWYU style: it forces a diagnostic at the use site and a concrete add/remove suggestion, using a `-direct` / `-indirect` header pair.
I also re-checked that the test fails without the fix and that this still fixes the issue found in Bitcoin Core.

bolshakov-a (approved):
As I said, this LGTM, but it would make sense to add a test case like this:
```cpp
class IndirectClass;  // A forward-decl to indicate that the return type is not provided by the operator.
IndirectClass operator""_ic(unsigned long long);

void Fn() {
  (void)1_ic;  // IWYU should report IndirectClass here.
}
```
This would test that `HandleFnReturnOnCallSite` function is called.

l0rinc:
Thanks @bolshakov-a, pushed something similar, updated PR description, please let me know if that's what you meant:

```bash
git range-diff --creation-factor=95 \
    afb8c0c2187e11f0ed45b70f5403fe3c33f73696^..f40d92a69300c6c610897a0c9d76e2580c650ee2 \
    a1bdbc1e21861a40e4c4f92a3176718e6aab6cb8^..d35b288834d501b9f06084896709613bb6c4627e
```

bolshakov-a:
Yeah, like this. Btw, there are dedicated common headers `direct.h` and `indirect.h` with `IndirectClass`, but defining a separate one in the test should go as well, if @kimgr doesn't mind.

kimgr:
Thanks! I think it's fine as-is, as you need custom headers for the UDL operator anyway.

hebasto:
Unfortunately, this PR introduced a [regression](https://github.com/include-what-you-use/include-what-you-use/issues/2057).

Issue #1759, opened 2025-05-14
Support "hidden friends"


carlosgalvezp:
Currently, IWYU does not require includes for member functions, probably as part of the "author intent":

```cpp
struct A
{
  void operator+(A const& rhs) const;
};
```

```
Ignoring use of A::operator+ (main.cpp:7:5): member of class
```

However, the more modern way to implement this is by using the "hidden friends" idiom:

```cpp
struct B
{
  friend void operator+(B const& lhs, B const& rhs);
};
```

In this case, however, IWYU considers this as a plain free-standing member function (which it is), and requires an include for it:

```
Mapped foo.h to "foo.h" for operator+ (only candidate)
```

It would be nice if the "hidden friends" pattern could also be ignored as part of the "author intent" just like it is with regular member functions.

carlosgalvezp:
The following change to `iwyu_output.cc` solves the problem, I haven't tested if it breaks something else. Essentially check not only the DeclContext (semantic) but also the LexicalDeclContext (where it's written).

```
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1330,8 +1330,12 @@ void ProcessFullUse(OneUse* use,
   }
 
   // (B5) Discard symbol uses for class members in the same file as parent.
-  if (const CXXRecordDecl* parent_decl =
-          DynCastFrom(use->decl()->getDeclContext())) {
+  const CXXRecordDecl* parent_a = DynCastFrom(use->decl()->getDeclContext());
+  const CXXRecordDecl* parent_b = DynCastFrom(use->decl()->getLexicalDeclContext());
+
+  if (parent_a || parent_b) {
+    const CXXRecordDecl* parent_decl = parent_a ? parent_a : parent_b;

```

kimgr:
I'm not sure I understand the fix, I'm not fully read up on semantic vs lexical decl contexts. What are examples of where they'd differ? I've only been reading this bug report on my phone, so I don't really see the bigger picture.

carlosgalvezp:
Semantic = where it semantically belongs, i.e. at namespace scope.
Lexical = where it's declared in the code, i.e. the scope of the class.

For example:

```cpp
namespace foo
{
class Bar
{ 
    // This function is not a member function, but a free-standing function outside of Foo
    // getDeclContext() == "namespace foo"
    // getLexicalDeclContext() == "class Bar"
    friend bool operator+(...);  
}
}
```

kimgr:
[...]

PR #2016, opened 2026-04-26
Add reduced bug test for #1250

kimgr:
I ran into this while debugging #1759, which cover hidden friend decls. When trying the suggested change, uses of Redeclarable were missed.

It turns out it's because operator!= was the only seen use of Redeclarable.h, all other uses were hidden behind UnresolvedLookupExpr due to being inside an uninstantiated template.

The testcase laid out here carries some echoes of the Redeclarable problem, but I think it's closer related to the "not-seen-in-template" problem described in #1250.

### DIFF
--- a/tests/bugs/1250/1250b.cc
+++ b/tests/bugs/1250/1250b.cc
@@ -1,0 +1,55 @@
+//===--- 1250b.cc - iwyu test ---------------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_XFAIL
+
+#include "1250b-direct.h"
+
+template<class T>
+// IWYU: Template...*needs a declaration
+void func_template(Template<T>* v) {
+  // IWYU: consume is...*1250b-i2.h
+  consume(
+    // IWYU: Template is...*1250b-i1.h
+    v->begin(),
+    // IWYU: Template is...*1250b-i1.h
+    v->end());
+}
+
+// IWYU seems to miss the two calls to begin() and end() in the function
+// template above, so Template is only reported for forward-decl use.
+
+// For contrast, a plain function with instantiated template reports everything
+// as expected. You can uncomment below to verify.
+
+// // IWYU: Template...*needs a declaration
+// void function(Template<int>* v) {
+//   // Curiosity: this reports as consume(:0, :0), not just consume.
+//   // IWYU: consume(:0, :0) is...*1250b-i2.h
+//   consume(
+//     // IWYU: Template is...*1250b-i1.h
+//     v->begin(),
+//     // IWYU: Template is...*1250b-i1.h
+//     v->end());
+// }
+
+/**** IWYU_SUMMARY
+
+tests/bugs/1250/1250b.cc should add these lines:
+#include "1250b-i1.h"
+#include "1250b-i2.h"
+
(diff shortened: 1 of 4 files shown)


kimgr:
@bolshakov-a @carlosgalvezp Let me know if this makes sense to you.

bolshakov-a:
I'm not _completely_ sure that this is a bug, because a forward-declaration of `Template` is enough actually. It should be rather a policy decision to require the `Template` definition, but such policy is not implemented yet. However, if implemented, it should rather require the definition as long as there is no explicit forward-declaration, similarly to typedef/fnreturn/autocast policy. So, probably, the test is good.

tests/bugs/1250/1250b.cc
```diff
@@ -0,0 +1,55 @@
...
+// IWYU seems to miss the two calls to begin() and end() in the function
+// template above, so Template is only reported for forward-decl use.
+
+// For contrast, a plain function with instantiated template reports everything
+// as expected. You can uncomment below to verify.
+
+// // IWYU: Template...*needs a declaration
+// void function(Template<int>* v) {
+//   // Curiosity: this reports as consume(:0, :0), not just consume.
+//   // IWYU: consume(:0, :0) is...*1250b-i2.h
```
bolshakov-a, line 41: As expected, after d0b415d5c73ccd269081b85ea9635c608462bbd7. It is more interesting why IWYU doesn't print template arguments inside `func_template`.

bolshakov-a: > It is more interesting why IWYU doesn't print template arguments inside `func_template`.

It may be a bug. Looks like `arbitrary_fn_decl` should be reported here instead of `first_decl`:
https://github.com/include-what-you-use/include-what-you-use/blob/ae18593bd2f3d4b16d8e9921aa95af0afb368490/iwyu.cc#L3047

because for templates, `getTemplatedDecl()` is called to obtain `arbitrary_fn_decl` as `FunctionDecl`.

kimgr: Yeah, this makes them consistent:
```diff
@@ -3044,7 +3044,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // only operator==() we see is in foo.h, we don't need to #include
     // foo.h if the only call to operator== we see is on two integers.)
     if (!arbitrary_fn_decl || !arbitrary_fn_decl->isOverloadedOperator())
-      ReportDeclUse(CurrentLoc(), first_decl);
+      ReportDeclUse(CurrentLoc(),
+                    arbitrary_fn_decl ? arbitrary_fn_decl : first_decl);
     return true;
   }
```

kimgr: But also breaks a couple of testcases.

kimgr:
> I'm not _completely_ sure that this is a bug, because a forward-declaration of `Template` is enough actually.

Hmm, yes! I did not expect it to be, but it looks like it.

I added a commit on top where `func_template` and `Template` are instantiated, and to me it looks bad that `consume` (an implementation detail of `func_template`) is reported there even if the language allows delaying the include until instantiation.

bolshakov-a:
> it looks bad that `consume` (an implementation detail of `func_template`) is reported there even if the language allows delaying the include until instantiation.

It is because `consume` is not considered as provided by template at instantiation site until the template definition file includes it directly. Such suggestions should go away at the second IWYU pass (assuming `func_template` definition is in its own file modified by IWYU).

tests/bugs/1250/1250b.cc
kimgr: Fix pending in #2020

PR #2019, opened 2026-05-07
Consider symbol mappings for definition headers

### LINKED ISSUE #224: seems to be some confusion with mapping files
I have this in my QT mapping file:

    { symbol: [ "QKeyEvent",      "private", "<QKeyEvent>", "public" ] },
    { symbol: [ "QMouseEvent", "private", "<QMouseEvent>", "public" ] },

in order to deal with the way QT defines these in the documentation as opposed to implementing them in the headers. However, this occasionally causes include-what-you-use to produce this output:

    x.cpp should add these lines:
    #include <QKeyEvent>          // for QKeyEvent
    #include <QMouseEvent>        // for QMouseEvent
    class QKeyEvent;
    class QMouseEvent;

which seems wrong.

bolshakov-a:
Prior to this, only include-mapped headers (if any) were taken into account to determine if a fwd-decl use is satisfied by a definition from any header that should be included.

This should fix #224.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1566,9 +1566,8 @@ void CalculateIwyuForForwardDeclareUse(
   const NamedDecl* dfn_from_desired_includes = nullptr;
   const NamedDecl* dfn_from_actual_includes = nullptr;
   for (const NamedDecl* dfn : dfns) {
-    vector<string> headers =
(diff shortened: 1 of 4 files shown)


kimgr:
Nice, thanks!

PR #2020, opened 2026-05-07
Report unresolved function template args

kimgr:
This showed up in an unrelated context:
https://github.com/include-what-you-use/include-what-you-use/pull/2016/changes#r3143816232

When a template was called from an uninstantiated template, e.g.

    template<typename T, typename U>
    void Tpl(T, U);

    template<typename T>
    void CallTpl(T v) {
      Tpl(v, int{});
    }

the IWYU diagnostics only included the function name, and not the :N
parameter types, i.e. we got:

    Tpl is defined in "blah.h" ...

instead of the expected:

    Tpl(:0, :1) is defined in "blah.h" ...

Improve the logic in VisitUnresolvedLookupExpr to report an actual
function decl rather than the first decl in the lookup set (which might
be a using-decl).

Now that we report a resolved function decl, we also have to explicitly
report any using-decl; that information is lost when arbitrary_fn_decl
is reported.

Add test coverage for both multiple-directly-included files, ADL
lookup (no use tracking) and arbitrary fn decl resolution.

Update existing test diagnostics to have parameters where necessary.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3044,8 +3044,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // end up being the built-in form of the operator.  (Even if the
     // only operator==() we see is in foo.h, we don't need to #include
     // foo.h if the only call to operator== we see is on two integers.)
-    if (!arbitrary_fn_decl || !arbitrary_fn_decl->isOverloadedOperator())
-      ReportDeclUse(CurrentLoc(), first_decl);
+    if (!arbitrary_fn_decl || !arbitrary_fn_decl->isOverloadedOperator()) {
+      ReportDeclUse(CurrentLoc(),
+                    arbitrary_fn_decl ? arbitrary_fn_decl : first_decl);
+    }
     return true;
   }
 
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -3015,24 +3015,28 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
           ReportDeclUse(CurrentLoc(), decl);
         // No AddProcessedOverloadLoc here: PublicHeaderIntendsToProvide should
         // already work well at the instantiation site.
-      } else if (!expr->requiresADL()) {
-        // Report any of the decls so as to keep the code compilable. If ADL is
-        // to be applied (i.e. the function name is unqualified), it is allowed
-        // to have no declaration.
-        ReportDeclUse(CurrentLoc(), first_decl);
-        // No AddProcessedOverloadLoc here: even first_decl may be different
-        // when scanning different translation units. It's better to
-        // double-report it on template-defn and instantiation sites than not
-        // to report an actually used decl at all.
+        return true;
+      } else if (expr->requiresADL()) {
+        // If ADL is to be applied (i.e. the function name is unqualified), it
+        // is allowed to have no declaration.
+        return true;
       }
-      return true;
     }
 
+    // Report any of the decls so as to keep the code compilable.
+    // No AddProcessedOverloadLoc here: the expr decls may be different when
+    // scanning different translation units. It's better to double-report on
+    // template-defn and instantiation sites than not to report an actually used
+    // decl at all.
     const FunctionDecl* arbitrary_fn_decl = nullptr;
     for (const NamedDecl* decl : expr->decls()) {
-      // Sometimes a UsingShadowDecl comes between us and the 'real' decl.
-      if (const UsingShadowDecl* using_shadow_decl = DynCastFrom(decl))
+      // Sometimes a UsingShadowDecl comes between us and the 'real' decl. Make
+      // sure we report it before resolving.
+      if (const UsingShadowDecl* using_shadow_decl = DynCastFrom(decl)) {
+        ReportDeclUse(CurrentLoc(), decl, nullptr, UF_None,
+                      /*report_using_decl_only = */ true);
         decl = using_shadow_decl->getTargetDecl();
+      }
       if (const FunctionDecl* fn_decl = DynCastFrom(decl)) {
         arbitrary_fn_decl = fn_decl;
         break;
@@ -3047,8 +3051,10 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     // end up being the built-in form of the operator.  (Even if the
     // only operator==() we see is in foo.h, we don't need to #include
     // foo.h if the only call to operator== we see is on two integers.)
-    if (!arbitrary_fn_decl || !arbitrary_fn_decl->isOverloadedOperator())
-      ReportDeclUse(CurrentLoc(), first_decl);
+    if (!arbitrary_fn_decl || !arbitrary_fn_decl->isOverloadedOperator()) {
+      ReportDeclUse(CurrentLoc(),
+                    arbitrary_fn_decl ? arbitrary_fn_decl : first_decl);
+    }
     return true;
   }
 
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -137,6 +137,7 @@ using namespace i1_ns2;
 using i1_ns3::i1_int_global3;
 namespace cc_ns_alias = i1_ns4;
 using i1_ns::I1_NamespaceStruct;
+// IWYU: i1_ns::I1_NamespaceTemplateFn is...*tests/cxx/badinc-i1.h
 using i1_ns::I1_NamespaceTemplateFn;
 // TODO(csilvers): mark this using declaration as redundant and remove it?
 // IWYU: i1_ns::I1_UnusedNamespaceStruct needs a declaration
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -346,7 +346,7 @@ template<class T,int> class Cc_TemplateSubclass : public H_TemplateTypedef { };
 // Let's test that we detect overloaded functions correctly when all
 // overloads are in the same file, and then when they're in different files.
 template<typename T> void CallOverloadedFunctionSameFile(T t) {
-  // IWYU: I1_OverloadedFunction is...*badinc-i1.h
+  // IWYU: I1_OverloadedFunction(const :0 &) is...*badinc-i1.h
   I1_OverloadedFunction(t);
 }
 
@@ -360,7 +360,7 @@ template<typename T> void CallOverloadWithUsingShadowDecl(T t) {
   // This is only defined in one place, but because we get to it via
   // a using expression, it generates a UsingShadowExpr.  Make sure we
   // "see through" that properly.
-  // IWYU: i1_ns::I1_NamespaceTemplateFn is...*badinc-i1.h
+  // IWYU: i1_ns::I1_NamespaceTemplateFn(:0) is...*badinc-i1.h
   I1_NamespaceTemplateFn(t);
 }
 
--- a/tests/cxx/badinc.cc
+++ b/tests/cxx/badinc.cc
@@ -347,7 +347,7 @@ template<class T,int> class Cc_TemplateSubclass : public H_TemplateTypedef { };
 // Let's test that we detect overloaded functions correctly when all
 // overloads are in the same file, and then when they're in different files.
 template<typename T> void CallOverloadedFunctionSameFile(T t) {
-  // IWYU: I1_OverloadedFunction is...*badinc-i1.h
+  // IWYU: I1_OverloadedFunction(const :0 &) is...*badinc-i1.h
   I1_OverloadedFunction(t);
 }
 
@@ -361,7 +361,7 @@ template<typename T> void CallOverloadWithUsingShadowDecl(T t) {
   // This is only defined in one place, but because we get to it via
   // a using expression, it generates a UsingShadowExpr.  Make sure we
   // "see through" that properly.
-  // IWYU: i1_ns::I1_NamespaceTemplateFn is...*badinc-i1.h
+  // IWYU: i1_ns::I1_NamespaceTemplateFn(:0) is...*badinc-i1.h
   I1_NamespaceTemplateFn(t);
 }
 
--- a/tests/cxx/overload_fn_mapping.cc
+++ b/tests/cxx/overload_fn_mapping.cc
@@ -122,6 +122,21 @@ void User() {
   TplFn(arr2);
 }
 
+// Do all template examples again inside an uninstantiated template.
+template<typename T>
+void UserTemplate(T v) {
+  // IWYU: TplFn(:0, :1) is...*-i2.h
+  TplFn(v, v);
+  // IWYU: TplFn(:1, :0 *) is...*-i3.h
+  TplFn(v, &v);
+  // IWYU: Tpl is...*-i1.h
+  // IWYU: TplFn(Tpl<:0>) is...*-i4.h
+  TplFn(Tpl<T>{});
+  T arr2[7];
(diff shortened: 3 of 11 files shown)


kimgr:
@bolshakov-a The `arbitrary_fn_decl` localized fix doesn't seem to work very well, do you have any more elaborate ideas?

bolshakov-a:
I don't know. Compilers don't even try to find out if the presented overload is acceptable:
```cpp
namespace ns {
  void Fn(int, int);
}

// This cannot be instantiated at all, but compilers don't care.
template <typename T>
void TplFn(T t) {
  // Qualified name means that ADL is not used. Fn overloads should appear before TplFn definition.
  ns::Fn(t);
}
```
https://godbolt.org/z/PaEsaTYcP

Nevertheless, it is usually better to require the public header of some weird overload than its private header.

kimgr:
So I guess we could wildcard-match overloads in tests? I pushed a commit on top that does that for the new test, but something similar would have to happen for the remaining failures in badinc and placement_new tests.

bolshakov-a:
> So I guess we could wildcard-match overloads in tests?

I think so, as well.

tests/cxx/overload_fn_mapping.cc
```diff
@@ -122,6 +122,24 @@ void User() {
...
 }
 
+// Do all template examples again inside an uninstantiated template.
+// TODO: note that all uses are reported, but IWYU doesn't know how to
+// distinguish between overloads in uninstantiated contexts, so we fuzzy-match
+// overload and definition filename.
+template <typename T>
+void UserTemplate(T v) {
+  // IWYU: TplFn(...*) is...*-i{{[0-9]}}.h
+  TplFn(v, v);
```
bolshakov-a: I've realized that IWYU should better not report here anything at all because it is allowed not to provide any `TplFn` declaration here, no one is included directly, and there are different _public_ headers for the declarations. The problem is that IWYU currently doesn't consider _public_ (mapped) headers to determine `same_file` (and all the decls here are really in the same _private_ file).

bolshakov-a: So I think that a more correct test case should involve some other template which maps to only one public file along with all its overloads, and/or a template inside some namespace so as to block ADL.

tests/cxx/badinc.cc
```diff
@@ -137,6 +137,7 @@ using namespace i1_ns2;
 using i1_ns3::i1_int_global3;
 namespace cc_ns_alias = i1_ns4;
 using i1_ns::I1_NamespaceStruct;
+// IWYU: i1_ns::I1_NamespaceTemplateFn is...*tests/cxx/badinc-i1.h
 using i1_ns::I1_NamespaceTemplateFn;
```
bolshakov-a, line 140: This shows another bug. IWYU reports now the decl for non-referenced using-decl, but it is actually referenced below.

iwyu.cc
bolshakov-a: Note that `first_decl` is still reported here.

tests/cxx/overload_fn_mapping.cc
kimgr: > So I think that a more correct test case should involve some other template which maps to only one public file along with all its overloads, and/or a template inside some namespace so as to block ADL.

I tried this in a separate test case, and it gives much more understandable results. Relying on unqualified/ADL lookups is what makes everything confusing.

tests/cxx/badinc.cc
kimgr: Fixed, we need to report the using-decl separately. Before this patch, `first_decl` *was* the using-decl, so `ReportDeclUse` took care of it automatically.

iwyu.cc
kimgr: Fixed

kimgr:
Updated. It seems like we had two paths for reporting an arbitrary decl, also before the arbitrary_fn_decl handling block:

```diff
-      } else if (!expr->requiresADL()) {
-        // Report any of the decls so as to keep the code compilable. If ADL is
-        // to be applied (i.e. the function name is unqualified), it is allowed
-        // to have no declaration.
```

I tweaked that to let it fall through to the subsequent arbitrary lookup instead.

kimgr:
@bolshakov-a Actually, I wanted to ask you about a related thing: https://github.com/include-what-you-use/include-what-you-use/blob/b9853a1927e8a51032386ed7ce4a557d5f71d18a/iwyu.cc#L3003-L3006

I've been thinking about that sort of on the side. I wonder how hard it would be to build an `iwyu_lookup` module that can do IWYU-tailored lookup of unresolved decls? It looks like `OverloadExpr` have all the candidates already, so if we could model it as a lookup from `OverloadExpr` => one decl (or possibly an ordered list of decls), we may be able to move some complexity out of the main IWYU flow. Particularly `AddProcessedOverloadLoc` and `IsProcessedOverloadLoc` seem pretty hacky to me, I wonder if that could all just be handled in `HandleFunctionCall` if the `CallExpr` has an `OverloadExpr` below it, or something.

iwyu.cc
```diff
@@ -3015,24 +3015,19 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
-        // to be applied (i.e. the function name is unqualified), it is allowed
-        // to have no declaration.
-        ReportDeclUse(CurrentLoc(), first_decl);
-        // No AddProcessedOverloadLoc here: even first_decl may be different
-        // when scanning different translation units. It's better to
-        // double-report it on template-defn and instantiation sites than not
-        // to report an actually used decl at all.
+      } else if (expr->requiresADL()) {
+        return true;
       }
```
bolshakov-a, line 3023: Why don't you want to keep some comment? It may be not obvious why ADL is important here.

bolshakov-a:
> I wonder how hard it would be to build an iwyu_lookup module that can do IWYU-tailored lookup of unresolved decls?

I think it is of low priority. Users can choose by themselves which overload candidates they want to provide along with the template.

> Particularly `AddProcessedOverloadLoc` and `IsProcessedOverloadLoc` seem pretty hacky to me, I wonder if that could all just be handled in `HandleFunctionCall` if the `CallExpr` has an `OverloadExpr` below it, or something.

Note that it is not used with `UnresolvedLookupExpr` at all now, only with `CXXNewExpr`.

iwyu.cc
kimgr: Restored/moved around.

iwyu.cc
```diff
@@ -3015,24 +3015,27 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
           ReportDeclUse(CurrentLoc(), decl);
         // No AddProcessedOverloadLoc here: PublicHeaderIntendsToProvide should
         // already work well at the instantiation site.
```
kimgr, line 3017: I wonder if we should `return true;` here to avoid falling through to the arbitrary-fn-decl reporting below? I tried adding it and no tests fail...

bolshakov-a: There was one in the line 3028. I think you should restore it, otherwise IWYU can report `arbitrary_fn_decl` not directly included and hence not intended by user for provision.

kimgr: Done

kimgr:
@bolshakov-a I'm feeling ready to merge this, did you want to take another look first?

tests/cxx/overload_fn_tpl.cc
```diff
@@ -0,0 +1,47 @@
...
+
+// Template examples mirrored from overload_fn_mapping.cc, but here without
+// mappings and inside an uninstantiated template. This is mostly to test that
+// function template arguments are properly rendered in diagnostics.
+template <typename T>
+void UserTemplate(T v) {
+  // IWYU: ns::TplFn(:0, :1) is...*-i1.h
+  ns::TplFn(v, bool{});
+  // IWYU: ns::TplFn(:0, :1) is...*-i1.h
+  ns::TplFn(1, 2);
```
bolshakov-a: Are you sure this one and below are `UnresolvedLookupExpr`s?

kimgr: Ah, no, they are not. I can make them so by passing `T` types or values, but then they fall back into the `arbitrary_fn_decl` resolution and are all reported as `ns::TplFn(:0, :1)`. It was a relief to not end up in that approximate reporting, but maybe that's actually what we want to get at here. Not sure what's the better test input here...

bolshakov-a: While [my TODO you were writing about](https://github.com/include-what-you-use/include-what-you-use/pull/2020#issuecomment-5231781028) is not done, such additional test cases are useless.

kimgr: You're thinking just reduce it like so:
```diff
diff --git a/tests/cxx/overload_fn_tpl-i1.h b/tests/cxx/overload_fn_tpl-i1.h
index 66971be6..72bca97c 100644
--- a/tests/cxx/overload_fn_tpl-i1.h
+++ b/tests/cxx/overload_fn_tpl-i1.h
@@ -8,20 +8,8 @@
 //===----------------------------------------------------------------------===//
 
 namespace ns {
-template <typename>
-class Tpl {};
 
 template <typename T, typename U>
 int TplFn(T, U);
 
-template <typename T>
-int TplFn(auto, T*);
-
-template <typename T>
-int TplFn(Tpl<T>);
-
-// Something like std::swap for arrays.
-template <class T, int N>
-void TplFn(T (&x)[N]);
-
 }  // namespace ns
diff --git a/tests/cxx/overload_fn_tpl.cc b/tests/cxx/overload_fn_tpl.cc
index e97b4b8e..2c5a2bd9 100644
--- a/tests/cxx/overload_fn_tpl.cc
+++ b/tests/cxx/overload_fn_tpl.cc
@@ -11,26 +11,13 @@
 
 #include "tests/cxx/overload_fn_tpl-d1.h"
 
-// Template examples mirrored from overload_fn_mapping.cc, but here without
-// mappings and inside an uninstantiated template. This is mostly to test that
-// function template arguments are properly rendered in diagnostics.
+// Function template instantiation with dependent type inside an uninstantiated
+// template. This is mostly to test that function template arguments are
+// properly rendered in diagnostics.
 template <typename T>
 void UserTemplate(T v) {
   // IWYU: ns::TplFn(:0, :1) is...*-i1.h
   ns::TplFn(v, bool{});
-  // IWYU: ns::TplFn(:0, :1) is...*-i1.h
-  ns::TplFn(1, 2);
-
-  int i = 0;
-  // IWYU: ns::TplFn(:1, :0 *) is...*-i1.h
-  ns::TplFn(1, &i);
-  // IWYU: ns::Tpl is...*-i1.h
-  // IWYU: ns::TplFn(ns::Tpl<:0>) is...*-i1.h
-  ns::TplFn(ns::Tpl<int>{});
-
-  int arr[7];
-  // IWYU: ns::TplFn(:0 (&)[:1]) is...*-i1.h
-  ns::TplFn(arr);
 }
 
 /**** IWYU_SUMMARY
@@ -42,6 +29,6 @@ tests/cxx/overload_fn_tpl.cc should remove these lines:
 - #include "tests/cxx/overload_fn_tpl-d1.h"  // lines XX-XX
 
 The full include-list for tests/cxx/overload_fn_tpl.cc:
-#include "tests/cxx/overload_fn_tpl-i1.h"  // for Tpl, TplFn
+#include "tests/cxx/overload_fn_tpl-i1.h"  // for TplFn
 
 ***** IWYU_SUMMARY */
```
?

bolshakov-a: Yes. But maybe the case of function template declarations from different files (`same_file == false`) is worth adding as well, not sure.

kimgr: Done + an `Undefined(v);` call to provoke ADL (though I don't think it hits the `expr->requiresADL()` branch, because  the lack of overload candidates `decls_begin() == decls_end()` will probably catch that early).

tests/cxx/overload_fn_tpl.cc
```diff
@@ -0,0 +1,45 @@
...
+// template. This is mostly to test that function template arguments are
+// properly rendered in diagnostics.
+template <typename T>
+void UserTemplate(T v) {
+  // IWYU: ns::TplFn(:0, :1) is...*-i1.h
+  ns::TplFn(v, bool{});
+
+  // No diagnostic here, the decls need to be directly included to satisfy use
+  // of both overloads. Both -d2 and -d3 are preserved in the summary below.
+  ns::TplFnOverload(v, int{});
```
bolshakov-a: I'd expect you put `TplFnOverload` into indirect headers so as to trigger diagnostics suggesting some of them.

kimgr: Would that still qualify as "directly-included", or would I need .imp mappings to make that happen?

kimgr: Hmm, I can't get directly-included to trigger even with this .imp:
```json
[
  { "symbol": ["ns::TplFnOverload(:0, :1)", "private", "\"tests/cxx/overload_fn_tpl-i2.h\"", "public"] },
  { "symbol": ["ns::TplFnOverload(:0, int)", "private", "\"tests/cxx/overload_fn_tpl-i3.h\"", "public"] }
]
```

bolshakov-a: > Would that still qualify as "directly-included", or would I need .imp mappings to make that happen?

Neither. I mean that `VisitUnresolvedLookupExpr` from the master branch has two places where `first_decl` is reported: when all the candidate functions are in the same file: https://github.com/include-what-you-use/include-what-you-use/blob/06fe577d33deb0e3cf50505717980552d64442fb/iwyu.cc#L3051 and when they are in different files: https://github.com/include-what-you-use/include-what-you-use/blob/06fe577d33deb0e3cf50505717980552d64442fb/iwyu.cc#L3022
It would be good to test that the behavior with function templates is fixed in both cases.

bolshakov-a: > I can't get directly-included to trigger even with this .imp

Probably, because your public headers `-i2.h` and `-i3.h` are not directly included actually? Anyway, I don't suggest testing that "directly-included" logic because you didn't change it.

kimgr: I see, thanks. I added scenarios for all of:

* TplFn with overloads in two different headers; arbitrary choice
* TplFnDirect with overloads in two different directly included headers; keep both
* Undefined; assumes ADL, ignored

kimgr: > Probably, because your public headers -i2.h and -i3.h are not directly included actually?

Exactly, yeah.

>  Anyway, I don't suggest testing that "directly-included" logic because you didn't change it.

Right, I re-added it before I saw this comment... Now I see it's almost covered in https://github.com/include-what-you-use/include-what-you-use/blob/master/tests/cxx/overload_expr-d1.h#L13 -- except for the fact that multiple headers will be kept.

bolshakov-a:
Now you have lost the case of `same_file == true`. And I don't think you need the separate `-d2.h` header, you can include `-i1.h` and `-i2.h` through `-d1.h`.

kimgr:
> Now you have lost the case of `same_file == true`. And I don't think you need the separate `-d2.h` header, you can include `-i1.h` and `-i2.h` through `-d1.h`.

Hahaha, this seems to be an unsolvable problem for me :). I'll revisit.

tests/cxx/overload_fn_tpl.cc
kimgr: Updated again, added a same-file case and improved some of the explanations.

bolshakov-a:
LGTM, thanks!

kimgr:
Thank you!

PR #2021, opened 2026-05-07
Fix header spelling in fwd-decl uses

bolshakov-a:
For the case of quoted includes with the path relative to the includer, `ConvertToQuotedInclude` with the defaulted second parameter returned the absolute file path which did not match any of the include lines inside `CalculateDesiredIncludesAndForwardDeclares`. Due to that, `(ptr only)` comment did not appear for the corresponding `#include` line. This fixes such behavior.

As an alternative fix, the correct 2nd argument could probably be passed to `ConvertToQuotedInclude`.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1565,12 +1565,15 @@ void CalculateIwyuForForwardDeclareUse(
   // fact.  Also if it's defined in one of the actual_includes.
   const NamedDecl* dfn_from_desired_includes = nullptr;
   const NamedDecl* dfn_from_actual_includes = nullptr;
+  string desired_hdr_with_dfn;
   for (const NamedDecl* dfn : dfns) {
     vector<string> headers = GlobalIncludePicker().GetMappedPublicHeaders(
         dfn, GetFilePath(use->use_loc()), GetFilePath(dfn));
     for (const string& header : headers) {
-      if (ContainsKey(desired_includes, header))
+      if (ContainsKey(desired_includes, header)) {
         dfn_from_desired_includes = dfn;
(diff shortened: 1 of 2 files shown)


iwyu_output.cc
```diff
@@ -1565,21 +1565,22 @@ void CalculateIwyuForForwardDeclareUse(
...
     for (const string& header : headers) {
-      if (ContainsKey(desired_includes, header))
+      if (ContainsKey(desired_includes, header)) {
         dfn_from_desired_includes = dfn;
+        desired_hdr_with_dfn = header;
+      }
       if (ContainsKey(actual_includes, header))
         dfn_from_actual_includes = dfn;
     }
-    // We ourself are always a 'desired' and 'actual' include (though
```
kimgr: It's not clear to me how the same-file changes relate to the mapped-header spelling...? The commit message also says "Moreover, ...", which seems to me there's more than one thing going on. Do they interact somehow?

bolshakov-a: It is just to not think about what `desired_hdr_with_dfn` should be set to in such cases. Because if `dfn_from_desired_includes` is set, `desired_hdr_with_dfn` is expected to be set as well (see code below).

kimgr: Ah, I see. So maybe do the simplification of same-file as a separate patch first? I think that would make things clearer.

bolshakov-a: Done in #2022. (Looks like a year.)

kimgr: Thanks, merged! Looks like we're getting closer to current year, I wonder who's going to get that issue/PR number :-)

PR #2023, opened 2026-05-10
Don't force exported fwd-decl when dfn is present

bolshakov-a:
Recategorizing fwd-decl uses to full uses switched off the handling from the D1 step, so that if the definition should be already present, IWYU still suggested the header with the forward-declaration along the header with the definition.

This change fixes the bug by means of keeping the uses of exported forward-declarations as fwd-decl uses and setting a suggested header to them. `CalculateDesiredIncludesAndForwardDeclares` has been changed so that any use with a suggested header require an include line, not only full uses.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1229,7 +1229,9 @@ void ProcessForwardDeclare(OneUse* use,
       const auto* decl = cast<NamedDecl>(redecl);
       if (preprocessor_info->ForwardDeclareIsExported(decl)) {
         use->reset_decl(decl);
-        use->set_full_use();
+        use->set_suggested_header(ConvertToQuotedInclude(
+            GetFilePath(decl->getLocation()),
+            MakeAbsolutePath(GetParentPath(GetFilePath(use->use_loc())))));
         break;
       }
     }
@@ -1862,13 +1864,12 @@ void CalculateDesiredIncludesAndForwardDeclares(
     if (use.ignore_use())
       continue;
 
-    if (use.is_full_use()) {
-      CHECK_(use.has_suggested_header() && "Full uses should have #includes");
+    if (use.has_suggested_header()) {
       if (!Contains(*lines, use.suggested_header())) { // must be added
         lines->push_back(OneIncludeOrForwardDeclareLine(
             use.decl_file(), use.suggested_header(), -1));
       }
-    } else if (!use.has_suggested_header()) {
+    } else {
       // Forward-declare uses that are already satisfied by an #include
       // have that as their suggested_header.  For the rest, we need to
       // make sure there's a forward-declare in the current file.
@@ -1934,13 +1935,14 @@ void CalculateDesiredIncludesAndForwardDeclares(
       for (auto it = range.first; it != range.second; ++it) {
         it->second->set_desired();
       }
-    } else if (ContainsKey(include_map, use.suggested_header())) {
-      // If we satisfy a forward-declare use from a file, let the file
-      // know (this is just for logging).
+    } else {
+      // If we satisfy a forward-declare use from a file, let the file know.
       const string symbol_name = use.short_symbol_name();
 
       auto range = include_map.equal_range(use.suggested_header());
+      CHECK_(range.first != range.second);
       for (auto it = range.first; it != range.second; ++it) {
+        it->second->set_desired();
         if (!it->second->HasSymbolUse(symbol_name))
           it->second->AddSymbolUse(symbol_name + " (ptr only)");
       }
--- a/tests/cxx/pragma_export_fwd-d1.h
+++ b/tests/cxx/pragma_export_fwd-d1.h
@@ -7,6 +7,9 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "tests/cxx/pragma_export_fwd-i1.h"
+#include "tests/cxx/pragma_export_fwd-i2.h"
+
 class FwdDecl3;  // IWYU pragma: export
 
 // IWYU pragma: begin_exports
--- a/tests/cxx/pragma_export_fwd-i1.h
+++ b/tests/cxx/pragma_export_fwd-i1.h
@@ -1,0 +1,10 @@
+//===--- pragma_export_fwd-i1.h - test input file for iwyu ----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+class FullAndFwdDeclUse {};
--- a/tests/cxx/pragma_export_fwd-i2.h
+++ b/tests/cxx/pragma_export_fwd-i2.h
@@ -1,0 +1,10 @@
+//===--- pragma_export_fwd-i2.h - test input file for iwyu ----------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+class FullAndFwdDeclUse;  // IWYU pragma: export
--- a/tests/cxx/pragma_export_fwd.cc
+++ b/tests/cxx/pragma_export_fwd.cc
@@ -14,18 +14,28 @@
 // Uses do not trigger warnings as they are already provided for.
 void a(const FwdDecl1&);
 void b(const FwdDecl2*);
+// IWYU: FwdDecl3 needs a declaration
 void c(const FwdDecl3&);
+// IWYU: FwdDecl4 needs a declaration
 void d(const FwdDecl4*);
+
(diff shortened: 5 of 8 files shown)


iwyu_output.cc
```diff
@@ -1229,7 +1229,9 @@ void ProcessForwardDeclare(OneUse* use,
       const auto* decl = cast<NamedDecl>(redecl);
       if (preprocessor_info->ForwardDeclareIsExported(decl)) {
         use->reset_decl(decl);
-        use->set_full_use();
+        use->set_suggested_header(ConvertToQuotedInclude(
+            GetFilePath(decl->getLocation()),
+            MakeAbsolutePath(GetParentPath(GetFilePath(use->use_loc())))));
```
bolshakov-a, line 1234: I'm not sure if some include mapping could make sense along with IWYU pragmas. If it could, `GetCandidateHeadersForFilepath` should better be used here (and some appropriate header should be selected from the returned array).

bolshakov-a:
The failed (succeeded?) bug test is unrelated. The change in behavior is apparently due to newer clang revision.

kimgr:
Has to be this one, right: https://github.com/llvm/llvm-project/commit/fb024337497f4c158ea16bfb90e50138d9a2a0c6? I think it came through in the Ubuntu packages we use for CI this morning.

kimgr:
I guess this default `Visit` implementation causes the `TemplateArgumentLoc`s to be visited where they were formerly overlooked: https://github.com/llvm/llvm-project/commit/fb024337497f4c158ea16bfb90e50138d9a2a0c6#diff-2e0b189003b6b8daa54926f1f71c88c5fb883080b5e37c6ce65079c62d33a1eeR686-R697.

But it also looks like the whole design for explicit instantiations has changed, so I wonder if we need to adjust something in IWYU or if we can simplify.

iwyu_output.cc
kimgr: I don't fully understand what you're suggesting. Can you give an example?

kimgr:
Yes, looks like it. Here's the `-v7` diff between IWYU with Clang 22 vs. IWYU with Clang 23 at latest:
```diff
diff --git a/before.txt b/after.txt
index eedb8573..d9c1ab8b 100644
--- a/before.txt
+++ b/after.txt
@@ -1,7 +1,7 @@
 INFO:root:Testing iwyu on tests/bugs/958/958.cc
->>> Running /home/kimgr/code/iwyu/out/main+deb22/bin/include-what-you-use -Xiwyu --verbose=3 -Xiwyu --verbose=7 tests/bugs/958/958.cc
+>>> Running /home/kimgr/code/iwyu/out/main+deb/bin/include-what-you-use -Xiwyu --verbose=3 -Xiwyu --verbose=7 tests/bugs/958/958.cc
 Setting verbose-level to 7
-Search path: /usr/lib/llvm-22/lib/clang/22/include/ (system)
+Search path: /usr/lib/llvm-23/lib/clang/23/include/ (system)
 Search path: /usr/include/x86_64-linux-gnu/c++/14/ (system)
 Search path: /usr/include/x86_64-linux-gnu/ (system)
 Search path: /usr/include/c++/14/backward/ (system)
@@ -62,6 +62,15 @@ tests/bugs/958/958.cc:17:13: (5) [ CXXDependentScopeMemberExpr ]  CXXDependentSc
 tests/bugs/958/958.cc:17:10: (6) [ DeclRefExpr ]  DeclRefExpr  'T' lvalue ParmVar  'val' 'T'
 
 Marked full-info use of decl  val (from tests/bugs/958/958.cc:16:17) at tests/bugs/958/958.cc:17:10
+tests/bugs/958/958.cc:22:1: (1) [ ExplicitInstantiationDecl ]  template int templFunc<X>(X val)
+tests/bugs/958/958.cc:22:10: (2) [ FunctionProtoTypeLoc ] int (X)
+tests/bugs/958/958.cc:22:10: (3) [ BuiltinTypeLoc ] int
+tests/bugs/958/958.cc:22:29: (3) [ ParmVarDecl ]  X val
+tests/bugs/958/958.cc:22:27: (4) [ RecordTypeLoc ] X
+Marked full-info use of decl  X (from tests/bugs/958/file1.h:10:8) at tests/bugs/958/958.cc:22:27
+tests/bugs/958/958.cc:22:24: (2) [ TemplateArgumentLoc ]  <X>
+tests/bugs/958/958.cc:22:24: (3, fwd decl) [ RecordTypeLoc ] X
+Marked fwd-decl use of decl  X (from tests/bugs/958/file1.h:10:8) at tests/bugs/958/958.cc:22:24
 --- Calculating IWYU violations for tests/bugs/958/958.h ---
 
 (tests/bugs/958/958.h has correct #includes/fwd-decls)
@@ -69,20 +78,19 @@ Marked full-info use of decl  val (from tests/bugs/958/958.cc:16:17) at tests/bu
 Ignoring full use of val (tests/bugs/958/958.cc:17:10): declared inside a function
 Marked use of include-file "958.h" (from tests/bugs/958/958.h) at tests/bugs/958/958.cc:1:1
 Mapped tests/bugs/958/958.h to "958.h" for templFunc(:0) (only candidate)
+Mapped tests/bugs/958/file1.h to "file1.h" for X (only candidate)
 Ignoring full use of templFunc(:0) (tests/bugs/958/958.cc:16:5): #including dfn from "958.h"
+Ignoring full use of X (tests/bugs/958/958.cc:22:27): #including dfn from "file1.h"
+Noting fwd-decl use of X (tests/bugs/958/958.cc:22:24) is satisfied by dfn in tests/bugs/958/file1.h:10:8
+Ignoring fwd-decl use of X (tests/bugs/958/958.cc:22:24): have definition at tests/bugs/958/file1.h:10:8
 Ignoring full use of  (tests/bugs/958/958.cc:1:1): #including dfn from "958.h"
 
-tests/bugs/958/958.cc should add these lines:
+(tests/bugs/958/958.cc has correct #includes/fwd-decls)
```

kimgr:
Attempt at test coverage: https://github.com/include-what-you-use/include-what-you-use/pull/2024

bolshakov-a:
> I guess this default `Visit` implementation causes the `TemplateArgumentLocs` to be visited

To be more correct, the full use of the type is reported not from the traversing of template arguments but of the function prototype in `RecursiveASTVisitor`, as you can see from the log. When scanning an ordinary function declaration, IWYU calls `set_in_forward_declare_context` if the declaration is not a definition:
https://github.com/include-what-you-use/include-what-you-use/blob/c36eacb7f104dd4349c0ca1dcc0e23433b52c423/iwyu.cc#L1765

If the function declaration is actually the definition, IWYU reports types of parameters likewise other variable types. But then `FunctionProtoTypeLoc` appears outside of any `FunctionDecl`, `set_in_forward_declare_context` is not called, and reporting parameter types is a bug:
```cpp
#include "tests/cxx/direct.h"

int main() {
  // IWYU wrongly reports that `IndirectClass` should be complete.
  int (*p)(IndirectClass);
}
```
The same situation is now with `ExplicitInstantiationDecl`. Whereas IWYU correctly reports now parameter (ordinary function parameters, not template parameters) types for explicit instantiation _definitions_, it reports them for explicit instantiation _declarations_ as well. Thus, one bug has been replaced with another bug.

iwyu_output.cc
bolshakov-a: I mean e.g.
```cpp
// private.h
class Class;  // IWYU pragma: export
```
and the following mapping file:
```yaml
[
{ "include": ["\"private.h\"", "private", "\"public1.h\"", "public"] },
{ "include": ["\"private.h\"", "private", "\"public2.h\"", "public"] },
]
```
Probably, such cases should not be considered because IWYU pragmas are for user's code and IWYU-friendly libraries, whereas mappings are for external libraries which are not IWYU-friendly.

kimgr: Yeah, that rationale makes sense to me. But conversely, is there a reason not to use `GetCandidateHeadersForFilepath`? More complexity in choosing which header to use?

iwyu_output.cc
```diff
@@ -1934,13 +1935,15 @@ void CalculateDesiredIncludesAndForwardDeclares(
       for (auto it = range.first; it != range.second; ++it) {
         it->second->set_desired();
       }
-    } else if (ContainsKey(include_map, use.suggested_header())) {
+    } else {
       // If we satisfy a forward-declare use from a file, let the file
       // know (this is just for logging).
```
kimgr: Is this comment -- `(this is just for logging)` -- now out of date? Seeing as how we're now setting the include-line as desired?

iwyu_output.cc
bolshakov-a: Yeah, exactly.

iwyu_output.cc
bolshakov-a: Indeed, thanks!

iwyu_output.cc
kimgr: Yeah, then good enough as-is!

Issue #1256, opened 2023-05-07
IWYU suggests replacing forward declaration with include when forward declaration is (maybe?) sufficient


ts826848:
I have a struct containing a function pointer with a non-built-in return type. I think the return type is forward-declarable in this case, but IWYU appears to favor including the header containing the definition instead.

I haven't been able to find a part of the standard which explicitly says that function pointer return types are forward-declarable, so I'm not 100% certain this is a bug, but since return types are forward-declarable in for free/member functions I'm assuming it's the same for function pointers

`results.hpp`:

```cpp
#ifndef RESULTS_HPP_INCLUDED
#define RESULTS_HPP_INCLUDED

struct results {};

#endif // RESULTS_HPP_INCLUDED
```

`test.hpp`:

```cpp
#ifndef TEST_HPP_INCLUDED
#define TEST_HPP_INCLUDED

struct results;
struct funcs {
  results (*get_results)();
};

#endif // TEST_HPP_INCLUDED
```

`test.cpp`:

```cpp
#include "test.hpp"
#include "results.hpp"

static auto f() { return results{}; }
static auto g() { funcs fs = {.get_results = &f}; }
```

IWYU output with `--verbose=10`:

```
% clang++ -std=c++17 -c test.cpp && include-what-you-use -Xiwyu --verbose=10 test.cpp
Setting verbose-level to 10
<[Removed "Adding mapping from ..."]>
<[Removed "Search path: ..."]>
Found include: <built-in>:-1 -> test.cpp
Added to main compilation unit: test.cpp
[ Exiting to  ] test.cpp:1:1
[ #include    ] "test.hpp" (test.hpp)
Found include: test.cpp:1 -> test.hpp
Adding test.hpp as associated header for test.cpp
Marked test.hpp as associated header of test.cpp.
Added an #include: test.cpp -> "test.hpp"
Added to main compilation unit: test.hpp
[ #ifndef     ] test.hpp:1:9: TEST_HPP_INCLUDED
[ #define     ] test.hpp:2:9: TEST_HPP_INCLUDED
[ Exiting to  ] test.cpp:2:1
[ #include    ] "results.hpp" (results.hpp)
Found include: test.cpp:2 -> results.hpp
Added an #include: test.cpp -> "results.hpp"
[ #ifndef     ] results.hpp:1:9: RESULTS_HPP_INCLUDED
[ #define     ] results.hpp:2:9: RESULTS_HPP_INCLUDED
[...]

PR #2024, opened 2026-05-10
Add test for explicit instantiation declarations/definitions

kimgr:
https://github.com/llvm/llvm-project/commit/fb024337497 added an
ExplicitInstantiationDecl in the AST, which affected how Clang visits
template arguments for explicit instantiations in its default
RecursiveASTVisitor.

Try to capture our current semantics in a regression test, and mark up
desired semantics with TODOs. Quite a bit of disconnect currently.

I've checked all TODOs with Clang's behavior as a golden standard.

### DIFF
--- a/tests/cxx/expl_inst_args.cc
+++ b/tests/cxx/expl_inst_args.cc
@@ -1,0 +1,103 @@
+//===--- expl_inst_args.cc - iwyu test ------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// This test checks that template arguments are reported from explicit
+// instantiation declarations and definitions. For the relationship between
+// explicit instantiations and their backing template, see other expl_inst_*
+// tests.
+
+// IWYU_ARGS: -I .
+
+#include "tests/cxx/expl_inst_args-d1.h"
+
+// Example covering all relevant entities, based on:
+// https://github.com/llvm/llvm-project/issues/191442
+template <typename T>
+void foo(T x) {}                          // (a) function template
+template <typename T> T bar = T{};        // (b) variable template
+template <typename T> struct S {          // (c) class template
+  void method(T x) {}                     // (d) member function
+  static T sval;                          // (e) static data member
+  template <typename U> void tmpl(U u) {} // (f) member function template
+  struct Inner { T val; };                // (g) nested class
+};
+template <typename T>
+void body() { T x; }                      // (h) function template body
+// TODO: These are misattributed from explicit instantiation definitions.
+// IWYU: Struct1 needs a declaration
+// IWYU: Struct2 needs a declaration
+// IWYU: Struct1 is...*expl_inst_args-i1.h
+// IWYU: Struct2 is...*expl_inst_args-i1.h
+template <typename T> T S<T>::sval = T{}; // (i) template static member dfn
+
+// Part 1: add explicit instantiation declarations for these template kinds.
+
+// TODO: Language does not need complete Struct1 here.
+// IWYU: Struct1 needs a declaration
+// IWYU: Struct1 is...*expl_inst_args-i1.h
+extern template void foo<Struct1>(Struct1);              // (a)
+// IWYU: Struct1 needs a declaration
+extern template Struct1 bar<Struct1>;                    // (b)
+// TODO: Language wants complete Struct1 for nested S::Inner.
+// IWYU: Struct1 needs a declaration
+extern template struct S<Struct1>;                       // (c)
+// TODO: Language does not need complete Struct2 here.
+// IWYU: Struct2 needs a declaration
+// IWYU: Struct2 is...*expl_inst_args-i1.h
+extern template void S<Struct2>::method(Struct2);        // (d)
+// IWYU: Struct2 needs a declaration
+extern template Struct2 S<Struct2>::sval;                // (e) & (i)
+// TODO: Language does not need complete Struct2 here.
+// IWYU: Struct1 needs a declaration
+// IWYU: Struct2 needs a declaration
+// IWYU: Struct2 is...*expl_inst_args-i1.h
+extern template void S<Struct1>::tmpl<Struct2>(Struct2); // (f)
+// IWYU: Struct2 needs a declaration
+extern template struct S<Struct2>::Inner;                // (g)
+// IWYU: Struct1 needs a declaration
+extern template void body<Struct1>();                    // (h)
+
+// Part 2: add explicit instantiation definitions for these template kinds.
+
+// IWYU: Struct1 needs a declaration
+// IWYU: Struct1 is...*expl_inst_args-i1.h
+template void foo<Struct1>(Struct1);              // (a)
+// TODO: Language wants complete Struct1 here for bar dfn.
+template Struct1 bar<Struct1>;                    // (b)
+// IWYU: Struct1 needs a declaration
+// IWYU: Struct1 is...*expl_inst_args-i1.h
+template struct S<Struct1>;                       // (c)
+// IWYU: Struct2 needs a declaration
+// IWYU: Struct2 is...*expl_inst_args-i1.h
+template void S<Struct2>::method(Struct2);        // (d)
+// TODO: Language wants complete Struct2 here for sval static member dfn.
+// IWYU: Struct2 needs a declaration
+template Struct2 S<Struct2>::sval;                // (e) & (i)
+// IWYU: Struct1 needs a declaration
+// IWYU: Struct2 needs a declaration
+// IWYU: Struct2 is...*expl_inst_args-i1.h
+template void S<Struct1>::tmpl<Struct2>(Struct2); // (f)
+// IWYU: Struct2 needs a declaration
+template struct S<Struct2>::Inner;                // (g)
+// TODO: Language wants complete Struct1 here for the body instantiation.
+// IWYU: Struct1 needs a declaration
+template void body<Struct1>();                    // (h)
+
+/**** IWYU_SUMMARY
+
+tests/cxx/expl_inst_args.cc should add these lines:
+#include "tests/cxx/expl_inst_args-i1.h"
+
+tests/cxx/expl_inst_args.cc should remove these lines:
+- #include "tests/cxx/expl_inst_args-d1.h"  // lines XX-XX
+
+The full include-list for tests/cxx/expl_inst_args.cc:
+#include "tests/cxx/expl_inst_args-i1.h"  // for Struct1, Struct2
+
+***** IWYU_SUMMARY */
--- a/tests/cxx/expl_inst_args-d1.h
+++ b/tests/cxx/expl_inst_args-d1.h
@@ -1,0 +1,10 @@
+//===--- expl_inst_args-d1.h - iwyu test ----------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/expl_inst_args-i1.h"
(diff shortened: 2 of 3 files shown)


kimgr:
@bolshakov-a I'm hoping you can help me reason about what the intended behavior is for the respective template kinds.

tests/cxx/expl_inst_args.cc
```diff
@@ -0,0 +1,67 @@
...
+// IWYU: Struct2 is...*expl_inst_args-i1.h
+template void ns::S<Struct2>::method(Struct2);          // (d)
+// IWYU: Struct2 needs a declaration
+// IWYU: Struct2 is...*expl_inst_args-i1.h
+template Struct2 ns::S<Struct2>::sval;                  // (e)
+// IWYU: Struct1 needs a declaration
+// IWYU: Struct1 is...*expl_inst_args-i1.h
+// IWYU: Struct2 needs a declaration
+// IWYU: Struct2 is...*expl_inst_args-i1.h
+template void ns::S<Struct1>::tmpl<Struct2>(Struct2);   // (f)
```
bolshakov-a: `Struct1` actually need not be complete here.

bolshakov-a (commented):
In fact, `explicit_instantiation` test already has some test cases regarding class template argument handling.

In general, the new test cases look good (but explicit instantiation _declarations_ may be worth adding), IWYU just doesn't currently handle all of them properly.

bolshakov-a:
> I'm hoping you can help me reason about what the intended behavior is for the respective template kinds.

I just want to note that you can always check the behavior (whether the code compiles without the full type info or not) with your compiler or at Godbolt's website (as I often do actually).

kimgr:
> > I'm hoping you can help me reason about what the intended behavior is for the respective template kinds.
>
> I just want to note that you can always check the behavior (whether the code compiles without the full type info
> or not) with your compiler or at Godbolt's website (as I often do actually).

Ah, yeah, of course. But is that the only goal, to suggest what the compiler would accept? I've been lost in woods of author intent so long that I no longer trust my judgment, but I guess that doesn't apply here.

bolshakov-a:
Luckily, we don't yet have any author-intent policy for explicit template instantiations.

kimgr:
Quite a few updates:

* Added both explicit instantiation declarations and definitions
* Marked up everywhere IWYU reports
* Also compared to what the compiler requires at each point, and added TODOs where they disagree

I guess if we merge this, we can remove the 958 bug test (that particular bug is fixed) and leave the TODOs as motivation for improving?

kimgr:
Oh, oops. IWYU segfaults somewhere in the middle. I'll look into it.

kimgr:
Something to do with variable templates; this is a minimal repro:
```c++
struct Struct1 {};
template <typename T> T bar = T{};

extern template Struct1 bar<Struct1>;
```

kimgr:
Aha. It only segfaults on higher verbosity levels, and when IWYU calls into `ExplicitInstantiationDecl::print`, so the bug is in there.

tests/cxx/expl_inst_args.cc
```diff
@@ -0,0 +1,100 @@
...
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// This test checks that template arguments are reported from explicit
+// instantiation declarations. For the relationship between explicit
+// instantiations and their backing template, see other expl_inst_* tests.
+
+// IWYU_ARGS: -I . -std=c++20
```
bolshakov-a: Looks like C++20 mode is not required here.

tests/cxx/expl_inst_args.cc
```diff
@@ -0,0 +1,100 @@
...
+// IWYU: Struct2 needs a declaration
+extern template struct ns::S<Struct2>::Inner;                // (g)
+
+// Part 2: add explicit instantiation definitions for these template kinds.
+
+// IWYU: Struct1 needs a declaration
+// IWYU: Struct1 is...*expl_inst_args-i1.h
+template void ns::foo<Struct1>(Struct1);              // (a)
+// TODO: Language wants full dfn of Struct1 here.
+// TODO: // IWYU: Struct1 needs a declaration
```
bolshakov-a: Is `// TODO` here a typo, or IWYU actually doesn't report even the fwd-decl use of `Struct1`?

bolshakov-a (approved):
> I guess if we merge this, we can remove the 958 bug test (that particular bug is fixed) and leave the TODOs as motivation for improving?

Probably, yes, but OTOH, you can make the following changes and still observe the same bug:
```diff
--- a/tests/bugs/958/958.cc
+++ b/tests/bugs/958/958.cc
@@ -13,13 +13,14 @@
 #include "file1.h"
 
 template <typename T>
-int templFunc(T val) {
+int templFunc() {
+  T val;
   return val.x;
 }
 
 // Explicit function template instantiation:
 // https://en.cppreference.com/w/cpp/language/function_template#Explicit_instantiation
-template int templFunc<X>(X val);
+template int templFunc<X>();
```

tests/cxx/expl_inst_args.cc
bolshakov-a: Interesting. It is attributed to the explicit instantiation declaration location. Probably, a deficiency in Clang, in `ExplicitInstantiationDecl` implementation.

kimgr: It's not reported, surprisingly. IWYU has nothing to say about line 71.

tests/cxx/expl_inst_args.cc
kimgr: Removed

bolshakov-a:
> Aha. It only segfaults on higher verbosity levels, and when IWYU calls into ExplicitInstantiationDecl::print, so the bug is in there.

I've opened https://github.com/llvm/llvm-project/issues/197797.

tests/cxx/expl_inst_args.cc
bolshakov-a: I think because the argument is attributed to line 45.

kimgr:
> I've opened https://github.com/llvm/llvm-project/issues/197797

Thanks! I was trying to build latest llvm to dig deeper, but I can no longer link the clang unittests with 40GB of memory :cold_sweat:.

tests/cxx/expl_inst_args.cc
kimgr: Yeah, I wonder if it's miswired in the AST:
```
tests/cxx/expl_inst_args.cc:71:1: (1) [ ExplicitInstantiationDecl ] 0x6336a5a620a8 <todo>                                                                                                     
tests/cxx/expl_inst_args.cc:71:18: (2) [ NestedNameSpecifier ] ns::                                                                                                                           
tests/cxx/expl_inst_args.cc:45:33: (2) [ TemplateArgumentLoc ] 0x7fff2c83f2c0 <Struct1>                                                                                                       
tests/cxx/expl_inst_args.cc:45:33: (3, fwd decl) [ RecordTypeLoc ] Struct1                                                                                                                    
Marked fwd-decl use of decl 0x6336a5a32930 Struct1 (from tests/cxx/expl_inst_args-i1.h:10:8) at tests/cxx/expl_inst_args.cc:45:33 
```
Looks like the traversal goes from the explicit instantiation on line 71 to the TemplateArgumentLoc on line 45. Maybe they're shared between decl and defn, somehow?

kimgr: Oh, you're way ahead of me; I didn't see this comment: https://github.com/include-what-you-use/include-what-you-use/pull/2024#discussion_r3243923870

kimgr: I asked on the PR: https://github.com/llvm/llvm-project/issues/191442#issuecomment-4457732314

bolshakov-a:
> I can no longer link the clang unittests with 40GB of memory

Even with one job at a time? (You can use `-j` option of `ninja` or `make` or specify [`LLVM_PARALLEL_LINK_JOBS` CMake variable](https://www.llvm.org/docs/CMake.html#frequently-used-llvm-related-variables).)

kimgr:
> > I can no longer link the clang unittests with 40GB of memory
>
> Even with one job at a time? (You can use -j option of ninja or make or specify [LLVM_PARALLEL_LINK_JOBS
> CMake variable](https://www.llvm.org/docs/CMake.html#frequently-used-llvm-related-variables).)

Yeah, I have those set routinely. But I'm using `mold`, not `lld`, so maybe it's more aggressive with memory usage. I gave up on GNU `ld` a long time ago.

tests/cxx/expl_inst_args.cc
```diff
@@ -0,0 +1,100 @@
...
+// instantiation declarations. For the relationship between explicit
+// instantiations and their backing template, see other expl_inst_* tests.
+
+// IWYU_ARGS: -I .
+
+#include "tests/cxx/expl_inst_args-d1.h"
+
+// Example covering all relevant entities, stolen from
+// https://github.com/llvm/llvm-project/issues/191442
+namespace ns {
```
bolshakov-a: I wonder if a namespace is actually needed for this test.

kimgr: Yeah, it's probably just a distraction here, I can remove it.

kimgr:
A comment on a merged PR that is more relevant for this one: https://github.com/include-what-you-use/include-what-you-use/pull/2023#issuecomment-4416194914

> > I guess this default `Visit` implementation causes the `TemplateArgumentLocs` to be visited
> 
> To be more correct, the full use of the type is reported not from the traversing of template arguments but of the function prototype in `RecursiveASTVisitor`, as you can see from the log. When scanning an ordinary function declaration, IWYU calls `set_in_forward_declare_context` if the declaration is not a definition:
> https://github.com/include-what-you-use/include-what-you-use/blob/c36eacb7f104dd4349c0ca1dcc0e23433b52c423/iwyu.cc#L1765
> 
> If the function declaration is actually the definition, IWYU reports types of parameters likewise other variable types. But then `FunctionProtoTypeLoc` appears outside of any `FunctionDecl`, `set_in_forward_declare_context` is not called, and reporting parameter types is a bug:
> ```cpp
> #include "tests/cxx/direct.h"
> 
> int main() {
>   // IWYU wrongly reports that `IndirectClass` should be complete.
>   int (*p)(IndirectClass);
> }
> ```
> The same situation is now with `ExplicitInstantiationDecl`. Whereas IWYU correctly reports now parameter (ordinary function parameters, not template parameters) types for explicit instantiation _definitions_, it reports them for explicit instantiation _declarations_ as well. Thus, one bug has been replaced with another bug.

kimgr:
@bolshakov-a 
> > I guess if we merge this, we can remove the 958 bug test (that particular bug is fixed) and leave the TODOs as motivation for improving?
> 
> Probably, yes, but OTOH, you can make the following changes and still observe the same bug:
> 
> ```diff
> --- a/tests/bugs/958/958.cc
> +++ b/tests/bugs/958/958.cc
> @@ -13,13 +13,14 @@
>  #include "file1.h"
>  
>  template <typename T>
> -int templFunc(T val) {
> +int templFunc() {
> +  T val;
>    return val.x;
>  }
>  
>  // Explicit function template instantiation:
>  // https://en.cppreference.com/w/cpp/language/function_template#Explicit_instantiation
> -template int templFunc<X>(X val);
> +template int templFunc<X>();
> ```

I added a `template<class T> void body() { T val; }` example covering this to the new test, with TODOs based on Clang's reporting.

Do you think it makes sense to remove the 958 bug test with that addition?

bolshakov-a:
I don't know, bug tests was your idea. If the newly added TODO contained the bug number, it would act as the bug test. But anyway, I hope we'll fix that bug soon, because there are now distinct AST nodes corresponding to explicit function template instantiations, so this can be done:
https://github.com/include-what-you-use/include-what-you-use/blob/3fdba2fb4cd3fddfe8a76250976d343aaf650b46/iwyu.cc#L5220-L5221

> But then `FunctionProtoTypeLoc` appears outside of any `FunctionDecl`, `set_in_forward_declare_context` is not called, and reporting parameter types is a bug

Btw, it is bug #1256.

tests/cxx/expl_inst_args.cc
kimgr: The fix has been merged, but not yet made its way to the Ubuntu packages, so I'll leave this as-is until apt-llvm catches up and our build breaks.

PR #2027, opened 2026-05-16
Add intermediate test header

bolshakov-a:
This causes emitting explicit warnings in the partial specialization explicit instantiation test case thus making the test more informative.

`-i1` is appended to the file name because I'm going to add `-i2` header.

No functional change.


kimgr:
Makes sense, thanks!

Issue #988, opened 2022-01-03
[Feature request] Add "forward" mapping


geimer:
The currently available `symbol` mapping already allows to map symbol _definitions_ to header files.  However, in various use cases (see e.g., #530) it is desirable to use an include file for _forward declarations_ rather than explicitly repeating them all over the place.  Unless I have missed something, this doesn't seem to be possible with the `symbol` mapping.  Thus, I propose to implement a `forward` mapping type to also allow such kinds of mappings.

An alternative way of implementing this feature would be to introduce a `forward` visibility type for the `symbol` mapping, though I can't really judge which option makes more sense.

kimgr:
Yep, something like that would be nice.

I like the idea of a `forward` visibility, but my mind is too cloudy to play out exactly how it would work :-)

I'm mostly typing this up for myself, so I can reason about mappings, I'm not entirely sure this is how it works in practice...

For headers:
* Private-to-private mapping says: "From-header includes to-header, but they are both private"
* Private-to-public mapping says: "From-header is private, to-header is public."
* Public-to-public mapping says: "From-header includes to-header, and they are both public"
* Public-to-private mapping should never exist

The outcome of that is that given a suggested header A;
* if A is public and present, leave it
* if A is public and not present, is there a set of public-to-public mapping from A?
    * for every to-header X mapped from A, is X present? If so, leave X.
    * if not, suggest A
* if A is private, whether present or not
    * if there is a private-to-public mapping from A to X, suggest X
    * if not, is there a private-to-private mapping from A to Y? If yes, recurse with Y to see if we can find a public header.
    * if no mapping available, suggest A as-is

For symbols, they are just folded into the mapping visibility structure for convenience. The from-visibility is ignored.
* to-private mapping says: "Symbol is defined in to-header, but it is private"
* to-public mapping says: "Symbol is defined in to-header and it is public."

In conjunction with header mappings this can resolve quite nicely. If a symbol is mapped to a private header and private header mapping eventually gives a public header, that strikes a nice balance between ease of generating automated mappings, e.g. `grep` for symbols and map them to their containing file, then `grep` for `#include` directives and generate header mappings. In practice it's not really that simple, of course.

So I wonder if it would be possible to smuggle a `forward-only` visibility into the unused from-visibility field of the symbol mapping... That would still allow to-private and to-public mappings for symbols, but resolution would have to start by deciding whether the symbol we're looking for is for a forward-declare use or a full use. Something like:

[...]

PR #2028, opened 2026-05-17
Add forward-declaration mapping

bolshakov-a:
This enables specifying headers in mapping files that should be used instead of forward-declarations of the corresponding tag types. The kind of a symbol mapping entry should be set in the second mapping data field and can be `full` or `fwd-decl`. `private` is also acceptable and is handled as `full` so that old symbol mappings are treated as full-use mappings.

If a source file under analysis contains a forward-declaration already, IWYU suggests replacing it with the mapped header, similar to the behavior with forward-declarations exported with pragmas.

This implements #988.

### DIFF
--- a/docs/IWYUMappings.md
+++ b/docs/IWYUMappings.md
@@ -115,17 +115,30 @@ header.
 Data for this directive is a list of four strings containing:
 
 * The symbol name to map from
-* The visibility of the symbol
+* The use kind of the symbol (`full` or `fwd-decl`, where `full` corresponds
+  to uses that require the complete type info for tag types)
 * The include name to map to
 * The visibility of the include name to map to
 
 For example;
 
-    { "symbol": ["NULL", "private", "<cstddef>", "public"] }
+    { "symbol": ["NULL", "full", "<cstddef>", "public"] }
 
-The symbol visibility is largely redundant -- it must always be `private`. It
-isn't entirely clear why symbol visibility needs to be specified, and it might
-be removed moving forward.
+As another example, consider class `X` defined in `"bits/x.h"`. Here is what
+IWYU would suggest for different cases across mappings and uses:
+
+|                     |       No mapping      | `["X", "full", "<x.h>", ...` | `["X", "fwd-decl", "<xfwd.h>", ...` |
+| ------------------- | --------------------- | ---------------------------- | ----------------------------------- |
+| Full use of `X`     | `#include "bits/x.h"` | `#include <x.h>`             | `#include "bits/x.h"`               |
+| Fwd-decl use of `X` | `class X;`            | `class X;`                   | `#include <xfwd.h>`                 |
+
+If you want the mapping both for complete type info and fwd-decl uses, specify
+the both entries in the mapping file.
+
+If `fwd-decl` mapping is specified, IWYU suggests to replace already present
+forward-declarations with the corresponding headers. For backward compatibility
+with the old mapping file format, `private` is acceptable and is handled as
+`full` use kind.
 
 Unlike `include`, `symbol` directives do not support the `@`-prefixed regex
 syntax in the first entry. Track the [following
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -16,6 +16,7 @@
 #include <map>                          // for map, map<>::mapped_type, etc
 #include <memory>
 #include <numeric>                      // for accumulate
+#include <optional>
 #include <string>                       // for string, basic_string, etc
 #include <system_error>                 // for error_code
 #include <utility>                      // for pair, make_pair
@@ -62,6 +63,7 @@ using llvm::yaml::document_iterator;
 using std::find;
 using std::make_pair;
 using std::map;
+using std::optional;
 using std::pair;
 using std::string;
 using std::unique_ptr;
@@ -1576,6 +1578,14 @@ void WriteMappings(StringRef directive,
   out << "]\n";
 }
 
+static optional<UseKind> ParseUseKind(StringRef use_kind) {
+  if (use_kind == "full" || use_kind == "private")
+    return UseKind::Full;
+  if (use_kind == "fwd-decl")
+    return UseKind::FwdDecl;
+  return {};
+}
+
 }  // anonymous namespace
 
 // Write out all internal mappings to files in dirpath.
@@ -1643,7 +1653,8 @@ void IncludePicker::AddInternalMappings(CStdLib cstdlib, CXXStdLib cxxstdlib) {
 
       // the canonical header is returned first
       for (const Header& header : sym.headers()) {
-        AddSymbolMapping(name, MappedInclude(header.name().str()), kPublic);
+        AddSymbolMapping(name, UseKind::Full,
+                         MappedInclude(header.name().str()), kPublic);
       }
     }
   }
@@ -1660,7 +1671,8 @@ void IncludePicker::AddInternalMappings(CStdLib cstdlib, CXXStdLib cxxstdlib) {
 
       // the canonical header is returned first
       for (const Header& header : sym.headers()) {
-        AddSymbolMapping(name, MappedInclude(header.name().str()), kPublic);
+        AddSymbolMapping(name, UseKind::Full,
+                         MappedInclude(header.name().str()), kPublic);
       }
     }
   }
@@ -1773,9 +1785,14 @@ void IncludePicker::AddIncludeMapping(const string& map_from,
 }
 
 void IncludePicker::AddSymbolMapping(const string& map_from,
+                                     UseKind use_kind,
                                      const MappedInclude& map_to,
                                      IncludeVisibility to_visibility) {
-  symbol_include_map_[map_from].push_back(map_to);
+  if (use_kind == UseKind::Full) {
+    symbol_include_map_[map_from].push_back(map_to);
+  } else {
+    fwd_decl_symbol_map_[map_from].push_back(map_to);
+  }
 
   MarkVisibility(&include_visibility_map_, map_to.quoted_include,
                  to_visibility);
@@ -1794,7 +1811,8 @@ void IncludePicker::AddSymbolMappings(const IncludeMapEntry* entries,
                                       size_t count) {
   for (size_t i = 0; i < count; ++i) {
     const IncludeMapEntry& e = entries[i];
-    AddSymbolMapping(e.map_from, MappedInclude(e.map_to), e.to_visibility);
+    AddSymbolMapping(e.map_from, UseKind::Full, MappedInclude(e.map_to),
+                     e.to_visibility);
   }
 }
 
@@ -1905,10 +1923,11 @@ void IncludePicker::FinalizeAddedIncludes() {
   // If a.h maps to b.h maps to c.h, we'd like an entry from a.h to c.h too.
   MakeMapTransitive(&filepath_include_map_);
   // Now that filepath_include_map_ is transitively closed, it's an
-  // easy task to get the values of symbol_include_map_ closed too.
-  for (IncludeMap::value_type& symbol_include : symbol_include_map_) {
+  // easy task to get the values of symbol maps closed too.
+  for (IncludeMap::value_type& symbol_include : symbol_include_map_)
     ExpandOnce(filepath_include_map_, &symbol_include.second);
-  }
+  for (IncludeMap::value_type& symbol_include : fwd_decl_symbol_map_)
+    ExpandOnce(filepath_include_map_, &symbol_include.second);
 
   has_called_finalize_added_include_lines_ = true;
 
@@ -1916,6 +1935,7 @@ void IncludePicker::FinalizeAddedIncludes() {
   if (ShouldPrint(9)) {
     PrintMappings(filepath_include_map_, "filepath_include_map_");
     PrintMappings(symbol_include_map_, "symbol_include_map_");
+    PrintMappings(fwd_decl_symbol_map_, "fwd_decl_symbol_map_");
   }
 }
 
@@ -1986,6 +2006,13 @@ vector<MappedInclude> IncludePicker::GetCandidateHeadersForSymbol(
     const string& symbol) const {
   CHECK_(has_called_finalize_added_include_lines_ && "Must finalize includes");
   return GetPublicValues(symbol_include_map_, symbol);
+}
+
+vector<string> IncludePicker::GetCandidateHeadersForSymbolFwdDecl(
+    const string& symbol, const string& including_filepath) const {
+  CHECK_(has_called_finalize_added_include_lines_ && "Must finalize includes");
+  return BestQuotedIncludesForIncluder(
+      GetPublicValues(fwd_decl_symbol_map_, symbol), including_filepath);
 }
 
 vector<string> IncludePicker::GetCandidateHeadersForSymbolUsedFrom(
@@ -2196,7 +2223,15 @@ void IncludePicker::AddMappingsFromFile(const string& filename,
           return;
         }
 
-        AddSymbolMapping(mapping[0], MappedInclude(mapping[2]), to_visibility);
+        optional<UseKind> use_kind = ParseUseKind(mapping[1]);
+        if (!use_kind) {
+          json_stream.printError(
+              current_node, "Unknown symbol use kind '" + mapping[1] + "'.");
+          return;
+        }
+
+        AddSymbolMapping(mapping[0], *use_kind, MappedInclude(mapping[2]),
+                         to_visibility);
       } else if (directive == "include") {
         // Include mapping.
         vector<string> mapping = GetSequenceValue(mapping_item_node.getValue());
--- a/iwyu_include_picker.h
+++ b/iwyu_include_picker.h
@@ -69,6 +69,8 @@ using std::vector;
 enum class RegexDialect;
 struct IncludeMapEntry;
 
+enum class UseKind { Full, FwdDecl };
+
 enum IncludeVisibility { kUnusedVisibility, kPublic, kPrivate };
 enum class CStdLib { None, ClangSymbols, Glibc };
 enum class CXXStdLib { None, ClangSymbols, Libstdcxx, Libcxx };
@@ -148,6 +150,8 @@ class IncludePicker {
   // most standard) header for the symbol.
   vector<MappedInclude> GetCandidateHeadersForSymbol(
       const string& symbol) const;
+  vector<string> GetCandidateHeadersForSymbolFwdDecl(
+      const string& symbol, const string& including_filepath) const;
 
   // As above, but given a specific including header it is possible to convert
   // mapped includes to quoted include strings (because we can for example know
@@ -219,9 +223,10 @@ class IncludePicker {
   // Adds a mapping from a a symbol to a quoted include. We use this to
   // maintain mappings of documented types, e.g.
   //  For std::map<>, include <map>.
-  void AddSymbolMapping(
-      const string& map_from, const MappedInclude& map_to,
-      IncludeVisibility to_visibility);
+  void AddSymbolMapping(const string& map_from,
+                        UseKind use_kind,
+                        const MappedInclude& map_to,
+                        IncludeVisibility to_visibility);
 
   // Adds mappings from sized arrays of IncludeMapEntry.
   void AddIncludeMappings(const IncludeMapEntry* entries, size_t count);
@@ -265,8 +270,10 @@ class IncludePicker {
   vector<string> BestQuotedIncludesForIncluder(
       const vector<MappedInclude>&, const string& including_filepath) const;
 
-  // From symbols to includes.
+  // From symbols to includes for full symbol uses.
   IncludeMap symbol_include_map_;
+  // From symbols to includes for uses that require only forward-declarations.
+  IncludeMap fwd_decl_symbol_map_;
 
   // From quoted filepath patterns to includes, where a pattern can be
   // either a quoted filepath (e.g. "foo/bar.h" or <a/b.h>) or @
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -278,9 +278,12 @@ string GetShortNameAsString(const NamedDecl* named_decl) {
 }  // namespace internal
 
 // Holds information about a single full or fwd-decl use of a symbol.
-OneUse::OneUse(const NamedDecl* decl, SourceLocation use_loc,
-               SourceLocation decl_loc, OneUse::UseKind use_kind,
-               UseFlags flags, const char* comment)
+OneUse::OneUse(const NamedDecl* decl,
+               SourceLocation use_loc,
+               SourceLocation decl_loc,
+               UseKind use_kind,
+               UseFlags flags,
+               const char* comment)
     : symbol_name_(internal::GetQualifiedNameAsString(decl)),
       short_symbol_name_(internal::GetShortNameAsString(decl)),
       decl_(decl),
@@ -296,7 +299,8 @@ OneUse::OneUse(const NamedDecl* decl, SourceLocation use_loc,
 }
 
 // This constructor always creates a full use.
-OneUse::OneUse(const string& symbol_name, OptionalFileEntryRef dfn_file,
+OneUse::OneUse(const string& symbol_name,
+               OptionalFileEntryRef dfn_file,
                SourceLocation use_loc)
     : symbol_name_(symbol_name),
       short_symbol_name_(symbol_name),
@@ -304,7 +308,7 @@ OneUse::OneUse(const string& symbol_name, OptionalFileEntryRef dfn_file,
       decl_file_(dfn_file),
       decl_filepath_(GetFilePath(dfn_file)),
       use_loc_(use_loc),
-      use_kind_(kFullUse),
+      use_kind_(UseKind::Full),
       use_flags_(UF_None),
       ignore_use_(false),
       is_iwyu_violation_(false) {
@@ -324,7 +328,7 @@ OneUse::OneUse(OptionalFileEntryRef included_file,
       decl_file_(included_file),
       decl_filepath_(GetFilePath(included_file)),
       use_loc_(include_loc),
-      use_kind_(kFullUse),
+      use_kind_(UseKind::Full),
       use_flags_(UF_None),
       ignore_use_(false),
       is_iwyu_violation_(false) {
@@ -333,12 +337,24 @@ OneUse::OneUse(OptionalFileEntryRef included_file,
   suggested_header_ = quoted_include;
 }
 
+bool OneUse::is_full_use() const {
+  return use_kind_ == UseKind::Full;
+}
+
 void OneUse::reset_decl(const NamedDecl* decl) {
   CHECK_(decl_ && "Need existing decl to reset it");
   CHECK_(decl && "Need to reset decl with existing decl");
   decl_ = decl;
   decl_file_ = GetFileEntry(decl);
   decl_filepath_ = GetFilePath(decl);
+}
+
+void OneUse::set_full_use() {
+  use_kind_ = UseKind::Full;
+}
+
+void OneUse::set_forward_declare_use() {
+  use_kind_ = UseKind::FwdDecl;
 }
 
 int OneUse::UseLinenum() const {
@@ -353,6 +369,13 @@ void OneUse::SetPublicHeaders() {
   // We should never need to deal with public headers if we already know
   // who we map to.
   CHECK_(suggested_header_.empty() && "Should not need a public header here");
+
+  if (use_kind_ == UseKind::FwdDecl) {
+    public_headers_ = GlobalIncludePicker().GetCandidateHeadersForSymbolFwdDecl(
+        symbol_name_, GetFilePath(use_loc_));
+    return;
+  }
+
   if (decl_) {
     public_headers_ = GlobalIncludePicker().GetMappedPublicHeaders(
         decl_, GetFilePath(use_loc_), decl_filepath());
@@ -367,7 +390,8 @@ void OneUse::SetPublicHeaders() {
 const vector<string>& OneUse::public_headers() {
   if (public_headers_.empty()) {
     SetPublicHeaders();
-    CHECK_(!public_headers_.empty() && "Should always have at least one hdr");
+    CHECK_(use_kind_ == UseKind::FwdDecl || !public_headers_.empty())
+        << "Full uses should always have at least one hdr";
   }
   return public_headers_;
 }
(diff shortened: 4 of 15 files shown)


iwyu_output.cc
```diff
@@ -882,14 +892,15 @@ set<string> CalculateMinimalIncludes(
...
-    if (use.ignore_use() || !use.is_full_use())
+    if (use.ignore_use())
       continue;
     // Special case #1: Some uses come with a suggested header already picked.
     if (use.has_suggested_header()) {
       desired_headers.insert(use.suggested_header());
       continue;
     }
+    if (!use.NeedsSuggestedHeader())
+      continue;
```
bolshakov-a, line 923: In fact, this check is redundant because, for fwd-decl uses, `NeedsSuggestedHeader` currently returns `false` if there are no public headers, but in this case the following conditional branches aren't executed anyway.

bolshakov-a:
@RT2Code, could you check if this implementation fits your needs?

iwyu_include_picker.cc
```diff
@@ -80,6 +82,8 @@ struct IncludeMapEntry {      // A POD so we can make the input static
   IncludeVisibility to_visibility;
 };
 
+enum class SymbolUseKind { Unused, Full, FwdDecl };
```
kimgr: I feel like we could probably use the same enum here and in `OneUse::UseKind`?

It creates some annoying coupling between `iwyu_output` and `iwyu_include_picker`, but I think we already have that dependency?

Also, instead of `Unused`, consider returning a `std::optional<SymbolUseKind>` from the parse function.

iwyu_include_picker.cc
```diff
@@ -1773,9 +1787,12 @@ void IncludePicker::AddIncludeMapping(const string& map_from,
 }
 
 void IncludePicker::AddSymbolMapping(const string& map_from,
+                                     SymbolUseKind use_kind,
                                      const MappedInclude& map_to,
                                      IncludeVisibility to_visibility) {
-  symbol_include_map_[map_from].push_back(map_to);
+  (use_kind == SymbolUseKind::Full ? symbol_include_map_
```
kimgr: Heh, I think we could probably live with a temporary here for readability.

iwyu_include_picker.cc
kimgr: Or maybe there's a good reason to keep `Unused` at the zero value? Like default-initialized values?

iwyu_output.cc
```diff
@@ -377,8 +385,10 @@ bool OneUse::PublicHeadersContain(const string& elt) {
   return ContainsValue(public_headers(), elt);
 }
 
-bool OneUse::NeedsSuggestedHeader() const {
-  return (!ignore_use() && is_full_use() && suggested_header_.empty());;
+bool OneUse::NeedsSuggestedHeader() {
```
kimgr, line 404: Can't this be `const` still?

kimgr: Ah, `public_headers()`. I'll push a little refactor that makes that const with a mutable member instead.

iwyu_output.cc
```diff
@@ -377,8 +385,10 @@ bool OneUse::PublicHeadersContain(const string& elt) {
...
+  if (ignore_use() || !suggested_header_.empty())
+    return false;
+  return is_full_use() ? true : !public_headers().empty();
```
kimgr: I'd prefer this without the ternary conditional, I think just if-guards would be clearer:
```
if (is_full_use())
  return true;
// Something about why public_headers might be empty for fwd-decls...
return !public_headers.empty();
```

But also, looks like `public_headers` can never be empty?

iwyu_include_picker.cc
bolshakov-a: I think, `optional` is a good idea. It is empty by default.

iwyu_output.cc
bolshakov-a: They are empty for fwd-decl uses that have no mapping.

iwyu_include_picker.cc
kimgr: I was thinking more for maps or arrays containing `SymbolUseKind`; if they contain default-initialized values (e.g. by `SymbolUseKind k[16] = {};`), it might be nice if it defaulted to a sentinel rather than `Full`.

iwyu_output.cc
kimgr: The assertions disagree:
```
const vector<string>& OneUse::public_headers() {
  if (public_headers_.empty()) {
    SetPublicHeaders();
    CHECK_(!public_headers_.empty() && "Should always have at least one hdr");
  }
  return public_headers_;
}
```

I don't see a path through that function where `public_headers_` is returned empty.

iwyu_output.cc
kimgr: https://github.com/include-what-you-use/include-what-you-use/pull/2029

iwyu_output.cc
bolshakov-a: But I've changed that assertion:
https://github.com/include-what-you-use/include-what-you-use/pull/2028/changes#diff-96c9906a158b909fda0f001cb5bd04615d46704df24173306894f3d228abbed1R377

kimgr: Oh, 🤦. Completely missed that.

iwyu_include_picker.cc
kimgr: or even:
```
if (use_kind == SymbolUseKind::Full) {
  symbol_include_map_[map_from].push_back(map_to);
} else {
  fwd_decl_symbol_map_[map_from].push_back(map_to);
}
```

iwyu_output.cc
kimgr: Nevermind. Keeping `NeedsSuggestedHeader` const seems more trouble than it's worth.

iwyu_include_picker.cc
kimgr: But yeah, seems better for the enum to be closed without sentinel if we can use `optional` for parsing.

bolshakov-a: Currently, there are no such arrays. `libc_symbol_map` and `stdlib_cxx_symbol_map` will probably contain `SymbolUseKind`. But they are constructed already initialized.

iwyu_ast_util.cc
```diff
@@ -1238,21 +1238,10 @@ const NamedDecl* GetInstantiatedFromDecl(const NamedDecl* decl) {
...
-      // decl is instantiated from a non-specialized template.
-      return tpl_decl;
-    } else if (const auto* partial_spec_decl =
-                   instantiated_from
-                       .dyn_cast<VarTemplatePartialSpecializationDecl*>()) {
-      // decl is instantiated from a template partial specialization.
-      return partial_spec_decl;
-    }
+    if (const auto* pattern = tpl_sp_decl->getTemplateInstantiationPattern())
+      return pattern;
```
bolshakov-a, line 1241: Some unrelated work leaked here, sorry.

iwyu_output.h
```diff
@@ -118,16 +117,20 @@ class OneUse {
...
 
   void reset_decl(const clang::NamedDecl* decl);
-  void set_full_use() { use_kind_ = kFullUse; }
-  void set_forward_declare_use() { use_kind_ = kForwardDeclareUse; }
+  void set_full_use() {
+    use_kind_ = SymbolUseKind::Full;
+  }
+  void set_forward_declare_use() {
+    use_kind_ = SymbolUseKind::FwdDecl;
+  }
```
bolshakov-a: Implementations of these functions can be moved to `.cpp` so as to allow an opaque-declaration of `SymbolUseKind` here instead of `#include "iwyu_include_picker.h"`.

kimgr: Sounds good.

iwyu_include_picker.h
```diff
@@ -219,9 +223,10 @@ class IncludePicker {
   // Adds a mapping from a a symbol to a quoted include. We use this to
   // maintain mappings of documented types, e.g.
   //  For std::map<>, include <map>.
-  void AddSymbolMapping(
-      const string& map_from, const MappedInclude& map_to,
-      IncludeVisibility to_visibility);
+  void AddSymbolMapping(const string& map_from,
+                        SymbolUseKind,
```
kimgr: Let's add a parameter name, even if it's pretty self-explanatory.

iwyu_include_picker.cc
```diff
@@ -1576,6 +1580,14 @@ void WriteMappings(StringRef directive,
   out << "]\n";
 }
 
+static optional<SymbolUseKind> ParseSymbolUseKind(string_view use_kind) {
```
kimgr: I'd prefer LLVM's `StringRef` over `string_view` still.

iwyu_include_picker.h
```diff
@@ -69,6 +69,8 @@ using std::vector;
 enum class RegexDialect;
 struct IncludeMapEntry;
 
+enum class SymbolUseKind { Full, FwdDecl };
```
kimgr: I wonder if we can now strip `Symbol` from this, and just call it `UseKind`?

iwyu_output.cc
kimgr: I think it looks quite clear to have it spelled out here even if it's redundant, and a bit more robust to change.

iwyu_output.h
```diff
@@ -46,17 +46,17 @@ using std::vector;
...
 class OneUse {
  public:
-  enum UseKind { kFullUse, kForwardDeclareUse };
-
   // This constructor is used to track nominal decl uses.
   OneUse(const clang::NamedDecl* decl,
          clang::SourceLocation use_loc,
          clang::SourceLocation decl_loc,
-         UseKind use_kind,
+         SymbolUseKind,
```
kimgr: I'd like to see a parameter name here too

tests/cxx/fwd_decl_mapping.cc
```diff
@@ -0,0 +1,64 @@
...
+
+#include "tests/cxx/fwd_decl_mapping-d1.h"
+
+// IWYU: OnlyFwdDeclUse needs a declaration
+OnlyFwdDeclUse* pOnlyFwdDeclUse;
+
+// IWYU: EnumClass needs a declaration
+EnumClass e;
+
+// The forward-declaration header is not needed because the definition
```
kimgr: Is there a reason to do the full and fwd-decl use on the same line? I feel like it would be a little clearer if they were two separate declarations with their respective assertions.

tests/cxx/fwd_decl_mapping.cc
```diff
@@ -0,0 +1,64 @@
...
+// should be available.
+// IWYU: FwdDeclAndFullUse needs a declaration
+// IWYU: FwdDeclAndFullUse is...*-i4.h
+FwdDeclAndFullUse fwdDeclAndFullUse, *pFwdDeclAndFullUse;
+
+// '-i5.h' should not be suggested because '-i6.h' should provide the both.
```
kimgr: typo: the both -> them both?

kimgr: OK, and `-i6.h` will be desired because of `FwdDeclInI6*` below?

tests/cxx/fwd_decl_mapping.cc
```diff
@@ -0,0 +1,64 @@
...
+// IWYU: HasIncludeMapping needs a declaration
+HasIncludeMapping* him;
+
+// IWYU: HasMappingForFullUse needs a declaration
+HasMappingForFullUse* hmffu;
+
+// Test that IWYU suggests the canonical (first mapped) header.
+// IWYU: FwdDeclInTwoHeaders2 needs a declaration
+FwdDeclInTwoHeaders2* fdth2;
+
```
kimgr, line 42: In general I found it a bit confusing that there are no diagnostics for which header, but they just magically appear in the summary. But that's a consequence of forward-declares just getting a "needs a declaration" diagnostic, and then mappings silently updating that to a mapped `#include`, right?

tests/cxx/fwd_decl_mapping.cc
bolshakov-a: Right, it is exactly to make `-i6.h` desired.

tests/cxx/fwd_decl_mapping.cc
bolshakov-a: Something like that. And it would be confusing to have "is **defined** in" comment for forward-declaration headers.

tests/cxx/fwd_decl_mapping.cc
bolshakov-a: > typo: the both -> them both?

Rather a lack of English grammar knowledge.

RT2Code:
> @RT2Code, could you check if this implementation fits your needs?

It does, thanks for the hard work. However, I find the naming a bit confusing:

> The symbol use kind can be `full` or `fwd_decl`. `full` corresponds to uses that
> require complete type info for tag types. If `fwd_decl` mapping is specified,
> IWYU suggests to include the corresponding header instead of a forward-
> declaration. For backward compatibility with the old mapping file format,
> `private` is acceptable and is handled as `full` use kind.

If I understand correctly, full ends up allowing forward declarations to be suggested, while fwd_decl does the opposite and forces including the corresponding header instead. Intuitively, I would expect the opposite behavior from those names.

bolshakov-a:
Makes sense. Maybe, `"definition"` would be better than `"full"`? CC @kimgr

OTOH, it would be confusing for non-tag kinds of symbols, e.g. functions.

bolshakov-a:
I've realized that having two symbols mapped to one header in the test was probably not good. Particularly, the test passed if `EnumClass` mapping kind was changed to `full`.

kimgr:
> > @RT2Code, could you check if this implementation fits your needs?
> 
> It does, thanks for the hard work. However, I find the naming a bit confusing:
> 
> > The symbol use kind can be `full` or `fwd_decl`. `full` corresponds to uses that
> > require complete type info for tag types. If `fwd_decl` mapping is specified,
> > IWYU suggests to include the corresponding header instead of a forward-
> > declaration. For backward compatibility with the old mapping file format,
> > `private` is acceptable and is handled as `full` use kind.
> 
> If I understand correctly, full ends up allowing forward declarations to be suggested, while fwd_decl does the opposite and forces including the corresponding header instead. Intuitively, I would expect the opposite behavior from those names.

Hmm, yeah, it is a little unfortunate that `full` allows forward-declarations while `fwd_decl` forces includes.

But I wonder if the confusion stems partly from the doc phrasing... As I understand it, the use kind is not about what IWYU will do, but where the mapping applies. Maybe an example would be better. Assuming `X` is declared in `"bits/x.h"`, here's how the mappings should play out:

|  | No mapping | `["X", "full", "<x.h>", "public"]` | `["X", "fwd_decl", "<xfwd.h>", "public"]` |
|--|--|--|--|
| Full use of `X` | `#include "bits/x.h"` | `#include <x.h>` | `#include "bits/x.h"` | 
| Fwd-decl use of `X` | `class X;` | `class X;` | `#include <xfwd.h>` |

bolshakov-a:
@kimgr, are you suggesting to add this table in the docs?

kimgr:
Maybe, if you think it's helpful.

bolshakov-a:
Maybe, `"complete"` would be better than `"full"` because it better corresponds to the standard wording? Another concern is that `"fwd-decl"` could be more ergonomic because you don't have to press Shift when typing.

kimgr:
I think the standard talks about "complete" and "incomplete" **types**, whereas in IWYU we talk about "full" or "fwd-decl" **uses**. It might be interesting to try and shape IWYU's wording more towards the standard language, but that feels like a v2 feature, when we can afford to break mapping backward compatibility anyway :-).

Btw, please rebase, I finally fixed up the 958 testcase based on your example.

bolshakov-a:
At first glance, I've read "I finally fixed up the 958 testcase**s**". Don't frighten me.

docs/IWYUMappings.md
```diff
@@ -115,17 +115,31 @@ header.
...
-isn't entirely clear why symbol visibility needs to be specified, and it might
-be removed moving forward.
+The symbol use kind can be `full` or `fwd_decl`. `full` corresponds to uses that
+require complete type info for tag types. As another example, consider class `X`
+defined in `"bits/x.h"`. Here is what IWYU would suggest for different mapping
+and use cases:
+
+|                     |       No mapping      | `["X", "full", "<x.h>", "public"]` | `["X", "fwd_decl", "<xfwd.h>", "public"]` |
+| ------------------- | --------------------- | ---------------------------------- | ----------------------------------------- |
+| Complete use of `X` | `#include "bits/x.h"` | `#include <x.h>`                   | `#include "bits/x.h"`                     |
```
kimgr: Complete use -> Full use

docs/IWYUMappings.md
```diff
@@ -115,17 +115,31 @@ header.
...
 
-The symbol visibility is largely redundant -- it must always be `private`. It
-isn't entirely clear why symbol visibility needs to be specified, and it might
-be removed moving forward.
+The symbol use kind can be `full` or `fwd_decl`. `full` corresponds to uses that
+require complete type info for tag types. As another example, consider class `X`
+defined in `"bits/x.h"`. Here is what IWYU would suggest for different mapping
+and use cases:
+
+|                     |       No mapping      | `["X", "full", "<x.h>", "public"]` | `["X", "fwd_decl", "<xfwd.h>", "public"]` |
```
kimgr: We can make the table a little bit narrower if we skip the `, "public"]` part of the mapping, something like;
```
| `["X", "full", "<x.h>", ...` | `["X", "fwd_decl", "<xfwd.h>", ...` |

File: docs/IWYUMappings.md [hunk 17]
```diff
@@ -115,17 +115,31 @@ header.
...
-    { "symbol": ["NULL", "private", "<cstddef>", "public"] }
+    { "symbol": ["NULL", "full", "<cstddef>", "public"] }
 
-The symbol visibility is largely redundant -- it must always be `private`. It
-isn't entirely clear why symbol visibility needs to be specified, and it might
-be removed moving forward.
+The symbol use kind can be `full` or `fwd_decl`. `full` corresponds to uses that
+require complete type info for tag types. As another example, consider class `X`
+defined in `"bits/x.h"`. Here is what IWYU would suggest for different mapping
+and use cases:
```
kimgr: > Here is what IWYU would suggest for different mapping and use cases:

different mapping**s**

bolshakov-a: I meant rather "different mapping cases and use cases", but I'm not sure what wording is more correct.

docs/IWYUMappings.md
```diff
@@ -115,17 +115,31 @@ header.
...
 For example;
 
-    { "symbol": ["NULL", "private", "<cstddef>", "public"] }
+    { "symbol": ["NULL", "full", "<cstddef>", "public"] }
 
-The symbol visibility is largely redundant -- it must always be `private`. It
-isn't entirely clear why symbol visibility needs to be specified, and it might
-be removed moving forward.
+The symbol use kind can be `full` or `fwd_decl`. `full` corresponds to uses that
+require complete type info for tag types. As another example, consider class `X`
```
kimgr: > As another example, ...

I wasn't sure what the first example was, but I see now the block before is preceded by `For example;`.

The way you explain symbol use kind in between made me forget. I wonder if you could move the `full` and `fwd_decl` info up into the bullet list, something like:

> * The use kind of the symbol (`full` or `fwd_decl`, where `full corresponds to uses that require the complete type for tag types)

(suitably line wrapped)

tests/cxx/fwd_decl_mapping_path_local.cc
```diff
@@ -0,0 +1,36 @@
...
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -Xiwyu --mapping_file=tests/cxx/fwd_decl_mapping_path_local.imp
+
+// Tests forward-declaration symbol mapping in the case of relative quoted
+// includes. '-I .' command line argument is not specified intentionally.
```
kimgr: not specified -> left out

kimgr:
> Another concern is that `"fwd-decl"` could be more ergonomic because you don't have to press Shift when typing.

I like kebab-case a bit more than snake_case, mostly because I think it looks a little more elegant. So I'd be happy to see `fwd-decl` over `fwd_decl`.

docs/IWYUMappings.md
kimgr: Hmm, I see. Maybe:
> Here is what IWYU would suggest for different cases across mappings and uses:

?

kimgr:
This looks good to me! Do you want to take it out of RFC/draft status, or are there any unknowns to worry about?

bolshakov-a:
I just waited if anyone else wanted to express an opinion about the proposed behavior. Perhaps, @firewave has to say something?

kimgr:
I'm just excited to see the followup where you map all the `std::xx` symbols to `<xxfwd>` :slightly_smiling_face:

bolshakov-a:
In fact, in many cases, IWYU already works well with I/O streams. Currently, problems appear in some extreme cases where all the template arguments are specified explicitly, like this:
```cpp
#include <iosfwd>
#include <ostream>

// IWYU suggests to keep <ostream> instead of <iosfwd> with libstdc++-12.
void Fn(std::basic_ostream<char, std::char_traits<char>>&);
```
So I don't expect any big impact of explicit `<iosfwd>` mappings.

firewave:
> I just waited if anyone else wanted to express an opinion about the proposed behavior. Perhaps, @firewave has to say something?

I am not the person to comment about mapping files, but thanks for the shout out. Maybe if I ever get around to looking at the Qt mapping generator again I might have something to add.

> I'm just excited to see the followup where you map all the `std::xx` symbols to `<xxfwd>` 🙂

That's the only thing on my mind regarding this change.

> In fact, in many cases, IWYU already works well with I/O streams.

The actual mappings are still rather blunt than fine - see #1540 and https://github.com/cppcheck-opensource/cppcheck/pull/6684. Although the later is quite stale, the current state should possibly still reflect that if the mappings in question are dropped.

bolshakov-a:
Yeah, [the mappings from `stdlib_cpp_include_map`](https://github.com/include-what-you-use/include-what-you-use/blob/b37561b32d43709d7644323993c6c5718b7ab00f/iwyu_include_picker.cc#L778) are quite misleading and should be cleaned up. Among them, the standard guarantees some transitive `#include`s only for [`<ios>`](https://timsong-cpp.github.io/cppwp/n4950/ios.syn) and [`<iostream>`](https://timsong-cpp.github.io/cppwp/n4950/iostream.syn).

bolshakov-a:
**Typedef mappings**

Currently, an explicitly spelled typedef is always a full use. When it is used in a non-forward-declarable context, its underlying type is reported or not depending on the IWYU heuristic called [author-intent policy](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhatIsAUse.md#automatic-re-export-typedefs).

I've tried to map the remaining symbols from [input.output] clause of the standard: https://github.com/bolshakov-a/include-what-you-use/commit/511de8087d0cba86f62b1fc3002cd3784c1678ba. Typedefs like `std::ostream` are of course mapped as `full` both to `<iosfwd>` and to `<ostream>`. But there is a problem: if some standard library implementation defines a typedef redeclaration _after_ the underlying template, IWYU considers it as "providing" and suggests to include `<iosfwd>` even where the forward-declaration is not enough. I was hoping that implementers don't do that because it is kinda nonsense to define a typedef in `<iosfwd>` and then redefine it in a header including `<iosfwd>`. But it turns out that libstdc++ do exactly that for `std::spanstream` at least.

I fixed the problem by hardcoding such typedefs as nonproviding: https://github.com/bolshakov-a/include-what-you-use/commit/4b3e813568d19a51b651faf648ce912b1ae188c8. But this, besides requiring to remember and manually update the list if some typedefs are added in some future standard revisions, cannot be done by users if some other libraries need similar handling. Another approach could be to consider only the first typedef redeclaration in a translation unit, but this could break other cases where IWYU would otherwise suggest a really providing typedef.

The whole concept of "providing typedefs" looks suspicious because redeclarations of the same typedef could be both providing and non-providing:
```cpp
class Class;
using Alias = Class; // non-providing
class Class {};
using Alias = Class; // providing
```
But it is required to block internal stuff of libraries from exposition to users, AFAIU. Probably, typdefs need some special handling in symbol mappings, but I have no complete picture of how it should look like.

Btw, @firewave, it would be nice if you test on cppcheck codebase that https://github.com/bolshakov-a/include-what-you-use/commits/ios_mapping/ doesn't introduce any regressions at least. You could also try the changeset on top of that with spurious mapping entries removed from `stdlib_cpp_include_map`: https://github.com/bolshakov-a/include-what-you-use/commit/a0d246921e928b5d2126d2230763599a089bf1d1.

firewave:
Since my systems lack easy access to LLVM 23 nightly builds I added jobs to the Cppcheck CI which utilizes an IWYU nightly - see https://github.com/cppcheck-opensource/cppcheck/pull/8595. You should be able to use that yourself in a fork for tests (just delete all other workflows and change the `on` to `[push, pull_request]`).

Current: https://github.com/cppcheck-opensource/cppcheck/actions/runs/26625348237/job/78460634902
ios_mapping: https://github.com/cppcheck-opensource/cppcheck/actions/runs/26625704365
remove_some_std_inc_mappings: https://github.com/cppcheck-opensource/cppcheck/actions/runs/26625714094

`ios_mapping` does not make a difference at all.

The other one looks like a mixed bag. Still keeps the full includes where the fwd would be sufficient. Suggests `<iostream>` where it should be `<istream>` and `<ostream>`. Biggest improvement is suggesting the full includes where they were missing before. So overall better but still not great. I'd say open a PR for it and we get properly into the weeds. I will also try to do manual "perfect" state to see how it deals with that and what the differences are.

bolshakov-a:
@firewave, thanks!

> `ios_mapping` does not make a difference at all.

As expected, generally, as I've noticed above.

> Still keeps the full includes where the fwd would be sufficient.

If you use typedefs like `ostream` and not underlying templates like `basic_ostream`, it is also expected. #1990 should fix such behavior (but I'm still not confident with that patch; see the discussion there). But if you manually replace complete definition headers with `<iosfwd>`, IWYU should accept that.

> Suggests `<iostream>` where it should be `<istream>` and `<ostream>`.

That's a deficiency (or a feature?) of [the IWYU "set cover" algorithm](https://github.com/include-what-you-use/include-what-you-use/blob/77cb2a4118adc3392348e3c0f70b2a81e9a8fc27/iwyu_output.cc#L1091-L1094), see https://github.com/include-what-you-use/include-what-you-use/pull/1990#issuecomment-4232647496. PR #1990 should also fix it, if merged.

> I'd say open a PR for it and we get properly into the weeds.

This PR and some other preparatory stuff should be merged first.

bolshakov-a:
> Suggests `<iostream>` where it should be `<istream>` and `<ostream>`.

I've tested on the following:
```cpp
// direct.h
#include <istream>
#include <ostream>

// fn.cpp
#include "direct.h"

void Fn(std::istream& is, std::ostream& os) {
    int i;
    is >> i;
    os << i;
}
```
The output of IWYU from the master branch is definitely wrong:
```console
fn.cpp should add these lines:
#include <istream>  // for basic_istream, basic_ostream, basic_istream::oper...

fn.cpp should remove these lines:

The full include-list for fn.cpp:
#include <istream>  // for basic_istream, basic_ostream, basic_istream::oper...
```
IWYU from [`remove_some_std_inc_mappings` branch](https://github.com/bolshakov-a/include-what-you-use/commits/remove_some_std_inc_mappings/) does suggest `<iostream>`:
```console
fn.cpp should add these lines:
#include <iostream>  // for basic_istream, basic_ostream, basic_istream::ope...

fn.cpp should remove these lines:

The full include-list for fn.cpp:
#include <iostream>  // for basic_istream, basic_ostream, basic_istream::ope...
```
I can confirm that `prefer_canonical` branch (from #1990), being rebased onto `remove_some_std_inc_mappings`, does fix the issue:
```console
fn.cpp should add these lines:
#include <istream>  // for basic_istream, basic_istream::operator>>, istream
#include <ostream>  // for basic_ostream, basic_ostream::operator<<, ostream

fn.cpp should remove these lines:

The full include-list for fn.cpp:
#include <istream>  // for basic_istream, basic_istream::operator>>, istream
#include <ostream>  // for basic_ostream, basic_ostream::operator<<, ostream
```

kimgr:
I think this PR is good to go. The typedef problem looks a bit thorny, but I saw the brute-force fix in the other branch -- that might be a good way to go for libstdc++ specifically. Let's merge this and work from the new baseline?

tests/cxx/fwd_decl_mapping_path_local.cc
```diff
@@ -0,0 +1,36 @@
...
+
+#include "fwd_decl_mapping_path_local-d1.h"
+
+// A forward-declaration of Class is mapped to '*-i2.h', which in turn is
+// exported by '*-i1.h', so including '*-i1.h' should be enough.
+Class* pClass;
+
+// Make sure '*-i1.h' is to be reported.
+// IWYU: ClassFromI1 is...*-i1.h
+ClassFromI1 cfi1;
```
kimgr, line 23: Hmm, do we have a simpler example where there's ONLY a forward-decl and no full use from the header providing the forward-decl? Something like:

```
#include "xxx-d1.h"

// this is the only use of xxx-i1.h
Class *pClass;
```

kimgr: Just want to make sure that the base case works without interference from other includes.

bolshakov-a: I think that the main `fwd_decl_mapping.cc` test should cover such cases, if I understand you correctly. It should be irrelevant to the include path form: note that the mapped headers from that test don't exist at all. This test has been written specifically to test `including_filepath`-related stuff of `GetCandidateHeadersForSymbolFwdDecl` function.

bolshakov-a:
> Let's merge this and work from the new baseline?

Yes, please!

tests/cxx/fwd_decl_mapping_path_local.cc
bolshakov-a: If you suggest to remove any other use of `-i1.h` except `Class`, `-i1.h` would not be reported at all in that case, because there is public-to-public mapping of `-i2.h` to `-i1.h` (not private to public). And IWYU would just report `-i2.h` (as spelled in the mapping file, I think).

kimgr: Yes, that's what I was getting at. The two-step mapping/export is good, but the simplest case comes first. I'll check the fwd_decl_mapping case when I have time (birthday party here today! 🎂)

kimgr: OK, I see the `OnlyFwdDeclUse` case captures this nicely.

PR #2029, opened 2026-05-17
Make public_headers() const

kimgr:
Inline SetPublicHeaders (only used from public_headers()), and mark public_headers() const.

Mark the public_headers_ member mutable to allow lazy-loading.

No functional change.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -349,26 +349,22 @@ string OneUse::PrintableUseLoc() const {
   return PrintableLoc(use_loc());
 }
 
-void OneUse::SetPublicHeaders() {
-  // We should never need to deal with public headers if we already know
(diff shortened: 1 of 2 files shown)


bolshakov-a:
I'm a little concerned about inlining `SetPublicHeaders`: the distinct function allowed me to place [an early return](https://github.com/include-what-you-use/include-what-you-use/pull/2028/changes#diff-96c9906a158b909fda0f001cb5bd04615d46704df24173306894f3d228abbed1R360) for fwd-decl mapped headers instead of increasing the nesting level of conditional constructs; otherwise LGTM.

kimgr:
@bolshakov-a I see, I missed that you were changing `SetPublicHeaders` in your PR. Let's forget about this one.

PR #2030, opened 2026-05-18
Report template static data member definitions

bolshakov-a:
Prior to this, instantiated static data members were attributed to their declarations inside the class template. Hence, their out-of-line definitions (which may reside in another header) were not suggested for inclusion in an instantiating file.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -166,8 +166,6 @@ using clang::UnresolvedLookupExpr;
 using clang::UsingDirectiveDecl;
 using clang::ValueDecl;
 using clang::VarDecl;
-using clang::VarTemplateDecl;
-using clang::VarTemplatePartialSpecializationDecl;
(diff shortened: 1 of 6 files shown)


bolshakov-a:
Perhaps, the word "handle" in the commit message is "too loud" because I suspect that template static data members are still not handled properly in all cases, e.g. the argument types needed for their instantiation aren't reported, probably.

bolshakov-a:
Yeah, e.g. in this case:
```cpp
#include "tests/cxx/direct.h"

template <typename T>
struct Tpl {
  static T t;
};

template <typename T>
T Tpl<T>::t;

void Fn() {
  (void)Tpl<IndirectClass>::t;
}
```
IWYU wrongly attributes the full use of `IndirectClass` to the static data member definition location instead of the instantiation location (bug #1181). So I've changed the commit message to some more humble one.

PR #2032, opened 2026-05-19
Add bug test for issue 2031

kimgr:


tests/bugs/2031/2031.cc
```diff
@@ -0,0 +1,32 @@
...
+  // std::string should be reported here, or possibly at the GetGreeting
+  // definition. But the macro expansion confuses IWYU.
+  std::string greeting = GetGreeting(BOB);
+  std::cout << greeting << std::endl;
+  return 0;
+}
+
+/**** IWYU_SUMMARY
+
+(tests/bugs/2031/2031.cc has correct #includes/fwd-decls)
```
bolshakov-a, line 30: Shouldn't `(tests/bugs/2031/2031.h has correct #includes/fwd-decls)` be here as well?

kimgr: Thanks, good catch. I'll follow up with something when I have a sec.

kimgr: Done, pushed directly to master at 3fdba2fb4cd3fddfe8a76250976d343aaf650b46.

PR #2037, opened 2026-05-22
Report explicitly instantiated fn and var decls

bolshakov-a:
This can be done now because after https://github.com/llvm/llvm-project/commit/fb024337497, there are distinct `ExplicitInstantiationDecl` nodes in the AST corresponding to every explicit instantiation declaration or definition.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -210,6 +210,7 @@ using clang::ElaboratedTypeKeyword;
 using clang::EnumConstantDecl;
 using clang::EnumDecl;
 using clang::EnumType;
+using clang::ExplicitInstantiationDecl;
 using clang::Expr;
 using clang::ExprResult;
 using clang::FriendDecl;
@@ -5101,9 +5102,58 @@ class IwyuAstConsumer
       return true;
     if (decl->isExplicitSpecialization())
       ReportDeclUse(CurrentLoc(), decl->getSpecializedTemplate());
-    // TODO(bolshakov): handle explicit instantiations. Looks like clang doesn't
-    // store all of the explicit instantiation redeclarations in the AST.
     return Base::VisitVarTemplateSpecializationDecl(decl);
+  }
+
+  bool VisitExplicitInstantiationDecl(ExplicitInstantiationDecl* decl) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+    const NamedDecl* spec = decl->getSpecialization();
+    if (const auto* fn_spec = dyn_cast<FunctionDecl>(spec)) {
+      if (decl->isExternTemplate()) {
+        // Explicit instantiation declarations require only any function
+        // declaration. If it is a class method, a declaration should be already
+        // provided by that class.
+        if (!fn_spec->isCXXClassMember()) {
+          ReportDeclUse(CurrentLoc(),
+                        fn_spec->getPrimaryTemplate()->getTemplatedDecl());
+        }
+      } else {
+        if (fn_spec->getTemplateSpecializationKind() ==
+            clang::TSK_ExplicitSpecialization) {
+          // Explicit instantiation definition doesn't do anything if it refers
+          // to an explicit specialization, so any redeclaration of the latter
+          // goes.
+          ReportDeclUse(CurrentLoc(), fn_spec);
+        } else {
+          // Otherwise, report the instantiation pattern. Another template
+          // redeclaration in the instantiating file is not enough,
+          // so UF_InstantiationPattern is specified.
+          ReportDeclUse(CurrentLoc(), spec, nullptr, UF_InstantiationPattern);
+          // TODO(bolshakov): report required template arguments.
+        }
+      }
+    } else if (const auto* var_tpl_spec =
+                   dyn_cast<VarTemplateSpecializationDecl>(spec)) {
+      if (decl->isExternTemplate()) {
+        // An explicit specialization (if present) should be reported even
+        // in the case of a member variable template because it may alter
+        // the variable type.
+        ReportDeclUse(CurrentLoc(), var_tpl_spec);
+      } else {
+        ReportDeclUse(CurrentLoc(), spec, nullptr, UF_InstantiationPattern);
+        // TODO(bolshakov): report required template arguments.
+      }
+    } else if (const auto* var_spec = dyn_cast<VarDecl>(spec)) {
+      // Static data member of a template class. The type cannot be changed, and
+      // there should not be any redeclaration but the definition.
+      if (!decl->isExternTemplate()) {
+        ReportDeclUse(CurrentLoc(), var_spec);
+        // TODO(bolshakov): report required template arguments.
+      }
+    }
+    // TODO(bolshakov): handle class template and member class instantiations.
+    return Base::VisitExplicitInstantiationDecl(decl);
   }
 
   // Track use of namespace in using directive decl
@@ -5217,8 +5267,6 @@ class IwyuAstConsumer
         ReportDeclUse(CurrentLoc(), tpl_decl);
       }
     }
-    // TODO(bolshakov): handle explicit instantiations. Looks like clang doesn't
-    // store all of the explicit instantiation redeclarations in the AST.
     return Base::VisitFunctionDecl(decl);
   }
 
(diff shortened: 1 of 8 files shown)


iwyu.cc
```diff
@@ -5101,11 +5102,62 @@ class IwyuAstConsumer
...
+        // provided by that class.
+        if (!fn_spec->isCXXClassMember()) {
+          ReportDeclUse(CurrentLoc(),
+                        fn_spec->getPrimaryTemplate()->getTemplatedDecl());
+        }
+      } else {
+        if (fn_spec->getTemplateSpecializationKind() ==
+            clang::TSK_ExplicitSpecialization) {
+          // Explicit instantiation definition doesn't do anything if it refers
+          // to an explicit speicalization, so any redeclaration of the latter
```
kimgr: typo: speicalization

iwyu.cc
```diff
@@ -5101,11 +5102,62 @@ class IwyuAstConsumer
...
+          // goes.
+          ReportDeclUse(CurrentLoc(), fn_spec);
+        } else {
+          // Otherwise, report the instantiation pattern. Another template
+          // redeclaration in the instantiating file is not enough,
+          // so UF_RedeclUse is specified.
```
kimgr: I originally decided for the use-flags that they were supposed to say something about the context in which the use occurred, not about how the use was to be treated downstream. I think because the original flag `UF_InCxxMethodBody` was promoted from a boolean `in_cxx_method_body`, and that flag really carries context info about the use location.

Does `UF_RedeclUse` here match the semantics for other redecl uses...? Almost, right? There's also an `UF_ExplicitInstantiation` with the same downstream semantics, but I don't think that matches -- it means the use _targets_ an explicit instantiation, but these uses seem to originate from an explicit instantiation.

This is rambling a bit, but I think if we want to keep using these flags, we should make sure to use them with the same intent, not just throw on a flag because we want an outcome. I may be overthinking, so if you really intended to model these links between explicit instantiation and underlying specialization as redecls, then all is good. But otherwise maybe we should just add a `UF_InstantiatedSpecialization` (or something) and add it to the mask checked in iwyu_output.

bolshakov-a: I agree with you, and I thought that "redeclaration use" was semantically correct in this case (after all, `GetDefinitionAsWritten(fn_spec)` below looks like a selection of that redeclaration which is actually the definition), but OTOH, a function template instantiation pattern really isn't a redeclaration of the specializations, strictly speaking.

kimgr: Looking at it from the other direction, is it OK if we hit this path: https://github.com/include-what-you-use/include-what-you-use/blob/b37561b32d43709d7644323993c6c5718b7ab00f/iwyu_output.cc#L1345?

>  OTOH, a function template instantiation pattern really isn't a redeclaration of the specializations, strictly speaking.

Yeah, this is what's bothering me. If we were to come up with a new policy in iwyu_output that applies to all `UF_RedeclUse`s, we don't want to accidentally include these specialization/instantiation relationships, but rather make a conscious choice whether to use nominal or canonical location, and whether to apply the `B3` policy.

But it's very fuzzy. We can also just fix it if it becomes a problem.

bolshakov-a: > Looking at it from the other direction, is it OK if we hit this path

It is exactly why I've used `UF_RedeclUse`. Any primary function template redeclaration in the using file should not suppress reporting the definition, as in the case of ordinary functions.

Maybe, it would be even more convenient to introduce a separate flag, so as not to call `GetDefinitionAsWritten` explicitly (it should be called inside `ReportFullSymbolUse`).

kimgr: Ah, so we already compensate? Yeah, then a separate flag would make sense. But it might also be nice to turn the flags around to actually _do_ say something about how the use should be handled (e.g. `UF_DefinitionLoc`, `UF_CanonicalLoc`, `UF_SomethingAboutFnDefaultArgs`), but then `UF_InCxxMethodBody` is hard to formulate.

bolshakov-a: Btw, I've discovered that for implicit specializations, a function redeclaration inside a using file does suppress reporting of the template definition:
```cpp
// fn_tpl.h
template <class T>
void TplFn() {}

// user.cpp
#include "fn_tpl.h"

template <class>
void TplFn();

void User() {
  TplFn<int>();
}
```
```console

- #include "fn_tpl.h"  // lines 1-1

---
```
But in contrast to the explicit instantiation case, the code still compiles after applying the suggestion. So I'm not completely sure this should be fixed. But if should, `UF_InstantiatedSpecialization` flag could help.

kimgr: Doesn't that look like good behavior? If there's a suitable cheaper decl available, don't require the `#include`. But it's probably unintentional, because I guess IWYU doesn't suggest to replace an existing `#include` with a fwd-decl?

bolshakov-a: > Doesn't that look like good behavior? If there's a suitable cheaper decl available, don't require the `#include`.

Maybe, but you may end up with a linker error if `TplFn<int>` is not instantiated in some other translation unit.

> I guess IWYU doesn't suggest to replace an existing `#include` with a fwd-decl?

No, IWYU doesn't suggest to insert a `TplFn` redeclaration instead of `#include "fn_tpl.h"` when there isn't one, if I understood your question correctly.

iwyu_output.cc
```diff
@@ -1341,8 +1341,11 @@ void ProcessFullUse(OneUse* use, const IwyuPreprocessorInfo* preprocessor_info,
   // those types, iwyu *does* care about the definition vs declaration.
   // All this is moot for decl uses triggered by definitions, all their redecls
```
bolshakov-a: > We don't bother with RedeclarableTemplate<> types (FunctionTemplateDecl), since for those types, iwyu *does* care about the definition vs declaration.

I don't know if it still is (was) relevant.

iwyu_use_flags.h
```diff
@@ -19,6 +19,7 @@ const UseFlags UF_None = 0;
 const UseFlags UF_InCxxMethodBody = 1;       // use is inside a C++ method body
 const UseFlags UF_RedeclUse = 2;             // use is due to some redeclaration
 const UseFlags UF_ExplicitInstantiation = 4; // use targets an explicit instantiation
+const UseFlags UF_InstantiatedSpecialization = 8; // primary template defn use by an instantiation
```
bolshakov-a: Should some formatting be applied here?

kimgr: Or we eliminate the problem? https://github.com/include-what-you-use/include-what-you-use/pull/2038

kimgr: Also
> UF_InstantiatedSpecialization = 8; // primary template defn use by an instantiation

maybe `UF_InstantiatedTemplate` if it refers to the primary template defn?

iwyu.cc
kimgr: I see, yeah.

iwyu_use_flags.h
bolshakov-a: > maybe `UF_InstantiatedTemplate` if it refers to the primary template defn?

Hard to say. As I noticed already, function template declarations without body can also be instantiated (and their instantiation may even require some argument type to be complete, like in the case `T::NestedType TplFn();`). But the key point here is to report the definition, and that it should not be suppressed by any function or variable template redeclaration.

bolshakov-a: So maybe `UF_InstantiatedTemplateDfn`? Or `UF_InstantiationPattern`?

kimgr: I like `UF_InstantiationPattern`, if only because it's slightly shorter. Let's try it and see how it looks as a whole. We can always rename it if we think of something better.

iwyu_output.cc
kimgr: Probably not, it looks like it predates the more advanced template reporting and `UF_RedeclUse`, so now we do the "filtering" for templates earlier?

bolshakov-a: Looks like `FunctionTemplateDecl` should not appear here (their "templated decls" should instead, which are `FunctionDecl`s). And where they appear, there are [bugs](https://github.com/include-what-you-use/include-what-you-use/pull/2016/changes#r3143816232).

Should I remove that part of the comment in this PR?

kimgr: Yes, please.

kimgr:
Looks great, thanks for persevering!

PR #2038, opened 2026-05-23
Move UseFlag comments to their own lines

kimgr:
The trailing line comments make it hard to add a new flag, as doing so may cause a reformat of all existing flags when the aligned comments don't fit.


bolshakov-a:
LGTM, thanks!

PR #2039, opened 2026-05-24
Report fully used func explicit instantiation args

### LINKED ISSUE #958: iwyu removes necessary header files when forward declaring template function
Consider this example:

file1.h
```
struct X{};
```

file2.h
```
template<typename T> 
void templFunc();
```

file2.cxx
```
#include "file2.h"
#include "file1.h"

template<typename T> 
void templFunc() {}

template<>
void templFunc<X>();

```

And the output of iwyu:
```
[iwans@home]$ include-what-you-use file2.cxx

(file2.h has correct #includes/fwd-decls)

file2.cxx should add these lines:
struct X;

file2.cxx should remove these lines:
- #include "file1.h"  // lines 2-2

The full include-list for file2.cxx:
#include "file2.h"
struct X;
```

iwyu has incorrectly removed file1.h and forward declared the struct X

bolshakov-a:
This fixes #958.


iwyu_ast_util.cc
```diff
@@ -1175,41 +1218,24 @@ TemplateInstantiationData GetTplInstDataForFunction(
...
-                 << PrintableType(type) << "\n";
-      }
-    }
-    // TODO(bolshakov): more fine-grained determination of provided types could
-    // e.g. consider only arguments that actually take part in type deduction.
-    // Simply moving the line below into the if-statement body above doesn't
-    // work because 'type' can be just an internal component of a 'typedef'ed
-    // argument.
-    InsertAllInto(provided_getter(type), &provided_types);
-  }
```
bolshakov-a, line 1196: I moved this into `GetDeducedTplArgData` so as to reuse in `GetTplExplicitInstData`.

kimgr:
Thanks -- in particular for thinking about the provision cases, I still have no intuition for how to handle those.

PR #2043, opened 2026-05-30
Don't report non-aggregate classes in some cases

bolshakov-a:
Reporting default-initialized non-aggregate parts of an aggregate is redundant when the initialization is not reflected in the source. In such cases, the type should already be complete.


kimgr:
Again rebased and ran tests locally without issue. LGTM!

PR #2046, opened 2026-05-31
Adjust to new Clang

bolshakov-a:
https://github.com/llvm/llvm-project/commit/6768170423542b772f9dfbfd21b06eefd983e188 changes `ExplicitInstantiationDecl::getNumTemplateArgs()` return type to optional. Moreover, it fixes traversing type of variable template explicit instantiation, but some handling was needed in `CanForwardDeclareType` method because only explicit instantiation definitions require the complete type.

https://github.com/llvm/llvm-project/commit/3a25f7b645fa68343d548eb116d884fd1ebb3edb removes obtaining definitions from `getTemplateInstantiationPattern` methods, so it should be done at the call sites now.

### DIFF
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -1263,8 +1263,10 @@ const NamedDecl* GetInstantiatedFromDecl(const NamedDecl* decl) {
       return pattern;
     }
   } else if (const auto* var_decl = dyn_cast<VarDecl>(decl)) {
-    if (const VarDecl* pattern = var_decl->getTemplateInstantiationPattern())
-      return pattern;
+    if (const VarDecl* pattern = var_decl->getTemplateInstantiationPattern()) {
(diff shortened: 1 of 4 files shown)


iwyu_ast_util.cc
```diff
@@ -1011,7 +1011,7 @@ static map<const Type*, const Type*> GetTplTypeResugarMapForExplicitTplArgs(
 static map<const Type*, const Type*> GetTplTypeResugarMapForExplicitTplArgs(
     const ExplicitInstantiationDecl* decl) {
   map<const Type*, const Type*> retval;
-  for (unsigned i = 0, ie = decl->getNumTemplateArgs(); i < ie; ++i) {
+  for (unsigned i = 0; i < decl->getNumTemplateArgs().value_or(0); ++i) {
```
bolshakov-a, line 1014: Removing `ie` is accidental, but with it the line exceeds the 80-col limit.

PR #2049, opened 2026-06-03
Update README to describe how to safely download IWYU releases

kimgr:
Explain where releases are available for download.

Cover how to get the public key, and how to use it to verify tamper-free downloads.


kimgr:
Read formatted here: https://github.com/kimgr/include-what-you-use/tree/readme-gpg#how-to-download

PR #2050, opened 2026-06-03
Use UseKind for internal symbol mappings

bolshakov-a:


iwyu_include_picker.cc
```diff
@@ -1562,18 +1581,23 @@ void WriteMappings(StringRef directive,
       << FormatISO8601(time(nullptr)) << "\n";
   out << "[\n";
   for (size_t i = 0; i < count; ++i) {
-    const IncludeMapEntry& entry = entries[i];
-    CHECK_(entry.from_visibility != kUnusedVisibility)
-        << "cannot export unknown from-visibility";
+    const auto& entry = entries[i];
+    if constexpr (is_same_v<MapEntry, IncludeMapEntry>) {
```
kimgr, line 1585: C++! You've grown up since I last checked!

kimgr:
Very nice!

iwyu_include_picker.cc
bolshakov-a: Yeah, it is more convenient than various template metaprogramming techniques.

PR #2051, opened 2026-06-05
Print "ptr only" uses according to comment_style

bolshakov-a:
Prior to this, printing comments for "ptr only" uses did not take into account `--comment_style` option, so they were printed without namespace qualification on `--comment_style=long`. Even worse, IWYU failed to detect if a header is already required for the full use, so the both comments appeared, like this:
```cpp
#include "class.h"  // for ns::Class, Class (ptr only)
```


bolshakov-a:
I've noticed this while exploring the logs from https://github.com/include-what-you-use/include-what-you-use/pull/2028#issuecomment-4572526349.

PR #2052, opened 2026-06-07
Add mappings for I/O stream headers

bolshakov-a:
Template declarations with default arguments from `<iosfwd>` should not be recategorized to full uses (at least before mapping), so A2 and A3 steps have been swapped around and made mutually exclusive.

I/O typedef list has been hard-coded in `iwyu_ast_util.cc` to handle the cases when an implementation (re-)defines typedefs after underlying type definitions, thus making them "providing" for IWYU author-intent policy (libstdc++ case).


mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -181,23 +182,27 @@ def __exit__(self, *t):
...
         self.__text = text
+        self.__fwd_decl_header = fwd_decl_header
         self.__cur_pos = 0
         self.__namespace_stack = []
         # __tpl_params holds parameters while parsing a template declaration.
         self.__tpl_params = []
-        self.found_names = []
+        # found_names is a dictionary that maps symbols to boolean values
+        # indicating whether it is a forward-declaration or not.
+        self.found_names = {}
```
bolshakov-a, line 198: Maybe it should better be a list of classes?

iwyu_output.cc
```diff
@@ -1169,27 +1169,30 @@ void ProcessForwardDeclare(OneUse* use,
...
+    // No return here: (A4) or (A5) may cause us to ignore this decl entirely.
+  }
+  // (A3) If it has default template parameters, recategorize as a full use.
+  // Do this even if the used decl is a fwd-decl so that IWYU doesn't insert
+  // another fwd-decl specifying the same default argument. But keep the use
+  // kind when the forward-declaration is in the same file so as not to suggest
+  // removing the forward-declaration as unused. Uses of fwd-decls from <iosfwd>
+  // should not be recategorized to full uses so that (D1) step works with them,
+  // hence 'else' has been placed here.
+  else if (tpl_decl && HasDefaultTemplateParameters(tpl_decl) &&
```
bolshakov-a, line 1192: @kimgr, I suspect you don't like such formatting.

iwyu_ast_util.cc
```diff
@@ -1570,6 +1570,56 @@ unsigned GetTplParamNumberWithoutPack(const TemplateDecl* tpl) {
...
+    "std::wofstream",
+    "std::wfstream",
+    "std::wsyncbuf",
+    "std::wosyncstream",
+    "std::streampos",
+    "std::wstreampos",
+    "std::u8streampos",
+    "std::u16streampos",
+    "std::u32streampos",
+  };
```
bolshakov-a, line 1620: Clang-format compresses this list if run on it. Should I put `// clang-format off`?

bolshakov-a (commented):
This change breaks at least this case with libc++ (with `-std=c++20` and `-D_LIBCPP_ENABLE_EXPERIMENTAL` specified):
```cpp
#include <syncstream>

void Fn(std::osyncstream&);
```
On master, IWYU correctly suggests replacing `<syncstream>` with `<iosfwd>` because there are no `syncstream` mappings there, and `osyncstream` typedef is defined in `<iosfwd>`. On this branch, IWYU outputs that the `#include` list is correct because `<syncstream>` is now a public header for `osyncstream` typedef along with `<iosfwd>`. So maybe #1990 should be merged first.

iwyu_ast_util.cc
kimgr: `// clang-format off` sounds good to me. Could you also sort the symbol names so it's easier to scan/find an insertion point for new ones?

iwyu_output.cc
kimgr: Yeah, I find they obscure the control flow a bit. But it's consistent with what's already there, so I can do a refactor pass at some point and move them around if it annoys me too much :-).

Looks good as-is.

iwyu_ast_util.cc
bolshakov-a: Currently, they are in the order they appear in [`<iosfwd>` synopsis](https://timsong-cpp.github.io/cppwp/n4950/iosfwd.syn). Do you think alphabetical order would be better?

firewave:
As this supersedes #1683 shouldn't it also get of those mappings as well? Or am I missing something?

iwyu_ast_util.cc
kimgr: Yeah, I think alphabetical is better, much quicker to see if something is missing, etc.

mapgen/iwyu-mapgen-std-symbol.py
kimgr: You're thinking something like list of `Symbol(name, fwd_decl)`? I think the plain dict is clearer for now, let's keep it as-is.

bolshakov-a:
> As this supersedes https://github.com/include-what-you-use/include-what-you-use/pull/1683 shouldn't it also get of those mappings as well?

I think it can be a separate step, after implementing include mappings for standardized dependencies.

kimgr:
Thank you!

bolshakov-a:
Hey @kimgr, are you sure this is not a problem? https://github.com/include-what-you-use/include-what-you-use/pull/2052#pullrequestreview-4444758439

> This change breaks at least this case with libc++ ...

kimgr:
@bolshakov-a I thought it looked unusual enough to live with for now. Sounds like you were saying #1990 will eventually fix it, so I think it's OK to be a little suboptimal until then.

bolshakov-a:
However, I still don't know when (and if) #1990 will be merged...

firewave:
FYI this doesn't appear to have any effect on Cppcheck code.

bolshakov-a:
Yeah, I remember. I don't expect this to have effect unless some code spells e.g. `basic_ostream<char, char_traits<char>>` instead of just `ostream`.

Issue #1122, opened 2022-10-06
Invalid suggestions for templates using operator overload from macro expansion


graveljp:
IWYU provides invalid recommendation when a template uses an overloaded operator defined via a macro expansion. See here for a minimalist example reproducing this problem: https://github.com/graveljp/iwyu-bugs/tree/main/operator_overload_macro

You can reproduce by doing:
```
git clone https://github.com/graveljp/iwyu-bugs.git
cd iwyu-bugs/operator_overload_macro
include-what-you-use main.cc
```

This outputs:
```
main.cc should add these lines:
#include "my_int.h"  // for operator*

main.cc should remove these lines:

The full include-list for main.cc:
#include "my_int.h"  // for operator*
#include "utils.h"   // for TimesTen
---
```

`main.cc` shouldn't be including `my_int.h`. Interestingly, the problem doesn't happens if the `operator*` isn't defined by a macro in `my_int.h`, or if the `operator*` is replaced with a normal `Multiply()` function.

Tests at on latest master (IWYU at https://github.com/include-what-you-use/include-what-you-use/commit/c1d8dd310a3b2e0601d44ece73fba9bc037e82d1 and LLVM at https://github.com/llvm/llvm-project/commit/bc4bcbcfc820b324f680e8f260691c38052eedc9).

graveljp:
It's worth pointing out that clangd (in [UnusedIncludes: Strict](https://clangd.llvm.org/guides/include-cleaner#configuration) mode) also agrees that `main.cc` shouldn't be including `my_int.h`.

kimgr:
At verbosity level 5, IWYU will explain a little bit about its reasoning for use locations in macros. You can add `-Xiwyu -v5` and look for `Attributing use of ... to location at ...`. The preceding log lines cover the decision tree there.

If behavior is different with/without macro, I suspect it would mis-attribute the use of `MyInt` in a way that causes it to also run into the template instantiation problem space. Or something.

graveljp:
I was able to get the same problem by replacing the macro invocation with an include statement to inject code in the header, though in that particular example, it means doing an `#include` at the end of a file, which is stretching it a bit:
```
struct MyInt {
  int value;
};
MyInt MakeMyInt(int value) {
  return {value};
}
#include "operations.h"
```
[...]

PR #2053, opened 2026-06-08
Improve checking macro locations in templates

bolshakov-a:
The primary goal of this fix was to handle [`Q_DECLARE_METATYPE`](https://doc.qt.io/qt-6/qmetatype.html#Q_DECLARE_METATYPE) macro from Qt. For the code like this:
```cpp
// my_struct.h
struct MyStruct {};

// my_struct_meta.h
#include "my_struct.h"
#include <QMetaType>

Q_DECLARE_METATYPE(MyStruct)

// fn.cpp
#include "my_struct_meta.h"
#include <QVariant>

void Fn() {
  MyStruct ms;
  QVariant v;
  v.setValue(ms);
}
```
IWYU suggested replacing `my_struct_meta.h` with `my_struct.h` which would break compilation because `Q_DECLARE_METATYPE` is needed to be able to put user's type to `QVariant`. It was because [`qvariant.h` includes `qmetatype.h`](https://github.com/qt/qtbase/blob/89f6199e17273b2cdb38fb93ddc152f55c093934/src/corelib/kernel/qvariant.h#L11) that [defines `Q_DECLARE_METATYPE`](https://github.com/qt/qtbase/blob/89f6199e17273b2cdb38fb93ddc152f55c093934/src/corelib/kernel/qmetatype.h#L1517), so `QMetaTypeId` specialization was considered as provided by `QVariant::setValue` template function.

Besides, this fixes the original issue of #1122.


iwyu.cc
```diff
@@ -4436,10 +4436,21 @@ class InstantiatedTemplateVisitor
...
-              GetFileEntry(ast_node->GetLocation()),
-              GetFileEntry(decl->getLocation())))
+              GetLocFileEntry(GetInstantiationLoc(ast_node->GetLocation())),
+              decl_expansion_file)) {
+        return GetInstantiationLoc(ast_node->GetLocation());
+      }
+      if (preprocessor_info().PublicHeaderIntendsToProvide(
+              GetLocFileEntry(GetSpellingLoc(ast_node->GetLocation())),
+              decl_expansion_file)) {
         return ast_node->GetLocation();
```
bolshakov-a, line 4452: Perhaps, this should be `return GetSpellingLoc(ast_node->GetLocation());` for consistency. However, this should not change behavior because `GetSpellingLoc` is called later inside `GetFileEntry` function called from `InstantiatedTemplateVisitor::ReportDeclUse`.

PR #2058, opened 2026-06-18
Add libstdcxx-specific operator""s mapping

### LINKED ISSUE #2057: Regression when using `operator""s`
A regression was introduced in https://github.com/include-what-you-use/include-what-you-use/pull/2014, where IWYU incorrectly resolves `operator""s` provided by <chrono>.

Consider the following MRE:
```c++
#include <chrono>
using namespace std::chrono_literals;
int main() {
    auto a = 10s;
    return 0;
}
```

At commit f6400c725f0de1fabd6edb6bacaae65ee52cc746 (prior to the PR changes), IWYU evaluates the source file correctly:
```
$ include-what-you-use test.cpp 

```

However, starting with commit ae18593bd2f3d4b16d8e9921aa95af0afb368490, IWYU fails:
```
$ include-what-you-use test.cpp 

#include <string>  // for operator""s

- #include <chrono>  // lines 1-1

#include <string>  // for operator""s
---
```

---

From https://en.cppreference.com/cpp/chrono/operator%22%22s:
> `std::string` also defines `operator""s`, to represent literal objects of type `std::string`, but it is a string literal: `10s` is ten seconds, but `"10"s` is a two-character string.

bolshakov-a:
`-Wno-user-defined-literals` has been added in tests because otherwise clang warns that literal suffixes not beginning with `_` are reserved.

This fixes #2057.

### DIFF
--- a/tests/cxx/std_symbol_mapping.cc
+++ b/tests/cxx/std_symbol_mapping.cc
@@ -7,7 +7,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-// IWYU_ARGS: -I .
+// IWYU_ARGS: -I . -Wno-user-defined-literals
 
 // Tests the internal IWYU standard library symbol mapping. In particular, tests
 // that the mapping works with types in the fwd-decl context.
@@ -84,6 +84,12 @@ void Fn() {
   std::basic_spanstream<char>* ss;
   // IWYU: std::basic_spanstream is...*<spanstream>
   (void)sizeof(*ss);
+
+  {
+    using namespace std::chrono_literals;
+    // IWYU: operator""s({{.*}}) is...*<chrono>
+    (void)10s;
+  }
 }
 
 /**** IWYU_SUMMARY
@@ -91,6 +97,7 @@ void Fn() {
 tests/cxx/std_symbol_mapping.cc should add these lines:
 #include <algorithm>
 #include <array>
+#include <chrono>
 #include <cmath>
 #include <functional>
 #include <iosfwd>
@@ -104,6 +111,7 @@ tests/cxx/std_symbol_mapping.cc should remove these lines:
 The full include-list for tests/cxx/std_symbol_mapping.cc:
 #include <algorithm>  // for move
 #include <array>  // for array, get, operator==
+#include <chrono>  // for operator""s
 #include <cmath>  // for pow
 #include <functional>  // for function, swap
 #include <iosfwd>  // for basic_ostream (ptr only), char_traits (ptr only)
--- a/mapgen/iwyu-mapgen-std-symbol.py
+++ b/mapgen/iwyu-mapgen-std-symbol.py
@@ -161,6 +161,10 @@ HANDWRITTEN_MAPPING = (
       'ostream'),
     ('std::print', 'print'),
     ('std::println', 'print'),
+# libstdc++ defines a literal operator for seconds as a numeric literal operator
+# template despite the <chrono> header synopsis. This may be standardized later.
+# See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85821.
+    ('std::operator""s()', 'chrono'),
 )
 
 def contains_unspec_alias(inp):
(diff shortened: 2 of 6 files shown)


tests/cxx/std_symbol_mapping.cc
```diff
@@ -84,13 +84,20 @@ void Fn() {
   std::basic_spanstream<char>* ss;
   // IWYU: std::basic_spanstream is...*<spanstream>
   (void)sizeof(*ss);
+
+  {
+    using namespace std::chrono_literals;
+    // IWYU: operator""s({{.*}}) is...*<chrono>
```
kimgr, line 90: I was a bit worried about this shadowing `operator""s` from `bits/basic_string.h`/`<string>`, but I added this and it appears to work beautifully:
```diff
diff --git a/tests/cxx/std_symbol_mapping-direct.h b/tests/cxx/std_symbol_mapping-direct.h
index 0aea1e1e..13c65782 100644
--- a/tests/cxx/std_symbol_mapping-direct.h
+++ b/tests/cxx/std_symbol_mapping-direct.h
@@ -106,4 +106,15 @@ template <char...>
 chrono::seconds operator""s();
 }  // namespace chrono_literals
 }  // namespace literals
+
+template <typename T, typename = char_traits<T>>
+class basic_string {};
+using string = basic_string<char>;
+
+inline namespace literals {
+inline namespace string_literals {
+basic_string<char> operator""s(const char*, size_t);
+}  // namespace string_literals
+}  // namespace literals
+
 }
diff --git a/tests/cxx/std_symbol_mapping.cc b/tests/cxx/std_symbol_mapping.cc
index 16f749fa..b364968a 100644
--- a/tests/cxx/std_symbol_mapping.cc
+++ b/tests/cxx/std_symbol_mapping.cc
@@ -90,6 +90,12 @@ void Fn() {
     // IWYU: operator""s({{.*}}) is...*<chrono>
     (void)10s;
   }
+
+  {
+    using namespace std::string_literals;
+    // IWYU: operator""s({{.*}}) is...*<string>
+    (void)"literally, a string"s;
+  }
 }
 
 /**** IWYU_SUMMARY
@@ -102,6 +108,7 @@ tests/cxx/std_symbol_mapping.cc should add these lines:
 #include <functional>
 #include <iosfwd>
 #include <spanstream>
+#include <string>
 #include <utility>
 #include <valarray>
 
@@ -114,8 +121,9 @@ The full include-list for tests/cxx/std_symbol_mapping.cc:
 #include <chrono>  // for operator""s
 #include <cmath>  // for pow
 #include <functional>  // for function, swap
-#include <iosfwd>  // for basic_ostream (ptr only), char_traits (ptr only)
+#include <iosfwd>  // for basic_ostream (ptr only)
 #include <spanstream>  // for basic_spanstream
+#include <string>  // for char_traits (ptr only), operator""s
 #include <utility>  // for get, move, operator==, pair, swap
 #include <valarray>  // for pow, valarray
```
The why-comment doesn't qualify which `operator""s` overload is pulled in, but that seems OK to me.

I don't know if it would make sense to add that test coverage (separately) now that there's some ambiguity over `operator""s`?

kimgr (approved):

tests/cxx/std_symbol_mapping.cc
bolshakov-a: > I was a bit worried about this shadowing `operator""s` from `bits/basic_string.h`/`<string>`

If you are worried that the regular expression is too broad, I can replace it with `{{(unsigned long long)?}}`.

> I don't know if it would make sense to add that test coverage (separately) now that there's some ambiguity over `operator""s`?

Maybe, but if some standard library implementers replaced the standardized versions of literal operators for `std::string` with a string literal operator template, the test would not show the problem.

kimgr: No, the test formulation looks fine, I was more concerned about the actual mappings getting ambiguous with both `string_literals` and `chrono_literals`, but it seems you have that covered :)

> Maybe, but if some standard library implementers replaced the standardized versions of literal
> operators for std::string with a string literal operator template, the test would not show the
> problem.

Alright, I'm not sure what that would look like, but if you feel it puts in unnecessary constraints, let's leave it out.

bolshakov-a: > I was more concerned about the actual mappings getting ambiguous with both `string_literals` and `chrono_literals`

They are not ambiguous unless literals for `std::string` are implemented through literal operator template as well. Recall that `()` in mappings means "empty parameter list", whereas no parentheses at all means a fallback mapping entry accepting any parameters not covered by explicit mappings. The fallback mapping for [`std::string` literals](https://timsong-cpp.github.io/cppwp/n4950/string.view.literals) is needed so as not to specify the (unspecified) `size_t` underlying type.

PR #2060, opened 2026-06-21
Log consistent messages whenever uses are modified by policy

kimgr:
Several of the places where we ignore or reclassify a use to and from
fwd-decl had no logging. This caused uses to silently disappear, which
made it harder to make policy changes.

Add an 'Ignoring ... use of ...' log wherever we set a use to ignored.

Add new 'Moving...' logs wherever we promote/demote uses to/from
fwd-decl. Make all 'Moving...' logs follow the same form.

Add new 'Updating target decl...' logs wherever we modify the target
decl using reset_decl.

### DIFF
--- a/iwyu_output.cc
+++ b/iwyu_output.cc
@@ -1204,9 +1204,9 @@ void ProcessForwardDeclare(OneUse* use,
   if (!use->is_full_use()) {
     if (preprocessor_info->ForwardDeclareIsInhibited(
             GetFileEntry(use->use_loc()), use->symbol_name())) {
-      VERRS(6) << "Changing fwd-decl use of " << use->symbol_name()
-               << " (" << use->PrintableUseLoc()
-               << ") to a full-info use: no_forward_declare pragma\n";
+      VERRS(6) << "Moving " << use->symbol_name()
+               << " from fwd-decl use to full use: no_forward_declare pragma"
+               << " (" << use->PrintableUseLoc() << ")\n";
       use->set_full_use();
     }
   }
@@ -1257,7 +1257,7 @@ void ProcessForwardDeclare(OneUse* use,
   }
 
   // (A7) If any arbitrary redeclaration is marked with IWYU pragma: export,
-  // reset use as a full use of this decl to keep its containing file included.
+  // reset use to this decl to keep its containing file included.
   if (!use->is_full_use()) {
     for (const Decl* redecl : use->decl()->redecls()) {
       const auto* decl = cast<NamedDecl>(redecl);
@@ -1266,6 +1266,10 @@ void ProcessForwardDeclare(OneUse* use,
         use->set_suggested_header(ConvertToQuotedInclude(
             GetFilePath(decl->getLocation()),
             MakeAbsolutePath(GetParentPath(GetFilePath(use->use_loc())))));
+        VERRS(6) << "Updating target decl for " << use->symbol_name()
+                 << " fwd-decl use: export pragma"
+                 << " (" << use->PrintableUseLoc() << "), using decl from"
+                 << " " << PrintableLoc(GetLocation(decl)) << "\n";
         break;
       }
     }
@@ -1287,11 +1291,18 @@ void ProcessForwardDeclare(OneUse* use,
         // don't keep another include that is not necessary.
         use->reset_decl(cast<NamedDecl>(decl));
         promote_to_full_use = true;
+        VERRS(6) << "Updating target decl for " << use->symbol_name()
+                 << " fwd-decl use: no_fwd_decls mode"
+                 << " (" << use->PrintableUseLoc() << "), using decl from"
+                 << " " << PrintableLoc(GetLocation(decl)) << "\n";
         break;
       }
     }
 
     if (promote_to_full_use) {
+      VERRS(6) << "Moving " << use->symbol_name()
+               << " from fwd-decl use to full use: no_fwd_decls mode"
+               << " (" << use->PrintableUseLoc() << ")\n";
       use->set_full_use();
     }
   }
@@ -1359,7 +1370,7 @@ void ProcessFullUse(OneUse* use, const IwyuPreprocessorInfo* preprocessor_info,
     // Just change us to a forward-declare use.  Later, we'll decide
     // which forward-declare is the best one to keep.
     VERRS(6) << "Moving " << use->symbol_name()
-             << " from full use to fwd-decl: definition found later in file"
+             << " from full use to fwd-decl use: definition found later in file"
              << " (" << use->PrintableUseLoc() << ")\n";
     use->set_forward_declare_use();
     return;
@@ -1514,6 +1525,9 @@ void ProcessFullUse(OneUse* use, const IwyuPreprocessorInfo* preprocessor_info,
               const NamedDecl* decl = use.decl();
               return decl && (ns_decl != decl) && IsInNamespace(decl, ns_decl);
             })) {
+      VERRS(6) << "Ignoring use of namespace " << use->symbol_name() << " ("
+               << use->PrintableUseLoc() << "):"
+               << " declaration in namespace already used\n";
       use->set_ignore_use();
       return;
     }
@@ -1618,6 +1632,9 @@ void CalculateIwyuForForwardDeclareUse(
     }
     if (IsBeforeInSameFile(dfn, use->use_loc())) {
       // TODO(bolshakov): this looks like a duplicate of A6 step. Deduplicate.
+      VERRS(6) << "Ignoring fwd-decl use of " << use->symbol_name() << " ("
+               << use->PrintableUseLoc() << "):"
+               << " dfn already present\n";
       use->set_ignore_use();
       return;
     }
@@ -1638,6 +1655,9 @@ void CalculateIwyuForForwardDeclareUse(
   if (!same_file_decl) {
     for (const NamedDecl* redecl : redecls) {
       if (ContainsKey(associated_includes, GetFileEntry(redecl))) {
+        VERRS(6) << "Ignoring fwd-decl use of " << use->symbol_name() << " ("
+                 << use->PrintableUseLoc() << "):"
+                 << " redecl present in associated header\n";
         use->set_ignore_use();
         return;
       }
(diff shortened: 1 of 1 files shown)


kimgr:
@bolshakov-a Inspired by this policy: https://github.com/include-what-you-use/include-what-you-use/blob/6bb3aa3094160b22ccad41e0ffc7db84dc6cbb16/iwyu_output.cc#L1508 -- which silently threw away any namespace decls I reported for 2009.

iwyu_output.cc
```diff
@@ -1266,6 +1266,11 @@ void ProcessForwardDeclare(OneUse* use,
         use->set_suggested_header(ConvertToQuotedInclude(
             GetFilePath(decl->getLocation()),
             MakeAbsolutePath(GetParentPath(GetFilePath(use->use_loc())))));
+        VERRS(6) << "Moving " << use->symbol_name()
+                 << " from fwd-decl use to full use: export pragma"
+                 << " (" << use->PrintableUseLoc() << "),"
+                 << " used dfn from " << PrintableLoc(GetLocation(decl))
+                 << "\n";
```
bolshakov-a: There is no promotion to full use here. I forgot to fix the comment in fb419bfdaaaf3f776015c00195c67046a26e68a8.

iwyu_output.cc
```diff
@@ -1292,6 +1297,11 @@ void ProcessForwardDeclare(OneUse* use,
     }
 
     if (promote_to_full_use) {
+      VERRS(6) << "Moving " << use->symbol_name()
+               << " from fwd-decl use to full use: no_fwd_decls mode"
+               << " (" << use->PrintableUseLoc() << "),"
+               << " used dfn from " << PrintableLoc(GetLocation(use->decl()))
```
bolshakov-a:
```suggestion
               << " used decl from " << PrintableLoc(GetLocation(use->decl()))
```
It is not necessarily the definition, as I can see.

iwyu_output.cc
```diff
@@ -1618,6 +1631,9 @@ void CalculateIwyuForForwardDeclareUse(
     }
     if (IsBeforeInSameFile(dfn, use->use_loc())) {
       // TODO(bolshakov): this looks like a duplicate of A6 step. Deduplicate.
+      VERRS(6) << "Ignoring fwd-decl use of " << use->symbol_name() << " ("
+               << use->PrintableUseLoc() << "):"
+               << " specialization dfn already present\n";
```
bolshakov-a: Tbh, I don't quite remember if it applies only to specializations here, and that the case of the (primary template) definition is always covered by (A6).

iwyu_output.cc
```diff
@@ -1638,41 +1654,52 @@ void CalculateIwyuForForwardDeclareUse(
...
-             << PrintableLoc(GetLocation(providing_decl)) << "\n";
-  }
-  if (providing_decl) {
-    // Change decl_ to point to this "better" redecl.
-    use->reset_decl(providing_decl);
+    use->reset_decl(same_file_decl);
+    VERRS(6) << "Moving " << use->symbol_name()
+             << " from fwd-decl use to full use: satisfied by same file"
+             << " (" << use->PrintableUseLoc() << "),"
+             << " used dfn from " << PrintableLoc(GetLocation(use->decl())) << "\n";
```
bolshakov-a:
```suggestion
             << " used decl from " << PrintableLoc(GetLocation(use->decl())) << "\n";
```
And there is also no moving to full use, but it is less important here.

iwyu_output.cc
```diff
@@ -1638,41 +1654,52 @@ void CalculateIwyuForForwardDeclareUse(
...
-             << PrintableLoc(GetLocation(providing_decl)) << "\n";
     // Mark that this use is another reason we want this header.
     CHECK_(!desired_hdr_with_dfn.empty());
+    use->reset_decl(dfn_from_desired_includes);
     use->set_suggested_header(desired_hdr_with_dfn);
+    VERRS(6) << "Moving " << use->symbol_name()
+             << " from fwd-decl use to full use: satisfied by desired file"
+             << " (" << use->PrintableUseLoc() << "),"
+             << " used dfn from " << PrintableLoc(GetLocation(use->decl()))
+             << "\n";
```
bolshakov-a: There is no moving to full use here as well. I think this fact is important for "(ptr only)" comment to appear.

iwyu_output.cc
```diff
@@ -1638,41 +1654,52 @@ void CalculateIwyuForForwardDeclareUse(
...
+             << " (" << use->PrintableUseLoc() << "),"
+             << " used dfn from " << PrintableLoc(GetLocation(use->decl())) << "\n";
   }
 
   // Be sure to store as a TemplateClassDecl if we're a templated
   // class.
   if (const ClassTemplateSpecializationDecl* spec_decl =
           DynCastFrom(use->decl())) {
     use->reset_decl(spec_decl->getSpecializedTemplate());
+    VERRS(6) << "Moving " << use->symbol_name() << " to specialization decl ("
```
bolshakov-a:
```suggestion
    VERRS(6) << "Moving " << use->symbol_name() << " to primary template decl ("
```
And I'm not sure `" used dfn from "` is correct here. AFAIU, it remains a fwd-decl use, so the definition header may still be discarded.

bolshakov-a (changes requested):
I've found some issues.

iwyu_output.cc
kimgr: Rephrased to `Moving X from fwd-decl use to *included decl*`.

And fixed the comment.

iwyu_output.cc
kimgr: Fixed, thanks

iwyu_output.cc
kimgr: I played it down a bit to the unspecified `template decl`, do you think that works?

iwyu_output.cc
kimgr: Nice, went with `included decl` here too.

iwyu_output.cc
kimgr: Nice, went with `included decl` here too, and updated `dfn` to `decl`

iwyu_output.cc
kimgr: Thanks; I updated `specialization` -> `primary template` and `dfn` -> `decl`.

iwyu_output.cc
```diff
@@ -1638,41 +1654,51 @@ void CalculateIwyuForForwardDeclareUse(
...
+             << " (" << use->PrintableUseLoc() << "),"
+             << " used decl from " << PrintableLoc(GetLocation(use->decl())) << "\n";
   }
 
-  // Be sure to store as a TemplateClassDecl if we're a templated
-  // class.
+  // Be sure to store as a ClassTemplateDecl if we're a templated class.
   if (const ClassTemplateSpecializationDecl* spec_decl =
           DynCastFrom(use->decl())) {
     use->reset_decl(spec_decl->getSpecializedTemplate());
```
kimgr, line 1688: Actually, this transform looks a little weird to me. If we have a forward-decl use of a template specialization, why do we reset it to a forward-decl use of the primary template? They can be in different files, right?

iwyu_output.cc
bolshakov-a: Are you sure this piece of code works only with templates? Probably, yes, but `" dfn already present\n"` would be sufficient, I think.

iwyu_output.cc
bolshakov-a: "Included decl" looks a little strange with "same file", but OTOH "included" is not necessarily to be read as "external header included".

iwyu_output.cc
bolshakov-a: That's a good question. I commented out this branch, and all the tests passed.

bolshakov-a:
A couple of minor notes, but now it LGTM!

iwyu_output.cc
kimgr: Ah, the [patch that introduced it](fb53a762eb96fde23ba198242d3bf3a9ded3874a) dropped a comment that said:
```
// OK, we're going to forward-declare this type.  Make sure it's
// a template class decl, and not an instantiation decl.
```
It didn't occur to me that "specialization" is the formal term for "instantiation".

That makes a bit more sense, but I'm surprised no test code triggers this... So maybe it's already resolved earlier.

iwyu_output.cc
kimgr: Not at all sure (anymore :slightly_smiling_face:) Thanks, that comment makes sense, and vibes with the block comment I overlooked earlier.

iwyu_output.cc
kimgr: Yeah, looking at it fresh, it does look a little off. I'll see if I can come up with something better.

kimgr: OK, I changed all the `Moving ... included decl` to distinct `Updating target decl for ... fwd-decl use`, which makes a bit more sense in context.

For `--no_fwd_decls`, I first log `Updating target decl` if the decl is reset, and then `Moving ...` when the use is promoted.

iwyu_output.cc
bolshakov-a: I've managed to find a case where removing this branch leads to an error:
```cpp
// tpl_fwd_decl.h
template <int> class Tpl;

// src.cpp
#include "tpl_fwd_decl.h"

using Alias = Tpl<1>;

template <> class Tpl<1> {};
```
Before removing, the output is correct:
```console
template <int> class Tpl;

- #include "tpl_fwd_decl.h"  // lines 1-1

template <int> class Tpl;
```
After removing:
```console
class Tpl;
template <int> class Tpl;

- #include "tpl_fwd_decl.h"  // lines 1-1

class Tpl;
template <int> class Tpl;
```
Btw, if there is a primary template definition of `Tpl` immediately before the explicit specialization definition:
```cpp
// src.cpp
...
template <int> class Tpl {};
template <> class Tpl<1> {};
```
the output on master is not very good:
```console
template <int> class Tpl;
template <int> class Tpl;

- #include "tpl_fwd_decl.h"  // lines 1-1

template <int> class Tpl;
template <int> class Tpl;
```
but it compiles at least.

bolshakov-a: > It didn't occur to me that "specialization" is the formal term for "instantiation".

In fact, they are different terms, and both formal, so it is better not to confuse them. However, "implicit instantiation" does mean nearly the same as "implicit specialization", despite the former is a process, and the latter is a result.

kimgr:
Updated, please take another look

bolshakov-a:
> Updated, please take another look

I think this is now good.

iwyu_output.cc
kimgr: Edited your example into `tests/cxx/template_specialization.cc` in https://github.com/include-what-you-use/include-what-you-use/pull/2067.

Issue #1667, opened 2024-12-11
STL-C headers seem to be ignored


pbielefeldt:
Dear all,

I am using iwyu on a C++ project on openSUSE Leap with CLang version 15.0.7, iwyu version 0.19. I would like to set it up in a way that it recommends replacing lines such as 
```c++
#include <stdlib.h>
```
with `#include <cstdio>`.

My understanding is that this should be done when using the mapping file [stl.c.headers.imp](https://github.com/include-what-you-use/include-what-you-use/blob/master/stl.c.headers.imp#L35).

I have a small dummy file:
```c++
// a.cxx
#include <iostream>
#include <chrono>
#include <stdlib.h>

int main()
{
	std::cout << "Hello World" << std::endl;
	return EXIT_SUCCESS;
}

```

On that file (`a.cxx`), I run
```bash
include-what-you-use -Xiwyu --mapping_file=<path>/include-what-you-use/stl.c.headers.imp a.cxx
```

## Actual Result ##

iwyu returns this result for me:
```
a.cxx should add these lines:

a.cxx should remove these lines:
- #include <chrono>  // lines 2-2

The full include-list for a.cxx:
#include <stdlib.h>  // for EXIT_SUCCESS
#include <iostream>  // for operator<<, endl, basic_ostream, cout, ostream
---

```

## Expected Result ##

The finding about `<chrono>` is as expected.

I would however expect it to propose exchanging `#include <stdlib.h>  // for EXIT_SUCCESS` with `#include <cstdlib>  // for EXIT_SUCCESS`.

Note that if I manually exchange stdlib.h for cstdlib, iwyu does not propose changing it back to stdlib.h.

hexagonrecursion:
stl.c.headers.imp is for a different purpose entirely. This file is one of the mappings that are **used by default**. It prevents iwyu from replacing `<cstdio>` with `<stdio.h>`, but it does not make iwyu replace `<stdio.h>` with `<cstdio>` either. Currently the only way to make iwyu replace `<stdio.h>` with `<cstdio>` is to `--no_default_mappings` and then make **your own** mapping for the entire C++ standard library by copying and editing the built-in mappings. Beware: I tried that and couldn't quite get it to work. Ended up using sed instead:
```sh
    find \( -name '*.cpp' -o -name '*.h' \) -execdir sed -i '
        s/^#include <assert.h>/#include <cassert>/;
        s/^#include <complex.h>/#include <ccomplex>/;
[...]

Issue #1126, opened 2022-10-12
IWYU suggests to add C header instead of C++ header


lovehina13:
I have a C++ source code with `size_t` but without its explicit header (inherited by another).

IWYU report suggests to add the line `#include <stddef.h>` instead of `#include <cstddef>` recommended by C++ guidelines.

I checked the `size_t` is associated with `#include <stddef.h>` in `libc_symbol_map` definition of `iwyu_include_picker.cc` file and it's actually normal. Also, the `stdlib_c_include_map` definition prevents to ignore C++ to C headers conversions.

But in case of C++ source code, it would be better that IWYU report suggest to add/replace C headers with C++ notation.

Is there an option to activate this behavior? `--help` command doesn't specify anything about this.

_Note: I tried to set custom mapping but it fails due to double definition._

Bu11etmagnet:
Note that `<cstddef>` does not define `size_t`. It defines `std::size_t`.

kimgr:
@Bu11etmagnet It defines both in most implementations.

As far I can tell from web searching without actually checking the standard text, the wording is that `<cname>` must define all the symbols from `<name.h>` inside namespace std, and is _allowed to_ also define them in the global namespace.

Apparently there is also work underway to deprecate the `<cname>` headers again in favor of just using the C headers. But that's probably still on the far horizon.

kimgr:
Should be fixed as of: https://github.com/include-what-you-use/include-what-you-use/commit/146292c5224f204166bb855ba8bf392685f92691

Issue #1093, opened 2022-08-09
stddef.h recommended in C++ code


martingalvan-nordic:
Hi, I'm running an IWYU I built manually from the clang_14 branch (--version shows include-what-you-use 0.18 (git:abd5d2b) based on Ubuntu clang version 14.0.6-++20220622053131+f28c006a5895-1~exp1~20220622173215.157). OS is Ubuntu 20.04.

I noticed that IWYU is recommending `#include <stddef.h>` for `size_t` in .cpp files. I'm invoking it through CMake as follows:

`/usr/bin/cmake -E __run_co_compile --iwyu="/usr/local/bin/include-what-you-use;-Xiwyu;--cxx17ns;-Xiwyu;--no_fwd_decls;-Xiwyu;--mapping_file=/usr/local/bin/../share/include-what-you-use/stl.c.headers.imp;--driver-mode=g++" <other args come here>`

You can see that I'm explicitly including the stl.c.headers.imp mapping file (I don't know whether the mappings provided by IWYU are used by default). However, I still see stddef.h being recommended instead of cstddef. Am I doing something wrong?

kimgr:
I can confirm that the equivalent of `stl.c.headers.imp` is baked into IWYU by default.

IWYU mappings are not keyed by language, so the expectation that `size_t` should come from `<cstddef>` for C++ and `<stddef.h>` for C unfortunately does not hold.

Perhaps you could add an additional mapping file that just says `<stddef.h>` (private) is mapped to `<cstddef.h>` (public). Or, alternatively, provide an override symbol mapping from `size_t` (private) to `<cstddef>` (public). Though I suspect both of those will fail due to artificial constraints in IWYU.

I see there's a public-to-public mapping from `<stddef.h>` to `<cstddef>`, which I think means `<cstddef>` is accepted as a stand-in for `<stddef.h>`. So if you were to search-and-replace `<stddef.h>` to `<cstddef>` manually, at least IWYU should leave the existing code alone. Try on one file first to verify.

bolshakov-a:
Isn't current behavior the correct one? Names in `std` namespace are declared in C++ headers, and names in global namespace are in C headers (despite the fact that the formers include the latters, really).

martingalvan-nordic:
> Isn't current behavior the correct one? Names in `std` namespace are declared in C++ headers, and names in global namespace are in C headers (despite the fact that the formers include the latters, really).

[...]

Issue #890, opened 2021-03-26
Suggest <cstdint> instead of <stdint.h>


carlosgalvezp:
Hi,

I want to use this amazing tool in our C++ project. We want to use C++ headers, so e.g. `cstdint` instead of `stdint.h`. 
However the tool is suggesting to include `stdint.h` for `uint32_t` (for example).

I see in the code written:
https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_include_picker.cc#L471

```
(If you *did* want to replace
  // assert.h with cassert, you'd change it to a public->private
  // mapping.)
```

Can you clarify how to implement this in practice? I've tried adding a mapping from `stdint.h` to `<cstdint>`, making `stdint.h` private, but then it complains that `stdint.h` has dual visibility (public and private).

The tool is great but for most symbols it finds it's asking us to use C headers, which is just wrong. The tool should be capable of suggesting either C headers or C++ headers depending on the file extension.

Thanks!

Krzmbrzl:
Looking at e.g.
https://github.com/include-what-you-use/include-what-you-use/blob/c5f3b0fe4c5407d6379626d946dc67089c0d3844/qt5_11.imp#L3
it seems that the `.h` version should actually be listed without the `.h` file extension in the mapping.

Did you do that?

(Warning: I don't actually have a clue about all this - this is just my first thought on this and I figured I might as well share it with you)

carlosgalvezp:
I don't think that's the case here. Qt doesn't have `.h` headers. Just like `cstdint.h` doesn't exist. There exists:

* C headers:       `<stdint.h>`
* C++ headers:  `<cstdint>`

I just have trouble understanding the sentence:
> you'd change it to a public->private mapping.

Change it where? Do I need to hack the source code, or can I do it in an .imp file? Change what? It would be nice with the exact line of code that accomplishes what the comment is suggesting.

kimgr:
@carlosgalvezp I think that may be a typo/thinko in the comment. It seems to me a private->public mapping would be more suitable.

Here's how IWYU works with mappings:

* There are the original, compiled-in mappings for libstdc++ and gnu libc (that's where you found the comment)
* There are external mappings in .imp files that can be applied on top of these built-in mappings. These are primarily meant to allow extensions for third-party libraries
* But there are also mirrors of the compiled-in mappings in external .imp files

[...]

PR #2062, opened 2026-06-24
Suggest `<c*>` headers instead of `<*.h>` in C++ mode

### LINKED ISSUE #270: IWYU suggests including <stddef.h> for C++ files
Actually `<cstddef>` should be suggested, but for my cpp files it says, I should add `<stddef.h>`

bolshakov-a:
This change makes `<cname>` headers the only acceptable option for the C standard library symbols in C++ mode. Strictly speaking, it is wrong for symbols from the global namespace, but the common implementations actually include `<name.h>` into `<cname>`. The new behavior is in line with the current C++ standard wording that discourages using `<name.h>` headers in source files not intended to be valid ISO C. If a file should be valid ISO C, IWYU should be launched on it in C mode (either implicitly based on the file extension, or explicitly with `-xc` or `-xc-header` flag).

For those who want to use `<name.h>` headers instead, a new command line flag `-Xiwyu --use_c_headers` has been introduced. It causes the opposite behavior: even if `<cname>` header is included directly, IWYU suggests replacing it with the corresponding `<name.h>` (except the case of explicitly spelled `std::` qualification). However, if a standard library implementation declares some tag type (like `timespec`) in `std` namespace first (which is unlikely to happen), it may not always work.

The order of adding the internal mappings has been changed so that `stdint.h` -> `cstdint` mapping path is preferred over `stdint.h` -> `inttypes.h` -> `cinttypes`.

The change in `iwyu_string_util.h` is because `<ctime>` now takes preference over `<cstddef>` for `size_t` because it is nailed down as the only candidate for `gmtime` and `strftime`.

This fixes #270, #890, #1093, #1126, and #1667.

### DIFF
--- a/include-what-you-use.1
+++ b/include-what-you-use.1
@@ -174,6 +174,10 @@ is already visible in the file's transitive includes.
 Print full include list with comments, even if there are no
 \(lqinclude-what-you-use\(rq violations.
 .TP
+.B \-\-use_c_headers
+In C++ mode, force use of C standard library headers \fB<\fIname\fP.h>\fR
+instead of their C++ counterparts \fB<c\fIname\fP>\fR.
+.TP
 .BI \-\-verbose= level
 Set verbosity. At the highest level, this will dump the AST of the source file
 and explain all decisions.
--- a/iwyu_globals.h
+++ b/iwyu_globals.h
@@ -112,6 +112,7 @@ struct CommandlineFlags {
   set<string> dbg_flags; // Debug flags.
   set<string> exp_flags;       // Experimental flags.
   RegexDialect regex_dialect;  // Dialect for regular expression processing.
+  bool use_c_headers;  // Force use C standard library headers in C++ mode.
 };
 
 const CommandlineFlags& GlobalFlags();
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -27,6 +27,7 @@
 #include "clang/Basic/LangOptions.h"
 #include "clang/Tooling/Inclusions/StandardLibrary.h"
 #include "iwyu_ast_util.h"
+#include "iwyu_globals.h"
 #include "iwyu_location_util.h"
 #include "iwyu_path_util.h"
 #include "iwyu_port.h"
@@ -1667,22 +1668,6 @@ void IncludePicker::AddInternalMappings(CStdLib cstdlib, CXXStdLib cxxstdlib) {
   using clang::tooling::stdlib::Lang;
   using clang::tooling::stdlib::Symbol;
 
-  if (cstdlib == CStdLib::Glibc) {
-    AddSymbolMappings(libc_symbol_map, IWYU_ARRAYSIZE(libc_symbol_map));
-    AddIncludeMappings(libc_include_map, IWYU_ARRAYSIZE(libc_include_map));
-  } else if (cstdlib == CStdLib::ClangSymbols) {
-    // Get canonical C standard library mappings from clang tooling
-    for (const Symbol& sym : Symbol::all(Lang::C)) {
-      string name = sym.name().str();
-
-      // the canonical header is returned first
-      for (const Header& header : sym.headers()) {
-        AddSymbolMapping(name, UseKind::Full,
-                         MappedInclude(header.name().str()), kPublic);
-      }
-    }
-  }
-
   if (cxxstdlib == CXXStdLib::Libstdcxx) {
     AddIncludeMappings(libstdcpp_include_map,
                        IWYU_ARRAYSIZE(libstdcpp_include_map));
@@ -1704,8 +1689,10 @@ void IncludePicker::AddInternalMappings(CStdLib cstdlib, CXXStdLib cxxstdlib) {
   if (cxxstdlib != CXXStdLib::None) {
     // Map C headers to associated C++ headers. The standard library
     // mappings shouldn't be mentioning the C headers.
-    AddIncludeMappings(stdlib_c_include_map,
-                       IWYU_ARRAYSIZE(stdlib_c_include_map));
+    if (!GlobalFlags().use_c_headers) {
+      AddIncludeMappings(stdlib_c_include_map,
+                         IWYU_ARRAYSIZE(stdlib_c_include_map));
+    }
     // C++ include mappings to allow different public headers that
     // generally include each other.
     AddIncludeMappings(stdlib_cpp_include_map,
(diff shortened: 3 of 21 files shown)


include-what-you-use.1
```diff
@@ -174,6 +174,10 @@ is already visible in the file's transitive includes.
 Print full include list with comments, even if there are no
 \(lqinclude-what-you-use\(rq violations.
 .TP
+.B \-\-use_c_headers
+In C++ mode, force using of C standard library headers <name.h> instead of their
+C++ counterparts <cname>.
```
bolshakov-a: @alejandro-colomar, could you propose some better wording or formatting? I'm not good at `roff`.

iwyu_include_picker.cc
```diff
@@ -1720,6 +1707,32 @@ void IncludePicker::AddInternalMappings(CStdLib cstdlib, CXXStdLib cxxstdlib) {
...
+      }
+    }
+  }
+
+  // HACK: Mark C standard library headers as private in C++ mode so that
+  // <cname> are suggested instead of <name.h>.
+  if (cxxstdlib != CXXStdLib::None && !GlobalFlags().use_c_headers) {
+    for (const IncludeMapEntry& entry : stdlib_c_include_map)
+      include_visibility_map_.at(entry.map_from) = IncludeVisibility::kPrivate;
+  }
```
bolshakov-a, line 1735: Perhaps, changing C standard header visibility should be done after loading external mapping files which may contain some more mappings for them? OTOH, it is probably better to notify users that their mapping files are not consistent with internal mappings rather than silently change the visibility, which may cause astonishment.

bolshakov-a (commented):
I was a bit concerned if it is correct to suggest C++ headers for POSIX symbols, but it turned out that there is no portable way to use them in a C++ program: https://discourse.llvm.org/t/portable-way-to-use-posix-stuff-in-c/91127.

One more concern is that the behavior of [the bug test](https://github.com/include-what-you-use/include-what-you-use/blob/6bb3aa3094160b22ccad41e0ffc7db84dc6cbb16/tests/bugs/1000/1000.cc) for #1000 has been changed: IWYU now suggests to completely replace `<cmath>` with `<cstddef>`. This can be fixed by means of adding the mapping of global-namespace `abs` functions. But OTOH, the new behavior is in line with #1990.

tests/cxx/quoted_includes_first.cc
```diff
@@ -27,13 +28,13 @@ IndirectSubDirClass global_var;
...
 
 tests/cxx/quoted_includes_first.cc should remove these lines:
+- #include <errno.h>  // lines XX-XX
 - #include <list>  // lines XX-XX
 
 The full include-list for tests/cxx/quoted_includes_first.cc:
 #include "tests/cxx/pch.h"
 #include "quoted_includes_first.h"
 #include "subdir/indirect_subdir.h"  // for IndirectSubDirClass
-#include <errno.h>  // for EACCES
```
kimgr, line 36: reminder to myself: what about `<cerrno>`?

kimgr: Aha, associated header.

iwyu_include_picker.cc
kimgr: Yeah, this echoes of my plans for #1271. I think we should keep the current, somewhat painful, behavior where mapping conflicts cause crashes/diagnostics, and then attack that systematically as a separate effort.

iwyu_globals.h
```diff
@@ -112,6 +112,7 @@ struct CommandlineFlags {
   set<string> dbg_flags; // Debug flags.
   set<string> exp_flags;       // Experimental flags.
   RegexDialect regex_dialect;  // Dialect for regular expression processing.
+  bool use_c_headers = false;
```
kimgr: Let's initialize `use_c_headers` in the constructor together with the other fields, and add a little comment here like for the others.

(I'd be open to a refactor PR to move initialization entirely into the header, but that seems low-prio)

include-what-you-use.1
kimgr: I'm also very new to *roff, but I see we've used `\fBword\fR` in other places to selectively bold keywords, include-names and commands, so maybe let's do that here too:
```
.B \-\-use_c_headers
In C++ mode, force use of C standard library headers \fB<name.h>\fR instead of
their C++ counterparts \fB<cname>\fR.
.TP
```
(slight wording suggestion there too: using -> use)

bolshakov-a: Maybe like that? `... \fB<\fIname\fP.h>\fR instead of their C++ counterparts \fB<c\fIname\fP>\fR.`
изображение

Here is some info about `roff`: https://man7.org/linux/man-pages/man7/groff_man_style.7.html.

kimgr: Nice, yeah, that makes sense.

alejandro-colomar: > @alejandro-colomar, could you propose some better wording or formatting? I'm not good at `roff`.

For some reason, I missed the notification for this.  Sorry.

Includes should go fully in italics.  At least, that's the common practice in the Linux man-pages project, and also the recommendation from groff_man(7).

That is:

```man
.I <cname>
```

or

```man
\f[I]<cname>\f[]
```

or

```man
\fI<cname>\fP
```

(in order of preference)

alejandro-colomar: I don't have the notification in my mailbox.  This is weird.

```sh
alx@devuan:~/mail/kernel/foss/github$ grep -rl formatting | xargs grep -h ^From: | sort | uniq
From: Alejandro Colomar <notifications@github.com>
From: EugeneZelenko <notifications@github.com>
From: Hadi Chokr <notifications@github.com>
From: Iker Pedrosa <notifications@github.com>
From: Karlson2k <notifications@github.com>
From: Kubernetes Triage Robot <notifications@github.com>
From: Richard Russon <notifications@github.com>
From: Sean McGinnis <notifications@github.com>
From: Tobias Deiminger <notifications@github.com>
From: Tobias Stoeckmann <notifications@github.com>
From: Tristan Miller <notifications@github.com>
From: Yi <notifications@github.com>
```

I also don't see the notification in my github notifications page.  Github is really crappy about notifications.

alejandro-colomar: > > @alejandro-colomar, could you propose some better wording or formatting? I'm not good at `roff`.
> 
> For some reason, I missed the notification for this. Sorry.
> 
> Includes should go fully in italics. At least, that's the common practice in the Linux man-pages project, and also the recommendation from groff_man(7).
> 
> That is:
> 
> ```man
> .I <cname>
> ```
> 
> or
> 
> ```man
> \f[I]<cname>\f[]
> ```
> 
> or
> 
> ```man
> \fI<cname>\fP
> ```
> 
> (in order of preference)

Actually, since 'name' is a variable part of the file name, it should go in roman.

Here's the relevant documentation from groff_man_style(7):

```
     .I [text]
            Set text in an italic or oblique face.  Given no argument, I plants
            a one‐line input trap; text on the next line, which can be  further
            formatted with a macro, is set in an italic or oblique face.

            Use italics for file and path names, for environment variables, for
            C  data  types, for enumeration or preprocessor constants in C, for
            variant (user‐replaceable) portions of  syntax  synopses,  for  the
            first  occurrence  (only)  of a technical concept being introduced,
            for names of journals and of literary works longer than an article,
            and anywhere a parameter requiring replacement by the user  is  en‐
            countered.  An exception involves variant text in a context already
            typeset  in  italics,  such  as file or path names with replaceable
            components; in such cases, follow the  convention  of  mathematical
            typography:  set  the file or path name in italics as usual but use
            roman for the variant part (see IR and RI below), and italics again
            in running roman text when referring to the variant material.
```

Thus:

```man
.IR <c name >
```

or

```man
\f[I]<c\f[R]name\f[I]>\f[P]
```

or

```man
\fI<c\fRname\fI>\fP
```

(in order of preference)

bolshakov-a: Thanks! To be done in some follow-up PR.

> I don't have the notification in my mailbox. This is weird.

Github doesn't work well sometimes.

bolshakov-a: I've opened #2065.

firewave:
I still see C headers being suggested in a C++ project:

```
/home/runner/work/cppcheck/cppcheck/lib/checkother.h should add these lines:
#include <stddef.h>      // for size_t
```

firewave:
> I still see C headers being suggested in a C++ project:
> 
> ```
> /home/runner/work/cppcheck/cppcheck/lib/checkother.h should add these lines:
> #include <stddef.h>      // for size_t
> ```

Nevermind, my branch was out-of-date.

hebasto:
Thanks for the patch! It works flawlessly in our [project](https://github.com/bitcoin/bitcoin), meaning we can finally drop our custom [patch](https://github.com/bitcoin/bitcoin/blob/f26c15bdd28f532f95ead48736146d88b66bbea7/ci/test/01_iwyu.patch#L17-L634) once the next IWYU release is out.

bolshakov-a:
I see you did nearly the same. But what was the reason to keep `<float.h>` public? https://github.com/bitcoin/bitcoin/blob/f26c15bdd28f532f95ead48736146d88b66bbea7/ci/test/01_iwyu.patch#L591

hebasto:
> I see you did nearly the same. But what was the reason to keep `<float.h>` public? https://github.com/bitcoin/bitcoin/blob/f26c15bdd28f532f95ead48736146d88b66bbea7/ci/test/01_iwyu.patch#L591

No particular reason. It was overlooked in https://github.com/bitcoin/bitcoin/commit/0b489886f85ff0430ae838dbc6a87da201033644.

bolshakov-a:
Ah, Ok.

PR #2063, opened 2026-06-28
Update symbol mapping documentation in manpage

kimgr:
The symbol mappings changed in cc406a285cc30168fc8115de2106e6b6bbfdc943; we can now specify 'full' or 'fwd-decl' mappings, which apply to the different kinds of uses.

Reflect this in the manpage.


kimgr:
@bolshakov-a

bolshakov-a:
Thanks!

PR #2065, opened 2026-06-30
Fix manpage according to groff style guide

bolshakov-a:
Filenames should be in italics with variant parts in roman.


bolshakov-a:
Now it looks like this:
изображение

Issue #1764, opened 2025-06-03
All intrinsics headers should be mapped to the umbrella header


hebasto:
Here is a minimum reproducible example (`intrinsics.cpp`):
```c++
#include <immintrin.h>

int main()
{
    __m128i a = _mm_set1_epi32(0);
    __m128i b = _mm_set1_epi32(1);
    __m128i r = _mm_blend_epi16(a, b, 0xFF);
    return _mm_extract_epi32(r, 3);
}
```

IWYU output:
```
$ include-what-you-use intrinsics.cpp

intrinsics.cpp should add these lines:
#include <emmintrin.h>  // for _mm_set1_epi32, __m128i
#include <smmintrin.h>  // for _mm_blend_epi16, _mm_extract_epi32

intrinsics.cpp should remove these lines:
- #include <immintrin.h>  // lines 1-1

The full include-list for intrinsics.cpp:
#include <emmintrin.h>  // for _mm_set1_epi32, __m128i
#include <smmintrin.h>  // for _mm_blend_epi16, _mm_extract_epi32
---
```

kimgr:
I suppose the intrinsics headers are different between compilers? So far I've found distinct GCC, Clang and Microsoft intrinsics.

In line with tradition, I guess we should support the GCC intrinsics by default, but it might be nice to think about the others too.

You should be able to come up with a mapping file for your environment to unblock yourself.

kimgr:
I went digging a bit, and I see Clang's intrinsic headers are part of the resource dir (`clang -print-resource-dir`).

That means IWYU will always see Clang's intrinsic headers (because it uses Clang as a library, and Clang assumes its intrinsic headers agree with the compiler implementation of said intrinsics) -- and never MSVC or GCC headers, even if the codebase may be portable across all three.

Is there some kind of agreement across compilers which header should provide which intrinsic? I can imagine GCC and Clang use the same public header names.

kimgr:
(btw I happened to notice we have external mappings for gcc and clang intrinsics in https://github.com/include-what-you-use/include-what-you-use/blob/a060293328e42028bf1a05622f3892c406d3fe26/clang-6.intrinsics.imp and gcc-8.intrinsics.imp)

PR #2066, opened 2026-07-02
Add intrinsics mappings for Clang 22

hebasto:
Derive private -> public mappings from the `#error` directives in the Clang 22.1.8 resource headers.

The directive in `avx512vp2intersectintrin.h` [misspells](https://github.com/llvm/llvm-project/blob/ca7933e47d3a3451d81e72ac174dcb5aa28b59d1/clang/lib/Headers/avx512vp2intersectintrin.h#L25) its own header name as `<avx512vp2intersect.h>`. The mapping uses the real file name.

Also map the component public headers (`<emmintrin.h>` etc.) to `<immintrin.h>` so IWYU accepts the umbrella header as a substitute.

Also see https://github.com/include-what-you-use/include-what-you-use/issues/1764.


firewave:
> The directive in `avx512vp2intersectintrin.h` [misspells](https://github.com/llvm/llvm-project/blob/ca7933e47d3a3451d81e72ac174dcb5aa28b59d1/clang/lib/Headers/avx512vp2intersectintrin.h#L25) its own header name as `<avx512vp2intersect.h>`. The mapping uses the real file name.

Should probably report that upstream.

kimgr:
@hebasto Thanks -- I'll tweak it a little to remove the (almost-)aligned formatting. It's nice to read, but it causes friction when editing.

kimgr:
Reformatted and merged as 6a113bcf7017b8d5b7b4358c8a0fb16be96ad39a. Thank you!

kimgr:
> > The directive in `avx512vp2intersectintrin.h` [misspells](https://github.com/llvm/llvm-project/blob/ca7933e47d3a3451d81e72ac174dcb5aa28b59d1/clang/lib/Headers/avx512vp2intersectintrin.h#L25) its own header name as `<avx512vp2intersect.h>`. The mapping uses the real file name.
> 
> Should probably report that upstream.

I'll open a quick LLVM PR for this.

kimgr:
... and it's merged. @hebasto do you have any scripting you could share for building these mappings? Or does it take human judgment to finalize them?

hebasto:
> @hebasto do you have any scripting you could share for building these mappings? Or does it take human judgment to finalize them?

Here is the Python script created by Claude Fable 5: https://gist.github.com/hebasto/d81a8f33875eb43d3f0833e9c3f50600.

kimgr:
Thanks! I'll let that inspire me to build a regular `iwyu-mapgen` and open a PR.

kimgr:
@hebasto Wait, that script doesn't cover the public-to-public mappings here, right: https://github.com/include-what-you-use/include-what-you-use/pull/2066/changes#diff-7e82e252e6ab18b8fba846cd112e258d66f753bc2fdf8874a763d063d2ca25d8R117 -- those are entirely curated?

hebasto:
> @hebasto Wait, that script doesn't cover the public-to-public mappings here, right: https://github.com/include-what-you-use/include-what-you-use/pull/2066/changes#diff-7e82e252e6ab18b8fba846cd112e258d66f753bc2fdf8874a763d063d2ca25d8R117 -- those are entirely curated?

Correct.

PR #2067, opened 2026-07-04
Add test coverage for template specialization reporting

kimgr:
In the review for cbded3f4a6c3a7b653f01560418a79b121d72a33, we started questioning the purpose of the reset from specialization to primary template decl was good for in CalculateIwyuForForwardDeclareUse.

Turns out it was not covered by tests; removing it had no effect.

@bolshakov-a came up with an example to exercise it; incorporate it into the template_specialization test case.

(Incidentally, the other branch in that conditional, which sets the described class template for CXXRecordDecls, is covered by the Specialized type in the same test case.)

### DIFF
--- a/tests/cxx/template_specialization-d2.h
+++ b/tests/cxx/template_specialization-d2.h
@@ -1,0 +1,11 @@
+//===--- template_specialization-d2.h - test input file for iwyu -*- C++ -*===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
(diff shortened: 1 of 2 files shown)


bolshakov-a:
Maybe, it's worth noting that the type alias is needed to trigger a full-use report because there is no forward-declaration in the alias-defining file. That full use is then recategorized to fwd-decl use because the definition is actually after the alias:
https://github.com/include-what-you-use/include-what-you-use/blob/6a113bcf7017b8d5b7b4358c8a0fb16be96ad39a/iwyu_output.cc#L1086-L1090

Otherwise LGTM.

kimgr:
Thanks, it actually wasn't entirely clear to me why the `using` was necessary, but I think I see it now. I'll make a little note.

PR #2068, opened 2026-07-05
Add mapping generator for Clang intrinsics

kimgr:
After 6a113bcf7017b8d5b7b4358c8a0fb16be96ad39a, @hebasto shared their approach for building mappings.

Implement it in IWYU's style, and include it as
mapgen/iwyu-mapgen-clang-intrin.py.

This mapping generator can produce both .imp for external mappings, and C++ initialization syntax (in the style of our existing std_symbol_map.inc).

The matching rules are somewhat limited, but they have been verified to work with Clang 23 mainline. We'll have to update them if Clang starts using more sophisticated messages (e.g. they used to have multiple public headers like `<a> or <b>`, that might happen again, and will need some more advanced parsing).

### DIFF
--- a/mapgen/iwyu-mapgen-clang-intrin.py
+++ b/mapgen/iwyu-mapgen-clang-intrin.py
@@ -1,0 +1,179 @@
+#!/usr/bin/env python3
+
+##===--- iwyu-mapgen-clang-intrin.py --------------------------------------===##
+#
+#                     The LLVM Compiler Infrastructure
+#
+# This file is distributed under the University of Illinois Open Source
+# License. See LICENSE.TXT for details.
+#
+##===----------------------------------------------------------------------===##
+
+"""Generates mappings for Clang intrinsics headers.
+
+The Clang intrinsics have a simple rule: each private header contains an #error
+directive which points to the public header:
+
+    #if !defined X86GPRINTRIN_H_
+    #error "Never use <bmiintrin.h> directly; include <x86gprintrin.h> instead."
+    #endif
+
+Scan over all intrinsics headers, match the #error messages and build mappings.
+"""
+import sys
+import os
+import re
+import glob
+import argparse
+import json
+import textwrap
+
+
+# In addition to the static patterns in the intrinsics headers, there are also
+# some known public-to-public mappings where public headers are interchangeable.
+# Keep a curated mapping.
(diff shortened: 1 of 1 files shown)


kimgr:
@hebasto Please let me know what you think

kimgr:
Oh, and also -- what would happen if we generated the C++ and baked it into IWYU? I can't quite see through how intrinsic headers from different compilers interact (cf #1764), so not sure if the status quo for portable projects gets better or worse if we assume Clang intrinsics internally?

hebasto (commented):
Should the script parse headers in `ppc_wrappers`?

Asking because it currently results in the following:
```console
$ cat /usr/lib/clang/22/include/bmi2intrin.h | grep "directly"
#error "Never use <bmi2intrin.h> directly; include <immintrin.h> instead."
$ cat /usr/lib/clang/22/include/bmiintrin.h | grep "directly"
#error "Never use <bmiintrin.h> directly; include <x86intrin.h> instead."
$ ./mapgen/iwyu-mapgen-clang-intrin.py --lang imp /usr/lib/clang/22/include | grep "<bmi"
  { "include": ["<bmi2intrin.h>", "private", "<immintrin.h>", "public"] },
  { "include": ["<bmi2intrin.h>", "private", "<x86gprintrin.h>", "public"] },
  { "include": ["<bmiintrin.h>", "private", "<x86gprintrin.h>", "public"] },
  { "include": ["<bmiintrin.h>", "private", "<x86intrin.h>", "public"] },
```

kimgr:
Oh, I see. I didn't notice that the header names were the same between root and ppc_wrappers :facepalm:.  OK, I think PowerPC is niche enough at this point that we can ignore mappings for these headers. I don't see a clear way of keeping them conditional.

kimgr:
Updated, PTAL.

hebasto (approved):
LGTM.

kimgr:
Thank you!

hebasto:
> Oh, and also -- what would happen if we generated the C++ and baked it into IWYU? I can't quite see through how intrinsic headers from different compilers interact (cf #1764), so not sure if the status quo for portable projects gets better or worse if we assume Clang intrinsics internally?

For many private headers, GCC instructs users to include `<x86gprintrin.h>` rather than `<immintrin.h>`.

PR #2069, opened 2026-07-05
Deduplicate fwd-decls using the canonical decl

### LINKED ISSUE #2064: Specializations and primary templates are considered interchangeable
We were discussing this line: https://github.com/include-what-you-use/include-what-you-use/blob/cbded3f4a6c3a7b653f01560418a79b121d72a33/iwyu_output.cc#L1688 in [a basically-unrelated PR](https://github.com/include-what-you-use/include-what-you-use/pull/2060#pullrequestreview-4587736778), and @bolshakov-a summarized:

I've managed to find a case where removing this branch leads to an error:
```cpp
// tpl_fwd_decl.h
template <int> class Tpl;

// src.cpp
#include "tpl_fwd_decl.h"

using Alias = Tpl<1>;

template <> class Tpl<1> {};
```
Before removing, the output is correct:
```console
template <int> class Tpl;

- #include "tpl_fwd_decl.h"  // lines 1-1

template <int> class Tpl;
```
After removing:
```console
class Tpl;
template <int> class Tpl;

- #include "tpl_fwd_decl.h"  // lines 1-1

class Tpl;
template <int> class Tpl;
```
Btw, if there is a primary template definition of `Tpl` immediately before the explicit specialization definition:
```cpp
// src.cpp
...
template <int> class Tpl {};
template <> class Tpl<1> {};
```
the output on master is not very good:
```console
template <int> class Tpl;
template <int> class Tpl;

- #include "tpl_fwd_decl.h"  // lines 1-1

template <int> class Tpl;
template <int> class Tpl;
```
but it compiles at least.

_Originally posted by @bolshakov-a in https://github.com/include-what-you-use/include-what-you-use/pull/2060#discussion_r3488380202_

There seem to be two problems that would be nice to address:

* There's no test coverage for the use modification from specialization -> primary template
* The duplicated forward-decl if both are available is not great

kimgr:
For forward-declarations inserted by IWYU, we try to deduplicate before
adding them. However, when there are multiple redeclarations, we would
sometimes end up with two different redecls in two different
uses (particularly when the target decl has been updated due to IWYU
policy).

The original repro case was:

    // tpl_fwd_decl.h
    template <int> class Tpl;

    // src.cpp
    #include "tpl_fwd_decl.h"

    using Alias = Tpl<1>;
    template <int> class Tpl {};

which would lead to:

    template <int> class Tpl;
    template <int> class Tpl;

because the Alias produces both a fwd-decl and a full use of Tpl<1>. The
full use is first demoted to fwd-decl use, and then updated to point to
the described template (primary template in header), so there are now
two fwd-decl uses, each of their own individual decl.

Canonicalize both use-decls before comparing, so we match even if the
uses originally referred to different redecls.

Fixes #2064.

### DIFF
--- a/tests/cxx/template_specialization.cc
+++ b/tests/cxx/template_specialization.cc
@@ -14,6 +14,7 @@
 
 #include "tests/cxx/template_specialization-d1.h"
 #include "tests/cxx/template_specialization-d2.h"
+#include "tests/cxx/template_specialization-d3.h"
 #include "tests/cxx/direct.h"
 
 template<typename T> class Foo;
@@ -61,24 +62,32 @@ using FwdDeclaredTplSpecAlias = FwdDeclaredTpl<1>;
 // IWYU: FwdDeclaredTpl needs a declaration
 template <> class FwdDeclaredTpl<1> {};
 
+// IWYU: DefinedBeforeSpec needs a declaration
+using DefinedBeforeSpecAlias = DefinedBeforeSpec<1>;
+// Define the primary template; no diagnostic here.
+template <int> class DefinedBeforeSpec {};
+
 /**** IWYU_SUMMARY
 
 tests/cxx/template_specialization.cc should add these lines:
 #include "tests/cxx/indirect.h"
 #include "tests/cxx/template_specialization-i1.h"
 #include "tests/cxx/template_specialization-i2.h"
+template <int> class DefinedBeforeSpec;
 template <int> class FwdDeclaredTpl;
 
 tests/cxx/template_specialization.cc should remove these lines:
 - #include "tests/cxx/direct.h"  // lines XX-XX
 - #include "tests/cxx/template_specialization-d1.h"  // lines XX-XX
 - #include "tests/cxx/template_specialization-d2.h"  // lines XX-XX
+- #include "tests/cxx/template_specialization-d3.h"  // lines XX-XX
 - template <typename T> class Foo;  // lines XX-XX
 
 The full include-list for tests/cxx/template_specialization.cc:
 #include "tests/cxx/indirect.h"  // for IndirectClass
 #include "tests/cxx/template_specialization-i1.h"  // for Foo
 #include "tests/cxx/template_specialization-i2.h"  // for Foo
+template <int> class DefinedBeforeSpec;
 template <int> class FwdDeclaredTpl;
 template <typename T> struct Specialized;  // lines XX-XX+1
 
--- a/iwyu_output.h
+++ b/iwyu_output.h
@@ -196,7 +196,8 @@ class OneIncludeOrForwardDeclareLine {
   }
 
   bool matches(const clang::NamedDecl* decl) const {
-    return !IsIncludeLine() && (fwd_decl_ == decl);
+    return !IsIncludeLine() &&
+           (fwd_decl_->getCanonicalDecl() == decl->getCanonicalDecl());
   }
 
   void set_present() { is_present_ = true; }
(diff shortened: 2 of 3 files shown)


bolshakov-a:
> One decl was processed, but the symbol from the original decl was presented. Fixing this gives some spurious changes in the test suite, and I'm not sure if they're good or bad.

I suspect the original idea was that `reset_decl` should be used only to choose some other redeclaration of the same entity, so it should have the same name, and rebuilding the names is redundant, in theory.

The test changes are definitely good. It is the primary template (forward-)declarations that is required in those cases,  and not specialization declarations.

Except there is still no test coverage of the original issue, AFAIU, this LGTM, thanks!

kimgr:
Cut the `reset_decl` changes out of this PR, I'll post those separately. Just the canonical dedup here now, please take a look.

tests/cxx/template_specialization.cc
```diff
@@ -61,24 +62,35 @@ using FwdDeclaredTplSpecAlias = FwdDeclaredTpl<1>;
 // IWYU: FwdDeclaredTpl needs a declaration
 template <> class FwdDeclaredTpl<1> {};
 
+// IWYU: DefinedBeforeSpec needs a declaration
+// IWYU: DefinedBeforeSpec<1> needs a declaration
+using DefinedBeforeSpecAlias = DefinedBeforeSpec<1>;
+// Define the primary template; no diagnostic here.
+template <int> class DefinedBeforeSpec {};
+// Specialize the primary template defined above; no diagnostic here.
+template <> class DefinedBeforeSpec<1> {};
```
bolshakov-a: Looks like the presence of the explicit specialization definition is irrelevant for this case.

And one more case worth checking is when there is no template definition at all. In that case, my IWYU version outputs:
```console
template <int> class Tpl;

#include "tpl_fwd_decl.h"  // for Tpl
template <int> class Tpl;
```
which is also wrong.

kimgr: I added some summary logging at the top of `IwyuFileInfo::CalculateAndReportIwyuViolations`, and it appears we get both a fwd- and a full use on the alias declaration:

```c++
// src3.cpp
#include "tpl_fwd_decl.h"

using Alias = Tpl<1>;
```

```
$ IWYU_VERBOSE=8 include-what-you-use local-repros/2064/src3.cpp 
...
Preliminary uses for local-repros/2064/src3.cpp
   fwd: local-repros/2064/src3.cpp:4:15: Tpl from local-repros/2064/tpl_fwd_decl.h:2:22
  full: local-repros/2064/src3.cpp:4:15: Tpl from local-repros/2064/tpl_fwd_decl.h:2:22

--- Calculating IWYU violations for local-repros/2064/src3.cpp ---
Mapped local-repros/2064/tpl_fwd_decl.h to "tpl_fwd_decl.h" for Tpl (only candidate)
Ignoring full use of Tpl (local-repros/2064/src3.cpp:4:15): #including dfn from "tpl_fwd_decl.h"
local-repros/2064/src3.cpp:4:15: warning: Tpl needs a declaration, but does not provide or directly #include one.

local-repros/2064/src3.cpp should add these lines:
template <int> class Tpl;

local-repros/2064/src3.cpp should remove these lines:

The full include-list for local-repros/2064/src3.cpp:
#include "tpl_fwd_decl.h"  // for Tpl
template <int> class Tpl;
---
```

When it says `Ignoring full use...`, it really means `already satisfied, leaving it around`.

And then there doesn't appear to be any dedup between full uses and fwd-decl uses. So this seems to be a deeper problem.

kimgr: Removed the specialization definition and updated the commit message.

Issue #2031, opened 2026-05-19
IWYU suggests unnecessary include in header


nick-potenski:
Given the following example:
```c++
// test.cpp

#include "test.hpp"

#include <iostream>
#include <string>

std::string GetGreeting(const std::string& name) {
  return "Hello, " + name + "!";
}

int main() {
  std::string greeting = GetGreeting(BOB);
  std::cout << greeting << std::endl;
  return 0;
}
``` 

```c++
// test.hpp

#ifndef TEST_HPP
#define TEST_HPP

#define BOB "bob"

#endif
``` 

IWYU suggests adding an unnecessary header to `test.hpp`:
```
$ include-what-you-use test.cpp

test.hpp should add these lines:
#include <string>  // for basic_string

test.hpp should remove these lines:

The full include-list for test.hpp:
#include <string>  // for basic_string
---

(test.cpp has correct #includes/fwd-decls)
```

kimgr:
Macro expansions are tricky. I added https://github.com/include-what-you-use/include-what-you-use/pull/2032 to repro this.

ArihantChawla:
Hi, would love to work on this!

kimgr:
@ArihantChawla You're more than welcome to look into this!

Just to give you some strings to start pulling on:

* The use location for a macro expansion is computed in iwyu.cc, `GetCanonicalUseLocation` -- it uses some heuristics to try and figure out if the definition or the expansion location should own the use. This is known to be incomplete, and I would start debugging here.
* The pair test.cpp/test.hpp, by virtue of sharing the same filename stem (`test`), are analyzed as a unit, so IWYU will change its analysis to avoid duplicates between the two. test.hpp is called the "associated header" in this model. You may want to try renaming test.hpp to something else, e.g. `macro.hpp` and see if the associated header policies have any negative impact here.
* There's quite a bit of logic in iwyu_output.cc which uses `GetSpellingLoc` and `GetInstantiationLoc` in various ways; this could affect the use location after it's reported from iwyu.cc. I believe `GetSpellingLoc` is supposed to return the location inside the macro definition, and `GetInstantiationLoc` the location at expansion, but I'm not sure they work 100%

Those are my primary suspects off the top of my head, don't hesitate to ping back if you get stuck.

ArihantChawla:
Hi, thanks for the quick reply!

[...]

Issue #1934, opened 2026-02-26
IWYU suggests forward declaring enumerations inside a class in the same file that defines the class


mwoehlke-kitware:
Consider this header:
```
class Foo {
public:
  enum Bar : int { ... };
  ...
};
```

IWYU (0.24, clang 20.1.8) reports:
```
foo.h should add these lines:
enum Foo::Bar : int;
```

Not only does this appear pointless, it isn't even valid C++; AFAIK it isn't actually *possible* to add the forward declaration IWYU wants.

bolshakov-a:
I cannot reproduce this. Could you provide some more self-contained example?

mwoehlke-kitware:
I can reproduce with the following command line:
```bash
/usr/bin/include-what-you-use -w --driver-mode=g++ /usr/lib64/ccache/c++ -std=c++17 -c iwyu1934.cxx
```
### `iwyu1934.h`:
```c++
#pragma once

#define FOR_EACH_ANIMAL(ACTION) \
  ACTION(NONE, CAT)             \
  ACTION(CAT, LION)             \
  ACTION(CAT, TIGER)

class iwyu1934
{
public:
  enum Animal : unsigned
  {
    NONE,
#define ANIMAL_ENUM(P, A) A,
    FOR_EACH_ANIMAL(ANIMAL_ENUM)
#undef DIAGNOSTIC_ENUM
    COUNT
  };
  constexpr static unsigned Count = static_cast<unsigned>(COUNT);

  struct Relation
  {
    Animal Parent;
  };

  // The problem seems to be some combination of this member and the macros
  constexpr static Animal
    Relations[Count] = {
      NONE, // NONE
#define RELATION(P, A) P,
      FOR_EACH_ANIMAL(RELATION)
#undef DIAGNOSTIC_CATEGORY_INFO
    };
};
```
### `iwyu1934.cxx`:
```c++
#include "iwyu1934.h"
```

bolshakov-a:
Thanks! Looks like there is a misattribution of the full use of the `Animal` enumeration (more precisely, its enumerators) inside `Relations` variable initializer to the `FOR_EACH_ANIMAL` macro spelling location  instead of the expansion location. In fact, IWYU cleans up fwd-decl uses of nested declarations:
https://github.com/include-what-you-use/include-what-you-use/blob/c88320f1303b0918b1b53c22bb2aa06c6c587e59/iwyu_output.cc#L1062
[...]

PR #2070, opened 2026-07-06
Fix implicit construction macro attribution

ArihantChawla:
* Fix for issue #2031 
* Added a function MacroBodyHasNoIdentifiers which checks for explicit declarations of type in a macro definition, if none are found it attributes ownership to expansion side


iwyu.cc
```diff
@@ -315,6 +322,41 @@ bool CanIgnoreLocation(SourceLocation loc) {
...
+// context, not something the macro author wrote.
+//
+// This deliberately stops short of checking whether the macro mentions
+// the *specific* decl being used: a macro can legitimately be
+// responsible for a type without ever spelling its name, e.g. by
+// dereferencing a pointer ('#define GET(n) (*(table + n))') or calling
+// one of its members ('#define CALL (obj).Method()') -- both need the
+// pointee/object's complete type, but neither macro spells that type's
+// name anywhere. Once a macro references *any* identifier, we fall back
+// to the existing (spelling-biased) heuristics rather than guessing.
```
bolshakov-a, line 342: Does this work correctly with user-defined literals? Like this
```cpp
#define TEST_STRING_VIEW "test"sv
```

bolshakov-a (commented):
I don't intend to block this, but this looks like an ad-hoc solution which obviously doesn't work in the common case (when there is e.g. a function call returning `const char*` inside a macro).

I still think that the present (on the master branch) heuristic [should be inverted](https://github.com/include-what-you-use/include-what-you-use/issues/1934#issuecomment-3976433021) because forward-declarations look like just a nonsense for macros. It is hard to imagine a case when they are really useful, so checking for their presence is strange.

kimgr:
Hey, thanks for the patch!

I'm also not sure this is a viable solution, but if you want to use it as a driver to learn more about the IWYU PR process, we can discuss around the formalities:

- The failing testcase is a trap: I try to add a "bug test" for all bug reports that are reproducible (see also the `reproducible` issue tag). A bug test is supposed to show what behavior the user wants, and is marked with `IWYU_XFAIL`. That way it succeeds as long as the bug exists, and if the bug is accidentally fixed (200+ open bugs, easy to miss one) it starts failing. And that's exactly what happened here. Read more here [about the bugs suite](https://github.com/include-what-you-use/include-what-you-use/tree/master/tests/bugs#bugs-suite) and the process for moving past these test failures. Which leads us to the next point...
- Fixes like this need to come with a reduced fully portable regression test (independent of any library code, including standard library). These are not the same as the bug test with `IWYU_XFAIL` removed, because bug tests are usually written in terms of success scenarios, and they often depend on the standard library. There is [documentation for the regression test framework](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#regression-tests), which describes good approaches. I usually borrow from an existing testcase, but do note that many of them don't comply with the guidelines, so be careful :dragon: 
- We use `clang-format` for new code to ensure a consistent format. Please do format your patches, but please don't format entire files as part of your submission (see [git-format](https://github.com/kimgr/git-format) for a Git-oriented wrapper.)
- We usually use a [rebase workflow for PRs](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#git-workflow) -- please keep the PR branch as you would like to see it on `master`. That's how I will review it -- does the commit message make sense? Are the commits (if more than one) reasonably partitioned? Does every commit reasonably work end-to-end?

Anything you learn in that process will be useful for the next patch, if you want to engage more with the IWYU project.

As for the patch itself, it's kind of an interesting approach. Even if it doesn't solve the entire issue, I think it might be a nice heuristic that macros without identifiers cannot own uses. I'm trying to think of counter-examples, and operator overloads spring to mind. Does something like this behave as expected:

```c++
// operator.h
bool operator==(const Foo& lhs, const Foo& rhs);

// foo.h
struct Foo {
  int val;
};

// macro.h
#define EQ ==

// t.cc
#include "macro.h"
#include "foo.h"
#include "operator.h"

void test(const Foo& a, const Foo& b) {
  if (a EQ b) {
     // cake!
  }
}
```
?

iwyu.cc
ArihantChawla: I'm not sure if sv was intended to be a part of this or typo, but I have tested string literals so "test" will work

kimgr: @ArihantChawla `sv` is a user-defined literal, see https://cppreference.com/cpp/string/basic_string_view/operator""sv

ArihantChawla:
> ```c++
> // operator.h
> bool operator==(const Foo& lhs, const Foo& rhs);
> 
> // foo.h
> struct Foo {
>   int val;
> };
> 
> // macro.h
> #define EQ ==
> 
> // t.cc
> #include "macro.h"
> #include "foo.h"
> #include "operator.h"
> 
> void test(const Foo& a, const Foo& b) {
>   if (a EQ b) {
>      // cake!
>   }
> }
> ```
> 
> ?
-------------
Giving no extra includes needed, the result is the same even if the name of macro.h is changed to t.h

 ./bin/include-what-you-use ../examples/t.cc 

(../examples/t.cc has correct #includes/fwd-decls)
-------------

ArihantChawla:
> I don't intend to block this, but this looks like an ad-hoc solution which obviously doesn't work in the common case (when there is e.g. a function call returning `const char*` inside a macro).

char wouldn't have any includes associated with it so not sure what would be the need for checking there, but if the macro had std::string as the function return call then that would need an include for std::string in the macro file as per my understanding of the existing code. Please correct me if I am wrong in this assumption.

> I still think that the present (on the master branch) heuristic [should be inverted](https://github.com/include-what-you-use/include-what-you-use/issues/1934#issuecomment-3976433021) because forward-declarations look like just a nonsense for macros. It is hard to imagine a case when they are really useful, so checking for their presence is strange.

This is an interesting point and something I was also pondering when discussing the issue with @kimgr 
This might be very basic but one thing to note is that the original issue was reported in associated header file (same name as the source file), I was not able to reproduce it when the name of the included header and source files was different.

iwyu.cc
bolshakov-a: Yes, in such cases, including `<string_view>` in the macro-defining header may still make sense, despite there are no identifiers here. (`sv` is not an identifier, AFAIU, but I can be wrong.)

bolshakov-a:
> char wouldn't have any includes associated with it so not sure what would be the need for checking there, but if the macro had std::string as the function return call then that would need an include for std::string in the macro file as per my understanding of the existing code. Please correct me if I am wrong in this assumption.

I mean the cases like this ([`ctermid`](https://pubs.opengroup.org/onlinepubs/009695399/functions/ctermid.html) is a POSIX function returning `char*`):
```cpp
// test.h
#include <cstdio>

#define CTERMID ctermid(NULL)

// test.cpp
#include "test.h"
#include <string>

void Fn(const std::string&);

int main() {
    Fn(CTERMID);
}
```
I suspect that the output on your branch is the same as on master (i.e. wrong):
```console
#include <string>  // for basic_string

#include <cstdio>  // for NULL, ctermid
#include <string>  // for basic_string
---

```
> This might be very basic but one thing to note is that the original issue was reported in associated header file (same name as the source file), I was not able to reproduce it when the name of the included header and source files was different.

I'm able to reproduce after renaming `test.hpp` to `macro.hpp` by specifying `-Xiwyu --check_also=macro.hpp` in the command line.

iwyu.cc
```diff
@@ -1415,6 +1465,13 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
       VERRS(5) << "Found a hint decl in macro definition file\n";
       use_loc = expansion_loc;
       side = "expansion";
+    } else if (MacroBodyHasNoIdentifiers(use_loc,
+                                        compiler()->getPreprocessor())) {
```
kimgr, line 1469: Please clang-format this section so it aligns.

iwyu.cc
```diff
@@ -1415,6 +1465,13 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+      VERRS(5) << "Macro body has no identifiers at all; can't be a "
```
kimgr, line 1470: The "use expansion location instead" is redundant, it will be printed on line 1423/1480 based on `side`. I think you can reduce this to
```
VERRS(5) << "No identifiers in macro body\n";
```

iwyu.cc
```diff
@@ -315,6 +322,41 @@ bool CanIgnoreLocation(SourceLocation loc) {
...
+//
+// This deliberately stops short of checking whether the macro mentions
+// the *specific* decl being used: a macro can legitimately be
+// responsible for a type without ever spelling its name, e.g. by
+// dereferencing a pointer ('#define GET(n) (*(table + n))') or calling
+// one of its members ('#define CALL (obj).Method()') -- both need the
+// pointee/object's complete type, but neither macro spells that type's
+// name anywhere. Once a macro references *any* identifier, we fall back
+// to the existing (spelling-biased) heuristics rather than guessing.
+bool MacroBodyHasNoIdentifiers(SourceLocation use_loc, Preprocessor& pp) {
```
kimgr, line 343: Can you move this to `iwyu_lexer_utils{.h,.cpp}`?

kimgr:
@ArihantChawla Are you still (interested in) working on this? I left a few stray comments, but primarily this is missing tests. Please add some.

kimgr:
Oh, and also, rebase on latest master to get a working CI build.

PR #2071, opened 2026-07-07
[clang compat] Use new getExplicitInstantiationInfo()

kimgr:
https://github.com/llvm/llvm-project/commit/ff11bbce9e87ab4ef59a7
removed the getExternKeywordLoc wrapper, and instead added a
getExplicitInstantiationInfo function to get the whole instantiation
info.

The patch in question also changes the AST so that function call exprs
in template instantiations no longer always point to the definition of
the callee. Instead, it seems they point at the last seen
declaration. It seems we assumed the definition in
InstantiatedTemplateVisitor::HandleFunctionCall, so explicitly get the
definition before handling it.

Migrate our only use to the new API.

Note: the upstream patch is already reverted on LLVM mainline, and under
revision to be re-landed. But the packages at apt.llvm.org have not
caught up with that change yet, so I'll merge this now to get our builds
working again, and then revert it when the Ubuntu packages are up to
date again.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -4035,6 +4035,10 @@ class InstantiatedTemplateVisitor
                           const Expr* calling_expr) {
     if (const Type* resugared_type = ResugarType(parent_type))
       parent_type = resugared_type;
+    if (callee) {
+      if (FunctionDecl* defn = callee->getDefinition())
+        callee = defn;
+    }
     if (!Base::HandleFunctionCall(callee, parent_type, calling_expr))
       return false;
     return TraverseFunctionIfInstantiatedTpl(callee, parent_type, calling_expr);
--- a/iwyu_ast_util.cc
+++ b/iwyu_ast_util.cc
@@ -98,6 +98,7 @@ using clang::EnumDecl;
 using clang::EnumType;
 using clang::ExplicitCastExpr;
 using clang::ExplicitInstantiationDecl;
+using clang::ExplicitInstantiationInfo;
 using clang::Expr;
 using clang::ExprResult;
 using clang::ExprValueKind;
@@ -1324,9 +1325,14 @@ bool IsExplicitInstantiationDefinitionAsWritten(
     const ClassTemplateSpecializationDecl* decl) {
   // When swithing instantiation declaration to definition, clang preserves
   // the 'extern' keyword location info.
-  return decl->getSpecializationKind() ==
-             clang::TSK_ExplicitInstantiationDefinition &&
-         decl->getExternKeywordLoc().isInvalid();
+  if (decl->getSpecializationKind() ==
+      clang::TSK_ExplicitInstantiationDefinition) {
+    if (const ExplicitInstantiationInfo* info =
+            decl->getExplicitInstantiationInfo()) {
+      return info->ExternKeywordLoc.isInvalid();
+    }
+  }
+  return false;
 }
 
 bool IsInInlineNamespace(const Decl* decl) {


kimgr:
```
/home/runner/work/include-what-you-use/include-what-you-use/iwyu_string_util.h should add these lines:
#include <algorithm>             // for max
```
Noooo, where did that come from???

kimgr:
I can't reproduce locally. Time to go to bed, clearly.

kimgr:
I did manage to reproduce. IWYU throws this when analyzing `iwyu_string_util.h` via `iwyu_regex.cc`.

This line triggers the use, after ~2K log lines of template analysis of `std::vector::emplace`: https://github.com/include-what-you-use/include-what-you-use/blob/13371e0d60e90928b1718516dcce8dc7e3a0c684/iwyu_string_util.h#L132.

I have not been able to reproduce it in a smaller context yet.

So I think in the course of fixing template instantiation args in https://github.com/llvm/llvm-project/commit/ff11bbce9e87ab4ef59a7, they also managed to change some aspect of the AST, so our analysis gets lost. Will try to find something deeper.

bolshakov-a:
I can reproduce with the following:
```cpp
#include <regex>
#include <vector>

void Fn() {
    std::vector<std::string> retval;
    std::string str;
    retval.emplace_back(str);
}
```
If the order of include directives is reversed, the bug disappears. It looks like `max` location misattribution.

kimgr:
@bolshakov-a Thanks! That brings the verbose log down to ~2MB, and I should be able to diff before/after to get a clue what happened. I see there's a small change in traversal here: https://github.com/llvm/llvm-project/blob/ff11bbce9e87ab4ef59a789cff1c9b9d771adc1e/clang/include/clang/AST/RecursiveASTVisitor.h#L2230 (can't link to the diff, because GitHub seems to have given up on larger diffs). I wonder if that throws IWYU traversal off somehow.

bolshakov-a:
Further reduced to reproduce without the standard library:
```cpp
// my_max.h
#pragma once

template <typename T>
T my_max(T t1, T t2) {
    return t1 > t2 ? t1 : t2;
}

// my_max_decl.h
#pragma once

template <typename T>
T my_max(T t1, T t2);

// fn_instantiating_my_max.h
#include "my_max.h"

inline void FnInstantiatingMyMax() {
    my_max(1, 2);
}

// tpl_fn.h
#include "my_max.h"

template <typename T>
void TplFn() {
    my_max(T{}, T{});
}

// other_hdr.h
#include "fn_instantiating_my_max.h"
#include "my_max_decl.h"

// main.cpp
#include "other_hdr.h"
#include "tpl_fn.h"

int main() {
    TplFn<int>();
}
```
The order of header inclusions is important.

After https://github.com/llvm/llvm-project/commit/ff11bbce9e87ab4ef59a7, `my_max` call expression inside the instantiated `TplFn` specialization points to the redeclaration from `my_max_decl.h` and not to the definition from `my_max.h`. Because `tpl_fn.h` header doesn't provide `my_max_decl.h`, `my_max` function template is considered as not provided by `TplFn` and hence is suggested for inclusion into `main.cpp`.

I don't still understand the exact reason of this change in clang behavior, but it is hardly related to the `RecursiveASTVisitor` changes. Getting to the definition inside `HandleFunctionCall` could help.

kimgr:
Finally rebuilt and reran everything. Here's the diff between IWYU from this PR built against the apt.llvm.org version (70646dd3eda3e64fc94b44d7403f946cd7455b92) and IWYU master built against latest llvm main (e62c0cb076fbcb70752530571726944648a48829), where the clang patch in question has been reverted:

```diff
diff --git a/spill_max_main.txt b/spill_max_patch.txt
index 806ab401..8bc26add 100644
--- a/spill_max_main.txt
+++ b/spill_max_patch.txt
@@ -23991,7 +23991,7 @@ Marked full-info use of decl retval (from local-repros/spill_max.cc:5:30) at loc
 local-repros/spill_max.cc:7:25: (4) [ DeclRefExpr ] DeclRefExpr  'std::string':'class std::basic_string<char>' lvalue Var  'str' 'std::string':'class std::basic_string<char>'
 
 Marked full-info use of decl str (from local-repros/spill_max.cc:6:17) at local-repros/spill_max.cc:7:25
-local-repros/spill_max.cc:7:5: (3) [ FunctionCall ] template<> reference vector<std::basic_string<char>, std::allocator<std::basic_string<char>>>::emplace_back<<std::basic_string<char> &>>(std::basic_string<char> &__args)
+local-repros/spill_max.cc:7:5: (3) [ FunctionCall ] template<> reference emplace_back<<std::basic_string<char> &>>(std::basic_string<char> &__args)
 Adding an implicit tpl-function type of interest: class std::basic_string<char> &
 Adding a type-components of interest: class std::basic_string<char>
 Remapping template arg of interest: class std::basic_string<char> -> std::string
@@ -24023,19 +24023,7 @@ Adding a type-components of interest: char
 Adding a template-class type of interest: class std::basic_string<char> -> std::string
 Adding a template-class default type of interest: class std::allocator<class std::basic_string<char> >
 Adding a type-components of interest: char
-/usr/include/c++/14/bits/vector.tcc:112:7: (4 in tpl) [ CXXMethodDecl ] template<> reference vector<std::basic_string<char>, std::allocator<std::basic_string<char>>>::emplace_back<<std::basic_string<char> &>>(std::basic_string<char> &__args)
-/usr/include/c++/14/bits/vector.tcc:111:7: (5 in tpl) [ NestedNameSpecifier ] vector<std::basic_string<char>, std::allocator<std::basic_string<char> > >::
-/usr/include/c++/14/bits/vector.tcc:111:7: (6 in tpl) [ TemplateSpecializationTypeLoc ] vector<class std::basic_string<char>, class std::allocator<class std::basic_string<char> > >
-/usr/include/c++/14/bits/vector.tcc:111:7: (7 in tpl) [ TemplateName ] std::vector
-/usr/include/c++/14/bits/vector.tcc:111:14: (7 in tpl) [ TemplateArgumentLoc ] <class std::basic_string<char> >
-/usr/include/c++/14/bits/vector.tcc:111:14: (8 in tpl, fwd decl) [ SubstTemplateTypeParmTypeLoc ] class std::basic_string<char>
-AnalyzeTemplateTypeParmUse: type = class std::basic_string<char>, actual_type = std::string
-/usr/include/c++/14/bits/vector.tcc:111:14: (9 in tpl, fwd decl) [ RecordType ] class std::basic_string<char>
-/usr/include/c++/14/bits/vector.tcc:111:19: (7 in tpl) [ TemplateArgumentLoc ] <class std::allocator<class std::basic_string<char> > >
-/usr/include/c++/14/bits/vector.tcc:111:19: (8 in tpl, fwd decl) [ SubstTemplateTypeParmTypeLoc ] class std::allocator<class std::basic_string<char> >
-AnalyzeTemplateTypeParmUse: type = class std::allocator<class std::basic_string<char> >, actual_type = class std::allocator<class std::basic_string<char> >
-/usr/include/c++/14/bits/vector.tcc:111:19: (9 in tpl, fwd decl) [ RecordType ] class std::allocator<class std::basic_string<char> >
-(Replaying full-use information from the cache for std::vector)
+/usr/include/c++/14/bits/vector.tcc:112:7: (4 in tpl) [ CXXMethodDecl ] template<> reference emplace_back<<std::basic_string<char> &>>(std::basic_string<char> &__args)
 /usr/include/c++/14/bits/stl_vector.h:1306:2: (5 in tpl) [ FunctionProtoTypeLoc ] reference (class std::basic_string<char> &)
 /usr/include/c++/14/bits/stl_vector.h:1306:2: (6 in tpl) [ TypedefTypeLoc ] reference
 /usr/include/c++/14/bits/stl_vector.h:1310:26: (6 in tpl) [ ParmVarDecl ] std::basic_string<char> &__args
@@ -25218,23 +25206,8 @@ Marked full-info use of decl std::basic_string (from /usr/include/c++/14/bits/ba
 
 /usr/include/c++/14/bits/basic_string.h:557:8: (20 in tpl) [ RecordTypeLoc ] std::forward_iterator_tag
 /usr/include/c++/14/bits/basic_string.h:557:8: (21 in tpl) [ NestedNameSpecifier ] std::
-/usr/include/c++/14/bits/basic_string.h:556:2: (18 in tpl) [ FunctionCall ] template<> void basic_string<char, std::char_traits<char>, std::allocator<char>>::_M_construct<char *>(char *__beg, char *__end, std::forward_iterator_tag)
-/usr/include/c++/14/bits/basic_string.tcc:222:7: (19 in tpl) [ CXXMethodDecl ] template<> void basic_string<char, std::char_traits<char>, std::allocator<char>>::_M_construct<char *>(char *__beg, char *__end, std::forward_iterator_tag)
-/usr/include/c++/14/bits/basic_string.tcc:221:7: (20 in tpl) [ NestedNameSpecifier ] basic_string<char, std::char_traits<char>, std::allocator<char> >::
-/usr/include/c++/14/bits/basic_string.tcc:221:7: (21 in tpl) [ TemplateSpecializationTypeLoc ] basic_string<char, struct std::char_traits<char>, class std::allocator<char> >
-/usr/include/c++/14/bits/basic_string.tcc:221:7: (22 in tpl) [ TemplateName ] std::basic_string
-/usr/include/c++/14/bits/basic_string.tcc:221:20: (22 in tpl) [ TemplateArgumentLoc ] <char>
-/usr/include/c++/14/bits/basic_string.tcc:221:20: (23 in tpl, fwd decl) [ SubstTemplateTypeParmTypeLoc ] char
-AnalyzeTemplateTypeParmUse: type = char, actual_type = char
-/usr/include/c++/14/bits/basic_string.tcc:221:20: (24 in tpl, fwd decl) [ BuiltinType ] char
-/usr/include/c++/14/bits/basic_string.tcc:221:28: (22 in tpl) [ TemplateArgumentLoc ] <struct std::char_traits<char> >
-/usr/include/c++/14/bits/basic_string.tcc:221:28: (23 in tpl, fwd decl) [ SubstTemplateTypeParmTypeLoc ] struct std::char_traits<char>
-AnalyzeTemplateTypeParmUse: type = struct std::char_traits<char>, actual_type = struct std::char_traits<char>
-/usr/include/c++/14/bits/basic_string.tcc:221:28: (24 in tpl, fwd decl) [ RecordType ] struct std::char_traits<char>
-/usr/include/c++/14/bits/basic_string.tcc:221:37: (22 in tpl) [ TemplateArgumentLoc ] <class std::allocator<char> >
-/usr/include/c++/14/bits/basic_string.tcc:221:37: (23 in tpl, fwd decl) [ SubstTemplateTypeParmTypeLoc ] class std::allocator<char>
-AnalyzeTemplateTypeParmUse: type = class std::allocator<char>, actual_type = class std::allocator<char>
-/usr/include/c++/14/bits/basic_string.tcc:221:37: (24 in tpl, fwd decl) [ RecordType ] class std::allocator<char>
+/usr/include/c++/14/bits/basic_string.h:556:2: (18 in tpl) [ FunctionCall ] template<> void _M_construct<char *>(char *__beg, char *__end, std::forward_iterator_tag)
+/usr/include/c++/14/bits/basic_string.tcc:222:7: (19 in tpl) [ CXXMethodDecl ] template<> void _M_construct<char *>(char *__beg, char *__end, std::forward_iterator_tag)
 /usr/include/c++/14/bits/basic_string.h:331:9: (20 in tpl) [ FunctionProtoTypeLoc ] void (char *, char *, std::forward_iterator_tag)
 /usr/include/c++/14/bits/basic_string.h:331:9: (21 in tpl) [ BuiltinTypeLoc ] void
 /usr/include/c++/14/bits/basic_string.h:332:35: (21 in tpl) [ ParmVarDecl ] char *__beg
@@ -26502,19 +26475,8 @@ AnalyzeTemplateTypeParmUse: type = class std::basic_string<char>, actual_type =
 /usr/include/c++/14/bits/vector.tcc:123:42: (9 in tpl) [ DeclRefExpr ] DeclRefExpr  'class std::basic_string<char>' lvalue ParmVar  '__args' 'class std::basic_string<char> &'
 
 /usr/include/c++/14/bits/vector.tcc:123:22: (8 in tpl) [ FunctionCall ] template<> constexpr std::basic_string<char> &forward<std::basic_string<char> &>(typename std::remove_reference<std::basic_string<char> &>::type &__t) noexcept
-/usr/include/c++/14/bits/vector.tcc:123:4: (7 in tpl) [ FunctionCall ] template<> void vector<std::basic_string<char>, std::allocator<std::basic_string<char>>>::_M_realloc_append<<std::basic_string<char> &>>(std::basic_string<char> &__args)
-/usr/include/c++/14/bits/vector.tcc:582:7: (8 in tpl) [ CXXMethodDecl ] template<> void vector<std::basic_string<char>, std::allocator<std::basic_string<char>>>::_M_realloc_append<<std::basic_string<char> &>>(std::basic_string<char> &__args)
-/usr/include/c++/14/bits/vector.tcc:581:7: (9 in tpl) [ NestedNameSpecifier ] vector<std::basic_string<char>, std::allocator<std::basic_string<char> > >::
-/usr/include/c++/14/bits/vector.tcc:581:7: (10 in tpl) [ TemplateSpecializationTypeLoc ] vector<class std::basic_string<char>, class std::allocator<class std::basic_string<char> > >
-/usr/include/c++/14/bits/vector.tcc:581:7: (11 in tpl) [ TemplateName ] std::vector
-/usr/include/c++/14/bits/vector.tcc:581:14: (11 in tpl) [ TemplateArgumentLoc ] <class std::basic_string<char> >
-/usr/include/c++/14/bits/vector.tcc:581:14: (12 in tpl, fwd decl) [ SubstTemplateTypeParmTypeLoc ] class std::basic_string<char>
-AnalyzeTemplateTypeParmUse: type = class std::basic_string<char>, actual_type = std::string
-/usr/include/c++/14/bits/vector.tcc:581:14: (13 in tpl, fwd decl) [ RecordType ] class std::basic_string<char>
-/usr/include/c++/14/bits/vector.tcc:581:19: (11 in tpl) [ TemplateArgumentLoc ] <class std::allocator<class std::basic_string<char> > >
-/usr/include/c++/14/bits/vector.tcc:581:19: (12 in tpl, fwd decl) [ SubstTemplateTypeParmTypeLoc ] class std::allocator<class std::basic_string<char> >
-AnalyzeTemplateTypeParmUse: type = class std::allocator<class std::basic_string<char> >, actual_type = class std::allocator<class std::basic_string<char> >
-/usr/include/c++/14/bits/vector.tcc:581:19: (13 in tpl, fwd decl) [ RecordType ] class std::allocator<class std::basic_string<char> >
+/usr/include/c++/14/bits/vector.tcc:123:4: (7 in tpl) [ FunctionCall ] template<> void _M_realloc_append<<std::basic_string<char> &>>(std::basic_string<char> &__args)
+/usr/include/c++/14/bits/vector.tcc:582:7: (8 in tpl) [ CXXMethodDecl ] template<> void _M_realloc_append<<std::basic_string<char> &>>(std::basic_string<char> &__args)
 /usr/include/c++/14/bits/stl_vector.h:1882:2: (9 in tpl) [ FunctionProtoTypeLoc ] void (class std::basic_string<char> &)
 /usr/include/c++/14/bits/stl_vector.h:1882:2: (10 in tpl) [ BuiltinTypeLoc ] void
 /usr/include/c++/14/bits/stl_vector.h:1883:31: (10 in tpl) [ ParmVarDecl ] std::basic_string<char> &__args
@@ -27264,66 +27226,19 @@ Marked full-info use of decl std::basic_string (from /usr/include/c++/14/bits/ba
 `-NestedNameSpecifier Namespace  'std'
 
 /usr/include/c++/14/bits/stl_vector.h:1911:36: (24 in tpl) [ NestedNameSpecifier ] std::
-/usr/include/c++/14/bits/stl_algobase.h:257:5: (24 in tpl) [ FunctionDecl ] template<> inline constexpr const unsigned long &max<unsigned long>(const unsigned long &__a, const unsigned long &__b)
-/usr/include/c++/14/bits/stl_algobase.h:256:18: (25 in tpl) [ FunctionProtoTypeLoc ] const unsigned long &(const unsigned long &, const unsigned long &)
-/usr/include/c++/14/bits/stl_algobase.h:256:18: (26 in tpl) [ LValueReferenceTypeLoc ] const unsigned long &
-/usr/include/c++/14/bits/stl_algobase.h:256:18: (27 in tpl) [ SubstTemplateTypeParmTypeLoc ] unsigned long
-/usr/include/c++/14/bits/stl_algobase.h:256:18: (28 in tpl) [ BuiltinType ] unsigned long
-/usr/include/c++/14/bits/stl_algobase.h:257:20: (26 in tpl) [ ParmVarDecl ] const unsigned long &__a
-/usr/include/c++/14/bits/stl_algobase.h:257:15: (27 in tpl) [ LValueReferenceTypeLoc ] const unsigned long &
-/usr/include/c++/14/bits/stl_algobase.h:257:15: (28 in tpl) [ SubstTemplateTypeParmTypeLoc ] unsigned long
-/usr/include/c++/14/bits/stl_algobase.h:257:15: (29 in tpl) [ BuiltinType ] unsigned long
-/usr/include/c++/14/bits/stl_algobase.h:257:36: (26 in tpl) [ ParmVarDecl ] const unsigned long &__b
-/usr/include/c++/14/bits/stl_algobase.h:257:31: (27 in tpl) [ LValueReferenceTypeLoc ] const unsigned long &
-/usr/include/c++/14/bits/stl_algobase.h:257:31: (28 in tpl) [ SubstTemplateTypeParmTypeLoc ] unsigned long
-/usr/include/c++/14/bits/stl_algobase.h:257:31: (29 in tpl) [ BuiltinType ] unsigned long
-/usr/include/c++/14/bits/stl_algobase.h:258:5: (25 in tpl) [ CompoundStmt ] CompoundStmt 
-|-IfStmt 
-| |-BinaryOperator  '_Bool' '<'
-| | |-ImplicitCastExpr  'unsigned long' <LValueToRValue>
-| | | `-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__a' 'const unsigned long &'
-| | `-ImplicitCastExpr  'unsigned long' <LValueToRValue>
-| |   `-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__b' 'const unsigned long &'
-| `-ReturnStmt 
-|   `-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__b' 'const unsigned long &'
-`-ReturnStmt 
-`-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__a' 'const unsigned long &'
-
-/usr/include/c++/14/bits/stl_algobase.h:262:7: (26 in tpl) [ IfStmt ] IfStmt 
-|-BinaryOperator  '_Bool' '<'
-| |-ImplicitCastExpr  'unsigned long' <LValueToRValue>
-| | `-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__a' 'const unsigned long &'
-| `-ImplicitCastExpr  'unsigned long' <LValueToRValue>
-|   `-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__b' 'const unsigned long &'
-`-ReturnStmt 
-`-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__b' 'const unsigned long &'
-
-/usr/include/c++/14/bits/stl_algobase.h:262:15: (27 in tpl) [ BinaryOperator ] BinaryOperator  '_Bool' '<'
-|-ImplicitCastExpr  'unsigned long' <LValueToRValue>
-| `-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__a' 'const unsigned long &'
-`-ImplicitCastExpr  'unsigned long' <LValueToRValue>
-`-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__b' 'const unsigned long &'
-
-/usr/include/c++/14/bits/stl_algobase.h:262:11: (28 in tpl) [ ImplicitCastExpr ] ImplicitCastExpr  'unsigned long' <LValueToRValue>
-`-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__a' 'const unsigned long &'
-
-/usr/include/c++/14/bits/stl_algobase.h:262:11: (29 in tpl) [ DeclRefExpr ] DeclRefExpr  'const unsigned long' lvalue ParmVar  '__a' 'const unsigned long &'
-
-/usr/include/c++/14/bits/stl_algobase.h:262:17: (28 in tpl) [ ImplicitCastExpr ] ImplicitCastExpr  'unsigned long' <LValueToRValue>
-`-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__b' 'const unsigned long &'
-
-/usr/include/c++/14/bits/stl_algobase.h:262:17: (29 in tpl) [ DeclRefExpr ] DeclRefExpr  'const unsigned long' lvalue ParmVar  '__b' 'const unsigned long &'
-
-/usr/include/c++/14/bits/stl_algobase.h:263:2: (27 in tpl) [ ReturnStmt ] ReturnStmt 
-`-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__b' 'const unsigned long &'
-
-/usr/include/c++/14/bits/stl_algobase.h:263:9: (28 in tpl) [ DeclRefExpr ] DeclRefExpr  'const unsigned long' lvalue ParmVar  '__b' 'const unsigned long &'
-
-/usr/include/c++/14/bits/stl_algobase.h:264:7: (26 in tpl) [ ReturnStmt ] ReturnStmt 
-`-DeclRefExpr  'const unsigned long' lvalue ParmVar  '__a' 'const unsigned long &'
-
-/usr/include/c++/14/bits/stl_algobase.h:264:14: (27 in tpl) [ DeclRefExpr ] DeclRefExpr  'const unsigned long' lvalue ParmVar  '__a' 'const unsigned long &'
-
+/usr/include/c++/14/bits/stl_algobase.h:257:5: (24 in tpl) [ FunctionDecl ] template<> constexpr const unsigned long &max<unsigned long>(const unsigned long &, const unsigned long &)
+/usr/include/c++/14/bits/algorithmfwd.h:402:11: (25 in tpl, fwd decl) [ FunctionProtoTypeLoc ] const unsigned long &(const unsigned long &, const unsigned long &)
+/usr/include/c++/14/bits/algorithmfwd.h:402:11: (26 in tpl, fwd decl) [ LValueReferenceTypeLoc ] const unsigned long &
+/usr/include/c++/14/bits/algorithmfwd.h:402:11: (27 in tpl, fwd decl) [ SubstTemplateTypeParmTypeLoc ] unsigned long
+/usr/include/c++/14/bits/algorithmfwd.h:402:11: (28 in tpl, fwd decl) [ BuiltinType ] unsigned long
+/usr/include/c++/14/bits/algorithmfwd.h:403:19: (26 in tpl, fwd decl) [ ParmVarDecl ] const unsigned long &
+/usr/include/c++/14/bits/algorithmfwd.h:403:15: (27 in tpl, fwd decl) [ LValueReferenceTypeLoc ] const unsigned long &
+/usr/include/c++/14/bits/algorithmfwd.h:403:15: (28 in tpl, fwd decl) [ SubstTemplateTypeParmTypeLoc ] unsigned long
+/usr/include/c++/14/bits/algorithmfwd.h:403:15: (29 in tpl, fwd decl) [ BuiltinType ] unsigned long
+/usr/include/c++/14/bits/algorithmfwd.h:403:31: (26 in tpl, fwd decl) [ ParmVarDecl ] const unsigned long &
+/usr/include/c++/14/bits/algorithmfwd.h:403:27: (27 in tpl, fwd decl) [ LValueReferenceTypeLoc ] const unsigned long &
+/usr/include/c++/14/bits/algorithmfwd.h:403:27: (28 in tpl, fwd decl) [ SubstTemplateTypeParmTypeLoc ] unsigned long
+/usr/include/c++/14/bits/algorithmfwd.h:403:27: (29 in tpl, fwd decl) [ BuiltinType ] unsigned long
 /usr/include/c++/14/bits/stl_vector.h:1911:46: (21 in tpl) [ MaterializeTemporaryExpr ] MaterializeTemporaryExpr  'const size_type':'const unsigned long' lvalue
 `-ImplicitCastExpr  'const size_type':'const unsigned long' <NoOp>
 `-CXXMemberCallExpr  'size_type':'unsigned long'
@@ -27350,7 +27265,8 @@ Marked full-info use of decl std::basic_string (from /usr/include/c++/14/bits/ba
 
 /usr/include/c++/14/bits/stl_vector.h:1911:54: (22 in tpl) [ DeclRefExpr ] DeclRefExpr  'size_type':'unsigned long' lvalue ParmVar  '__n' 'size_type':'unsigned long'
 
-/usr/include/c++/14/bits/stl_vector.h:1911:35: (20 in tpl) [ FunctionCall ] template<> inline constexpr const unsigned long &max<unsigned long>(const unsigned long &__a, const unsigned long &__b)
+/usr/include/c++/14/bits/stl_vector.h:1911:35: (20 in tpl) [ FunctionCall ] template<> constexpr const unsigned long &max<unsigned long>(const unsigned long &, const unsigned long &)
+Marked full-info use of decl std::max(const :0 &, const :0 &) (from /usr/include/c++/14/bits/stl_algobase.h:257:5) at local-repros/spill_max.cc:7:5
 /usr/include/c++/14/bits/stl_vector.h:1912:2: (15 in tpl) [ ReturnStmt ] ReturnStmt 
 `-ConditionalOperator  'size_type':'unsigned long'
 |-ParenExpr  '_Bool'
@@ -30446,6 +30362,7 @@ Mapped /usr/include/c++/14/bits/basic_string.h to <string> for std::basic_string
 Mapped /usr/include/c++/14/bits/basic_string.h to <string> for std::basic_string (only candidate)
 Mapped /usr/include/c++/14/bits/basic_string.h to <string> for std::basic_string (only candidate)
 Mapped /usr/include/c++/14/bits/basic_string.h to <string> for std::basic_string (only candidate)
+Mapped /usr/include/c++/14/bits/stl_algobase.h to <algorithm> for std::max(const :0 &, const :0 &) (only candidate)
 Mapped /usr/include/c++/14/bits/basic_string.h to <string> for std::basic_string (only candidate)
 Mapped /usr/include/c++/14/bits/basic_string.h to <string> for std::basic_string (only candidate)
 Mapped /usr/include/c++/14/bits/basic_string.h to <string> for std::basic_string (only candidate)
@@ -30482,14 +30399,17 @@ local-repros/spill_max.cc:6:5: warning: std::basic_string is defined in <string>
 local-repros/spill_max.cc:6:5: warning: std::string is defined in <string>, which isn't directly #included.
 local-repros/spill_max.cc:7:5: warning: std::basic_string is defined in <string>, which isn't directly #included (for explicit instantiation).
 local-repros/spill_max.cc:7:5: warning: std::basic_string is defined in <string>, which isn't directly #included.
+local-repros/spill_max.cc:7:5: warning: std::max(const :0 &, const :0 &) is defined in <algorithm>, which isn't directly #included.
 
 local-repros/spill_max.cc should add these lines:
-#include <string>  // for basic_string, string
+#include <algorithm>  // for max
+#include <string>     // for basic_string, string
 
 local-repros/spill_max.cc should remove these lines:
 - #include <regex>  // lines 1-1
 
 The full include-list for local-repros/spill_max.cc:
-#include <string>  // for basic_string, string
-#include <vector>  // for vector
+#include <algorithm>  // for max
+#include <string>     // for basic_string, string
+#include <vector>     // for vector
 ---
```

There's quite a bit of traversal that disappears, and the `FunctionCall` printing no longer includes nested name specifiers for the containing type.

bolshakov-a:
The traversal of `CompoundStmt` of `max` disappeared because `max` redeclaration (without body) from `bits/algorithmfwd.h` was referred to in the AST instead of the definition from `bits/stl_algobase.h`. But the other changes caused by the clang patch are really strange, not only the lack of NNS but also the double angle brackets in e.g. `emplace_back<<std::basic_string<char> &>>`.

Probably, we just have to wait, but if we should still adjust IWYU to that Clang patch, maybe it would be better not to go to the definition in `HandleFunctionCall` but to check if _any_ of the redeclarations is provided by template in `InstantiatedTemplateVisitor::ReportDeclUse` (https://github.com/include-what-you-use/include-what-you-use/blob/13371e0d60e90928b1718516dcce8dc7e3a0c684/iwyu.cc#L3978).

bolshakov-a:
> but also the double angle brackets in e.g. `emplace_back<<std::basic_string<char> &>>`.

I've overlooked that this oddity is present even without that clang patch.

kimgr:
Oh, wow. Yeah, that looks like a little bug somewhere in the DeclPrinter.

kimgr:
> The traversal of CompoundStmt of max disappeared because max redeclaration (without body) from bits/algorithmfwd.h was referred to in the AST instead of the definition from bits/stl_algobase.h. 

Ah, OK, so that could be considered an improvement -- instead of always the definition, the AST points to the actually used declaration. I'll just try forcing the definition in `HandleFunctionCall`, and see what happens.

kimgr:
```
@@ -4035,6 +4037,9 @@ class InstantiatedTemplateVisitor
                           const Expr* calling_expr) {
     if (const Type* resugared_type = ResugarType(parent_type))
       parent_type = resugared_type;
+    if (callee) {
+      callee = const_cast<FunctionDecl*>(cast<FunctionDecl>(GetDefinitionAsWritten(callee)));
+    }
       return false;
     return TraverseFunctionIfInstantiatedTpl(callee, parent_type, calling_expr);
```

Makes your minimal repro work consistent with IWYU master, but on the other hand fails 4 tests.

FWIW, there's a new revision of the Clang patch pending in: https://github.com/llvm/llvm-project/pull/208601

kimgr:
OK, using a simple `->getDefinition` instead of `GetDefinitionAsWritten` seems to work with the PR Clang, and passes all the tests.

iwyu.cc
```diff
@@ -4035,6 +4035,11 @@ class InstantiatedTemplateVisitor
                           const Expr* calling_expr) {
     if (const Type* resugared_type = ResugarType(parent_type))
       parent_type = resugared_type;
+    if (callee) {
+      if (FunctionDecl* defn = callee->getDefinition()) {
+        callee = defn;
+      }
```
bolshakov-a: Nit: excessive curly braces.

kimgr:
The upstream patch has been reverted on main, and in sitting in a draft PR here: https://github.com/llvm/llvm-project/pull/208601 waiting until the 23 branch is taken. Turn this into a draft until it lands again.

kimgr:
Unfortunately the apt.llvm.org builds are stuck on an unrelated patch conflict since Jul 7, so there are no builds for the current main: https://llvm-jenkins.debian.net/job/llvm-toolchain-noble-source/

iwyu.cc
kimgr: Thanks, done

kimgr:
@bolshakov-a 
> But the other changes caused by the clang patch are really strange, [...] the double angle brackets in
> e.g. `emplace_back<<std::basic_string<char> &>>`

This appears to be something Clang does for parameter packs.

The following minimal repro:
```c++
// t.cc
template<class T>
struct Vector {
  template<typename... Args>
  void emplace_back(Args&&...);

  template<typename U>
  void emplace_back2(T&&);
};

void t() {
  Vector<int> v;
  v.emplace_back(short(10));
  v.emplace_back2(short(10));
}
```

can be used to demonstrate with `clang -fsyntax-only -Xclang -ast-dump t.cc`. Clang's default printer does not show the `<< ... >>` format, but IWYU does. It may have something to do with the `TerseOutput` flag we set by default when printing declarations. I wonder if it can be considered a bug.

bolshakov-a:
> Clang's default printer does not show the `<< ... >>` format

It does with `-ast-print` option: https://godbolt.org/z/v6ET8Ges5.

kimgr:
> > Clang's default printer does not show the `<< ... >>` format
> 
> It does with `-ast-print` option: https://godbolt.org/z/v6ET8Ges5.

I reported https://github.com/llvm/llvm-project/issues/211737 to track this. I'm not sure if the goal is to print valid C++ code, or if it's more important to capture the fact that there is a parameter pack, we'll see what the upstream project thinks.

hebasto:
> [llvm/llvm-project@ff11bbce9e87ab4ef59a7](https://github.com/llvm/llvm-project/commit/ff11bbce9e87ab4ef59a7) removed the getExternKeywordLoc wrapper, and instead added a getExplicitInstantiationInfo function to get the whole instantiation info.

That commit has been reverted in https://github.com/llvm/llvm-project/commit/63e757382099c92c2f91d0f4acb7d74a84c09ef9.

hebasto:
> > [llvm/llvm-project@ff11bbce9e87ab4ef59a7](https://github.com/llvm/llvm-project/commit/ff11bbce9e87ab4ef59a7) removed the getExternKeywordLoc wrapper, and instead added a getExplicitInstantiationInfo function to get the whole instantiation info.
> 
> That commit has been reverted in [llvm/llvm-project@63e7573](https://github.com/llvm/llvm-project/commit/63e757382099c92c2f91d0f4acb7d74a84c09ef9).

On the `release/23.x` branch:
1. Reverted: https://github.com/llvm/llvm-project/commit/6d1566cecf40143012674376f5c9250316c29634
2. Relanded: https://github.com/llvm/llvm-project/commit/031b773b01700acf82f5977a5aa6024621b8211c
3. Reverted: https://github.com/llvm/llvm-project/commit/092f4451cecaf52c14f1be6c8741882f0b3c9de6

kimgr:
@hebasto Yes, but the snapshot LLVM apt packages are still not up to date with these changes. I tried opening https://github.com/include-what-you-use/include-what-you-use/pull/2085 to explicit use the mainline version (24). Will see if I can revert this #2071 PR as part of that change.

PR #2074, opened 2026-07-24
Add symbol mappings for float.h

### LINKED ISSUE #2073: IWYU proposes to replace float.h with __float_float.h and __float_infinity_nan.h
With the following (very simplified) reproducer.c file

```
#include <float.h>

static double epsilon(void) {
        return DBL_EPSILON;
}

static int nonumber(void) {
        return NAN;
}
```

IWYU 0.26 wants to replace `float.h` with `__float_float.h` and `__float_infinity_nan.h`

```
include-what-you-use reproducer.c

reproducer.c should add these lines:
#include <__float_float.h>         // for DBL_EPSILON
#include <__float_infinity_nan.h>  // for NAN

reproducer.c should remove these lines:
- #include <float.h>  // lines 1-1

The full include-list for reproducer.c:
#include <__float_float.h>         // for DBL_EPSILON
#include <__float_infinity_nan.h>  // for NAN
---
```

I'm not a C expert, but nevertheless I doubt that replacing `float.h` with `__float_*.h` is a good idea.

For the records: The original reproducer is [cJSON.c](https://gitlab.com/apparmor/apparmor/-/blob/master/binutils/cJSON.c) as used in AppArmor.

kimgr:
The float.h header ships as a compiler builtin.

Do a brute force manual mapping of all numeric limits symbols, as
specified in https://en.cppreference.com/c/header/float.

Add an additional mapping for INFINITY and NAN to math.h, where they
were provided before C23. This is not going to be perfect for pre-C23
code, as mappings will prefer float.h before math.h if both are included
in the same compilation unit, but it will be better than the status quo
for the rest of these symbols.

Fixes #2073


kimgr:
@bolshakov-a I've tested this with both C and C++, and `<cfloat>` is proposed for C++ still.

@cboltz If you can try this out and let me know how it works out, that would be appreciated.

iwyu_include_picker.cc
```diff
@@ -366,6 +366,52 @@ const SymbolMapEntry libc_symbol_map[] = {
...
+  { "LDBL_MANT_DIG", UseKind::Full, "<float.h>", kPublic },
+  { "LDBL_MAX", UseKind::Full, "<float.h>", kPublic },
+  { "LDBL_MAX_10_EXP", UseKind::Full, "<float.h>", kPublic },
+  { "LDBL_MAX_EXP", UseKind::Full, "<float.h>", kPublic },
+  { "LDBL_MIN", UseKind::Full, "<float.h>", kPublic },
+  { "LDBL_MIN_10_EXP", UseKind::Full, "<float.h>", kPublic },
+  { "LDBL_MIN_EXP", UseKind::Full, "<float.h>", kPublic },
+  { "LDBL_SNAN", UseKind::Full, "<float.h>", kPublic },
+  { "LDBL_TRUE_MIN", UseKind::Full, "<float.h>", kPublic },
+  { "NAN", UseKind::Full, "<float.h>", kPublic },
```
bolshakov-a, line 414: I'm a little concerned that `INFINITY` and `NAN` macros are defined in `<math.h>` as well, so it is probably better to add the corresponding mapping entries simultaneously because there may be users who rely on that. But `<float.h>` should still be preferred (canonical) for them, so maybe #1990 should be merged first.

OTOH, C23 p. 7.12.11 states that those macros in `<math.h>` are an obsolescent feature and suggests including `<float.h>` instead, so the mapping to `<math.h>` should probably be avoided, indeed.

CC @alejandro-colomar

bolshakov-a:
> I've tested this with both C and C++, and `<cfloat>` is proposed for C++ still.

Exactly as it should be?

iwyu_include_picker.cc
kimgr: These two bother me a little too, because they're only in `<float.h>` as of C23 (together with some `SNAN` symbols), which is quite recent:
https://en.cppreference.com/c/header/float

So it's fair for a code base targeting an older C version to pull these macros from `<math.h>`, and it would be annoying for IWYU to force them onto C23 just for the sake of it.

Maybe I should leave the C23-specific macros out for now?

kimgr: OTOH, @cboltz's test case uses `NAN`, so that's going to be annoying in another way :)

I wish we had key-able mappings.

I wonder if I should make this a small set of header mappings specifically for clang's `<float.h>` instead.

bolshakov-a: Good catch! Seems that IWYU should have language-version-specific mappings... Or the both headers should be canonical for these macros.

> because they're only in `<float.h>` as of C23

C23 final draft (N3220) still specifies them in `<math.h>` along with `<float.h>`. Btw, I have trouble accessing cppreference lately. Looks like it is blocked in/for Russia.

> I wonder if I should make this a small set of header mappings specifically for clang's `<float.h>` instead.

This could be a workaround.

kimgr:
Added an alternative here: https://github.com/include-what-you-use/include-what-you-use/pull/2076. I still think I prefer this one, I'll try adding `math.h` as alternative providers of `INFINITY` and `NAN`, and then it should be equivalent.

iwyu_include_picker.cc
kimgr: Added special-case mappings for `INFINITY`/`NAN` to `<math.h>` in this PR too. It might be good enough.

bolshakov-a: > Added special-case mappings for `INFINITY/NAN` to `<math.h>` in this PR too. It might be good enough.

#1990 would break it.

kimgr: The new private headers were introduced in https://github.com/llvm/llvm-project/commit/38473c5d351d6b2ae21b5eb1ce66c2bcc2f7c6f7, and it's hard to say for sure, but it looks like they're expecting the libc's `<math.h>` to use a config macro, `__need_infinity_nan`, to decide whether INFINITY and NAN should be defined by `<float.h>`.

So this means both `<math.h>` and `<float.h>` are or can be canonical providers of `INFINITY` and `NAN`. Is there no way to express that after #1990?

bolshakov-a: > Is there no way to express that after https://github.com/include-what-you-use/include-what-you-use/pull/1990?

It could be done by means of mapping `<float.h>` to `<math.h>`, but it would be wrong. I'd prefer to write language-standard-specific symbol mappings for C23/C++26 and for prior standards for these macros.

However, I noticed that there should probably be a way to specify multiple canonical headers in symbol mappings: https://github.com/include-what-you-use/include-what-you-use/pull/1990#issuecomment-4640351239.

kimgr: Ah, actually we can achieve language-standard-specific internal mappings now, since we can add different sets conditionally in `AddInternalMappings` based on input args. Just need to pass in `LangStandard::Kind` or something there too.

alejandro-colomar: I think I'd accept either mapping from <float.h> or <math.h>, regardless of the language version.  Let's consider that some programs may target several versions, and want to use a header that works for all of them.

kimgr:
@bolshakov-a  I'm now leaning towards the include mappings in https://github.com/include-what-you-use/include-what-you-use/issues/2073 for now, until we can find a nice design for conditional mappings.

I see I added mappings from `__float_infinity_nan.h` to `math.h` there too, but I suppose that can be removed? There's no evidence for `math.h` including that private header, as far as I can see.

bolshakov-a:
> I see I added mappings from __float_infinity_nan.h to math.h there too, but I suppose that can be removed? There's no evidence for math.h including that private header, as far as I can see.

Looks like that, because the used `math.h` is from the system's libc, not from clang-supplied headers like `float.h`.

kimgr:
Closed in favor of #2076

PR #2076, opened 2026-07-24
Add mapping generator and mappings for Clang builtin headers

### LINKED ISSUE #2073: IWYU proposes to replace float.h with __float_float.h and __float_infinity_nan.h

kimgr:
Some of the compiler builtin headers have gradually been split into
private underscore-prefixed header names on the form
'__$public_$purpose.h', which maps to '$public.h'.

Add a mapping generator which does simple name translation to produce
mappings, both for C++ and .imp.

Add a new internal mapping set for the Clang compiler builtins, based on
the output from the mapping generator.

Add a testcase for C23 and float.h. This should be reasonably portable
since float.h is part of the compiler builtin headers, which have to be
available in the resource dir for Clang and IWYU to work (but note that
it #include_next's into the system float.h for some platforms, so we
could see an entirely different include tree outside of the mainstream).

Fixes #2073.

### DIFF
--- a/clang_builtin_include_map.inc
+++ b/clang_builtin_include_map.inc
@@ -1,0 +1,24 @@
+// Clang builtin mappings, generated with:
+// mapgen/iwyu-mapgen-clang-builtins.py --lang=c++ /usr/lib/llvm-23/lib/clang/23/include/
+// Do not edit!
+// Private-to-public #include mappings.
+{ "<__float_float.h>", kPrivate, "<float.h>", kPublic },
+{ "<__float_header_macro.h>", kPrivate, "<float.h>", kPublic },
+{ "<__float_infinity_nan.h>", kPrivate, "<float.h>", kPublic },
+{ "<__stdarg___gnuc_va_list.h>", kPrivate, "<stdarg.h>", kPublic },
+{ "<__stdarg___va_copy.h>", kPrivate, "<stdarg.h>", kPublic },
+{ "<__stdarg_header_macro.h>", kPrivate, "<stdarg.h>", kPublic },
+{ "<__stdarg_va_arg.h>", kPrivate, "<stdarg.h>", kPublic },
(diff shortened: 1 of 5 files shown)


kimgr:
I tried to write up a testcase for C++ as well, but I can't get the C++ compiler to see `NAN` from either `float.h` or `cfloat`. I tried a few different variations with `-std=c++26`, but then gave up.

kimgr:
@bolshakov-a Do you still have a Windows machine where you could try the test case? It looks like Clang's `<float.h>` goes looking for system `<float.h>` via `#include_next` for MINGW and MSVC environments. Not sure what that header looks like, and if it might need other mappings.

Maybe it's just not worth adding tests for include mappings.

bolshakov-a:
> Do you still have a Windows machine where you could try the test case?

Yes, but I need some time to build the actual LLVM.

bolshakov-a:
The test passes on Windows.

kimgr:
Thanks! I'll let it stay then, maybe add a comment that it's a little semi-portable.

PR #2077, opened 2026-07-24
Preserve includes for function template default arguments

### LINKED ISSUE #1942: IWYU removes #include needed for function template default argument
> #1940 fixes the crash. However, there remains a case with wrong behavior when there is no template definition in the translation unit, like this:
> ```cpp
> // f-def_arg.h
> template <typename>
> void f(int = 0);
> 
> // f.h
> #include "f-def_arg.h"
> 
> template <typename>
> void f(int);
> 
> // main.cpp
> #include "f.h"
> 
> int main() {
>   f<char>();
> }
> ```
> ```
> f.h should add these lines:
> 
> f.h should remove these lines:
> - #include "f-def_arg.h"  // lines 1-1
> 
> The full include-list for f.h:
> ---
> 
> (main.cpp has correct #includes/fwd-decls)
> ```
> The code doesn't compile after applying the suggestion. Remapping of the implicit instantiation to the first template declaration inside `GetRedeclSpecifyingDefArg` would probably be wrong because even in this case the implicit instantiation should be scanned because a template parameter may be fully used in a function template signature like this:
> ```cpp
> template <typename T>
> T::Internal f(int = 0);
> ``` 

 _Originally posted by @bolshakov-a in [#1938](https://github.com/include-what-you-use/include-what-you-use/issues/1938#issuecomment-4028694270)_

lite-tx:
## Summary

- preserve the exact declaration that provides a default argument when an ordinary call, placement-new expression, or function-template specialization uses it
- keep source-file function-template redeclarations independent, while making header redeclarations retain inherited-default providers through their reusable header boundary
- unify default-provider lookup in `IwyuBaseAstVisitor` and keep the original specialization for instantiated-function analysis
- add focused regressions for issue #1942, placement new, and dependent calls resolved during nested template instantiation

## Testing

- clean LLVM 24/Ninja configure and build
- targeted `cxx.test_fn_def_args` and `cxx.test_fn_def_args_tpl_redecl` tests (2/2)
- full CTest suite (240/240)
- known-bugs suite (60 expected failures)
- stdin tests (3/3), mapping export (7 files), install smoke, and license-header check (958 files)
- clang-format on edited code lines, `git diff --check`, and IWYU dogfood (IWYU 0, fix-includes 0, zero diff)

Fixes #1942

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -718,19 +718,8 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
       return false;
     if (CanIgnoreCurrentASTNode())
       return true;
-    FunctionDecl* func = expr->getDirectCallee();
-    if (!func)
-      return true;
-    // If the call makes use of default arguments, select exactly that
-    // redeclaration which specifies the first default argument used.
-    for (unsigned i = 0, ie = expr->getNumArgs(); i < ie; ++i) {
-      if (isa<CXXDefaultArgExpr>(expr->getArg(i))) {
-        func = GetRedeclSpecifyingDefArg(i, func);
-        break;
-      }
-    }
     return this->getDerived().HandleFunctionCall(
-        func, TypeOfParentIfMethod(expr), expr);
+        expr->getDirectCallee(), TypeOfParentIfMethod(expr), expr);
   }
 
   bool TraverseUserDefinedLiteral(UserDefinedLiteral* expr) {
@@ -827,18 +816,6 @@ class BaseAstVisitor : public RecursiveASTVisitor<Derived> {
     const Type* parent_type = expr->getAllocatedType().getTypePtrOrNull();
     // 'new' calls operator new in addition to the ctor of the new-ed type.
     if (FunctionDecl* operator_new = expr->getOperatorNew()) {
-      // If the placement-new call makes use of default arguments, select
-      // exactly that redecl which specifies the first default argument used.
-      for (unsigned i = 0, ie = expr->getNumPlacementArgs(); i < ie; ++i) {
-        if (isa<CXXDefaultArgExpr>(expr->getPlacementArg(i))) {
-          // The 1st parameter is the size of the allocated type, there is not
-          // any corresponding argument.
-          CHECK_(i + 1 < operator_new->getNumParams());
-          operator_new = GetRedeclSpecifyingDefArg(i + 1, operator_new);
-          break;
-        }
-      }
-
       // If operator new is a method, it must (by the semantics of
       // per-class operator new) be a method on the class we're newing.
       const Type* op_parent = nullptr;
@@ -3172,14 +3149,27 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (!callee || CanIgnoreCurrentASTNode() || CanIgnoreDecl(callee))
       return true;
 
-    HandleFnReturnOnCallSite(callee, parent_type, calling_expr);
+    FunctionDecl* default_arg_provider =
+        GetUsedDefaultArgumentProvider(callee, calling_expr);
+    const bool is_template_specialization =
+        callee->isFunctionTemplateSpecialization();
+    FunctionDecl* callee_for_use =
+        default_arg_provider && !is_template_specialization
+            ? default_arg_provider
+            : callee;
+
+    HandleFnReturnOnCallSite(callee_for_use, parent_type, calling_expr);
+    if (default_arg_provider && is_template_specialization) {
+      ReportDeclUse(CurrentLoc(), default_arg_provider, nullptr,
+                    UF_DefaultArgument);
+    }
 
     // We may have already been checked in a previous
     // VisitUnresolvedLookupExpr() call.  Don't check again in that case.
     if (IsProcessedOverloadLoc(CurrentLoc()))
       return true;
 
-    ReportDeclUse(CurrentLoc(), callee);
+    ReportDeclUse(CurrentLoc(), callee_for_use);
 
     return true;
   }
@@ -3513,6 +3503,57 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
  private:
   template <typename T> friend class IwyuBaseAstVisitor;
+
+  FunctionDecl* GetUsedDefaultArgumentProvider(FunctionDecl* callee,
+                                               const Expr* calling_expr) {
+    unsigned param_index = callee->getNumParams();
+    // Argument positions do not always match parameter indexes. For example,
+    // member operator calls include the implicit object argument, while a
+    // placement-new argument list omits the implicit allocation-size parameter.
+    // Use the ParmVarDecl as the source of truth.
+    if (const auto* call_expr = dyn_cast_or_null<CallExpr>(calling_expr)) {
+      for (const Expr* arg : call_expr->arguments()) {
+        const auto* default_arg = dyn_cast<CXXDefaultArgExpr>(arg);
+        if (!default_arg)
+          continue;
+
+        param_index = default_arg->getParam()->getFunctionScopeIndex();
+        break;
+      }
+    } else if (const auto* new_expr =
+                   dyn_cast_or_null<CXXNewExpr>(calling_expr)) {
+      for (const Expr* arg : new_expr->placement_arguments()) {
+        const auto* default_arg = dyn_cast<CXXDefaultArgExpr>(arg);
+        if (!default_arg)
+          continue;
+
+        param_index = default_arg->getParam()->getFunctionScopeIndex();
+        break;
+      }
+    }
+
+    if (param_index >= callee->getNumParams())
+      return nullptr;
+
+    if (callee->isFunctionTemplateSpecialization()) {
+      FunctionDecl* pattern = callee->getTemplateInstantiationPattern();
+      if (!pattern)
+        return nullptr;
+
+      // Header patterns already retain their inherited-default providers in
+      // VisitFunctionTemplateDecl. Only source patterns need the call site to
+      // report the provider.
+      if (IsInHeader(pattern))
+        return nullptr;
+
+      CHECK_(param_index < pattern->getNumParams());
+      if (!pattern->getParamDecl(param_index)->hasInheritedDefaultArg())
+        return nullptr;
+      callee = pattern;
+    }
+
+    return GetRedeclSpecifyingDefArg(param_index, callee);
+  }
 
   bool IsProcessedOverloadLoc(SourceLocation loc) const {
     return ContainsKey(visitor_state_->processed_overload_locs, loc);
@@ -5322,6 +5363,30 @@ class IwyuAstConsumer
       }
     }
     return Base::VisitFunctionDecl(decl);
+  }
+
+  bool VisitFunctionTemplateDecl(FunctionTemplateDecl* decl) {
+    if (CanIgnoreCurrentASTNode())
+      return true;
+
+    // Default function arguments for function templates can only be specified
+    // in the initial declaration. A header redeclaration with inherited
+    // defaults therefore requires the declaration that provides them.
+    FunctionDecl* function = decl->getTemplatedDecl();
+    if (!IsInHeader(function))
+      return Base::VisitFunctionTemplateDecl(decl);
+
+    for (const ParmVarDecl* param : function->parameters()) {
+      if (!param->hasInheritedDefaultArg())
+        continue;
+
+      const unsigned param_index = param->getFunctionScopeIndex();
+      ReportDeclUse(CurrentLoc(),
+                    GetRedeclSpecifyingDefArg(param_index, function), nullptr,
+                    UF_RedeclUse);
+      break;
+    }
+    return Base::VisitFunctionTemplateDecl(decl);
   }
 
   // Avoid forward-declaration warnings for types which _should_ be already
--- a/tests/cxx/fn_def_args.cc
+++ b/tests/cxx/fn_def_args.cc
@@ -9,14 +9,35 @@
 
 // IWYU_ARGS: -Xiwyu --check_also="tests/cxx/fn_def_args-d1.h" -I .
 
-// Tests handling function default arguments. In particular, tests that IWYU
-// doesn't report them when analyzing call site.
+// Tests include requirements introduced by function default arguments.
 
 #include "tests/cxx/fn_def_args-d1.h"
 
 // IWYU: FnWithSmearedDefArgs3(int, int, int) is...*-i1.h
 void FnWithSmearedDefArgs3(int = 0, int, int);
 void FnWithDefArg(int);
+
+// Source-file redeclarations do not require their inherited default arguments
+// until a call actually uses one.
+template <typename T>
+void FnTplSourceDefault(int);
+
+template <typename T>
+void FnTplSourceExplicit(int);
+
+template <typename T>
+void FnTplSourceUnused(int);
+
+template <typename T>
+void* operator new(decltype(sizeof(0)), T, short, long) noexcept;
+
+template <typename T>
+void FnTplDefinedInHeader(int);
+
+template <typename T>
+void CallFnTplDefinedInHeader() {
+  FnTplDefinedInHeader<T>();
+}
 
 template <typename T>
 void TplFn() {
@@ -68,10 +89,26 @@ void Fn() {
   FnTplWithDefArg<float>();
   // IWYU: FnTplWithDefArg(int) is...*-i1.h
   FnTplWithDefArg<float>(1);
+
+  // IWYU: FnTplSourceDefault(int) is...*fn_def_args-i2.h
+  FnTplSourceDefault<int>();
+  FnTplSourceExplicit<int>(0);
+  // IWYU: FnTplDefinedInHeader(int) is...*fn_def_args-i2.h
+  FnTplDefinedInHeader<int>();
+  // The dependent call in CallFnTplDefinedInHeader is resolved only when the
+  // wrapper is instantiated, revealing the actual definition and its provider.
+  // IWYU: FnTplDefinedInHeader(int) is...*fn_def_args-i2.h
+  CallFnTplDefinedInHeader<int>();
+
   // IWYU: operator new(unsigned long, :0, int) is...*-i1.h
   new ('a', 1) int;
   // IWYU: operator new(unsigned long, :0, int) is...*-i1.h
   new ('a') int;
+
+  // Use a separate overload to cover a source-file template redeclaration.
+  new ('a', short{}, 1L) int;
+  // IWYU: operator new(unsigned long, :0, short, long) is...*-i2.h
+  new ('a', short{}) int;
 }
 
 /**** IWYU_SUMMARY
@@ -86,7 +123,7 @@ tests/cxx/fn_def_args.cc should remove these lines:
 The full include-list for tests/cxx/fn_def_args.cc:
 #include "tests/cxx/fn_def_args-d1.h"  // for FnWithSmearedDefArgs, Struct, operator new
 #include "tests/cxx/fn_def_args-i1.h"  // for FnTplWithDefArg, FnWithSmearedDefArgs2, FnWithSmearedDefArgs3, operator new
-#include "tests/cxx/fn_def_args-i2.h"  // for FnWithDefArg, FnWithSmearedDefArgs, FnWithSmearedDefArgs2, operator new
+#include "tests/cxx/fn_def_args-i2.h"  // for FnTplDefinedInHeader, FnTplSourceDefault, FnWithDefArg, FnWithSmearedDefArgs, FnWithSmearedDefArgs2, operator new
 #include "tests/cxx/fn_def_args-i3.h"  // for FnWithSmearedDefArgs2
 
 ***** IWYU_SUMMARY */
--- a/tests/cxx/fn_def_args_tpl_redecl-d1.h
+++ b/tests/cxx/fn_def_args_tpl_redecl-d1.h
@@ -1,0 +1,12 @@
+//===--- fn_def_args_tpl_redecl-d1.h - iwyu test --------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/fn_def_args_tpl_redecl-i1.h"
+#include "tests/cxx/fn_def_args_tpl_redecl-i2.h"
+#include "tests/cxx/fn_def_args_tpl_redecl-i3.h"
--- a/tests/cxx/fn_def_args_tpl_redecl.cc
+++ b/tests/cxx/fn_def_args_tpl_redecl.cc
@@ -1,0 +1,31 @@
+//===--- fn_def_args_tpl_redecl.cc - iwyu test ----------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I .
+
+#include "tests/cxx/fn_def_args_tpl_redecl.h"
+
+int main() {
+  // IWYU: TemplateArgument needs a declaration
+  // IWYU: TemplateArgument is...*fn_def_args_tpl_redecl-i3.h
+  FnTplDeclOnly<TemplateArgument>();
+}
+
+/**** IWYU_SUMMARY
+
+tests/cxx/fn_def_args_tpl_redecl.cc should add these lines:
+#include "tests/cxx/fn_def_args_tpl_redecl-i3.h"
+
+tests/cxx/fn_def_args_tpl_redecl.cc should remove these lines:
+
+The full include-list for tests/cxx/fn_def_args_tpl_redecl.cc:
+#include "tests/cxx/fn_def_args_tpl_redecl.h"
+#include "tests/cxx/fn_def_args_tpl_redecl-i3.h"  // for TemplateArgument
+
+***** IWYU_SUMMARY */
--- a/tests/cxx/fn_def_args_tpl_redecl.h
+++ b/tests/cxx/fn_def_args_tpl_redecl.h
@@ -1,0 +1,31 @@
+//===--- fn_def_args_tpl_redecl.h - iwyu test -----------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "tests/cxx/fn_def_args_tpl_redecl-d1.h"
+
+template <typename T>
+// IWYU: FnTplDeclOnly(int, int, int) is...*fn_def_args_tpl_redecl-i1.h
+typename T::Internal FnTplDeclOnly(int, int, int);
+
+template <typename T>
+// IWYU: FnTplNotCalled(int) is...*fn_def_args_tpl_redecl-i1.h
+void FnTplNotCalled(int);
+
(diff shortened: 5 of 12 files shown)


kimgr:
@bolshakov-a You've worked on default arguments in the past, and this patch has a slightly different shape than yours did:
* 90bf93b6955c7d8c04a98d5eb032f124af6f2b06
* f39823818b962824c178873fa2fe0c3243f9fb42
* afd879c353e4a1ade9e2c773dbeacadab900cd7a
* 2d4e4f4b9eedd773b47b5fbc824a92c992364a5a
* 43f3f1f2bab88738e13125312cc932596a9edf7e

So I wonder if there are quirks with "stacked" default arguments from many different redecls. Could you take a look?

@lite-tx Thank you! Not to be rude, but I get an LLM vibe from the change. If you did use some kind of AI service, please disclose which model was used in the commit message. If not, I apologize :)

Speaking of commit messages, we typically use a rebase workflow, so please keep the branch squashed to one commit on top of latest master, and put some rationale in the commit message (besides just a subject line).

tests/cxx/fn_def_args_tpl_redecl.h
```diff
@@ -7,16 +7,13 @@
...
-int main() {
-  f<char>();
-}
+template <typename T>
+typename T::Internal FnTplDeclOnly(int);
 
 /**** IWYU_SUMMARY
 
-(tests/bugs/1942/1942.cc has correct #includes/fwd-decls)
+(tests/cxx/fn_def_args_tpl_redecl.h has correct #includes/fwd-decls)
```
kimgr: This is typically a warning sign for new tests -- we only get detailed IWYU diagnostics if it suggests changes, so we can get much more detailed information about behavior if the test is written so that changes are expected. See https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUStyle.md#force-diagnostics for some recommendations.

tests/cxx/fn_def_args_tpl_redecl.h
```diff
@@ -7,16 +7,13 @@
 //
 //===----------------------------------------------------------------------===//
 
-// IWYU_XFAIL
+#include "tests/cxx/fn_def_args_tpl_redecl-i1.h"
```
kimgr: The `-direct`/`-indirect` and `-d1`/`-i1` pattern is a little subtle, but it's also described in the *Force diagnostics* section of the developer guide.

The idea is that *d*irect headers are included directly from the main test file (or its associated header), and *i*ndirect headers are included from direct headers only. This is exactly the kind of indirection IWYU loves to complain about, and since we want to force diagnostics, these -d/-i headers show up all over the place.

So I'm not sure if these headers were consciously laid out this way, but I would argue that they should be partitioned so `fn_def_args_tpl_redecl-d*.h` include `fn_def_args_tpl_redecl-i*.h`, and the main test files `fn_def_args_tpl_redecl.h` and `fn_def_args_tpl_redecl.cc` should just include -d headers. Then put the symbols in the right place to provoke IWYU suggestions.

bolshakov-a:
I think that reporting for a file not being currently analyzed is wrong because it makes IWUY result dependent on which translation unit it is launched (or it may be launched on the header directly, without a .cpp file calling the function). So the first template declaration should better be reported not at the call site (`HandleFunctionCall`) but at the 2nd declaration site visiting (in `VisitFunctionTemplateDecl`, IIRC).

tests/cxx/fn_def_args_tpl_redecl.h
lite-tx: Reworked this to force diagnostics. The regression now checks the exact provider warning and include replacement.

tests/cxx/fn_def_args_tpl_redecl.h
lite-tx: Done. The main test files now include only `-d` headers, and those headers include the `-i` providers.

lite-tx:
Thanks. I moved the reporting to `VisitFunctionTemplateDecl` and added an uncalled-template case, so the result no longer depends on a call site. Specialization scanning is unchanged.

I also squashed the branch to one commit on current master and added the requested rationale and AI disclosure.

kimgr:
Thanks! This LGTM, but I'd like to hear from @bolshakov-a too.

bolshakov-a:
I think that `-i3.h` header which is just a copy of `-i2.h` is not very useful for the test purpose. I would rather also get rid of `-d2.h` header by placing `#include "...-i4.h"` into `-d1.h`. Otherwise LGTM. However, for some more fine-grained handling, I'd suggest to check if the function template redeclaration is in a header file and not in a cpp-file because if it is in a cpp-file, it is probably placed there intentionally so as not to depend on any other redeclaration, so IWYU should better not pull another declaration there. But in this case, the declaration specifying default arguments should be reported at the call site if used...

tests/cxx/fn_def_args.cc
```diff
@@ -68,9 +82,13 @@ void Fn() {
...
-  // IWYU: operator new(unsigned long, :0, int) is...*-i1.h
+
+  // IWYU: FnTplSourceDefault(int) is...*fn_def_args-i2.h
+  FnTplSourceDefault<int>();
+  FnTplSourceExplicit<int>(0);
+
   new ('a', 1) int;
-  // IWYU: operator new(unsigned long, :0, int) is...*-i1.h
+  // IWYU: operator new(unsigned long, :0, int) is...*-i2.h
   new ('a') int;
```
bolshakov-a, line 106: I think it is better to invent some `operator new` prototype unrelated to the existing ones so as not to touch existing test cases.

tests/cxx/fn_def_args_tpl_redecl-d1.h
```diff
@@ -7,13 +7,6 @@
...
-void f(int);
-
-/**** IWYU_SUMMARY
-
-(tests/bugs/1942/1942.h has correct #includes/fwd-decls)
-
-***** IWYU_SUMMARY */
+#include "tests/cxx/fn_def_args_tpl_redecl-i1.h"
+#include "tests/cxx/fn_def_args_tpl_redecl-i2.h"
+#include "tests/cxx/fn_def_args_tpl_redecl-i4.h"
```
bolshakov-a: Please rename `-i4.h` -> `-i3.h`.

tests/cxx/fn_def_args_tpl_redecl.cc
```diff
@@ -0,0 +1,45 @@
...
+template <typename T>
+void FnTplNotCalled(int);
+
+template <typename T>
+void FnTplExplicitArgs(int);
+
+int main() {
+  // The source-file redeclaration does not itself require an earlier
+  // declaration, but this call uses its inherited default arguments.
+  // IWYU: FnTplDeclOnly(int, int, int) is...*fn_def_args_tpl_redecl-i1.h
```
bolshakov-a: What was the reason to change this test? Doesn't `fn_def_args` cover the added code?

iwyu.cc
```diff
@@ -5649,6 +5671,55 @@ class IwyuAstConsumer
...
       return false;
+
+    // Function template specializations must remain the callee so their
+    // instantiations are scanned. If a source-file redeclaration inherited
+    // default arguments, separately report the declaration providing them at
+    // the call site where they are used.
+    if (callee && !CanIgnoreCurrentASTNode() && !CanIgnoreDecl(callee) &&
+        callee->isFunctionTemplateSpecialization()) {
+      FunctionDecl* pattern =
+          callee->getTemplateInstantiationPattern(/*ForDefinition=*/false);
```
bolshakov-a: Could you explain why `/*ForDefinition=*/false` is needed? Is it covered in tests?

iwyu.cc
```diff
@@ -5649,6 +5671,55 @@ class IwyuAstConsumer
...
+            break;
+          }
+        }
+
+        if (param_index < pattern->getNumParams() &&
+            pattern->getParamDecl(param_index)->hasInheritedDefaultArg()) {
+          ReportDeclUse(
+              CurrentLoc(),
+              GetRedeclSpecifyingDefArg(param_index, pattern), nullptr,
+              UF_RedeclUse);
```
bolshakov-a: @kimgr, I remember you said that using `UF_RedeclUse` at the call site is not correct.

iwyu.cc
```diff
@@ -5649,6 +5671,55 @@ class IwyuAstConsumer
...
+        }
+      }
+    }
```
bolshakov-a: This handling should apply not only to function template calls from non-templated functions, but also to calls from other function templates, so it should be placed not in `IwyuAstConsumer` but in `IwyuBaseAstVisitor`.

bolshakov-a (changes requested):

iwyu.cc
```diff
@@ -3165,14 +3165,62 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // When we call (or potentially call) a function, do an IWYU check
   // via ReportDeclUse() to make sure the definition of the function
   // is properly #included.
-  bool HandleFunctionCall(FunctionDecl* callee, const Type* parent_type,
-                          const Expr* calling_expr) {
+  void ReportUsedDefaultArgumentProvider(FunctionDecl* callee,
```
bolshakov-a: Nit: can be in `private` section.

iwyu.cc
```diff
@@ -5649,6 +5722,7 @@ class IwyuAstConsumer
                           const Expr* calling_expr) {
     if (!Base::HandleFunctionCall(callee, parent_type, calling_expr))
       return false;
+
```
bolshakov-a: Nit: unrelated new line addition.

tests/cxx/fn_def_args.cc
```diff
@@ -9,15 +9,36 @@
...
+void FnTplSourceExplicit(int);
+
+template <typename T>
+void FnTplSourceUnused(int);
+
+template <typename T>
+void* operator new(decltype(sizeof(0)), T, short, long) noexcept;
+
+template <typename T>
+void FnTplSourceDefinition(int);
```
bolshakov-a: Nit: the name looks like the definition is in the source, but it is in a header actually. But I can't currently come up with a better name.

tests/cxx/fn_def_args.cc
```diff
@@ -68,10 +89,25 @@ void Fn() {
...
+
+  // IWYU: FnTplSourceDefault(int) is...*fn_def_args-i2.h
+  FnTplSourceDefault<int>();
+  FnTplSourceExplicit<int>(0);
+  // IWYU: FnTplSourceDefinition(int) is...*fn_def_args-i2.h
+  FnTplSourceDefinition<int>();
+  // IWYU: FnTplSourceDefinition(int) is...*fn_def_args-i2.h
+  CallFnTplSourceDefinition<int>();
+
+  // Preserve the existing placement-new coverage unchanged.
```
bolshakov-a: I think this comment is not needed here.

tests/cxx/fn_def_args_tpl_redecl.cc
bolshakov-a: One more time: I think that the version of this test from 533ffae was good enough except redundant headers. You've covered source file redeclarations in `fn_def_args`.

bolshakov-a (approved):
Excepting some nits, looks good, thanks!

iwyu.cc
kimgr: Yeah, I'm a little nervous about that flag 🙂. @lite-tx Could you try removing it and see if behavior changes? I think it's changed in semantics to mean something like "this is a use of a specific redecl".

bolshakov-a: It is already replaced by the new `UF_DefaultArgument` flag. I haven't tested, but I think the flag is needed so that IWYU doesn't remap the needed redeclaration to another one.

tests/cxx/fn_def_args.cc
lite-tx: Done. I kept the existing placement-new cases unchanged and added a separate overload taking `(T, short, long = 0)` for the source-redeclaration regression.

tests/cxx/fn_def_args_tpl_redecl-d1.h
lite-tx: Done; the fixture is now `-i3.h`.

tests/cxx/fn_def_args_tpl_redecl.cc
lite-tx: Agreed. I restored the focused shape from `533ffae`: `FnTplDeclOnly` is redeclared in the header and called directly, while the source-redeclaration and nested-call coverage remains in `fn_def_args`. The redundant direct header is gone.

iwyu.cc
lite-tx: This path no longer uses the `ForDefinition` override. It now reports the provider with the dedicated `UF_DefaultArgument` flag, and the focused ordinary-call and placement-new regressions cover it.

iwyu.cc
lite-tx: Done. `UF_RedeclUse` was replaced with the dedicated `UF_DefaultArgument` flag, which preserves the exact provider without treating the call as a redeclaration. Replacing it with `UF_None` breaks the intended regression.

iwyu.cc
lite-tx: Done. The shared handling now lives in `IwyuBaseAstVisitor`, and the nested templated-caller regression in `fn_def_args` exercises it.

iwyu.cc
lite-tx: Done. `ReportUsedDefaultArgumentProvider` is now in the private section.

iwyu.cc
lite-tx: Done; removed the unrelated blank line.

tests/cxx/fn_def_args.cc
lite-tx: Renamed it to `FnTplDefinedInHeader` (and the wrapper accordingly) to make the definition location explicit.

tests/cxx/fn_def_args.cc
lite-tx: Done; removed the redundant comment.

tests/cxx/fn_def_args.cc
```diff
@@ -68,10 +89,24 @@ void Fn() {
...
   // IWYU: FnTplWithDefArg(int) is...*-i1.h
   FnTplWithDefArg<float>(1);
+
+  // IWYU: FnTplSourceDefault(int) is...*fn_def_args-i2.h
+  FnTplSourceDefault<int>();
+  FnTplSourceExplicit<int>(0);
+  // IWYU: FnTplDefinedInHeader(int) is...*fn_def_args-i2.h
+  FnTplDefinedInHeader<int>();
+  // IWYU: FnTplDefinedInHeader(int) is...*fn_def_args-i2.h
+  CallFnTplDefinedInHeader<int>();
```
kimgr, line 101: @lite-tx @bolshakov-a Should `FnTplDefinedInHeader` really be reported here? I would expect `CallFnTplDefinedInHeader` to take ownership. But. maybe this is a separate issue?

iwyu.cc
```diff
@@ -3514,6 +3517,51 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
  private:
   template <typename T> friend class IwyuBaseAstVisitor;
 
+  // When we call (or potentially call) a function, do an IWYU check
+  // via ReportDeclUse() to make sure the definition of the function
+  // is properly #included.
+  void ReportUsedDefaultArgumentProvider(FunctionDecl* callee,
+                                         const Expr* calling_expr) {
+    if (!callee->isFunctionTemplateSpecialization())
```
kimgr: Doesn't the same kind of logic apply for function decls in general? i.e. something like:
```
// i1.h
void fn(int = 0, const char* = nullptr);

// d1.h
#include "i1.h"

void fn(int, const char*);

// t.cc
#include "d1.h"

void t() {
  fn();  // expecting this to point to i1.h, because it relies on default args
}
```

kimgr: (I'm not really asking for changes at this point, just trying to understand if there's a reason to keep this so template-focused)

iwyu.cc
```diff
@@ -5328,6 +5379,28 @@ class IwyuAstConsumer
...
+    FunctionDecl* function = decl->getTemplatedDecl();
+    if (!IsInHeader(function))
+      return Base::VisitFunctionTemplateDecl(decl);
+
+    for (unsigned i = 0, ie = function->getNumParams(); i < ie; ++i) {
+      if (!function->getParamDecl(i)->hasInheritedDefaultArg())
+        continue;
+
+      ReportDeclUse(CurrentLoc(), GetRedeclSpecifyingDefArg(i, function),
+                    nullptr, UF_RedeclUse);
```
kimgr: Not `UF_DefaultArgument`?

tests/cxx/fn_def_args.cc
bolshakov-a: > Should `FnTplDefinedInHeader` really be reported here?

Yes, it should. Consider the following:
```cpp
// fn_tpl_1.h
#include <type_traits>
template <typename T>
std::enable_if_t<sizeof(T) <= sizeof(int)> FnTpl(int = 0) {
}

// fn_tpl_2.h
#include <type_traits>
template <typename T>
std::enable_if_t<(sizeof(T) > sizeof(int))> FnTpl() {
}

// src.cpp
#include "fn_tpl_1.h"
#include "fn_tpl_2.h"

template <typename T>
void CallFnTpl() {
    FnTpl<T>();
}

int main() {
    CallFnTpl<int>();
}
```
`#include "fn_tpl_2.h"` in `src.cpp` is redundant if `main` instantiates only `FnTpl` from `fn_tpl_1.h` (or vice versa). Note that the two `FnTpl` declarations are definitions of two different templates, not redeclarations of the same template (they differ in the parameter lists).

kimgr: Nice example! So it's not until `CallFnTpl` instantiation time we know which `FnTpl` decls are actually used. It would be nice with some comments in the test to describe that connection.

bolshakov-a: Also note the difference between things like `CallTplFn` in source and header files. In the last case, it is better to report (i.e. keep) the both directly included `TplFn` definitions in line with the provision policy (I think `VisitUnresolvedLookupExpr` should already do this).

iwyu.cc
bolshakov-a: For non-templates, it should already work (#1827). But that approach doesn't scale in a straightforward way to templates.

iwyu.cc
bolshakov-a: This is a use of a declaration by another redeclaration, so `UF_RedeclUse` fits here.

iwyu.cc
kimgr: I was thinking the other way around -- if this function can be generalized to functions or templates, are we maybe able to call it from 1827's `TraverseCallExpr` and `TraverseCXXNewExpr` as well?

bolshakov-a: `ReportDeclUse` cannot be called from `BaseAstVisitor` where `TraverseCallExpr` and `TraverseCXXNewExpr` are located, and you cannot replace a function template specialization obtained by `getDirectCallee()` with some primary template redeclaration because it is the specialization that should be passed to `ScanInstantiatedFunction`. OTOH, finding the needed function redeclaration could probably be moved into `HandleFunctionCall`, but it would be not very good due to `CallExpr` and `CXXNewExpr` type info losing.

bolshakov-a: > it would be not very good due to CallExpr and CXXNewExpr type info losing.

But downcasting is already done here. So yeah, it should be possible to unify.

kimgr:
@lite-tx Please rebase on latest master to get CI running.

tests/cxx/fn_def_args.cc
lite-tx: Added the requested comment at the `CallFnTplDefinedInHeader<int>()` instantiation. It now explains that the dependent call is resolved only during wrapper instantiation, which reveals the selected definition and its default-argument provider.

iwyu.cc
lite-tx: Unified this in `IwyuBaseAstVisitor`. `CallExpr` and `CXXNewExpr` now pass the original callee through; ordinary calls use the selected provider as the effective callee, while template specializations report the provider separately and retain the specialization for instantiation analysis. The extra `callee_for_default_arg` path is no longer needed after the upstream normalization removal.

lite-tx:
Rebased onto latest master and force-pushed `f7f52ee`. I also unified the default-argument provider lookup in `IwyuBaseAstVisitor`, removed the now-redundant extra callee channel, and added the requested instantiation-time test comment.

Locally, the LLVM 24/Ninja configure and build pass, as do both targeted regressions, all 240 CTest tests, the known-bugs suite (60 expected failures), stdin tests, mapping export, install, license, changed-lines clang-format, git diff checks, and dogfood. The [new Actions run](https://github.com/include-what-you-use/include-what-you-use/actions/runs/31250392839) is waiting for maintainer approval.

iwyu.cc
```diff
@@ -3514,6 +3504,48 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+  FunctionDecl* GetUsedDefaultArgumentProvider(FunctionDecl* callee,
+                                               const Expr* calling_expr) {
+    unsigned param_index = callee->getNumParams();
+    if (const auto* call_expr = dyn_cast_or_null<CallExpr>(calling_expr)) {
+      for (const Expr* arg : call_expr->arguments()) {
+        const auto* default_arg = dyn_cast<CXXDefaultArgExpr>(arg);
+        if (!default_arg)
+          continue;
+
+        param_index = default_arg->getParam()->getFunctionScopeIndex();
```
kimgr, line 3520: I'm not sure what `getFunctionScopeIndex` means, but I see it delegates immediately to `getParameterIndex`, which reads a bit more clearly. But why is it necessary to ask the arg for the param's index, rather than just using a C-style for-loop and using the arg index? I mean, maybe this way is more rigorous, in a way, but it seems a bit more back-and-forth. I'm thinking something like:

```
for (unsigned i = 0; i < call_expr->getNumArgs(); ++i) {
   // ... default arg check...
   param_index = i;
   break;
}
```

What are the pros/cons?

iwyu.cc
```diff
@@ -3514,6 +3504,48 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        break;
+      }
+    }
+
+    if (param_index >= callee->getNumParams())
+      return nullptr;
+
+    if (callee->isFunctionTemplateSpecialization()) {
+      FunctionDecl* pattern = callee->getTemplateInstantiationPattern();
+      if (!pattern || IsInHeader(pattern))
```
kimgr: I'd love a comment to explain why we don't find default args for patterns in headers. I can sort of see that it's probably a reusable declaration if it's in a header, but I'm not sure why it doesn't qualify to provide a default arg...?

iwyu.cc
lite-tx: `CXXOperatorCallExpr` is the counterexample. For `fun()` where `Fun::operator()(int = 0)`, Clang 24 gives the call two arguments: the object at index 0 and the `CXXDefaultArgExpr` at index 1, while the callee has one parameter at index 0. Placement new has a similar mismatch because its placement-argument list omits the implicit allocation-size parameter. Also, `ParmVarDecl::getParameterIndex()` is private; `getFunctionScopeIndex()` is the public API. I kept the parameter-derived index and added a comment explaining this.

iwyu.cc
lite-tx: Good point. I added a comment explaining the split. `VisitFunctionTemplateDecl` already makes a header pattern with inherited defaults depend on the provider, so callers only need the pattern's header. A source-file pattern has no reusable header boundary, so its provider is reported only when a call actually uses the inherited default.

lite-tx:
Rebased once more onto `b2e4d19`, which includes the LLVM 24 target-check workaround, and force-pushed `4e0a90c`. I also added comments explaining why the parameter index comes from `CXXDefaultArgExpr` (operator calls and placement new have argument/parameter offsets) and why header template patterns are handled by `VisitFunctionTemplateDecl`.

Locally, the clean LLVM 24/Ninja build, both targeted regressions, all 240 CTest tests, 60 expected known-bug failures, stdin tests, mapping export, install, license, format/diff checks, and clean dogfood pass. The [new Actions run](https://github.com/include-what-you-use/include-what-you-use/actions/runs/31254107963) is awaiting maintainer approval.

iwyu.cc
kimgr: Thanks, that helps a lot!

iwyu.cc
kimgr: Ah, nice. So it sounds like this fixes a bug for non-template member operators as well. If it's easy, maybe add a testcase for that as well? Either way, makes sense!

lite-tx: I tried a focused A/B regression here, but the black-box IWYU result does not distinguish `getFunctionScopeIndex()` from the raw `CXXOperatorCallExpr` argument index for a non-template member operator. The class definition already provides the member/default declaration, while an out-of-class inline definition still has to be reported; forcing the default-argument redecl instead drops that definition, and compiling the resulting include set fails with an undefined reference. A meaningful index assertion would require factoring the AST mapping into a unit-testable helper, which feels disproportionate for this optional case, so I left the smoke-only testcase out. Both default-argument regressions still pass.

kimgr: Cool, thanks.

iwyu.cc
kimgr: Nice, thanks

iwyu.cc
```diff
@@ -3514,6 +3504,58 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+        const auto* default_arg = dyn_cast<CXXDefaultArgExpr>(arg);
+        if (!default_arg)
+          continue;
+
+        param_index = default_arg->getParam()->getFunctionScopeIndex();
+        break;
+      }
+    } else if (const auto* new_expr =
+                   dyn_cast_or_null<CXXNewExpr>(calling_expr)) {
+      for (unsigned i = 0, ie = new_expr->getNumPlacementArgs(); i < ie; ++i) {
```
kimgr: There is a `placement_arguments()` range wrapper on `CXXNewExpr` too, I think you should use that here for symmetry with the `CallExpr`/`arguments` above.

iwyu.cc
```diff
@@ -5324,6 +5366,28 @@ class IwyuAstConsumer
...
+      return true;
+
+    // Default function arguments for function templates can only be specified
+    // in the initial declaration. A header redeclaration with inherited
+    // defaults therefore requires the declaration that provides them.
+    FunctionDecl* function = decl->getTemplatedDecl();
+    if (!IsInHeader(function))
+      return Base::VisitFunctionTemplateDecl(decl);
+
+    for (unsigned i = 0, ie = function->getNumParams(); i < ie; ++i) {
```
kimgr: You might consider `FunctionDecl::parameters()` here too, and use `ParmVarDecl::getFunctionScopeIndex` instead of `i`. It's nice because all default-arg handling uses the same form, but also no big deal if they're different (since we almost only care about the index here)

iwyu.cc
lite-tx: Good point. Switched this to `placement_arguments()` for symmetry. It covers the same placement arguments in the same order, while parameter selection still comes from the `ParmVarDecl`. The default-argument and placement-new regressions pass.

iwyu.cc
lite-tx: Updated this to iterate over `FunctionDecl::parameters()` and derive the provider index from each `ParmVarDecl::getFunctionScopeIndex()`, matching the call-site handling. The targeted regressions and full test suite pass.

kimgr:
@lite-tx This looks good to me! Could you add back a `Co-authored-by:` attribution trailer in the commit message to say which agent and model you were using for this work?

@bolshakov-a Could you please take another look to make sure I haven't lured @lite-tx off track with my review? :)

lite-tx:
Added the requested `Co-authored-by:` trailer to the commit message in 5e49403.

bolshakov-a:
@kimgr, I tracked the introduced changes, so I think this PR is still good.

PR #2078, opened 2026-07-24
Handle provision policy for sugared types

### LINKED ISSUE #1905: Provision policy doesn't work with sugared types
I've stumbled on this while trying to fix #1665. Consider the following:
```cpp
// indirect.h
class Ret {};

class Arg {
public:
  Arg(int);
};

// using.h
class Ret;
class Arg;

namespace ns {
  using ::Ret;
  using ::Arg;
}

extern Ret r;
extern Arg a;

// direct.h
#include "indirect.h"
#include "using.h"

// fn.h
#include "direct.h"

ns::Ret Fn1(ns::Arg);
decltype(r) Fn2(decltype(a));

// main.cpp
#include "fn.h"

void User() {
  Fn1(5);
  Fn2(7);
}
```
The IWYU output is:
```
fn.h should add these lines:
#include "using.h"  // for Arg, Ret, a, r
class Arg;
class Ret;

fn.h should remove these lines:

The full include-list for fn.h:
#include "using.h"  // for Arg, Ret, a, r
class Arg;
class Ret;
---

(using.h has correct #includes/fwd-decls)

(main.cpp has correct #includes/fwd-decls)
```
That is, it suggested full argument and return types neither at function declaration nor at call site, hence compilation fails after applying this suggestion. At declaration site, besides template specialization types, only directly written tag types are analyzed for being provided (`IsProvidedTypeComponent` call inside `VisitTagType`); `UsingType` and `DecltypeType` are considered as being in fwd-decl context. At call site, the types are desugared to `RecordType`s and handled as if IWYU reported the need of `#include`s at declaration site.

I think that fixing this inconsistency by forcing the same policy as for bare `RecordType`s, including the check for any forward-declaration, would be wrong because forward-declarations are senseless both for `decltype` and for `UsingType`. I see two acceptable options:

- check only for the presence of any direct `#include` for the canonical type, regardless of forward-declarations;
- always consider such sugared types as non-providing.

I'm not sure about how critical this problem is, nor about implications on users' or library code.

lite-tx:
## Summary

- handle provision for opaque type sugar generically instead of enumerating `UsingType` and `DecltypeType`
- preserve sugar provenance inside compound types such as `Tpl<decltype(var)>`, so hidden template arguments follow the same provision policy as top-level sugar
- treat a hidden type as provided only when the declaring file defines it, directly includes its definition, or is mapped to intend to provide it
- preserve written sugar while computing return-type and autocast provision at call sites, keeping declaration and caller decisions consistent
- keep `UsingType` on its existing `UsingShadowDecl` reporting path, avoiding a redundant underlying `ReportTypeUse`

Fixes #1905.

## Regression coverage

The tests cover these independent boundaries:

- directly provided `using`, `decltype`, and `__typeof__` return/autocast types
- a one-hop transitive wrapper that must not count as a direct provider
- a definition later in the same header, without suggesting a redundant forward declaration
- nested opaque sugar in `SugarTemplate<decltype(var)>`, for both direct-provider and caller-owned cases
- both the target definition header and the using-declaration header for `UsingType`

## Validation

- rebased onto `upstream/master` at `673cc339`
- LLVM 23 build successful
- focused provider/non-provider sugar tests: 2/2
- full CTest suite: 241/241
- license-header check
- `git diff --check`


iwyu.cc
```diff
@@ -3451,9 +3446,51 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+    if (!use_file_info)
+      return false;
+
+    for (OptionalFileEntryRef include :
+         use_file_info->direct_includes_as_fileentries()) {
+      if (preprocessor_info().PublicHeaderIntendsToProvide(include,
+                                                           definition_file)) {
+        return true;
+      }
+    }
```
bolshakov-a: Considering headers included by directly included headers is a non-obvious level of indirection. I think it is better to avoid this and consider only directly-included headers: only if a header defining some class is included directly in the function-declaring header (or if the latter "intends to provide" the former, to handle external library cases), treat that class as provided and keep the include.

iwyu.cc
```diff
@@ -5458,18 +5497,35 @@ class IwyuAstConsumer
...
-    if (CanForwardDeclareType(current_ast_node())) {
+    const Type* underlying_type = type->desugar().getTypePtr();
+    const bool is_provided =
+        IsProvidedSugaredTypeComponent(current_ast_node(), underlying_type);
+    if (CanForwardDeclareType(current_ast_node()) && !is_provided) {
       ReportDeclForwardDeclareUse(CurrentLoc(), type->getDecl());
     } else {
       ReportDeclUse(CurrentLoc(), type->getDecl());
+      if (is_provided)
+        ReportTypeUse(CurrentLoc(), underlying_type, DerefKind::None);
```
bolshakov-a: Is this needed? Doesn't `ReportDeclUse` already report the underlying type in non-typedef cases? I commented out these two lines, and the tests passed.

bolshakov-a (commented):
Thanks for working on this! Generally, looks good, but the explicit enumeration of sugar types is better to be avoided. `UsingType` and `DecltypeType` are just examples of type sugar in my issue. I can think also of at least C23 `TypeOfExprType`. So I'd prefer some more general handling.

I've also found the following little issue:
```cpp
// var.h
class Class;

extern Class var;

// hdr.h
#include "var.h"

decltype(var) Fn();

class Class {
};
```
On your branch, IWYU (launched as `include-what-you-use -xc++-header hdr.h`) outputs:
```console
hdr.h should add these lines:
class Class;

hdr.h should remove these lines:

The full include-list for hdr.h:
#include "var.h"  // for var
class Class;
```
On the master branch:
```console
(hdr.h has correct #includes/fwd-decls)
```
which is more correct.

Issues with the test are similar the ones in #2077.

iwyu.cc
```diff
@@ -1134,6 +1134,50 @@ class AstFlattenerVisitor : public BaseAstVisitor<AstFlattenerVisitor> {
...
+      target_ = GetCanonicalType(target);
+      TraverseType(QualType(root, 0));
+      return found_;
+    }
+
+    bool TraverseType(QualType type, bool traverse_qualifier = true) {
+      if (type.isNull() || found_)
+        return true;
+
+      if (depth_ > 0 && GetCanonicalType(type.getTypePtr()) == target_) {
```
bolshakov-a: Does it work correctly in cases like this?
```cpp
extern Class var;

__remove_reference_t(decltype(var)) GetClass();
```
Looks like this condition evaluates to `true` when traversing `DecltypeType` here because its canonical type coincides with the `target_`, but the intent is that the desugared type should not be found in this case, IIUC?

bolshakov-a: I've tested with the following configuration:
```cpp
// var.h
class Class;

extern Class var;

// class.h
class Class {};

// direct.h
#include "class.h"

// test.h
#include "direct.h"
#include "var.h"

__remove_pointer(decltype(var)) GetClass();

// test.cpp
#include "test.h"

void Test() {
  GetClass();
}
```
The output of IWYU from the branch is wrong:
```console

#include "var.h"  // for var
---

```

lite-tx: Good catch. The canonical comparison was matching the intermediate `DecltypeType`, even though traversal never reached the desugared `Class`. I changed the finder to compare exact `Type*` identity and added a regression based on this case. It now removes the transitive include from the declaring header and adds the definition header at the call site. Applying both suggestions compiles cleanly. The exact comparison also made the depth tracking unnecessary, so I removed it.

bolshakov-a:
IWYU still doesn't report sugared component types, like in this case:
```cpp
// class.h
class Class {};

// direct.h
#include "class.h"

// var.h
class Class;

extern Class var;

// tpl.h
template <typename T>
class Tpl {
    T t;
};

// test.h
#include "direct.h"
#include "tpl.h"
#include "var.h"

Tpl<decltype(var)> GetTpl();

// test.cpp
#include "test.h"

void Test() {
    GetTpl();
}
```
```console

#include "tpl.h"  // for Tpl
#include "var.h"  // for var
---

```
So the `Class` definition that is needed for `Tpl` instantiation is lost.

iwyu.cc
lite-tx: Applied. Sugared components are now considered provided only when their definition is in the declaring file, the definition header is directly included, or the declaring file maps directly to that definition header. The new non-provider regression uses a one-hop wrapper include and verifies that the caller still receives the definition requirement.

iwyu.cc
lite-tx: Removed. The generic VisitType path now handles opaque sugar, while VisitUsingType relies on ReportDeclUse for the non-typedef underlying declaration and keeps only the separate typedef/template-instantiation handling that is still needed.

lite-tx:
Reproduced and fixed. The missing piece was that component enumeration retained the canonical `Class` but discarded the fact that it came from nested opaque sugar in `Tpl<decltype(var)>`; the ordinary return-type policy then incorrectly marked it as provided. The current revision collects concealed sugar across the full written type and applies the direct-provider rule to each hidden component. I added paired regressions: a transitive/non-provider case requires the caller to include the definition, while a direct-provider case remains quiet. The LLVM 23 build and the full 241-test CTest suite pass.

iwyu.cc
```diff
@@ -1546,7 +1611,14 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
         continue;
-      for (const Type* type : autocast_types) {
+      for (const ParmVarDecl* param : fn_redecl->parameters()) {
+        const Type* param_type = GetTypeOf(param);
+        if (!HasImplicitConversionConstructor(param_type))
+          continue;
+        const Type* type = GetConcealedDesugaredType(param_type)
+                               ? param_type
+                               : RemovePointersAndReferencesAsWritten(
+                                     param_type);
```
bolshakov-a, line 1621: A comment that this is needed because `RemovePointersAndReferencesAsWritten` desugars the given type may be worth leaving here.

iwyu.cc
```diff
@@ -3476,8 +3595,12 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
         }
       }
       const Type* canonical = GetCanonicalType(component);
-      if (!CodeAuthorWantsJustAForwardDeclare(canonical, loc))
+      if (canonical == provided_sugar_component ||
+          (concealed_sugar_components.count(canonical)
+               ? IsSugaredTypeProvidedByFile(canonical, loc)
+               : !CodeAuthorWantsJustAForwardDeclare(canonical, loc))) {
```
bolshakov-a, line 3601: I wonder if it can be done simpler:
```cpp
if (concealed_sugar_components.count(canonical)
    ? IsSugaredTypeProvidedByFile(canonical, loc)
    : !CodeAuthorWantsJustAForwardDeclare(canonical, loc)) {
```
Or, am I missing something?

Perhaps, things can be made even more simple if desugaring is moved out from `TypeEnumeratorWithoutSubstituted`. Looks like `GetConcealedDesugaredTypes` is not needed in that case.

iwyu.cc
```diff
@@ -3451,9 +3523,56 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
...
+
+    const IwyuFileInfo* use_file_info =
+        preprocessor_info().FileInfoFor(use_file);
+    if (!use_file_info)
+      return false;
+
+    return use_file_info->direct_includes_as_fileentries().count(
+               definition_file) ||
+           preprocessor_info().PublicHeaderIntendsToProvide(use_file,
+                                                            definition_file);
```
bolshakov-a, line 3548: The separate `direct_includes_as_fileentries` check is redundant, `PublicHeaderIntendsToProvide` is enough. Moreover, I'd like to see the specific behavior with external public headers covered in tests (a mapping file is needed).

tests/cxx/provided_sugar-types.h
```diff
@@ -0,0 +1,32 @@
+//===--- provided_sugar-types.h - test input file for iwyu ----------------===//
```
bolshakov-a, line 1: Why `...-types.h`? Maybe, `...-variables.h`?

tests/cxx/provided_sugar_nonprovider.cc
```diff
@@ -0,0 +1,47 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+// IWYU_ARGS: -I . \
+//   -Xiwyu --check_also=tests/cxx/provided_sugar_nonprovider-api.h
+
+#include "tests/cxx/provided_sugar-decltype-defs.h"
+#include "tests/cxx/provided_sugar_nonprovider-api.h"
+#include "tests/cxx/provided_sugar-typeof-defs.h"
+#include "tests/cxx/provided_sugar-using-defs.h"
```
bolshakov-a, line 16: `...-defs.h` headers should better not be included here directly so as to trigger diagnostics.

tests/cxx/provided_sugar_nonprovider-api.h
```diff
@@ -0,0 +1,52 @@
+//===--- provided_sugar_nonprovider-api.h - test input file for iwyu ------===//
```
bolshakov-a, line 1: Nit: there is an inconsistency between `provided_sugar_nonprovider-api.h` and `provided_sugar_provider.h` header names.

tests/cxx/provided_sugar-decltype-defs.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifndef INCLUDE_WHAT_YOU_USE_TESTS_CXX_PROVIDED_SUGAR_DECLTYPE_DEFS_H_
+#define INCLUDE_WHAT_YOU_USE_TESTS_CXX_PROVIDED_SUGAR_DECLTYPE_DEFS_H_
+
+class DecltypeReturn {};
+
+class DecltypeArgument {
+ public:
+  DecltypeArgument(int);
+};
```
bolshakov-a, line 18: Unluckily, _every_ class definition used as an argument or return type should be in its own separate header to test `provided_sugar_provider.h` completely. OTOH, the helper stuff from `-template.h`, `-types.h`, and `-using.h` can be united in one header.

bolshakov-a (commented):
Ok, generally, looks good, though suboptimal results may appear in some cases like:
```cpp
// class.h
#pragma once

class Class {};

// direct.h
#include "class.h"

// var.h
class Class;

extern Class var;

// get_tpl.h
#include "direct.h"
#include "var.h"

template <typename T1, typename T2>
class Tpl {
    T1 t1;
    T2 t2;
};

Tpl<decltype(var), Class> GetTpl();

// test.cpp
#include "get_tpl.h"

void Test() {
  GetTpl();
}
```
```console
$ include-what-you-use test.cpp -Xiwyu --check_also=get_tpl.h

get_tpl.h should add these lines:
#include "class.h"  // for Class

get_tpl.h should remove these lines:

The full include-list for get_tpl.h:
#include "class.h"  // for Class
#include "var.h"    // for var
---

#include "class.h"    // for Class

#include "class.h"    // for Class
#include "get_tpl.h"  // for GetTpl
```
However, I think, it can be accepted. Moreover, the redundant `Class` reporting from `test.cpp` goes away after the second IWYU pass.

kimgr:
@lite-tx Thanks for working on this, very ambitious patch!

On that note, is there any way to split it? It's very hard for me to keep such a big change set in my head while reviewing.

It would be nice to change the test cases to follow the `-dX.h`/`-iY.h` form. We have some testcases with `-xxx.h` where `xxx` is a signifier, like in your suggested tests here, but I think those were just added accidentally. Also, if there's any way to reduce the number of test files, that would also help make it more reviewable.

iwyu.cc
```diff
@@ -1134,6 +1134,83 @@ class AstFlattenerVisitor : public BaseAstVisitor<AstFlattenerVisitor> {
 map<const Decl*, AstFlattenerVisitor::NodeSet>
 AstFlattenerVisitor::nodeset_decl_cache_;
 
+// Return the desugared type if normal AST traversal does not visit it, or
```
kimgr, line 1137: Feels like these two fit better in `iwyu_ast_util`, since they don't rely on any state in `iwyu.cc`.

bolshakov-a:
@kimgr, if @lite-tx doesn't plan to work on this any longer, I suggest merging this as-is (while it doesn't conflict with the master branch) and making the refactoring later.

bolshakov-a:
Now it conflicts with master and needs to be rebased first.

PR #2079, opened 2026-07-25
HAXX: Play around with conditional mappings

kimgr:
I started thinking about what conditional mappings might look like. I don't think this is entirely terrible as an optional, more dynamic add-on for builtin mappings.

I just wired this up quickly, hoping to hear ideas on how to make the criteria nicer :-)

### DIFF
--- a/iwyu_include_picker.cc
+++ b/iwyu_include_picker.cc
@@ -92,10 +92,56 @@ struct SymbolMapEntry {
   IncludeVisibility to_visibility;
 };
 
+// POD to describe criteria under which a mapping applies.
+// Zero values for criteria are assumed to mean "any".
+struct MapCriteria {
+  clang::Language language;
+  int min_stdver;
+  int max_stdver;
+};
+
+// A Conditional is a POD that binds a MapType to a set of critera under which
+// it applies.
+template<class MapType>
+struct Conditional {
+  MapType mapping;
+  MapCriteria criteria;
+};
+
+// Environment implementation
+Environment::Environment(clang::Language language, int stdver)
+    : language(language),
+      stdver(stdver) {
+}
+
+bool Environment::matches(const MapCriteria& criteria) const {
+  if (criteria.language != clang::Language::Unknown &&
+      criteria.language != language)
+    return false;
+  if (criteria.min_stdver != 0 && criteria.min_stdver < stdver)
+    return false;
+  if (criteria.max_stdver != 0 && criteria.max_stdver > stdver)
+    return false;
+  return true;
+}
+
 namespace {
 
 // Listed below are all IWYU's native symbol and include mappings,
 // loosely based on GCC 4.4's libc and libstdc++.
+
+const Conditional<SymbolMapEntry> numeric_limits_symbol_map[] = {
+  // C ranges
+  { { "INFINITY", UseKind::Full, "<math.h>", kPublic }, { clang::Language::C, 199901, 202311 } },
+  { { "INFINITY", UseKind::Full, "<float.h>", kPublic }, { clang::Language::C, 202311, 0 } },
+  { { "NAN", UseKind::Full, "<math.h>", kPublic }, { clang::Language::C, 199901, 202311 } },
+  { { "NAN", UseKind::Full, "<float.h>", kPublic }, { clang::Language::C, 202311, 0 } },
+  // C++ ranges (update when 2026 version numbers come alive)
+  { { "INFINITY", UseKind::Full, "<math.h>", kPublic }, { clang::Language::CXX, 0, 202400 } },
+  { { "INFINITY", UseKind::Full, "<float.h>", kPublic }, { clang::Language::CXX, 202400, 0 } },
+  { { "NAN", UseKind::Full, "<math.h>", kPublic }, { clang::Language::CXX, 0, 202400 } },
+  { { "NAN", UseKind::Full, "<float.h>", kPublic }, { clang::Language::CXX, 202400, 0 } },
+};
 
 // Symbol -> include mappings for GNU libc
 const SymbolMapEntry libc_symbol_map[] = {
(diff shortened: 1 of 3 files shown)


iwyu_include_picker.cc
```diff
@@ -92,11 +92,57 @@ struct SymbolMapEntry {
   IncludeVisibility to_visibility;
 };
 
+// POD to describe criteria under which a mapping applies.
+// Zero values for criteria are assumed to mean "any".
+struct MapCriteria {
+  clang::Language language;
+  int min_stdver;
+  int max_stdver;
```
kimgr, line 100: Might be nice to add criteria for the CStdLib and CXXStdLib enums here too

iwyu_include_picker.cc
```diff
@@ -92,11 +92,57 @@ struct SymbolMapEntry {
...
+}
+
 namespace {
 
 // Listed below are all IWYU's native symbol and include mappings,
 // loosely based on GCC 4.4's libc and libstdc++.
 
+const Conditional<SymbolMapEntry> numeric_limits_symbol_map[] = {
+  // C ranges
+  { { "INFINITY", UseKind::Full, "<math.h>", kPublic }, { clang::Language::C, 199901, 202311 } },
```
kimgr, line 135: This would be noisier, but a lot clearer, with C++20 designated initializers:
```
{ { "INFINITY", UseKind::Full, "<math.h>", kPublic },
  { .language: clang::Language::C, .min_stdver: 199901, .max_stdver: 202311 } },
```
... and it would also make it possible to omit optional criteria.

iwyu_include_picker.cc
```diff
@@ -92,11 +92,57 @@ struct SymbolMapEntry {
...
+  { { "INFINITY", UseKind::Full, "<float.h>", kPublic }, { clang::Language::C, 202311, 0 } },
+  { { "NAN", UseKind::Full, "<math.h>", kPublic }, { clang::Language::C, 199901, 202311 } },
+  { { "NAN", UseKind::Full, "<float.h>", kPublic }, { clang::Language::C, 202311, 0 } },
+  // C++ ranges (update when 2026 version numbers come alive)
+  { { "INFINITY", UseKind::Full, "<math.h>", kPublic }, { clang::Language::CXX, 0, 202400 } },
+  { { "INFINITY", UseKind::Full, "<float.h>", kPublic }, { clang::Language::CXX, 202400, 0 } },
+  { { "NAN", UseKind::Full, "<math.h>", kPublic }, { clang::Language::CXX, 0, 202400 } },
+  { { "NAN", UseKind::Full, "<float.h>", kPublic }, { clang::Language::CXX, 202400, 0 } },
+};
+
```
kimgr, line 145: Hmm, actually seeing this now, I wonder if we would be better off just taking something like the `Environment` to `AddInternalMappings`, and then just do conditional logic:
```
if (env.matches(Language::C, 202311, 0)) {
    AddSymbolMappings(numeric_limits_c23, IWYU_ARRAYSIZE(numeric_limits_c23));
}
if (env.matches(Language::CXX, 202611, 0)) {
    AddSymbolMappings(numeric_limits_cxx26, IWYU_ARRAYSIZE(numeric_limits_cxx26));
}
...
```
The logic becomes quite complex and noisy, but it's nice that there's not a condition for each individual symbol.

A benefit of declarative per-entry (or per-set) conditions is they could potentially be extended to external .imp mappings, though.

iwyu_include_picker.cc
```diff
@@ -92,11 +92,57 @@ struct SymbolMapEntry {
...
+Environment::Environment(clang::Language language, int stdver)
+    : language(language),
+      stdver(stdver) {
+}
+
+bool Environment::matches(const MapCriteria& criteria) const {
+  if (criteria.language != clang::Language::Unknown &&
+      criteria.language != language)
+    return false;
+  if (criteria.min_stdver != 0 && criteria.min_stdver < stdver)
```
bolshakov-a, line 121:
```suggestion
  if (criteria.min_stdver != 0 && criteria.min_stdver > stdver)
```
The same below for `max_stdver`.

bolshakov-a (commented):
I was thinking about just something like
```cpp
if (!lang_opts.CPlusPlus11)
  AddSymbolMappings(mapping_for_cxx03);
else if (!lang_opts.CPlusPlus26)
  AddSymbolMappings(mapping_for_cxx11_through_cxx23);
else
  AddSymbolMappings(mapping_for_cxx26);
```
But your version looks more flexible. However, are you sure it may be useful for external mappings?

kimgr:
> However, are you sure it may be useful for external mappings?

That's a very good point. Maybe not. I tend to want to keep feature parity between them, because it makes it possible for users to tweak mappings in their local environment without rebuilding IWYU (e.g. by exporting mappings, hand-editing to make necessary changes, and running with `-Xiwyu --no_internal_mappings -Xiwyu --mapping_file=patched.imp`).

So I don't think there's usually any need for conditional external mappings, but it would be nice if it was just two views of the same data.

kimgr:
I was thinking about a spin on this. A whole mapping group protected by a condition, something like:
```c++
template<class EntryType>
struct ConditionalMapping {
   Criteria criteria;
   EntryType[] entries;
};
```
And this can then be aggregated into a single POD-inited set:
```c++
const ConditionalMapping<SymbolMapEntry> numeric_limits_symbol_map[] = {
  {
    { clang::Language::C, 199901, 202311 },
    {
      { "INFINITY", UseKind::Full, "<math.h>", kPublic },
      ...
    },
  },
  {
    { clang::Language::C, 202311, 0 },
    {
      { "INFINITY", UseKind::Full, "<float.h>", kPublic },
      ...
    },
  },
};
```

I like the simplicity of just conditionally checking the various `.CPlusPlusXY` flags, like you suggested, but it annoys me to have to have a named global variable for every single variation.

bolshakov-a:
> I tend to want to keep feature parity between them, because it makes it possible for users to tweak mappings in their local environment without rebuilding IWYU

It should be not very difficult for users to specify the mapping for the used language standard, like this: `-std=c++20 -Xiwyu --no_internal_mappings -Xiwyu --mapping_file=std_symbol_common.imp -Xiwyu --mapping_file=std_symbol_cxx11_through_cxx23.imp`.

kimgr:
> > I tend to want to keep feature parity between them, because it makes it possible for users to tweak mappings in their local environment without rebuilding IWYU
> 
> It should be not very difficult for users to specify the mapping for the used language standard, like this: `-std=c++20 -Xiwyu --no_internal_mappings -Xiwyu --mapping_file=std_symbol_common.imp -Xiwyu --mapping_file=std_symbol_cxx11_through_cxx23.imp`.

Right, but it requires that they can get `std_symbol_common.imp` and `std_symbol_cxx11_through_cxx23.imp` from somewhere. So if we had a declarative C++ representation, we'd have to translate it to multiple consistently named files in `--export_mappings`.

With your suggested chain-of-ifs adding different named sets this would fall out naturally, but with my nested structure it wouldn't map nicely.

PR #2081, opened 2026-07-26
Fix comment in `std_size_t` test

bolshakov-a:
After 146292c5224f204166bb855ba8bf392685f92691, IWYU suggests removing `#include <stdio.h>` just because it is launched in the C++ mode.


kimgr:
Thanks!

PR #2083, opened 2026-08-03
fix_includes.py: add --always_first_include flag

ericch1:
## Summary

Add a new `--always_first_include` commandline flag to `fix_includes.py`
that lets users designate one or more `#include` basenames (e.g.
`config.h`) that must always be sorted before every other `#include`
in a file, in the order given, regardless of `--reorder` or
`--quoted_includes_first`.

This is useful for projects that have a convention of always including
a specific project-wide header first (e.g. a compatibility/portability
header, or a header that must be included before any system headers
for correctness), a common pattern in several C++ codebases.

**The flag is opt-in and defaults to an empty list, so behavior is
unchanged from before this patch when the flag is not used.**

## Usage

```
# Always sort config.h first.
fix_includes.py --reorder --always_first_include=config.h < iwyu_out.txt

# Multiple basenames, sorted first in the order given (repeatable flag
# and/or comma-separated list are both supported).
fix_includes.py --reorder --always_first_include=config.h --always_first_include=platform.h < iwyu_out.txt
fix_includes.py --reorder --always_first_include=config.h,platform.h < iwyu_out.txt
```

## Implementation

* Introduced a new `_ALWAYS_FIRST_INCLUDE_KIND` (ordinal `0`, sorting
  before `_MAIN_CU_INCLUDE_KIND`) alongside the existing `*_KIND`
  constants used by `_FirstReorderSpanWith`/`_GetLineKind`, and wired it
  into both `SORT_ORDER_DEFAULT` and `SORT_ORDER_QUOTED_FIRST`.
* Added `_IsAlwaysFirstInclude()`, which matches an `#include` line's
  basename against the (ordered) list of basenames from
  `flags.always_first_include`.
* `_GetLineKind()` now accepts an optional `always_first_includes`
  parameter and returns `_ALWAYS_FIRST_INCLUDE_KIND` for matching
  `#include` lines (checked after the existing main-CU-include check).
* `_DecoratedMoveSpanLines()` overrides the `sort_key` for
  always-first `#include`s with a zero-padded index reflecting their
  position in `flags.always_first_include`, so that multiple
  always-first `#include`s keep the relative order the user asked for
  (rather than being sorted alphabetically amongst themselves).
* `_FirstReorderSpanWith()`'s assertion and "backup kind" search range
  were extended to include the new kind.
* `main()` gained the `--always_first_include` `argparse` option
  (`action='append'`, default `[]`). Its values are flattened (each
  occurrence may itself be a comma-separated list) into a single
  ordered list of basenames stored back onto `flags.always_first_include`.

No existing behavior changes when `--always_first_include` is not
passed: with an empty list, `_IsAlwaysFirstInclude()` always returns
`False`, so `_ALWAYS_FIRST_INCLUDE_KIND` is never assigned and sorting
proceeds exactly as before.

## Testing

Per `CONTRIBUTING.md`, added test coverage to `fix_includes_test.py`:

* `FakeFlags.always_first_include` (new attribute, defaults to `[]`) so
  existing tests are unaffected.
* `testAlwaysFirstInclude`: verifies that two always-first `#include`s
  given out of alphabetical order (`zconfig.h`, `aplatform.h`) are
  sorted first, and in the exact order specified, ahead of ordinary
  system and quoted `#include`s.
* `testMainAlwaysFirstIncludeFlagParsing`: verifies `main()`'s
  commandline parsing of a comma-separated `--always_first_include`
  value, run through `--sort_only --reorder`.

Ran the full existing suite locally:

```
$ python3.11 fix_includes_test.py
Ran 148 tests in 0.056s

OK
```

(148 = 146 pre-existing tests + 2 new tests, all passing.)

## Checklist (per CONTRIBUTING.md)

- [x] Added a test case to `fix_includes_test.py` for this
      `fix_includes.py` change.
- [x] Ran `python3.11 fix_includes_test.py` — all tests pass.
- [x] No behavior change when the new flag is not used (verified via
      the full existing test suite passing unmodified).

This work was almost all done by copilot/Sonnet 5.0.

Hope this may be useful for others,

Eric

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -1536,6 +1536,7 @@ def _GetNamespaceLevelReorderSpans(file_lines):
 
 
 # These are potential 'kind' arguments to _FirstReorderSpanWith.
+_ALWAYS_FIRST_INCLUDE_KIND = 0    # e.g. #include "config.h", via a flag
 _MAIN_CU_INCLUDE_KIND = 1         # e.g. #include "foo.h" when editing foo.cc
 _C_SYSTEM_INCLUDE_KIND = 2        # e.g. #include <stdio.h>
 _CXX_SYSTEM_INCLUDE_KIND = 3      # e.g. #include <vector>
@@ -1547,11 +1548,12 @@ _EOF_KIND = 7                     # used at eof
 # The span kinds are defined in default sort order, so generate a default
 # identity mapping.
 SORT_ORDER_DEFAULT = {
-  kind: kind for kind in range(_MAIN_CU_INCLUDE_KIND, _EOF_KIND + 1)
+  kind: kind for kind in range(_ALWAYS_FIRST_INCLUDE_KIND, _EOF_KIND + 1)
 }
 
 # In quoted-first mode, we sort all quoted kinds before system kinds.
 SORT_ORDER_QUOTED_FIRST = {
+  _ALWAYS_FIRST_INCLUDE_KIND: 0,
   _MAIN_CU_INCLUDE_KIND: 1,
   _NONSYSTEM_INCLUDE_KIND: 2,
   _PROJECT_INCLUDE_KIND: 3,
@@ -1565,6 +1567,26 @@ def _IsSystemInclude(line_info):
   """Given a line-info, return true iff the line is a <>-style #include."""
   # The key for #includes includes the <> or "", so this is easy. :-)
   return line_info.type == _INCLUDE_RE and line_info.key[0] == '<'
+
+
+def _IsAlwaysFirstInclude(line_info, always_first_includes):
+  """Given a line-info, return true iff its #include is always sorted first.
+
+  Arguments:
+    line_info: a LineInfo structure with .type and .key filled in.
+    always_first_includes: a sequence of #include basenames (as given by
+       the --always_first_include flag) that should always sort before
+       every other #include, in the order they're listed.
+
+  Returns:
+    True if line_info is an #include of one of the always_first_includes
+    basenames, False else.
+  """
+  if not always_first_includes or line_info.type != _INCLUDE_RE:
+    return False
+  # .key includes the surrounding <> or "" of the #include.
+  basename = os.path.basename(line_info.key[1:-1])
+  return basename in always_first_includes
 
 
 def _IsMainCUInclude(line_info, filename):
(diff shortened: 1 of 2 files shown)


fix_includes.py
```diff
@@ -1665,13 +1687,16 @@ def _IsSameProject(line_info, edited_file, project):
   return (included_root and edited_root and included_root == edited_root)
 
 
-def _GetLineKind(file_line, filename, separate_project_includes):
+def _GetLineKind(file_line, filename, separate_project_includes,
+                 always_first_includes=()):
```
kimgr: Let's not default `always_first_includes` here. It's default value is decided in the argument parsing anyway, so better to require it in here.

fix_includes.py
```diff
@@ -1712,6 +1737,7 @@ class Bang;
   *before* the using line!
 
   kind is one of the following enums, with examples:
+     _ALWAYS_FIRST_INCLUDE_KIND: #include "config.h", via a commandline flag
```
kimgr: I read `#include "config.h", via a commandline flag` as force-included via command-line, e.g. `-include config.h`.

Can you rephrase that comment a bit? Something like

```
 _ALWAYS_FIRST_INCLUDE_KIND: #include "a.h", via --always_first_include=a.h
```

fix_includes_test.py
```diff
@@ -4515,6 +4516,57 @@ def testQuotedFirst(self):
...
+#include "a.h"
+#include "b.h"
+
+int main() { return 0; }
+"""
+    iwyu_output = "(simple.cc has correct #includes/fwd-decls)\n"
+    self.RegisterFileContents({'simple.cc': infile})
+    fix_includes.sys.stdin = StringIO(iwyu_output)
+    fix_includes.main(['fix_includes', '--sort_only', '--reorder',
+                       '--always_first_include=b.h,a.h', 'simple.cc'])
```
kimgr: Could add another `--always_first_include=c.h` here to make sure repeated uses are parsed correctly.

kimgr:
> This work was almost all done by copilot/Sonnet 5.0.

Could you add a trailer to your commit message indicating this? Something like a `Co-authored-by:` seems to be conventional. You can probably ask your agent to rewrite the commit message with whatever self-attribution it usually uses.

kimgr:
Could you please rebase on latest master to catch up with recent CI fixes?

ericch1:
@kimgr , I think I did all you asked for... thanks for the review!

kimgr:
Looks good, thanks!

PR #2085, opened 2026-08-07
[ci] Pin llvm-24

kimgr:
The snapshot builds seem to be in an in-between state. Use llvm-24 explicitly for the master branch for now, to see if we can get builds through.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
 
     env:
-      LLVM_TAG:
+      LLVM_TAG: -24
 
     steps:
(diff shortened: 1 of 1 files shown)


kimgr:
Nope, the 24 builds are not really done yet, either. I'll just leave it at snapshot, the dependencies will eventually straighten out.

hebasto:
> Nope, the 24 builds are not really done yet, either. I'll just leave it at snapshot, the dependencies will eventually straighten out.

Please consider an alternative approach in https://github.com/include-what-you-use/include-what-you-use/pull/2088.

kimgr:
@hebasto Considered and merged. Thanks!

PR #2088, opened 2026-08-07
[ci] Install versioned variant of Clang snapshot

hebasto:
This is an [alternative](https://github.com/include-what-you-use/include-what-you-use/pull/2085#issuecomment-5220118227) to https://github.com/include-what-you-use/include-what-you-use/pull/2085.

The unversioned snapshot metapackages on apt.llvm.org become uninstallable while the trunk major version rolls over (currently 23 -> 24; see #2085), which breaks CI on master.

Add the snapshot suite for the runner's actual codename, detect the newest clang-N available in the package index, and install the versioned packages instead. The detected `LLVM_TAG` is exported via `GITHUB_ENV` so that subsequent steps keep using it; a preset `LLVM_TAG` short-circuits the detection.

The versioned packages co-install cleanly with the toolchains shipped on the runner image, so removing the base-dist LLVM/Clang packages is no longer necessary.

As part of compatibility effort:
1.  The `TypeTraits.h` header is renamed to `BuiltinTraits.h`. See https://github.com/llvm/llvm-project/commit/6bb30c03650c9ebc816d089ea78b36a246116f07
2. The b8b7ad1aa1675d42e18cddc0cdfb881cc52497d1 commit is [reverted](https://github.com/include-what-you-use/include-what-you-use/pull/2071#issuecomment-5219444608). See https://github.com/llvm/llvm-project/commit/63e757382099c92c2f91d0f4acb7d74a84c09ef9.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,15 +24,12 @@ jobs:
       - name: Install prerequisites
         run: |
           wget -O- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/llvm-snapshot.asc
-          sudo add-apt-repository "deb http://apt.llvm.org/noble/ llvm-toolchain-noble$LLVM_TAG main"
+          . /etc/os-release
+          REPO_NAME="deb http://apt.llvm.org/${VERSION_CODENAME} llvm-toolchain-${VERSION_CODENAME} main"
+          sudo add-apt-repository -y "${REPO_NAME}"
           sudo apt update
-          # Remove any base dist LLVM/Clang installations
-          sudo apt remove -y \
-               "libclang*" \
-               "clang*" \
-               "llvm*" \
-               "libc++*"
-          # Reinstall tagged versions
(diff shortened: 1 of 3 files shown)


kimgr:
Thank you! That's better along all axes, besides fixing the issue.

hebasto:
The CI [failure](https://github.com/include-what-you-use/include-what-you-use/actions/runs/31248127921) is caused by https://github.com/llvm/llvm-project/pull/213331.

Re-disabling CMake target checks should help:
```diff
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Work around broken packaging
         run: |
           # Disable CMake target checks.
-          # sudo ./iwyu-fixup-llvm-target-checks.bash "$LLVM_PREFIX"
+          sudo ./iwyu-fixup-llvm-target-checks.bash "$LLVM_PREFIX"
 
       - name: Build include-what-you-use with LLVM ${{ env.LLVM_MAJOR }}
         run: |
```

kimgr:
Thanks! I hadn't even noticed we were still failing :-(. Suggested fix in https://github.com/include-what-you-use/include-what-you-use/pull/2089

PR #2090, opened 2026-08-09
Take account of standard-guaranteed header mapping

bolshakov-a:
Because the standard library headers should not be handled as interchangeable, the mapping has been done in the symbol mapping and not as an include mapping, keeping the headers defining the symbols as the first (canonical) ones.


bolshakov-a:
FWIW, this is the sorted content of `providing_header_map`:
```
'compare': ['array', 'bitset', 'chrono', 'coroutine', 'deque', 'filesystem', 'flat_map', 'flat_set', 'forward_list', 'iterator', 'list', 'map', 'memory', 'optional', 'queue', 'ranges', 'regex', 'set', 'stack', 'stacktrace', 'string', 'string_view', 'system_error', 'thread', 'tuple', 'typeindex', 'unordered_map', 'unordered_set', 'utility', 'variant', 'vector']
'concepts': ['iterator', 'ranges']
'cstdint': ['cinttypes']
'initializer_list': ['algorithm', 'array', 'bitset', 'deque', 'flat_map', 'flat_set', 'forward_list', 'list', 'map', 'queue', 'random', 'ranges', 'regex', 'set', 'stack', 'string', 'unordered_map', 'unordered_set', 'utility', 'valarray', 'vector']
'ios': ['iostream']
'iosfwd': ['bitset', 'ios', 'iostream']
'istream': ['iostream']
'iterator': ['ranges']
'ostream': ['iostream', 'syncstream']
'streambuf': ['iostream']
'string': ['bitset']
```

kimgr:
Nice -- so now we can remove that annoying `stdlib_cpp_include_map`?

bolshakov-a:
> Nice -- so now we can remove that annoying `stdlib_cpp_include_map`?

I think it is better to fix [the bug with inline namespaces](https://github.com/include-what-you-use/include-what-you-use/issues/988#issuecomment-5184227261) first, otherwise for the following case:
```cpp
#include <ios>    

std::basic_streambuf<char>* sb;
std::streamoff offset;
```
IWYU reports with libc++:
```console
#include <streambuf>  // for basic_streambuf

#include <ios>        // for streamoff
#include <streambuf>  // for basic_streambuf
```
`#include <streambuf>` is unwanted and appears because the `basic_streambuf` fwd-decl use is converted to a full use. While there is `stdlib_cpp_include_map`, `<streambuf>` is mapped to `<ios>`, so IWYU accepts the code.

PR #2091, opened 2026-08-09
Fix fwd-decl mapping of types in inline namespaces

bolshakov-a:
This fixes the issue described in https://github.com/include-what-you-use/include-what-you-use/issues/988#issuecomment-5184227261.

There still remains a bug that the export pragma doesn't work with forward-declarations in inline namespaces.


kimgr:
Nice, thanks!

PR #2093, opened 2026-08-09
Add mappings for classes with duplicated decls

### LINKED ISSUE #2084: IWYU suggests `<variant>` for `std::tuple`
Consider the following MRE:
```c++
// test_correct.cpp
#include <variant>
#include <tuple>
using st = std::tuple<int, int, int>;
```

IWYU's output is correct as expected:
```console
$ include-what-you-use test_correct.cpp

test_correct.cpp should add these lines:

test_correct.cpp should remove these lines:
- #include <variant>  // lines 2-2

The full include-list for test_correct.cpp:
#include <tuple>  // for tuple
---
```

When the headers are reordered:
```c++
// test_wrong.cpp
#include <tuple>
#include <variant>
using st = std::tuple<int, int, int>;
```

IWYU produces incorrect output:
```console
$ include-what-you-use test_wrong.cpp 

(test_wrong.cpp has correct #includes/fwd-decls)
```

In real-world [codebases](https://github.com/bitcoin/bitcoin), this issue results in a false suggestion: `#include <variant>  // for tuple`.

---

```console
include-what-you-use 0.27 (git:673cc33) based on Ubuntu clang version 23.0.0 (++20260707081847+70646dd3eda3-1~exp1~20260707082012.1709)
```

bolshakov-a:
This adds the mapping entries for `std::allocator`, `std::atomic`, `std::compare_three way`, `std::tm`, and `std::tuple`. `std::allocator` should be declared not only in `<memory>` but also in `<iosfwd>`. This is because it was excluded from the mapping when `<iosfwd>` stuff was not mapped. The standard also requires that some headers provide declarations (but not definitions) of `std::atomic`, `std::tm`, and `std::tuple`. These classes are still excluded from the automapping, and the corresponding mapping entries have been added in `HANDWRITTEN_MAPPING`.

I don't remember why I excluded `compare_three_way`. Looks like there were no problem with it because the both headers (`<compare>` and `<functional>`) provide the complete definition.

This fixes #2084.


mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -152,42 +150,84 @@
 HANDWRITTEN_MAPPING = (
 # Some namespace-scope symbols are not defined in the corresponding header
 # synopses.
-    ('std::hash', 'bitset'),
-    ('std::hash<std::bitset<:0>>', 'bitset'),
-    ('std::hash', 'thread'),
-    ('std::hash<std::thread::id>', 'thread'),
+    ('std::hash', False, 'bitset'),
```
kimgr, line 153: What does the `False` mean here? Forward-decl mapping?

mapgen/iwyu-mapgen-std-symbol.py
```diff
@@ -837,9 +877,8 @@ def print_content(std_source_path, lang):
                 headers_by_symbol.setdefault(symbol, []).append(headername)
             for provided, providing in include_map:
                 providing_header_map.setdefault(provided, []).append(providing)
-    for symbol, headername in HANDWRITTEN_MAPPING:
-        # Currently, all entries in HANDWRITTEN_MAPPING are for full uses.
-        headers_by_symbol.setdefault((symbol, False), []).append(headername)
+    for symbol, flag, headername in HANDWRITTEN_MAPPING:
```
kimgr, line 880: `flag` could probably have a better name. Judging from the deleted comment it is `True` for fwd-decl and `False` for full uses?

kimgr:
Can I ask for some more detail in the commit message? It looks like `allocator`, `atomic`, `compare_three_way` and the streams headers are getting special treatment here, but only `compare_three_way` is mentioned.

mapgen/iwyu-mapgen-std-symbol.py
bolshakov-a: `True` - fwd-decl use, `False` - full use, as above in `CANONICAL_HEADERS`.

mapgen/iwyu-mapgen-std-symbol.py
bolshakov-a: > flag could probably have a better name.

Yes, but `symbol_and_flag` variable name is already used here below.

kimgr: Aha, I might look into stronger naming/typing here separately. Thanks.

bolshakov-a:
> Can I ask for some more detail in the commit message? It looks like `allocator`, `atomic`, `compare_three_way` and the streams headers are getting special treatment here, but only `compare_three_way` is mentioned.

Done. The I/O stream header mapping is not changed here, the corresponding `HANDWRITTEN_MAPPING` entries are just affected by addition of the fwd-decl flag and reformatting. But I think that these stream extraction and insertion operator entries are to be moved into a separate list, so as to multiply them by other I/O stream headers. I read @jwakely's [words](https://discourse.llvm.org/t/do-i-o-stream-headers-provide-std-char-traits/91519/4) about a defect report as "It is currently not guaranteed, but in practice and by intent, it should be." So maybe such moving should be done first, to minimize changes in this PR?

kimgr:
Thanks! This patch makes a lot more sense with that additional information. Let's deal with the I/O streams later, this LGTM.

PR #2096, opened 2026-08-11
Fix taking MemberExpr location in multiline macros

### LINKED ISSUE #1546: Invalid recommendations caused by multi-line macros
Consider the following files:

**macro.h**
```C++
#pragma once
#include <string>

void UseString(const char* message);
std::string GetString();

#define MY_MACRO() UseString(GetString() \
    .c_str())
```

**main.cc**
```C++
#include "macro.h"

int main() {
  MY_MACRO();
  return 0;
}
```

Running IWYU on `main.cc` incorrectly recommends including `<string>`:
```
$ include-what-you-use main.cc

main.cc should add these lines:
#include <string>   // for basic_string

main.cc should remove these lines:

The full include-list for main.cc:
#include <string>   // for basic_string
#include "macro.h"  // for MY_MACRO
---
```

Interestingly, the problem disappears if the macro is defined on a single line:
```C++
#define MY_MACRO() UseString(GetString().c_str())
```

```
$ include-what-you-use main.cc

(main.cc has correct #includes/fwd-decls)
```

Output of `include-what-you-use --version`:
```
include-what-you-use 0.23 (git:bc86a8f) based on clang version 19.0.0git (https://github.com/llvm/llvm-project.git 55e59083cb610f30ad21fe8c8cdb9900534937ec)
```

bolshakov-a:
`GetMemberExprLocation` was written back in 2011 (37c67ac3631457a47ff2267ca3a95a646de41ba0) when `clang::MemberExpr` did not have `getOperatorLoc()` method (was added in 2015: https://github.com/llvm/llvm-project/commit/f4cb2be05a51c6ee3aaf3939d729fe63bb81960f) and had a deficiency that it wrongly attributed a member expression to the macro instantiation location if its base expression was on the one line of a multiline macro whereas the member name was on the other line.

This fixes #1546.


bolshakov-a:
However, I'm not pretty confident that I'm correct. Consider the following (rather weird) code:
```cpp
// macro.h
#define DUMP .dump()

// class1.h
struct Class1 {
  void dump();
};

// class2.h
struct Class2 {
  void dump();
};

// src.cpp
#include "class1.h"
#include "class2.h"
#include "macro.h"

void Fn(Class1& c1, Class2& c2) {
  c1 DUMP;
  c2 DUMP;
}
```
On the master branch, IWYU accepts the code, but after my patch, `Class1` and `Class2` are attributed to the macro location which is hardly expected:
```console
macro.h should add these lines:
#include "class1.h"  // for Class1
#include "class2.h"  // for Class2

macro.h should remove these lines:

The full include-list for macro.h:
#include "class1.h"  // for Class1
#include "class2.h"  // for Class2
---

struct Class1;
struct Class2;

- #include "class1.h"  // lines 1-1
- #include "class2.h"  // lines 2-2

#include "macro.h"  // for DUMP
struct Class1;
struct Class2;
```
So maybe
```cpp
if (!IsInMacro(base_end))
  return base_end;
```
check should be kept in `GetMemberExprLocation`?

bolshakov-a:
Btw, looks like `member_expr` is a more appropriate place for the test case than `macro_location` because it already has test cases for member expressions in macros?

tests/cxx/member_expr.cc
kimgr, line 126: Given your example with `c1 DUMP`, I wonder if we should be focusing on the dot... What happens if we consider the `base` the point-of-attribution?

kimgr:
> Btw, looks like `member_expr` is a more appropriate place for the test case than `macro_location` because it already has test cases for member expressions in macros?

I think `macro_location` is basically a unittest for your favorite, `GetCanonicalUseLocation`, so `member_expr` might make more sense for this one, yeah.

kimgr:
> So maybe
> 
> ```c++
> if (!IsInMacro(base_end))
>   return base_end;
> ```
> 
> check should be kept in `GetMemberExprLocation`?

Doesn't it need to be something like:
```
// If expression is partially in a macro, pick the side that is not.
if (!IsInMacro(base_end) && IsInMacro(operator_loc))
  return base_end;
if (IsInMacro(base_end) && !IsInMacro(operator_loc))
  return operator_loc;
```
(except for the case where `operator_loc` is invalid)

tests/cxx/member_expr.cc
bolshakov-a: These cases would be broken:
https://github.com/include-what-you-use/include-what-you-use/blob/06fe577d33deb0e3cf50505717980552d64442fb/tests/cxx/member_expr.cc#L45-L49

but maybe it is Ok?

kimgr: Ah right, if the base is behind a macro then we have the same kind of issue. I guess they chose the dot because it's least likely to be hidden...

I wonder if we can use a permutation table to think about what makes sense:

| base | operator | member | effective loc |
|--------|-------------|------------|------------------|
| macro | macro | macro | macro |
| ... | ... | ... | ... |
| text | text | text | text |

It feels like if _any_ part of the expression is `text` (i.e. `!IsInMacro()`), then it should be attributed to the non-macro location.

kimgr: Ah, right, and that's what the logic was before.

bolshakov-a: However, if a macro looks like
```cpp
#define TMP_CLASS_READY_TO_USE Class{}.
```
the macro-defining header should better provide the complete `Class` type info although the macro doesn't specify a member name.

kimgr: Makes sense, yeah!

kimgr: ... and maybe same if there's no dot in the macro?
```
#define CALLABLE Class{}

// Attribute to macro? Or expansion-side?
CALLABLE.method();
```

bolshakov-a: In theory, `CALLABLE` may look like
```cpp
#define CALLABLE obj
```
and `obj` could be an external variable or a reference, so `CALLABLE` could be used where the complete type is not needed, so it would make sense not to provide it with the macro.

bolshakov-a:
> Doesn't it need to be something like:

I think it is effectively the same as
```cpp
if (!IsInMacro(base_end) && IsInMacro(operator_loc))
  return base_end;
return operator_loc;
```

tests/cxx/member_expr.cc
kimgr: The same is true if the dot is in the macro, though, right?
```c++
#define CALLABLE Class{}.
// vs.
#define CALLABLE obj.
```

bolshakov-a: Yeah, such macros are useless without the complete type.

bolshakov-a:
I've updated the PR and added the test cases for ` && IsInMacro(operator_loc)` part of the condition (look for `Test that the use location is the dot location`). But I'm not sure if such fine-grained location attribution makes much sense.

tests/cxx/member_expr.cc
```diff
@@ -111,13 +131,23 @@ void ViaMacro(const IndirectTemplate<IndirectClass>& ic) {
...
+  // IWYU: IndirectClass is...*indirect.h
+  ic DOT_METHOD;
+
   // But this member-expr is entirely in the macro, so we don't own it.
   IC_CALL_METHOD;
+  IC_CALL_METHOD_MULTILINE;
+  // IC_DOT is useless without the complete type, so its header should better
+  // provide it.
+  IC_DOT Method();
 }
```
kimgr, line 151: Looks like the only permutation we're missing is `IC DOT_METHOD` -- feels like the macro should provide it.

kimgr: OTOH `IC.CALL_METHOD` seems to be attributed to expansion loc, so not sure :)

bolshakov-a: I think that in this case, the class should be attributed to expansion location. But achieving this would involve figuring out if the operator and the base are in the same macro or not, so we'd have the problem similar to what Craig tried to solve. So I'd propose leaving it as-is and avoiding tests written for doubtful cases.

kimgr: OK, fair enough. LGTM!

PR #2097, opened 2026-08-12
Report fwd-decl used by qualified class-head-name

### LINKED ISSUE #1573: Incorrect removal of required forward declarations
In any header with a construct similar to the following example IWYU will delete forward declaration resulting in a compile error:

```
namespace foo
{
struct Bar;
}

struct foo::Bar 
{
    void frombulate();
};
```

The only way to prevent the removal is to manually mark the forward declaration with `// IWYU pragma: keep` which can be tedious.

bolshakov-a:
Constructs like
```cpp
namespace ns {
class C;
}
class ns::C {};
```
require the forward-declaration prior to the definition.

A6 step should be skipped because `DeclIsVisibleToUseInSameFile` returns true not only when the definition is before the use but also when their locations coincide, and that is the case here.

This fixes #1573.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1600,7 +1600,8 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
   // with the warning message that iwyu emits.
   virtual void ReportDeclForwardDeclareUse(SourceLocation used_loc,
                                            const NamedDecl* used_decl,
-                                           const char* comment = nullptr) {
+                                           const char* comment = nullptr,
+                                           UseFlags extra_use_flags = 0) {
     const NamedDecl* target_decl = used_decl;
 
     // Sometimes a shadow decl comes between us and the 'real' decl.
@@ -1614,7 +1615,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
     if (CanIgnoreDecl(target_decl))
       return;
 
-    UseFlags use_flags = ComputeUseFlags(current_ast_node());
+    UseFlags use_flags = ComputeUseFlags(current_ast_node()) | extra_use_flags;
 
     // Canonicalize the use location and report the use.
     used_loc = GetCanonicalUseLocation(used_loc, target_decl);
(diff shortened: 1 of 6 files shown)


tests/cxx/defn_is_use.cc
```diff
@@ -16,8 +16,31 @@ void ns1::ns2::Foo() {
...
+// to compile the definitions below.
+namespace ns1 {
+class ClassInNs1;
+enum class EnumInNs1;
+}  // namespace ns1
+
+class ns1::ClassInNs1 {};
+enum class ns1::EnumInNs1 {};
+
+// IWYU should suggest adding a forward-declaration of ClassInNs2.
```
kimgr, line 30: Should this be a TODO?

kimgr: Oh, I see. A fwd-decl is already included, but we still want a local fwd-decl?

iwyu.cc
```diff
@@ -5141,6 +5144,12 @@ class IwyuAstConsumer
       preprocessor_info().FileInfoFor(CurrentFileEntry())->AddForwardDeclare(
           decl_to_fwd_declare, definitely_keep_fwd_decl);
     }
+    // If class-head-name of the definition has a nested-name-specifier, like
+    // 'class ns::C {};', IWYU should report that a forward-declaration is
+    // required before such definition.
+    if (decl->getQualifier())
+      ReportDeclForwardDeclareUse(CurrentLoc(), decl->getPreviousDecl(),
```
kimgr, line 5152: `getPreviousDecl` looks a little scary here, but I guess it never returns nullptr? And also, does a qualified name imply that it must already be declared in the enclosing namespace?

Maybe expand the comment a bit to explain.

tests/cxx/defn_is_use.cc
bolshakov-a: In fact, emitting of "needs a declaration" warning message is suppressed in `CalculateIwyuForForwardDeclareUse`, which is probably not good, so `UF_RedeclUse` flag may be worth checking there too. But nevertheless, the suggested fwd-decl still appears in the summary.

iwyu.cc
bolshakov-a: Yes, there should be a previous redeclaration, otherwise the user's code would not compile.

tests/cxx/defn_is_use.cc
bolshakov-a: Ok, I've added the check and tested the case of a fwd-decl provided by an associated header which didn't work properly.

iwyu.cc
kimgr: I also noticed that the redecl chain is circular, so worst case we report a fwd decl use of the definition itself.

tests/cxx/defn_is_use.cc
bolshakov-a: Adding `UF_Redecl` check there was wrong, "same file" logic should work even in this case. Fixed. Changes in `fwd_decl_nested_class` have been added just to test that there is no regression.

kimgr: Cool, thanks!

PR #2101, opened 2026-08-15
Keep includes embedded in language constructs

### LINKED ISSUE #1743: IWYU should not mess with includes inside of functions or other constructs
In systemd we have the following file:

```
/* SPDX-License-Identifier: LGPL-2.1-or-later */

#include "filesystems-gperf.h"
#include "stat-util.h"

const char* fs_type_to_string(statfs_f_type_t magic) {

        switch (magic) {
#include "filesystem-switch-case.h"
        }

        return NULL;
}

int fs_type_from_string(const char *name, const statfs_f_type_t **ret) {
        const struct FilesystemMagic *fs_magic;

        assert(name);
        assert(ret);

        fs_magic = filesystems_gperf_lookup(name, strlen(name));
        if (!fs_magic)
                return -EINVAL;

        *ret = fs_magic->magic;
        return 0;
}

bool fs_in_group(const struct statfs *s, FilesystemGroups fs_group) {
        int r;

        NULSTR_FOREACH(fs, filesystem_sets[fs_group].value) {
                const statfs_f_type_t *magic;

                r = fs_type_from_string(fs, &magic);
                if (r >= 0)
                        for (size_t i = 0; i < FILESYSTEM_MAGIC_MAX; i++) {
                                if (magic[i] == 0)
                                        break;

                                if (is_fs_type(s, magic[i]))
                                        return true;
                        }
        }

        return false;
}

const FilesystemSet filesystem_sets[_FILESYSTEM_SET_MAX] = {
        [FILESYSTEM_SET_BASIC_API] = {
                .name = "@basic-api",
                .help = "Basic filesystem API",
                .value =
                "cgroup\0"
                "cgroup2\0"
                "devpts\0"
                "devtmpfs\0"
                "mqueue\0"
                "proc\0"
                "sysfs\0"
        },
        [FILESYSTEM_SET_ANONYMOUS] = {
                .name = "@anonymous",
                .help = "Anonymous inodes",
                .value =
                "anon_inodefs\0"
                "pipefs\0"
                "sockfs\0"
        },
        [FILESYSTEM_SET_APPLICATION] = {
                .name = "@application",
                .help = "Application virtual filesystems",
                .value =
                "autofs\0"
                "fuse\0"
                "overlay\0"
        },
        [FILESYSTEM_SET_AUXILIARY_API] = {
                .name = "@auxiliary-api",
                .help = "Auxiliary filesystem API",
                .value =
                "binfmt_misc\0"
                "configfs\0"
                "efivarfs\0"
                "fusectl\0"
                "hugetlbfs\0"
                "rpc_pipefs\0"
                "securityfs\0"
        },
        [FILESYSTEM_SET_COMMON_BLOCK] = {
                .name = "@common-block",
                .help = "Common block device filesystems",
                .value =
                "btrfs\0"
                "erofs\0"
                "exfat\0"
                "ext4\0"
                "f2fs\0"
                "iso9660\0"
                "ntfs3\0"
                "squashfs\0"
                "udf\0"
                "vfat\0"
                "xfs\0"
        },
        [FILESYSTEM_SET_HISTORICAL_BLOCK] = {
                .name = "@historical-block",
                .help = "Historical block device filesystems",
                .value =
                "ext2\0"
                "ext3\0"
                "minix\0"
        },
        [FILESYSTEM_SET_NETWORK] = {
                .name = "@network",
                .help = "Well-known network filesystems",
                .value =
                "afs\0"
                "ceph\0"
                "cifs\0"
                "gfs\0"
                "gfs2\0"
                "ncp\0"
                "ncpfs\0"
                "nfs\0"
                "nfs4\0"
                "ocfs2\0"
                "orangefs\0"
                "pvfs2\0"
                "smb3\0"
                "smbfs\0"
        },
        [FILESYSTEM_SET_PRIVILEGED_API] = {
                .name = "@privileged-api",
                .help = "Privileged filesystem API",
                .value =
                "bpf\0"
                "debugfs\0"
                "pstore\0"
                "tracefs\0"
        },
        [FILESYSTEM_SET_SECURITY] = {
                .name = "@security",
                .help = "Security/MAC API VFS",
                .value =
                "apparmorfs\0"
                "selinuxfs\0"
                "smackfs\0"
        },
        [FILESYSTEM_SET_TEMPORARY] = {
                .name = "@temporary",
                .help = "Temporary filesystems",
                .value =
                "ramfs\0"
                "tmpfs\0"
        },
        [FILESYSTEM_SET_KNOWN] = {
                .name = "@known",
                .help = "All known filesystems declared in the kernel",
                .value =
#include "filesystem-list.h"
        },
};

const FilesystemSet *filesystem_set_find(const char *name) {
        if (isempty(name) || name[0] != '@')
                return NULL;

        for (FilesystemGroups i = 0; i < _FILESYSTEM_SET_MAX; i++)
                if (streq(filesystem_sets[i].name, name))
                        return filesystem_sets + i;

        return NULL;
}
```

Running IWYU on this file, it will suggest removing the `#include "filesystem-switch-case.h"` and `#include "filesystem-list.h"` includes, which breaks compilation.

IWYU should probably not touch includes that are within other constructs such as functions as the potential for harm here is much bigger than the potential to do anything useful.

bolshakov-a:
This should fix #1743.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -227,6 +227,7 @@ using clang::MemberExpr;
 using clang::MemberPointerType;
 using clang::NamedDecl;
 using clang::NamespaceAliasDecl;
+using clang::NamespaceDecl;
 using clang::NestedNameSpecifier;
 using clang::NestedNameSpecifierLoc;
 using clang::OpaqueValueExpr;
@@ -5050,6 +5051,26 @@ class IwyuAstConsumer
 
   // --- Visitors of types derived from Decl.
 
+  bool VisitDecl(Decl* decl) {
+    KeepIncludeIfEmbedded();
+    return Base::VisitDecl(decl);
+  }
+
+  bool VisitStmt(Stmt* stmt) {
+    KeepIncludeIfEmbedded();
+    return Base::VisitStmt(stmt);
+  }
+
+  bool VisitType(Type* type) {
+    KeepIncludeIfEmbedded();
+    return Base::VisitType(type);
+  }
+
+  bool VisitTemplateArgumentLoc(const TemplateArgumentLoc& arg) {
+    KeepIncludeIfEmbedded();
+    return Base::VisitTemplateArgumentLoc(arg);
+  }
+
   bool VisitNamespaceAliasDecl(NamespaceAliasDecl* decl) {
     if (CanIgnoreCurrentASTNode())
       return true;
@@ -5847,6 +5868,28 @@ class IwyuAstConsumer
   }
 
  private:
+  // This is to keep includes under non-top-level declarations and other AST
+  // nodes, like in this case:
+  //
+  // int arr[] = {
+  // #include "array_items.h"
+  // };
+  void KeepIncludeIfEmbedded() const {
+    const ASTNode* parent = current_ast_node()->parent();
+    if (!parent || parent->IsA<TranslationUnitDecl>() ||
+        parent->IsA<NamespaceDecl>()) {
+      return;
+    }
+    OptionalFileEntryRef current_file =
+        GetFileEntry(GetInstantiationLoc(CurrentLoc()));
+    OptionalFileEntryRef parent_file = GetFileEntry(parent->GetLocation());
+    if (current_file != parent_file) {
+      preprocessor_info()
+          .FileInfoFor(parent_file)
+          ->ReportKnownDesiredFile(current_file);
+    }
+  }
+
   set<const Type*> GetProvidedTypeComponents(const Type* type) const {
     set<const Type*> res;
     const Type* desugared_until_typedef_or_tpl = Desugar(type);
--- a/tests/bugs/1743/enumerators.def
+++ b/tests/bugs/1743/enumerators.def
@@ -1,3 +1,0 @@
-FOO,
-BAR,
-BAZ,
--- a/tests/cxx/keep_embedded_include-d1.h
+++ b/tests/cxx/keep_embedded_include-d1.h
@@ -1,4 +1,4 @@
-//===--- 1743.cc - iwyu test ----------------------------------------------===//
+//===--- keep_embedded_include-d1.h - test input file for iwyu ------------===//
 //
 //                     The LLVM Compiler Infrastructure
 //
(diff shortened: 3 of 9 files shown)


iwyu.cc
```diff
@@ -5847,6 +5868,28 @@ class IwyuAstConsumer
...
+  // nodes, like in this case:
+  //
+  // int arr[] = {
+  // #include "array_items.h"
+  // };
+  void KeepIncludeIfEmbedded() const {
+    const ASTNode* parent = current_ast_node()->parent();
+    if (!parent || parent->IsA<TranslationUnitDecl>() ||
+        parent->IsA<NamespaceDecl>()) {
+      return;
```
bolshakov-a, line 5881: I had an idea that includes in top-level function blocks should also follow the old logic. But then I realized that a header included may contain e.g. some RAII guard not referred to outside that header, like this:
```cpp
// guard.h
std::lock_guard<std::mutex> lock(&m_mutex);
lock;  // to prevent spurious warnings that the variable is unused

// class.h
class Class {
  mutable std::mutex m_mutex;
  int m_val;
public:
  int GetVal() const {
#include "guard.h"
    return m_val;
  }
  void SetVal(int val) {
#include "guard.h"
    m_val = val;
  }
};
```
So these `#include`s should be kept although the header content is not referenced explicitly in `class.h`.

kimgr:
Elegant! Can you just describe the problem and the approach in the commit message so we have the background info preserved in the git log?

bolshakov-a:
> Elegant! Can you just describe the problem and the approach in the commit message so we have the background info preserved in the git log?

Done

iwyu.cc
```diff
@@ -5855,6 +5868,28 @@ class IwyuAstConsumer
...
+    }
+    OptionalFileEntryRef current_file =
+        GetFileEntry(GetInstantiationLoc(CurrentLoc()));
+    OptionalFileEntryRef parent_file = GetFileEntry(parent->GetLocation());
```
bolshakov-a, line 5885: Tbh, I'm not sure if `GetInstantiationLoc` should be used here as well.

kimgr: Seems fine to assume no containing macros for now, and revisit if we have a concrete case to work from.

bolshakov-a: I've found a (somehow doubtful) case and fixed it in #2104.

PR #2103, opened 2026-08-16
Remove "keep" pragmas at embedded includes

bolshakov-a:
They are not needed after 6601674725703326fc87c832772749ab84e6574d.

No functional change.


kimgr:
Nice, I was just thinking about this.

PR #2106, opened 2026-08-20
Avoid re-export policy for decls in source files

bolshakov-a:
Provision policies make sense only for header files, where IWYU cannot determine how a declaration is going to be used. As opposed to headers, source files are usually not intended to be included in other source files, so IWYU can analyze all the uses of a typedef, a function, or a template. Requiring to include more than needed in a source file for automatic re-export was just a bug.

IWYU already didn't suggest providing return and parameter types for function declarations in source files. This change extends this to typedefs, type aliases, alias templates, and default arguments of template type parameters. Moreover, this fixes an inconsistency in handling function return types in cases like
```cpp
#include "direct.h"

IndirectClass LocalGetIndirectClass();

void Use() {
  LocalGetIndirectClass();
}
```
In such cases, IWYU reported `IndirectClass` type neither at `LocalGetIndirectClass()` declaration nor at its call site, and wrongly suggested replacing `#include` with a forward-declaration.

The new test cases have been added into `iwyu_stricter_than_cpp` and `typedef_in_template` (look for `ProvidingLocal`). A lot of other tests have been changed to maintain their old behavior. In some of them, declarations intended to be providing have been moved into headers. In other tests, some test cases from source files have been duplicated in headers.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -1522,8 +1522,6 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
               GetFileEntry(call_expr), GetFileEntry(*fn_redecl))) {
         continue;
       }
-      if (fn_redecl->isThisDeclarationADefinition() && !IsInHeader(*fn_redecl))
-        continue;
       for (const Type* type : autocast_types) {
         const set<const Type*> components =
             GetProvidedTypes(type, GetLocation(*fn_redecl));
@@ -3473,6 +3471,9 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
           continue;
         }
       }
+      // Source file declarations should not provide more than actually needed.
+      if (!IsInHeader(loc))
+        continue;
       const Type* canonical = GetCanonicalType(component);
       if (!CodeAuthorWantsJustAForwardDeclare(canonical, loc))
         retval.insert(canonical);
@@ -5337,7 +5338,9 @@ class IwyuAstConsumer
   // This is called from Traverse*() because Visit*()
   // can't call HandleFunctionCall().
   bool HandleAliasedClassMethods(TypedefNameDecl* decl) {
-    if (CanIgnoreCurrentASTNode())
+    // If a typedef is defined in a source file (as opposed to headers), it is
+    // already known how it is used. Providing more than needed makes no sense.
+    if (CanIgnoreCurrentASTNode() || !IsInHeader(decl))
       return true;
 
     const Type* underlying_type = decl->getUnderlyingType().getTypePtr();
@@ -5920,6 +5923,11 @@ class IwyuAstConsumer
 
   void CollectProvidedByDefaultTplArg(SourceLocation template_loc,
                                       TemplateInstantiationData* data) const {
+    // Only declarations in headers can provide. In source files, it is already
+    // known how the template is used, hence default arguments should be
+    // reported at instantiation sites if actually needed.
+    if (!IsInHeader(template_loc))
+      return;
     for (auto [canonical, sugared] : data->resugar_map) {
       if (!sugared) {  // This designates a default argument.
         if (const NamedDecl* decl =
@@ -6004,15 +6012,18 @@ class IwyuAstConsumer
 
   pair<bool, const char*> CanBeProvidedTypeComponent(
       const ASTNode* node) const {
+    // No point in author-intent analysis in source files or for builtins.
+    if (!IsInHeader(node->GetLocation()))
+      return pair(false, nullptr);
+
     if (node->HasAncestorOfType<TypedefNameDecl>() ||
         node->HasAncestorOfType<TemplateTypeParmDecl>()) {
       return pair(true, nullptr);
     }
 
     if (const FunctionDecl* decl = node->GetAncestorAs<FunctionDecl>()) {
-      // No point in author-intent analysis of function definitions
-      // in source files, or for builtins, or for friend declarations.
-      if (!IsInHeader(decl) || IsFriendDecl(decl))
+      // No point in author-intent analysis for friend function declarations.
+      if (IsFriendDecl(decl))
         return pair(false, nullptr);
       if (const auto* method = dyn_cast<CXXMethodDecl>(decl)) {
         // Only the primary (least derived) declaration can provide types.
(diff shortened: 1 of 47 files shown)


bolshakov-a:
#2107 fixes the problem with `iwyu_output.cc`. `<functional>` was suggested for `std::less` default argument of the aliased `multimap`s inside `CalculateDesiredIncludesAndForwardDeclares` function.

tests/cxx/type_pack_elem.cc
bolshakov-a, line 49: @kimgr, you added this test in 423301855beecfb74398048202cd5c1f9d0c8731. Do you remember whether the provision policy matters for this alias template?

iwyu.cc
```diff
@@ -5920,6 +5925,11 @@ class IwyuAstConsumer
 
   void CollectProvidedByDefaultTplArg(SourceLocation template_loc,
                                       TemplateInstantiationData* data) const {
+    // Only declarations in headers can provide. In source files, it is already
```
bolshakov-a, line 5926: In fact, I'm not sure if "source files" are the correct term. Header files are also source files, after all. The term "type provision" used by me may also be confusing because "re-export" is what is used in [the docs](https://github.com/include-what-you-use/include-what-you-use/blob/0.26/docs/WhatIsAUse.md#re-exporting).

iwyu.cc
bolshakov-a: This `IsInHeader` check is not needed after this change: `GetProvidedTypes` now has its own. Should I remove it in this PR or in a follow-up NFC one?

kimgr: I think it would make sense to remove it in this PR.

tests/cxx/type_pack_elem.cc
kimgr: I'm not sure. Are you thinking to move types out into headers if provision needs to be demonstrated?

iwyu.cc
kimgr: Yeah, C++ terminology is vague here. I think "source file" is the most intuitive and succinct there is if we can live with the fact that headers are source files too.

Also, `IsHeader` is currently only looking at the file extension. We could also look at the role in the compilation (i.e. something like `!IsMainFile(...)`). I've gone back and forth there before, now I think it's more useful to decide based on name, so a reusable header is considered a header even if we analyze it directly:`include-what-you-use -xc++header a.h`

kimgr:
This bothers me a little to use `IsInHeader` in central analysis. It's a heuristic at best, and I worry that we might eventually be looking at the wrong thing.

I don't have any concrete ideas for alternatives, but it would be nicer to be able to look at AST-level language details instead of filenames. Something like internal vs. external visibility, but I don't know how that's represented, and it doesn't really say the same thing.

tests/cxx/type_pack_elem.cc
bolshakov-a: Yeah, it was what I did in this changeset when a typedef should be "providing" for the purposes of a specific test.

bolshakov-a:
I think that, at the AST-level, source and header files are indistinguishable.

bolshakov-a:
Rebased.

iwyu.cc
bolshakov-a: > Also, `IsHeader` is currently only looking at the file extension.

In the worst case, IWYU would consider a source file as a header and suggest more `#include`s than needed, as the current master does.

kimgr: That's a good way to think about it, I hadn't considered that tradeoff.

tests/cxx/type_pack_elem.cc
kimgr: I'm not sure what I was thinking in terms of type provision -- probably nothing :). I think I was just trying to construct a case where T is full-used in the template, but the use is still attributed to the instantiation, even if that instantiation is part of the instantiation of a compiler builtin: `__type_pack_element`. In other words, just throwing some stuff around and see what happens.

I think you see this a lot more clearly than I do, so if you think it makes sense to test provision policy with such a construct, feel free to move it around a bit.

As a contrast, it would make sense to leave `Tuple1617` below alone, because that is a reduction of an actual reported bug.

bolshakov-a: In that case, it is probably Ok to have `Alias` in the non-providing position.

kimgr:
> I think that, at the AST-level, source and header files are indistinguishable.

Right. I meant something more like get to the bottom of what we're _really_ asking, when we say `IsInHeaderFile`. This comment is itching in the back of my head: https://github.com/include-what-you-use/include-what-you-use/blob/master/iwyu_preprocessor.cc#L506

I think in many cases `IsInHeaderFile` is probably the right question to ask, but sometimes it might be "is this decl visible to more than one file?" or maybe "is this an externally visible declaration?", or even "is this the canonical declaration?".

They don't really answer the same thing, and I agree that `IsInHeaderFile` is probably a good/better approximation. But it might be a good idea to add a wrapper abstraction to capture what this is really about, maybe `IsReusableDeclaration`, and use that where the primary question is not "is in header" but more like "what role does the declaration play in the program".

I'm just thinking out loud, I don't think there's anything actionable here.

tests/cxx/type_pack_elem.cc
kimgr: Also, I saw in the PR for this test, I mentioned I'd stolen "a few nice examples" from alias_template.cc, so this probably maps directly to: https://github.com/include-what-you-use/include-what-you-use/pull/2106/changes#diff-60191070addfab3b374eb2f2d1d02404ba1fcbcc35cb142fd9a14c1f4c721d6cR27

bolshakov-a: I'm not sure about `alias_template.cc` test cases either, and duplicated them in the header "just in case". Btw, it was you again who developed `alias_template.cc`: 8fd93dc87f1aef2541a56d6f5576715b8b761ae1.

bolshakov-a:
When you see in a preprocessed source a line
```cpp
typedef IndirectClass Type;
```
or the corresponding AST subtree, how can you decide if it is intended to be used externally or not?

The idea of checking if a file as a whole has any "non-inlined" definition looks interesting to me as a way to determine if a file is a header or not (so this is probably what should be inside `IsHeaderFile` function) , but it doesn't work when analyzing a distinct declaration, in isolation from others.

kimgr:
Yeah, I don't think there's any clear cut way to do it without more context. Maybe for functions `isExternallyVisible` or `getFormalLinkage` would be one interesting characteristic, since it says whether a declaration is local to the translation unit. Something more would be necessary to reason about whether it's local to _the file_, but yeah. Anyway. `IsInHeaderFile` is available and seems useful as-is.

kimgr:
Thanks, this is a major improvement!

PR #2107, opened 2026-08-20
Fix obtaining aliased template type location

bolshakov-a:
When an aliased type was passed to `GetTplInstData`, `TypeToDeclAsWritten` returned the alias declaration, not the class template declaration, so determination of default argument provision status didn't work correctly.


bolshakov-a:
The problem arose on scanning methods of the template, not the template itself.

PR #2111, opened 2026-09-01
Check also in unit tests

ericch1:
Fixes `run_iwyu_tests.py`, which fails a
test whenever IWYU reports on a file its globs did not collect. 

**`run_iwyu_tests.py`** — expectations are read only from `foo-*.*`, `foo.h` and
`foo.cc`, but summaries are compared over every file IWYU reports on, so a summary
for any other file gets diffed against nothing and fails the test. That made it
impossible to cover another file of a component, even with `--check_also` aimed
straight at it. The globs now take in every sibling sharing the test's stem.
New test: `check_also_sibling.{cc,inc}`.

The work has been done by Claude Opus 5.

Hope this help!

Eric


kimgr:
@ericch1 Hey, thanks! First off, can you split this into two separate PRs? The unrelated changes will just end up blocking each other, otherwise. The fix_includes feature looks good already, but I have some reservations with the run_iwyu_tests change.

ericch1:
ok, I removed the case insensitive commit from here and putted it here #2113

ericch1:
Here too, the C++ build failures can't be caused by this commit, I just modified a test and the python script...

kimgr:
Hmm, it took me a while to see how `--check_also` and the file collection globs were even related.

But I see now that all `--check_also`s are matching pretty broadly (e.g. `tests/cxx/*-d*.h`), but the test inputs in fact mostly/only include files with a matching stem. So there's kind of a hidden assumption that any included files matched by `--check_also` are also in the glob set wired in `TestIwyuOnRelevantFiles`... and then that hidden assumption doesn't cover `.inc`, which is what this patch fixes.

I wonder if we could remove that hidden assumption instead?

Just spitballing, but maybe look at the actual summary output from `include-what-you-use` (which should cover all `--check_also`, associated headers, and main file), extract filenames, and use those filenames to extract expected summary information and perform assertions?

ericch1:
Ok, good remark. I removed the assumption instead of widening the globs: the harness now takes IWYU's summary output as the authority on which files were checked, and reads expectations out of those. The globs remain only for the direction the output can't cover — noticing a file whose IWYU_SUMMARY isn't reported on at all — plus the files a test names outright in --check_also. That leftover is written down where it's made now, rather than hidden, and it can only hide a lost expectation, never invent one.

Turns out this covers more than .inc: tests/bugs/1905/fn.h was never in the glob set either. I also added iwyu_test_util_test.py, since the harness had no unit tests of its own. Full suite is green locally (ctest + the bugs suite) against LLVM 24, with a local patch for the two clang API changes that are breaking the C++ build on master right now.

iwyu_test_util_test.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
```
kimgr, line 1: I think this might be overshooting, with tests for the test engine. I wonder if we can achieve something like it with internal assertions in `iwyu_test_util.py`? I can imagine this was quite useful when developing the summary detection, etc, but it feels a bit much for regression testing. I'm on the fence whether we should keep it... Sway me, please :)

kimgr:
I see now I hadn't fully understood the duality you were trying to solve in the initial patch.

* we have a set of "files to check", and we only use them to discover expected summary assertions
* when IWYU runs we get a set of actual summaries
* I think there were two problems:
  - the patterns for "files to check" didn't include more unusual extensions like `mytest.inc`
  - expected summaries from `--check_also`-files were only considered if they happened to match the patterns (e.g. `*-d*.h` matching `mytest-d1.h`). Other files were simply not considered.

I'm not sure if `_CompareExpectedAndActualSummaries` would complain about missing actual _and_ missing expected, but I think it might.

I had the same instinct at first, to parse `--check_also`, but I don't think it helps much, because it's very unusual that they specify actual files (almost always globs). And it feels surprising that they are sometimes considered, but mostly not.

Also, `--check_also` globs are not necessarily tied to the filename being tested... The case you found for bug 1905 is a good example. But I wrote that recently, and I had no idea this was so involved :-)

I'll need to think a bit more about this... It feels like there's something straightforward hiding.

ericch1:
You were right: `_CompareExpectedAndActualSummaries` does complain both ways. It
walks `set(actual) | set(expected)`, so if either side is missing, the diff is not
empty. Both failures print the same header line, and only the sign of the diff tells
them apart.

I think this is also the straightforward thing you said might be hiding, so let me put
the two cases side by side. Here is what each one does on master, and what this branch
makes of it:

| | on master | on this branch |
|---|---|---|
| **A** actual, no expected — IWYU reports on a file, but nobody read that file's expectations | `@@ -0,0 +1,6 @@` with `+` lines, which looks like a real IWYU finding | fixed: the filenames now come from the summary output |
| **B** expected, no actual — a file has an IWYU_SUMMARY, but IWYU never reported on it | `@@ -1 +0,0 @@` with `-` lines | prints `No summary reported for X, which has an IWYU_SUMMARY` |

I had not separated these two cases before, and they are not equally easy to fix:

* **A** can be solved exactly. The summary output *is* the list of files that were
  checked. So the hidden assumption you asked me to remove really does disappear, and
  nothing has to be collected before the run.
* **B** cannot. Nothing records which files *should* have been checked. The globs do
  not, and `--check_also` does not either: a glob also matches headers that belong to
  other tests and are not part of this translation unit. Any answer is a guess.

So I would like to remove the **B** half again: `_GetCheckAlsoFiles`, its two regexes,
`_HasExpectedSummary`, and the message in the table. Column three for **B** would then
say the same as column two. I will also remove `iwyu_test_util_test.py` — you are right
about that one. It helped me while I was investigating this, but it is too much to
maintain for regression testing, and `check_also_sibling` already covers the behaviour.

What remains is `files_to_check = caller's list | actual summaries`, about fifteen
lines, and `run_iwyu_tests.py` is not touched at all. It still fixes the original
problem: `tests/cxx/check_also_sibling.inc` and `tests/bugs/1905/fn.h` were both
invisible to the harness, and both are case **A**.

If **B** is worth doing one day, I do not think it belongs in a single test. For the
whole suite there is an exact answer: every file under `tests/` with a non-empty
IWYU_SUMMARY should be reported by at least one test. Comparing those two sets needs no
globs and no `--check_also`. The problem is that CMake registers one ctest per test
file, so the union is not available where the check would run. A separate PR, if you
like the idea.

kimgr:
I'm afk at the moment, but this makes sense to me. Just to add one maybe useful detail: I think it was a silent assumption that check_also would only match the globbed files_to_check.

I wonder if we could solve B by making that manifest. Something like evaluate check_also against files_to_check and diagnose if something is missing (which indicates you need to change your check_also).

But maybe that's in conflict with your plan for A...

ericch1:
Pushed, **A** only. `files_to_check` is now the caller's list plus whatever IWYU
reported on. Gone: `_GetCheckAlsoFiles`, its two regexes, `_HasExpectedSummary`, the
"no summary reported" message, and `iwyu_test_util_test.py`.

**B** is left as it is on master.

iwyu_test_util.py
```diff
@@ -585,6 +585,31 @@ def _GetLaunchArguments(cc_file):
   return shlex.split(args)
 
 
+def _CollectFilesToCheck(cpp_files_to_check, actual_summaries):
+  """Returns the files to read expectations from.
+
+  IWYU summarizes the main file, its associated headers and everything matched
```
kimgr, line 591: Very nice, this makes a lot of sense to me now.

iwyu_test_util.py
```diff
@@ -585,6 +585,31 @@ def _GetLaunchArguments(cc_file):
...
+  by --check_also, so the summary output is the authority on what was checked.
+  The caller's candidates need not cover that set -- they are collected by
+  globbing for the test's stem, which misses e.g. a --check_also target with a
+  different name or extension -- so take their union.
+  """
+  files_to_check = list(cpp_files_to_check)
+  known = set(files_to_check)
```
kimgr: I think we can collapsed these two;
```
files_to_check = set(cpp_files_to_check)
```
(order might change, but I think that's neither here nor there)

iwyu_test_util.py
```diff
@@ -585,6 +585,31 @@ def _GetLaunchArguments(cc_file):
...
+  for filename in sorted(actual_summaries):
```
kimgr: Do the summaries need to be sorted? I don't think so...?

kimgr: Also, if `actual_summaries` is a dict, I'd prefer if we explicitly used `.keys()` to get an iterator over the keys. The implicit iteration over keys for dicts has always confused me :)

iwyu_test_util.py
```diff
@@ -585,6 +585,31 @@ def _GetLaunchArguments(cc_file):
...
+      continue
+    # Only files in the test tree can carry expectations.  IWYU also reports on
+    # library headers (--check_also='*' matches those too), and prints paths
+    # that need not resolve from the test runner's working directory.
+    if os.path.isabs(filename) or not os.path.isfile(filename):
+      continue
+    known.add(filename)
+    files_to_check.append(filename)
+
+  return files_to_check
```
kimgr: ... and then if `files_to_check` is a set, we can return `sorted(files_to_check)` here to get a consistently-shaped list.

ericch1: ok, I push all these suggestions, thanks!

kimgr: Thank you! LGTM!

PR #2113, opened 2026-09-05
fix_includes.py: add --case_insensitive sorting flag

ericch1:
#includes are sorted by their raw text, so ASCII order puts every capitalised name ahead of every lowercase one: "Zebra.h" sorts before "apple.h".  Projects that name headers after the types they declare end up with their #include blocks split in two by capitalisation alone.

Add --case_insensitive, which lowercases the sort key so that names interleave the way a reader expects.  Off by default; nothing changes unless it is asked for.

The flag only affects the comparison, not the text: the #include lines themselves are written back unchanged.

The work has been done by Claude Opus 5.

Hope this help!

Eric

### DIFF
--- a/fix_includes.py
+++ b/fix_includes.py
@@ -2209,6 +2209,8 @@ def FixFileLines(iwyu_record, file_lines, flags, fileinfo):
   def key(decorated_span):
     reorder_span, kind, sort_key, all_lines = decorated_span
     kind_key = GetLineSortOrdinal(kind, flags.quoted_includes_first)
+    if flags.sort_nocase:
+      sort_key = sort_key.lower()
     if flags.reorder:
       return reorder_span, kind_key, sort_key, all_lines
     else:
@@ -2462,6 +2464,9 @@ def main(argv):
   parser.add_argument('--noreorder', action='store_false', dest='reorder',
                       help=('Do not re-order lines relative to other similar '
                             'lines.'))
+  parser.add_argument('--sort_nocase', action='store_true', default=False,
+                      help=('Sort #includes ignoring case, so that "apple.h" '
+                            'sorts before "Banana.h" instead of after it.'))
 
   parser.add_argument('-s', '--sort_only', action='store_true',
                       help=('Just sort #includes of files listed on cmdline;'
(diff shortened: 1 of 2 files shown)


ericch1:
The C++ build faillures can't be caused by this commit since it just modify the python script... ???

fix_includes.py
```diff
@@ -2462,6 +2464,9 @@ def main(argv):
   parser.add_argument('--noreorder', action='store_false', dest='reorder',
                       help=('Do not re-order lines relative to other similar '
                             'lines.'))
+  parser.add_argument('--case_insensitive', action='store_true', default=False,
```
kimgr: Hmm, I think `--case_insensitive` is both a bit too long and too generic (case-insensitive in what respect?).

How about `--sort_nocase`?

kimgr:
> The C++ build faillures can't be caused by this commit since it just modify the python script... ???

Yeah, upstream LLVM has changed in some breaking way. We'll figure that out separately.

ericch1:
Ok, `--case_insensitive` doesn't say case-insensitive in *what*. I'm happy to rename it..

I asked Claude for suggestions, pick whichever you prefer and I'll amend the commit:

- `--sort_nocase` (your suggestion) — shortest of the specific ones, and the sort_ prefix leaves room for future sort-related options. My only hesitation: elsewhere in the script "no" is a negation prefix (--noreorder, --nocomments), so --sort_nocase can read for a moment like "don't sort by case".
- `--sort_ignore_case` — no such ambiguity, at the cost of a few more characters.
- `--sort_casefold` — names the actual operation and is the Unicode-correct one, but it's Python jargon and probably too obscure for a user-facing flag.
- `--sort_order=ascii|nocase` — extensible without adding more booleans later, but it breaks with the all-boolean style of the other flags, so I'd rather not unless you see other orderings coming.

I changed it to `--sort_nocase` for now, but I'll pick the one you choose if you prefer another one...

kimgr:
I think the `--sort_order=ascii|nocase` is tempting, but I think it's quite unlikely to grow more options. If we do end up with another sort order, I'd be fine with introducing an enum-style switch at that point.

This LGTM, thank you!

PR #2114, opened 2026-09-06
Adapt to Clang RecursiveASTVisitor and ConceptReference changes

ericch1:
TraverseTemplateName now takes a TraverseQualifier parameter; forward it, as the TraverseType and TraverseTypeLoc overrides above already do.

ConceptReference::getNamedConcept now returns a TemplateName, which need not name a decl, so report a use only when it does.

Work done by Claude Opus 5.

### DIFF
--- a/iwyu.cc
+++ b/iwyu.cc
@@ -5540,7 +5540,9 @@ class IwyuAstConsumer
   bool VisitConceptReference(ConceptReference* concept_ref) {
     if (CanIgnoreCurrentASTNode())
       return true;
-    ReportDeclUse(CurrentLoc(), concept_ref->getNamedConcept());
+    if (const TemplateDecl* concept_decl =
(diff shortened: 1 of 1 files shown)


kimgr:
Thank you! I was already dreading trying to root-cause the breakage :-)

Let's split this into two _commits_ -- can be the same PR since they are both necessary for a green build. And check the Git log for similar changes, there's some regularity:

* Prefix the commit message with `[clang compat]` -- I use this for release management to quickly see/revert breaking changes
* Reference the LLVM-side commit that introduced the breaking change; this make it easier for us to revert again if the change is reverted LLVM-side. See other `[clang compat]` commits for examples, I think we usually use a GitHub URL.

Thanks again.

ericch1:
Done, hoping it is like what you expect.

Thanks,
Eric

kimgr:
Excellent, thanks!

PR #2117, opened 2026-09-16
Add associated_header mapping directive

ericch1:
# Associate a header with the file holding its inline definitions

Some projects keep a header's inline definitions in a separate file, and the
header `#include`s that file itself, usually at the very end:

| | header | includes |
|---|---|---|
| [libstdc++](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/basic_ios.h) | `bits/basic_ios.h` | `bits/basic_ios.tcc` |
| [ITK](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImage.h) | `itkImage.h` | `itkImage.hxx` |
| [Boost.Asio](https://github.com/boostorg/asio/blob/develop/include/boost/asio/error.hpp) | `boost/asio/error.hpp` | `boost/asio/impl/error.ipp` |
| [OpenFOAM](https://github.com/OpenFOAM/OpenFOAM-dev/blob/master/src/OpenFOAM/primitives/Vector/Vector.H) | `Vector.H` | `VectorI.H` |

Such a file is only ever compiled as part of its header, so it should be able to
use what the header `#include`s — exactly as `foo.cc` may use what `foo.h`
provides. IWYU analyzed it on its own instead, and asked it to repeat those
`#include`s.

`IWYU pragma: associated` cannot express this. It has to sit on an `#include` of
the associated header, and here the header includes the implementation file, not
the other way round — the pragma would need an `#include` written only to carry
it.

## Two commits

**1. The `associated_header` mapping directive.**

```json
{ "associated_header": ['@"(.*)\.icc"', '"\1.h"'] }
```

`[from, to]`: a quoted include or `@`-regex for the implementation file, and the
quoted include of its associated header, with capture groups as for `include`
mappings. A mapping rather than a list of extensions, because of the table above
— only the first pair shares a stem and differs only in extension. ITK keeps the
stem but uses `.hxx`, Asio moves the file to `impl/`, OpenFOAM changes the stem.

The relation stays one-way. A mapping file can name both directions, and two
files that are each other's associated header make the two analyses depend on
each other, tripping `Assertion failed: desired_includes_have_been_calculated_`.

**2. `.icc`, `.inl`, `.ipp` and `.tcc` recognized without configuration.**

Separable: drop it and the directive still works, it just costs those projects a
two-line mapping file.

## Tests

`associated_icc.*` for the `.icc` case, `associated_mapping_custom.*` for a
`.inline` convention IWYU knows nothing about, `associated_icc_redundant_include.*`
for a file that `#include`s its header back, `associated_mapping_cycle.*` for a
mapping file naming both directions, and `associated_icc_default.*` for the
built-in extensions with no mapping file at all.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bolshakov-a:
I wonder if the problem can be solved by a new pragma to be placed on `#include "*.icc"` in the associated header? Mapping files are primarily for external libraries which are not under user's control.

ericch1:
Good idea. I shared and discussed with Claude, here are our answer and propositions:

---
A pragma is the right tool when a project has only a few such headers,
or when it prefers to keep the information next to the code. I have it working
locally as a separate commit, but I would like to agree on the name below before I
push it:

    #include "component/component.icc"  // IWYU pragma: associated_impl

`IWYU pragma: associated` could not be reused. It has to sit on an `#include` of
the associated header, and here the `#include` points the other way. Both forms
run through the same code, so the one-way rule and the rest are shared.

That brings up a naming question, and I would like your opinion. The new directive
in the mapping file is currently called `associated_header`, which does not match
the pragma. In the other directives, the key says what the first entry is:
`include` maps from an include, `symbol` maps from a symbol, `ref` names a file to
read. Mine maps from the implementation file, so `associated_impl` would be the
consistent name, and the pragma and the directive would then use the same word for
the same relation. The opposite choice does not work: `// IWYU pragma:
associated_header` on `#include "foo.icc"` would read as "foo.icc is my associated
header", which is what `IWYU pragma: associated` already means. I have not renamed
anything yet.

I would still like to keep the directive next to the pragma, because they answer
different questions. The pragma marks one file. The directive states a naming rule
that a whole repository already follows:

| project | files following the convention | pragmas needed | mapping rules needed |
|---|---|---|---|
| [ITK](https://github.com/InsightSoftwareConsortium/ITK/blob/main/Modules/Core/Common/include/itkImage.h) | 1313 `.hxx` | 1313 | 1 |
| [OpenFOAM](https://github.com/OpenFOAM/OpenFOAM-dev/blob/master/src/OpenFOAM/primitives/Vector/Vector.H) | 494 `*I.H` | 494 | 1 |
| [libstdc++](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/basic_ios.h) | 10 `.tcc` | 10 | 1 |

Editing 1313 headers, to say what the project already does everywhere, is a lot to
ask.

About mapping files being mainly for external libraries: IWYU already lets the same
relation be written both ways. `// IWYU pragma: private, include "public.h"` in a
header, and `{ "include": ["<bits/unique_ptr.h>", "private", "<memory>", "public"] }`
in a `.imp` file, say the same thing. The docs also say such mappings are "usually"
needed for third-party libraries, and they mention project-specific mappings.

A mapping rule also stays with the code. The `.imp` file is checked into the repo,
next to the headers it describes, and versioned with them. Only the
`--mapping_file` flag has to be added to the build, once, not per file. That flag
is the one real advantage of the pragma: it still works when someone runs IWYU by
hand.

So I think we want both.

---

Thanks,
Eric

bolshakov-a:
> About mapping files being mainly for external libraries: IWYU already lets the same
relation be written both ways.

Currently, writing a mapping file is the main way to denote a header as belonging to an external library. It is important (at least) for template "provision policy". Consider the following:
```cpp
// internal_fn.h
template <typename T>
void InternalFn(T) {}

// intermediate.h
#include "internal_fn.h"

// private.h
// IWYU pragma: private, include "public.h"
#include "intermediate.h"

template <typename T>
void TplFn(T t) {
    InternalFn(t);
}

// public.h
#include "private.h"

// user.cpp
#include "public.h"

void UseTplFn() {
    TplFn(1);
}
```
Note that `InternalFn` definition is included into `private.h` not directly but through the `intermediate.h` header. If you write your own code in line with the IWYU principle, you'll hardly do that. It would rather mean that `InternalFn` is a customization point that should be provided explicitly by `TplFn` users. So, IWYU outputs:
```console
$ include-what-you-use -I . user.cpp 

#include "internal_fn.h"  // for InternalFn

#include "internal_fn.h"  // for InternalFn
#include "public.h"       // for TplFn
```
But if all the headers are part of some external library written ignoring the IWYU principle, `InternalFn` is rather an implementation detail that should not be exposed to users. Hence, when running with the following mapping file:
```yaml
[
{"include": ["\"private.h\"", "private", "\"public.h\"", "public"]}
]
```
the IWYU output is different:
```console
$ include-what-you-use -I . -Xiwyu --mapping_file=map.imp user.cpp 

(user.cpp has correct #includes/fwd-decls)
```
> `associated_impl` would be the consistent name

I agree, but let's wait for @kimgr.

ericch1:
@bolshakov-a , I worked with Claude based on your reply and I propose something new, here is the whole answer Claude has written:

---
You are right about mapping files, and my earlier sentence was wrong: a public
file provides everything it includes transitively, and a mapping file is the
only way to mark one public. But that meaning lives in the visibility field,
and this directive has none — it never reaches `IsPublic()`. I ran your
example: swapping the `include` mapping for an `associated_header` one leaves
`user.cpp` still needing `"internal_fn.h"`.

Testing that turned up a real gap, though. On its own the directive reports
nothing. A file is only reported on if it reaches
`AddGlobToReportIWYUViolationsFor()`, which today happens through
`GetCanonicalName()` matching the main file's stem. So `foo.icc` works without
`--check_also` only because of the last commit, and a convention declared in a
mapping file, such as `foo.inline`, never does.

The fix is four lines next to the existing
`AddGlobToReportIWYUViolationsFor(includer_path)`: also report on the
implementation file when its associated header belongs to the main compilation
unit. The directive then stands on its own, and the last commit goes back to
being only a list of extensions. Full suite green, and a `.inline` reached
through an unrelated header is still left alone.

On the name: I agree with `associated_impl`, and I am happy to wait for @kimgr

---

Hope this help,
Eric

PR #2118, opened 2026-09-19
Fix iwyu crash

tpadioleau:
This follows the spirit of https://github.com/include-what-you-use/include-what-you-use/pull/2055.


tpadioleau:
Here is a small reproducer for the change l1604:
```cpp
#include <cstddef>
#include <tuple>
#include <type_traits>

class Foo;

template <>
struct std::tuple_size<Foo> : std::integral_constant<std::size_t, 1>
{
};
```

tpadioleau:
cc @bolshakov-a

iwyu.cc
```diff
@@ -1601,7 +1601,7 @@ class IwyuBaseAstVisitor : public BaseAstVisitor<Derived> {
 
     // Sometimes a shadow decl comes between us and the 'real' decl.
     const UsingDecl* using_decl = nullptr;
-    if (const auto* shadow_decl = dyn_cast<UsingShadowDecl>(used_decl)) {
+    if (const auto* shadow_decl = dyn_cast_or_null<UsingShadowDecl>(used_decl)) {
```
bolshakov-a, line 1604: Hmm, is it Ok that `nullptr` ever reaches `ReportDeclForwardDeclareUse`? CC @kimgr

bolshakov-a: Heh, @kimgr, [your fears](https://github.com/include-what-you-use/include-what-you-use/pull/2097#discussion_r3768943754) were not in vain! @tpadioleau, thanks for finding this! So yeah, either `decl->getPreviousDecl()` should be checked for `nullptr` in `VisitTagDecl`, or `ClassTemplateSpecializationDecl`s should be filtered out from there (forward-declarations of their primary templates are [reported](https://github.com/include-what-you-use/include-what-you-use/blob/84b0e86f4d0b959ee20a3be59edf7e1ec207e98a/iwyu.cc#L5187) from `VisitClassTemplateSpecializationDecl`). A simpler reproducer for testing:
```cpp
namespace ns {
template <int>
class Tpl;
}

template <>
class ns::Tpl<1> {};
```

tpadioleau: Glad if I can help. I let you decide on how to proceed. I honestly don't fully understand the consequences of my changes. I only know that it allowed me to use the branch `clang_23` on my project.

bolshakov-a: I've opened #2119.

tpadioleau: I remove this part of the current PR ?

bolshakov-a: Yes, it should not be needed after #2119. I also think that the code and the test changes in this PR should be squashed into a single commit.

tpadioleau: Done

